PR #399, opened 2023-05-05
[router] Fix initialization bug in router quota throttler

majisourav99:
## [router] Fix initialization bug in router quota throttler

There is a bug in how `ReadRequestThrottler` initialized store throttlers in the constructor. 
 `buildAllStoreReadThrottlers` relies on various config values , but they were initialized after the method was called. So after a router bounce it will initialize the store throttlers with buffer values of `0` but inside store update handler it will rebuild the throttler with proper value of the configs. This leads to diverted per store router quota in the same cluster depending on the state of start time of various hosts and store version changes.
This PR fixes the order of initialization and also removed code to update `quotaRebalanceEnabled` which is never called in router code.

## How was this PR tested?

CI unit tests.

### DIFF
--- a/services/venice-router/src/main/java/com/linkedin/venice/router/throttle/ReadRequestThrottler.java
+++ b/services/venice-router/src/main/java/com/linkedin/venice/router/throttle/ReadRequestThrottler.java
@@ -110,13 +110,13 @@ public class ReadRequestThrottler implements RouterThrottler, RoutersClusterMana
     this.zkRoutersManager.subscribeRouterCountChangedEvent(this);
     this.storeRepository.registerStoreDataChangedListener(this);
     this.stats = stats;
-    this.idealTotalQuotaPerRouter = calculateIdealTotalQuotaPerRouter();
     this.maxRouterReadCapacity = maxRouterReadCapacity;
     this.perStorageNodeReadQuotaBuffer = perStorageNodeReadQuotaBuffer;
     this.perStorageNodeThrottlerEnabled = perStorageNodeThrottlerEnabled;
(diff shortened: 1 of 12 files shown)


services/venice-router/src/main/java/com/linkedin/venice/router/throttle/ReadRequestThrottler.java
```diff
@@ -110,13 +110,13 @@ public ReadRequestThrottler(
...
     this.stats = stats;
-    this.idealTotalQuotaPerRouter = calculateIdealTotalQuotaPerRouter();
     this.maxRouterReadCapacity = maxRouterReadCapacity;
     this.perStorageNodeReadQuotaBuffer = perStorageNodeReadQuotaBuffer;
     this.perStorageNodeThrottlerEnabled = perStorageNodeThrottlerEnabled;
-    this.storesThrottlers = new AtomicReference<>(buildAllStoreReadThrottlers());
     this.lastRouterCount = zkRoutersManager.getExpectedRoutersCount();
     this.perStoreRouterQuotaBuffer = perStoreRouterQuotaBuffer;
+    this.idealTotalQuotaPerRouter = calculateIdealTotalQuotaPerRouter();
+    this.storesThrottlers = new AtomicReference<>(buildAllStoreReadThrottlers());
```
FelixGV, line 119: So this here is the fix, right?

majisourav99: right this is the fix.

FelixGV (approved):
Thanks Sourav!

PR #400, opened 2023-05-05
[server] Restart meta store writer when hitting produce exception

sixpluszero:
## [server] Restart meta store writer when hitting produce exception
This PR tries to enhance the produce logic in meta store writer. When VW#sendMessage throws exception that might be result from for example stale DNS information or failed internal state (like finalized checksum issue) due to bug, we might want to retry by restarting a new VW and clean internal state so it might be able to have healthy producer and internal state to produce the message. If it fails all retry we should log the error message.

## How was this PR tested?

Add new unit test to cover the retry logic.

### DIFF
--- a/internal/venice-common/src/main/java/com/linkedin/venice/system/store/MetaStoreWriter.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/system/store/MetaStoreWriter.java
@@ -400,12 +400,12 @@ public class MetaStoreWriter implements Closeable {
         messageProduceRetryCount++;
         if (messageProduceRetryCount < maxMessageProduceRetryCount) {
           LOGGER.warn(
-              "Caught exception while trying to write message, will retry {}/{}",
+              "Caught exception while trying to write message, will restart Venice Writer and retry {}/{}",
               messageProduceRetryCount,
               maxMessageProduceRetryCount);
-        } else {
-          LOGGER.error("Caught exception while trying to write message, will shutdown writer.", e);
           closeVeniceWriter(metaStoreName, writer, true);
+        } else {
+          LOGGER.error("Fail to write message after {} attempts.", maxMessageProduceRetryCount, e);
           break;
         }
       } finally {
--- a/internal/venice-common/src/main/java/com/linkedin/venice/system/store/MetaStoreWriter.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/system/store/MetaStoreWriter.java
@@ -40,6 +40,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.locks.ReentrantLock;
+import java.util.function.Consumer;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
 import org.apache.avro.Schema;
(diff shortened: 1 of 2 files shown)


majisourav99:
can we create a test capturing this?

internal/venice-common/src/main/java/com/linkedin/venice/system/store/MetaStoreWriter.java
```diff
@@ -371,16 +372,31 @@ private void update(
...
+      writer.update(
+          key,
+          update,
+          AvroProtocolDefinition.METADATA_SYSTEM_SCHEMA_STORE.currentProtocolVersion.get(),
+          derivedComputeSchemaId,
+          null);
+      writer.flush();
+    } catch (Exception e) {
+      LOGGER.error("Caught exception while trying to update, will respawn a new writer.", e);
+      closeVeniceWriter(metaStoreName, writer, true);
```
majisourav99, line 389: writer could be null here ?

sixpluszero: getOrCreateMetaStoreWriter should always return a VW is my understanding here.

internal/venice-common/src/main/java/com/linkedin/venice/system/store/MetaStoreWriter.java
```diff
@@ -371,16 +373,52 @@ private void update(
...
+        writer.flush();
+        messageProduced = true;
+      } catch (Exception e) {
+        messageProduceRetryCount++;
+        if (messageProduceRetryCount < maxMessageProduceRetryCount) {
+          LOGGER.warn(
+              "Caught exception while trying to write message, will restart Venice Writer and retry {}/{}",
+              messageProduceRetryCount,
+              maxMessageProduceRetryCount);
+          closeVeniceWriter(metaStoreName, writer, true);
```
gaojieliu, line 406: Is it possible that `closeVeniceWriter` could throw another exception?
Can we catch the potential exception to avoid breaking the loop?

sixpluszero: Hi Gaojie, I don't think there is any exception to be thrown here, the closeVeniceWrtier with skipTopicCheck = true means it will just call vw.close() with graceful close = true.
Inside VW.close() it exception is caught and handled with WARN logging.

gaojieliu: This is mainly for safety purpose.
There is no guarantee that function won't throw exception in the future, right?

internal/venice-common/src/main/java/com/linkedin/venice/system/store/MetaStoreWriter.java
```diff
@@ -371,16 +372,61 @@ private void update(
...
+
+  void writeMessageWithRetry(String metaStoreName, Consumer<VeniceWriter> writerConsumer) {
+    ReentrantLock lock = getOrCreateMetaStoreWriterLock(metaStoreName);
+    int messageProduceRetryCount = 0;
+    int maxMessageProduceRetryCount = 3;
+    VeniceWriter writer = null;
+    boolean messageProduced = false;
+    while (!messageProduced) {
+      try {
+        lock.lock();
```
gaojieliu, line 393: Not a big issue, just try to align with the existing code style.
Can you move this line up right above `try`?

gaojieliu (approved):

internal/venice-common/src/main/java/com/linkedin/venice/system/store/MetaStoreWriter.java
sixpluszero: Sure, just updated.

PR #401, opened 2023-05-07
[WIP][samza] Make a generic module for Venice nearline producers and remove tight coupling on Samza

nisargthakkar:
## Make a generic module for Venice nearline producers and remove tight coupling on Samza

Currently, Venice only supports Samza as a stream processing framework. This PR adds a generic `venice-nearline-producer` framework that can be used by any stream processing framework to write to Venice.

## How was this PR tested?

Modified all tests to use this new framework. Tests in `VeniceSystemFactoryTest` still use VeniceSystemFactory and `VeniceSystemProducer` to validate the Samza integration.

Internal CI passed. #7980


internal/venice-nearline-producer/build.gradle
adamxchen, line 1: Shouldn't we add this under `client` , since `venice-samza` was there? But I guess that might be fine too since `venice-client-common` in the `internal`. Maybe we can rename it to `venice-nearline-producer-common`?

adamxchen (commented):
Thanks for the change! Would like to get this merged sooner since we have `VenicePulsarSink` now. What's left for this PR?

nisargthakkar:
Will reopen after resolving conflicts

PR #402, opened 2023-05-08
[server] Additional safeguard to make sure server will always be able to shutdown successfully with ingestion isolation

sixpluszero:
## [server] Additional safeguard to make sure server will always be able to shutdown successfully with ingestion isolation

With all the recent bug fixes, it is unlikely that main process and isolation process will run out of sync for a particular resource's metadata. However, when real bug happens, current infinite retry logic might lead to blocking Venice server from shutting down, which is not acceptable as it needs manual shutdown for all impacted hosts in the cluster.
This PR adds an additional safeguard to guard this behavior. When HelixParticipantService is shutting down, we will inject a boolean into IsolatedIngestionBackend notifying it is preparing to shutdown. When Helix manager is sending stop consumption command to all the resources, if a resource has out-of-sync metadata and reject remote execution, we will skip the retry.

## How was this PR tested?

Added new unit test to test the behavior. 

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/IsolatedIngestionBackend.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/IsolatedIngestionBackend.java
@@ -58,6 +58,7 @@ public class IsolatedIngestionBackend extends DefaultIngestionBackend
   private final VeniceConfigLoader configLoader;
   private final ExecutorService completionReportHandlingExecutor = Executors.newFixedThreadPool(10);
   private Process isolatedIngestionServiceProcess;
+  private boolean isShuttingDown = false;
 
   public IsolatedIngestionBackend(
       VeniceConfigLoader configLoader,
@@ -113,7 +114,7 @@ public class IsolatedIngestionBackend extends DefaultIngestionBackend
         topicName,
         partition,
         STOP_CONSUMPTION,
-        () -> mainIngestionRequestClient.stopConsumption(storeConfig.getStoreVersionName(), partition),
+        () -> getMainIngestionRequestClient().stopConsumption(storeConfig.getStoreVersionName(), partition),
         () -> super.stopConsumption(storeConfig, partition));
   }
 
@@ -228,6 +229,11 @@ public class IsolatedIngestionBackend extends DefaultIngestionBackend
     }
   }
 
+  @Override
+  public void prepareForShutdown() {
+    isShuttingDown = true;
+  }
+
   public void setIsolatedIngestionServiceProcess(Process process) {
     isolatedIngestionServiceProcess = process;
   }
@@ -242,6 +248,10 @@ public class IsolatedIngestionBackend extends DefaultIngestionBackend
 
   public MainIngestionRequestClient getMainIngestionRequestClient() {
     return mainIngestionRequestClient;
+  }
+
+  public boolean isShuttingDown() {
+    return isShuttingDown;
   }
 
   void removeTopicPartitionLocally(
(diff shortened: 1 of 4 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/IsolatedIngestionBackend.java
```diff
@@ -58,6 +58,7 @@ public class IsolatedIngestionBackend extends DefaultIngestionBackend
   private final VeniceConfigLoader configLoader;
   private final ExecutorService completionReportHandlingExecutor = Executors.newFixedThreadPool(10);
   private Process isolatedIngestionServiceProcess;
+  private boolean isShuttingDown = false;
```
lluwm, line 61: Does it have concurrent accesses? Feels like we might want to use an AtomicBoolean if it happens to.

clients/da-vinci-client/src/test/java/com/linkedin/davinci/ingestion/IsolatedIngestionBackendTest.java
```diff
@@ -258,4 +258,36 @@ public void testBackendCanMaintainMetadataCorrectlyForDroppingPartition() {
...
+      doCallRealMethod().when(backend).stopConsumption(any(), anyInt());
+      doCallRealMethod().when(backend).executeCommandWithRetry(anyString(), anyInt(), any(), any(), any());
+      ReentrantReadWriteLock readWriteLock = new ReentrantReadWriteLock();
+      when(monitorService.getForkProcessActionLock()).thenReturn(readWriteLock);
+      KafkaStoreIngestionService storeIngestionService = mock(KafkaStoreIngestionService.class);
+      when(storeIngestionService.isPartitionConsuming(anyString(), anyInt())).thenReturn(false);
+      when(backend.getStoreIngestionService()).thenReturn(storeIngestionService);
+      when(ingestionRequestClient.stopConsumption(anyString(), anyInt())).thenReturn(false);
+      when(backend.isShuttingDown()).thenReturn(true);
+      backend.stopConsumption(storeVersionConfig, partition);
```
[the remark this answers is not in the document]
sixpluszero: Sure, will do

clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/IsolatedIngestionBackend.java
sixpluszero: No, the only call is from HelixParticipantService for once during stop(). Each server has only 1 instance of this service class. If we have concurrent access in the future(which I don't think would be true for now) we can evolve to use Atomic I think.

lluwm: I see, but isn't `executeCommandWithRetry` which reads it this value can run from a thread pool, or maybe I misunderstood the way how II works

sixpluszero: My understanding is: During the shutdown, only HelixParticipantService will modify its value. All Helix threads will read it if it hits this condition (conflict metadata for two processes). Your concern of race condition should not be the case as even if it is missing the update for this round of check it will be caught in the next round of infinite loop and it will break the loop.

lluwm: Thanks for the discussion and updates, Jialin!

lluwm (approved):
LGTM!

PR #403, opened 2023-05-08
[server][da-vinci] Remove hot path logging in ingestion task

nisargthakkar:
## Remove hot path logging in ingestion task

I noticed up to 450k logs in a single day on one host that got logged by this. I don't think this adds much value

## How was this PR tested?

N/A

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
@@ -2951,7 +2951,6 @@ public abstract class StoreIngestionTask implements Runnable, Closeable {
       elapsedTime = System.currentTimeMillis() - startTime;
 
       if (state != null) {
-        LOGGER.info("Version state is available for {} after {} ms", kafkaTopic, elapsedTime);
         return state;
       }
 


clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -2951,7 +2951,6 @@ protected StoreVersionState waitVersionStateAvailable(String kafkaTopic) throws
       elapsedTime = System.currentTimeMillis() - startTime;
 
       if (state != null) {
-        LOGGER.info("Version state is available for {} after {} ms", kafkaTopic, elapsedTime);
```
sixpluszero, line 2954: Looks like it will be called only in chunking store. Do you think it is worth using redundant filter or it is completely not worth logging here?

sixpluszero: Actually I think we can reduce the logging in another way, we call the getVersionState() first one time, if it is null then we enter the for loop and thus the logging will only be once per SIT...but maybe it is fine just to remove it completely.

majisourav99: I vote to remove it completely.

nisargthakkar: Yeah, if having the log doesn't help, I also would like to remove it. If there are issues, it will be captured by the error log eventually

PR #406, opened 2023-05-09
[producer] Add an implementation for online producer

nisargthakkar:
## Add an implementation for online writer

This PR adds an implementation for the `VeniceProducer` interface. The implementation allows for online applications to write to Venice

## How was this PR tested?

GH CI


clients/venice-producer/src/main/java/com/linkedin/venice/producer/VeniceProducerMetrics.java
```diff
@@ -0,0 +1,85 @@
...
+
+    if (metricsRepository != null) {
+      enableMetrics = true;
+      requestSensor = registerSensor("write_request", new OccurrenceRate());
+      putRequestSensor = registerSensor("put_request", new OccurrenceRate());
+      deleteRequestSensor = registerSensor("delete_request", new OccurrenceRate());
+      updateRequestSensor = registerSensor("update_request", new OccurrenceRate());
+
+      successRequestSensor = registerSensor("success_write_request", new OccurrenceRate());
+      failedRequestSensor = registerSensor("failed_write_request", new OccurrenceRate());
```
FelixGV, line 34: Hmm... I would prefer avoiding the term "request" as it seems to imply a meaning of transactionality. How about replacing "request" with "operation" or with "event"?

clients/venice-producer/src/main/java/com/linkedin/venice/producer/VeniceProducerMetrics.java
```diff
@@ -0,0 +1,85 @@
...
+      produceLatencySensor = registerSensor("produce_latency", new Avg(), new Max());
+
+      pendingRequestSensor = registerSensor("pending_write_request", new Total());
```
FelixGV, line 37: Hmm... I wonder if it's ideal to have a `Total` here... I see that you're incrementing and decrementing it in the functions below, so it'll be accurate at any point in time, but as far as the metric is concerned, it will take just a single snapshot per minute, which could miss a lot of the action in-between data points. An alternative might be to define this metric as, let's say `Avg` and `Max`, and then outside of the metric, we'd keep an `AtomicInteger` and then call `pendingRequestSensor.record(pendingRequestAtomicCounter.incrementAndGet());` so that the metric gives us a sense of the range of values.

clients/venice-producer/src/main/java/com/linkedin/venice/producer/VeniceProducerMetrics.java
```diff
@@ -0,0 +1,85 @@
...
+    if (metricsRepository != null) {
+      enableMetrics = true;
+      requestSensor = registerSensor("write_request", new OccurrenceRate());
+      putRequestSensor = registerSensor("put_request", new OccurrenceRate());
+      deleteRequestSensor = registerSensor("delete_request", new OccurrenceRate());
+      updateRequestSensor = registerSensor("update_request", new OccurrenceRate());
+
+      successRequestSensor = registerSensor("success_write_request", new OccurrenceRate());
+      failedRequestSensor = registerSensor("failed_write_request", new OccurrenceRate());
+      produceLatencySensor = registerSensor("produce_latency", new Avg(), new Max());
```
FelixGV, line 35: This is probably going to be the most scrutinized metric, and I would like to avoid confusion or misinterpretation. Can we perhaps call it `produce_to_durable_buffer_latency` or something like that? Also, given that it's a latency, I think we should include p50/p95/p99.

internal/venice-client-common/src/test/java/com/linkedin/venice/stats/ThreadPoolStatsTest.java
```diff
@@ -1,6 +1,5 @@
 package com.linkedin.venice.stats;
 
-import com.linkedin.davinci.stats.ThreadPoolStats;
```
FelixGV, line 3: Home at last :D ...

internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
```diff
@@ -1868,4 +1868,15 @@ private ConfigKeys() {
    * Config to control the queue capacity for the thread pool executor used for ssl handshake in servers.
    */
   public static final String SERVER_SSL_HANDSHAKE_QUEUE_CAPACITY = "server.ssl.handshake.queue.capacity";
+
+  /**
+   * Number of threads for online Venice producer. This controls how many messages can be processed at a time
```
FelixGV, line 1873: What is the implication of this? Can the config description be a bit more detailed?

internal/venice-test-common/src/integrationtest/java/com/linkedin/venice/endToEnd/TestHybrid.java
```diff
@@ -1065,10 +1068,23 @@ public void testHybridMultipleVersions() throws Exception {
       }
       producer.stop();
 
+      Properties onlineProducerProps = new Properties();
+      try (VeniceProducer veniceOnlineProducer = OnlineProducerFactory.createProducer(
+          ClientConfig.defaultGenericClientConfig(storeName).setVeniceURL(cluster.getRandomRouterURL()),
+          new VeniceProperties(onlineProducerProps),
```
FelixGV, line 1074: I think we do this kind of stuff fairly often. Maybe we should have a static helped function so we can just do `VeniceProperties.empty()`...

clients/venice-producer/src/main/java/com/linkedin/venice/producer/AbstractVeniceProducer.java
```diff
@@ -0,0 +1,445 @@
...
+      };
+
+      byte[] keyBytes = keySerializer.serialize(key);
+      byte[] valueBytes = getSerializer(valueSchema).serialize(value);
+
+      if (logicalTime < 0) {
+        veniceWriter.put(keyBytes, valueBytes, valueSchemaId, callback);
+      } else {
+        veniceWriter.put(keyBytes, valueBytes, valueSchemaId, logicalTime, callback);
+      }
```
FelixGV, line 241: Can we simplify this and always use the API with `logicalTime` since we pass `APP_DEFAULT_LOGICAL_TS` anyway, which is the same default that `VeniceWriter` would use?

clients/venice-producer/src/main/java/com/linkedin/venice/producer/AbstractVeniceProducer.java
```diff
@@ -0,0 +1,445 @@
...
+  @Override
+  public CompletableFuture<DurableWrite> asyncPut(long logicalTime, K key, V value) {
+    String error = validateProducerWithLogicalTime(logicalTime);
+    if (!StringUtils.isEmpty(error)) {
+      return getFutureCompletedExceptionally(error);
+    }
+    return asyncPutInternal(logicalTime, key, value);
+  }
+
+  private CompletableFuture<DurableWrite> asyncPutInternal(long logicalTime, K key, V value) {
```
FelixGV, line 198: So IIUC, the difference between `asyncPut(K, V)` and `asyncPut(time, K, V)` is just that `validateProducerWithLogicalTime` does an extra check compared to `validateProducer`. I don't think it's worth tying ourselves into knots over this extra check. We could make the code simpler IMHO if `asyncPut(time, K, V)` contains the full check as well as the full content of `asyncPutInternal`, and `asyncPut(K, V)` just delegates to `asyncPut(APP_DEFAULT_LOGICAL_TS, K, V)`. We would do an extra boolean check on the time-free API as compared to the currently proposed implementation, but I don't think that's a big deal in this context. The online apps using this shouldn't be churning out millions of writes per sec per node, so these minor details probably don't matter here.

If you agree, then the same refactoring could apply to the update and delete functions as well.

clients/venice-producer/src/main/java/com/linkedin/venice/producer/AbstractVeniceProducer.java
```diff
@@ -0,0 +1,445 @@
...
+    return asyncUpdateInternal(logicalTime, key, updateFunction);
+  }
+
+  private CompletableFuture<DurableWrite> asyncUpdateInternal(
+      long logicalTime,
+      K key,
+      Consumer<UpdateBuilder> updateFunction) {
+    producerMetrics.recordUpdateRequest();
+    return CompletableFuture.supplyAsync(() -> {
+      // Caching to avoid race conditions during processing of the function
```
FelixGV, line 329: Good call.

FelixGV (commented):
Thanks Nisarg, this looks really good. I left a few minor comments, but I think this could be merged pretty soon.

internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
nisargthakkar: This sets the number of threads of `AbstractVeniceProducer#producerExecutor`. It controls how many concurrent write operations can be present. Right now, the default value is set to 10, but I was planning on tuning it through some experimentation

clients/venice-producer/src/main/java/com/linkedin/venice/producer/VeniceProducerMetrics.java
nisargthakkar: Makes sense. Given a uniform load, it is better to get a statistic on the whole window rather than a point-in-time value

clients/venice-producer/src/main/java/com/linkedin/venice/producer/AbstractVeniceProducer.java
nisargthakkar: Yeah, that sounds okay. I did it this way to order the error messages to check producer state then validate the logical_ts. But I don't mind simplifying code and change the order of checks

nisargthakkar: The issue is that `asyncPut(K, V)` calls `asyncPut(APP_DEFAULT_LOGICAL_TS, K, V)`. Since `APP_DEFAULT_LOGICAL_TS` is = -2, it will end up failing the if check. So, I decided to add the `asyncPutInternal` without any checks.

I can move the `validateProducer` call inside of the `asyncPutInternal` function and leave the `logicalTs` validation in the APIs with `logicalTs`. Does this work?

nisargthakkar:
```java
@Override
public CompletableFuture<DurableWrite> asyncPut(K key, V value) {
  return asyncPutInternal(APP_DEFAULT_LOGICAL_TS, key, value);
}

@Override
public CompletableFuture<DurableWrite> asyncPut(long logicalTime, K key, V value) {
  if (logicalTime < 0) {
    return getFutureCompletedExceptionally("Logical time must be a non-negative value. Got: " + logicalTime);
  }

  return asyncPutInternal(logicalTime, key, value);
}

private CompletableFuture<DurableWrite> asyncPutInternal(long logicalTime, K key, V value) {
  String error = validateProducer();
  if (!StringUtils.isEmpty(error)) {
    return getFutureCompletedExceptionally(error);
  }
  ...
}

```

clients/venice-producer/src/main/java/com/linkedin/venice/producer/AbstractVeniceProducer.java
```diff
@@ -0,0 +1,445 @@
...
+      } catch (Exception e) {
+        producerMetrics.recordFailedRequest();
+        throw e;
+      }
+      // Might block
+      int valueSchemaId;
+      try {
+        valueSchemaId = schemaReader.getValueSchemaId(valueSchema);
+      } catch (Exception e) {
+        valueSchemaId = SchemaData.INVALID_VALUE_SCHEMA_ID;
```
xunyin8, line 213: Are we just going to swallow this exception? I understand that we want to keep the amount of stuff we log minimal as a client library but perhaps we can afford to put the underlying exception name in the VeniceException message that we throw later in case the error persists and we need it for debugging.

clients/venice-producer/src/main/java/com/linkedin/venice/producer/online/OnlineVeniceProducer.java
```diff
@@ -0,0 +1,120 @@
...
+      InternalAvroStoreClient<K, V> storeClient,
+      SchemaReader kmeSchemaReader,
+      VeniceProperties producerConfigs,
+      MetricsRepository metricsRepository,
+      ICProvider icProvider) {
+    this.storeClient = storeClient;
+    this.storeName = storeClient.getStoreName();
+    this.icProvider = icProvider;
+
+    Duration schemaRefreshPeriod;
```
xunyin8, line 54: Do we want the schema reader on the producer to auto refresh? Does this mean registering a bad user schema could potentially break things if online writers pick it up automatically?

clients/venice-producer/src/main/java/com/linkedin/venice/producer/AbstractVeniceProducer.java
nisargthakkar: Makes sense. Will add this information

clients/venice-producer/src/main/java/com/linkedin/venice/producer/online/OnlineVeniceProducer.java
nisargthakkar: In online producers, `SchemaReader` is used to fetch the schema ids for data that a user has provided. We want to reduce the number of times that a request to write data has to wait since these are online applications.

Since the SchemaReader object is owned by the producer, this will not leak to other parts of the code that use `SchemaReader`

xunyin8: Right, perhaps I should rephrase the question. In the online producers world, do we want users to pick the writer schema similar to the offline and stream world today or always use the latest schema. The later will come with a risk which is bad user schema could break online write instantly and doesn't really give user a chance to canary writes using the new schema.

nisargthakkar: We want users to provide whichever writer schema they want like Samza jobs. For partial updates, we need to pick the latest schema (and the primary reason why we need the schema refresh). This will be a superset schema since superset schemas are always enabled for WC stores

xunyin8: Got it, thanks for the explanation.

clients/venice-producer/src/main/java/com/linkedin/venice/producer/AbstractVeniceProducer.java
```diff
@@ -0,0 +1,425 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+
+/**
+ * A generic implementation of the {@link VeniceProducer} interface
+ *
+ * @see VeniceProducer
+ */
+public abstract class AbstractVeniceProducer<K, V> implements VeniceProducer<K, V> {
```
adamxchen, line 57: Does this PR have to wait for https://github.com/linkedin/venice/pull/401?

nisargthakkar: No. PR #401 is waiting on this

clients/venice-producer/src/main/java/com/linkedin/venice/producer/VeniceProducerMetrics.java
```diff
@@ -0,0 +1,92 @@
...
+  private Sensor pendingOperationSensor = null;
+
+  private final AtomicInteger pendingOperationCounter = new AtomicInteger(0);
+
+  public VeniceProducerMetrics(MetricsRepository metricsRepository, String storeName) {
+    super(metricsRepository, storeName);
+
+    if (metricsRepository != null) {
+      enableMetrics = true;
+      operationSensor = registerSensor("write_operation", new OccurrenceRate());
```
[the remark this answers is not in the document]
nisargthakkar: `write` is expected to be a sum of `put`, `update`, `delete`. I don't see a difference b/w `write` and `produce`

clients/venice-producer/src/main/java/com/linkedin/venice/producer/AbstractVeniceProducer.java
```diff
@@ -0,0 +1,425 @@
...
+  private RecordSerializer<Object> keySerializer;
+  private static final Schema STRING_SCHEMA = Schema.create(Schema.Type.STRING);
+  private static final Schema INT_SCHEMA = Schema.create(Schema.Type.INT);
+  private static final Schema LONG_SCHEMA = Schema.create(Schema.Type.LONG);
+  private static final Schema FLOAT_SCHEMA = Schema.create(Schema.Type.FLOAT);
+  private static final Schema DOUBLE_SCHEMA = Schema.create(Schema.Type.DOUBLE);
+  private static final Schema BYTES_SCHEMA = Schema.create(Schema.Type.BYTES);
+  private static final Schema BOOL_SCHEMA = Schema.create(Schema.Type.BOOLEAN);
+
+  protected void configure(
```
majisourav99, line 79: can we validate the store for if its write enabled or not ?

nisargthakkar: Sure. I can make that change on the router. I will raise a separate PR

clients/venice-producer/src/main/java/com/linkedin/venice/producer/VeniceProducerMetrics.java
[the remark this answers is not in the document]
majisourav99: I was thinking `write` == `put` before I saw the code.

clients/venice-producer/src/main/java/com/linkedin/venice/producer/AbstractVeniceProducer.java
```diff
@@ -0,0 +1,425 @@
...
+      }
+      final CompletableFuture<Void> completableFuture = new CompletableFuture<>();
+      final Instant sendStartTime = Instant.now();
+      final PubSubProducerCallback callback = (PubSubProduceResult produceResult, Exception exception) -> {
+        Duration sendDuration = Duration.between(sendStartTime, Instant.now());
+        if (exception == null) {
+          producerMetrics.recordSuccessfulRequestWithLatency(sendDuration.toMillis());
+          completableFuture.complete(null);
+        } else {
+          producerMetrics.recordFailedRequest();
```
majisourav99, line 233: can we add a log here?

nisargthakkar: Yeah, I can. Do you think this will flood the logs? I guess it's okay, though

clients/venice-producer/src/main/java/com/linkedin/venice/producer/AbstractVeniceProducer.java
nisargthakkar: https://github.com/linkedin/venice/pull/422

clients/venice-producer/src/main/java/com/linkedin/venice/producer/AbstractVeniceProducer.java
```diff
@@ -0,0 +1,428 @@
...
+      return getFutureCompletedExceptionally("Logical time must be a non-negative value. Got: " + logicalTime);
+    }
+
+    return asyncPutInternal(logicalTime, key, value);
+  }
+
+  private CompletableFuture<DurableWrite> asyncPutInternal(long logicalTime, K key, V value) {
+    String error = validateProducer();
+    if (!StringUtils.isEmpty(error)) {
+      return getFutureCompletedExceptionally(error);
```
majisourav99, line 197: Does it make sense to record metrics around this?

nisargthakkar: I don't think so. This should occur when the producer starts or is being shutdown (probably only during development). Users would also log the error on their end too when they encounter failed futures

clients/venice-producer/src/main/java/com/linkedin/venice/producer/VeniceProducerMetrics.java
```diff
@@ -0,0 +1,92 @@
...
+      updateOperationSensor = registerSensor("update_operation", new OccurrenceRate());
+
+      successOperationSensor = registerSensor("success_write_operation", new OccurrenceRate());
+      failedOperationSensor = registerSensor("failed_write_operation", new OccurrenceRate());
+      String produceLatencySensorName = "produce_to_durable_buffer_latency";
+      produceLatencySensor = registerSensor(
+          produceLatencySensorName,
+          TehutiUtils.getPercentileStat(getName() + AbstractVeniceStats.DELIMITER + produceLatencySensorName));
+
+      pendingOperationSensor = registerSensor("pending_write_operation", new Avg(), new Max(), new Gauge());
```
FelixGV, line 44: Do we need the `Gauge`? I don't think it's intended to be used in this way...

FelixGV (commented):
Looks good overall. Just a minor comment on the metrics.

clients/venice-producer/src/main/java/com/linkedin/venice/producer/VeniceProducerMetrics.java
nisargthakkar: We probably don't need it. I added it to test the metrics code. With just `Avg` and `Max`, the test cases won't make much sense.

I can do two things:
1. Remove that Gauge metric and the test lines for it
2. Add the Gauge only for tests. I'm not keen on doing this as it doesn't seem to add any value

nisargthakkar: The test cases I've written are like:
```
Assert.assertEquals(metricsRepository.getMetric(PENDING_OPERATION_METRIC_NAME).value(), 0.0);
```

These use the `Gauge` currently. The value for the `Avg` is 0.5 and for the `Max` it is 1. `Gauge` will give us the last measured metric value (obv. assuming that the test case doesn't run for too long that the metric value gets reset)

nisargthakkar: Replaced `Gauge` and `Avg` with `Min` since `Avg` is slightly tricky to reason about

FelixGV (approved):
Thanks Nisarg! Great work!

PR #409, opened 2023-05-10
[router] Added openssl support to Router http client and removed Kafka openssl support

gaojieliu:
This code change is leveraging `conscrypt-openjdk-uber` lib to support openssl in Http Client lib being used by Router. Once this is verified in Router, we will apply the similar optimization to fast-client.

New Router config:
router.http.client.openssl.enabled: default false

This code change also removed the Kafka openssl support, which was proved to be unstable.

Resolves #XXX

## How was this PR tested?

CI


FelixGV (approved):
Thanks Gaojie!

PR #411, opened 2023-05-10
[VIP-1] Authentication Service API

eolivelli:


docs/proposals/vip-1.md
```diff
@@ -0,0 +1,162 @@
...
+we need to support at least only mechanisms that work well on HTTP and that can be performed in a single step, 
+like basic HTTP Authentication, that needs only to use an HTTP Header
+or TLS based Authentication, that is based on the client certificate exchanged during the TLS handshake.
+
+
+## Scope
+
+1. **What is in scope?**
+
+   - Define an API to write plugins to perform authentication checks.
```
ZacAttack, line 41: Can we add some note (note necessarily here, but somewhere) that this is specific to authentication of Read/Push?

I don't think this would apply to things like nearline produce or davinci local ingestion since that authentication is done directly against the pubsub broker.

Speaking of the pubsub broker (rambling at this point) one thing that we've tried to have set up (but.... haven't been always successful) is trying to come up with some harmony between the authentication/authorization set up of Venice and the underlying pubsub broker.  The notion here is that the Venice controller is able to support API's which abstract what needs to be set up where.  Within LinkedIn, this looks like having the Venice controller set up ACL's on both Venice and Kafka, and both of those services talk to the same authentication service for fetching rules.  This is a very LinkedIn centric view, do you forsee any problems with this setup with what you have in mind with JWT based authenticaiton?

manujose0 (approved):

PR #412, opened 2023-05-10
[router] Various perf-related cleanups

FelixGV:
Removed lots of object allocation and string parsing from the router's hot path.

Venice changes:

- ResourceAssignment::getPartitionAssignment does 1 fewer map lookup. Called at least once per request.

- Partition now carries two extra EnumMap instances, which are faster than the previous String-keyed map and also pre-populates the results of ready instances and working instances. Previously, there were lots of hot path collection instantiation in these functions.

- PartitionAssignment now carries an extra array to do lookups faster than with the Integer-keyed map.

- VenicePartitionFinder improvements:
  - Added findPartitionNumber to get the partition for a key without string parsing.
  - Also made findPartitioner public so that it can be looked up once per multi-key request, rather than once per key.

- Removed the partitionNames set from KeyPartitionSet, which was only used for error logging.

- VeniceDelegateMode improvements:
  - getHelixGroupNum and getAssignedHelixGroupId don't use Optionals anymore.
  - ScatterGatherModeForMultiKeyRequest now reuses an ArrayList of ArrayLists when grouping keys by partitions, rather than a throwaway map of LinkedLists.

- Adopted Fast-Avro everywhere in the router.

- VenicePath improvements:
  - Removed the keyIdxToRouterKey map from VenicePath, which is dead code since the router cache days. This map was getting populated with every key of each request, needlessly.
  - Removed string parsing of resource names from the VenicePath.
  - Removed Optional wrapping of the chunkedResponse, responseDecompressor and body.

- VeniceChunkedResponse::write now does not synchronize every time.

Alpini changes:

- Removed Optionals from Headers. Also removed unused functions getAsDate, getAsInt, getAsLong.

- Removed IOUtils::getFilterInputStreamIn (and the RateLimitedStream that uses it) since it is blocking JDK17 and is not used by Venice.

- Added a more efficient overload of findPartitionNumber in PartitionFinder and AsyncPartitionFinder.

- Made ScatterGatherRequest use an empty set rather than a real set for the unused _partitionIdsToQuery property. Ideally, we would remove this entirely but this change is already big enough.


internal/venice-common/src/main/java/com/linkedin/venice/meta/Partition.java
```diff
@@ -30,10 +31,52 @@ public class Partition {
   private final int id;
 
   private final Map<String, List<Instance>> stateToInstancesMap;
+  private final Map<HelixState, List<Instance>> helixStateToInstancesMap;
```
majisourav99, line 34: Looks like we can get aways with just storing the Instance::getNodeID instead of the full instance objects. but savings probably not worth the refactoring.

FelixGV: Hmm... perhaps, but this `List<Instance>` ends up being used in a ton of places, so the cost of refactoring seems steep (if it is possible at all, not sure).

In terms of overhead, IDK if it changes anything. Either way it'll be a list of pointers to objects on the heap. Whether the pointers point to Strings or `Instance` objects doesn't matter, assuming those objects are going to be on the heap no matter what, due to being referenced from elsewhere.

majisourav99: yeah, Instances are already pre-allocated before Partition class creates List of those.

services/venice-router/src/main/java/com/linkedin/venice/router/api/VeniceDelegateMode.java
```diff
@@ -471,77 +471,108 @@ public <H, P extends ResourcePath<K>, K, R> Scatter<H, P, K> scatter(
...
           }
+          // Important to clear the inner list since it is thread-local state, which will be re-used by the next request
+          keysForCurrentPartition.clear();
+        }
+      } finally {
+        for (; currentPartition < partitionCount; currentPartition++) {
+          // This would only happen if the body of the try threw an exception.
+          // If that were to happen we would do a final cleanup here out of an abundance of caution.
+          keysForCurrentPartition = keysPerPartition.get(currentPartition);
+          keysForCurrentPartition.clear();
```
gaojieliu, line 567: So the `inner` list will become larger if there are some large requests or some hot partitions.
And here we have logic to clear the list belonging to every partition, so the total complexity will be O(inner_list_size * partition_count) for cleanup.
Not sure whether this is a concern or not, but we should be aware of such side effect.

internal/venice-common/src/main/java/com/linkedin/venice/meta/Partition.java
```diff
@@ -30,10 +31,52 @@ public class Partition {
...
     this.stateToInstancesMap = stateToInstancesMap;
+
+    // Populate enum maps which are more efficient than the string-based map
+    this.helixStateToInstancesMap = new EnumMap<>(HelixState.class);
+    this.executionStatusToInstancesMap = new EnumMap<>(ExecutionStatus.class);
+    for (Map.Entry<String, List<Instance>> entry: stateToInstancesMap.entrySet()) {
+      try {
+        HelixState helixState = HelixState.valueOf(entry.getKey());
+        this.helixStateToInstancesMap.put(helixState, Collections.unmodifiableList(entry.getValue()));
+      } catch (IllegalArgumentException e) {
```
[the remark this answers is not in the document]
FelixGV: The string-keyed map combines both enums in its key space. So if looking for a string that doesn't exist in one the enums, it will throw an IAE. I can leave a comment for that.

internal/venice-common/src/main/java/com/linkedin/venice/meta/PartitionAssignment.java
```diff
@@ -28,41 +22,38 @@ public PartitionAssignment(String topic, int numberOfPartition) {
           "Expected number of partition should be larger than 0 for resource '" + topic + "'. Current value:"
               + numberOfPartition);
     }
-    this.expectedNumberOfPartitions = numberOfPartition;
-    idToPartitionMap = new HashMap<>();
+    this.idToPartitionMap = new HashMap<>(numberOfPartition);
+    this.partitionsArrayIndexedById = new Partition[numberOfPartition];
```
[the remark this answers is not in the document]
FelixGV: It may be possible to use just the array but that would require a bigger refactoring. I canged only the hot paths so far...

services/venice-router/src/main/java/com/linkedin/venice/router/api/VeniceDelegateMode.java
FelixGV: Well `clear` only goes as far as the size of the list, even if its capacity is larger. So the time complexity is basically O(3*N) where N is the number of keys queried (once to populate lists, a second time to populate the treemaps, and a third time to clear). Having treemaps in the middle is kind of a PITA, maybe we could change that (with further refactorings in Alpini).

gaojieliu: I see. Essentially, it only clears the entries current request is using.

internal/venice-common/src/main/java/com/linkedin/venice/meta/Partition.java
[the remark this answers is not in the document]
gaojieliu: I see.
Javadoc should be good enough.
At the first glance, the code is quite confusing.

services/venice-router/src/main/java/com/linkedin/venice/router/api/VeniceDelegateMode.java
```diff
@@ -221,20 +217,14 @@ public <H, P extends ResourcePath<K>, K, R> Scatter<H, P, K> scatter(
...
         RouterExceptionAndTrackingUtils.FailureType failureType = RouterExceptionAndTrackingUtils.FailureType.REGULAR;
         if (venicePath.isRetryRequest()) {
           // don't record it as unhealthy request.
           failureType = RouterExceptionAndTrackingUtils.FailureType.RETRY_ABORTED_BY_NO_AVAILABLE_REPLICA;
         }
         String isRetry = venicePath.isRetryRequest() ? "retry " : "";
-        String errMsg = "Partitions : " + partitions + " not available to serve " + isRetry + "request of type: "
-            + venicePath.getRequestType();
+        String errMsg =
+            resourceName + " not available to serve " + isRetry + "request of type: " + venicePath.getRequestType();
```
gaojieliu, line 227: Can we keep this piece of logic for debugging purpose?
The non-available case won't happen at steady state and if it happens, the partition info in the log will help investigate the problem.

FelixGV: I didn't remove this because of the overhead in the error-handling path, but because the collection returned by `scatterGatherRequest.getPartitionsNames()` appear to be populated only within `BROADCAST_BY_PARTITION` and `BROADCAST_BY_PRIMARY_HOST`, which I believe are unused strategies. Actually, I'd like to delete these two strategies entirely, and the partition names from `ScatterGatherRequest`, but there were a lot of test changes needed (I could still do it though). Essentially, the current code ends up instantiating a new `HashSet` for each `ScatterGatherRequest`, even though it is never used. It's possible that I didn't understand the code properly though. I am not sure.

gaojieliu: I think it is populated by the caller:
```
scatter.addOfflineRequest(new ScatterGatherRequest<>(Collections.emptyList(), keySet, partitionName));
```
If we want to get rid of `partitionNames` set from `ScatterGatherRequest`, I guess we can calculate the partition from one key of the offline `ScatterGatherRequest` on the fly, which should be lightweight.

FelixGV: Oh yeah, that's right, that's where it was happening, and I removed it in this PR. Good idea to extract the partition from the first key... I'll do that.

internal/venice-common/src/main/java/com/linkedin/venice/meta/Partition.java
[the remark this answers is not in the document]
FelixGV: Done.

internal/venice-common/src/main/java/com/linkedin/venice/utils/ArrayCollection.java
```diff
@@ -0,0 +1,147 @@
...
+      }
+      return populatedSize;
+    };
+  }
+
+  /**
+   * @param array backing the collection
+   * @param populatedSizeSupplier for when the caller has a more efficient way of knowing the populated size of the array
+   */
+  public ArrayCollection(E[] array, IntSupplier populatedSizeSupplier) {
```
gaojieliu, line 35: As a general utility, shall we check whether `IntSupplier` will produce a size, which is larger than the underlying `array` length?

internal/venice-common/src/main/java/com/linkedin/venice/utils/ArrayCollection.java
```diff
@@ -0,0 +1,147 @@
...
+        } else {
+          return true;
+        }
+      }
+      return false;
+    }
+
+    @Override
+    public E next() {
+      if (this.index == array.length) {
```
gaojieliu, line 141: These functions are tricky.
Shall we check against `populatedSize` or `array.length`?

internal/venice-common/src/main/java/com/linkedin/venice/utils/ArrayCollection.java
FelixGV: Sure, I can add that.

internal/venice-common/src/main/java/com/linkedin/venice/utils/ArrayCollection.java
FelixGV: The intent is that the iterator traverses the entire array, but skips the null entries. I'll add JavaDoc for this.

internal/alpini/router/alpini-router-api/src/main/java/com/linkedin/alpini/router/api/BroadcastScatterGatherRequest.java
```diff
@@ -0,0 +1,26 @@
...
+
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Objects;
+import java.util.Set;
+import javax.annotation.Nonnull;
+
+
+public class BroadcastScatterGatherRequest<H, K> extends ScatterGatherRequest<H, K> {
```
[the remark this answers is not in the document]
FelixGV: Done.

gaojieliu (approved):
Great, thanks for the optimizations!

PR #413, opened 2023-05-10
[da-vinci-client] Do not attempt to restore meta system store on DaVinciBackend initialization

xunyin8:
Users that have ran DaVinci based metadata or store repository will run into the following exception: java.lang.UnsupportedOperationException: The implementation ThinClientMetaStoreBasedRepository should not be subscribing to system store...
when they try to run DaVinci client with thin client based store repository. This is because DaVinci backend will try to restore all on disk stores upon initialization, including meta system stores. User would then have to perform manual cleanup in the data.base.path in order to proceed.

1. Added logic to skip restoring meta system stores so they will only be subscribed explictly when needed.

2. Added an integration test with static data.base.path to reproduce the exact problem.

## How was this PR tested?
Existing tests and added new integration test

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
@@ -28,6 +28,7 @@ import com.linkedin.davinci.store.cache.backend.ObjectCacheBackend;
 import com.linkedin.davinci.store.cache.backend.ObjectCacheConfig;
 import com.linkedin.venice.client.store.ClientConfig;
 import com.linkedin.venice.client.store.ClientFactory;
+import com.linkedin.venice.common.VeniceSystemStoreType;
 import com.linkedin.venice.exceptions.VeniceException;
 import com.linkedin.venice.exceptions.VeniceNoStoreException;
 import com.linkedin.venice.kafka.protocol.state.PartitionState;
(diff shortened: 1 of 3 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
```diff
@@ -261,6 +262,12 @@ protected synchronized void bootstrap(Optional<Set<String>> managedClients) {
     for (AbstractStorageEngine storageEngine: storageEngines) {
       String kafkaTopicName = storageEngine.getStoreName();
       String storeName = Version.parseStoreFromKafkaTopicName(kafkaTopicName);
+      if (VeniceSystemStoreType.META_STORE.isSystemStore(storeName)) {
+        // Do not bootstrap meta system store via DaVinci backend initialization since the operation is not supported by
+        // ThinClientMetaStoreBasedRepository. This shouldn't happen normally, but it's possible if the user was using
+        // DVC based metadata for the same store and switched to thin client based metadata.
+        continue;
```
anaberezhnov, line 269: Should we remove the obsolete state from the disk in such case?

anaberezhnov (approved):

clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
xunyin8: Will update the code to remove it the obsolete state once we are fully on system store free FC (request based metadata).

PR #416, opened 2023-05-11
[vpj][controller] Add Targeted Region Push feature to VPJ

adamxchen:
## Summary

Venice by design would push data to all regions when using VPJ. While it makes data available globally sooner, it also introduces a risk that a bad VPJ could cause global outage especially for users using da-vinci clients. In this PR, it introduces the first part of Single/Targeted Region Push feature that would allow to push to one region only and consider it is completed. The idea is very straightforward. Basically we ask non-canary-region to skip the add version admin message such that only the canary region will add version and consume the data. 

Will add the functionality that pushes data to the rest of world later.

### Suggested review order:
1. VenicePushJob
2. ControllerClient
3. VeniceParentHelixAdmin
4. AdminExecutionTask
5. others

### Admin protocol upgrade
Added a new field to 
```
            {
              "name": "targetedRegions",
              "doc": "The list of regions that is separated by comma for targeted region push. If set, this admin message should only be consumed by the targeted regions",
              "type": [
                "null",
                {
                  "type": "array",
                  "items":  "string"
                }
              ],
              "default": null
            }
```
to the admin protocol and bump the version to 71.

## How was this PR tested?

Newly added integ tests in `TestPushJobWithNativeReplication.java` and new unit tests in `AdminConsumptionTaskTest.java`
Some sample logs
```
JobStatusQueryResponse(
version: 2,
status: COMPLETED,
statusDetails: This is targeted region push and following regions dc-0 are expected to be COMPLETED. Parent Kafka topic truncated,
extraInfo: {dc-0=COMPLETED, dc-1=NOT_CREATED},
extraDetails: {dc-0=http://adchen-mn1.linkedin.biz:55268 1/1 partitions completed in0 Da Vinci instances., dc-1=http://adchen-mn1.linkedin.biz:55372 Offline job hasn't been created yet. and Version for topic testTargetedRegionPushJob_b0f6447cea393_99a4386a_v2 isn't created since this is a targeted region push},
ControllerResponse(cluster: venice-cluster0, name: testTargetedRegionPushJob_b0f6447cea393_99a4386a, error: null, errorType: null, exceptionType: null))
2023-05-16 16:54:12 - [] INFO [VenicePushJob] [TestNG-method=testTargetedRegionPushJob-1] Specific status: {dc-0=COMPLETED, dc-1=NOT_CREATED}
2023-05-16 16:54:12 - [] INFO [VenicePushJob] [TestNG-method=testTargetedRegionPushJob-1] 		New overall details: This is targeted region push and following regions dc-0 are expected to be COMPLETED. Parent Kafka topic truncated
2023-05-16 16:54:12 - [] INFO [VenicePushJob] [TestNG-method=testTargetedRegionPushJob-1] 		New specific details for dc-0: http://adchen-mn1.linkedin.biz:55268 1/1 partitions completed in0 Da Vinci instances.
2023-05-16 16:54:12 - [] INFO [VenicePushJob] [TestNG-method=testTargetedRegionPushJob-1] Successfully pushed testTargetedRegionPushJob_b0f6447cea393_99a4386a_v2
```

Users now can turn on this feature using the flag `targeted.regions.push` to specify a list of regions, separated by comma or "single.region.push", a boolean, to push a single region


clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
```diff
@@ -387,6 +387,17 @@ public class VenicePushJob implements AutoCloseable {
...
   public static final int NOT_SET = -1;
   private static final Logger LOGGER = LogManager.getLogger(VenicePushJob.class);
+  /**
+   * Config to enable single-colo/region canary push for VPJ.
+   * In this mode, the VPJ will only push data to the source colo/region first, perform validation and finally
+   * leverage data recovery/repush to propagate data globally.
+   * The source colo is determined by Controller using the given priority order:
+   *     1. Parent controller emergency source fabric config.
+   *     2. VPJ plugin source grid fabric config, aka {@link #SOURCE_GRID_FABRIC}.
+   *     3. Store level source fabric config.
```
majisourav99, line 397: you mean `nativeReplicationSourceFabric` store config or you are planning to add new one?

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
```diff
@@ -2606,13 +2637,25 @@ private void pollStatusUntilComplete(
       });
 
       if (overallStatus.isTerminal()) {
+        if (pushJobSetting.canaryRegionPush) {
+          if (completedDatacenters.size() == 1) {
+            LOGGER.info("Successfully pushed {} to {}", topicInfo.topic, completedDatacenters.iterator().next());
+            return;
+          } else {
+            throw new VeniceException(
+                "Canary region push job ended up with zero or more than one region ending with unexpected COMPLETED status, "
```
majisourav99, line 2646: it cannot be more than 1 right ? may be less than 1. accordingly update the message.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -5332,6 +5348,10 @@ private OfflinePushStatusInfo getOfflinePushStatusInfo(
       // Check whether cluster is in maintenance mode or not
       if (isClusterInMaintenanceMode(clusterName)) {
         moreDetailsBuilder.append("Cluster: ").append(clusterName).append(" is in maintenance mode");
+      } else if (isCanaryRegionPush) {
+        moreDetailsBuilder.append("Version for topic ")
+            .append(kafkaTopic)
+            .append(" isn't created since this is a canary region push");
```
majisourav99, line 5354: can we add the cluster name and fabric name in the log?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
```diff
@@ -1664,13 +1667,24 @@ Version addVersionAndTopicOnly(
         versionSwapDeferred);
     Version newVersion = result.getSecond();
     if (result.getFirst()) {
+      // fail early if canary region push is enabled but source region is not set.
+      if (canaryRegionPush && StringUtils.isEmpty(newVersion.getNativeReplicationSourceFabric())) {
```
majisourav99, line 1671: shouldnt we look up cluster level config if store is not set?

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
adamxchen: I mean `nativeReplicationSourceFabric` store config.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
adamxchen: I'm being defensive here in case the canary region push feature is broken for whatever reasons and 0 or more than 1 regions ingest the data.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
adamxchen: We can but I prefer to not add the information since `fabric` name is already present in the log and `cluster` seems has not good value?

Sample log
```
INFO [VenicePushJob] [TestNG-method=testCanaryRegionPushJob-1] JobStatusQueryResponse(
    version: 2,
    status: COMPLETED,
    statusDetails: null,
    extraInfo: {dc-0=COMPLETED, dc-1=NOT_CREATED},
    extraDetails: {dc-0=http://adchen-mn1.linkedin.biz:51779 1/1 partitions completed in Da Vinci., dc-1=http://adchen-mn1.linkedin.biz:51880 Offline job hasn't been created yet. and Version for topic testCanaryRegionPushJob_9419d10a6d17b_4a94b9ac_v2 isn't created since this is a canary region push},
```

The `extraDetails` already contains the fabric name.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
adamxchen: No we don't need to. The `NativeReplicationSourceFabric` should have been decided and set by the `AddTopicAndVersionOnly` method above ([line](https://github.com/linkedin/venice/blob/faaea30a22ccc36756659f6a400d19987c91e574/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java#L1651)) using various values, e.g. cluster config and store config.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
```diff
@@ -708,6 +721,19 @@ private PushJobSetting getPushJobSetting(VeniceProperties props) {
       if (pushJobSettingToReturn.isSourceETL) {
         throw new VeniceException("Source ETL is not supported while using Kafka Input Format");
       }
+
+      if (pushJobSettingToReturn.canaryRegionPush) {
+        throw new VeniceException("Canary region push is not supported while using Kafka Input Format");
+      }
+    }
+
+    if (pushJobSettingToReturn.canaryRegionPush) {
```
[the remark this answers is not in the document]
adamxchen: Do you mean merging these validation with KIF validation above?

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
```diff
@@ -1046,7 +1072,9 @@ public void run() {
         // Update and send push job details with new info to the controller
         pushJobDetails.pushId = pushId;
         pushJobDetails.partitionCount = kafkaTopicInfo.partitionCount;
-        pushJobDetails.valueCompressionStrategy = kafkaTopicInfo.compressionStrategy.getValue();
+        pushJobDetails.valueCompressionStrategy = kafkaTopicInfo.compressionStrategy != null
```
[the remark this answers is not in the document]
adamxchen: In my test 😂 . 
It used to be no warning but in my recent tests for `VenicePushJob`, this field is set to null so Spotbug doesn't like this approach.

internal/venice-common/src/main/java/com/linkedin/venice/controllerapi/ControllerClient.java
```diff
@@ -270,7 +272,8 @@ public VersionCreationResponse requestTopicForWrites(
    * @param batchJobHeartbeatEnabled whether batch push job enables the heartbeat
    * @param rewindTimeInSecondsOverride if a valid value is specified (>=0) for hybrid store, this param will override
    *                                       the default store-level rewindTimeInSeconds config.
-   *
+   * @param deferVersionSwap whether to defer version swap after the push is done
+   * @param canaryRegionPush whether this is a topic requested for a canary region push
    * @return VersionCreationResponse includes topic and partitioning
    */
   public VersionCreationResponse requestTopicForWrites(
```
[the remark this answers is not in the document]
adamxchen: Yep, will do that!

FelixGV:
Just a general comment on the naming... I think "Canary Region Push" sounds a bit confusing, for various reasons. Some other naming ideas that I think would be better suited include: "Single Region Push", "Targeted Push" or "Non-Global Push"... Would any of those sound ok to you?

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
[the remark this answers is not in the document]
nisargthakkar: yes. Both the `if` checks are `if (pushJobSettingToReturn.canaryRegionPush) {`. These can be merged into one conditional block

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
```diff
@@ -708,6 +721,19 @@ private PushJobSetting getPushJobSetting(VeniceProperties props) {
       if (pushJobSettingToReturn.isSourceETL) {
         throw new VeniceException("Source ETL is not supported while using Kafka Input Format");
       }
+
+      if (pushJobSettingToReturn.canaryRegionPush) {
```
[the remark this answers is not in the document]
nisargthakkar: If there are no implementation concerns and overheads, let's not add synthetic blockers. We never know when we'll need any feature

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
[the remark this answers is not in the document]
nisargthakkar: Umm, then let's modify the test. Test should mimic code behavior. Having a different code paths in test vs actual code seems sketchy

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
```diff
@@ -387,6 +387,17 @@ public class VenicePushJob implements AutoCloseable {
...
+   * Config to enable single-colo/region canary push for VPJ.
+   * In this mode, the VPJ will only push data to the source colo/region first, perform validation and finally
+   * leverage data recovery/repush to propagate data globally.
+   * The source colo is determined by Controller using the given priority order:
+   *     1. Parent controller emergency source fabric config.
+   *     2. VPJ plugin source grid fabric config, aka {@link #SOURCE_GRID_FABRIC}.
+   *     3. Store level source fabric config.
+   *     4. Cluster level source fabric config.
+   */
+  public static final String CANARY_REGION_PUSH = "canary.region.push";
```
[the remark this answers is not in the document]
nisargthakkar: Hmm. Controller can look at the `regions_filter` field in the `request_topic` call and use a source region from that list. After this, we won't need any more refactoring, right?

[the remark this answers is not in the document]
majisourav99: yeah making the canary region extensible to contain multiple colos is a good idea. we might fail the push if set of target colo does not contain the NR source fabric.

adamxchen:
> Just a general comment on the naming... I think "Canary Region Push" sounds a bit confusing, for various reasons. Some other naming ideas that I think would be better suited include: "Single Region Push", "Targeted Push" or "Non-Global Push"... Would any of those sound ok to you?

Hey @FelixGV I was hesitating between "Canary Region Push" and "Single Region Push". "Single Region Push" is definitely easy to understand but would it implicitly indicate the push only happens to a single region, whereas we actually push to one region first but still push to the rest of world later? What are those reasons you find "Canary Region Push" confusing?

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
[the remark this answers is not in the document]
adamxchen: Okay, so turned out my tests actually set this field but Spotbug wants to eliminate this possibility of setting this field to null. Should be okay to keep a check.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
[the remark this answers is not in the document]
adamxchen: After a second thought, let's do it.

FelixGV:
> Hey @FelixGV I was hesitating between "Canary Region Push" and "Single Region Push". "Single Region Push" is definitely easy to understand but would it implicitly indicate the push only happens to a single region, whereas we actually push to one region first but still push to the rest of world later? What are those reasons you find "Canary Region Push" confusing?

Sure, let me elaborate.

1. I know that canary is used in several companies, but I am not sure that the meaning is standardized. I think it could convey different things to different people depending on their background. 
2. At LinkedIn, we use canary to mean deploying a change to one or few hosts. What we are doing here is different. I know that you added "region" to it and the two ideas are indeed analogous, but I wouldn't want our internal users to onboard to this thinking that it is a "canary push" (as opposed to "canary _region_ push"). We may or may not build true canary push functionality at some point, but let's reserve that keyword for when we do.
3. The project has two parts: 1) pushing to a subset of regions, and 2) orchestrating a push to a subset of regions, followed by a health check of that region, followed by pushing to the remaining regions. I don't think we should club these two together under the same name. We should have precise names for each functionality, and if one functionality is a superset of another, that is fine.

adamxchen:
> 

Ack. Let's use "Single Region Push" then!

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -2612,7 +2612,14 @@ public Version incrementVersionIdempotent(
       Optional<X509Certificate> requesterCert,
       long rewindTimeInSecondsOverride,
       Optional<String> emergencySourceRegion,
-      boolean versionSwapDeferred) {
+      boolean versionSwapDeferred,
+      boolean canaryRegionPush) {
+    if (canaryRegionPush) {
+      // only parent controller should handle this request. See @{link
+      // VeniceParentHelixAdmin#incrementVersionIdempotent}
```
lluwm, line 2619: use /** xxx */ to access the link easily?

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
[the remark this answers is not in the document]
adamxchen: Updated the PR and now we call it "targeted region push" 😄

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
```diff
@@ -3408,64 +3442,100 @@ private OfflinePushStatusInfo getOffLineJobStatus(
...
-    int successCount = childClusters.size() - failCount;
-    if (!(successCount >= (childClusters.size() / 2) + 1)) {
-      // Strict majority must be reachable, otherwise keep polling
-      currentReturnStatus = ExecutionStatus.PROGRESS;
+    ExecutionStatus currentReturnStatus = ExecutionStatus.NEW;
+    StringBuilder currentReturnStatusDetails = new StringBuilder();
+    if (isCanaryRegionPush) {
+      // for canary region push, ignore other child regions' status and canary region is expected to
+      // be ERROR/COMPLETED, which has higher priority, as end states
+      currentReturnStatus = statuses.get(statuses.size() - 1);
```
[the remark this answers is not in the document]
adamxchen: Thanks for the review! I've removed this logic so plz feel free to take a look again!

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
nisargthakkar: Also, I believe it should be `{@link ...}` instead of `@{link}`

adamxchen: Updated. Please review again.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -5682,7 +5772,16 @@ public String getNativeReplicationSourceFabric(
     }
 
     if (sourceFabric == null || sourceFabric.isEmpty()) {
-      sourceFabric = multiClusterConfigs.getControllerConfig(clusterName).getNativeReplicationSourceFabric();
+      sourceFabric = getMultiClusterConfigs().getControllerConfig(clusterName).getNativeReplicationSourceFabric();
+    }
+
+    if (targetedRegions != null) {
+      Set<String> regions = RegionUtils.parseRegionsFilterList(targetedRegions);
+      if (!regions.contains(sourceFabric) && !emergencySourceRegion.isPresent()) {
```
majisourav99, line 5780: Do we want to warn the user for this case? Ideally user should use the originating data fabric + other colo if they want a 2nd colo. Since its a user given config now, user could provide non-native colo, which would be very inefficient.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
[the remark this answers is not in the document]
majisourav99: How about using `destination region push` ? src/dst is more standardized than target IMHO.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
adamxchen: Native colo/region is set by this method, so when user provides a region that isn't selected by criteria above,  whatever they provide will be selected as native colo/region.

Though we may have preferred native region to choose from, it's not necessary inefficient from my understanding and it's okay to let customers to specify it. Let me know if I miss anything here.

internal/venice-common/src/main/java/com/linkedin/venice/utils/RegionUtils.java
```diff
@@ -44,4 +48,17 @@ public static String getRegionSpecificMetricPrefix(String localRegionName, Strin
...
+
+  /**
+   * A helper function to split a region list with {@link #REGION_FILTER_LIST_SEPARATOR}
+   */
+  public static Set<String> parseRegionsFilterList(String regionsFilterList) {
+    if (StringUtils.isEmpty(regionsFilterList)) {
+      return Collections.emptySet();
+    }
+    Set<String> regions = new HashSet<>();
+    String[] tokens = regionsFilterList.trim().toLowerCase().split(REGION_FILTER_LIST_SEPARATOR);
```
nisargthakkar, line 60: I don't think we enforce lower case for region names. I might be wrong, though

services/venice-controller/src/main/resources/avro/AdminOperation/v71/AdminOperation.avsc
```diff
@@ -0,0 +1,1012 @@
...
+              "default": -1
+            },
+            {
+              "name": "versionSwapDeferred",
+              "doc": "Indicates if swapping this version to current version after push completion should be initiated or not",
+              "type": "boolean",
+              "default": false
+            },
+            {
+              "name": "targetedRegions",
```
nisargthakkar, line 731: Can we make this an array instead of comma-separated-list?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
nisargthakkar: Let's instead fail the job if user doesn't specify the source region as a targeted region?

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
```diff
@@ -2585,14 +2611,19 @@ private void pollStatusUntilComplete(
...
+              .queryOverallJobStatus(topicToMonitor, incrementalPushVersion, pushJobSetting.targetedRegions));
+
       if (response.isError()) {
         // status could not be queried which could be due to a communication error.
         throw new VeniceException(
             "Failed to connect to: " + pushJobSetting.veniceControllerUrl + " to query job status, after "
                 + pushJobSetting.controllerStatusPollRetries + " attempts. Error: " + response.getError());
       }
 
+      LOGGER.info("DEBUGGING");
```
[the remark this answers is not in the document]
adamxchen: Uh... Sure!

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
adamxchen: One gotcha might be if the user provides an invalid region, which cannot be used to find the corresponding kakfa bootstrap url, the existing behavior of this is parent controller will use the existing kafka config ([code](https://github.com/linkedin/venice/blob/d689f173e461b51e531e41bcfb685057637995cb/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java#L2358)), and none of existing child regions will create a new version, eventually the push job will fail due to time out.

internal/venice-common/src/main/java/com/linkedin/venice/utils/RegionUtils.java
adamxchen: This method was originally in VeniceHelixAdmin and moved to this `venice-common` so I suppose it has been working this way.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
```diff
@@ -710,6 +724,15 @@ private PushJobSetting getPushJobSetting(VeniceProperties props) {
       }
     }
 
+    if (StringUtils.isNotEmpty(pushJobSettingToReturn.targetedRegions)) {
+      if (pushJobSettingToReturn.isSourceETL) {
```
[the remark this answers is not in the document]
adamxchen: Ack. Is ETL snapshot per colo/region instead of global view?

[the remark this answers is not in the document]
nisargthakkar: Yes. ETL is from a specific region

services/venice-controller/src/main/resources/avro/AdminOperation/v71/AdminOperation.avsc
adamxchen: Sure, a list of string is more performant!

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/consumer/AdminExecutionTask.java
```diff
@@ -627,8 +625,9 @@ private void handleAddVersion(AddVersion message) {
             replicationMetadataVersionId);
       }
     } else {
-      String regions = message.targetedRegions != null ? message.targetedRegions.toString() : null;
-      if (StringUtils.isNotEmpty(regions) && !RegionUtils.parseRegionsFilterList(regions).contains(this.regionName)) {
+      boolean skipConsumption = message.targetedRegions != null
+          && !message.targetedRegions.stream().map(Object::toString).collect(Collectors.toSet()).contains(regionName);
```
nisargthakkar, line 629: Instead of `.collect().contains`, we can use `. anyMatch`

adamxchen: Cool! Updated to `message.targetedRegions.stream().map(Object::toString).noneMatch(regionName::equals);`

services/venice-controller/src/main/java/com/linkedin/venice/controller/Admin.java
```diff
@@ -225,7 +259,8 @@ Version incrementVersionIdempotent(
       Optional<X509Certificate> requesterCert,
       long rewindTimeInSecondsOverride,
       Optional<String> emergencySourceRegion,
-      boolean versionSwapDeferred);
+      boolean versionSwapDeferred,
+      String targetedRegions);
```
lluwm, line 263: Personally, I would prefer to do the string parsing outside the `incrementVersionIdempotent` and pass a list/set of regions to it to keep the API implementation simple.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
```diff
@@ -3383,37 +3409,106 @@ private OfflinePushStatusInfo getOffLineJobStatus(
...
+        ExecutionStatus status = statuses.get(region);
+        /**
+         * We consider failures for the following cases:
+         * 1. the provided region isn't available for the cluster, in other words, invalid targeted region is provided.
+         * 2. the provided region has troubles to query the job status.
+         */
+        if (status != null) {
+          sortedStatuses.add(status);
+          if (status.equals(ExecutionStatus.UNKNOWN)) {
+            targetedRegionFailCount += 1;
```
lluwm, line 3466: nit: targetedRegionFailCount += 1; -> targetedRegionFailCount++;

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
```diff
@@ -3383,37 +3409,106 @@ private OfflinePushStatusInfo getOffLineJobStatus(
...
+      }
+      sortedStatuses = sortedStatuses.stream()
+          .sorted(Comparator.comparingInt(VeniceHelixAdmin.STATUS_PRIORITIES::indexOf))
+          .collect(Collectors.toList());
+
+      currentReturnStatusDetails.append("This is targeted region push and following regions ")
+          .append(targetedRegions)
+          .append(" are expected to be COMPLETED. ");
+      currentReturnStatus =
+          getFinalReturnStatus(sortedStatuses, targetedRegionsSet, targetedRegionFailCount, currentReturnStatusDetails);
```
lluwm, line 3480: I am hoping that If we can move `currentReturnStatusDetails` before @L3472, then the remaining part becomes a common code for L3446-L3451, so that we can move it outside the branch and change would be much simpler.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
```diff
@@ -3383,37 +3409,106 @@ private OfflinePushStatusInfo getOffLineJobStatus(
...
+  }
+
+  private ExecutionStatus getFinalReturnStatus(
+      List<ExecutionStatus> sortedStatuses,
+      Set<String> childRegions,
+      int failCount,
+      StringBuilder currentReturnStatusDetails) {
+    ExecutionStatus currentReturnStatus = ExecutionStatus.NEW;
+
+    if (sortedStatuses.size() > 0) {
```
lluwm, line 3506: nit: if (sortedStatuses.size() > 0) -> if (!sortedStatuses.isEmpty())

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
```diff
@@ -3383,37 +3409,106 @@ private OfflinePushStatusInfo getOffLineJobStatus(
...
+      currentReturnStatus = sortedStatuses.get(0);
+    }
+
+    int successCount = childRegions.size() - failCount;
+    if (!(successCount >= (childRegions.size() / 2) + 1)) {
```
lluwm, line 3511: nit: if (!(successCount >= (childRegions.size() / 2) + 1)) -->
if (successCount < (childRegions.size() / 2 + 1))

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
adamxchen: Discussed offline with @majisourav99 . The reason why it can be inefficient is HDFS/Samza reprocessing job may be always from one region A (in other words, largely depends on how offline job is setup), and if the targeted regions are region B and C, although it should work with no issue, additional data hop from region A to region B or region C can add up the overall push time. It's preferred that users always include the region A in this config and specify it as the source fabric, such that it will be always get selected to reduce the extra hop

adamxchen: I will try to add some logic in VPJ.
When controller query store details in VPJ, we also check the store's config of native region setup to see if the provided targeted regions contain the region or not. If not, we will give a warning in VPJ so people can see it in the logs. Maybe we can expose one controller API for it but I don’t think it is worth the effort just for a warning msg.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
adamxchen: For some weird reasons, github is not showing you the latest changes. I've done it already..

services/venice-controller/src/main/java/com/linkedin/venice/controller/Admin.java
adamxchen: Let's keep it same way as regionFilter if you have no strong objection

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
adamxchen: We decided to introduce 2 configs. One config is boolean (easy mode) so store config will automatically be used as the targeted region, so customers don't have to do anything. The advanced mode is we can take a list of strings as regions but it can impact the performance if customers specifies a region other than where data comes from.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
```diff
@@ -388,6 +389,24 @@ public class VenicePushJob implements AutoCloseable {
...
   private static final Logger LOGGER = LogManager.getLogger(VenicePushJob.class);
 
+  /**
+   * Config to enable single targeted region push mode in VPJ.
+   * In this mode, the VPJ will only push data to a single colo/region first, perform validation and finally
+   * leverage data recovery/repush to propagate data globally.
+   * The single solo/region is decided by the store config in {@link StoreResponse#getStore()}}.
+   * For multiple targeted regions push, may use the advanced mode. See {@link #TARGETED_REGIONS_PUSH}.
+   */
+  public static final String SINGLE_REGION_PUSH = "single.region.push";
```
majisourav99, line 399: can we make name like "enable.single.region.push" or "single.region.push.enabled" ?

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
```diff
@@ -2271,6 +2298,18 @@ private StoreSetting getSettingsFromController(ControllerClient controllerClient
     storeSetting.isWriteComputeEnabled = storeResponse.getStore().isWriteComputationEnabled();
     storeSetting.isIncrementalPushEnabled = storeResponse.getStore().isIncrementalPushEnabled();
     storeSetting.storeRewindTimeInSeconds = DEFAULT_RE_PUSH_REWIND_IN_SECONDS_OVERRIDE;
+    if (setting.singleRegionPush && setting.targetedRegions == null) {
```
majisourav99, line 2301: should we fail if both set ? or could we make the logic be  must enable feature flag, then optionally set the list of target regions else it will use the NR source fabric ?

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
adamxchen: Hmm, I'd like to keep it this way because I want both configs follow the same format as "xxx.xxx.push
e.g. `single.region.push` and `targeted.regions.push`. Lmk if you still prefer to change it.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
adamxchen: > could we make the logic be must enable feature flag, then optionally set the list of target regions else it will use the NR source fabric ?

I think this is a great idea. Will do it

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
```diff
@@ -666,6 +688,14 @@ private PushJobSetting getPushJobSetting(VeniceProperties props) {
     pushJobSettingToReturn.deferVersionSwap = props.getBoolean(DEFER_VERSION_SWAP, false);
     pushJobSettingToReturn.repushTTLEnabled = props.getBoolean(REPUSH_TTL_ENABLE, false);
     pushJobSettingToReturn.repushTTLInSeconds = NOT_SET;
+    pushJobSettingToReturn.isTargetedRegionPush = props.getBoolean(TARGETED_REGION_PUSH_ENABLED, false);
```
majisourav99, line 691: nit: variable name to match config name, `isTargetedRegionPushEnabled`

adamxchen: Updated.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
```diff
@@ -2271,6 +2303,18 @@ private StoreSetting getSettingsFromController(ControllerClient controllerClient
     storeSetting.isWriteComputeEnabled = storeResponse.getStore().isWriteComputationEnabled();
     storeSetting.isIncrementalPushEnabled = storeResponse.getStore().isIncrementalPushEnabled();
     storeSetting.storeRewindTimeInSeconds = DEFAULT_RE_PUSH_REWIND_IN_SECONDS_OVERRIDE;
+    if (setting.isTargetedRegionPush && setting.targetedRegions == null) {
+      // only override the targeted regions if it is not set and it is a single region push
+      setting.targetedRegions = storeResponse.getStore().getNativeReplicationSourceFabric();
+      if (setting.targetedRegions == null) {
```
majisourav99, line 2309: I see getNativeReplicationSourceFabric default value is empty string, regardless good to use `StringUtils.isEmpty` here

adamxchen: That's my miss.  Thanks for catching. Let me update it

internal/venice-common/src/main/java/com/linkedin/venice/utils/RegionUtils.java
majisourav99, line 60: can do 

`return  Stream.of(regionsFilterList.trim().split("\\s*,\\s*")).collect(Collectors.toSet());`

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -5680,7 +5770,16 @@ public String getNativeReplicationSourceFabric(
...
```
majisourav99, line 5778: Add a detailed log here indicating the target colo list, NR fabric and which one we are choosing.

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/consumer/AdminExecutionTask.java
```diff
@@ -622,18 +625,30 @@ private void handleAddVersion(AddVersion message) {
...
-          storeName,
-          pushJobId,
-          versionNumber,
-          numberOfPartitions,
-          pushType,
-          remoteKafkaBootstrapServers,
-          rewindTimeInSecondsOverride,
-          replicationMetadataVersionId,
-          message.versionSwapDeferred);
+      boolean skipConsumption = message.targetedRegions != null && !message.targetedRegions.isEmpty()
```
majisourav99, line 628: can use string utils and RegionUtils methods here

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
majisourav99: Good. can we add a test to catch this?

adamxchen: done

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
```diff
@@ -2606,13 +2653,28 @@ private void pollStatusUntilComplete(
...
-          throw new VeniceException(
-              "Push job error reported by controller: " + pushJobSetting.veniceControllerUrl + "\ncontroller response: "
-                  + response.toString());
+          // 2) For targeted region push, not all targeted regions have completed.
+
+          StringBuilder errorMsg = new StringBuilder().append("Push job error reported by controller: ")
+              .append(pushJobSetting.veniceControllerUrl)
+              .append(
+                  StringUtils.isNotEmpty(pushJobSetting.targetedRegions)
+                      ? ". Zero or more than specified regions are in COMPLETED states. "
```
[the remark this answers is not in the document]
adamxchen: When the job fails, the detailed controller response will be printed out, which contains the targeted regions, and region status so don't need to add it here.

e.g.
```
JobStatusQueryResponse(
version: 2,
status: COMPLETED,
statusDetails: This is targeted region push and following regions dc-0 are expected to be COMPLETED. Parent Kafka topic truncated,
extraInfo: {dc-0=COMPLETED, dc-1=NOT_CREATED},
```

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/consumer/AdminExecutionTask.java
adamxchen: `targetedRegions` is a list of String so these methods don't apply. We can keep it this way.

internal/venice-common/src/main/java/com/linkedin/venice/utils/RegionUtils.java
adamxchen: We still support java 8 so cool syntax like this doesn't compile :(

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
adamxchen: done

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
```diff
@@ -2271,6 +2303,19 @@ private StoreSetting getSettingsFromController(ControllerClient controllerClient
...
     storeSetting.isIncrementalPushEnabled = storeResponse.getStore().isIncrementalPushEnabled();
     storeSetting.storeRewindTimeInSeconds = DEFAULT_RE_PUSH_REWIND_IN_SECONDS_OVERRIDE;
+    if (setting.isTargetedRegionPushEnabled && setting.targetedRegions == null) {
+      // only override the targeted regions if it is not set and it is a single region push
+      setting.targetedRegions = storeResponse.getStore().getNativeReplicationSourceFabric();
+      if (StringUtils.isEmpty(setting.targetedRegions)) {
+        throw new VeniceException(
+            "The store either does not have native replication mode enabled or set up default source fabric.");
+      }
+      if (StringUtils.isNotEmpty(setting.targetedRegions)) {
```
majisourav99, line 2313: most of the time this will trigger as all stores are on NR today. Can we remove the inc_push + target region combo validation earlier at line ~691.

adamxchen: Line 691 is still in the constructor where we haven't made the call to controller yet. That's why I moved here.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
```diff
@@ -2748,7 +2811,9 @@ protected void setupDefaultJobConf(
     conf.set(KAFKA_BOOTSTRAP_SERVERS, topicInfo.kafkaUrl);
     conf.set(PARTITIONER_CLASS, topicInfo.partitionerClass);
     // flatten partitionerParams since JobConf class does not support set an object
-    topicInfo.partitionerParams.forEach(conf::set);
+    if (topicInfo.partitionerParams != null) {
```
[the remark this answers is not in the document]
adamxchen: There's a chance they can be null if the store has no `partitionerConfig`

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -5330,6 +5415,10 @@ private OfflinePushStatusInfo getOfflinePushStatusInfo(
       // Check whether cluster is in maintenance mode or not
       if (isClusterInMaintenanceMode(clusterName)) {
         moreDetailsBuilder.append("Cluster: ").append(clusterName).append(" is in maintenance mode");
+      } else if (StringUtils.isNotEmpty(targetedRegions)) {
+        moreDetailsBuilder.append("Version for topic ")
+            .append(kafkaTopic)
+            .append(" isn't created since this is a targeted region push");
```
majisourav99, line 5421: shouldnt we print this line only if target region != current region?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -5680,7 +5770,22 @@ public String getNativeReplicationSourceFabric(
...
+      sourceFabric = getMultiClusterConfigs().getControllerConfig(clusterName).getNativeReplicationSourceFabric();
+    }
+
+    if (targetedRegions != null) {
+      Set<String> regions = RegionUtils.parseRegionsFilterList(targetedRegions);
+      if (!regions.contains(sourceFabric) && !emergencySourceRegion.isPresent()) {
+        // user specifies a targeted region that isn't the chosen source fabric, so randomly choose one
+        // but emergency source region should always be respected.
+        String selection = regions.iterator().next();
+        LOGGER.info(
```
majisourav99, line 5782: maybe info->warn?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
```diff
@@ -3383,37 +3409,106 @@ private OfflinePushStatusInfo getOffLineJobStatus(
         leaderControllerUrl = controllerClient.getLeaderControllerUrl();
       } catch (VeniceException exception) {
         LOGGER.warn("Couldn't query {} for job status of {}", region, kafkaTopic, exception);
-        statuses.add(ExecutionStatus.UNKNOWN);
+        statuses.put(region, ExecutionStatus.UNKNOWN);
```
majisourav99, line 3412: Should we even query non-target regions ? just check the target regions statuses here?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
adamxchen: If that's a targeted region, the `executionStatus` will not be `NOT_CREATED` so we don't need this check, though it's probably more accurate in case kafka has some issues

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
adamxchen: I think it's fine to query all regions since we can detect if non-targeted regions mistakenly ingest the message

majisourav99: why do we want to do it ? It will throw unnecessarily exception `VeniceNoHelixResourceException` if we query non-existent topics.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
adamxchen: Sure

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
adamxchen: No, it doesn't throw. Since the push isn't created, the status will always be `ExecutionStatus.NOT_CREATED`.  [Ref](https://github.com/linkedin/venice/blob/664ba727bbf5edabd49c22dfcfc61b01b9eedbbf/services/venice-controller/src/main/java/com/linkedin/venice/pushmonitor/AbstractPushMonitor.java#L314)

services/venice-controller/src/test/java/com/linkedin/venice/controller/TestVeniceParentHelixAdmin.java
```diff
@@ -3,21 +3,7 @@
...
-import static org.mockito.Mockito.doThrow;
-import static org.mockito.Mockito.eq;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.never;
-import static org.mockito.Mockito.spy;
-import static org.mockito.Mockito.timeout;
-import static org.mockito.Mockito.times;
-import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.*;
```
adamxchen: , along with the querying the targeted regions only.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
adamxchen: Discussed offline, we now query the targeted regions

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
```diff
@@ -2585,7 +2630,9 @@ private void pollStatusUntilComplete(
       JobStatusQueryResponse response = ControllerClient.retryableRequest(
           controllerClient,
           pushJobSetting.controllerStatusPollRetries,
-          client -> client.queryOverallJobStatus(topicToMonitor, incrementalPushVersion));
+          client -> client
+              .queryOverallJobStatus(topicToMonitor, incrementalPushVersion, pushJobSetting.targetedRegions));
```
nisargthakkar, line 2634: Can you test in an integration test if `emergencySourceFabric` is set, but it is not one of the targeted regions, can the push still complete?

As discussed offline, I think it will work fine (with the emergency source region not ingesting any data), but it would be good to confirm it.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -5680,7 +5758,22 @@ public String getNativeReplicationSourceFabric(
...
-      sourceFabric = multiClusterConfigs.getControllerConfig(clusterName).getNativeReplicationSourceFabric();
+      sourceFabric = getMultiClusterConfigs().getControllerConfig(clusterName).getNativeReplicationSourceFabric();
+    }
+
+    if (targetedRegions != null) {
+      Set<String> regions = RegionUtils.parseRegionsFilterList(targetedRegions);
+      if (!regions.contains(sourceFabric) && !emergencySourceRegion.isPresent()) {
+        // user specifies a targeted region that isn't the chosen source fabric, so randomly choose one
+        // but emergency source region should always be respected.
+        String selection = regions.iterator().next();
```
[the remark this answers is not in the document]
adamxchen: > by checking if any of the fabrics in the previous if/else chain is present in the targeted region list

That's the plan so I added `!regions.contains(sourceFabric)` at line 5766 to check whether the pre-selected region is in the targeted region list or not first.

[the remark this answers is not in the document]
nisargthakkar: There are subtle differences. Consider a scenario where there are the following regions: `dc-[0..5]`

`emergencySourceRegion` - None
`sourceGridFabric` - `dc-1`
`nativeReplicationSourceFabric in storeConfig` - `dc-2`
`nativeReplicationSourceFabric in app config` - `dc-3`
`targetedRegionsList` - `dc-2,dc-4`

In the current logic, it is possible for `dc-4` to get selected as the source region (`regions.iterator().next()`), but `dc-2` would be a better choice.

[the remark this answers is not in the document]
majisourav99: good example, yes we should fall back to a known config rather than throwing a die.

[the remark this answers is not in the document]
adamxchen: Uh, I see. This example makes sense. Let me update this!

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
adamxchen: Yep, you are right.

Wrote a small test case for it. When emergency region is dc-0, and targeted region is dc-1, the jobStatus response will be the following when we query all regions.

```
2023-05-25 17:36:48 - [] INFO [VenicePushJob] [TestNG-method=testTargetedRegionPushJobOverride-1] JobStatusQueryResponse(
version: 2,
status: NOT_CREATED,
statusDetails: ,
extraInfo: {dc-0=NOT_CREATED, dc-1=COMPLETED},
extraDetails: {dc-0=http://adchen-mn1.linkedin.biz:63635 Offline job hasn't been created yet. and Version creation for topic: testTargetedRegionPushJob_1ee5d864cf496_21a1e3f6_v2 got delayed, dc-1=http://adchen-mn1.linkedin.biz:63730 1/1 partitions completed in0 Da Vinci instances.},
ControllerResponse(cluster: venice-cluster0, name: testTargetedRegionPushJob_1ee5d864cf496_21a1e3f6, error: null, errorType: null, exceptionType: null))
```


Since the current implementation is only query targeted region (dc-1), in reality, this job will be reported as COMPLETED.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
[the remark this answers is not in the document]
adamxchen: Updated. And also added a unit test for this!

internal/venice-test-common/src/integrationtest/java/com/linkedin/venice/endToEnd/TestPushJobWithNativeReplication.java
```diff
@@ -943,6 +926,25 @@ public void testTargetedRegionPushJob() throws Exception {
...
+
+            TestUtils.waitForNonDeterministicAssertion(30, TimeUnit.SECONDS, () -> {
+              Map<String, Integer> coloVersions =
+                  parentControllerClient.getStore(storeName).getStore().getColoToCurrentVersions();
+
+              coloVersions.forEach((colo, version) -> {
+                if (colo.equals(DEFAULT_NATIVE_REPLICATION_SOURCE)) {
+                  Assert.assertEquals((int) version, 3);
+                } else {
+                  Assert.assertEquals((int) version, 4);
```
nisargthakkar, line 943: nit: Can you change the check to:
```
                if (colo.equals("dc-1")) {
                  Assert.assertEquals((int) version, 4);
                } else {
                  Assert.assertEquals((int) version, 3);
                }
```

This is to avoid the test from failing if the number of colos are increased

nisargthakkar (commented):
Just one nit. Rest LGTM

nisargthakkar (approved):
LGTM. Great work, Adam. Thanks

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
adamxchen: I have no idea why I didn't take this suggestion but somehow revisit it again today, I think it makes sense to move up to 691 since `TARGETED_REGION_PUSH_ENABLED` should be set to true regardless the value of `TARGETED_REGION_PUSH_LIST`. I will add this to my next commit!

PR #418, opened 2023-05-12
[fast-client] Fix routing strategy issue when using request based metadata update

xunyin8:
## Summary

1. RequestBasedMetadata was forcing the fast client routing strategy to be HelixScatterGatherRoutingStrategy which is unintended. The routing strategy should be selected based on config.

2. HelixScatterGatherRoutingStrategy did not take InstanceHealthMonitor into consideration. Meaning it will continue to route requests to bad hosts when routing table is out of sync. This is witnessed when servers are going through deployment in our certification cluster and causing latency spikes.

3. ClientRoutingStrategy is difficult to set/instantiate in factory. Changing the config to an enum and instantiate the actual strategy in AbstractStoreMetadata instead.

## How was this PR tested?
Existing unit and integration tests. Added new unit test for HelixScatterGatherRoutingStrategy.

### DIFF
--- a/clients/venice-client/src/main/java/com/linkedin/venice/fastclient/meta/AbstractStoreMetadata.java
+++ b/clients/venice-client/src/main/java/com/linkedin/venice/fastclient/meta/AbstractStoreMetadata.java
@@ -7,6 +7,7 @@ import com.linkedin.venice.compression.CompressorFactory;
 import com.linkedin.venice.compression.VeniceCompressor;
 import com.linkedin.venice.fastclient.ClientConfig;
 import com.linkedin.venice.utils.Utils;
+import com.linkedin.venice.utils.concurrent.VeniceConcurrentHashMap;
 import java.io.IOException;
 import java.nio.ByteBuffer;
 import java.util.ArrayList;
@@ -23,8 +24,10 @@ public abstract class AbstractStoreMetadata implements StoreMetadata {
 
   public AbstractStoreMetadata(ClientConfig clientConfig) {
     this.instanceHealthMonitor = new InstanceHealthMonitor(clientConfig);
-    if (clientConfig.getClientRoutingStrategy() != null) {
-      this.routingStrategy = clientConfig.getClientRoutingStrategy();
+    ClientRoutingStrategyType clientRoutingStrategyType = clientConfig.getClientRoutingStrategyType();
+    if (clientRoutingStrategyType == ClientRoutingStrategyType.HELIX_ASSISTED) {
+      this.routingStrategy =
+          new HelixScatterGatherRoutingStrategy(instanceHealthMonitor, new VeniceConcurrentHashMap<>());
     } else {
       this.routingStrategy = new LeastLoadedClientRoutingStrategy(this.instanceHealthMonitor);
     }
--- a/clients/venice-client/src/main/java/com/linkedin/venice/fastclient/meta/AbstractStoreMetadata.java
+++ b/clients/venice-client/src/main/java/com/linkedin/venice/fastclient/meta/AbstractStoreMetadata.java
@@ -37,6 +37,13 @@ public abstract class AbstractStoreMetadata implements StoreMetadata {
     this.storeName = clientConfig.getStoreName();
   }
 
+  /**
+   * For testing only.
+   */
+  public void setRoutingStrategy(ClientRoutingStrategy routingStrategy) {
+    this.routingStrategy = routingStrategy;
+  }
+
   @Override
   public String getStoreName() {
     return storeName;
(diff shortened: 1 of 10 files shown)


clients/venice-client/src/main/java/com/linkedin/venice/fastclient/meta/HelixScatterGatherRoutingStrategy.java
```diff
@@ -15,8 +15,12 @@
 public class HelixScatterGatherRoutingStrategy implements ClientRoutingStrategy {
   private Map<String, Integer> helixGroupInfo;
   private List<Integer> groupIds;
+  private final InstanceHealthMonitor instanceHealthMonitor;
 
-  public HelixScatterGatherRoutingStrategy(Map<String, Integer> helixGroupInfo) {
+  public HelixScatterGatherRoutingStrategy(
+      InstanceHealthMonitor instanceHealthMonitor,
+      Map<String, Integer> helixGroupInfo) {
```
m-nagarajan, line 22: Function definition looks odd. This implies that `helixGroupInfo` is setup during initing `HelixScatterGatherRoutingStrategy`, but that is the test's flow. non test flow actually inits  `HelixScatterGatherRoutingStrategy` with `new VeniceConcurrentHashMap<>()` and then uses `updateHelixGroupInfo` to replace that during the initial warmup. so I feel that it's cleaner like below
```
public HelixScatterGatherRoutingStrategy(InstanceHealthMonitor instanceHealthMonitor) {
....
}
```

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/meta/AbstractStoreMetadata.java
```diff
@@ -23,8 +24,10 @@ public abstract class AbstractStoreMetadata implements StoreMetadata {
 
   public AbstractStoreMetadata(ClientConfig clientConfig) {
     this.instanceHealthMonitor = new InstanceHealthMonitor(clientConfig);
-    if (clientConfig.getClientRoutingStrategy() != null) {
-      this.routingStrategy = clientConfig.getClientRoutingStrategy();
+    ClientRoutingStrategyType clientRoutingStrategyType = clientConfig.getClientRoutingStrategyType();
```
m-nagarajan, line 27: Let this return `LEAST_LOADED` as default strategy if its not explicitly configured and probably a switch case to setup `this.routingStrategy`. What do you think?

m-nagarajan (commented):
Thanks @xunyin8 for finding the RC and fixing it. Left some comments.

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/meta/AbstractStoreMetadata.java
xunyin8: Sure, let me make that change.

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/meta/HelixScatterGatherRoutingStrategy.java
```diff
@@ -36,7 +40,7 @@ public List<String> getReplicas(long requestId, List<String> replicas, int requi
         if (selectedReplicas.size() == requiredReplicaCount) {
           return selectedReplicas;
         }
-        if (helixGroupInfo.get(replica) == groupId) {
+        if (helixGroupInfo.get(replica) == groupId && !instanceHealthMonitor.isInstanceBlocked(replica)) {
```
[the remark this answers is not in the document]
xunyin8: No, that'd defeat some purpose of Helix assisted routing. If we do want to add some notion of least loaded we also need to make some changes since HAR is mostly going to be used for batch gets and currently the instance monitor only tracks request not key count.

clients/venice-client/src/test/java/com/linkedin/venice/fastclient/meta/HelixScatterGatherRoutingStrategyTest.java
```diff
@@ -1,11 +1,13 @@
 package com.linkedin.venice.fastclient.meta;
 
+import static org.mockito.Mockito.*;
```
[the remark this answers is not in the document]
xunyin8: Good catch, I think my IDE keeps reverting the correct settings for auto import...

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/meta/HelixScatterGatherRoutingStrategy.java
xunyin8: Yes, let me change it.

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/meta/ClientRoutingStrategyType.java
```diff
@@ -0,0 +1,5 @@
+package com.linkedin.venice.fastclient.meta;
+
+public enum ClientRoutingStrategyType {
+  LEAST_LOADED, HELIX_ASSISTED
```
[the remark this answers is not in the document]
xunyin8: Done

m-nagarajan (approved):
LGTM, thanks Xun.

m-nagarajan (approved):

PR #419, opened 2023-05-12
[samza] Use Fast-Avro serializer in VeniceSystemProducer

sixpluszero:
## [samza] Use Fast-Avro serializer in VeniceSystemProducer

This PR tries to resolve slowness in VeniceSystemProducer's send() method. 
Through profiling we found that serializer is slow, in this PR we use FastAvro serializer instead.
Also schema-string based map is also found slow in get() as schema.toString() is heavy. We rely on FastAvro Factory's internal map cache to speed up. 

## How was this PR tested?

TODO


FelixGV (approved):
Thanks Jialin!

PR #422, opened 2023-05-15
[router] Check if a store has writes enabled in a request_topic call

nisargthakkar:
## Check if a store has writes enabled in a request_topic call

When routers receive a request_topic API call, they will check if the store has writes enabled

## How was this PR tested?

Added unit test. GH CI


majisourav99 (approved):
LGTM. Thanks

PR #423, opened 2023-05-16
[WIP] Added client-side computation support for thin-client

anaberezhnov:

WIP, DO NOT MERGE! This is just to share the direction I'm going. Structure-wise the code is far from what I'd like to see. Also it misses client-side configuration, proper unit-tests and stats to cover the new functionality. I'm pretty sure there're bugs as well.

## How was this PR tested?

Modified TestStreaming, TestRead. Also added placeholder for a new integration test ReadComputeTest.

### DIFF
--- a/clients/venice-thin-client/src/main/java/com/linkedin/venice/client/store/AbstractAvroStoreClient.java
+++ b/clients/venice-thin-client/src/main/java/com/linkedin/venice/client/store/AbstractAvroStoreClient.java
@@ -8,6 +8,7 @@ import static com.linkedin.venice.streaming.StreamingConstants.KEY_ID_FOR_STREAM
 import com.linkedin.avroutil1.compatibility.AvroCompatibilityHelper;
 import com.linkedin.avroutil1.compatibility.AvroVersion;
 import com.linkedin.venice.HttpConstants;
+import com.linkedin.venice.VeniceConstants;
 import com.linkedin.venice.client.exceptions.ServiceDiscoveryException;
 import com.linkedin.venice.client.exceptions.VeniceClientException;
 import com.linkedin.venice.client.exceptions.VeniceClientHttpException;
@@ -27,10 +28,16 @@ import com.linkedin.venice.client.store.transport.TransportClientResponse;
 import com.linkedin.venice.client.store.transport.TransportClientStreamingCallback;
 import com.linkedin.venice.compression.CompressionStrategy;
 import com.linkedin.venice.compression.CompressorFactory;
+import com.linkedin.venice.compute.ComputeOperationUtils;
 import com.linkedin.venice.compute.ComputeRequestWrapper;
+import com.linkedin.venice.compute.ReadComputeOperator;
+import com.linkedin.venice.compute.protocol.request.ComputeOperation;
+import com.linkedin.venice.compute.protocol.request.enums.ComputeOperationType;
 import com.linkedin.venice.compute.protocol.response.ComputeResponseRecordV1;
 import com.linkedin.venice.exceptions.VeniceException;
 import com.linkedin.venice.exceptions.VeniceNoStoreException;
+import com.linkedin.venice.read.ComputationStrategy;
+import com.linkedin.venice.read.RequestType;
 import com.linkedin.venice.read.protocol.response.MultiGetResponseRecordV1;
 import com.linkedin.venice.read.protocol.response.streaming.StreamingFooterRecordV1;
 import com.linkedin.venice.schema.SchemaData;
@@ -67,8 +74,10 @@ import java.util.function.BiFunction;
 import java.util.function.Function;
 import java.util.function.Supplier;
 import org.apache.avro.Schema;
+import org.apache.avro.generic.GenericData;
 import org.apache.avro.generic.GenericRecord;
 import org.apache.avro.io.BinaryEncoder;
+import org.apache.avro.specific.SpecificRecord;
 import org.apache.commons.codec.binary.Hex;
 import org.apache.commons.io.IOUtils;
 import org.apache.logging.log4j.LogManager;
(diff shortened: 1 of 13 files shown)


clients/venice-thin-client/src/main/java/com/linkedin/venice/client/store/AbstractAvroStoreClient.java
```diff
@@ -884,6 +926,17 @@ public void onHeaderReceived(Map<String, String> headers) {
...
+      if (responseTypeHeader != null) {
+        responseType = Integer.parseInt(responseTypeHeader);
+      }
+      if (responseFormatProvider != null) {
+        ResponseFormat format = responseFormatProvider.apply(RequestType.valueOf(responseType));
+        this.envelopeDeserializerFunc = format.envelopeDeserializerFunc();
+        this.recordDeserializerFunc = format.recordDeserializerFunc();
+        this.indexRetrievalFunc = format.indexRetrievalFunc();
+        this.streamingFooterRecordDeserializer = format.streamingFooterRecordDeserializer();
+      }
```
gaojieliu, line 939: This piece seems to belong the constructor function.
Do you think `responseFormatProvider` could change after the initialization?

services/venice-router/src/main/java/com/linkedin/venice/router/api/VenicePathParser.java
```diff
@@ -177,6 +179,23 @@ public VenicePath parseResourceUri(String uri, HTTP_REQUEST request) throws Rout
               routerConfig.getSmartLongTailRetryAbortThresholdMs(),
               routerConfig.isComputeFastAvroEnabled(),
               routerConfig.getLongTailRetryMaxRouteForMultiKeyReq());
+          String supportedComputationStrategy =
+              request.headers().get(HttpConstants.VENICE_SUPPORTED_COMPUTATION_STRATEGY);
+          ComputationStrategy computationStrategy =
+              !storeRepository.isReadComputationEnabled(storeName) && supportedComputationStrategy != null
+                  ? ComputationStrategy.valueOf(Integer.parseInt(supportedComputationStrategy))
+                  : ComputationStrategy.SERVER_SIDE_COMPUTATION;
```
gaojieliu, line 187: So if (!storeRepository.isReadComputationEnabled(storeName) && supportedComputationStrategy == null), shall we throw exception here?

I feel the logic here is a little confusing..
I assume the following should be right:
1. If store supports read compute, always use server-side compute.
2. If store doesn't support read compute, and if client-side compute is available, use multiget (client-side compute).
3. if store doesn't support read compute and client-side compute is not available, throw exception.

gaojieliu (commented):
There is a small perf concern with this approach:
Say some store won't enable read compute for a long time, but it starts using Venice read compute api.
The client will always construct compute request, which includes result schema and Router will need to convert the compute path to multiget path after deserializing the compute request.
This overhead can be avoided with more complex client-server communication, such as read compute API will send a multiget request when the previous response tells the store doesn't have read compute enabled, when the response tells the client that store read compute is enabled at runtime, the client can switch to the compute request. 
It is good to still keep the router-side logic in the PR, but I think the client can be smarter.


Also from the code structure perspective, I think we can add a layer inside venice-thin-client to handle the detection of server-side read compute support and automatically convert the request/response.
Essentially, the conversion logic can be put in the new layer since Venice Client is using multi-layer approach, which is slightly different from DaVinci, where most of the logic is in the same layer.
What do you think?

anaberezhnov:
> There is a small perf concern with this approach: Say some store won't enable read compute for a long time, but it starts using Venice read compute api. The client will always construct compute request, which includes result schema and Router will need to convert the compute path to multiget path after deserializing the compute request. This overhead can be avoided with more complex client-server communication, such as read compute API will send a multiget request when the previous response tells the store doesn't have read compute enabled, when the response tells the client that store read compute is enabled at runtime, the client can switch to the compute request. It is good to still keep the router-side logic in the PR, but I think the client can be smarter.
> 
> Also from the code structure perspective, I think we can add a layer inside venice-thin-client to handle the detection of server-side read compute support and automatically convert the request/response. Essentially, the conversion logic can be put in the new layer since Venice Client is using multi-layer approach, which is slightly different from DaVinci, where most of the logic is in the same layer. What do you think?

A more client-heavy implementation is the approach I was thinking about initially. I went with the opposite because it allows a service operator to quickly change settings without bouncing clients or relying on some periodic check logic. With the current approach, the main overhead is the request conversion on the router side. We can minimize it by converging MultiGetRouterRequestKeyV1 & ComputeRouterRequestKeyV1 to avoid new object allocation. These are exactly the same records, just with a different type names.

services/venice-router/src/main/java/com/linkedin/venice/router/api/VenicePathParser.java
anaberezhnov: We're on the same page in terms of expected behavior. In case `!storeRepository.isReadComputationEnabled(storeName) && supportedComputationStrategy == null` this block won't have any effect and the exception will be thrown from the original check in `VeniceDispatcher::dispatch`. I suspect it's possible to move that check here, but need to see first how it will affect metric tracking, etc.

anaberezhnov:
@CodiumAI-Agent reflect

anaberezhnov:
@CodiumAI-Agent /improve

clients/venice-thin-client/src/main/java/com/linkedin/venice/client/store/AbstractAvroStoreClient.java
```diff
@@ -8,6 +8,7 @@
 import com.linkedin.avroutil1.compatibility.AvroCompatibilityHelper;
 import com.linkedin.avroutil1.compatibility.AvroVersion;
 import com.linkedin.venice.HttpConstants;
+import com.linkedin.venice.VeniceConstants;
```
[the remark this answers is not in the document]
anaberezhnov: @CodiumAI-Agent /review

anaberezhnov:
@CodiumAI-Agent /review

PR #427, opened 2023-05-17
[all][controller][router] Fix several Controller API and ControllerClient bugs. Make routers fetch store name from query params

nisargthakkar:
## Fix several Controller API and ControllerClient bugs. Make routers fetch store name from query params

1. Return proper `ErrorType` in controller calls when store is not found
2. Updated some response messages and included `ErrorType` in many cases where we can infer it
3. Extract proper exception and `ErrorType` from API calls that use `D2Client`
4. Make `D2ControllerClient.discoverCluster` not throw exceptions and follow the same behavior as `ControllerClient.discoverCluster`
5. Make all variations of `D2ControllerClientFactory.discoverAndConstructControllerClient` behave consistently
6. Make `D2ServiceDiscoveryResponse` take over functionality from `D2ServiceDiscoveryResponseV2`. Marked `D2ServiceDiscoveryResponseV2` as deprecated
7. In `discover_cluster` API calls, routers now handle accept the store name via query params too.
    1. Previously, routers only supported accessing store name as a URL path param `/discover_cluster/<store_name>`
    2. Controllers use `/discover_cluster?store_name=<store_name>`
    3. So, `ControllerClient` sends a request with query param first and trying if the host is a controller. If there are errors, the `ControllerClient` tries to send a request with store name as a path param
    4. This change removes the need for a 2nd round trip as routers will also be able to handle the query in the first attempt
8. When metadata requests in router fail due to unsupported paths, return 400 instead of 500
9. Handle controller response errors in `ControllerTransport` as gracefully as possible at the moment. Previously, all error responses were marked as `BAD_REQUEST`

## How was this PR tested?

GH CI


internal/venice-client-common/src/main/java/com/linkedin/venice/D2/D2ClientUtils.java
```diff
@@ -112,7 +114,11 @@ public static RestResponse sendD2GetRequest(String requestPath, D2Client client)
     try {
       response = client.restRequest(createD2GetRequest(requestPath)).get();
     } catch (Exception e) {
-      throw new VeniceException("D2 client failed to sent request, " + requestPath, e);
+      Throwable t = ExceptionUtils.getRecursiveCause(e, RestException.class);
+      if (t == null) {
+        throw new VeniceException("D2 client failed to send request, " + requestPath, e);
+      }
+      response = ((RestException) t).getResponse();
```
adamxchen, line 121: It make sense to get the true root cause of the exception but why do we not throw exception now? Is `RestException` okay to not return error?

adamxchen (commented):
Overall looks great. One minor comment. Thank you!

internal/venice-client-common/src/main/java/com/linkedin/venice/D2/D2ClientUtils.java
nisargthakkar: If a D2 request fails with any reason other than a 2XX, it gets wrapped in a `RestException` ([Link](https://github.com/linkedin/rest.li/blob/master/r2-core/src/main/java/com/linkedin/r2/transport/http/common/HttpBridge.java#L70-L83)). We also return 4XX and 5XX in our codes.

In most cases, we return errors in `ErrorType` field in `ControllerResponse` along with 400 status code that we want to handle gracefully

services/venice-controller/src/main/java/com/linkedin/venice/controller/server/StoresRoutes.java
```diff
@@ -827,7 +835,7 @@ public void internalHandle(Request request, MultiStoreTopicsResponse veniceRespo
           });
           veniceResponse.setTopics(deletableTopicsList);
         } catch (Exception e) {
-          veniceResponse.setError("Failed to list deletable store topics. Message: " + e.getMessage());
+          veniceResponse.setError("Failed to list deletable store topics. Message: " + e.getMessage(), e);
```
sixpluszero, line 838: nit: Just wondering if we use ", e" pattern, do we need to log error message in the error log?

services/venice-controller/src/main/java/com/linkedin/venice/controller/server/StoragePersonaRoutes.java
```diff
@@ -77,7 +78,8 @@ public void internalHandle(Request request, StoragePersonaResponse veniceRespons
         } catch (Exception e) {
           veniceResponse.setError(
               "Failed when getting persona " + personaName + ". Exception type: " + e.getClass().toString()
-                  + ". Detailed message = " + e.getMessage());
+                  + ". Detailed message = " + e.getMessage(),
```
sixpluszero, line 81: Hmm, I see there is a lot of this kind of error message pattern "customized error msg + Exception type = xxx, Detailed message = xxx", wondering if we could do it in a single method and use it everywhere

services/venice-controller/src/main/java/com/linkedin/venice/controller/server/StoresRoutes.java
nisargthakkar: Here, `setError("...", e)` will only extract the `ErrorType` from the exception. It will be sent in the response to the client. The log is for the server-side logging. https://github.com/linkedin/venice/blob/main/internal/venice-client-common/src/main/java/com/linkedin/venice/controllerapi/ControllerResponse.java#L68

nisargthakkar: I just re-read the code. It seems the message will get appended twice. Will update it

services/venice-controller/src/main/java/com/linkedin/venice/controller/server/StoragePersonaRoutes.java
nisargthakkar: Yeah, we can

sixpluszero (approved):
Looks good to me, just the part around 400/500 error handling
// cluster from both cluster and router, we send the path "/discover_cluster?storename=$storeName" at first,
          // if it does not work, try "/discover_cluster/$storeName"
looks weird to me. Not sure why we had this different path, but I guess we can deal with it later.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/AvroGenericDaVinciClient.java
```diff
@@ -39,7 +39,7 @@
 import com.linkedin.venice.client.store.transport.TransportClient;
```
nisargthakkar, line 39: > Looks good to me, just the part around 400/500 error handling // cluster from both cluster and router, we send the path "/discover_cluster?storename=$storeName" at first, // if it does not work, try "/discover_cluster/$storeName" looks weird to me. Not sure why we had this different path, but I guess we can deal with it later.

I'll check this out a bit more before merging

nisargthakkar: @sixpluszero You were right. This would have broken in production. When routers cannot handle a metadata request, they currently return 500 error code. I will handle this and make it return 400.

I will also remove the specific 400/500 check in ControllerClient

internal/venice-client-common/src/main/java/com/linkedin/venice/controllerapi/D2ServiceDiscoveryResponseV2.java
```diff
@@ -3,31 +3,13 @@
 import com.fasterxml.jackson.annotation.JsonIgnore;
 
 
+@Deprecated
```
sixpluszero, line 6: Just curious - why don't we just nuke it?

services/venice-router/src/main/java/com/linkedin/venice/router/MetaDataHandler.java
```diff
@@ -158,63 +156,67 @@ public void channelRead0(ChannelHandlerContext ctx, HttpRequest req) throws IOEx
...
+          handleRequestTopic(ctx, helper, req);
+          break;
+        default:
+          // SimpleChannelInboundHandler automatically releases the request after channelRead0 is done.
+          // since we're passing it on to the next handler, we need to retain an extra reference.
+          ReferenceCountUtil.retain(req);
+          ctx.fireChannelRead(req);
+      }
+    } catch (VeniceHttpException e) {
+      setupResponseAndFlush(HttpResponseStatus.valueOf(e.getHttpStatusCode()), e.getMessage().getBytes(), false, ctx);
```
sixpluszero, line 219: My understanding is you add this try catch here. Just wondering what would happen in the past if handler throws exception in Router request?

services/venice-router/src/main/java/com/linkedin/venice/router/MetaDataHandler.java
```diff
@@ -372,54 +374,45 @@ private void handleUpdateSchemaLookup(ChannelHandlerContext ctx, VenicePathParse
...
     String clusterName = config.get().getCluster();
     String d2Service = getD2ServiceByClusterName(clusterName);
     if (StringUtils.isEmpty(d2Service)) {
       String errorMsg = "D2 service for store: " + storeName + " doesn't exist";
-      setupErrorD2DiscoveryResponseAndFlush(NOT_FOUND, errorMsg, headers, ctx);
+      setupErrorD2DiscoveryResponseAndFlush(NOT_FOUND, errorMsg, ctx);
       return;
     }
-    String serverD2Service = getServerD2ServiceByClusterName(clusterName);
-    if (headers.contains(D2_SERVICE_DISCOVERY_RESPONSE_V2_ENABLED)) {
```
sixpluszero, line 393: Hmm, I got a question to confirm for this V2->V1 migration: If an existing client is sending d2 discovery request via v2 and received by new Router, this should not work?

internal/venice-common/src/main/java/com/linkedin/venice/controllerapi/ControllerClient.java
```diff
@@ -966,7 +978,7 @@ public ClusterStaleDataAuditResponse getClusterStaleStores(String clusterName, S
           ClusterStaleDataAuditResponse.class);
     } catch (Exception e) {
       return makeErrorResponse(
-          "controllerapi:ControllerClient:getClusterStaleStores - ",
+          "Failed to get stale stores in cluster: " + clusterName,
```
sixpluszero, line 981: Thanks for adding this change. Make it more readable.

sixpluszero (commented):
Thanks Nisarg, I took another pass, overall I think it looks good and thanks for cleaning up code. My major question was the comment about v2->v1 migration, I am not sure if this will break things between different version of components. Feel free to explain a bit in the comment.

internal/venice-client-common/src/main/java/com/linkedin/venice/controllerapi/D2ServiceDiscoveryResponseV2.java
nisargthakkar: It can cause `ClassNotFoundException` due to transitive dependencies if we remove it now. We need to adopt a proper practice to note deprecation versions and then remove deprecated classes in major

services/venice-router/src/main/java/com/linkedin/venice/router/MetaDataHandler.java
nisargthakkar: Previously, router framework would send a 500 error to client with the exception message. But here, we set a proper status code in the response

services/venice-router/src/main/java/com/linkedin/venice/router/MetaDataHandler.java
nisargthakkar: Previously, We only return the `zkAddress` and `kafkaBootstrapServers` when `d2.service.discovery.response.v2.enabled` is set to true in request headers. This is defined in D2ServiceDiscoveryResponseV2.java.

This was added when jackson in clients couldn't handle missing fields. We have made the code handle missing fields in json by default since https://github.com/linkedin/venice/commit/e8e13025.

There are 2 evidences to support that this won't break actual users:
1. We didn't face issues when we added `errorType`
2. We have added `serverD2Service` as a field in `D2ServiceDiscoveryResponse`

When we were implementing `serverD2Service`, I had verified that no one was using the legacy clients that couldn't handle unknown fields

nisargthakkar: With this change, serializing and deserializing using `D2ServiceDiscoveryResponse` and `D2ServiceDiscoveryResponseV2` will work the same since `D2ServiceDiscoveryResponseV2 extends D2ServiceDiscoveryResponse` and doesn't add any fields of it's own

PR #428, opened 2023-05-17
[server] Make sure containsTopic check spent desired amount of time on retries

sixpluszero:
## [server] Make sure containsTopic check spent desired amount of time on retries

The contains topic check expects to spent at most 180s with 3 retries, however with current impl, it will spent a few hundred ms on 3 fast attempts and fail as we don't wait enough time to begin the next retry attempt.
This PR uses DelayFunction in FailSafe module to add a new retry mechanism to make sure after each attempt, it will wait until the desired attempt total duration is spent before it begins next attempt.
With the setup, in the worst case (topic does not exist), it will be:
Attempt 0, wait until T=60s, Retry 1, wait until T=120s, Retry 2, wait until T=180s, Retry 3, Fail loudly.

## How was this PR tested?

Add a new unit test to make sure new try mechanism is working as expected.

### DIFF
--- a/internal/venice-client-common/src/main/java/com/linkedin/venice/utils/RetryUtils.java
+++ b/internal/venice-client-common/src/main/java/com/linkedin/venice/utils/RetryUtils.java
@@ -257,9 +257,9 @@ public class RetryUtils {
       List<Class<? extends Throwable>> retryFailureTypes) {
     RetryPolicy<Object> retryPolicy = new RetryPolicy<>().handle(retryFailureTypes)
         .withDelay(
-            (result, failure, context) -> (durationPerAttempt.toMillis() > context.getElapsedAttemptTime().toMillis()
-                ? Duration.ofMillis(durationPerAttempt.toMillis() - context.getElapsedAttemptTime().toMillis())
-                : Duration.ZERO))
+            (result, failure, context) -> durationPerAttempt.compareTo(context.getElapsedAttemptTime()) > 0
+                ? durationPerAttempt.minus(context.getElapsedAttemptTime())
+                : Duration.ZERO)
         .withMaxRetries(maxRetry);
     retryPolicy.onFailedAttempt(RetryUtils::logAttemptWithFailure);
     return Failsafe.with(retryPolicy).get(supplier::get);
--- a/internal/venice-client-common/src/main/java/com/linkedin/venice/utils/RetryUtils.java
+++ b/internal/venice-client-common/src/main/java/com/linkedin/venice/utils/RetryUtils.java
@@ -257,8 +257,9 @@ public class RetryUtils {
       List<Class<? extends Throwable>> retryFailureTypes) {
     RetryPolicy<Object> retryPolicy = new RetryPolicy<>().handle(retryFailureTypes)
         .withDelay(
-            (result, failure, context) -> Duration
-                .ofMillis(durationPerAttempt.toMillis() - context.getElapsedAttemptTime().toMillis()))
+            (result, failure, context) -> (durationPerAttempt.toMillis() > context.getElapsedAttemptTime().toMillis()
+                ? Duration.ofMillis(durationPerAttempt.toMillis() - context.getElapsedAttemptTime().toMillis())
+                : Duration.ZERO))
         .withMaxRetries(maxRetry);
     retryPolicy.onFailedAttempt(RetryUtils::logAttemptWithFailure);
     return Failsafe.with(retryPolicy).get(supplier::get);
--- a/internal/venice-client-common/src/main/java/com/linkedin/venice/utils/RetryUtils.java
+++ b/internal/venice-client-common/src/main/java/com/linkedin/venice/utils/RetryUtils.java
@@ -240,6 +240,30 @@ public class RetryUtils {
     return Failsafe.with(retryPolicy).get(supplier::get);
   }
 
+  /**
+   * Execute a {@link Supplier} with maximum retries and fix duration per retry attempt. If all attempts are made and
+   * still no success, the last thrown exception will be thrown.
+   *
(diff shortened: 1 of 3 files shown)


internal/venice-client-common/src/test/java/com/linkedin/venice/utils/RetryUtilsTest.java
```diff
@@ -266,4 +266,36 @@ public void testExponentialBackoffDelayRetryOnSupplier() {
...
+        .thenReturn(2);
+
+    Assert.assertEquals(
+        (int) RetryUtils.executeWithMaxRetriesAndFixedAttemptDuration(
+            () -> obj.getAnInteger() + 1,
+            2,
+            Duration.ofMillis(100),
+            Arrays.asList(IllegalStateException.class, IllegalArgumentException.class)),
+        3);
+    verify(obj, times(3)).getAnInteger();
```
m-nagarajan, line 298: can we add some sort of check that it waited for at least 2* `durationPerAttempt` to finish 3 attempts? probably also by providing a bigger `durationPerAttempt`.

m-nagarajan (commented):
Thanks Jialin for the changes. Added some comments.

internal/venice-client-common/src/main/java/com/linkedin/venice/utils/RetryUtils.java
```diff
@@ -240,6 +240,30 @@ public static <T> T executeWithMaxAttemptAndExponentialBackoff(
...
+   */
+  public static <T> T executeWithMaxRetriesAndFixedAttemptDuration(
+      VeniceCheckedSupplier<T> supplier,
+      final int maxRetry,
+      Duration durationPerAttempt,
+      List<Class<? extends Throwable>> retryFailureTypes) {
+    RetryPolicy<Object> retryPolicy = new RetryPolicy<>().handle(retryFailureTypes)
+        .withDelay(
+            (result, failure, context) -> Duration
+                .ofMillis(durationPerAttempt.toMillis() - context.getElapsedAttemptTime().toMillis()))
```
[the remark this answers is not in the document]
sixpluszero: Hmm looks like there is no such API in retry policy that limits single attempt time. Maybe we add a simple safeguard to check and wait 0 ms if it is this case.

internal/venice-client-common/src/test/java/com/linkedin/venice/utils/RetryUtilsTest.java
sixpluszero: Hmm what do you mean by this? If it is 3rd attempt then it should have waited 2*durationPerAttempt? Are you saying that the execution time might be > 100ms so no wait in between?

sixpluszero: Added the timeout to 1s

m-nagarajan: I meant to add a time check before and after `executeWithMaxRetriesAndFixedAttemptDuration` to measure that the total time taken is > 2 secs here? Some way to measure that it did wait for 1secs (including the execution time) between attempts.

internal/venice-client-common/src/main/java/com/linkedin/venice/utils/RetryUtils.java
```diff
@@ -240,6 +240,31 @@ public static <T> T executeWithMaxAttemptAndExponentialBackoff(
...
+   */
+  public static <T> T executeWithMaxRetriesAndFixedAttemptDuration(
+      VeniceCheckedSupplier<T> supplier,
+      final int maxRetry,
+      Duration durationPerAttempt,
+      List<Class<? extends Throwable>> retryFailureTypes) {
+    RetryPolicy<Object> retryPolicy = new RetryPolicy<>().handle(retryFailureTypes)
+        .withDelay(
+            (result, failure, context) -> (durationPerAttempt.toMillis() > context.getElapsedAttemptTime().toMillis()
+                ? Duration.ofMillis(durationPerAttempt.toMillis() - context.getElapsedAttemptTime().toMillis())
```
m-nagarajan, line 261: Just curious: Is there anything here that prevents the actual task to be stuck forever? something like `maxDuration`.

gaojieliu: What is the issue we are facing?
`ContainsTopic` check stuck forever?
I guess there is no way for RetryUtil to control the actual runtime of task unless, the RetryUtil spawns a thread for every call and setup a timer against it.

sixpluszero: I don't think retryutil fully control the attempted task. Also, in our case it is containsTopic check, with succeed/failed very very fast.
The issue we are facing here is: We actually want to wait up to 180s for containsTopic check for Kafka metadata. However, since each check failed pretty fas and we were using exponential backoff retry strategy and started from 100ms, and it will be like:
check1 (some ms), wait 100ms, check 2 (some ms), wait 200ms, check 3 -> fail loudly within a second or so. 
I use the new API to make sure each attempt will be a full 60s and we retry 3 times as there is not point to aggressively query the Kafka.

sixpluszero: Thanks Lei's comment on the nit, will address it.

m-nagarajan: Makes sense. Will mark as resolved.

m-nagarajan (approved):
LGTM, Thanks Jialin.

lluwm (approved):
LGTM. Thanks Jialin!

PR #429, opened 2023-05-17
[changelog-consumer][client] Add seek by time API's to Venice Changelog Consumer

ZacAttack:
## Add Seek by time API's to Venice Changelog Consumer
This was a client feature request.  We expect underlying pubsub implementations to provide this functionality, so it seemed a low hanger.  Expected semantics of the behavior of the new API's s documented in the interface javadoc attached to this PR.

## How was this PR tested?
Added new integration test cases

Yes, but only extends the interface.  Will not impact current users.

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumer.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumer.java
@@ -1,8 +1,10 @@
 package com.linkedin.davinci.consumer;
 
 import com.linkedin.venice.annotation.Experimental;
+import com.linkedin.venice.exceptions.VeniceException;
 import com.linkedin.venice.pubsub.api.PubSubMessage;
 import java.util.Collection;
+import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 
@@ -106,6 +108,27 @@ public interface VeniceChangelogConsumer<K, V> {
    * @throws a VeniceException if subscribe operation failed for any of the partitions
    */
   CompletableFuture<Void> subscribeAll();
+
+  /**
+   * Seek to the provided timestamps for the specified partitions.
+   *
+   * Note, this API can only be used to seek on nearline data applied to the current serving version in Venice.
+   * This will not seek on data transmitted via Batch Push. If the provided timestamp is lower then the earliest
+   * timestamp on a given stream, the earliest event will be returned.
+   *
+   * @param timeStamps a map keyed by a partition ID, and the timestamp checkpoints to seek for each partition.
+   * @return
+   * @throws VeniceException if seek operations failed for any of the specified partitions.
+   */
+  CompletableFuture<Void> seekToTimestamp(Map<Integer, Long> timeStamps);
+
+  /**
+   * Seek to the specified timestamp for all subscribed partitions. See {@link #seekToTimestamp(Map)} for more information.
+   *
+   * @return a future which completes when the operation has succeeded for all partitions.
+   * @throws VeniceException if seek operation failed for any of the partitions.
+   */
+  CompletableFuture<Void> seekToTimestamp(Long timestamp);
 
   /**
    * Stop ingesting messages from a set of partitions for a specific store.
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumer.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumer.java
@@ -114,7 +114,9 @@ public interface VeniceChangelogConsumer<K, V> {
    *
    * Note, this API can only be used to seek on nearline data applied to the current serving version in Venice.
    * This will not seek on data transmitted via Batch Push. If the provided timestamp is lower then the earliest
-   * timestamp on a given stream, the earliest event will be returned.
+   * timestamp on a given stream, the earliest event will be returned. THIS WILL NOT SEEK TO DATA WHICH WAS APPLIED
+   * ON A PREVIOUS VERSION.  You should never seek back in time to a timestamp which is smaller than the current time -
+   * rewindTimeInSeconds configured in the hybrid settings for this Venice store.
    *
    * @param timeStamps a map keyed by a partition ID, and the timestamp checkpoints to seek for each partition.
    * @return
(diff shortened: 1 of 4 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumerImpl.java
```diff
@@ -302,6 +303,36 @@ public CompletableFuture<Void> subscribeAll() {
     return this.subscribe(allPartitions);
   }
 
+  @Override
+  public CompletableFuture<Void> seekToTimestamp(Map<Integer, Long> timeStamps) {
+    // Get the latest change capture topic
+    storeRepository.refresh();
+    Store store = storeRepository.getStore(storeName);
+    int currentVersion = store.getCurrentVersion();
+    String topic = Version.composeKafkaTopic(storeName, currentVersion) + ChangeCaptureView.CHANGE_CAPTURE_TOPIC_SUFFIX;
```
haoxu07, line 312: One thing needs to be clarified here is: is it possible this timestamp user will derive from the time when user is doing RT producing. As we query offset based on this timestamp for kafka topic data generate by venice, so there might be time gap here.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumerImpl.java
```diff
@@ -302,6 +303,36 @@ public CompletableFuture<Void> subscribeAll() {
     return this.subscribe(allPartitions);
   }
 
+  @Override
+  public CompletableFuture<Void> seekToTimestamp(Map<Integer, Long> timeStamps) {
```
haoxu07, line 307: Nit:  `timestamps` instead of `timeStamps`. Or maybe the api name `seekToTimestamps` ?

haoxu07 (commented):
Thanks for the quick action! Just want to clarify on the input source.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumerImpl.java
ZacAttack: It's a good question.  The problem is that it's not going to be very easy to seek by user timestamp.  User timestamp need not line up with wall clock time or ordering.  So in order for this to work reliably, we're gonna have to scan the entirety of the topic to find a record with a matching timestamp. Moreover, events can share a timestamp, so it's kind of vague where you'd like to land.

Add to it, we can't even binary search topics for the right record for the above reason and the inability to do math on offsets in Xinfra.

We can document it in the javadoc, but I don't see this working easily.  I think wallclock is best we can do given the above limitations.  Maybe later we can add a 'findVeniceLogicalTimestamp' kind of api?  Not sure.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumerImpl.java
ZacAttack: Sure

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumerImpl.java
haoxu07: Yeah, we can clarify on the java doc. As there is always time gap between the topic produced by venice and topic produced by users. Searching through kafka records by our own is bit complicated, so I think we do not need to support this at this moment.

ZacAttack: Jotting this down for future work after thinking on this more, some points: 

- With the current seek API's a user could duplicate the logic we'd just put into the client with the current API's (filtering until you find what you need via seekToEndOfPush).  

- We don't expose produce metadata right now in the pubsubMessage since it's contained the KafkaMessageEnvelope that we strip away.  So for a future api where we'd seek to some producerTimestamp we'll probably need to backfill that piece of information to the PubSubMessage along with the new interface api's.

- Doing this for a batch of partitions might be a bit confusing if a set of producers aren't using wall clock time, but some other input from an upstream source, maybe it should be discouraged?

- Assuming RT write rate is small, and the hybrid rewind isn't enormous, it might not be that much data that need be consumed client side to find the right offset.  Maybe for most cases it would actually be reasonable?  Not sure.

ZacAttack: Nevermind we do provide the timestamp in the pubsubmessage, and looks like it's the producer timestamp, so from the consumer side with the current api's I think this would work nicely.

We'd later need to add a getter for upstream producer timestamp.

ZacAttack: Updated javadoc to reflect this.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumer.java
```diff
@@ -107,6 +109,35 @@
...
+   * timestamp on a given stream, the earliest event will be returned. THIS WILL NOT SEEK TO DATA WHICH WAS APPLIED
+   * ON A PREVIOUS VERSION.  You should never seek back in time to a timestamp which is smaller than the current time -
+   * rewindTimeInSeconds configured in the hybrid settings for this Venice store.
+   *
+   * The timestamp passed to this function should be associated to timestamps processed by this interface. The timestamp
+   * returned by {@link PubSubMessage.getPubSubMessageTime()} refers to the time when Venice processed the event, and
+   * calls to this method will seek based on that sequence of events.  Note: it bears no relation to timestamps provided by
+   * upstream producers when writing to Venice where a user may optionally provide a timestamp at time of producing a record.
+   *
+   * @param timestamps a map keyed by a partition ID, and the timestamp checkpoints to seek for each partition.
```
sixpluszero, line 127: Why do we have different TS for different partition? Shouldn't the user of the consumer client seek to a specific timestamp for all the partitions they are interested in?

haoxu07 (approved):

PR #430, opened 2023-05-17
[router] Add Property keys for configuring SSL for RouterServer.main()

ZacAttack:
## Add Property keys for configuring SSL for RouterServer.main()

VeniceRouter.main() was not honoring passed in SSL parameters when running, opting to instead use the test environment settings by relying on a self signed certificate.  This PR preserves this default behavior (as it is useful in testing), but adds the ability to override this behavior so that custom options can be passed.

## How was this PR tested?
running normal test suite which relied on this path.

No

### DIFF
--- a/internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
@@ -405,7 +405,7 @@ public class ConfigKeys {
    * provided via other properties. This should only be used for testing and defaults to true when running
    * RouterServer.main().
    */
-  public static final String ROUTER_USE_LOCAL_SSL_SETTINGS = "router.ssl.local.settings";
+  public static final String ROUTER_USE_LOCAL_SSL_SETTINGS = "router.local.ssl";
 
   /**
    * This instructs the router to open an ssl port. This defaults to true.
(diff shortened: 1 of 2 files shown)


services/venice-router/src/main/java/com/linkedin/venice/router/RouterServer.java
```diff
@@ -1,5 +1,8 @@
 package com.linkedin.venice.router;
 
+import static com.linkedin.venice.CommonConfigKeys.*;
```
adamxchen, line 3: nit: not import all? Intellij keeps reverting the config

internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
```diff
@@ -400,6 +400,18 @@ private ConfigKeys() {
   public static final String ROUTER_MAX_READ_CAPACITY = "router.max.read.capacity";
   public static final String ROUTER_QUOTA_CHECK_WINDOW = "router.quota.check.window";
 
+  /**
+   * This instructs the router to start running with self signed TLS certificates as opposed to those
+   * provided via other properties. This should only be used for testing and defaults to true when running
+   * RouterServer.main().
+   */
+  public static final String ROUTER_USE_LOCAL_SSL_SETTINGS = "router.ssl.local.settings";
```
adamxchen, line 408: Can we call it "router.local.ssl"?

adamxchen (dismissed):
Minor comments on the naming and import statements. Thanks!

services/venice-router/src/main/java/com/linkedin/venice/router/RouterServer.java
ZacAttack: freaking intellij

internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
ZacAttack: Sure

PR #431, opened 2023-05-17
[fc][test] Add a config for batchGet to use streamingBatchGet

m-nagarajan:
## Summary
1. The first version of `batchGet` in fast client used single get in a loop to support a customer request for two-key `batchGet`. `streamingBatchGet` support was added later but the `batchGet` API still uses the old method. Adding a config to allow switching between the two implementations. The current default is single get based batchGet, but once the `streamingBatchget` is validated, the default should be changed to `streamingBatchget` based batchGet, and probably remove the single get based batchGet support.
2. Added this config to integration tests, added unit tests and verified via metrics for the same. 

## How was this PR tested?
GH CI

Introduces fast client config `setUseStreamingBatchGetAsDefault(boolean)` which if set to `true` will use the streaming batchGet get based batchGet. Default is `false` which uses the existing single get based batchGet.


clients/venice-client/src/main/java/com/linkedin/venice/fastclient/ClientConfig.java
```diff
@@ -96,7 +106,8 @@ private ClientConfig(
       boolean isVsonStore,
       StoreMetadataFetchMode storeMetadataFetchMode,
       D2Client d2Client,
-      String clusterDiscoveryD2Service) {
+      String clusterDiscoveryD2Service,
+      boolean batchGetDefaultsToStreamingBatchGet) {
```
xunyin8, line 110: NIT: the variable sounds a little verbose, maybe rename it to `useStreamingBatchGetAsDefault`?

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/ClientConfig.java
```diff
@@ -206,6 +217,12 @@ private ClientConfig(
         && this.storeMetadataFetchMode != StoreMetadataFetchMode.SERVER_BASED_METADATA) {
       throw new VeniceClientException("Helix assisted routing is only available with server based metadata enabled");
     }
+    this.batchGetDefaultsToStreamingBatchGet = batchGetDefaultsToStreamingBatchGet;
+    if (this.batchGetDefaultsToStreamingBatchGet) {
+      LOGGER.info("BatchGet will use streamingBatchGet");
```
xunyin8, line 222: NIT: let's avoid camel case or code variables in logs if possible, can just replace with `Batch get will use streaming batch get implementation`.

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/ClientConfig.java
```diff
@@ -367,6 +388,7 @@ public String getClusterDiscoveryD2Service() {
     private StoreMetadataFetchMode storeMetadataFetchMode = StoreMetadataFetchMode.DA_VINCI_CLIENT_BASED_METADATA;
     private D2Client d2Client;
     private String clusterDiscoveryD2Service;
+    private boolean batchGetDefaultsToStreamingBatchGet = true;
```
xunyin8, line 391: Should we keep the default to `false` until we get some validation on performance in cert?

m-nagarajan: sounds good. We should probably clean up the single get based batchGet after making sure the streaming batch get works. Code looks sort of messy now.

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/ClientConfig.java
m-nagarajan: good suggestion, renamed it.

m-nagarajan (commented):
Thanks @xunyin8 for the review. Addressed the comments and also added unit tests for the diffCoverage. We probably need to add a new PR adding some more unit tests especially validating the counters.

xunyin8 (approved):
Thanks

PR #433, opened 2023-05-18
[router] some SSLConfig fields are not considered

nicoloboschi:

It's not possible to configure the ssl field `sslRequireClientCerts` via configuration. 
With the new authentication provider, mTLS auth is not required when TLS is enabled and therefore must be configurable in the router component.

Changes: 
* Copy the value `ssl.needs.client.cert` from the properties
* Reduce log for `SSLPeerUnverifiedException` when `needClientAuth=false`
* Allow storeAcl (both for Router and Server) to perform acl without a client certificate 

## How was this PR tested?
Added unit tests and tested manually.

`ssl.needs.client.cert` is now considered in the router SSL config

### DIFF
--- a/internal/alpini/netty4/alpini-netty4-base/src/main/java/com/linkedin/alpini/netty4/ssl/SslInitializer.java
+++ b/internal/alpini/netty4/alpini-netty4-base/src/main/java/com/linkedin/alpini/netty4/ssl/SslInitializer.java
@@ -49,6 +49,7 @@ import javax.annotation.Nonnull;
 import javax.net.ssl.SSLEngine;
 import javax.net.ssl.SSLHandshakeException;
 import javax.net.ssl.SSLParameters;
+import javax.net.ssl.SSLPeerUnverifiedException;
 import javax.net.ssl.SSLSession;
 import javax.net.ssl.SSLSessionContext;
 import javax.security.auth.x500.X500Principal;
@@ -293,6 +294,11 @@ public class SslInitializer extends ChannelInitializer<Channel> {
         } else {
           _handshakesFailed.increment();
         }
+        boolean needClientAuth = true;
+        SSLParameters sslParameters = _sslFactory.getSSLParameters();
+        if (sslParameters != null) {
+          needClientAuth = sslParameters.getNeedClientAuth();
+        }
         if (ctx.channel().hasAttr(SSL_HANDSHAKE_START_TS)) {
           SslHandler handler = ctx.pipeline().get(SslHandler.class);
           SSLSession session = handler.engine().getSession();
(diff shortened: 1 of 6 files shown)


internal/alpini/netty4/alpini-netty4-base/src/main/java/com/linkedin/alpini/netty4/ssl/SslInitializer.java
```diff
@@ -307,6 +313,10 @@ public void userEventTriggered(ChannelHandlerContext ctx, Object evt) throws Exc
                 }
               }
             }
+          } catch (SSLPeerUnverifiedException ex) {
+            if (needClientAuth) {
+              LOG.warn("Unable to obtain remote CN for {}", ctx.channel().remoteAddress(), ex);
```
ZacAttack, line 318: I wonder if we can simplify this slightly.  It occurs to me that if needClientAuth is true and we have an unverified principal in the session request, then the request was likely a failure as I would expect the sslEngine to have kicked in and enforced the rule.

So instead of doing the above chunk where we get the clientAuth/sslparameters/etc. we could probably just do 

`if (!succeed) {
  LOG.warn(....
}`

since we already spent the time getting that information and the log would help us out to puzzle a failure since it would correlate with the failed request.

internal/venice-common/src/main/java/com/linkedin/venice/acl/handler/StoreAclHandler.java
```diff
@@ -48,15 +49,22 @@ protected String extractStoreName(String resourceName) {
...
        * In HTTP/2, the SSLHandler is k  parent channel pipeline and the child channels won't have the SSL Handler.
        */
       sslHandler = ctx.channel().parent().pipeline().get(SslHandler.class);
     }
-    return SslUtils.getX509Certificate(sslHandler.engine().getSession().getPeerCertificates()[0]);
+    final Certificate[] peerCertificates;
+    try {
+      peerCertificates = sslHandler.engine().getSession().getPeerCertificates();
+    } catch (SSLPeerUnverifiedException exception) {
+      return null;
```
ZacAttack, line 64: This I'm a bit less sure about.  The intention of this server handler is such that it it's only actually part of the request pipeline if ACL's are enabled.  If ACL's are enabled, then we want this thing to be able to extract identity's of actors and scream at us if it can't do that.

Now, what's bothersome today is that it's taken for granted that the only way to determine the identity of a requester is via client certificate.  What'd be ideal is to make that identity extraction pluggable (so instead of it being strictly peer cert, maybe it's a crypto header or something idk).

So, my thinking is that we can maybe leave this alone for now with the intention of a bigger refactor based on @eolivelli's vip and you can get unblocked by just not enabling ACLs?

internal/venice-client-common/src/test/java/com/linkedin/venice/security/SSLConfigTest.java
```diff
@@ -0,0 +1,30 @@
+package com.linkedin.venice.security;
+
+import static org.testng.Assert.*;
```
adamxchen, line 3: nit: please do not import all.

adamxchen:
Hey @nicoloboschi , do you have any bandwidth to iterate the PR as Zac suggested above? We'd like to keep the open PR list short so folks can focus on high-priority ones and get changes into production ASAP.

PR #435, opened 2023-05-18
[controller] Report push error if too many DVC instances are dead

majisourav99:
## [controller] Report push error if too many DVC instances are dead"

If many instances are not reporting  heartbeat status to DaVinci pushstatus store, we can assume the ongoing push is not healthy. In this PR if more than 50% instances are not alive, and its been  more 5 mins its returns ERROR and fails the push job.

## How was this PR tested?

CI

### DIFF
--- a/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
+++ b/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
@@ -94,6 +94,7 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.Properties;
 import java.util.Set;
+import java.util.concurrent.TimeUnit;
 import org.apache.avro.Schema;
 import org.apache.avro.generic.GenericData;
 import org.apache.avro.mapred.AvroInputFormat;
@@ -557,6 +558,7 @@ public class VenicePushJob implements AutoCloseable {
     long storeRewindTimeInSeconds;
     Schema keySchema;
     HybridStoreConfig hybridStoreConfig;
+    StoreResponse storeResponse;
   }
 
   protected StoreSetting storeSetting;
@@ -2294,6 +2296,7 @@ public class VenicePushJob implements AutoCloseable {
     if (storeResponse.isError()) {
       throw new VeniceException("Can't get store info. " + storeResponse.getError());
     }
+    storeSetting.storeResponse = storeResponse;
     storeSetting.storeStorageQuota = storeResponse.getStore().getStorageQuotaInByte();
     storeSetting.isSchemaAutoRegisterFromPushJobEnabled =
         storeResponse.getStore().isSchemaAutoRegisterFromPushJobEnabled();
@@ -2590,7 +2593,7 @@ public class VenicePushJob implements AutoCloseable {
    * If any datacenter report an explicit error status, we throw an exception and fail the job. However, datacenters
    * with COMPLETED status will be serving new data.
    */
-  private void pollStatusUntilComplete(
+  void pollStatusUntilComplete(
       Optional<String> incrementalPushVersion,
       ControllerClient controllerClient,
       PushJobSetting pushJobSetting,
@@ -2611,6 +2614,7 @@ public class VenicePushJob implements AutoCloseable {
      * no more than {@link DEFAULT_JOB_STATUS_IN_UNKNOWN_STATE_TIMEOUT_MS}.
      */
     long unknownStateStartTimeMs = 0;
+    long pollStartTimeMs = System.currentTimeMillis();
 
     String topicToMonitor = getTopicToMonitor(topicInfo, pushJobSetting);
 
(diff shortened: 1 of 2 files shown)


sixpluszero:
There are a few tests failing. Could you share some motivation for this change? Is it a short-cut to fail job? Because at the end it will fail the job anyway.

majisourav99:
> 

Today it will skip recording any status for dead instances, so it push starts failing instances (`isInstanceAlive` returns false), we wont recognized as it failed. This PR sets some threshold on how many instances are going bad and fails the push

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
adamxchen, line 1: Not related to your change but I feel the logic of deciding the DVC health status is a bit counterintuitive. For example,
```
        if (status == middleStatus) {
          if (allInstancesCompleted && isInstanceAlive) {
            allInstancesCompleted = false;
          }
```
Logically it seems right but would be easier to reason about we do it in opposite way

adamxchen:
> > 
> 
> Today it will skip recording any status for dead instances, so it push starts failing instances (`isInstanceAlive` returns false), we wont recognized as it failed. This PR sets some threshold on how many instances are going bad and fails the push

Just want to make sure I understand. So, DVC instances can be DEAD or ERROR. We report failure for ERROR but nothing for DEAD?

majisourav99:
> > > 
> > 
> > 
> > Today it will skip recording any status for dead instances, so it push starts failing instances (`isInstanceAlive` returns false), we wont recognized as it failed. This PR sets some threshold on how many instances are going bad and fails the push
> 
> Just want to make sure I understand. So, DVC instances can be DEAD or ERROR. We report failure for ERROR but nothing for DEAD?

Right,   not `live` or dead instances are ignored completely. And aliveness is detected a instances reporting heartbeats. So if a bunch of hosts die from OOM, they will stop reporting heartbeat and will be ignored in the status report

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -5432,10 +5436,31 @@ private Pair<ExecutionStatus, String> getDaVinciPushStatusAndDetails(
...
+      if (lastUpdateTime != null) {
+        if (lastUpdateTime + DAVINCI_DEAD_INSTANCE_REPORT_ERROR_WAIT < System.currentTimeMillis()) {
+          storeVersionToDVCDeadInstanceTimeMap.remove(storeName + version.getNumber());
+          return new Pair<>(
+              ExecutionStatus.ERROR,
+              " To many dead instances " + (totalInstanceCount - liveInstanceCount) + ", total instances "
+                  + totalInstanceCount);
+        }
+      } else {
+        storeVersionToDVCDeadInstanceTimeMap.put(storeName + version.getNumber(), System.currentTimeMillis());
```
gaojieliu, line 5456: This doesn't look right since the function is triggered when AZ poling the job status.
If the AZ job polling interval is bigger than 5 mins, the update interval recorded in `storeVersionToDVCDeadInstanceTimeMap` will be higher than 5 mins.

Also if the `liveInstanceCount` > 0.5 * totalInstanceCount later, shall we reset the tracking map?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -5432,10 +5436,31 @@ private Pair<ExecutionStatus, String> getDaVinciPushStatusAndDetails(
         incompletePartition.add(partitionId);
       }
     }
+
+    if (totalInstanceCount == 0) {
+      return new Pair<>(completeStatus, "No DaVinci instance found yet.");
+    }
+
+    // Report error if too many davinci instances are not alive for over 5 mins
+    if (liveInstanceCount < 0.5 * totalInstanceCount) {
```
gaojieliu, line 5445: Potentially we need to specify some minimal threshold, such as the totalInstanceCount needs to be large than some number (such as 3 or 5), otherwise, a rolling deployment may trigger a false positive event.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
gaojieliu: @sixpluszero is working on another change to refactor the deciding logic of push job completed, where Controller would spawn some async thread to poll push status system store, and I think this idea can be implemented better on top of Jialin's change.

majisourav99: I see default AZ job polling period  is 5 mins, so we should definitely wait for more than than, may be 10mins or so. I will check with Jialin on this. Are we trying to converge both AZ job poll status and backend push status in the same code path?

PR #437, opened 2023-05-19
[controller] Offline push status will also monitor Da Vinci push status store if enabled

sixpluszero:
## [controller] Offline push status will also monitor Da Vinci push status store if enabled
Offline Push monitor today only monitors Venice Server push status (changes from ZK). However, when Da Vinci is also reporting ingestion status to this store push job, only VPJ is monitoring the DVC push status store but not push monitor. When the failure from DVC is deterministic (OOM detected by memory limiter (not yet implemented) for example), it will lead to infinite retry of DVC future version ingestion (as the bad version becomes current version after backend push completed).

This PR fixes the issue by introducing a new PushStatusCollector class that has a scheduler thread to periodically scan DVC push status enabled store and aggregate server push status with Da Vinci push status to come up with the aggregated status. With this change, when controller detects server push status change, it will not directly mark overall push as COMPLETED unless DVC result is also COMPLETED. 

## How was this PR tested?

Added a new unit test.

### DIFF
--- a/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerConfig.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerConfig.java
@@ -62,6 +62,8 @@ import static com.linkedin.venice.ConfigKeys.MIN_NUMBER_OF_UNUSED_KAFKA_TOPICS_T
 import static com.linkedin.venice.ConfigKeys.NATIVE_REPLICATION_FABRIC_ALLOWLIST;
 import static com.linkedin.venice.ConfigKeys.NATIVE_REPLICATION_FABRIC_WHITELIST;
 import static com.linkedin.venice.ConfigKeys.NATIVE_REPLICATION_SOURCE_FABRIC;
+import static com.linkedin.venice.ConfigKeys.OFFLINE_PUSH_MONITOR_DAVINCI_PUSH_STATUS_ENABLED;
+import static com.linkedin.venice.ConfigKeys.OFFLINE_PUSH_MONITOR_DAVINCI_PUSH_STATUS_SCAN_INTERVAL_IN_SECONDS;
 import static com.linkedin.venice.ConfigKeys.PARENT_CONTROLLER_MAX_ERRORED_TOPIC_NUM_TO_KEEP;
 import static com.linkedin.venice.ConfigKeys.PARENT_CONTROLLER_WAITING_TIME_FOR_CONSUMPTION_MS;
 import static com.linkedin.venice.ConfigKeys.PARENT_KAFKA_CLUSTER_FABRIC_LIST;
@@ -186,6 +188,9 @@ public class VeniceControllerConfig extends VeniceControllerClusterConfig {
    * To decide whether to initialize push status store related components.
    */
   private final boolean isDaVinciPushStatusStoreEnabled;
+
+  private final boolean offlinePushMonitorDaVinciPushStatusEnabled;
+  private final int offlinePushMonitorDaVinciPushStatusScanIntervalInSeconds;
 
   private final boolean zkSharedDaVinciPushStatusSystemSchemaStoreAutoCreationEnabled;
 
@@ -410,6 +415,10 @@ public class VeniceControllerConfig extends VeniceControllerClusterConfig {
     this.pushStatusStoreHeartbeatExpirationTimeInSeconds =
         props.getLong(PUSH_STATUS_STORE_HEARTBEAT_EXPIRATION_TIME_IN_SECONDS, TimeUnit.MINUTES.toSeconds(10));
     this.isDaVinciPushStatusStoreEnabled = props.getBoolean(PUSH_STATUS_STORE_ENABLED, false);
+    this.offlinePushMonitorDaVinciPushStatusEnabled =
+        props.getBoolean(OFFLINE_PUSH_MONITOR_DAVINCI_PUSH_STATUS_ENABLED, true) && isDaVinciPushStatusStoreEnabled;
+    this.offlinePushMonitorDaVinciPushStatusScanIntervalInSeconds =
+        props.getInt(OFFLINE_PUSH_MONITOR_DAVINCI_PUSH_STATUS_SCAN_INTERVAL_IN_SECONDS, 30);
     this.zkSharedDaVinciPushStatusSystemSchemaStoreAutoCreationEnabled =
         props.getBoolean(CONTROLLER_ZK_SHARED_DAVINCI_PUSH_STATUS_SYSTEM_SCHEMA_STORE_AUTO_CREATION_ENABLED, false);
     this.systemStoreAclSynchronizationDelayMs =
@@ -715,6 +724,14 @@ public class VeniceControllerConfig extends VeniceControllerClusterConfig {
 
   public boolean isDaVinciPushStatusStoreEnabled() {
     return isDaVinciPushStatusStoreEnabled;
+  }
+
+  public int getOfflinePushMonitorDaVinciPushStatusScanIntervalInSeconds() {
+    return offlinePushMonitorDaVinciPushStatusScanIntervalInSeconds;
+  }
+
+  public boolean isOfflinePushMonitorDaVinciPushStatusEnabled() {
+    return offlinePushMonitorDaVinciPushStatusEnabled;
   }
 
   public boolean isZkSharedDaVinciPushStatusSystemSchemaStoreAutoCreationEnabled() {
--- a/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerConfig.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerConfig.java
@@ -64,6 +64,7 @@ import static com.linkedin.venice.ConfigKeys.NATIVE_REPLICATION_FABRIC_WHITELIST
 import static com.linkedin.venice.ConfigKeys.NATIVE_REPLICATION_SOURCE_FABRIC;
 import static com.linkedin.venice.ConfigKeys.OFFLINE_PUSH_MONITOR_DAVINCI_PUSH_STATUS_ENABLED;
 import static com.linkedin.venice.ConfigKeys.OFFLINE_PUSH_MONITOR_DAVINCI_PUSH_STATUS_SCAN_INTERVAL_IN_SECONDS;
+import static com.linkedin.venice.ConfigKeys.OFFLINE_PUSH_MONITOR_DAVINCI_PUSH_STATUS_SCAN_THREAD_NUMBER;
 import static com.linkedin.venice.ConfigKeys.PARENT_CONTROLLER_MAX_ERRORED_TOPIC_NUM_TO_KEEP;
 import static com.linkedin.venice.ConfigKeys.PARENT_CONTROLLER_WAITING_TIME_FOR_CONSUMPTION_MS;
 import static com.linkedin.venice.ConfigKeys.PARENT_KAFKA_CLUSTER_FABRIC_LIST;
@@ -191,6 +192,8 @@ public class VeniceControllerConfig extends VeniceControllerClusterConfig {
 
   private final boolean offlinePushMonitorDaVinciPushStatusEnabled;
   private final int offlinePushMonitorDaVinciPushStatusScanIntervalInSeconds;
+
+  private final int offlinePushMonitorDaVinciPushStatusScanThreadNumber;
 
   private final boolean zkSharedDaVinciPushStatusSystemSchemaStoreAutoCreationEnabled;
 
@@ -419,6 +422,9 @@ public class VeniceControllerConfig extends VeniceControllerClusterConfig {
         props.getBoolean(OFFLINE_PUSH_MONITOR_DAVINCI_PUSH_STATUS_ENABLED, true) && isDaVinciPushStatusStoreEnabled;
     this.offlinePushMonitorDaVinciPushStatusScanIntervalInSeconds =
         props.getInt(OFFLINE_PUSH_MONITOR_DAVINCI_PUSH_STATUS_SCAN_INTERVAL_IN_SECONDS, 30);
+    this.offlinePushMonitorDaVinciPushStatusScanThreadNumber =
+        props.getInt(OFFLINE_PUSH_MONITOR_DAVINCI_PUSH_STATUS_SCAN_THREAD_NUMBER, 4);
+
     this.zkSharedDaVinciPushStatusSystemSchemaStoreAutoCreationEnabled =
         props.getBoolean(CONTROLLER_ZK_SHARED_DAVINCI_PUSH_STATUS_SYSTEM_SCHEMA_STORE_AUTO_CREATION_ENABLED, false);
     this.systemStoreAclSynchronizationDelayMs =
@@ -732,6 +738,10 @@ public class VeniceControllerConfig extends VeniceControllerClusterConfig {
 
   public boolean isOfflinePushMonitorDaVinciPushStatusEnabled() {
     return offlinePushMonitorDaVinciPushStatusEnabled;
+  }
+
+  public int getOfflinePushMonitorDaVinciPushStatusScanThreadNumber() {
+    return offlinePushMonitorDaVinciPushStatusScanThreadNumber;
   }
 
   public boolean isZkSharedDaVinciPushStatusSystemSchemaStoreAutoCreationEnabled() {
--- a/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
@@ -5343,13 +5343,11 @@ public class VeniceHelixAdmin implements Admin, StoreCleaner {
     if (store.isDaVinciPushStatusStoreEnabled() && (versionNumber > 1 || incrementalPushVersion.isPresent())) {
       if (store.getVersion(versionNumber).isPresent()) {
         Version version = store.getVersion(versionNumber).get();
-        ExecutionStatusWithDetails daVinciStatusAndDetails = incrementalPushVersion.isPresent()
-            ? PushMonitorUtils.getDaVinciPushStatusAndDetails(
-                pushStatusStoreReader.orElse(null),
-                version.kafkaTopicName(),
-                version.getPartitionCount(),
-                incrementalPushVersion)
-            : monitor.getDaVinciPushStatus(version.kafkaTopicName(), version.getPartitionCount());
+        ExecutionStatusWithDetails daVinciStatusAndDetails = PushMonitorUtils.getDaVinciPushStatusAndDetails(
+            pushStatusStoreReader.orElse(null),
+            version.kafkaTopicName(),
+            version.getPartitionCount(),
+            incrementalPushVersion);
         ExecutionStatus daVinciStatus = daVinciStatusAndDetails.getStatus();
         String daVinciDetails = daVinciStatusAndDetails.getDetails();
         executionStatus = getOverallPushStatus(executionStatus, daVinciStatus);
--- a/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
@@ -152,10 +152,12 @@ import com.linkedin.venice.pubsub.adapter.kafka.consumer.ApacheKafkaConsumerAdap
 import com.linkedin.venice.pubsub.api.PubSubConsumerAdapterFactory;
 import com.linkedin.venice.pubsub.api.PubSubTopic;
 import com.linkedin.venice.pushmonitor.ExecutionStatus;
+import com.linkedin.venice.pushmonitor.ExecutionStatusWithDetails;
 import com.linkedin.venice.pushmonitor.KillOfflinePushMessage;
 import com.linkedin.venice.pushmonitor.OfflinePushStatus;
 import com.linkedin.venice.pushmonitor.PushMonitor;
 import com.linkedin.venice.pushmonitor.PushMonitorDelegator;
+import com.linkedin.venice.pushmonitor.PushMonitorUtils;
 import com.linkedin.venice.pushmonitor.PushStatusDecider;
 import com.linkedin.venice.pushmonitor.StatusSnapshot;
 import com.linkedin.venice.pushstatushelper.PushStatusStoreReader;
@@ -5341,10 +5343,13 @@ public class VeniceHelixAdmin implements Admin, StoreCleaner {
     if (store.isDaVinciPushStatusStoreEnabled() && (versionNumber > 1 || incrementalPushVersion.isPresent())) {
       if (store.getVersion(versionNumber).isPresent()) {
         Version version = store.getVersion(versionNumber).get();
-        Pair<ExecutionStatus, String> daVinciStatusAndDetails =
-            getDaVinciPushStatusAndDetails(version, incrementalPushVersion);
-        ExecutionStatus daVinciStatus = daVinciStatusAndDetails.getFirst();
-        String daVinciDetails = daVinciStatusAndDetails.getSecond();
+        ExecutionStatusWithDetails daVinciStatusAndDetails = PushMonitorUtils.getDaVinciPushStatusAndDetails(
+            pushStatusStoreReader.orElse(null),
+            version.kafkaTopicName(),
+            version.getPartitionCount(),
+            incrementalPushVersion);
+        ExecutionStatus daVinciStatus = daVinciStatusAndDetails.getStatus();
+        String daVinciDetails = daVinciStatusAndDetails.getDetails();
         executionStatus = getOverallPushStatus(executionStatus, daVinciStatus);
         if (details != null || daVinciDetails != null) {
           String overallDetails = "";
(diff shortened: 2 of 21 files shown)


services/venice-controller/src/main/java/com/linkedin/venice/pushmonitor/PushStatusCollector.java
```diff
@@ -0,0 +1,194 @@
...
+        pushStatus.setMonitoring(false);
+        pushCompletedHandler.accept(topicName);
+      } else if (isOverallPushError(serverStatus, daVinciStatus)) {
+        pushStatus.setMonitoring(false);
+        String overallErrorString = "";
+        if (serverStatus.getStatus().equals(ExecutionStatus.ERROR)) {
+          overallErrorString = overallErrorString + "Server push error: " + serverStatus.getDetails() + "\n";
+        }
+        if (daVinciStatus.getStatus().equals(ExecutionStatus.ERROR)) {
+          overallErrorString = overallErrorString + "Da Vinci push error: " + daVinciStatus.getDetails() + "\n";
```
majisourav99, line 124: can we concat the string instead of overwriting the previous status?

services/venice-controller/src/main/java/com/linkedin/venice/pushmonitor/PushMonitorUtils.java
```diff
@@ -0,0 +1,106 @@
...
+import com.linkedin.venice.utils.Pair;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+
+public class PushMonitorUtils {
```
gaojieliu, line 15: Can you add some javadoc in this new class?

services/venice-controller/src/main/java/com/linkedin/venice/pushmonitor/PushMonitorUtils.java
```diff
@@ -0,0 +1,106 @@
...
+            allInstancesCompleted = false;
+          }
+        } else if (status != completeStatus) {
+          if (allInstancesCompleted || allMiddleStatusReceived) {
+            if (isInstanceAlive) {
+              allInstancesCompleted = false;
+              allMiddleStatusReceived = false;
+              if (status == ExecutionStatus.ERROR) {
+                erroredInstance = Optional.of(entry.getKey().toString());
+                break;
```
gaojieliu, line 66: This logic is not easy to read.
I assume the non-live instance will be ignored, right?
can we refactor this to be ?
```
if (! isInstanceAlive) {
   continue;
}
....
```

services/venice-controller/src/main/java/com/linkedin/venice/pushmonitor/PushStatusCollector.java
```diff
@@ -0,0 +1,194 @@
...
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.function.BiConsumer;
+import java.util.function.Consumer;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+
+public class PushStatusCollector {
```
gaojieliu, line 20: Add some javadoc?

services/venice-controller/src/main/java/com/linkedin/venice/pushmonitor/PushStatusCollector.java
```diff
@@ -0,0 +1,194 @@
...
+    this.daVinciPushStatusScanPeriodInSeconds = daVinciPushStatusScanPeriodInSeconds;
+  }
+
+  public void start() {
+    offlinePushCheckScheduler
+        .scheduleAtFixedRate(this::scanDaVinciPushStatus, 0, daVinciPushStatusScanPeriodInSeconds, TimeUnit.SECONDS);
+  }
+
+  public void subscribeTopic(String topicName, int partitionCount) {
+    String storeName = Version.parseStoreFromKafkaTopicName(topicName);
```
gaojieliu, line 54: @haoxu07 
Do we need to use pubsub abstraction here?

services/venice-controller/src/main/java/com/linkedin/venice/pushmonitor/PushStatusCollector.java
```diff
@@ -0,0 +1,194 @@
...
+    }
+  }
+
+  public void unsubscribeTopic(String topicName) {
+    daVinciStoreTopicToPartitionCountMap.remove(topicName);
+  }
+
+  public void handleServerPushStatusUpdate(String topicName, ExecutionStatus executionStatus, String detailsString) {
+
+    TopicPushStatus topicPushStatus = daVinciStoreTopicToPartitionCountMap.get(topicName);
```
gaojieliu, line 72: If the DaVinci push status hasn't been collected, it will be `null`, and this class will act on the server push status right away, which may be wrong.

Can we have some extra logic there?
For example, if the DaVinci push status collection hasn't been done at all for a given topic, we should mark the DaVinci status as `unknown` or `progress`, and after collecting the DaVinci status at least once, we can update the aggregated status.

services/venice-controller/src/main/java/com/linkedin/venice/pushmonitor/PushStatusCollector.java
```diff
@@ -0,0 +1,194 @@
...
+      ReadWriteStoreRepository storeRepository,
+      PushStatusStoreReader pushStatusStoreReader,
+      Consumer<String> pushCompletedHandler,
+      BiConsumer<String, String> pushErrorHandler,
+      int daVinciPushStatusScanPeriodInSeconds) {
+    this.storeRepository = storeRepository;
+    this.pushStatusStoreReader = pushStatusStoreReader;
+    this.pushCompletedHandler = pushCompletedHandler;
+    this.pushErrorHandler = pushErrorHandler;
+    this.daVinciPushStatusScanPeriodInSeconds = daVinciPushStatusScanPeriodInSeconds;
```
gaojieliu, line 45: I am wondering whether we should introduce a feature flag to disable DVC push status collection or not.
In case DaVinci push status collection is problematic, it will affect all the pushes.
I am not sure, and what is your opinion on this?

gaojieliu (changes requested):

services/venice-controller/src/main/java/com/linkedin/venice/pushmonitor/PushStatusCollector.java
sixpluszero: Do you mean a controller level flag to enable/disable collection of DVC status?

services/venice-controller/src/main/java/com/linkedin/venice/pushmonitor/PushStatusCollector.java
sixpluszero: I think for this line, the map (I need to rename the map) is populated when subscribeTopic is called. It will create an object. 
Also TopicPushStatus's getDaVinciStaus is getOrDefault(topic, UNKNOWN) as you are suggesting here.

services/venice-controller/src/main/java/com/linkedin/venice/pushmonitor/PushStatusCollector.java
sixpluszero: Hmm Could you give an example here for the conditional concat string?

services/venice-controller/src/main/java/com/linkedin/venice/pushmonitor/PushStatusCollector.java
```diff
@@ -0,0 +1,194 @@
...
+  private final BiConsumer<String, String> pushErrorHandler;
+  private final Map<String, TopicPushStatus> daVinciStoreTopicToPartitionCountMap = new VeniceConcurrentHashMap<>();
+
+  private final PushStatusStoreReader pushStatusStoreReader;
+
+  private final ReadWriteStoreRepository storeRepository;
+
+  private final int daVinciPushStatusScanPeriodInSeconds;
+
+  private final ScheduledExecutorService offlinePushCheckScheduler = Executors.newScheduledThreadPool(1);
```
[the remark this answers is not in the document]
gaojieliu: yeah, one store version will keep many partitions, and there could be many ongoing pushes, the latency of one round of scan could be high.

services/venice-controller/src/main/java/com/linkedin/venice/pushmonitor/PushStatusCollector.java
gaojieliu: yeah.
Since now, the backend version swap will depend on the DaVinci status as well, and if DVC status collection is not working properly, the impact will be huge.

sixpluszero: I see, in order to achieve this - also above comment to make other stuff configurable, I am going to add more argument to AbstarctPushMonitor constructor arg list, hopefully it is not too long, and if that's the case we can consider refactor it in future PR.

sixpluszero: Actually, let me do it in this PR. Just passing in VeniceControllerConfig can solve a lot of trouble.

services/venice-controller/src/main/java/com/linkedin/venice/pushmonitor/PushMonitorUtils.java
```diff
@@ -0,0 +1,112 @@
...
+        liveInstanceCount++;
+        if (status == middleStatus) {
+          if (allInstancesCompleted) {
+            allInstancesCompleted = false;
+          }
+        } else if (status != completeStatus) {
+          if (allInstancesCompleted || allMiddleStatusReceived) {
+            allInstancesCompleted = false;
+            allMiddleStatusReceived = false;
+            if (status == ExecutionStatus.ERROR) {
```
lluwm, line 71: why the ERROR status detection needs to be in the `allInstancesCompleted || allMiddleStatusReceived` branch?

services/venice-controller/src/main/java/com/linkedin/venice/pushmonitor/PushMonitorUtils.java
```diff
@@ -0,0 +1,112 @@
...
+    if (incompleteSize > 0 && incompleteSize <= 5) {
+      details += ". Following partitions still not complete " + incompletePartition + ". live instances: "
+          + liveInstanceCount + " total instances : " + totalInstanceCount;
+    }
+    if (details.length() != 0) {
+      statusDetail = details;
+    }
+    if (completedPartitions == partitionCount) {
+      return new ExecutionStatusWithDetails(completeStatus, statusDetail);
+    } else if (allMiddleStatusReceived) {
```
lluwm, line 104: nit: we can simplify it a bit by removing the `else`, i.e.
`else if (allMiddleStatusReceived) {` ->  `if (allMiddleStatusReceived) {`

services/venice-controller/src/main/java/com/linkedin/venice/pushmonitor/PushStatusCollector.java
```diff
@@ -0,0 +1,231 @@
...
+
+  private boolean isOverallPushError(
+      ExecutionStatusWithDetails serverStatus,
+      ExecutionStatusWithDetails daVinciStatus) {
+    return serverStatus.getStatus().equals(ExecutionStatus.ERROR)
+        || daVinciStatus.getStatus().equals(ExecutionStatus.ERROR);
+  }
+
+  public void clear() {
+    topicToPushStatusMap.clear();
```
lluwm, line 177: Do we need to shutdown `pushStatusStoreScanExecutor` and `offlinePushCheckScheduler` here?

services/venice-controller/src/main/java/com/linkedin/venice/pushmonitor/AbstractPushMonitor.java
```diff
@@ -902,7 +918,7 @@ protected void handleOfflinePushUpdate(
     routingDataRepository.unSubscribeRoutingDataChange(pushStatus.getKafkaTopic(), this);
 
     if (status.equals(ExecutionStatus.COMPLETED)) {
-      handleCompletedPush(pushStatus);
+      pushStatusCollector.handleServerPushStatusUpdate(pushStatus.getKafkaTopic(), COMPLETED, null);
```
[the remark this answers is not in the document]
sixpluszero: but that is filling in different argument right?

services/venice-controller/src/main/java/com/linkedin/venice/pushmonitor/PushStatusCollector.java
```diff
@@ -0,0 +1,231 @@
...
+        pushCompletedHandler.accept(topicName);
+      } else if (executionStatus.equals(ExecutionStatus.ERROR)) {
+        pushErrorHandler.accept(topicName, detailsString);
+      }
+    } else {
+      // Update the server topic status in the data structure and wait for async DVC status scan thread to pick up.
+      topicToPushStatusMap.computeIfPresent(topicName, (topic, pushStatus) -> {
+        pushStatus.setServerStatus(new ExecutionStatusWithDetails(executionStatus, detailsString));
+        return pushStatus;
+      });
```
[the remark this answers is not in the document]
majisourav99: can we extract the handler outside in `AbstractPushMonitor`, ie not pass in the `Consumer` this class and handle all the work here only for `daVinciPushStatusScanEnabled` case. we can directly call handleCompletedPush/errorPush inside `handleOfflinePushUpdate` when the config is false.

services/venice-controller/src/main/java/com/linkedin/venice/pushmonitor/PushStatusCollector.java
sixpluszero: Looks like we cannot do it explicitly here. The other PushMonitor does not implement VeniceService interface and does not have place to stop() explicitly. The place to call clear() is stopMonitoringAllPushes(), where it is not stopping the service and could actually be called during controller state transition.
So far I think it is not easy to do that, but maybe we could refactor it in a separate PR I guess?

services/venice-controller/src/main/java/com/linkedin/venice/pushmonitor/PushStatusCollector.java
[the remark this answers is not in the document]
sixpluszero: Not exactly. When subscribeTopic is called we also check whether store.isPushStatusStoreEnabled() is true. (Even though in theory every store has companion push status store today), only when it is true do we monitor DVC pushes.

services/venice-controller/src/main/java/com/linkedin/venice/pushmonitor/PushStatusCollector.java
sixpluszero: Actually, let me double check whether it is easy to add some stop() like thing to PushMonitor

services/venice-controller/src/main/java/com/linkedin/venice/pushmonitor/PushStatusCollector.java
```diff
@@ -0,0 +1,231 @@
...
+      TopicPushStatus pushStatus;
+      try {
+        pushStatus = future.get();
+      } catch (Exception e) {
+        LOGGER.error("Caught exception when getting future result of push status.", e);
+        continue;
+      }
+      ExecutionStatusWithDetails daVinciStatus = pushStatus.getDaVinciStatus();
+      ExecutionStatusWithDetails serverStatus = pushStatus.getServerStatus();
+      if (serverStatus == null) {
```
majisourav99, line 120: curious, when this can happen?

sixpluszero: Do you mean serverStatus == null?
It will happen when DVC status scanner is scanning a topic but no server terminal status is reported yet.

services/venice-controller/src/main/java/com/linkedin/venice/pushmonitor/PushMonitorUtils.java
sixpluszero: I guess it is ok anywhere, we could move it out one layer I suppose from my current understanding.

services/venice-controller/src/main/java/com/linkedin/venice/pushmonitor/PushStatusCollector.java
```diff
@@ -0,0 +1,231 @@
...
+        continue;
+      }
+      LOGGER.info(
```
gaojieliu, line 123: I think we need to log the topic name here.

services/venice-controller/src/main/java/com/linkedin/venice/pushmonitor/AbstractPushMonitor.java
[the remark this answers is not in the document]
lluwm: Never mind, I was thinking we can simply use `status` for the argument in handleServerPushStatusUpdate, but seems that it would simply skip for status other than COMPLETED and ERROR

services/venice-controller/src/main/java/com/linkedin/venice/pushmonitor/PushMonitorUtils.java
sixpluszero: Refactored the whole block.

lluwm: I didn't pay attention that you just move this block of code around. Let't just keep it for now, the logic here is quite convoluted and we may want to clean it later.

services/venice-controller/src/main/java/com/linkedin/venice/pushmonitor/PushStatusCollector.java
```diff
@@ -0,0 +1,231 @@
...
+        }
+        pushErrorHandler.accept(pushStatus.getTopicName(), pushErrorDetailStringBuilder.toString());
+      }
+    }
+  }
+
+  public void handleServerPushStatusUpdate(String topicName, ExecutionStatus executionStatus, String detailsString) {
+
+    TopicPushStatus topicPushStatus = topicToPushStatusMap.get(topicName);
+    if ((!daVinciPushStatusScanEnabled) || topicPushStatus == null) {
```
gaojieliu, line 152: can this condition happen when `daVinciPushStatusScanEnabled` is `true` ?
```
topicPushStatus == null
```

services/venice-controller/src/main/java/com/linkedin/venice/pushmonitor/PushMonitorUtils.java
sixpluszero: No worries, looks like everyone struggle to read this part so I tried to understand it myself and refactor a bit in the latest commit.

services/venice-controller/src/main/java/com/linkedin/venice/pushmonitor/PushStatusCollector.java
```diff
@@ -0,0 +1,231 @@
...
+      ExecutionStatusWithDetails serverStatus,
+      ExecutionStatusWithDetails daVinciStatus) {
+    return serverStatus.getStatus().equals(ExecutionStatus.COMPLETED)
+        && daVinciStatus.getStatus().equals(ExecutionStatus.COMPLETED);
+  }
+
+  private boolean isOverallPushError(
+      ExecutionStatusWithDetails serverStatus,
+      ExecutionStatusWithDetails daVinciStatus) {
+    return serverStatus.getStatus().equals(ExecutionStatus.ERROR)
```
gaojieliu, line 171: Shall we do a `null` check against the params since the `TopicPushStatus` doesn't init these variables at construction time?
```
    private ExecutionStatusWithDetails serverStatus;
    private ExecutionStatusWithDetails daVinciStatus;
```

gaojieliu: Maybe we should initialize these two variables as `NEW`?

services/venice-controller/src/main/java/com/linkedin/venice/pushmonitor/PushStatusCollector.java
gaojieliu: It seems when the push status system store is not enabled in store-level, this condition can be true.

gaojieliu: Can you add some javadoc for the above scenario?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -5341,10 +5343,13 @@ private OfflinePushStatusInfo getOfflinePushStatusInfo(
...
-            getDaVinciPushStatusAndDetails(version, incrementalPushVersion);
-        ExecutionStatus daVinciStatus = daVinciStatusAndDetails.getFirst();
-        String daVinciDetails = daVinciStatusAndDetails.getSecond();
+        ExecutionStatusWithDetails daVinciStatusAndDetails = PushMonitorUtils.getDaVinciPushStatusAndDetails(
+            pushStatusStoreReader.orElse(null),
+            version.kafkaTopicName(),
+            version.getPartitionCount(),
+            incrementalPushVersion);
+        ExecutionStatus daVinciStatus = daVinciStatusAndDetails.getStatus();
+        String daVinciDetails = daVinciStatusAndDetails.getDetails();
```
gaojieliu, line 5352: I guess querying DaVinci status logic won't be required any more since `PushStatusCollector` will consolidate DaVinciStatus and Server Status and update `OfflinePush` znode, right?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerConfig.java
```diff
@@ -410,6 +418,13 @@ public VeniceControllerConfig(VeniceProperties props) {
     this.pushStatusStoreHeartbeatExpirationTimeInSeconds =
         props.getLong(PUSH_STATUS_STORE_HEARTBEAT_EXPIRATION_TIME_IN_SECONDS, TimeUnit.MINUTES.toSeconds(10));
     this.isDaVinciPushStatusStoreEnabled = props.getBoolean(PUSH_STATUS_STORE_ENABLED, false);
+    this.offlinePushMonitorDaVinciPushStatusEnabled =
```
gaojieliu, line 421: Can we consolidate these two configs?

services/venice-controller/src/main/java/com/linkedin/venice/pushmonitor/AbstractPushMonitor.java
```diff
@@ -500,7 +516,7 @@ public Pair<ExecutionStatus, String> getPushStatusAndDetails(String topic) {
           String errorMsg = "After waiting for " + elapsedTimeInSec + " seconds, resource assignment for: " + topic
               + " timed out, strategy=" + strategy.toString() + ", replicationFactor=" + replicationFactor + ", reason="
               + notReadyReason.get();
-          handleErrorPush(pushStatus, errorMsg);
```
gaojieliu, line 503: Refactoring idea:
It seems `handleErrorPush` and `handleCompletedPush` can be easily misused.
Maybe we should fold these two functions into `handleOfflinePushUpdate`.

services/venice-controller/src/main/java/com/linkedin/venice/pushmonitor/PushStatusCollector.java
sixpluszero: The check is actually done right before where it is called. I am wondering if this is actually confusing, if so I can actually fold it back to the caller method to make it more clear..

services/venice-controller/src/main/java/com/linkedin/venice/pushmonitor/PushStatusCollector.java
sixpluszero: Actually this is done in the latest commit.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerConfig.java
sixpluszero: It is actually serving different purpose.
For the above one, when it is enabled, the controller will be able to r/w from push status store.
For the below one which is newly added, the collector will consider DVC push status during offline push status check

gaojieliu: Emm..
When does Controller need to write to push status store?
For incremental push? or to materialize the push status system store at store creation time?

services/venice-controller/src/main/java/com/linkedin/venice/pushmonitor/PushMonitorUtils.java
```diff
@@ -0,0 +1,112 @@
...
+        ? ExecutionStatus.END_OF_INCREMENTAL_PUSH_RECEIVED
+        : ExecutionStatus.COMPLETED;
+    ExecutionStatus middleStatus = incrementalPushVersion.isPresent()
+        ? ExecutionStatus.START_OF_INCREMENTAL_PUSH_RECEIVED
+        : ExecutionStatus.END_OF_PUSH_RECEIVED;
+    Optional<String> erroredInstance = Optional.empty();
+    String storeName = Version.parseStoreFromKafkaTopicName(topicName);
+    int completedPartitions = 0;
+    int totalInstanceCount = 0;
+    int liveInstanceCount = 0;
```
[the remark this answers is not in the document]
sixpluszero: So, I read the code again, it is actually counting replicas of all partitions. It is not considering du-dup instance for the logging purpose. So maybe just rename the instance -> replica?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerConfig.java
sixpluszero: There is a call in getOfflinePushStatus to manipulate with inc push status. Although it is not very clear to me what it is doing exactly for now, will need to check. Also there is deleter that deletes push status.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
sixpluszero: I guess what you mean here is we can try to fetch from collector's current status? Probably yeah...need some refactoring though.

services/venice-controller/src/main/java/com/linkedin/venice/pushmonitor/AbstractPushMonitor.java
sixpluszero: Hmm, the handleErrorPush and handleCompletedPush now is only used as the callback of the PushStatusCollector.
handleOfflinePushUpdate is the entry point to call to PushStatusCollector

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
sixpluszero: Added...but I somehow feel like the logic between Admin/Monitor class(3 layers of basically one thing)/ new Collector is becoming more complicated... I don't know if that's also your feeling or not but it would be nice if we do the real refactor in some other PR to clean it up.

services/venice-controller/src/main/java/com/linkedin/venice/pushmonitor/AbstractPushMonitor.java
gaojieliu: I see, then we can keep them.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
gaojieliu: No, I was thinking querying `OfflinePushStatus` is good enough since the aggregated status on top of DaVinci/Server will be persisted in `OfflinePushStatus` eventually.

sixpluszero: But actually here for incremental push it is not the case right? It is not monitored there if I understand it correctly.

sixpluszero: And I am not sure my latest implementation of querying collector is good, I feel like it is making the logic more complicated.

gaojieliu: I see.
For incremental push, you are right.
For batch-only push, querying offline push status should be good enough.
I guess incremental push should use a different code path from batch push since it doesn't rely on the `offline push status` at all.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerConfig.java
sixpluszero: Synced offline, decided to leave it for now.

gaojieliu (approved):
Thanks a lot!

PR #440, opened 2023-05-22
[test] Fix flaky unit test RetryUtilsTest#testFixAttemptDurationOnSupplier

sixpluszero:
## [test] Fix flaky unit test RetryUtilsTest#testFixAttemptDurationOnSupplier

This PR fixes the flaky test by introducing some non-trivial time on the attempted action. With additional logging we found that the total time spent was 1999ms/2000ms, less than expected > 2000ms. This might be because the RetryUtils's calculation precision loss on wait time. In theory, the logic should be good enough, hence just make the action a bit longer to make sure 2 * 1 second wait + 3rd attempt > 2seconds.

## How was this PR tested?

Unit test.

### DIFF
--- a/internal/venice-client-common/src/test/java/com/linkedin/venice/utils/RetryUtilsTest.java
+++ b/internal/venice-client-common/src/test/java/com/linkedin/venice/utils/RetryUtilsTest.java
@@ -288,13 +288,11 @@ public class RetryUtilsTest {
         .thenThrow(IllegalStateException.class)
         .thenReturn(2);
     long startTime = System.currentTimeMillis();
-    Assert.assertEquals(
-        (int) RetryUtils.executeWithMaxRetriesAndFixedAttemptDuration(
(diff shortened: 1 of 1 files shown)


adamxchen:
Also feel free to skip other tests, GH actions to speed up the testing cycle

sixpluszero:
> Also feel free to skip other tests, GH actions to speed up the testing cycle

Yeah, thanks for this comment. Actually I think about this but I am afraid that I might not be able to reproduce the flakiness after I remove too much tests (although in theory flakiness should be there but I don't know for sure what leads to the flakiness). I will try to run a few more times to see.

FelixGV:
Note that it's not necessary to open a PR to run the tests. You need only push to a branch on your own fork of the repo, and the actions will run there, in the Actions tab of your repo. After iterating like this, you can then open a PR.

m-nagarajan (approved):
Thanks for the fixing this. LGTM.

PR #395, opened 2023-05-03
[admin-tool][changelog-client][controller][server][test] Replace legacy kafka consumer usage with PubSubConsumerAdaptor and some kafka config cleanups.

haoxu07:
## Summary, imperative, start upper case, don't end with a period

This PR focus on:  
1. Constructing kafka consumer with `PubSubConsumerAdaptor` for two components: Admin Tool and Changelog Client. This will simplify the adaption of other pub sub system in the future.  
2. Besides, add unit test for kafka consumption logics inside admin tool by replacing previous unit test: `TopicMessageFinderTest` with `TestAdminToolConsumption`. 
3. Some kafka configs cleanup, avoid directly using kafka configs in server and controller. 

## How was this PR tested?

internal CI 

## Does this PR introduce any user-facing changes?

- [x] No. You can skip the rest of this section.
- [ ] Yes. Make sure to explain your proposed changes and call out the behavior change.

PR #441, opened 2023-05-23
[changelog] Add pause/Resume API's to changelog consumer

ZacAttack:
## [changelog] Add pause/Resume API's to changelog consumer
This is needed to support poll() in venice systemconsumer interface where what is required is selective partition polling. Pause/Resume is one way to do it, and is something that was historically done for the KafkaSystemConsumer, but another way would be to implement the BlockingEnvelopeMap provided by the samza library. I think we can iterate on this as this approach would give better performance, however it gives it at the cost of a larger memory footprint on the consumer side.  So proposing to take this simple route initially and then iterate on it depending on what we need.

Resolves #XXX

## How was this PR tested?
Added new unit and integration tests.

Yes, but only new API's, old API's are left alone and should not be effected.


clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumerImpl.java
```diff
@@ -248,6 +248,52 @@ public CompletableFuture<Void> seekToEndOfPush(Set<Integer> partitions) {
     return internalSeek(partitions, topic, kafkaConsumer::seekToBeginning);
   }
 
+  @Override
+  public void pause() {
+    this.pause(
+        kafkaConsumer.assignment()
```
FelixGV, line 254: Why is this tightly coupled to Kafka, rather than the pub sub abstractions?

ZacAttack: [I'm waiting on a PR from Hao to get this out](https://github.com/linkedin/venice/pull/395).  The Pubsub abstraction has pause, so should be a one line replacement I think.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumer.java
```diff
@@ -58,6 +58,35 @@
    */
   CompletableFuture<Void> seekToEndOfPush(Set<Integer> partitions);
 
+  /**
+   * Pause the client on all subscriptions. See {@link #pause(Set)} for more information.
+   *
+   * @throws VeniceException if seek operation failed for any of the partitions.
```
haoxu07, line 64: Minor:    @throws `VeniceException` if `pause` operation failed for any of the partitions, serveral cases as below.

haoxu07 (changes requested):

haoxu07 (approved):

PR #442, opened 2023-05-23
[admin-tool] Support a new datetime argument in data-recovery monitor

lluwm:
This rb supports a new argument (datetime) in the admin-tool data-recovery monitor command. With this new argument, monitor command checks if there exists a successfully pushed version (current version) for the given store in the target region whose start time is newer than the input argument. If it does, monitor can report success, otherwise, it keeps polling/reporting a future version push job status until the future version is finished. During the real data recovery scenario, we can set the datetime argument to a timestamp at which we believe all related components (e.g. Kafka service) are recovered. If the argument is unspecified, we choose `now` as the default value.

## How was this PR tested?
- Unit test
- Internal CI test
- Manually tested against EI environment
example output (with masks)
```
lelu@lelu-mn1 venice % java -jar ./clients/venice-admin-tool/build/libs/venice-admin-tool-all.jar --monitor-data-recovery --url xxxxx --dest-fabric ei4 --stores xxxxx --interval 60
[WARN] Running admin tool without SSL.
Using default date time: 2023-05-23T22:41:25.948936
...
2023-05-23 15:47:27 INFO [DataRecoveryMonitor] [main] [store: xxxxx, msg: No ongoing offline pushes detected after given date time (2023-05-23T22:41:25.948936), keep polling]
2023-05-23 15:47:27 INFO [DataRecoveryWorker] [main] Total: 1, Succeeded: 0, Error: 0, Uncompleted: 1
2023-05-23 15:48:28 INFO [DataRecoveryMonitor] [main] [store: xxxxx, msg: ver: 16082, status: STARTED, uncompleted ptn: 4/4]
2023-05-23 15:48:28 INFO [DataRecoveryWorker] [main] Total: 1, Succeeded: 0, Error: 0, Uncompleted: 1
2023-05-23 15:49:27 INFO [DataRecoveryMonitor] [main] [store: xxxxx, msg: ver: 16082, status: STARTED, uncompleted ptn: 4/4]
2023-05-23 15:49:27 INFO [DataRecoveryWorker] [main] Total: 1, Succeeded: 0, Error: 0, Uncompleted: 1
...
2023-05-23 15:54:27 INFO [DataRecoveryMonitor] [main] [store: xxxxx, msg: ver: 16082, status: COMPLETED]
2023-05-23 15:54:27 INFO [DataRecoveryWorker] [main] Total: 1, Succeeded: 1, Error: 0, Uncompleted: 0
```

### DIFF
--- a/clients/venice-admin-tool/src/main/java/com/linkedin/venice/Command.java
+++ b/clients/venice-admin-tool/src/main/java/com/linkedin/venice/Command.java
@@ -17,6 +17,7 @@ import static com.linkedin.venice.Arg.CLUSTER;
 import static com.linkedin.venice.Arg.CLUSTER_DEST;
 import static com.linkedin.venice.Arg.CLUSTER_SRC;
 import static com.linkedin.venice.Arg.COMPRESSION_STRATEGY;
+import static com.linkedin.venice.Arg.DATETIME;
 import static com.linkedin.venice.Arg.DEBUG;
 import static com.linkedin.venice.Arg.DERIVED_SCHEMA;
 import static com.linkedin.venice.Arg.DERIVED_SCHEMA_ID;
@@ -454,7 +455,7 @@ public enum Command {
   ),
   MONITOR_DATA_RECOVERY(
       "monitor-data-recovery", "Monitor data recovery progress for a group of stores",
-      new Arg[] { URL, STORES, DEST_FABRIC }, new Arg[] { INTERVAL }
+      new Arg[] { URL, STORES, DEST_FABRIC }, new Arg[] { INTERVAL, DATETIME }
   );
 
   private final String commandName;
--- a/clients/venice-admin-tool/src/main/java/com/linkedin/venice/Command.java
+++ b/clients/venice-admin-tool/src/main/java/com/linkedin/venice/Command.java
@@ -455,7 +455,7 @@ public enum Command {
   ),
   MONITOR_DATA_RECOVERY(
       "monitor-data-recovery", "Monitor data recovery progress for a group of stores",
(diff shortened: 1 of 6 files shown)


clients/venice-admin-tool/src/main/java/com/linkedin/venice/Command.java
```diff
@@ -454,7 +455,7 @@ public enum Command {
   ),
   MONITOR_DATA_RECOVERY(
       "monitor-data-recovery", "Monitor data recovery progress for a group of stores",
-      new Arg[] { URL, STORES, DEST_FABRIC }, new Arg[] { INTERVAL }
+      new Arg[] { URL, STORES, DEST_FABRIC }, new Arg[] { INTERVAL, DATETIME }
```
shuhui-liu, line 458: Thanks for adding this new argument. I'm thinking whether we should make the argument required.

For example, operator runs monitor-data-recovery for store1 but does not specify a datetime, then current timestamp is used. Store1's data recovery completed before this monitor command runs. Even though store1's data recovery is done, as store1's current version timestamp is older than the start timestamp of monitor command, the monitor will unnecessarily keep waiting for a new version for store1 . This will make operator mistakenly think that store1's data recovery hasn't started yet.

If there is any benefit of making the argument optional that I'm not aware of, please don't hesitate to let me know.

lluwm: Shuhui, thank you for your comments! I agree that if we allow `datetime` to be optional, then we could have some misuse cases as you mentioned. Let me change it to required argument.

shuhui-liu (approved):

PR #443, opened 2023-05-24
[router] Fixed a NPE occurring in router's streaming batch get

FelixGV:
This commit fixes a regression introduced in PR #412, where a race condition would result in VeniceChunkedResponse's responseMetadata to be not initialized yet. The field is actually not necessary at all, and its content can be determined the one and only time it's used, rather than stored ahead of time. Also eliminated the responseMetadataWriteInitiated AtomicBoolean, since the responseCompression field being null already has an equivalent meaning. Therefore, VeniceChunkedResponse now carries two fewer properties.

### DIFF
--- a/services/venice-router/src/main/java/com/linkedin/venice/router/streaming/VeniceChunkedResponse.java
+++ b/services/venice-router/src/main/java/com/linkedin/venice/router/streaming/VeniceChunkedResponse.java
@@ -284,8 +284,8 @@ public class VeniceChunkedResponse {
       }
     }
 
-    boolean isMultiGet = this.requestType.equals(RequestType.MULTI_GET_STREAMING);
-    if (isMultiGet && !this.responseCompression.equals(compression)) {
+    boolean isMultiGetStreaming = this.requestType.equals(RequestType.MULTI_GET_STREAMING);
+    if (isMultiGetStreaming && !this.responseCompression.equals(compression)) {
       // Defensive code, not expected
       LOGGER.error(
           "Received inconsistent compression for the new write: {}, and previous compression: {}",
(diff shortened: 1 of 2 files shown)


services/venice-router/src/main/java/com/linkedin/venice/router/streaming/VeniceChunkedResponse.java
```diff
@@ -274,40 +274,41 @@ public CompletableFuture<Long> write(
...
-         * when all the responses returned by storage nodes are ready, and when the inconsistency happens,
-         * {@link VeniceResponseAggregator} will return an error response instead to indicate the issue.
-         */
-        return CompletableFuture.completedFuture(0l);
-      }
-    } else {
-      responseMetadata = RESPONSE_META_FOR_COMPUTE;
+    }
+
+    boolean isMultiGet = this.requestType.equals(RequestType.MULTI_GET_STREAMING);
```
gaojieliu, line 287: Can we rename it to be `isMultiGetStreaming`?

FelixGV: Sure, done.

services/venice-router/src/main/java/com/linkedin/venice/router/streaming/VeniceChunkedResponse.java
```diff
@@ -246,7 +246,7 @@ public boolean whetherToSkipMessage(Object msg, ChannelPromise promise) {
               "Unexpected response status: " + status + ", and only non 200 status is expected here");
         }
       }
-      throw new VeniceException("Unexpected message type received: " + msg.getClass());
+      throw new VeniceException("Unexpected message type received: " + (msg == null ? "null" : msg.getClass()));
```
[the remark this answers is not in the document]
FelixGV: It used to be possible prior to the fix, and it resulted in a NPE bubbling up from this line, rather than a `VeniceException`.

With the fix, I believe it's not possible for a null to come in, nor for this line to ever be reached at all. So this is just defensive code (as it was always intended to be, AFAICT), but just ensuring that it can't throw a NPE anymore if a future regression were to shove a null `msg` down here.

gaojieliu (approved):

PR #444, opened 2023-05-24
[test] Mitigate FC integration tests taking long time in CI

m-nagarajan:
## Summary
1. Moved `AvroStoreClientGzipEndToEndTest` and `AvroStoreClientZstdEndToEndTest` to `IntegrationTestsG` as it was taking too much time in `IntegrationTestsI`
2. Trimmed down the number of tests run for the above classes
3. Trimmed down the number of tests in `testFastClientWithLongTailRetry` which needs to be fixed further as its flaky, will handle it as a separate task, as these issues are blocking other PRs to be merged due to test failures.
4. Added timeouts for all tests

These tests will be be cleaned up further:
1. once we validate `SERVER_BASED_METADATA`, we will probably remove the cases with `DA_VINCI_CLIENT_BASED_METADATA` and `THIN_CLIENT_BASED_METADATA`
2. once `useStreamingBatchGetAsDefault` is validated, the code and tests for single get based batch get will be removed

## How was this PR tested?
GH CI

### DIFF
--- a/internal/venice-test-common/src/integrationtest/java/com/linkedin/venice/fastclient/AvroStoreClientEndToEndTest.java
+++ b/internal/venice-test-common/src/integrationtest/java/com/linkedin/venice/fastclient/AvroStoreClientEndToEndTest.java
@@ -425,7 +425,7 @@ public class AvroStoreClientEndToEndTest extends AbstractClientEndToEndSetup {
         storeMetadataFetchMode);
   }
 
-  @Test(dataProvider = "True-and-False", timeOut = TIME_OUT)
+  @Test(dataProvider = "FastClient-One-Boolean", timeOut = TIME_OUT)
   public void testFastClientWithLongTailRetry(boolean batchGet) throws Exception {
     ClientConfig.ClientConfigBuilder clientConfigBuilder =
         new ClientConfig.ClientConfigBuilder<>().setStoreName(storeName).setR2Client(r2Client);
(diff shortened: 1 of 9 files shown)


internal/venice-test-common/src/integrationtest/java/com/linkedin/venice/fastclient/AvroStoreClientEndToEndTest.java
```diff
@@ -416,53 +425,39 @@ public void testFastClientGetWithDifferentHTTPVariants(
...
+  public void testFastClientWithLongTailRetry(boolean batchGet) throws Exception {
     ClientConfig.ClientConfigBuilder clientConfigBuilder =
         new ClientConfig.ClientConfigBuilder<>().setStoreName(storeName).setR2Client(r2Client);
 
     if (batchGet) {
       clientConfigBuilder
           // default maxAllowedKeyCntInBatchGetReq is 2. configuring it to test different cases.
           .setMaxAllowedKeyCntInBatchGetReq(recordCnt)
-          .setUseStreamingBatchGetAsDefault(useStreamingBatchGetAsDefault);
+          .setUseStreamingBatchGetAsDefault(true);
```
sixpluszero, line 437: Is this expected to be true always?

sixpluszero, line 437: Just to confirm, we will only test the true case here?

sixpluszero (approved):
Thanks for putting up the fix.
I think it is good to merge as it pass everything for the purpose of unblocking, but if you think some permutations are actually needed and still used in current case, we should add it back, maybe we split it into different classes and assign them into different integration groups.

internal/venice-test-common/src/integrationtest/java/com/linkedin/venice/fastclient/AvroStoreClientEndToEndTest.java
m-nagarajan: These tests needs to be rewritten to remove flakiness, will add the case for both true/false then. For now, this should be okay.

m-nagarajan (commented):
Thanks @sixpluszero for the review. Will take a look at this again later this week and see what more can be cleaned up or added back. Merging this for now to unblock the test failures.

PR #445, opened 2023-05-25
[Server] Increase default worker threads for admin tasks to avoid starvation

asaxena76:
[Server] When processing per-store Admin messages, the default threadpool size of 1 would result in one bad admin message
to cause timeouts for other stores admin operations. In effect breaking the isolation of stores.
Added test to demonstrate this effect. Changed default thread pool size to 3 to prevent this situation

## How was this PR tested?

Added unit tests

### DIFF
--- a/services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/consumer/AdminConsumptionTask.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/consumer/AdminConsumptionTask.java
@@ -64,6 +64,10 @@ import org.apache.logging.log4j.Logger;
  * This class is used to create a task, which will consume the admin messages from the special admin topics.
  */
 public class AdminConsumptionTask implements Runnable, Closeable {
+  // Setting this to a value so that admin queue will not go out of memory in case of too many admin messages.
+  // If we hit this number , there is most likely something seriously wrong with the system.
+  private static final int MAX_WORKER_QUEUE_SIZE = 10000;
+
   private static class AdminErrorInfo {
     long offset;
     Exception exception;
@@ -280,7 +284,7 @@ public class AdminConsumptionTask implements Runnable, Closeable {
         maxWorkerThreadPoolSize,
         60,
         TimeUnit.SECONDS,
-        new LinkedBlockingQueue<>(),
+        new LinkedBlockingQueue<>(MAX_WORKER_QUEUE_SIZE),
         new DaemonThreadFactory("Venice-Admin-Execution-Task"));
     this.undelegatedRecords = new LinkedList<>();
     this.stats.setAdminConsumptionFailedOffset(failingOffset);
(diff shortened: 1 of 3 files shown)


services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/consumer/AdminConsumptionTask.java
```diff
@@ -274,8 +274,9 @@ public AdminConsumptionTask(
 
     this.storeAdminOperationsMapWithOffset = new ConcurrentHashMap<>();
     this.problematicStores = new ConcurrentHashMap<>();
+    // since we use an unbounded queue the core pool size is really the max pool size
     this.executorService = new ThreadPoolExecutor(
-        1,
+        maxWorkerThreadPoolSize,
         maxWorkerThreadPoolSize,
```
majisourav99, line 280: So this 2nd param will  not be triggered since we are passing a unbounded work queue or we should consider limiting the queue size as well ?

huangminchn: Right, I don't think setting core thread number matters though.

asaxena76: If we use an unbounded queue then setting corePoolSize to more than 1 is **required** if we want multi threaded . The scheduler will just keep queueing results otherwise . 

https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ThreadPoolExecutor.html
"Unbounded queues. Using an unbounded queue (for example a [LinkedBlockingQueue](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/LinkedBlockingQueue.html) without a predefined capacity) will cause new tasks to wait in the queue when all corePoolSize threads are busy. Thus, no more than corePoolSize threads will ever be created. (And the value of the maximumPoolSize therefore doesn't have any effect.)"

Event if we bound the queue to a small number , it will wait until the queue is full before it starts spawning new threads. What we want in this case is parallel execution for "isolation" . The simplest way to achieve is to create a fixed size threadpool for which both the core and max should be set to the same value. Queue growth is not a concern because the message arrival rate is tiny. In any case no new threads will be created.

From the same source "By setting corePoolSize and maximumPoolSize the same, you create a fixed-size thread pool."

You can confirm this behavior by modifying the count and running the test . Basically this means that the config setting had never worked in achieving multi threaded behavior.

majisourav99: Right the max pool size will not be even triggered here, but the core pool size will determine the number of threads its going to use as we use unbounded work queue.  Thats why I said we might want to limit the queue size or not.

asaxena76: What would you suggest the value should be? Lets say we  change the unbounded queue to queue with capacity 10. If we keep the core value to be 1 , the head message will still block all others if it is not able to finish in time. So for us to get any benefit we need to change the core pool size to the maxthread value. The only effect of limiting the queue size will be that the console will start rejecting messages when the queue gets full. Guess we can add that . what value do you suggest?

majisourav99: No, we need to increase the core pool size for sure (thru config). But worker queue size we put a large number, infinite seems too big. may be 10k ?

huangminchn (approved):
Thanks Amit!

PR #446, opened 2023-05-25
[vpj] Fail pushjobs running over 24 hr bootstrap limit

majisourav99:
## Fail pushjobs running over 24 hr bootstrap limit

Today pushjobs are failed if they run over `bootstrapToOnlineTimeoutInHours` time limit, but its checked in server in `checkLongRunningTaskState` which kicks in only if its actively ingesting. But if due to some reason it does not ingest the job is left hanging forever. This PR add a check in vpj to honor the bootstrap time limit config and fail the push proactively.

### DIFF
--- a/clients/venice-push-job/src/test/java/com/linkedin/venice/hadoop/VenicePushJobTest.java
+++ b/clients/venice-push-job/src/test/java/com/linkedin/venice/hadoop/VenicePushJobTest.java
@@ -154,7 +154,7 @@ public class VenicePushJobTest {
     ControllerClient client = mock(ControllerClient.class);
     JobStatusQueryResponse response = mock(JobStatusQueryResponse.class);
     doReturn("UNKNOWN").when(response).getStatus();
-    doReturn(response).when(client).queryOverallJobStatus(anyString(), eq(Optional.empty()));
+    doReturn(response).when(client).queryOverallJobStatus(anyString(), eq(Optional.empty()), eq(null));
     VenicePushJob pushJob = getSpyVenicePushJob(vpjProps, client);
     VenicePushJob.PushJobSetting pushJobSetting = pushJob.getPushJobSetting();
     pushJobSetting.jobStatusInUnknownStateTimeoutMs = 10;
--- a/clients/venice-push-job/src/test/java/com/linkedin/venice/hadoop/VenicePushJobTest.java
+++ b/clients/venice-push-job/src/test/java/com/linkedin/venice/hadoop/VenicePushJobTest.java
@@ -22,6 +22,7 @@ import static org.mockito.ArgumentMatchers.anyInt;
 import static org.mockito.ArgumentMatchers.anyLong;
 import static org.mockito.ArgumentMatchers.anyString;
 import static org.mockito.Mockito.doNothing;
+>>>>>>> eb14ad121 (added unit test)
 import static org.mockito.Mockito.doReturn;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.spy;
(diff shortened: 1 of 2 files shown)


nisargthakkar:
nit: Can you edit the component in the PR title to `[vpj]`? Rest LGTM. I'm not sure if we can test this, though. Might have to override and merge

clients/venice-push-job/src/test/java/com/linkedin/venice/hadoop/VenicePushJobTest.java
```diff
@@ -16,11 +16,7 @@
 import static com.linkedin.venice.hadoop.VenicePushJob.VENICE_DISCOVER_URL_PROP;
 import static com.linkedin.venice.hadoop.VenicePushJob.VENICE_STORE_NAME_PROP;
 import static com.linkedin.venice.status.BatchJobHeartbeatConfigs.HEARTBEAT_ENABLED_CONFIG;
-import static org.mockito.ArgumentMatchers.any;
-import static org.mockito.ArgumentMatchers.anyBoolean;
-import static org.mockito.ArgumentMatchers.anyInt;
-import static org.mockito.ArgumentMatchers.anyLong;
-import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.ArgumentMatchers.*;
```
adamxchen, line 19: nit: not use the wild card import, but not a big deal.

clients/venice-push-job/src/test/java/com/linkedin/venice/hadoop/VenicePushJobTest.java
```diff
@@ -150,6 +147,31 @@ public void testPushJobSettingWithLivenessHeartbeat() {
...
+    topicInfo.version = 1;
+    topicInfo.topic = "abc";
+    pushJob.storeSetting = new VenicePushJob.StoreSetting();
+    pushJob.storeSetting.storeResponse = new StoreResponse();
+    StoreInfo storeInfo = new StoreInfo();
+    storeInfo.setBootstrapToOnlineTimeoutInHours(1);
+    pushJob.storeSetting.storeResponse.setStore(storeInfo);
+    Assert.expectThrows(
+        VeniceException.class,
+        () -> pushJob.pollStatusUntilComplete(Optional.empty(), client, pushJobSetting, topicInfo));
```
adamxchen, line 172: Can we also assert the exception thrown contain the right msg? It seems like the exception is thrown due to being in UNKNOWN state instead of being over the time limit set by `bootstrapToOnlineTimeoutInHours`.

In terms of testing, maybe we can set `BootstrapToOnlineTimeoutInHours` to be 0 and compare the time difference in milliseconds instead of hour so it's easier to reproduce this in tests.

adamxchen: From this [javadoc](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/TimeUnit.html#convert-long-java.util.concurrent.TimeUnit-)
>  Conversions from finer to coarser granularities truncate, so lose precision. For example, converting 999 milliseconds to seconds results in 0

So, I think it would always be 0 unless it hits one hour mark, I haven't tested it though.

majisourav99: yeah it will be 0 if elapsed time is < 1hr.  we can either set the config to -1 or convert everything to millisec and compare to trigger the code.

sixpluszero: Yeah. Personally converting hours->ms might be better I guess...

nisargthakkar (approved):

PR #447, opened 2023-05-25
[server] Fix edge case in isolated ingestion handover

sixpluszero:
## [server] Fix edge case in isolated ingestion handover
This PR address an edge case in isolated ingestion.
The current ingestion handover of isolated ingestion goes as follows:
(1) [child] stop consumption (2) [child] report ingestion process (3)[main] receive report, start consumption locally
This works fine, however, considering the following incoming Helix request sequence that could happen in the same time.
(1) standby->offline (2) offline -> dropped (3) offline -> standby

If standby -> offline is accepted by child process when executing stop consumption, it will be accepted for now as child process think it has the resource. Then it will go into offline->drop, which is essentially dropping partition(potentially engine as well) and cleaning up metadata, while the handover step (2) is happening in between. In the edge case, it will lead to first dropping partition and then reopen partition in main process. When Helix is issuing (3), since metadata is cleaned up during the race condition, it will by default send start consumption request to forked process and try to open the same RocksDB in the forked process and thus throw exception.

Although with previous ingestion isolation shutdown safeguard, all problematic host can be mitigated by restart, but this is a race condition that need to be tackled.
To avoid clashing the Helix transition with intermediate states, the solution proposed here is: 
(1) When child process is trying to hand over resource, it will first unsubscribe the resource (clean up metadata), thus all Helix request being sent to the host will be rejected and retried, until handover is completed. If report is failed to delievered (after retries), that should be network issue and ingestion should fail anyway (not this PR's scope)
(2) When main process receives COMPLETED report, it will not adjust resource metadata right away, but instead it will wait till startConsumptionLocally is completed (either successfully or exceptionally) and then set topic partition to local. Thus Helix actions will then be able to be executed on the main process correctly.

Also, this PR should also fixes the flaky unit test and unify two unit tests into same file.

## How was this PR tested?
Adjust unit tests to reflect the changed logic.


clients/da-vinci-client/src/test/java/com/linkedin/davinci/ingestion/IsolatedIngestionBackendTest.java
```diff
@@ -191,18 +193,36 @@ public void testIsolatedIngestionNotifierAsyncCompletionHandling() {
...
-    when(backend.isTopicPartitionIngesting(topic, 0)).thenReturn(false);
-    when(backend.isTopicPartitionIngesting(topic, 1)).thenReturn(true);
+    when(backend.isTopicPartitionHosted(topic, 0)).thenReturn(false);
+    when(backend.isTopicPartitionHosted(topic, 1)).thenReturn(true);
+    when(backend.isTopicPartitionHosted(topic, 2)).thenReturn(true);
     backend.getIsolatedIngestionNotifier(ingestionNotifier).completed(topic, 0, 123L, "", Optional.empty());
-    verify(backend, times(0)).getCompletionHandlingExecutor();
+    TestUtils.waitForNonDeterministicAssertion(5, TimeUnit.SECONDS, true, () -> {
+      verify(backend, times(0)).getCompletionHandlingExecutor();
+    });
```
nisargthakkar, line 212: `times(0)` doesn't need `waitForNonDeterministicAssertion`

clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/IsolatedIngestionBackend.java
```diff
@@ -305,13 +312,28 @@ public void completed(
...
-        if (isTopicPartitionIngesting(kafkaTopic, partition)) {
+        if (isTopicPartitionHosted(kafkaTopic, partition)) {
           getCompletionHandlingExecutor().submit(() -> {
             VeniceStoreVersionConfig config = getConfigLoader().getStoreConfig(kafkaTopic);
             config.setRestoreDataPartitions(false);
             config.setRestoreMetadataPartition(false);
             // Start partition consumption locally.
-            startConsumption(config, partition, leaderState);
+            try {
+              startConsumptionLocally(config, partition, leaderState);
```
nisargthakkar, line 322: Can you add a comment/javadoc to `startConsumption` to denote that it is for consumption in the isolated process?

Actually, it seems my understanding is wrong and `startConsumption` can handle both cases. Can you please clarify this?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/IsolatedIngestionBackend.java
```diff
@@ -305,13 +312,28 @@ public void completed(
...
+            } catch (Exception e) {
+              LOGGER.warn(
+                  "Caught exception when resuming ingestion of topic: {}, partition: {} in main process",
+                  kafkaTopic,
+                  partition,
+                  e);
+            } finally {
+              getMainIngestionMonitorService().setVersionPartitionToLocalIngestion(kafkaTopic, partition);
```
nisargthakkar, line 330: (This is dependent on the answer to the prev comment) Shouldn't we be setting this config before invoking `startConsumptionLocally`?

clients/da-vinci-client/src/test/java/com/linkedin/davinci/ingestion/IsolatedIngestionBackendTest.java
sixpluszero: Yeah that's true, I was doing cmd+c/v :)

clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/IsolatedIngestionBackend.java
sixpluszero: The idea here is that:
Before we start consumption on the main process successfully (or exceptionally if store is deleted in between, this only happens in weird EI env), the resource should still be considered as "handover/in transistion" state. If we set the state config before this is completed, pending Helix transition on the same resource will get executed on local and may lead to unexpected result.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/IsolatedIngestionBackend.java
sixpluszero: Yeah, IsolatedIngestionBackend.startConsumption will start consumption based on resource location (if not exist will by default start in child process)
startConsumptionLocally is to enforce it to start locally as here is the handover logic from child -> main

sixpluszero: I can add more Java docs

nisargthakkar (approved):
LGTM

clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/IsolatedIngestionBackend.java
nisargthakkar: Thanks. That makes sense. I had a misunderstanding that child process and main process could end up ingesting at the same time. Confirmed offline that this is not possible.

PR #448, opened 2023-05-25
[da-vinci] Added DaVinci memory limiter support

gaojieliu:
When DaVinci is using `MEMORY_BACKED_BY_DISK` mode, the application could encounter OOM issue when configuring mlockall (pre-allocate/load all the data when invoking mmap) with a big data push (required memory exceeds RAM capacity) regardless of Ingestion Isolation.

To mitigate this risk triggered by unexpected large data pushes, this code change introduces two configs:
 /*
   * The memory up-limit for the ingestion path while using RocksDB Plaintable format.
   * Currently, this option is only meaningful for DaVinci use cases. */ public static final String INGESTION_MEMORY_LIMIT = "ingestion.memory.limit";

  /**
   * Whether the ingestion is using mlock or not.
   * Currently, this option is only meaningful for DaVinci use cases. *
   * Actually, this config option is being actively used, and it is a placeholder for the future optimization.
   * The memory limit logic implemented today is assuming mlock usage, and to make it backward compatible when
   * we want to do more optimization for non-mlock usage, we will ask the mlock user to enable this flag. */ public static final String INGESTION_MLOCK_ENABLED = "ingestion.mlock.enabled";

Internally it is mainly using SSTFileManager to control max size of all SST files: https://github.com/facebook/rocksdb/wiki/Managing-Disk-Space-Utilization There is a shortcoming with this feature since it doesn't check sst file size when opening up an existing database, so this code change backfills this feature gap.

Actions when hitting the memory limit:
1. For the new pushes (future versions): DaVinci ingestion will fail the new push and report error to the push job status system store, and the push job should fail with such error.

2. For the current versions. DaVinci will try to kill the ingestions belonging to non-current versions, and once there are spaces reclaimed, the ingestion of current versions will resume.

Current version ingestion resumption is still a best effort, and there could be unknown corner cases, which current logic can't handle, and the following metric will notify the DaVinci users to take more actions if it lasts for a long time, such as restarting the instance or reduce the total number of DaVinci stores being used.

This code change exposes a metric to monitor the ingestion status:
--stuck_ingestion.Gauge : store-level and aggregate

In this code change, I also discovered some inproper implementations in RocksDB SSTFileManager regarding enforcing size limit:
1. SSTFileManager doesn't enforce the size limit when opening up an existing database.
2. SSTFileManager still tracks the sst files event after closing the RocksDB and RocksDB destroy function will completely remove the tracking (maybe this is expected?)
3. An existing RocksDB can't resume even there are more space left after removing some other database. To workaround this issue, this code change will abandon the in-memory memtable by reopening the database and re-consume from the previous checkpoint.

## How was this PR tested?

unit test/integration test

Please refer to the above section to find new configs.

This PR depends on https://github.com/linkedin/venice/pull/437

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
@@ -283,7 +283,7 @@ public class DaVinciBackend implements Closeable {
         }
       } catch (VeniceNoStoreException e) {
         // The store does not exist in Venice anymore, so it will be deleted.
-        LOGGER.info("Store does not exist, will delete invalid local version: {}", storageEngineName);
+        LOGGER.warn("Store does not exist, will delete invalid local version: {}", storageEngineName);
         storeShouldBeDeleted = true;
       }
       if (storeShouldBeDeleted) {
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
@@ -71,6 +71,7 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
+import java.util.function.Function;
 import org.apache.helix.zookeeper.impl.client.ZkClient;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
@@ -104,7 +105,7 @@ public class DaVinciBackend implements Closeable {
       Optional<Set<String>> managedClients,
       ICProvider icProvider,
       Optional<ObjectCacheConfig> cacheConfig) {
-    LOGGER.info("Creating Da Vinci backend");
+    LOGGER.info("Creating Da Vinci backend with managed clients: {}", managedClients);
     try {
       VeniceServerConfig backendConfig = configLoader.getVeniceServerConfig();
       this.configLoader = configLoader;
@@ -155,13 +156,19 @@ public class DaVinciBackend implements Closeable {
       // Add extra safeguards here to ensure we have released RocksDB database locks before we initialize storage
       // services.
       IsolatedIngestionUtils.destroyLingeringIsolatedIngestionProcess(configLoader);
+      /**
+       * The constructor of {@link #storageService} will take care of unused store/store version cleanup.
+       */
       storageService = new StorageService(
           configLoader,
           aggVersionedStorageEngineStats,
           rocksDBMemoryStats,
           storeVersionStateSerializer,
           partitionStateSerializer,
-          storeRepository);
+          storeRepository,
+          true,
+          true,
+          functionToCheckWhetherStorageEngineShouldBeKeptOrNot(managedClients));
       storageService.start();
 
       VeniceWriterFactory writerFactory = new VeniceWriterFactory(backendProps.toProperties());
@@ -237,7 +244,7 @@ public class DaVinciBackend implements Closeable {
             "Ingestion isolated and Cache are incompatible configs!!  Aborting start up!");
       }
 
-      bootstrap(managedClients);
+      bootstrap();
 
       storeRepository.registerStoreDataChangedListener(storeChangeListener);
       cacheBackend.ifPresent(
@@ -251,14 +258,69 @@ public class DaVinciBackend implements Closeable {
     }
   }
 
-  protected synchronized void bootstrap(Optional<Set<String>> managedClients) {
+  private Function<String, Boolean> functionToCheckWhetherStorageEngineShouldBeKeptOrNot(
+      Optional<Set<String>> managedClients) {
+    return storageEngineName -> {
+      String storeName = Version.parseStoreFromKafkaTopicName(storageEngineName);
+      if (VeniceSystemStoreType.META_STORE.isSystemStore(storeName)) {
+        // Do not bootstrap meta system store via DaVinci backend initialization since the operation is not supported by
+        // ThinClientMetaStoreBasedRepository. This shouldn't happen normally, but it's possible if the user was using
+        // DVC based metadata for the same store and switched to thin client based metadata.
+        return true;
+      }
+      /**
+       * If the corresponding Venice store doesn't even exist, the local storage engine will be removed.
+       * If the managed client feature is enabled, but the storage engine is not on the list, the local
+       * storage engine will be removed.
+       */
+      boolean storeShouldBeDeleted = false;
+      try {
+        StoreBackend storeBackend = getStoreOrThrow(storeName); // throws VeniceNoStoreException
+        if (managedClients.isPresent() && !managedClients.get().contains(storeName) && storeBackend.isManaged()) {
+          // If the store is not-managed, all its versions will be removed.
+          LOGGER.info("Will delete unused managed version: {}", storageEngineName);
+          storeShouldBeDeleted = true;
+        }
+      } catch (VeniceNoStoreException e) {
+        // The store does not exist in Venice anymore, so it will be deleted.
+        LOGGER.info("Store does not exist, will delete invalid local version: {}", storageEngineName);
+        storeShouldBeDeleted = true;
+      }
+      if (storeShouldBeDeleted) {
+        /**
+         * Clean up the local state.
+         * Since it is possible there could multiple versions for the same store, the following cleanup
+         * logic should work for multiple times of cleanup for the same store.
+         */
+        deleteStore(storeName);
+        String baseDataPath = configLoader.getVeniceServerConfig().getDataBasePath();
+        new StoreBackendConfig(baseDataPath, storeName).delete();
+        return false;
+      }
+
+      // Check whether the local storage engine belongs to any valid store version or not
+      Set<Integer> validVersionNumbers = new HashSet<>();
+      Optional<Version> latestVersion = getVeniceLatestNonFaultyVersion(storeName, Collections.emptySet());
+      Optional<Version> currentVersion = getVeniceCurrentVersion(storeName);
+      currentVersion.ifPresent(version -> validVersionNumbers.add(version.getNumber()));
+      latestVersion.ifPresent(version -> validVersionNumbers.add(version.getNumber()));
+
+      int versionNumber = Version.parseVersionFromKafkaTopicName(storageEngineName);
+      // The version is no longer valid (stale version), it will be deleted.
+      if (!validVersionNumbers.contains(versionNumber)) {
+        LOGGER.info("Will delete obsolete local version: {}", storageEngineName);
+        return false;
+      }
+      return true;
+    };
+  }
+
+  private synchronized void bootstrap() {
     List<AbstractStorageEngine> storageEngines =
         storageService.getStorageEngineRepository().getAllLocalStorageEngines();
-    LOGGER.info("Starting bootstrap, storageEngines: {}, managedClients: {}", storageEngines, managedClients);
-    Map<String, Set<Integer>> expectedBootstrapVersions = new HashMap<>();
+    LOGGER.info("Starting bootstrap, storageEngines: {}", storageEngines);
     Map<String, Version> storeNameToBootstrapVersionMap = new HashMap<>();
     Map<String, List<Integer>> storeNameToPartitionListMap = new HashMap<>();
-    Set<String> unusedStores = new HashSet<>();
     for (AbstractStorageEngine storageEngine: storageEngines) {
       String kafkaTopicName = storageEngine.getStoreName();
       String storeName = Version.parseStoreFromKafkaTopicName(kafkaTopicName);
@@ -270,39 +332,12 @@ public class DaVinciBackend implements Closeable {
       }
 
       try {
-        StoreBackend storeBackend = getStoreOrThrow(storeName); // throws VeniceNoStoreException
-        if (managedClients.isPresent() && !managedClients.get().contains(storeName) && storeBackend.isManaged()) {
-          // If the store is not-managed, all its versions will be removed.
-          LOGGER.info("Deleting unused managed version: {}", kafkaTopicName);
-          unusedStores.add(storeName);
-          storageService.removeStorageEngine(kafkaTopicName);
-          continue;
-        }
+        getStoreOrThrow(storeName); // throws VeniceNoStoreException
       } catch (VeniceNoStoreException e) {
-        // The store does not exist in Venice anymore, so it will be deleted.
-        LOGGER.info("Store does not exist, deleting invalid local version: {}", kafkaTopicName);
-        unusedStores.add(storeName);
-        storageService.removeStorageEngine(kafkaTopicName);
-        continue;
-      }
-
-      // Initialize expected version numbers for each store.
-      if (!expectedBootstrapVersions.containsKey(storeName)) {
-        Set<Integer> validVersionNumbers = new HashSet<>();
-        Optional<Version> latestVersion = getVeniceLatestNonFaultyVersion(storeName, Collections.emptySet());
-        Optional<Version> currentVersion = getVeniceCurrentVersion(storeName);
-        currentVersion.ifPresent(version -> validVersionNumbers.add(version.getNumber()));
-        latestVersion.ifPresent(version -> validVersionNumbers.add(version.getNumber()));
-        expectedBootstrapVersions.put(storeName, validVersionNumbers);
+        throw new VeniceException("Unexpected to encounter non-existing store here: " + storeName);
       }
 
       int versionNumber = Version.parseVersionFromKafkaTopicName(kafkaTopicName);
-      // The version is no longer valid (stale version), it will be deleted.
-      if (!expectedBootstrapVersions.get(storeName).contains(versionNumber)) {
-        LOGGER.info("Deleting obsolete local version: {}", kafkaTopicName);
-        storageService.removeStorageEngine(kafkaTopicName);
-        continue;
-      }
 
       Version version = storeRepository.getStoreOrThrow(storeName)
           .getVersion(versionNumber)
@@ -321,21 +356,6 @@ public class DaVinciBackend implements Closeable {
           && (storeNameToBootstrapVersionMap.get(storeName).getNumber() < versionNumber))) {
         storeNameToBootstrapVersionMap.put(storeName, version);
         storeNameToPartitionListMap.put(storeName, storageService.getUserPartitions(kafkaTopicName));
-      }
-    }
-    /**
-     * Remove unused store's {@link StoreBackend} to avoid duplicate metrics registration issues when iterating storage
-     * engines for different versions of the same store.
-     */
-    for (String unusedStoreName: unusedStores) {
-      deleteStore(unusedStoreName);
-    }
-
-    // Cleanup stale StaleBackendConfig
-    String baseDataPath = configLoader.getVeniceServerConfig().getDataBasePath();
-    for (String storeName: StoreBackendConfig.listConfigs(baseDataPath)) {
-      if (!storeByNameMap.containsKey(storeName)) {
-        new StoreBackendConfig(baseDataPath, storeName).delete();
       }
     }
 
@@ -594,9 +614,13 @@ public class DaVinciBackend implements Closeable {
       ingestionReportExecutor.submit(() -> {
         VersionBackend versionBackend = versionByTopicMap.get(kafkaTopic);
         if (versionBackend != null) {
+          /**
+           * Report push status needs to be executed before deleting the {@link VersionBackend}.
+           */
+          reportPushStatus(kafkaTopic, partitionId, ExecutionStatus.ERROR);
+
           versionBackend.completePartitionExceptionally(partitionId, e);
           versionBackend.tryStopHeartbeat();
-          reportPushStatus(kafkaTopic, partitionId, ExecutionStatus.ERROR);
         }
       });
     }
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/AvroGenericDaVinciClient.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/AvroGenericDaVinciClient.java
@@ -1,5 +1,6 @@
 package com.linkedin.davinci.client;
 
+import static com.linkedin.davinci.ingestion.utils.IsolatedIngestionUtils.INGESTION_ISOLATION_CONFIG_PREFIX;
 import static com.linkedin.davinci.store.rocksdb.RocksDBServerConfig.ROCKSDB_LEVEL0_FILE_NUM_COMPACTION_TRIGGER;
 import static com.linkedin.davinci.store.rocksdb.RocksDBServerConfig.ROCKSDB_LEVEL0_FILE_NUM_COMPACTION_TRIGGER_WRITE_ONLY_VERSION;
 import static com.linkedin.davinci.store.rocksdb.RocksDBServerConfig.ROCKSDB_LEVEL0_SLOWDOWN_WRITES_TRIGGER;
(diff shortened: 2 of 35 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -1071,19 +1075,63 @@ private void processIngestionException() {
...
+           * will end. Even later on, there are avaiable memory space, we can't resume the ingestion task.
+           */
+          pauseConsumption(pubSubTopicPartition.getPubSubTopic().getName(), pubSubTopicPartition.getPartitionNumber());
+          LOGGER.info(
+              "Pausing consumption of topic: {}, partition: {} because of hitting memory limit",
+              pubSubTopicPartition.getPubSubTopic().getName(),
+              pubSubTopicPartition.getPartitionNumber());
+          runnableForKillIngestionTasksForNonCurrentVersions.run();
+          if (storageEngine.hasMemorySpaceLeft()) {
+            if (partitionConsumptionStateMap.containsKey(exceptionPartition)) {
```
sixpluszero, line 1100: Looks like the contains key check is unnecessary? It must be in PCS map unless there are some edge case?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -1071,19 +1075,63 @@ private void processIngestionException() {
           */
         partitionIngestionExceptionList.set(exceptionPartition, null);
       } else {
-        if (!partitionConsumptionState.isCompletionReported()) {
-          reportError(partitionException.getMessage(), exceptionPartition, partitionException);
+        PubSubTopicPartition pubSubTopicPartition = new PubSubTopicPartitionImpl(versionTopic, exceptionPartition);
+        /**
+         * Special handling for current version when encountering {@link MemoryLimitExhaustedException}.
+         */
+        if (partitionException instanceof MemoryLimitExhaustedException
```
sixpluszero, line 1082: Do we want to check if it is DVC as well?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -3480,6 +3537,20 @@ protected String getKafkaVersionTopic() {
     return kafkaVersionTopic;
   }
 
+  public boolean isStuck() {
```
sixpluszero, line 3540: nit: isStuck is a bit too general, here we only check memory limit exceed, maybe we extend the name to be a bit more specific?

sixpluszero: It seems like you are registering isStuck metric..hmm, okay maybe we leave a comment to note that for now we only track memory issue for DVC at least?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBStorageEngine.java
```diff
@@ -253,4 +254,17 @@ private String getRocksDbEngineConfigPath() {
...
   }
+
+  @Override
+  public boolean hasMemorySpaceLeft() {
+    SstFileManager sstFileManager = factory.getSstFileManager();
+    if (sstFileManager.isMaxAllowedSpaceReached() || sstFileManager.isMaxAllowedSpaceReachedIncludingCompactions()) {
+      return false;
+    }
+    long currentUsage = sstFileManager.getTotalSize();
+    if (factory.getMemoryLimit() - currentUsage >= 2 * factory.getMemtableSize()) {
```
sixpluszero, line 265: Why 2x memtable here?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBStorageEngineFactory.java
```diff
@@ -275,6 +295,34 @@ public synchronized void removeStorageEngine(AbstractStorageEngine engine) {
...
   }
 
+  /**
+   * The following function shouldn't be invoked at runtime (only at startup time), so we didn't apply
+   * any throttling here, and if the above condition changes in the future, we may need to consider
+   * throttling the deletion to reduce the IO impact to the database disk, which may result in some
+   * side effect in the read path.
+   */
+  @Override
+  public synchronized void removeStorageEngine(String storeName) {
```
sixpluszero, line 305: As far as I can see, it is only invoked in restoreAllStores, Just wondering why we could not reuse the existing API removeStorageEngine(AbstractStoreageEngine engine)?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBStoragePartition.java
```diff
@@ -249,6 +270,46 @@ public RocksDBStoragePartition(
         Collections.singletonList(RocksDB.DEFAULT_COLUMN_FAMILY));
   }
 
+  private void checkMemoryLimit(long memoryLimit, SstFileManager sstFileManager, String dbPath) {
```
sixpluszero, line 273: So the intention of this PR is to check SST total file size during RocksDB#open(), it could not dynamically track it during hybrid ingestion when new SST files are being materialized continuously into the opened RocksDB? Would it be a concern>?

sixpluszero (commented):
Thanks Gaojie for the implementation! This is a large PR and I did one pass and left some comments, please help correct my understanding in comments if they are not correct.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
```diff
@@ -594,9 +614,13 @@ public void error(String kafkaTopic, int partitionId, String message, Exception
       ingestionReportExecutor.submit(() -> {
         VersionBackend versionBackend = versionByTopicMap.get(kafkaTopic);
         if (versionBackend != null) {
+          /**
+           * Report push status needs to be executed before deleting the {@link VersionBackend}.
+           */
+          reportPushStatus(kafkaTopic, partitionId, ExecutionStatus.ERROR);
+
           versionBackend.completePartitionExceptionally(partitionId, e);
           versionBackend.tryStopHeartbeat();
```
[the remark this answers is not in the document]
gaojieliu: Personally, I don't have a preference about heartbeat order.

For error status reporting, it has to go before deleting `VeniceBackend` since `reportPushStatus` will check whether the corresponding `VersionBackend` exists or not:
https://github.com/linkedin/venice/blob/main/clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java#L505

clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
```diff
@@ -251,14 +258,69 @@ public DaVinciBackend(
...
-  protected synchronized void bootstrap(Optional<Set<String>> managedClients) {
+  private Function<String, Boolean> functionToCheckWhetherStorageEngineShouldBeKeptOrNot(
+      Optional<Set<String>> managedClients) {
+    return storageEngineName -> {
+      String storeName = Version.parseStoreFromKafkaTopicName(storageEngineName);
+      if (VeniceSystemStoreType.META_STORE.isSystemStore(storeName)) {
+        // Do not bootstrap meta system store via DaVinci backend initialization since the operation is not supported by
+        // ThinClientMetaStoreBasedRepository. This shouldn't happen normally, but it's possible if the user was using
+        // DVC based metadata for the same store and switched to thin client based metadata.
+        return true;
```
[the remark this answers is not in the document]
gaojieliu: Maybe, I am not sure.
I copied the logic from the previous `bootstrap` function.
@xunyin8 
Can we delete meta system store automatically here?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
```diff
@@ -270,39 +332,12 @@ protected synchronized void bootstrap(Optional<Set<String>> managedClients) {
...
       try {
-        StoreBackend storeBackend = getStoreOrThrow(storeName); // throws VeniceNoStoreException
-        if (managedClients.isPresent() && !managedClients.get().contains(storeName) && storeBackend.isManaged()) {
-          // If the store is not-managed, all its versions will be removed.
-          LOGGER.info("Deleting unused managed version: {}", kafkaTopicName);
-          unusedStores.add(storeName);
-          storageService.removeStorageEngine(kafkaTopicName);
-          continue;
-        }
+        getStoreOrThrow(storeName); // throws VeniceNoStoreException
```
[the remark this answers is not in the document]
gaojieliu: This is mainly for defensive coding.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/IsolatedIngestionBackend.java
```diff
@@ -307,11 +307,22 @@ public void completed(
...
+            /**
+             * Start partition consumption locally.
+             * If any error happens when starting the consumption, error will be reported.
+             */
+            try {
+              VeniceStoreVersionConfig config = getConfigLoader().getStoreConfig(kafkaTopic);
+              config.setRestoreDataPartitions(false);
+              config.setRestoreMetadataPartition(false);
+              startConsumption(config, partition, leaderState);
+            } catch (Exception e) {
```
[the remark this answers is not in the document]
gaojieliu: This logic is a must!
During the handover from II process to main process, `startConsumption` could fail since the database reopen could breach the memory limit.

internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
```diff
@@ -1912,4 +1920,20 @@ private ConfigKeys() {
    */
   public static final String CLIENT_PRODUCER_SCHEMA_REFRESH_INTERVAL_SECONDS =
       "client.producer.schema.refresh.interval.seconds";
+
+  /*
+   * The memory up-limit for the ingestion path while using RocksDB Plaintable format.
+   * Currently, this option is only meaningful for DaVinci use cases.
+   */
+  public static final String INGESTION_MEMORY_LIMIT = "ingestion.memory.limit";
```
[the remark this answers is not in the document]
gaojieliu: This one doesn't have a unit which means the customer can use any unit, such as:
```
128
2KB
200MB
2GB
```
They are all supported.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/AvroGenericDaVinciClient.java
```diff
@@ -618,6 +620,8 @@ private VeniceConfigLoader buildVeniceConfig() {
         .put(KAFKA_BOOTSTRAP_SERVERS, kafkaBootstrapServers)
         .put(ROCKSDB_PLAIN_TABLE_FORMAT_ENABLED, daVinciConfig.getStorageClass() == StorageClass.MEMORY_BACKED_BY_DISK)
         .put(INGESTION_USE_DA_VINCI_CLIENT, true)
+        .put(INGESTION_ISOLATION_CONFIG_PREFIX + "." + INGESTION_MEMORY_LIMIT, -1) // Explicitly disable memory limiter
```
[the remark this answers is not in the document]
gaojieliu: The reason I want to put it here to highlight that this is DaVinci specific to avoid confusion.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -1071,19 +1075,63 @@ private void processIngestionException() {
...
-        if (!partitionConsumptionState.isCompletionReported()) {
-          reportError(partitionException.getMessage(), exceptionPartition, partitionException);
+        PubSubTopicPartition pubSubTopicPartition = new PubSubTopicPartitionImpl(versionTopic, exceptionPartition);
+        /**
+         * Special handling for current version when encountering {@link MemoryLimitExhaustedException}.
+         */
+        if (partitionException instanceof MemoryLimitExhaustedException
+            || partitionException.getCause() instanceof MemoryLimitExhaustedException
+                && isCurrentVersion.getAsBoolean()) {
+          LOGGER.info(
```
adamxchen, line 1085: nit: this message should be a WARN since we are trying to recover from a relatively bad situation.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/config/VeniceServerConfig.java
```diff
@@ -609,6 +622,87 @@ public VeniceServerConfig(VeniceProperties serverProperties, Map<String, Map<Str
...
+          "Config: " + INGESTION_MEMORY_LIMIT + " is only meaningful when using RocksDB plaintable format");
+    }
+    long totalMemtableUsage = rocksDBServerConfig.getRocksDBTotalMemtableUsageCapInBytes();
+    // Check ingestion mode
+    if (configuredIngestionMode.equals(IngestionMode.ISOLATED)) {
+      /**
+       * When ingestion isolation is enabled, we need to subtract the usages from the following componnets:
+       * 1. Main process total memtable usage limit.
+       * 2. Heap size of isolated JVM process.
+       * 3. Total memtable usage limit in isolated process.
```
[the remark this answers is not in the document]
gaojieliu: memtable is not part of heap memory and it is off-heap, and currently in Venice, we are using write buffer manager to limit the max amount of memory used by memtables.
The default write buffer manager limit is 2GB.. 
For DaVinci usage, we should reduce it.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
gaojieliu: Not necessary since there are logic upper layer in `VeniceServerConfig` to only enable this feature for DaVinci use case.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
gaojieliu: yeah, I can remove it.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBStorageEngine.java
gaojieliu: 2x just some magic number and we would like to have some headroom before resuming the consumption.
The database stuck can be caused by a flush or a write, and we would like to resume the consumption at least for one flush, and 2 was the number I chose.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
gaojieliu: Will do.

adamxchen: I agree that we should make the name more specific. Also a question for my curiosity, I think we have host-level ingestion metric so if the ingestion is stuck, customers can use that and log message to infer the situation so a new metric is not needed. Can we do that?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBStorageEngine.java
adamxchen: I think we try to count both current version and future version, so that's the reason?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBStorageEngineFactory.java
gaojieliu: Good question.
Today, sst file manager doesn't check space limit when opening a new database, which means DaVinci should be able to open the database even its size is bigger than memory limit.
But this is not a good long-term solution since database opening should check the space limit as well, and I am planning to work with RocksDB team to fix this issue.
Once that fix is in, opening up a storage engine and then deleting it won't work.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBStoragePartition.java
gaojieliu: Today, RocksDB SSTFileManager impl doesn't check space limit when opening up an existing database, and there is the reason we have this logic here.

when the database is opened, SSTFileManager will track every write and it is not a concern.

internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
```diff
@@ -1918,4 +1926,20 @@ private ConfigKeys() {
...
+
+  /**
+   * Whether the ingestion is using mlock or not.
```
adamxchen, line 1937: I am confused by the name. Is the config indicating whether the customer's host is using `mlock` or whether Venice ingestion process would be using `mlock`?

internal/venice-test-common/src/integrationtest/java/com/linkedin/venice/endToEnd/DaVinciClientMemoryLimitTest.java
```diff
@@ -0,0 +1,485 @@
...
+  private D2Client d2Client;
+
+  @BeforeClass
+  public void setUp() {
+    Utils.thisIsLocalhost();
+    Properties clusterConfig = new Properties();
+    clusterConfig.put(SERVER_PROMOTION_TO_LEADER_REPLICA_DELAY_SECONDS, 10L);
+    clusterConfig.put(OFFLINE_PUSH_MONITOR_DAVINCI_PUSH_STATUS_SCAN_NO_DAVINCI_STATUS_REPORT_RETRY_MAX_ATTEMPTS, 5); // To
+                                                                                                                     // allow
+                                                                                                                     // more
```
adamxchen, line 82: this is a long interesting comment 😄

adamxchen (commented):
Thanks for the work!

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
gaojieliu: No, this memory constraint stuck needs to be tracked separately since it is indicating some serious issue in DVC.
I can potentially rename it to be `isStuckByMemoryConstraint` to make it more specific.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
```diff
@@ -251,14 +258,69 @@ public DaVinciBackend(
...
+      try {
+        StoreBackend storeBackend = getStoreOrThrow(storeName); // throws VeniceNoStoreException
+        if (managedClients.isPresent() && !managedClients.get().contains(storeName) && storeBackend.isManaged()) {
+          // If the store is not-managed, all its versions will be removed.
+          LOGGER.info("Will delete unused managed version: {}", storageEngineName);
+          storeShouldBeDeleted = true;
+        }
+      } catch (VeniceNoStoreException e) {
+        // The store does not exist in Venice anymore, so it will be deleted.
+        LOGGER.info("Store does not exist, will delete invalid local version: {}", storageEngineName);
```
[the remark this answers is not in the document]
gaojieliu: Sure. WARN log should be fine.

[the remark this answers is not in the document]
gaojieliu: Sure

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
gaojieliu: Sure

internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
gaojieliu: To customer, they are essentially same thing.

adamxchen: Let me clarify this. Our intention is to ask customers to enable it if mlockall is being used so Venice can take this into memory limiter consideration. 
One misleading interpretation of this config is regardless customers has this `mlockall` or not, Venice ingestion process will enable it for them. I could be missing something but I don't think we are doing this in this PR

adamxchen: That being said, I've been thinking if there's better naming and feel that `ingestion.mlock.enabled` isn't bad, maybe `application.mlock.enabled`?  Please feel free to drop this comment. Not a big deal anyway.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/IsolatedIngestionBackend.java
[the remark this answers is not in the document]
sixpluszero: I see...oh right got it. I mean before that, I never see this part failed, but this new logic do have the chance to fail.

internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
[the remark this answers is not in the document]
sixpluszero: Got it. One more nit - maybe we limit the config to DAVINCI_INGESTION_MEMORY_LIMIT? If we don't foresee server having this kind of limit. (either way is fine, not a big deal)

internal/venice-test-common/src/integrationtest/java/com/linkedin/venice/endToEnd/DaVinciClientMemoryLimitTest.java
sixpluszero: Is there anyway to configure the plugin to make the long comment showing in one line above? I do see a lot of long comment in our codebase getting formatted like this, which looks pretty bad..

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBStoragePartition.java
sixpluszero: Oh I see... I think I missed this line:

>         this.sstFileManager.setMaxAllowedSpaceUsage(this.memoryLimit);
Now I get how it is tracking the usage dynamically.

internal/venice-test-common/src/integrationtest/java/com/linkedin/venice/endToEnd/DaVinciClientMemoryLimitTest.java
gaojieliu: yeah, the best way would be to move the tailing comments to be above the java statement.

gaojieliu:
@adamxchen 
Regarding this comment:
```
I agree that we should make the name more specific. Also a question for my curiosity, I think we have host-level ingestion metric so if the ingestion is stuck, customers can use that and log message to infer the situation so a new metric is not needed. Can we do that?
```
We need an explicit signal to tell customer about hitting the memory limit since it is not recoverable and other ingestion stuck, which is caused by network/kafka instability normally can be recovered, and memory constraints need explicit human intervention.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/config/VeniceServerConfig.java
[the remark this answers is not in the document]
sixpluszero: Sure, we can tune it in seas-careers along with your memory limiter config once it is in and released.
Is my understanding correct that that total memtable memory would be memtable per store * store number, which would be either (32M/64M) * 13, and this number is the peak time (all hybrid store ingestion at the same time). So I guess 0.5G at most should be big enough? Or maybe even less.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBStorageEngine.java
sixpluszero: Thanks for the explanation. Could you maybe put a simple comment there to explain the number of 2? Otherwise it could be fairly confusing for other ppl.

gaojieliu: Sure

clients/da-vinci-client/src/main/java/com/linkedin/davinci/config/VeniceServerConfig.java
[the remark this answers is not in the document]
gaojieliu: No, it is per partition.
That is the reason we want to limit the max usage.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBStorageEngineFactory.java
sixpluszero: Got it.

sixpluszero (approved):
Thanks Gaojie for the fix!

PR #449, opened 2023-05-25
[admin-tool][data-recovery] Executor Idempotency

lyssarzr:
## Summary
With this update, we fix an issue with potential duplicate/additional push jobs in Azkaban when re-running the data recovery store repush command from the admin-tool. To accomplish this, there are two additional input parameters needed to run the store repush command: the `--url` of a valid Venice parent controller, and a `--datetime` parameter, to be input in `dd-MM-yyyy HH:mm:ss` format. For example:

```
java -jar venice-admin-tool-all.jar --execute-data-recovery --recovery-command venice-tools --url "http://parentcontrollerurl:1576" --stores storeName --source-fabric ei-ltx1 --datetime "05-25-2023 07:57:32" --extra-command-args 'repush kafka --force --format'
```

If the input `--datetime` parameter is earlier than the input store's latest push job start date and time, then it will be skipped - we assume that we have already created a repush job for this store. Else, we will trigger the repush as normal.

Resolves VENG-10365

## How was this PR tested?
We initiated a repush on the heartbeat store in EI, and then attempted to create a second push job, and was appropriately unable to do so.

See Summary section for usage notes


clients/venice-admin-tool/src/main/java/com/linkedin/venice/datarecovery/DataRecoveryClient.java
```diff
@@ -45,8 +52,35 @@ public DataRecoveryMonitor getMonitor() {
...
+  public void filterStoresWithOngoingRepush(Set<String> storesList, StoreRepushCommand.Params params) {
+    String url = params.getUrl();
+    ControllerClient cli = params.getPCtrlCliWithoutCluster();
+    LocalDateTime timestamp = params.getTimestamp();
+    String sourceFabric = params.getSourceFabric();
+    for (String s: storesList) {
+      String clusterName = cli.discoverCluster(s).getCluster();
+      try (ControllerClient parentCtrlCli = new ControllerClient(clusterName, url, Optional.empty())) {
+        StoreHealthAuditResponse storeHealthInfo = parentCtrlCli.listStorePushInfo(s, false);
+        RegionPushDetails regionPushDetails = storeHealthInfo.getRegionPushDetails().get(sourceFabric);
```
shuhui-liu, line 64: I don't think we care about source fabric. Instead, dest fabric is what we should check for.

clients/venice-admin-tool/src/main/java/com/linkedin/venice/datarecovery/DataRecoveryClient.java
```diff
@@ -45,8 +52,35 @@ public DataRecoveryMonitor getMonitor() {
...
+        String latestTimestamp = regionPushDetails.getPushStartTimestamp();
```
shuhui-liu, line 65: `getPushStartTimestamp` only returns the start timestamp of current version. If the store has ongoing push job (future version), we should also skip the repush job as Venice does not support concurrent push jobs, right?

shuhui-liu (commented):
Thanks for the implementation. I left some comments. Also, please fix the VeniceCI/UnitTestsAndCodeCoverage failure.

clients/venice-admin-tool/src/main/java/com/linkedin/venice/datarecovery/DataRecoveryClient.java
```diff
@@ -45,8 +52,35 @@ public DataRecoveryMonitor getMonitor() {
...
+        DateTimeFormatter formatter = DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss.SSSSSS");
```
lluwm, line 66: Can we use the predefined constant [DateTimeFormatter.ISO_LOCAL_DATE_TIME](https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html#ISO_LOCAL_DATE_TIME) in this case?

clients/venice-admin-tool/src/main/java/com/linkedin/venice/datarecovery/DataRecoveryClient.java
```diff
@@ -45,8 +52,35 @@ public DataRecoveryMonitor getMonitor() {
...
+        LocalDateTime latestPushStartTime = LocalDateTime.parse(latestTimestamp, formatter);
+        if (latestPushStartTime.compareTo(timestamp) > 0) {
+          LOGGER.warn(
+              "Removing store " + s
+                  + " from stores list, as the has a push job started more recently than the input timestamp.");
+          storesList.remove(s);
+        }
```
lluwm, line 73: Two thoughts:
1.  it doesn't sound like a warning message.
2. we should looking for a better way to report, e.g. a summary of which stores are skipped and for what reason?

What do you think?

clients/venice-admin-tool/src/main/java/com/linkedin/venice/datarecovery/StoreRepushCommand.java
```diff
@@ -158,6 +201,16 @@ public void setExtraCommandArgs(String args) {
     public void setSourceFabric(String fabric) {
       this.sourceFabric = fabric;
     }
+
+    public void setTimestamp(String timestamp) {
+      DateTimeFormatter formatter = DateTimeFormatter.ofPattern("MM-dd-yyyy HH:mm:ss");
```
lluwm, line 206: Let's also use the [DateTimeFormatter.ISO_LOCAL_DATE_TIME](https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html#ISO_LOCAL_DATE_TIME) constant here.

clients/venice-admin-tool/src/main/java/com/linkedin/venice/datarecovery/DataRecoveryClient.java
```diff
@@ -45,8 +52,35 @@ public DataRecoveryMonitor getMonitor() {
...
   }
 
+  public void filterStoresWithOngoingRepush(Set<String> storesList, StoreRepushCommand.Params params) {
+    String url = params.getUrl();
+    ControllerClient cli = params.getPCtrlCliWithoutCluster();
+    LocalDateTime timestamp = params.getTimestamp();
+    String sourceFabric = params.getSourceFabric();
+    for (String s: storesList) {
+      String clusterName = cli.discoverCluster(s).getCluster();
+      try (ControllerClient parentCtrlCli = new ControllerClient(clusterName, url, Optional.empty())) {
```
lluwm, line 62: We might want to use `params.sslFactory` here, instead of passing an empty optional.

clients/venice-admin-tool/src/main/java/com/linkedin/venice/datarecovery/DataRecoveryClient.java
lyssarzr: I added a new system to report the skipped stores before the shutdown step in a block, with error message. Let me know how this looks :)

clients/venice-admin-tool/src/main/java/com/linkedin/venice/Command.java
```diff
@@ -447,7 +447,8 @@ public enum Command {
   ),
   EXECUTE_DATA_RECOVERY(
       "execute-data-recovery", "Execute data recovery for a group of stores",
-      new Arg[] { RECOVERY_COMMAND, STORES, SOURCE_FABRIC }, new Arg[] { EXTRA_COMMAND_ARGS, DEBUG, NON_INTERACTIVE }
+      new Arg[] { URL, RECOVERY_COMMAND, STORES, DEST_FABRIC, DATETIME },
```
lluwm, line 450: SOURCE_FABRIC is useful, I think we should add DEST_FABRIC as a new Arg, but not replacing the existing SOURCE_FABRIC

clients/venice-admin-tool/src/main/java/com/linkedin/venice/datarecovery/DataRecoveryClient.java
```diff
@@ -45,17 +56,69 @@ public DataRecoveryMonitor getMonitor() {
...
+      String clusterName = cli.discoverCluster(s).getCluster();
+      try (ControllerClient parentCtrlCli = buildControllerClient(clusterName, url, params.getSSLFactory())) {
+        StoreHealthAuditResponse storeHealthInfo = parentCtrlCli.listStorePushInfo(s, false);
+        RegionPushDetails regionPushDetails = storeHealthInfo.getRegionPushDetails().get(destFabric);
+        String latestTimestamp = regionPushDetails.getPushStartTimestamp();
+        LocalDateTime latestPushStartTime = LocalDateTime.parse(latestTimestamp, DateTimeFormatter.ISO_LOCAL_DATE_TIME);
+        if (latestPushStartTime.isBefore(timestamp)) {
+          ret.put(s, Pair.of(true, ""));
+        } else {
+          ret.put(s, Pair.of(false, "ongoing repush"));
```
lluwm, line 77: if I am not wrong, `listStorePushInfo` only reports the current version info, so to determine if there is an ongoing push, we need a different api call. one example is [here](https://github.com/linkedin/venice/blob/main/clients/venice-admin-tool/src/main/java/com/linkedin/venice/datarecovery/MonitorCommand.java#LL86C20-L86C20)

clients/venice-admin-tool/src/main/java/com/linkedin/venice/Command.java
lyssarzr: Re-added this parameter

clients/venice-admin-tool/src/main/java/com/linkedin/venice/datarecovery/DataRecoveryClient.java
lyssarzr: Updated the error handling and added the additional gate based on getFutureVersions

clients/venice-admin-tool/src/main/java/com/linkedin/venice/datarecovery/DataRecoveryClient.java
```diff
@@ -45,17 +58,84 @@ public DataRecoveryMonitor getMonitor() {
...
   public void execute(DataRecoveryParams drParams, StoreRepushCommand.Params cmdParams) {
     Set<String> storeNames = drParams.getRecoveryStores();
-    if (storeNames == null || storeNames.isEmpty()) {
-      LOGGER.warn("store list is empty, exit.");
-      return;
+    Map<String, Pair<Boolean, String>> pushMap = getRepushViability(storeNames, cmdParams);
+    Set<String> filteredStoreNames = new HashSet<>();
+
+    for (Map.Entry<String, Pair<Boolean, String>> e: pushMap.entrySet()) {
+      if (e.getValue().getLeft() == true) {
```
lluwm, line 114: Nit. if (e.getValue().getLeft() == true) { -> if (e.getValue().getLeft()) {

clients/venice-admin-tool/src/main/java/com/linkedin/venice/datarecovery/DataRecoveryClient.java
```diff
@@ -45,17 +58,84 @@ public DataRecoveryMonitor getMonitor() {
...
+      LOGGER.warn("store list is empty, exit.");
     }
 
-    getExecutor().perform(storeNames, cmdParams);
+    // check if we filtered stores based on push info, report them
+    if (storeNames.size() != filteredStoreNames.size()) {
+      LOGGER.info("================");
+      LOGGER.info("STORES STORES WERE SKIPPED:");
+      for (Map.Entry<String, Pair<Boolean, String>> e: pushMap.entrySet()) {
+        if (e.getValue().getLeft() == false) {
```
lluwm, line 133: Nit. if (e.getValue().getLeft() == false) { --> if (!e.getValue().getLeft()) {

clients/venice-admin-tool/src/main/java/com/linkedin/venice/datarecovery/DataRecoveryClient.java
```diff
@@ -45,17 +58,84 @@ public DataRecoveryMonitor getMonitor() {
...
+
+    for (Map.Entry<String, Pair<Boolean, String>> e: pushMap.entrySet()) {
+      if (e.getValue().getLeft() == true) {
+        filteredStoreNames.add(e.getKey());
+      }
     }
-    if (!drParams.isNonInteractive && !confirmStores(storeNames)) {
-      return;
+
+    if (filteredStoreNames != null && !filteredStoreNames.isEmpty()) {
```
lluwm, line 119: Nit, we can skip `filteredStoreNames != null` check, because it cannot be null, (allocated several lines before @ L111)

clients/venice-admin-tool/src/main/java/com/linkedin/venice/datarecovery/DataRecoveryClient.java
```diff
@@ -114,7 +194,7 @@ public Set<String> getRecoveryStores() {
     private Set<String> calculateRecoveryStoreNames(String multiStores) {
       Set<String> storeNames = null;
       if (multiStores != null && !multiStores.isEmpty()) {
-        storeNames = Utils.parseCommaSeparatedStringToSet(multiStores);
+        storeNames = new HashSet<>(Utils.parseCommaSeparatedStringToSet(multiStores));
```
lluwm, line 197: why we need to new a hashset here?

clients/venice-admin-tool/src/main/java/com/linkedin/venice/datarecovery/StoreRepushCommand.java
```diff
@@ -59,7 +64,7 @@ private List<String> generateRepushCommand() {
     cmd.add(this.params.command);
     cmd.add(this.params.extraCommandArgs);
     cmd.add(String.format("--store '%s'", this.params.store));
-    cmd.add(String.format("--fabric '%s'", this.params.sourceFabric));
+    cmd.add(String.format("--fabric '%s'", this.params.destFabric));
```
lluwm, line 67: This is not right, we should still use the sourceFabric for this command, it specifies where we copy data from

clients/venice-admin-tool/src/test/java/com/linkedin/venice/TestDataRecoveryClient.java
```diff
@@ -2,11 +2,7 @@
 
 import static org.mockito.ArgumentMatchers.*;
 import static org.mockito.ArgumentMatchers.anyString;
-import static org.mockito.Mockito.any;
-import static org.mockito.Mockito.doCallRealMethod;
-import static org.mockito.Mockito.doReturn;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.*;
```
lluwm, line 5: avoid using * for import

clients/venice-admin-tool/src/test/java/com/linkedin/venice/TestDataRecoveryClient.java
```diff
@@ -135,9 +134,84 @@ private void estimateRecovery() {
...
+    doReturn(true).when(dataRecoveryClient).confirmStores(any());
+    doCallRealMethod().when(dataRecoveryClient).getRepushViability(any(), any());
+    doReturn(controllerClient).when(dataRecoveryClient).buildControllerClient(anyString(), anyString(), any());
+    // client executes three store recovery.
+    DataRecoveryClient.DataRecoveryParams drParams = new DataRecoveryClient.DataRecoveryParams("store1,store2,store3");
+    drParams.setNonInteractive(true);
+    dataRecoveryClient.execute(drParams, cmdParams);
+  }
+
+  private void testBlockRepush(boolean isSuccess) {
```
lluwm, line 201: looks that there are many duplicate code between `testBlockRepush` and `executeRecovery`. If possible we should abstract them out into a common function

lluwm (approved):
Alyssa, thanks for all the effort to address the comments and the offline discussions. The change looks good to me!

PR #452, opened 2023-05-26
[controller][server][clients] End-To-End support for SASL authentication with Kafka

eolivelli:
## Summary
Ensure that SASL configuration is correctly propagated to all the components

Details:
- copy sasl properties to producer, consumer and admin clients to Kafka
- add support for additional writer configuration in the Pulsar Sink
- add End-To-End test (it uses a Push job, we could follow up with all the writer APIs) 

## How was this PR tested?
There is a new end-to-end test

- [N] No. You can skip the rest of this section.


eolivelli:
All tests passed :-)

PR #453, opened 2023-05-29
[client] Allow to configure the number of connections and upgrade HttpClient to 5.x async with HTTP2 support

eolivelli:
## Summary

The Store Client currently has some limitations about performances: it uses only HTTP 1.1 and opens only at most 2 connections. This means that you cannot have more than 2 inflight requests.

Un order to achieve better latencies this patch introduces:
- configuration parameters to tune the number. of HTTP 1.x connections
- upgrade of HttpClient to 5.x, with HTTP2 support
- a configuration flag to require HTTP2

Please note that HTTP2 works only with TLS enabled on the Router  

## How was this PR tested?
The code change is almost covered by existing tests, as the core of the change is the upgrade of HC to 5.x.

Currently there is no documentation about the client

### DIFF
--- a/clients/venice-thin-client/src/main/java/com/linkedin/venice/client/store/ClientConfig.java
+++ b/clients/venice-thin-client/src/main/java/com/linkedin/venice/client/store/ClientConfig.java
@@ -60,6 +60,13 @@ public class ClientConfig<T extends SpecificRecord> {
   private boolean isHttps = false;
   private SSLFactory sslFactory = null;
 
+  // HttpTransport settings
+  private int maxConnectionsPerRoute; // only for HTTP1
+
+  private int maxConnectionsTotal; // only for HTTP1
+
+  private boolean httpClient5Http2Enabled;
+
   // Test settings
   private Time time = new SystemTime();
 
@@ -107,11 +114,15 @@ public class ClientConfig<T extends SpecificRecord> {
         // Security settings
         .setHttps(config.isHttps())
         .setSslFactory(config.getSslFactory())
-
         .setForceClusterDiscoveryAtStartTime(config.isForceClusterDiscoveryAtStartTime())
         .setProjectionFieldValidationEnabled(config.isProjectionFieldValidationEnabled())
         .setPreferredSchemaFilter(config.getPreferredSchemaFilter().orElse(null))
         .setSchemaRefreshPeriod(config.getSchemaRefreshPeriod())
+
+        // HttpTransport settings
+        .setMaxConnectionsPerRoute(config.getMaxConnectionsPerRoute())
+        .setMaxConnectionsTotal(config.getMaxConnectionsTotal())
+        .setHttpClient5Http2Enabled(config.isHttpClient5Http2Enabled())
 
         // Test settings
         .setTime(config.getTime());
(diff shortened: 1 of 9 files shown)


FelixGV (commented):
Thanks for the PR Enrico!

This looks fine in theory but I wonder if there's a risk. We've occasionally seen some instabilities with httpclient5, and I wonder if we should keep the support for the previous version. Since the `TransportClient` is pluggable, and the package names are different, it seems feasible to support both.

This doesn't really affect us within LI since we use the `D2TransportClient` anyway, so I'm not sure what's the likelihood that HC5 could have a regression compared to 4.

@gaojieliu do you have an opinion on the above?

gaojieliu:
> Thanks for the PR Enrico!
> 
> This looks fine in theory but I wonder if there's a risk. We've occasionally seen some instabilities with httpclient5, and I wonder if we should keep the support for the previous version. Since the `TransportClient` is pluggable, and the package names are different, it seems feasible to support both.
> 
> This doesn't really affect us within LI since we use the `D2TransportClient` anyway, so I'm not sure what's the likelihood that HC5 could have a regression compared to 4.
> 
> @gaojieliu do you have an opinion on the above?

Yeah, what you said makes sense.
Let us keep both httpsyncclient (for http/1.1) and httpclient5 for (http2) for the time being, since we did discover a couple issues with httpclient5 in the past.
Once the new httpclient5 based implementation is stablized, we could delete the httpasyncclient based imp.

clients/venice-thin-client/src/main/java/com/linkedin/venice/client/store/ClientConfig.java
```diff
@@ -60,6 +60,13 @@
   private boolean isHttps = false;
   private SSLFactory sslFactory = null;
 
+  // HttpTransport settings
+  private int maxConnectionsPerRoute = 2; // AsyncHttpClient Default, only for HTTP1
+
+  private int maxConnectionsTotal = 20; // AsyncHttpClient Default, only for HTTP1
+
```
gaojieliu, line 67: So Venice Thin Client doesn't AsyncHttpClient any more, right?
I guess you can remove `AsyncHttpClient` from the javadoc.

clients/venice-thin-client/src/main/java/com/linkedin/venice/client/store/ClientConfig.java
```diff
@@ -60,6 +60,13 @@
...
+  private boolean requireHTTP2;
```
gaojieliu, line 68: Can this config to be more specific? such as `httpClient5Http2Enabled`? since D2 based impl doesn't read this config.

clients/venice-thin-client/src/main/java/com/linkedin/venice/client/store/transport/HttpsTransportClient.java
```diff
@@ -2,21 +2,54 @@
...
   }
 
+  private static CloseableHttpAsyncClient buildClient(
+      int maxConnectionsTotal,
+      int maxConnectionsPerRoute,
+      boolean requireHHtp2,
+      SSLFactory sslFactory) {
+    if (requireHHtp2) {
+      // HTTP2 only client doesn't allow you to configure the ConnectionManager
+      return HttpAsyncClients.customHttp2()
```
gaojieliu, line 37: Can we leverage this utility class to instantiate a httpclient5 based h2 client?
https://github.com/linkedin/venice/blob/main/internal/venice-client-common/src/main/java/com/linkedin/venice/httpclient5/HttpClient5Utils.java

The utility class has been proved inside LinkedIn.

gaojieliu (changes requested):

eolivelli:
@gaojieliu @FelixGV I have rebased onto the latest main branch.

After the discussion we had yesterday at the community meeting I believe that patch is good to go.
We said that at LI nobody is currently using this Transport and it is fine to upgrade HC5 to fine.

DataStax will provide feedback on the stability of this client, as we are going to use this for our testing and for our users

clients/venice-thin-client/src/main/java/com/linkedin/venice/client/store/transport/HttpTransportClient.java
```diff
@@ -30,14 +33,18 @@
  * {@link CloseableHttpAsyncClient} based TransportClient implementation.
  */
 public class HttpTransportClient extends TransportClient {
-  private final Logger logger = LogManager.getLogger(HttpTransportClient.class);
+  private final static Logger LOGGER = LogManager.getLogger(HttpTransportClient.class);
```
FelixGV, line 36: Thanks, that's better.

FelixGV (approved):

gaojieliu (approved):

PR #454, opened 2023-05-30
[test][server][dvc] Change test dependency and add consumer and admin adapter to PubSubClientFactory.

haoxu07:
This PR focus on:
1) Change dependency of `MapreduceClientJobClient` from `testImplementation` to implementation, as internal project needs to depend on this module and run test implementation. 
2) Add `PubSubConsumerAdapterFactory` and `PubSubAdminAdapterFactory` to `PubSubClientFactory`'s constructor.
3) Change `VeniceController` constructor to make `PubSubClientFactory` configurable for test and downstream project depending on it. 

Resolves #XXX

## How was this PR tested?

internal CI

### DIFF
--- a/internal/venice-test-common/src/integrationtest/java/com/linkedin/venice/integration/utils/KafkaBrokerFactory.java
+++ b/internal/venice-test-common/src/integrationtest/java/com/linkedin/venice/integration/utils/KafkaBrokerFactory.java
@@ -3,6 +3,8 @@ package com.linkedin.venice.integration.utils;
 import static com.linkedin.venice.integration.utils.ProcessWrapper.DEFAULT_HOST_NAME;
 
 import com.linkedin.venice.exceptions.VeniceException;
+import com.linkedin.venice.pubsub.adapter.kafka.admin.ApacheKafkaAdminAdapterFactory;
+import com.linkedin.venice.pubsub.adapter.kafka.consumer.ApacheKafkaConsumerAdapterFactory;
 import com.linkedin.venice.pubsub.adapter.kafka.producer.ApacheKafkaProducerAdapterFactory;
 import com.linkedin.venice.pubsub.api.PubSubClientsFactory;
 import com.linkedin.venice.utils.KafkaSSLUtils;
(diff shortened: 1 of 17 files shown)


services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerContext.java
```diff
@@ -0,0 +1,170 @@
...
+import com.linkedin.venice.utils.VeniceProperties;
+import io.tehuti.metrics.MetricsRepository;
+import java.util.Collections;
+import java.util.List;
+
+
+/**
+ * VeniceControllerContext contains dependencies required by {@link VeniceController}
+ */
+public class VeniceControllerContext {
```
ZacAttack, line 23: So the usual pattern is that the constructor args to a builder like this are the mandatory configs and everything else is optional.  It seems to be implied by this code that everything is optional, but I'm not sure that thats true?  My sense is that maybe most of these are optional (d2client, authorizerservice, etc.) but can a controller possibly work without the pubsubclientsfactory?

haoxu07: A controller will not work without `PubSubClientsFactory`, as the consumer, producer and admin adapter factory are all from it. We use the default setting of `PubSubClientsFactory` with `ApacheKafkaAdapters`. In integration test, it comes from `PubSubBrokerFactory#getClientsFactory()` by
`System.getProperty(PUBSUB_BROKER_FACTORY_FQCN, "com.linkedin.venice.integration.utils.KafkaBrokerFactory");`, we plan to override this property if we want to use other `PubSubClientFactory`.

ZacAttack: Wait, we're using a jvm variable to configure this?  Is that in test only or production code?

haoxu07: This is only for test only code. In production, this PR has changed the constructor of `VeniceController` and `VeniceServer` , specific `PubSubClientFactory` can be passed by our internal MP.

internal/venice-test-common/src/integrationtest/java/com/linkedin/venice/integration/utils/KafkaBrokerFactory.java
```diff
@@ -139,7 +143,10 @@ private KafkaBrokerWrapper(
       this.sslPort = sslPort;
       this.zkServerWrapper = zkServerWrapper;
       this.shouldCloseZkServer = shouldCloseZkServer;
-      pubSubClientsFactory = new PubSubClientsFactory(new ApacheKafkaProducerAdapterFactory());
+      pubSubClientsFactory = new PubSubClientsFactory(
+          new ApacheKafkaProducerAdapterFactory(),
```
sixpluszero, line 147: I see this pattern is used everywhere. Why not just have a new PubSubClientsFactory() constructor that takes 0 argument and do the initialization here?

haoxu07: The expectation of the `PubSubClientFactory` is to take different combination of PubSubAdapters for consumer, producer or admin. As for a specific component like controller, we may want to change just producer, this `PubSubClientFactory` can support hybrid clients for one component.

sixpluszero (approved):

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerContext.java
haoxu07: Add null checking in `VeniceController` constructor to force passing instances in.

sixpluszero (approved):

PR #455, opened 2023-05-30
[controller] Make Controller Resilient to Parent Broker Unavailability

ZacAttack:
## Make Controller Resilient to Parent Broker Unavailability
Fix for prod issue when decomming old parent broker. This should be all thats needed based on what calls we make to read the controller parent broker list. It's ok to barf an exception here at time of truncation and move on as if/when the broker comes back, topic clean up service will run through and clean up anything left behind.


lluwm (approved):
Zac, thanks for the fix and it looks good to me!

PR #459, opened 2023-06-01
[controller] system store admin messages should be processed in the same queue as their user store

xunyin8:
## Summary,

1. Execution order of admin messages for user store and their corresponding system store matters.

If the messages are processed in separate queues in AdminConsumptionTask then the following race is possible: expected order: create user store -> empty push to system store -> delete user store execution order: empty push to system store -> create user store -> delete user store or create user store -> delete user store -> empty push to system store.

The empty push to system store will fail due to VeniceNoStoreException.

2. Added new unit test for similar scenario as the race described above.

## How was this PR tested?
Existing tests and added a new test

### DIFF
--- a/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
@@ -6265,8 +6265,9 @@ public class VeniceHelixAdmin implements Admin, StoreCleaner {
     if (storeName == null) {
       return getLastSucceedExecutionId(clusterName);
     } else {
+      String userStoreName = VeniceSystemStoreType.extractUserStoreName(storeName);
       return adminConsumerServices.containsKey(clusterName)
-          ? adminConsumerServices.get(clusterName).getLastSucceededExecutionId(storeName)
+          ? adminConsumerServices.get(clusterName).getLastSucceededExecutionId(userStoreName)
           : null;
     }
   }
--- a/services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/consumer/AdminConsumptionTask.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/consumer/AdminConsumptionTask.java
@@ -829,10 +829,7 @@ public class AdminConsumptionTask implements Runnable, Closeable {
                   + " because it does not contain a storeName field");
         }
     }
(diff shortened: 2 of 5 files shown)


services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/consumer/AdminConsumptionTask.java
```diff
@@ -828,6 +829,10 @@ private String extractStoreName(AdminOperation adminOperation) {
                   + " because it does not contain a storeName field");
         }
     }
+    VeniceSystemStoreType systemStoreType = VeniceSystemStoreType.getSystemStoreType(storeName);
```
majisourav99, line 832: can we add this as an util method, so many places does this regular store name extraction ?

majisourav99 (commented):
LGTM. left a minor comment.

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/consumer/AdminConsumptionTask.java
xunyin8: Will do

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -6265,8 +6265,9 @@ public Long getLastSucceededExecutionId(String clusterName, String storeName) {
     if (storeName == null) {
       return getLastSucceedExecutionId(clusterName);
     } else {
+      String userStoreName = VeniceSystemStoreType.extractUserStoreName(storeName);
       return adminConsumerServices.containsKey(clusterName)
-          ? adminConsumerServices.get(clusterName).getLastSucceededExecutionId(storeName)
```
majisourav99, line 6269: Good catch! but depending on the deployment of this fix we may or may not get the execution ID ?

xunyin8: This code path is only used by the parent controller when pending for local verification of the admin message is written successfully. I don't think deployment order matters here but feel free to give a concrete example if I missed something.

majisourav99: I see this is used only in-memory map query in parent controller, so it should be fine.

majisourav99 (approved):
LGTM. Thanks Xun

PR #460, opened 2023-06-02
[server] Fix isolation ingestion regression during startup

sixpluszero:
## [server] Fix isolation ingestion regression during startup
Recent improvement on ingestion isolation logic introduces a regression found in test server: When performing resource handover, we will first wait until pending async ingestion task count for the resource goes down to 0 in order to avoid race condition between handover and stop consumption request. However, for existing resource, it could report COMPLETED right away when handling startConsumption ingestion action. If we try to wait in the same thread, the pending action count will never go down to 0 until the wait times out for 30s. This leads to all ingestion tasks waiting for 30s before being able to shutdown resource and report completion and it took EI server forever to complete bootstrap.

This PR fixes the issue by making sure the wait pending ingestion action is also executed in an async way, same as following stopConsumption and closeStoragePartition call that's already executed in async. With this change, the report thread will never be blocked and pending action will go down to 0 after the action is completed and removed from consumerActionQueue. The async wait method will then be able to complete fast.

## How was this PR tested?

Modify existing unit test.

Perform FrankenWar in test server, verified that bootstrap is completed fast and no STARTED replica lingering after a short period. 


nisargthakkar (approved):
LGTM

PR #462, opened 2023-06-02
[build] Add protobuf compiler plugin and required dependencies for generating gRPC Java stubs

rkunds:
## How was this PR tested?

wpi: Local and GHCI


services/venice-server/build.gradle
```diff
@@ -63,6 +70,25 @@ jar {
   }
 }
 
+def grpcVersion = '1.54.1'
+def protobufVersion = '3.21.7'
+
+protobuf {
```
sushantmane, line 76: let's move this to the parent `build.gradle` and then make this plugin and task available via `subprojects {..}`

build.gradle
```diff
@@ -114,7 +119,13 @@ ext.libraries = [
...
     zkclient: 'com.101tec:zkclient:0.7', // For Kafka AdminUtils
     zookeeper: 'org.apache.zookeeper:zookeeper:3.5.9',
-    zstd: 'com.github.luben:zstd-jni:1.5.2-3'
+    zstd: 'com.github.luben:zstd-jni:1.5.2-3',
+
+    grpcNettyShaded: 'io.grpc:grpc-netty-shaded:' + grpcVersion,
+    grpcProtobuf: 'io.grpc:grpc-protobuf:' + grpcVersion,
+    grpcStub: 'io.grpc:grpc-stub:' + grpcVersion,
+    tomcatAnnotations: 'org.apache.tomcat:annotations-api:6.0.53',
+    guava: 'com.google.guava:guava:30.1-jre'
```
sushantmane, line 128: We don't need `guava` dependency for gRPC. Let's remove it.

build.gradle
```diff
@@ -201,7 +212,6 @@ subprojects {
     }
     compileOnly {
       // These dependencies are transitively used at runtime, so we cannot exclude them further than compileOnly
-      exclude group: 'com.google.guava'
```
sushantmane, line 204: this might cause dependency conflict. Let's not remove this line

sushantmane (changes requested):
Congrats on your first PR! I've left some comments.

build.gradle
```diff
@@ -114,7 +119,13 @@ ext.libraries = [
     xerces: 'xerces:xercesImpl:2.9.1',
     zkclient: 'com.101tec:zkclient:0.7', // For Kafka AdminUtils
     zookeeper: 'org.apache.zookeeper:zookeeper:3.5.9',
-    zstd: 'com.github.luben:zstd-jni:1.5.2-3'
+    zstd: 'com.github.luben:zstd-jni:1.5.2-3',
+
+    grpcNettyShaded: 'io.grpc:grpc-netty-shaded:' + grpcVersion,
+    grpcProtobuf: 'io.grpc:grpc-protobuf:' + grpcVersion,
+    grpcStub: 'io.grpc:grpc-stub:' + grpcVersion,
+    tomcatAnnotations: 'org.apache.tomcat:annotations-api:6.0.53',
```
adamxchen, line 127: is this dependency needed? Seems nothing related to gPRC. 😕

services/venice-server/build.gradle
rkunds: done, moved dependencies and tasks to parent `build.gradle` file.

build.gradle
rkunds: To my understanding it should be necessary to include this as a `compileOnly` dependency to obtain annotations for generated code. For reference https://github.com/grpc/grpc-java/issues/9179.

adamxchen: Got it.

sushantmane (approved):
LGTM. Thanks, @rkunds!

PR #463, opened 2023-06-02
[test] Fix argument match bug in TestDataRecoveryClient.testMon…

lluwm:
…itor

mockito.anyString matches any non-null String, but in the unit test, a null String is possible, thus cannot match the function signature, which leads to a Nullpointer in later steps.

This rb fixes the issue by using a mockito.any() to include null.

## How was this PR tested?
Internal CI (Venice-WcTest8, #1223; Venice-WcTest, #8033, no TestDataRecoveryClient.testMonitor failure) 

### DIFF
--- a/clients/venice-admin-tool/src/test/java/com/linkedin/venice/TestDataRecoveryClient.java
+++ b/clients/venice-admin-tool/src/test/java/com/linkedin/venice/TestDataRecoveryClient.java
@@ -217,7 +217,7 @@ public class TestDataRecoveryClient {
     doReturn(storeResponse).when(mockedCli).getStore(anyString());
     doReturn(discoveryResponse).when(mockedCli).discoverCluster(anyString());
     doReturn(statusResponse).when(mockedCli).getFutureVersions(anyString(), anyString());
-    doReturn(queryResponse).when(mockedCli).queryDetailedJobStatus(anyString(), anyString());
+    doReturn(queryResponse).when(mockedCli).queryDetailedJobStatus(any(), anyString());
     doReturn(storeHealthResp).when(mockedCli).listStorePushInfo(anyString(), anyBoolean());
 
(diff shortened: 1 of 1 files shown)


clients/venice-admin-tool/src/test/java/com/linkedin/venice/TestDataRecoveryClient.java
```diff
@@ -217,7 +217,7 @@ private void monitorRecovery(ExecutionStatus status, boolean isCurrentVersionNew
     doReturn(storeResponse).when(mockedCli).getStore(anyString());
     doReturn(discoveryResponse).when(mockedCli).discoverCluster(anyString());
     doReturn(statusResponse).when(mockedCli).getFutureVersions(anyString(), anyString());
-    doReturn(queryResponse).when(mockedCli).queryDetailedJobStatus(anyString(), anyString());
+    doReturn(queryResponse).when(mockedCli).queryDetailedJobStatus(any(), anyString());
```
sixpluszero, line 220: The fix itself looks good to me, but I am curious - why it is not failing consistently in this case? Is the check itself also not deterministic in terms of input?

lluwm: That's a good question and I don't know the exact reason, I suspect it might be some optimization done by the JVM compiler. I also did some reproduction experiments myself, e.g. run the same tests 1000 times with/out changing it from anyString -> any, and I did see few failure cases if I didn't change, but no failure if I changed.

sixpluszero: okay...that's weird, but I guess we could live with it for now :)

sixpluszero (approved):

sixpluszero:
Oh, it would be great if you could tweak the commit title, so it does not break into two lines.

PR #464, opened 2023-06-02
[vpj] Consolidate controller requests for retrieving store info

adamxchen:

In the VPJ, we call the controller 4 times to get the `StoreResponse` which is wasteful, so this PR caches the response so we can reduce the duplicate calls by 2.

## How was this PR tested?

GHCI


adamxchen:
The changes aren't very clean as I want it to be but it requires more refactoring, e.g. either use class variables more extensive or reduce the scope, right now, it's mixed. We can start from here.

adamxchen:
The merge conflicts messed up with the setup. Going to have a new PR for this.

PR #465, opened 2023-06-03
[test] Fix flaky test PushStatusStoreTest#testIncrementalPushStatusStoredInPushStatusStore

sixpluszero:
## [test] Fix flaky test PushStatusStoreTest#testIncrementalPushStatusStoredInPushStatusStore
This test is flaky because when new push job is completed, router will refresh its metadata, if the client get happens faster than metadata refresh, Router will throw 5xx error indicating the resource does not exist.
This can be fixed in two ways:
(1) Add retry to client get
(2) Force router metadata refresh before client call.
This PR takes the 1st approach.

There might be other flaky tests in the codebase having the same symptom, will tackle them once we located it in test log.

## How was this PR tested?
Locally run the integration test 20 times. Before fix, every 10 invocation there will be 2-3 error due to resource not exist. After the test, all test passed.


FelixGV (approved):
Awesome, thanks Jialin!

PR #467, opened 2023-06-05
[test]Fix flakiness of fastclient/AvroStoreClientEndToEndTest::testFastClientGet

m-nagarajan:
## Summary
The metrics checks are done too early considering the async nature of metrics increment. Adding a non deterministic wait to check these metrics. testFastClientWithLongTailRetry in this class is still flaky, but that needs more time. 

## How was this PR tested?
GH CI


sixpluszero (approved):
Thanks Manoj! 

TIL metrics action in the client impl is done in kind of async way

m-nagarajan:
Thanks Jialin for the review.

PR #468, opened 2023-06-06
[vpj][doc] Update targeted region push doc

adamxchen:

Update some doc for targeted region push. Removed some contexts that are still under development. Will add them back once we ship them.

## How was this PR tested?

GHCI.

### DIFF
--- a/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
+++ b/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
@@ -393,7 +393,7 @@ public class VenicePushJob implements AutoCloseable {
 
   /**
    * Config to enable single targeted region push mode in VPJ.
-   * In this mode, the VPJ will only push data to a single colo/region.
+   * In this mode, the VPJ will only push data to a single region.
    * The single region is decided by the store config in {@link StoreInfo#getNativeReplicationSourceFabric()}}.
    * For multiple targeted regions push, may use the advanced mode. See {@link #TARGETED_REGION_PUSH_LIST}.
    */
--- a/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
+++ b/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
@@ -403,7 +403,7 @@ public class VenicePushJob implements AutoCloseable {
    * This is experimental config to specify a list of regions used for targeted region push in VPJ.
    * {@link #TARGETED_REGION_PUSH_ENABLED} has to be enabled to use this config.
    * In this mode, the VPJ will only push data to the provided regions.
-   * The input should be split by comma, e.g. "dc-0,dc-1,dc-2".
+   * The input should be split by comma, e.g. "dc-0, dc-1, dc-2".
    * For single targeted region push, see {@link #TARGETED_REGION_PUSH_ENABLED}.
    */
   public static final String TARGETED_REGION_PUSH_LIST = "targeted.region.push.list";
--- a/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
+++ b/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
@@ -58,6 +58,7 @@ import com.linkedin.venice.message.KafkaKey;
 import com.linkedin.venice.meta.BufferReplayPolicy;
 import com.linkedin.venice.meta.HybridStoreConfig;
 import com.linkedin.venice.meta.Store;
+import com.linkedin.venice.meta.StoreInfo;
 import com.linkedin.venice.meta.Version;
 import com.linkedin.venice.partitioner.DefaultVenicePartitioner;
 import com.linkedin.venice.partitioner.VenicePartitioner;
@@ -392,9 +393,8 @@ public class VenicePushJob implements AutoCloseable {
 
   /**
    * Config to enable single targeted region push mode in VPJ.
-   * In this mode, the VPJ will only push data to a single colo/region first, perform validation and finally
-   * leverage data recovery/repush to propagate data globally.
-   * The single solo/region is decided by the store config in {@link StoreResponse#getStore()}}.
+   * In this mode, the VPJ will only push data to a single colo/region.
+   * The single region is decided by the store config in {@link StoreInfo#getNativeReplicationSourceFabric()}}.
    * For multiple targeted regions push, may use the advanced mode. See {@link #TARGETED_REGION_PUSH_LIST}.
    */
   public static final String TARGETED_REGION_PUSH_ENABLED = "targeted.region.push.enabled";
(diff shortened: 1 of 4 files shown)


clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
```diff
@@ -392,20 +393,18 @@ public class VenicePushJob implements AutoCloseable {
 
   /**
    * Config to enable single targeted region push mode in VPJ.
-   * In this mode, the VPJ will only push data to a single colo/region first, perform validation and finally
-   * leverage data recovery/repush to propagate data globally.
-   * The single solo/region is decided by the store config in {@link StoreResponse#getStore()}}.
+   * In this mode, the VPJ will only push data to a single colo/region.
```
FelixGV, line 396: Nitpick: let's remove "colo" from here.

docs/user_guide/write_api/push_job.md
```diff
@@ -9,10 +9,11 @@ permalink: /docs/user_guide/write_api/push_job
 The Push Job takes data from a Hadoop grid and writes it to Venice. 
 
 ## Use Cases
-There are two modes the Push Job can run in:
+There are three modes the Push Job can run in:
 
 - Full Push (default)
 - Incremental Push
+- Targeted Region Push
```
FelixGV, line 16: Hmm... I think it's confusing to put this here. Targeted pushes are an option on Full Pushes, but in theory they could apply to inc pushes as well (with more changes, which we're not working on ATM). So IMHO it is orthogonal to inc/full push, rather than a third mode.

docs/user_guide/write_api/push_job.md
```diff
@@ -27,6 +28,14 @@ When performing an Incremental Push, no new dataset versions are created, and da
 of the dataset. This leverages the same mechanism as Streaming Writes, and requires that the store be configured as 
 Hybrid.
 
+### Targeted Region Push
+Technically, targeted region push is a type of full push, but it allows writing data into a subset of global regions/data
+centers, whereas full push writes globally at once. 
+
+Please note that this feature is still under active development. It doesn't push data to the rest of unspecified regions
```
FelixGV, line 35: Let's use "functionality" instead of "feature" since the latter is an overloaded term in ML parlance.

docs/user_guide/write_api/push_job.md
```diff
@@ -53,6 +62,9 @@ The user may choose to specify the following configs:
 - `extended.schema.validity.check.enabled`: Whether to perform extended schema validation on the input (equivalent to
   the `STRICT` mode in avro-util's [SchemaParseConfiguration](https://github.com/linkedin/avro-util/blob/master/helper/helper-common/src/main/java/com/linkedin/avroutil1/compatibility/SchemaParseConfiguration.java)). 
   If set to `false`, it becomes equivalent to avro-util's `LOOSE` mode. Default: `true`
+- `targeted.region.push.enabled`: Whether to perform targeted region push. Default: `false`
+- `targeted.region.push.list`: Optionally specify a list of target region(s) to push data into. See full details at 
+  [TARGETED_REGION_PUSH_LIST](https://venicedb.org/javadoc/com/linkedin/venice/hadoop/VenicePushJob.html#TARGETED_REGION_PUSH_LIST).
```
FelixGV, line 67: Pretty cool to link into Javadoc directly. Just wondering: can we remove the domain name, so that it links relatively from whatever domain it's on? If this works it could make it easier to test doc links on our personal forks (even if linking to Javadoc that only exists in that branch).

FelixGV (commented):
Thanks for the doc changes! This is a great habit to form! Left some minor comments.

docs/user_guide/write_api/push_job.md
adamxchen: I agree this may be confusing if we consider it as 3rd mode. I'd like to still introduce this in this `push_job.md` though since we don't need to create a new file for each new feature. Let me move this to a new section.

docs/user_guide/write_api/push_job.md
adamxchen: I couldn't find a way.. I tried to use relative path but somehow it will show the plain HTML file, instead of rendering it.. 
I think it has something to do with our documentation theme but not sure which config is related... 

Example: https://github.com/adamxchen/venice/blob/javadoc/docs/javadoc/com/linkedin/venice/hadoop/VenicePushJob.html#TARGETED_REGION_PUSH_LIST

docs/user_guide/write_api/push_job.md
```diff
@@ -27,6 +27,14 @@ When performing an Incremental Push, no new dataset versions are created, and da
 of the dataset. This leverages the same mechanism as Streaming Writes, and requires that the store be configured as 
 Hybrid.
 
+### Targeted Region Push
+Technically, targeted region push is an option of full push _(hence not a new mode)_, but it allows writing data into a
```
majisourav99, line 31: Are we not supporting target colo with inc_push ?

docs/user_guide/write_api/push_job.md
```diff
@@ -53,6 +61,9 @@ The user may choose to specify the following configs:
 - `extended.schema.validity.check.enabled`: Whether to perform extended schema validation on the input (equivalent to
   the `STRICT` mode in avro-util's [SchemaParseConfiguration](https://github.com/linkedin/avro-util/blob/master/helper/helper-common/src/main/java/com/linkedin/avroutil1/compatibility/SchemaParseConfiguration.java)). 
   If set to `false`, it becomes equivalent to avro-util's `LOOSE` mode. Default: `true`
+- `targeted.region.push.enabled`: Whether to perform targeted region push. Default: `false`
+- `targeted.region.push.list`: Optionally specify a list of target region(s) to push data into. See full details at
```
majisourav99, line 65: Mention this config takes effect only target push is enabled, otherwise not clear what `optionally` means here

docs/user_guide/write_api/push_job.md
FelixGV: Got it, no worries, thanks for trying.

docs/user_guide/write_api/push_job.md
adamxchen: We don't support it now so better just not to mention it?

docs/user_guide/write_api/push_job.md
adamxchen: Updated.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
```diff
@@ -392,20 +393,18 @@ public class VenicePushJob implements AutoCloseable {
...
    */
   public static final String TARGETED_REGION_PUSH_ENABLED = "targeted.region.push.enabled";
 
   /**
    * This is experimental config to specify a list of regions used for targeted region push in VPJ.
    * {@link #TARGETED_REGION_PUSH_ENABLED} has to be enabled to use this config.
-   * In this mode, the VPJ will only push data to the provided regions first, perform validation and finally
-   * leverage data recovery/repush to propagate data globally.
+   * In this mode, the VPJ will only push data to the provided regions.
    * The input should be split by comma, e.g. "dc-0,dc-1,dc-2".
```
majisourav99, line 406: nit: The input is comma separated list of region names

adamxchen: updated!

majisourav99 (approved):
Thanks Adam!

FelixGV (approved):

PR #469, opened 2023-06-06
[server] Do not throw exception in MetaStoreNotifier for DROPPED action when store is being deleted

sixpluszero:
## [server] Do not throw exception in MetaStoreNotifier for DROPPED action when store is being deleted
Found this some time ago in Venice Server in test env with (both ingestion isolation on and off)
This may result in OFFLINE->DROPPED being rollbacked by Helix mechanism, which results in storage engine lingering. (Although it might eventually be cleaned up by cleaner service, I still think it is not good action to have).

## How was this PR tested?

Existing tests.

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/notifier/MetaSystemStoreReplicaStatusNotifier.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/notifier/MetaSystemStoreReplicaStatusNotifier.java
@@ -33,7 +33,7 @@ public class MetaSystemStoreReplicaStatusNotifier implements VeniceNotifier {
     this.instance = instance;
   }
 
-  private void report(String kafkaTopic, int partitionId, ExecutionStatus status) {
+  void report(String kafkaTopic, int partitionId, ExecutionStatus status) {
     String storeName = Version.parseStoreFromKafkaTopicName(kafkaTopic);
     VeniceSystemStoreType systemStoreType = VeniceSystemStoreType.getSystemStoreType(storeName);
     if (systemStoreType != null && systemStoreType.equals(VeniceSystemStoreType.META_STORE)) {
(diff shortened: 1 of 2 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/notifier/MetaSystemStoreReplicaStatusNotifier.java
```diff
@@ -39,7 +40,21 @@ private void report(String kafkaTopic, int partitionId, ExecutionStatus status)
...
+    try {
+      store = storeRepository.getStoreOrThrow(storeName);
+    } catch (VeniceNoStoreException e) {
+      /**
+       * For store deletion, store removal in store repo might happen faster then {@link ExecutionStatus.DROPPED} action.
+       * This is to make sure the meta store notifier does not throw exception when this happens. For other status, this
+       * is not expected, so we should throw exception.
+       */
+      if (status.equals(ExecutionStatus.DROPPED)) {
+        LOGGER.warn("Store {} does not exist in store repository. Skip reporting status: {}", storeName, status);
```
FelixGV, line 53: It seems like this is a benign scenario. If this is the right understanding, then I would prefer a lower log level than WARN. Perhaps INFO, or even DEBUG?

FelixGV (commented):
Thanks Jialin!

clients/da-vinci-client/src/main/java/com/linkedin/davinci/notifier/MetaSystemStoreReplicaStatusNotifier.java
sixpluszero: Sure, will update it to INFO then. I don't think DEBUG will ever be printed out with current Venice Component logging level, since this is not excessive, I think INFO should be fine.

FelixGV (approved):
Thanks!

PR #470, opened 2023-06-06
[all] Do not create schema refresh executors unless necessary

nisargthakkar:
## Do not create schema refresh executors unless necessary
It was noticed that many threads were being created from `RouterBackedSchemaReader` which caused OOM issues in controllers. While the root cause of why so many threads were being created and not getting cleaned up is still unknown, this is a preemptive optimization to not create executors where they are unnecessary

## How was this PR tested?

GH CI


FelixGV (approved):
Thanks Nisarg!

FelixGV (approved):

PR #471, opened 2023-06-06
[controller][router][server][security] VIP-1 Implementation and support for JWT Authentication

eolivelli:
## Summary
This PR adds the API introduced by VIP-1 and an implementation of AuthenticationService based on JSON Web Tokens (JWT). The implementation covers authentication end-to-end on all the components and clients (also some CLI tools).
It is actually the porting of the work that has been done in the DataStax fork.

Changes, in order of importance, high level APIs:

**AuthenticationService**
- Add the new AuthenticationService API
- Implement the lifecycle of the AuthenticationService on the controller, server and router components
- Add the new ClientAuthenticationProvider API, this API allows to configure pluggable Authentication mechanisms on all the clients (ControllerClient and the other clients)
- Also ClientAuthenticationProvider is used in Router to Server communications
- ClientAuthenticationProvider can be loaded from the Properties of the component with a standard factory (router, pushjob....)
- AuthenticationService and ClientAuthenticationProvider don't come into play while talking to Kafka (PubSub)

**AuthorizerService**
- Implement the lifecycle of the AuthorizerService on the controller, server and router components
- Ensure that AuthorizerService is always executed (having authentication without authorisation is almost useless) 
- Keep DynamicAccessController, do not touch it and the way it is called, this is out of the scope of this work
- Add a new method "isSuperUser" to the AuthorizerService API, to cover some Controller APIs
- Provide a simple implementation that requires that the Principal is not null (so basically the user MUST be authenticated), this is useful for tests, especially while focusing on the AuthenticationService part

**JWT AuthenticationService**
- Implement an AS that validates the HTTP Authorisation header, with a Bearer token
- The implementation is inspired by Apache Pulsar
- Add a ClientAuthenticationProvider that injects the HTTP Header
- On the client the token can be a String or a path to a local file
- Add end-to-end tests
- Allow CLI tools (Admin and Query) to accept an optional "--token"  parameter (in the future it would be great to allow a configuration file, so that we could you the ClientAuthenticationProvider factory)

Other changes:
- PulsarSink, allow to configure the "token"
- Samza, OnlineProducer, PushJob: you can configure the ClientAuthenticationProvider with the Factory 

Follow up works (not implemented in this patch):
- implement a AuthenticationService that extracts the TLS X509 certificate and remove all the hard coded references to it (extractCertificate...), it can build a Principal whose name is extracted fro the certificate
- Implement an AuthorizerService that implement the ACLs and works with any Principal and not only X509 certificates 
- Add a way on the client tools to read a configuration file with the most common parameters (service urls, clusterName, authentication provider configuration....)

## How was this PR tested?
This test adds unit tests and end-to-end tests.

Also on k8s you can fully configure JTW authentication while deploying Venice with an [helm chart](https://github.com/datastax/venice-helm-chart) 

List of new configuration options:
- authentication.service.class: class name of the AuthenticationService
- authorizer.service.class: class name of the AuthorizerService

JWT AuthenticationService
- authentication.jwt.secretKey: secret key to validate tokens
- authentication.jwt.publicKey: public key to validate tokens
-  authentication.jwt.authClaim: claim to search for into the token
- authentication.jwt.publicAlg: algorithm
- authentication.jwt.audienceClaim: clam that represents the "audience"
- authentication.jwt.audience: audience
- authentication.jwt.jwksHostsAllowlist: list of valid hosts for external validation of the tokens
      
JWT client side authentication (usable on the router configuration and on jobs or clients):
- authentication.token: token or path to a token


internal/venice-common/src/main/java/com/linkedin/venice/authentication/AuthenticationService.java
```diff
@@ -0,0 +1,30 @@
...
+  }
+
+  default Principal getPrincipalFromHttpRequest(HttpRequestAccessor requestAccessor) {
+    return null;
+  }
+
+  interface HttpRequestAccessor {
+    String getHeader(String headerName);
+
+    X509Certificate getCertificate();
```
FelixGV, line 27: I don't see any code calling this function... is it needed?

internal/venice-common/src/main/java/com/linkedin/venice/authentication/jwt/TokenAuthenticationService.java
```diff
@@ -0,0 +1,62 @@
...
+  @Override
+  public void initialise(VeniceProperties veniceProperties) throws Exception {
+    TokenProperties tokenProperties = new TokenProperties(
+        veniceProperties.getString("authentication.jwt.secretKey", ""),
+        veniceProperties.getString("authentication.jwt.publicKey", ""),
+        veniceProperties.getString("authentication.jwt.authClaim", ""),
+        veniceProperties.getString("authentication.jwt.publicAlg", ""),
+        veniceProperties.getString("authentication.jwt.audienceClaim", ""),
+        veniceProperties.getString("authentication.jwt.audience", ""),
+        veniceProperties.getString("authentication.jwt.jwksHostsAllowlist", ""));
```
FelixGV, line 28: Should we have constants (e.g. probably in `ConfigKeys`) for these?

FelixGV (commented):
LGTM overall. I left a few minor comments. Please take a look. Would be good also to rebase and fix the test/coverage issues. Thanks for this major contribution!

clients/venice-admin-tool/src/main/java/com/linkedin/venice/Arg.java
```diff
@@ -11,6 +11,7 @@ public enum Arg {
   URL("url", "u", true, "Venice url, eg. http://localhost:1689  This can be a router or a controller"),
   VENICE_ZOOKEEPER_URL("venice-zookeeper-url", "vzu", true, "Venice Zookeeper url, eg. localhost:2622"),
   CLUSTER("cluster", "c", true, "Name of Venice cluster"),
+  TOKEN("token", "t", true, "Authentication Token for JWT authentication"),
```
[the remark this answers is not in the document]
eolivelli: IMHO for users who use always a "token" it is more user-friendly "--token" (json-web- is redundant).

clients/venice-thin-client/src/main/java/com/linkedin/venice/client/store/transport/D2TransportClient.java
```diff
@@ -250,11 +258,26 @@ private String getD2RequestUrl(String requestPath) {
...
   }
 
+  private Map<String, String> injectSecurityHeaders(Map<String, String> headers) {
+    if (authenticationProvider != null) {
+      if (headers == null) {
+        return authenticationProvider.getHTTPAuthenticationHeaders();
+      } else {
+        Map<String, String> copy = new HashMap<>(headers);
+        copy.putAll(authenticationProvider.getHTTPAuthenticationHeaders());
+        return copy;
```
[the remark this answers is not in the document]
eolivelli: In any case we end up in doing a copy.
See createD2GetRequest, it needs a whole Map, so at some point we have to do the copy.
`return new RestRequestBuilder(requestUri).setMethod(HTTP_GET).setHeaders(headers).build();`

we could make a loop over the two maps (headers and getHTTPAuthenticationHeaders) in createD2GetRequest and us RestRequestBuilder.addHeaderValue

WDYT ?

internal/venice-client-common/src/main/java/com/linkedin/venice/CommonConfigKeys.java
```diff
@@ -38,4 +38,9 @@ public class CommonConfigKeys {
    *    (Container 25.2.15)
    */
   public static final String SSL_FACTORY_CLASS_NAME = "ssl.factory.class.name";
+
+  /**
+   * This configuration is the JWT token to use to connect to the backend.
+   */
+  public static final String AUTHENTICATION_TOKEN = "authentication.token";
```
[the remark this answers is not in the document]
eolivelli: let's talk about this at the meeting :-)

internal/venice-client-common/src/main/java/com/linkedin/venice/authentication/ClientAuthenticationProviderFactory.java
```diff
@@ -0,0 +1,27 @@
...
+public final class ClientAuthenticationProviderFactory {
+  private ClientAuthenticationProviderFactory() {
+
+  }
+
+  public static ClientAuthenticationProvider build(VeniceProperties properties) {
+    return build(properties.toProperties());
+  }
+
+  public static ClientAuthenticationProvider build(Map properties) {
```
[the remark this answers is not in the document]
eolivelli: this method exists because in some points we pass something that implements "Map", like org.apache.samza...Config   (implenents Map<String, String)
like here
https://github.com/linkedin/venice/blob/f654de347c97e7808ff8b96037d4a75abdd9eb95/clients/venice-samza/src/main/java/com/linkedin/venice/samza/VeniceSystemFactory.java#L378
and sometimes we pass a Properties object (that is Map<Object, Object>)

We can keep Properties and adapt the code that uses Samza Config, or use here Map<Object, Object>.

I would use Map<String, Object>, but we will have to add some casts/conversions

internal/venice-common/src/main/java/com/linkedin/venice/authentication/AuthenticationService.java
eolivelli: This is here:
- to show that it can access the X509Certificate
- to show a path to implement an AuthenticationService based on X509Certificate (that is the current behaviour)

internal/venice-common/src/main/java/com/linkedin/venice/authentication/jwt/TokenAuthenticationService.java
eolivelli: I didn't want to pollute the main "ConfigKeys",
ideally AuthenticationService plugins are not hard coded into Venice, so I am not sure that it is good to have so many specific parameters in that class.

Maybe we can create a ConfigKeys class here in this package

clients/venice-admin-tool/src/main/java/com/linkedin/venice/Arg.java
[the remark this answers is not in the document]
eolivelli: in the community meeting we decided to not expose this parameter but read the credentials from the ssl config file https://github.com/linkedin/venice/blob/4738167344a7884655fea6263ea8d745a267eeec/clients/venice-admin-tool/src/main/java/com/linkedin/venice/Arg.java#L149C20-L149C35 

in the future we can rename the option to "auth-file" or something like that

clients/venice-thin-client/src/main/java/com/linkedin/venice/client/store/transport/D2TransportClient.java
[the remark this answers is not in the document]
eolivelli: at the meeting: we can do it later

internal/venice-client-common/src/main/java/com/linkedin/venice/authentication/ClientAuthenticationProviderFactory.java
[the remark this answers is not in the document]
eolivelli: let's keep Properties

internal/venice-common/src/main/java/com/linkedin/venice/authentication/AuthenticationService.java
eolivelli: let's keep it

internal/venice-common/src/main/java/com/linkedin/venice/authentication/jwt/TokenAuthenticationService.java
eolivelli: let's create constants

clients/venice-admin-tool/src/main/java/com/linkedin/venice/Arg.java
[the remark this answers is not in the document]
eolivelli: @FelixGV I have changed the code in order to read the Authentication configuration from the SSL config file.

it actually looks better than before, because this way it will be easier to pass any configurations to the ClientAuthenticationProviderFactory and we are not sticking the CLI tool to a specific AuthenticationProvider

clients/venice-admin-tool/src/main/java/com/linkedin/venice/AdminTool.java
```diff
@@ -588,7 +588,7 @@ static CommandLine getCommandLine(String[] args) throws ParseException, IOExcept
        * Don't throw exception yet until all controllers are deployed with SSL support and the script
        * that automatically generates SSL config file is provided.
        */
-      System.out.println("[WARN] Running admin tool without SSL.");
+      System.out.println("[WARN] Running admin tool without any SSL or authentication configuration.");
```
FelixGV, line 591: Nitpick: since it's a negation, I think we should change `s/or/nor/`

internal/venice-common/src/main/java/com/linkedin/venice/authentication/jwt/ConfigKeys.java
```diff
@@ -0,0 +1,12 @@
+package com.linkedin.venice.authentication.jwt;
+
+public final class ConfigKeys {
```
FelixGV, line 3: Nitpick: Can we call this `JwtConfigKeys` or similar? Otherwise, auto-completion will bring up many `ConfigKeys` which is a bit annoying.

FelixGV (commented):
Hi Enrico, LGTM overall, left just a few minor nitpicks. The main issue though is... can we increase the unit test coverage?

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/DispatchingAvroGenericStoreClient.java
```diff
@@ -201,7 +204,13 @@ protected CompletableFuture<V> get(GetRequestContext requestContext, K key) thro
       requestContext.routeRequestMap.put(route, routeRequestFuture);
       try {
         String url = route + uri;
-        CompletableFuture<TransportClientResponse> transportFuture = transportClient.get(url);
+        Map<String, String> headers;
+        if (authenticationProvider != null) {
```
ZacAttack, line 208: Does this change at all if the transport provider is gRPC?  Maybe something we can do as a TODO later since we're not using gRPC widely, but it makes me wonder if we need to adjust this abstraction to be more wrapped with the transport provider?

ZacAttack: Nevermind, the transportclient handles this.

ZacAttack: ehh.  Almost.  Looks like the grpcTransportClient we implemented has headers in the interface but doesn't do anything with them yet :/

So, bit of a warning/TODO that this might now work yet with gRPC

adamxchen:
Hey @eolivelli , thanks a lot for those changes! Do you have any bandwidth to iterate the PR? We'd like to keep the open PR list short so folks can focus on high-priority ones and get changes into production ASAP. If there's no traction or interest to get this PR merged, we will close this PR later.

eolivelli:
I am really sorry, but I don't have cycles.

Thanks

PR #473, opened 2023-06-06
[router][controller][server] Cleanups to partition/replica state

FelixGV:
Various parts of the code dealt with status in the form of strings, and those strings corresponded to values of the ExecutionStatus or HelixState enums. Worse, in some cases, a string could represent either of these enums. This commit fixes this by using strong types instead.

- Partition is the most significantly refactored class, leading to the elimination of the string-keyed map of states to host list. The old constructor that takes in a string-keyed map is still there but only used by tests and marked deprecated. All main code paths now construct instances via a new more efficient constructor that takes in EnumMaps directly.

- ReplicaState now carries just an ExecutionStatus rather than both states, each in a separate string. Also removed the isReadyToServe field, since it could be inferred from the venicePushStatus instead.

- PartitionStatus removes some redundant map lookups.

- ReplicaStatus now avoids processing statusHistory in code paths that don't need it.

- HelixExternalViewRepository::getReplicaStates now throws an exception. It was never used in the previous code, and carried lingering uses of code paths that are now cleaned up.

- HelixCustomizedViewOfflinePushRepository::onCustomizedViewDataChange has a few efficiency improvements related to Partition states and also reduces logging volume in cases instances were found in the CV, but missing from Live Instances. This is now going to be printed once per ZK watch invocation, with all relevant hostnames included, rather than once per missing replica.

- Deleted RoutingDataRepository::refreshRoutingDataForResource and all implementations.

## How was this PR tested?
GHCI


internal/venice-common/src/test/java/com/linkedin/venice/pushmonitor/PartitionStatusTest.java
```diff
@@ -1,5 +1,7 @@
 package com.linkedin.venice.pushmonitor;
 
+import static org.testng.Assert.*;
```
sixpluszero, line 3: nit: Let's not use * import here.

internal/venice-common/src/main/java/com/linkedin/venice/meta/Partition.java
```diff
@@ -26,19 +33,33 @@
 public class Partition {
   private static final Logger LOGGER = LogManager.getLogger(Partition.class);
   /**
-   * ID of partition. One of the number between [0 ~ total number of partition)
+   * ID of partition. One of the number between [0, total number of partition[
```
sixpluszero, line 36: maybe ] instead of [

internal/venice-common/src/main/java/com/linkedin/venice/meta/Partition.java
```diff
@@ -102,13 +128,23 @@ public List<Instance> getInstancesInState(String state) {
   public List<Instance> getReadyToServeInstances() {
```
sixpluszero, line 128: Unrelated to your change but ... the JavaDoc says remove this API once we are fully on L/F.. do you think we could remove it or it is the doc has problem?

internal/venice-common/src/main/java/com/linkedin/venice/meta/Partition.java
```diff
@@ -102,13 +128,23 @@ public List<Instance> getInstancesInState(String state) {
...
     List<Instance> completed = this.executionStatusToInstancesMap.get(ExecutionStatus.COMPLETED);
     List<Instance> online = this.helixStateToInstancesMap.get(HelixState.ONLINE);
-    return completed.size() > online.size() ? completed : this.readyInstances;
+    return completed.size() > online.size() ? completed : getReadyInstances();
   }
 
   public List<Instance> getWorkingInstances() {
```
sixpluszero, line 134: Also, just to be cautious as I am confused by the existing variable/API naming here. Is working instance containing the ready-to-serve and bootstrapping ones?

internal/venice-common/src/main/java/com/linkedin/venice/meta/Partition.java
```diff
@@ -131,52 +167,96 @@ public Instance getLeaderInstance() {
...
+    for (Map.Entry<HelixState, List<Instance>> entry: this.helixStateToInstancesMap.entrySet()) {
       List<Instance> instances = entry.getValue();
       for (Instance instance: instances) {
         if (instance.getNodeId().equals(instanceId)) {
-          return status;
+          return entry.getKey();
+        }
+      }
+    }
+    return null;
```
sixpluszero, line 220: Should we return HelixState.UNKNOWN instead?

sixpluszero: Or here Unknown has different meaning than not found instance id...

internal/venice-common/src/main/java/com/linkedin/venice/pushmonitor/PushStatusDecider.java
```diff
@@ -251,42 +252,29 @@ private void logConditionally(String newReason, Optional<String> previousReason,
...
     // partitions can be completed. Vice versa, partitions will be in error state if leader is in error
     // state.
     boolean isLeaderCompleted = true;
     int previouslyDisabledErrorReplica = 0;
-    for (Map.Entry<Instance, String> entry: instanceToStateMap.entrySet()) {
+    for (Map.Entry<Instance, HelixState> entry: instanceToStateMap.entrySet()) {
       ExecutionStatus currentStatus =
           getReplicaCurrentStatus(partitionStatus.getReplicaHistoricStatusList(entry.getKey().getNodeId()));
-      if (entry.getValue().equals(HelixState.LEADER_STATE)) {
+      if (entry.getValue() == HelixState.LEADER) {
```
sixpluszero, line 269: Hmm, I thought .equals() should be the better way?

internal/venice-common/src/main/java/com/linkedin/venice/helix/HelixExternalViewRepository.java
```diff
@@ -47,42 +49,7 @@ public HelixExternalViewRepository(SafeHelixManager manager) {
   }
 
   public List<ReplicaState> getReplicaStates(String kafkaTopic, int partitionId) {
-    Partition partition = resourceAssignment.getPartition(kafkaTopic, partitionId);
```
sixpluszero, line 50: not in the scope of this PR: This change looks good for this cleanup PR but I'd suggest we move the API to CV based repo only and force MetaDataHandler to instantiate HelixCVRepo instead general repo.

sixpluszero (commented):
Thanks for the cleanup. I took a pass and left some comments. Also, it looks like HelixState.java has some xxxx_STATE which can all be replaced by enum.name()?

internal/venice-common/src/main/java/com/linkedin/venice/meta/Partition.java
sushantmane, line 36: IMO original syntax was correct. It is half-closed interval.

`[0, total number of partition) --> [a, b) ==> {x belons to N | a <= x && x < b}`

internal/venice-common/src/main/java/com/linkedin/venice/pushmonitor/PushStatusDecider.java
```diff
@@ -63,14 +64,16 @@ public Pair<ExecutionStatus, Optional<String>> checkPushStatusAndDetails(
...
       int completedReplicasCount = partition.getReadyToServeInstances().size();
-      int assignedReplicasCount = partition.getAllInstances().size();
-      logger.debug(
-          "Checking Push status for offline push for topic: {} Partition: {} has {} assigned replicas including {} error replicas, {} completed replicas.",
-          pushStatus.getKafkaTopic(),
-          partition.getId(),
-          assignedReplicasCount,
-          errorReplicasCount,
-          completedReplicasCount);
+      if (logger.isDebugEnabled()) {
```
sushantmane, line 67: Is this check necessary if we use lambda for ` partition.getNumOfTotalInstances()`?

sushantmane (commented):
Thanks, @FelixGV! I left a few comments, please take a look.

internal/venice-common/src/main/java/com/linkedin/venice/pushmonitor/PushStatusDecider.java
majisourav99: enum `==` projects against NPE.

FelixGV: Enum values are singletons, so we can do `==` on them, just like with primitive.

I prefer doing `==` whenever possible since it is more concise (and for me, easier to read), it gets compile-time type safety (whereas you could call `Enum1.BLAH.equals(Enum2.BLEH)` and technically it compiles), and you can't get NPEs on `==`.

internal/venice-common/src/main/java/com/linkedin/venice/helix/HelixExternalViewRepository.java
FelixGV: SG, done.

internal/venice-common/src/main/java/com/linkedin/venice/meta/Partition.java
FelixGV: Yeah... although I think that's a hangover from the O/B/O days, since we don't even have `BOOTSTRAP` anymore... so most likely, `getWorkingInstances()` could be substituted for `getReadyToServeInstances()` and it would work just as well...

internal/venice-common/src/main/java/com/linkedin/venice/meta/Partition.java
sixpluszero: Hi Felix, closing this issue as Sushant also posted the same one. I might agree what he said - original syntax might be correct?

FelixGV: Seems like both syntaxes are correct... I didn't know about the parenthesis syntax for exclusion!

https://en.wikipedia.org/wiki/Interval_(mathematics)#Including_or_excluding_endpoints

internal/venice-common/src/main/java/com/linkedin/venice/pushmonitor/PushStatusDecider.java
FelixGV: Lambda could be an alternative way to do it, but it would instantiate a lambda object, which isn't necessary IMHO...

internal/venice-common/src/main/java/com/linkedin/venice/helix/HelixExternalViewRepository.java
```diff
@@ -47,42 +49,7 @@ public HelixExternalViewRepository(SafeHelixManager manager) {
   }
 
   public List<ReplicaState> getReplicaStates(String kafkaTopic, int partitionId) {
```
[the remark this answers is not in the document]
FelixGV: Instead I followed Jialin's advice to remove this from the interface and leave it only inside `HelixCustomizedViewOfflinePushRepository`.

internal/venice-common/src/main/java/com/linkedin/venice/meta/Partition.java
```diff
@@ -26,19 +33,33 @@
...
+    this.helixStateToInstancesMap = helixStateToInstancesMap;
+    this.executionStatusToInstancesMap = executionStatusToInstancesMap;
+    populateEmptyKeysOfEnumMaps();
+  }
 
+  /** Legacy constructor, used only by tests... TODO: Move everything over to the new constructor */
+  @Deprecated
   public Partition(int id, Map<String, List<Instance>> stateToInstancesMap) {
     this.id = id;
-    this.stateToInstancesMap = stateToInstancesMap;
```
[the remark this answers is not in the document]
FelixGV: I completely eliminated the deprecated constructor in my latest diff.

internal/venice-common/src/main/java/com/linkedin/venice/meta/Partition.java
FelixGV: I ended up cleaning this up. Deeeeeep rabbit hole. But I think with this we'll be completely rid of O/B/O. I even deleted `ONLINE` and `BOOTSTRAP` from the `HelixState` enum!

internal/venice-common/src/main/java/com/linkedin/venice/meta/Partition.java
FelixGV: I cleaned this up.

FelixGV:
> Thanks for the cleanup. I took a pass and left some comments. Also, it looks like HelixState.java has some xxxx_STATE which can all be replaced by enum.name()?

This is not possible... the annotations require a fully static constant, which apparently does not allow calling this function...

internal/venice-test-common/src/integrationtest/java/com/linkedin/venice/controller/TestHAASController.java
```diff
@@ -87,7 +87,7 @@ public void testStartHAASHelixControllerAsControllerClusterLeader() {
     }
   }
 
-  @Test(timeOut = 60 * Time.MS_PER_SECOND)
+  @Test(timeOut = 120 * Time.MS_PER_SECOND)
```
sixpluszero, line 90: Is this for mitigating flaky test?

internal/venice-test-common/src/integrationtest/java/com/linkedin/venice/controller/TestHAASController.java
```diff
@@ -125,6 +125,7 @@ public void testTransitionToHAASControllerAsControllerClusterLeader() {
               () -> assertEquals(
                   assertCommand(controllerClient.queryJobStatus(topicName)).getStatus(),
                   ExecutionStatus.COMPLETED.toString())));
+      System.out.println("End of the test");
```
sixpluszero, line 128: Same here, but since this it test, probably fine to keep it (probably for the sake of debugging flakiness..)

sixpluszero (approved):
LGTM from my side! Thanks for cleaning up legacy implementations.

internal/venice-test-common/src/integrationtest/java/com/linkedin/venice/controller/TestHAASController.java
FelixGV: Yeah, I was getting flakiness on these two tests on my local, and increasing the timeout seemed to help.

internal/venice-test-common/src/integrationtest/java/com/linkedin/venice/controller/TestHAASController.java
FelixGV: Thanks for catching it... I thought I had removed it, but forgot. Will clear it in the next commit.

FelixGV:
I have rebased and pushed one more commit, which removes the unnecessary log @sixpluszero pointed out, as well as some dead code I found. I also rewrote the commit message to include all details from the various iterations. Thanks for the reviews!

sixpluszero (approved):
LGTM thanks!

FelixGV:
> LGTM thanks!

Thanks for the review!

PR #475, opened 2023-06-06
[Server] [FC] Add gRPC support for Get & BatchGet Requests to VeniceServer and Fast Client

rkunds:
## Add gRPC support for Get & BatchGet Requests to VeniceServer and Fast Client

We introduce gRPC functionality for the read path between Venice Server and Fast Client. 

- Add Guava Dependency (temporarily) in order to be able to generate gRPC Java code stubs without package not found errors. 

- Update Proto schema to accommodate requests between client and servers
- Created gRPC transport client as a drop in replacement for current R2TransportClient
- Able to create and start a gRPC server instance alongside a specific Venice Server
- gRPC enabled Fast Client can perform Get and BatchGet Requests
- Create Integration test to test gRPC Fast Client against Thin Client and standard Fast Client
- Create initial unit testing for new functionality

## How was this PR tested?

CI, Unit Testing and Integration Testing for new features

Any gRPC related functionality such as enabling it on the server side will have to be explicitly enabled. The default setting is to disable gRPC for both Venice Server and Fast Client, so no behavior change other than additional configuration parameters have been created.


services/venice-server/src/main/java/com/linkedin/venice/listener/grpc/interceptors/AuthorizationInterceptor.java
```diff
@@ -0,0 +1,36 @@
...
+  @Override
+  public <ReqT, RespT> Listener<ReqT> interceptCall(
+      final ServerCall<ReqT, RespT> serverCall,
+      final Metadata metadata,
+      final ServerCallHandler<ReqT, RespT> serverCallHandler) {
+    final String auth_token = metadata.get(Key.of("auth_token", Metadata.ASCII_STRING_MARSHALLER)); // encodes the
+                                                                                                    // string via the
+                                                                                                    // marshaller
+
+    if (auth_token == null || !auth_token.equals("valid_token")) {
```
sushantmane, line 23: Use camel case for variable names, e.g., `authToken` instead of `auth_token`

services/venice-server/src/main/java/com/linkedin/venice/listener/grpc/VeniceReadServiceClient.java
```diff
@@ -0,0 +1,34 @@
...
+    originChannel = ManagedChannelBuilder.forTarget(ipString).usePlaintext().build();
+    ClientInterceptor interceptor = new AuthTokenProvideInterceptor();
+    Channel newChannel = ClientInterceptors.intercept(originChannel, interceptor);
+    stub = VeniceReadServiceGrpc.newBlockingStub(newChannel);
+  }
+
+  public String get(String key) {
+    VeniceClientRequest request = VeniceClientRequest.newBuilder().setKey(key).build();
+    // handle failed preconditions
+    VeniceServerResponse response = stub.get(request);
```
sushantmane, line 27: Could you add error handling for `get()`? 

For example, 

```java
try {
  response = stub.get(request);
} catch(StatusRuntimeException e) {
  LOGGER.error("Fail to get value for key: {}", key, e) 
  throw new VeniceException(e);
} catch(OtherException) {...}
```

internal/venice-test-common/src/integrationtest/java/com/linkedin/venice/server/VeniceServerGrpcTest.java
```diff
@@ -0,0 +1,87 @@
...
+  private void createServer(int port) {
+    this.server = new VeniceReadServiceServer(port);
+    this.server.start();
+  }
+
+  private void createClient(String addr, int port) {
+    this.client = new VeniceReadServiceClient(addr, port);
+  }
+
+  private void stopServer() {
```
sushantmane, line 22: IMO `stopServer()` and `shutdownClient()` methods here are not needed. We can directly call stop and shutdown.

internal/venice-test-common/src/integrationtest/java/com/linkedin/venice/server/VeniceServerGrpcTest.java
```diff
@@ -0,0 +1,87 @@
...
+    this.server.stop();
```
sushantmane, line 23: `this` is optional. I usually avoid using it unless it is required...

internal/venice-test-common/src/integrationtest/java/com/linkedin/venice/server/VeniceServerGrpcTest.java
```diff
@@ -0,0 +1,87 @@
...
+  @Test
+  public void testValidToken() {
+    createServer(50051);
+
+    createClient("localhost", 50051);
+    String value = client.get("valid");
+    Assert.assertEquals(value, "validvalid"); // get function simply returns the key twice, there is no accessing data
+                                              // yet.
+
+    stopServer();
```
sushantmane, line 39: Can we move `stopServer()` and `shutdownClient()` to something like
```java
  @AfterMethod
  public void tearDown() {
    server.stop();
    client.shutdown();
  }
```

internal/venice-test-common/src/integrationtest/java/com/linkedin/venice/server/VeniceServerGrpcTest.java
```diff
@@ -0,0 +1,87 @@
...
+    shutdownClient();
+  }
+
+  @Test
+  public void testServerNotStarted() {
```
sushantmane, line 44: Consider using `@Test(expectedExceptions = ..) ` instead of `Assert.fail()`

Your test will look something like this:
```java
@Test(expectedExceptions = StatusRuntimeException.class, expectedExceptionsMessageRegExp = "UNAVAILABLE: io exception")
  public void testServerNotStarted() {
    client.get("valid");
  }
  ```

File: internal/venice-test-common/src/integrationtest/java/com/linkedin/venice/server/VeniceServerGrpcTest.java [hunk 7]
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  @Test
+  public void testInvalidToken() {
+    createServer(50051);
+    createClient("localhost", 50051);
+
+    try {
+      client.get("invalid");
+      Assert.fail("Expected StatusRuntimeException");
```
sushantmane, line 63: Same as the above: `expectedExceptions` is preferred method

services/venice-server/src/main/java/com/linkedin/venice/grpc/VeniceGrpcServer.java
```diff
@@ -0,0 +1,35 @@
...
+    server = Grpc.newServerBuilderForPort(config.getPort(), config.getCredentials())
+        .addService(ServerInterceptors.intercept(config.getService(), config.getInterceptors()))
+        .build();
+  }
+
+  public void start() {
+    try {
+      server.start();
+    } catch (IllegalStateException | IOException illegalStateException) {
+      LOGGER.error("Failed to start {}", VeniceGrpcServer.class.getSimpleName(), illegalStateException);
```
sushantmane, line 25: Let's log the port and service name for which we are building the server 
```java
LOGGER.error("Failed to start gRPC server for service:{} on port:{}", sericeName, port, e);
```

services/venice-server/src/main/java/com/linkedin/venice/grpc/VeniceGrpcServerConfig.java
```diff
@@ -0,0 +1,101 @@
...
+import java.util.List;
+
+
+public class VeniceGrpcServerConfig {
+  private final int port;
+  private final ServerCredentials credentials;
+  private final BindableService service;
+  private final List<? extends ServerInterceptor> interceptors;
+
+  VeniceGrpcServerConfig(Builder builder) {
```
sushantmane, line 16: Change `VeniceGrpcServerConfig` constructor to `private`

services/venice-server/src/main/java/com/linkedin/venice/grpc/VeniceGrpcServerConfig.java
```diff
@@ -0,0 +1,101 @@
...
+    return service;
+  }
+
+  public List<? extends ServerInterceptor> getInterceptors() {
+    return interceptors;
+  }
+
+  @Override
+  public String toString() {
+    return "";
```
sushantmane, line 41: would be good to add the service name and port instead of having an empty string

services/venice-server/src/main/java/com/linkedin/venice/grpc/VeniceGrpcServerConfig.java
```diff
@@ -0,0 +1,101 @@
...
+
+  public static class Builder {
+    private Integer port;
+    private ServerCredentials credentials;
+    private BindableService service;
+    private List<? extends ServerInterceptor> interceptors;
+
+    public Builder() {
+      port = 0;
+      credentials = null;
```
sushantmane, line 52: The constructor for the `Builder` can be removed. It's unnecessary here

services/venice-server/src/main/java/com/linkedin/venice/grpc/VeniceGrpcServerConfig.java
```diff
@@ -0,0 +1,101 @@
...
+      verifyAndAddDefaults();
+      return new VeniceGrpcServerConfig(this);
+    }
+
+    public void verifyAndAddDefaults() {
+      if (port == null) {
+        throw new IllegalArgumentException("Port must be set");
+      }
+      if (credentials == null) {
+        throw new IllegalArgumentException("Credentials must be set");
```
sushantmane, line 92: let's add `InsecureServerCredentials.create()` as default value for `credentials`

services/venice-server/src/main/java/com/linkedin/venice/listener/grpc/VeniceReadServiceClient.java
```diff
@@ -0,0 +1,34 @@
+package com.linkedin.venice.listener.grpc;
+
+import com.linkedin.venice.listener.grpc.interceptors.AuthTokenProvideInterceptor;
+import com.linkedin.venice.protocols.VeniceClientRequest;
+import com.linkedin.venice.protocols.VeniceReadServiceGrpc;
+import com.linkedin.venice.protocols.VeniceServerResponse;
+import io.grpc.*;
```
sushantmane, line 7: Replace with the single class imports

services/venice-server/src/main/java/com/linkedin/venice/listener/grpc/VeniceReadServiceClient.java
```diff
@@ -0,0 +1,34 @@
...
+
+
+public class VeniceReadServiceClient {
+  private final ManagedChannel originChannel;
+
+  private final VeniceReadServiceGrpc.VeniceReadServiceBlockingStub stub;
+
+  public VeniceReadServiceClient(String addr, int port) {
```
sushantmane, line 15: We can just pass one argument `serverAddress` to `VeniceReadServiceClient`

services/venice-server/src/main/java/com/linkedin/venice/listener/grpc/VeniceReadServiceServer.java
```diff
@@ -0,0 +1,33 @@
...
+
+import com.linkedin.venice.listener.grpc.interceptors.AuthorizationInterceptor;
+import io.grpc.Grpc;
+import io.grpc.InsecureServerCredentials;
+import io.grpc.Server;
+import io.grpc.ServerInterceptors;
+import java.io.IOException;
+
+
+public class VeniceReadServiceServer {
```
sushantmane, line 11: You can either extend `VeniceGrpcServer` or have it as one of the fields in this class  to avoid code duplication

services/venice-server/src/main/java/com/linkedin/venice/listener/grpc/interceptors/AuthTokenProvideInterceptor.java
```diff
@@ -0,0 +1,45 @@
...
+    return new ForwardingClientCall.SimpleForwardingClientCall<ReqT, RespT>(
+        channel.newCall(methodDescriptor, callOptions)) {
+      String auth_token;
+      private Metadata headers;
+
+      @Override
+      public void sendMessage(ReqT message) {
+        String key = ((VeniceClientRequest) message).getKey();
+        if (key.equals("invalid")) {
+          auth_token = "invalid";
```
sushantmane, line 29: Use camel case for variable names, e.g.,` authToken` instead of `auth_token`

services/venice-server/src/main/java/com/linkedin/venice/listener/grpc/interceptors/AuthorizationInterceptor.java
```diff
@@ -0,0 +1,36 @@
...
+import io.grpc.StatusRuntimeException;
+
+
+public class AuthorizationInterceptor implements ServerInterceptor {
+  @Override
+  public <ReqT, RespT> Listener<ReqT> interceptCall(
+      final ServerCall<ReqT, RespT> serverCall,
+      final Metadata metadata,
+      final ServerCallHandler<ReqT, RespT> serverCallHandler) {
+    final String auth_token = metadata.get(Key.of("auth_token", Metadata.ASCII_STRING_MARSHALLER)); // encodes the
```
sushantmane, line 19: Can you move the comment above the statement?

sushantmane (changes requested):

sushantmane (commented):
Can you also add unit tests for newly added code, especially for `VeniceGrpcServer`?

internal/venice-test-common/src/integrationtest/java/com/linkedin/venice/fastclient/grpc/VeniceGrpcEndtoEndTest.java
```diff
@@ -0,0 +1,155 @@
...
+import java.util.concurrent.TimeUnit;
+import org.apache.avro.generic.GenericRecord;
+import org.apache.avro.specific.SpecificRecord;
+import org.testng.Assert;
+import org.testng.annotations.AfterClass;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Test;
+
+
+public class VeniceGrpcEndtoEndTest extends AbstractClientEndToEndSetup {
```
sushantmane, line 42: We don't need to extend `AbstractClientEndToEndSetup` since this test is not anything from it

internal/venice-test-common/src/integrationtest/java/com/linkedin/venice/fastclient/grpc/VeniceGrpcEndtoEndTest.java
```diff
@@ -0,0 +1,155 @@
...
+        controllerClient -> TestUtils.waitForNonDeterministicAssertion(60, TimeUnit.SECONDS, () -> {
+          StoreResponse storeResponse = controllerClient.getStore(uniqueStoreName);
+          Assert.assertEquals(storeResponse.getStore().getCurrentVersion(), 1);
+        }));
+
+    return uniqueStoreName;
+  }
+
+  @Test
+  public void testThinClient() throws Exception {
```
sushantmane, line 101: Let's create one test in which we verify the data first using thin client and then fast client

internal/venice-test-common/src/integrationtest/java/com/linkedin/venice/fastclient/grpc/VeniceGrpcEndtoEndTest.java
```diff
@@ -0,0 +1,155 @@
...
+
+    try {
+      for (Set<String> key: keys) {
+        Map<String, GenericRecord> ret = genericFastClient.batchGet(key).get();
+        for (Map.Entry<String, GenericRecord> entry: ret.entrySet()) {
+          Assert.assertEquals(entry.toString(), entry.getKey() + "=test_name_" + entry.getKey());
+        }
+      }
+    } catch (ExecutionException | InterruptedException e) {
+      throw new RuntimeException(e);
```
sushantmane, line 152: Can we use throws in method signature

clients/da-vinci-client/src/main/java/com/linkedin/davinci/config/VeniceServerConfig.java
```diff
@@ -405,6 +409,8 @@ public VeniceServerConfig(VeniceProperties serverProperties, Map<String, Map<Str
       throws ConfigurationException {
     super(serverProperties, kafkaClusterMap);
     listenerPort = serverProperties.getInt(LISTENER_PORT, 0);
+    grpcPort = serverProperties.getInt(GRPC_READ_SERVER_PORT, 0);
+    isGrpcEnabled = serverProperties.getBoolean(ENABLE_GRPC_READ_SERVER, true);
```
sushantmane, line 413: IMO default should be false
` serverProperties.getBoolean(ENABLE_GRPC_READ_SERVER, false);`

clients/da-vinci-client/src/main/java/com/linkedin/davinci/config/VeniceServerConfig.java
```diff
@@ -709,6 +715,14 @@ public int getListenerPort() {
     return listenerPort;
   }
 
+  public int getGrpcPort() {
+    return grpcPort;
+  }
+
+  public boolean getIsGrpcEnabled() {
```
sushantmane, line 722: rename `getIsGrpcEnabled` to `isGrpcEnabled`

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/integration/utils/VeniceServerWrapper.java
```diff
@@ -199,6 +201,9 @@ static StatefulServiceProvider<VeniceServerWrapper> generateService(
 
       // Generate server.properties in config directory
       int listenPort = TestUtils.getFreePort();
+      boolean isGrpcEnabled = Boolean.parseBoolean(configProperties.getProperty(ENABLE_GRPC_READ_SERVER, "false"));
+      System.out.println(isGrpcEnabled);
```
sushantmane, line 205: delete sout before merging PR

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/integration/utils/VeniceServerWrapper.java
```diff
@@ -227,6 +232,11 @@ static StatefulServiceProvider<VeniceServerWrapper> generateService(
         serverPropsBuilder.put(KafkaTestUtils.getLocalCommonKafkaSSLConfig(SslUtils.getTlsConfiguration()));
       }
 
+      if (isGrpcEnabled) {
```
sushantmane, line 235: What if in the future we want to turn on gRPC by default and in the tests we want to turn it off?

IMO this block should be rewritten as
```java
      serverPropsBuilder.put(ENABLE_GRPC_READ_SERVER, isGrpcEnabled);
      if (isGrpcEnabled) {
        serverPropsBuilder.put(GRPC_READ_SERVER_PORT, TestUtils.getFreePort());
      }
```

internal/venice-test-common/src/integrationtest/java/com/linkedin/venice/fastclient/grpc/VeniceGrpcEndtoEndTest.java
```diff
@@ -0,0 +1,157 @@
...
+    // 4. Create thin client
+    AvroGenericStoreClient<Object, Object> avroClient = ClientFactory.getAndStartGenericAvroClient(
+        ClientConfig.defaultGenericClientConfig(storeName).setVeniceURL(cluster.getRandomRouterURL()));
+
+    // 4. Create fastClient
+    r2Client = ClientTestUtils.getR2Client(ClientTestUtils.FastClientHTTPVariant.HTTP_2_BASED_R2_CLIENT);
+    d2Client = D2TestUtils.getAndStartHttpsD2Client(cluster.getZk().getAddress());
+
+    com.linkedin.venice.fastclient.ClientConfig.ClientConfigBuilder<Object, Object, SpecificRecord> clientConfigBuilder =
+        new com.linkedin.venice.fastclient.ClientConfig.ClientConfigBuilder<>().setStoreName(storeName)
```
sushantmane, line 119: Remove dependencies on`AbstractClientE2ESetup class

we can rewrite this block something like this:
```java


    ClientConfigBuilder<Object, Object, SpecificRecord> clientConfigBuilder =
        new ClientConfigBuilder<>().setStoreName(storeName)
            .setStoreMetadataFetchMode(StoreMetadataFetchMode.SERVER_BASED_METADATA)
            .setR2Client(ClientTestUtils.getR2Client(ClientTestUtils.FastClientHTTPVariant.HTTP_2_BASED_R2_CLIENT))
            .setD2Client(D2TestUtils.getAndStartHttpsD2Client(cluster.getZk().getAddress()))
            .setClusterDiscoveryD2Service(VeniceRouterWrapper.CLUSTER_DISCOVERY_D2_SERVICE_NAME)
            .setMaxAllowedKeyCntInBatchGetReq(150)
            .setMetadataRefreshIntervalInSeconds(5)
            .setMetricsRepository(new MetricsRepository())
            .setUseStreamingBatchGetAsDefault(true)
            .setSpeculativeQueryEnabled(false);

    AvroGenericStoreClient<String, GenericRecord> genericFastClient =
        ClientFactory.getAndStartGenericStoreClient(clientConfigBuilder.build());
        
        
```

internal/venice-test-common/src/integrationtest/java/com/linkedin/venice/fastclient/grpc/VeniceGrpcEndtoEndTest.java
```diff
@@ -0,0 +1,157 @@
...
+    // generate list of keys
+    List<Set<String>> keySets = new ArrayList<>();
+    for (int i = 1; i <= 100; i++) {
+      Set<String> key = new HashSet<>();
+      key.add(Integer.toString(i));
+      keySets.add(key);
+    }
+
+    // iterate through list and obtain records through both clients
+    for (Set<String> key: keySets) {
```
sushantmane, line 145: lets rename `key` to either `keySet` of `keys`

services/venice-server/src/main/java/com/linkedin/venice/grpc/VeniceGrpcServer.java
```diff
@@ -0,0 +1,39 @@
...
+        .addService(ServerInterceptors.intercept(config.getService(), config.getInterceptors()))
+        .build();
+  }
+
+  public void start() {
+    try {
+      server.start();
+    } catch (IllegalStateException | IOException illegalStateException) {
+      LOGGER.error(
+          "Failed to start {} on port {}",
```
sushantmane, line 26: Add service name in log statement and there is no need to include `VeniceGrpcServer.class.getSimpleName(),`

`"Failed to start gRPC server for service: {} on port:{}"...`

services/venice-server/src/main/java/com/linkedin/venice/listener/ListenerService.java
```diff
@@ -150,6 +158,18 @@ public boolean startInner() throws Exception {
     serverFuture = bootstrap.bind(port).sync();
     LOGGER.info("Listener service started on port: {}", port);
 
+    grpcServer = new VeniceGrpcServer(
```
sushantmane, line 161: IMO we should not create `VeniceGrpcServer ` object if `isGrpcEnabled` is set to false.
```java
if (isGrpcEnabled) {
    grpcServer = new VeniceGrpcServer( new VeniceGrpcServerConfig.Builder().setPort(grpcPort).setService(new VeniceReadServiceImpl()).build());)
}
```
Move `grpcServer` object creation on line number 154.

In `startInner` method you should only keep start logic and no instantiation 
```java
if (grpcServer != null) {
      grpcServer.start();
      LOGGER.info("gRPC service started on port: {}", grpcPort);
}
```

services/venice-server/src/main/java/com/linkedin/venice/listener/ListenerService.java
```diff
@@ -169,6 +189,11 @@ public void stopInner() throws Exception {
     workerGroup.shutdownGracefully();
     bossGroup.shutdownGracefully();
     shutdown.sync();
+
+    if (grpcServer != null) {
+      System.out.println("Stopping gRPC service on port: " + grpcPort);
```
sushantmane, line 194: Use log statement instead of SOUT

sushantmane (changes requested):

tlamlert:
This is a very good PR. Great job!

services/venice-server/src/main/java/com/linkedin/venice/listener/ListenerService.java
```diff
@@ -143,13 +155,25 @@ public ListenerService(
...
+              .build());
+    }
   }
 
   @Override
   public boolean startInner() throws Exception {
     serverFuture = bootstrap.bind(port).sync();
     LOGGER.info("Listener service started on port: {}", port);
 
+    if (isGrpcEnabled) {
```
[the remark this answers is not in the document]
rkunds: `grpcServer` is always going to be null at this point as we create the server object after. What is the point in checking this here?

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/transport/GrpcTransportClient.java
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  @Override
+  public void close() throws IOException {
+    for (Map.Entry<String, ManagedChannel> entry: nettyServerToGrpcServerChannelMap.entrySet()) {
+      entry.getValue().shutdown();
+    }
+  }
+
+  public class GrpcTransportClientCallback {
```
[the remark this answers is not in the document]
rkunds: in `R2TransportClient`, we have a similar functionality but in a nested class. Should we follow that convention or still move this to a different class?

services/venice-server/src/main/java/com/linkedin/venice/listener/ListenerService.java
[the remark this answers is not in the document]
sushantmane: yeah that's right. thanks!

services/venice-server/src/test/java/com/linkedin/venice/grpc/VeniceReadServiceImplTest.java
```diff
@@ -0,0 +1,80 @@
...
+    when(mockStorageReadRequestsHandler.handleSingleGetGrpcRequest(any(GetRouterRequest.class)))
+        .thenReturn(storageResponse);
+
+    veniceReadService.get(request, mockResponseObserver);
+
+    verify(mockResponseObserver, times(1)).onNext(any(VeniceServerResponse.class));
+    verify(mockResponseObserver, times(1)).onCompleted();
+
+    // reset mock counter
+    reset(mockResponseObserver);
```
sushantmane, line 59: is reset really necessary

sushantmane (changes requested):
Thanks, @rkunds. Left some comments let me know if you have any questions regarding review comments.

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/transport/GrpcTransportClient.java
```diff
@@ -0,0 +1,191 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+
+public class GrpcTransportClient extends InternalTransportClient {
+  private static final Logger LOGGER = LogManager.getLogger(GrpcTransportClient.class);
+  private final TransportClient r2TransportClient; // used for metadata requests
+
+  // field has public visibility for unit testing
+  public final Map<String, ManagedChannel> serverGrpcChannels;
```
gaojieliu, line 28: For unit test, we normally use package access.
Also I wonder whether we should expose a function to return a `ManagedChannel` for a given server address and keep this variable as private or not.

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/transport/GrpcTransportClient.java
```diff
@@ -0,0 +1,191 @@
...
+  public ManagedChannel getChannel(String requestPath) {
+    String portKey = requestPath.split("/")[2];
+
+    if (!veniceAddressToGrpcAddress.containsKey(portKey)) {
+      throw new VeniceException("No grpc server found for port: " + portKey);
+    }
+
+    return serverGrpcChannels.computeIfAbsent(
+        portKey,
+        k -> ManagedChannelBuilder.forTarget(veniceAddressToGrpcAddress.get(k)).usePlaintext().build());
```
gaojieliu, line 53: So we can't use SSL here?

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/transport/GrpcTransportClient.java
```diff
@@ -0,0 +1,191 @@
...
+  }
+
+  public ManagedChannel getChannel(String requestPath) {
+    String portKey = requestPath.split("/")[2];
+
+    if (!veniceAddressToGrpcAddress.containsKey(portKey)) {
+      throw new VeniceException("No grpc server found for port: " + portKey);
+    }
+
+    return serverGrpcChannels.computeIfAbsent(
```
gaojieliu, line 51: It seems this function can be invoked concurrently, so can we use `VeniceConcurrentHashMap` for `serverGrpcChannels`?

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/transport/GrpcTransportClient.java
```diff
@@ -0,0 +1,191 @@
...
+        portKey,
+        k -> ManagedChannelBuilder.forTarget(veniceAddressToGrpcAddress.get(k)).usePlaintext().build());
+  }
+
+  @Override
+  public CompletableFuture<TransportClientResponse> get(String requestPath, Map<String, String> headers) {
+    String[] requestParts = RequestHelper.getRequestParts(requestPath);
+
+    if (!requestParts[1].equals("storage")) {
```
gaojieliu, line 60: Can we use the predefined String Const instead of plain string here?

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/transport/GrpcTransportClient.java
```diff
@@ -0,0 +1,191 @@
...
+      LOGGER.debug("Requesting metadata from R2 client");
+      return r2TransportClient.get(requestPath, headers);
+    }
+
+    // Lazily create channel for each venice server
+    ManagedChannel channel = getChannel(requestPath);
+
+    VeniceClientRequest request = VeniceClientRequest.newBuilder()
```
gaojieliu, line 68: Is it possible to do a lightweight mutation of the request path, such as replacing the first part by some different sub-path?

Ideally, we want to reuse the same request path..
Another way would be to have a special header in `headers` map to indicate whether this is a metadata request or not, and we can reuse the same request path between Netty request and GPRC request.

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/transport/GrpcTransportClient.java
```diff
@@ -0,0 +1,191 @@
...
+  @Override
+  public CompletableFuture<TransportClientResponse> post(
+      String requestPath,
+      Map<String, String> headers,
+      byte[] requestBody) {
+    String[] requestParts = RequestHelper.getRequestParts(requestPath);
+
+    ManagedChannel channel = getChannel(requestPath);
+    VeniceClientRequest request = VeniceClientRequest.newBuilder()
+        .setStoreName(requestParts[2])
```
gaojieliu, line 90: `StoreName` is typically different from `ResourceName`, so why we initialize them with the same value here?

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/transport/GrpcTransportClient.java
```diff
@@ -0,0 +1,191 @@
...
+        .setResourceName(requestParts[2])
+        .setPartition(-1)
+        .setKeyBytes(ByteString.copyFrom(requestBody))
```
gaojieliu, line 93: Can we just reuse the same `requestBody` instead of making a copy?

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/transport/GrpcTransportClient.java
```diff
@@ -0,0 +1,191 @@
...
+    VeniceReadServiceGrpc.VeniceReadServiceStub clientStub = VeniceReadServiceGrpc.newStub(channel);
+    GrpcTransportClientBatchCallback callback = new GrpcTransportClientBatchCallback(clientStub, request);
+
+    return callback.post();
+  }
+
+  @Override
+  public void close() throws IOException {
+    for (Map.Entry<String, ManagedChannel> entry: serverGrpcChannels.entrySet()) {
+      entry.getValue().shutdown();
```
gaojieliu, line 105: Can we wrap the `shutdown` call with try/catch?
If one channel fails to shutdown, this function can still continue to shutdown other channels.

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/transport/GrpcTransportClient.java
```diff
@@ -0,0 +1,191 @@
...
+      this.clientStub = clientStub;
+      this.request = request;
+      this.valueFuture = new CompletableFuture<>();
+    }
+
+    public CompletableFuture<TransportClientResponse> get() {
+      clientStub.get(request, new StreamObserver<VeniceServerResponse>() {
+        @Override
+        public void onNext(VeniceServerResponse value) {
+          valueFuture.complete(
```
gaojieliu, line 127: So can this function be invoked multiple times?

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/transport/GrpcTransportClient.java
```diff
@@ -0,0 +1,191 @@
...
+        VeniceClientRequest request) {
+      this.clientStub = clientStub;
+      this.request = request;
+      this.valueFuture = new CompletableFuture<>();
+    }
+
+    public CompletableFuture<TransportClientResponse> post() {
+      clientStub.batchGet(request, new StreamObserver<VeniceServerResponse>() {
+        @Override
+        public void onNext(VeniceServerResponse value) {
```
gaojieliu, line 167: Can this function be invoked multiple times?
Also it seems it doesn't support the proper compression header here.

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/transport/GrpcTransportClient.java
```diff
@@ -0,0 +1,191 @@
...
+          LOGGER.info("Completed gRPC request");
+        }
+      });
+
+      return valueFuture;
+    }
+
+  }
+
+  public static class GrpcTransportClientBatchCallback {
```
gaojieliu, line 151: What is the difference between `GrpcTransportClientBatchCallback` and `GrpcTransportClientCallback`?
They look quite similar, so can we reuse it?

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/transport/GrpcTransportClient.java
```diff
@@ -0,0 +1,191 @@
...
+
+        @Override
+        public void onError(Throwable t) {
+          LOGGER.error("Error in gRPC request", t);
+          valueFuture.completeExceptionally(t);
+        }
+
+        @Override
+        public void onCompleted() {
+          LOGGER.info("Completed batch get gRPC request");
```
gaojieliu, line 184: What the sequence of these function calls?
For a regular request, `onNext` will be invoked and then `onCompleted`?

internal/venice-common/src/main/proto/VeniceReadService.proto
```diff
@@ -0,0 +1,27 @@
...
+
+service VeniceReadService {
+  rpc get (VeniceClientRequest) returns (VeniceServerResponse) {}
+  rpc batchGet(VeniceClientRequest) returns (VeniceServerResponse) {}
+}
+
+message VeniceClientRequest {
+  uint32 partition = 1;
+  string keyString = 2;           // used for single get
+  bytes  keyBytes = 3;            // used for batch get
```
gaojieliu, line 14: Can we reuse one field: `keyBytes` for both request types?

internal/venice-common/src/main/proto/VeniceReadService.proto
```diff
@@ -0,0 +1,27 @@
+syntax = 'proto3';
+package com.linkedin.venice.protocols;
+
+option java_multiple_files = true;
+
+service VeniceReadService {
```
gaojieliu, line 6: It seems the GPRC implementation doesn't leverage request/response headers, is it expected?

internal/venice-common/src/main/proto/VeniceReadService.proto
```diff
@@ -0,0 +1,27 @@
...
+  string storeName = 6;
+  bool   isStreamingRequest = 7;
+  bool   isRetryRequest = 8;
+}
+
+message VeniceServerResponse {
+  sint32 schemaId = 1;
+  bytes  data = 2;
+  uint32 dataSize = 3;
+  bytes  serializedArr = 4;
```
gaojieliu, line 26: Is `dataSize` necessary since `data` is already a byte array?

Also what is `serializedArr`?
I suggest we add enough documents in the proto definition, and I assume they will be carried over to the generated java classes, right?

internal/venice-common/src/main/proto/VeniceReadService.proto
```diff
@@ -0,0 +1,27 @@
...
+  string keyString = 2;           // used for single get
+  bytes  keyBytes = 3;            // used for batch get
+  sint32 requestTimeoutInNS = 4;
+  string resourceName = 5;
+  string storeName = 6;
+  bool   isStreamingRequest = 7;
+  bool   isRetryRequest = 8;
+}
+
+message VeniceServerResponse {
```
gaojieliu, line 22: It seems this definition doesn't include compression header here.

services/venice-server/src/main/java/com/linkedin/venice/grpc/VeniceGrpcServer.java
```diff
@@ -0,0 +1,46 @@
...
+
+
+public class VeniceGrpcServer {
+  private static final Logger LOGGER = LogManager.getLogger(VeniceGrpcServer.class);
+  private final Server server;
+  private final int port;
+
+  public VeniceGrpcServer(VeniceGrpcServerConfig config) {
+    port = config.getPort();
+    server = Grpc.newServerBuilderForPort(config.getPort(), config.getCredentials())
```
gaojieliu, line 19: This seems to be too simple..
I wonder whether there is a way to configure thread pool size or not.

services/venice-server/src/main/java/com/linkedin/venice/listener/grpc/VeniceReadServiceImpl.java
```diff
@@ -0,0 +1,70 @@
...
+
+    StorageResponseObject response =
+        (StorageResponseObject) storageReadRequestsHandler.handleSingleGetGrpcRequest(getRouterRequest);
+
+    ValueRecord valueRecord = response.getValueRecord();
+
+    return VeniceServerResponse.newBuilder()
+        .setSchemaId(valueRecord.getSchemaId())
+        .setDataSize(valueRecord.getDataSize())
+        .setData(ByteString.copyFrom(valueRecord.getData().array(), 4, valueRecord.getDataSize()))
```
gaojieliu, line 55: Is it possible to avoid data copy here?

services/venice-server/src/main/java/com/linkedin/venice/listener/grpc/VeniceReadServiceImpl.java
```diff
@@ -0,0 +1,70 @@
...
+
+  private VeniceServerResponse handleMultiGetRequest(VeniceClientRequest request) {
+    MultiGetRouterRequestWrapper multiGetRouterRequestWrapper =
+        MultiGetRouterRequestWrapper.parseMultiGetGrpcRequest(request);
+
+    ReadResponse readResponse = storageReadRequestsHandler.handleMultiGetGrpcRequest(multiGetRouterRequestWrapper);
+    int schemaId = readResponse.getResponseSchemaIdHeader();
+    ByteBuf data = readResponse.getResponseBody();
+
+    return VeniceServerResponse.newBuilder().setData(ByteString.copyFrom(data.array())).setSchemaId(schemaId).build();
```
gaojieliu, line 68: Can we avoid the data copy here?

gaojieliu (changes requested):

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/transport/GrpcTransportClient.java
```diff
@@ -0,0 +1,191 @@
...
+  // field has public visibility for unit testing
+  public final Map<String, ManagedChannel> serverGrpcChannels;
+  final Map<String, String> veniceAddressToGrpcAddress;
+
+  public GrpcTransportClient(Map<String, String> nettyServerToGrpc, TransportClient r2TransportClient) {
+    serverGrpcChannels = new HashMap<>();
+    veniceAddressToGrpcAddress = new HashMap<>();
+    this.r2TransportClient = r2TransportClient;
+
+    for (Map.Entry<String, String> entry: nettyServerToGrpc.entrySet()) {
```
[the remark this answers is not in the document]
rkunds: yup! I have changed how we obtain the mapping through the cluster creation options so this translation is no longer needed.

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/transport/GrpcTransportClient.java
```diff
@@ -0,0 +1,191 @@
...
+  public CompletableFuture<TransportClientResponse> get(String requestPath, Map<String, String> headers) {
+    String[] requestParts = RequestHelper.getRequestParts(requestPath);
+
+    if (!requestParts[1].equals("storage")) {
+      LOGGER.debug("Requesting metadata from R2 client");
+      return r2TransportClient.get(requestPath, headers);
+    }
+
+    // Lazily create channel for each venice server
+    ManagedChannel channel = getChannel(requestPath);
```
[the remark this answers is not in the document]
rkunds: `requestParts` does not include data related to the address, only the path. e.g: `https://localhost:1234/storage/new-store_v1/1/abcd` --> `["", "storage", "new-store_v1", "1", "abcd"]` . I think making another helper function, in this case, would be beneficial.

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/transport/GrpcTransportClient.java
rkunds: we will be able to use SSL to create secure connections. In order to enable TLS on a server we will have to provide the Certificate Chain File and the Private Key file. From what I have seen, there are no functions within Venice wrt to SSL utils that provides the path to these files. It should be straightforward otherwise. One solution is to add additional fields within `SSLConfig` which would have the path of the necessary files. https://grpc.io/docs/guides/auth/

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/transport/GrpcTransportClient.java
rkunds: Yes, sorry. We do not need `StoreName` when sending the request to the server.

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/transport/GrpcTransportClient.java
```diff
@@ -0,0 +1,191 @@
...
+      String requestPath,
+      Map<String, String> headers,
+      byte[] requestBody) {
+    String[] requestParts = RequestHelper.getRequestParts(requestPath);
+
+    ManagedChannel channel = getChannel(requestPath);
+    VeniceClientRequest request = VeniceClientRequest.newBuilder()
+        .setStoreName(requestParts[2])
+        .setResourceName(requestParts[2])
+        .setPartition(-1)
```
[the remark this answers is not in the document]
rkunds: No 😅

internal/venice-common/src/main/proto/VeniceReadService.proto
rkunds: it is not necessary anymore. I had utilized those fields at a previous iteration while testing and understanding how responses are sent back to the client. We no longer need those fields.

internal/venice-common/src/main/proto/VeniceReadService.proto
rkunds: we can, but then we would have to obtain the `keyBytes` on the client side. As of today, when the server receives a `single-get` request, `GetRouterRequest.parseGetHttpRequest` constructs a `GetRouterRequest` object to pass to `StorageReadRequestsHandler`. Here, `GetRouterRequest` has the key in the form of a string, where the method `getKeyBytesFromUrlKeyString` will decode the query string to get corresponding bytes for the key. We would have to perform this step on the client side if we were to reuse the same field for both request types.

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/transport/GrpcTransportClient.java
rkunds: post() will be invoked once, and in this implementation onNext will also be invoked once since we are not streaming any of the responses/requests. The compression header has been added.

internal/venice-common/src/main/proto/VeniceReadService.proto
rkunds: As of right now, we can use gRPC interceptors in order to inject additional data into messages as they are sent/received from the client or server. For fields that are added to headers when we perform `STORAGE` requests, we will add another field in the client/server message in order to achieve the same functionality.

services/venice-server/src/main/java/com/linkedin/venice/grpc/VeniceGrpcServer.java
rkunds: It is, however, it is not recommended based on the workload and may not be as performant. We can pass a custom executor when creating the gRPC server to best suit the workload.

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/transport/GrpcTransportClient.java
rkunds: After looking at the method again, `ManagedChannel.shutdown()` does not throw any kind of exception.

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/transport/GrpcTransportClient.java
rkunds: I do not believe we can. When creating the gRPC message, the Java type for a .proto byte array is a `com.google.protobuf.ByteString`. A `ByteString` is an immutable sequence of bytes, and I did not find any exposed API to create a ByteString that uses the underlying byte allocation of an existing array or ByteBuf.

services/venice-server/src/main/java/com/linkedin/venice/grpc/VeniceGrpcServer.java
```diff
@@ -0,0 +1,46 @@
...
+        .addService(ServerInterceptors.intercept(config.getService(), config.getInterceptors()))
+        .build();
+  }
+
+  public void start() throws VeniceException {
+    try {
+      server.start();
+    } catch (IOException exception) {
+      LOGGER.error("Failed to start gRPC Server for service {} on port {}", server.getServices(), port, exception);
```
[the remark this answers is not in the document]
rkunds: we do not have access to `config` at this point. Should we add it as a class variable?

services/venice-server/src/test/java/com/linkedin/venice/grpc/VeniceReadServiceImplTest.java
```diff
@@ -0,0 +1,80 @@
...
+
+public class VeniceReadServiceImplTest {
+  private StorageReadRequestsHandler mockStorageReadRequestsHandler;
+  private StreamObserver<VeniceServerResponse> mockResponseObserver;
+
+  private VeniceReadServiceImpl veniceReadService;
+
+  @BeforeTest
+  public void setUp() {
+    MockitoAnnotations.initMocks(this);
```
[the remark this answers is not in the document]
rkunds: we do not, I previously used annotations and neglected to remove that line.

services/venice-server/src/test/java/com/linkedin/venice/grpc/VeniceReadServiceImplTest.java
rkunds: it isn't we can achieve the same with `Mockito.clearInvocations`

services/venice-server/src/main/java/com/linkedin/venice/listener/grpc/VeniceReadServiceImpl.java
```diff
@@ -0,0 +1,70 @@
...
+  StorageReadRequestsHandler storageReadRequestsHandler;
+
+  public VeniceReadServiceImpl(StorageReadRequestsHandler storageReadRequestsHandler) {
+    LOGGER.info("Created gRPC Server for VeniceReadService");
+    this.storageReadRequestsHandler = storageReadRequestsHandler;
+  }
+
+  @Override
+  public void get(VeniceClientRequest request, StreamObserver<VeniceServerResponse> responseObserver) {
+    VeniceServerResponse grpcResponse = handleSingleGetRequest(request);
```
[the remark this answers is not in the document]
rkunds: I will address this in the follow up pr regarding gRPC integration with Venice Handlers.

services/venice-server/src/main/java/com/linkedin/venice/listener/grpc/VeniceReadServiceImpl.java
rkunds: I do not believe so. We need to perform a data copy when creating the ByteString object. https://protobuf.dev/reference/java/api-docs/com/google/protobuf/ByteString. 
https://protobuf.dev/reference/java/api-docs/com/google/protobuf/ByteString.html#readFrom-java.io.InputStream-

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/ClientConfig.java
```diff
@@ -78,6 +78,13 @@ public class ClientConfig<K, V, T extends SpecificRecord> {
    * get based batchGet support.
    */
   private final boolean useStreamingBatchGetAsDefault;
+  private final boolean useGrpc;
+  /**
+   * This is a temporary solution to support gRPC with Venice, we will replace this with retrieving information about
+   * gRPC servers when we make a request to receive Metadata from a server to obtain information in order to successfully
+   * route requests to the correct server/partition
+   */
+  private final Map<String, String> nettyServerToGrpcAddressMap;
```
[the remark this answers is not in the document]
sushantmane: Yeah this exactly what we are thinking about. Will also check with Helix team before making any change. Thanks

services/venice-server/src/main/java/com/linkedin/venice/listener/grpc/VeniceReadServiceImpl.java
```diff
@@ -0,0 +1,85 @@
...
+import com.linkedin.venice.protocols.VeniceServerResponse;
+import io.grpc.stub.StreamObserver;
+import io.netty.buffer.ByteBuf;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+
+public class VeniceReadServiceImpl extends VeniceReadServiceGrpc.VeniceReadServiceImplBase {
+  private static final Logger LOGGER = LogManager.getLogger(VeniceReadServiceImpl.class);
+  StorageReadRequestHandler storageReadRequestHandler;
```
sushantmane, line 22: storageReadRequestHandler  should be private?

sushantmane (changes requested):

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/transport/GrpcTransportClient.java
```diff
@@ -0,0 +1,182 @@
...
+import io.grpc.stub.StreamObserver;
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+
+public class GrpcTransportClient extends InternalTransportClient {
+  private static final Logger LOGGER = LogManager.getLogger(GrpcTransportClient.class);
+  private static final String STORAGE_ACTION = "storage";
```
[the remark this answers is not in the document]
rkunds: do we check action types in this manner in other areas?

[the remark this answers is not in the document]
sushantmane: I do see them at few places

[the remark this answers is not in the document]
rkunds: i see, then yes. It would be a good idea.

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/transport/GrpcTransportClient.java
rkunds: I will look into this.

sushantmane (approved):
Thank you, @rkunds! Great work.

Don't forget to edit the commit message before clicking on squash & merge.

internal/venice-common/src/main/proto/VeniceReadService.proto
rkunds: we have updated it to include the compression value.

build.gradle
```diff
@@ -730,7 +729,10 @@ ext.createDiffFile = { ->
         ':!services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerService.java',
 
         // Other files that have tests but are not executed in the regular unit test task
-        ':!internal/alpini/*'
+        ':!internal/alpini/*',
+
+        // unit test for gRPC Transport Client is not straightforward, adding to exclusion list for now
```
gaojieliu, line 734: This needs to be addressed in the near future since it is critical to the GRPC path.
I assume it should be doable by passing some mock GRPC implementation.

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/ClientConfig.java
```diff
@@ -108,13 +115,19 @@ private ClientConfig(
...
+      boolean useStreamingBatchGetAsDefault,
+      boolean useGrpc,
+      Map<String, String> nettyServerToGrpcAddressMap) {
     if (storeName == null || storeName.isEmpty()) {
       throw new VeniceClientException("storeName param shouldn't be empty");
     }
     if (r2Client == null) {
       throw new VeniceClientException("r2Client param shouldn't be null");
     }
+    if (useGrpc && nettyServerToGrpcAddressMap == null) {
```
gaojieliu, line 127: BTW, with such step, it won't work in EI/Prod env, which means we can't test/benchmark it in EI env.
Since leveraging Helix Instance config can be a more involved change, can we create a new endpoint in Netty Server, so that when grpc fast-client discovers the netty address via metadata request, it can send an request to the Netty server to figure out the GRPC port and cache it locally.
This may not be the ultimate solution, but it can work with minimal overhead since cluster setup doesn't change often.

If you agree, you can build such mechanism in a following PR.

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/transport/GrpcTransportClient.java
```diff
@@ -0,0 +1,182 @@
...
+      LOGGER.debug(
+          "performing unsupported gRPC transport client action ({}), passing request to R2 client",
+          requestParts[3]);
+      return r2TransportClient.get(requestPath, headers);
+    }
+
+    ManagedChannel channel = getChannel(requestParts[2]);
+
+    VeniceClientRequest request = VeniceClientRequest.newBuilder()
+        .setResourceName(requestParts[4])
```
gaojieliu, line 65: Does GRPC have capability to carry request headers to the backend?
Such as client compression or compute header, and it will be good to be able to carry these headers to the backend.

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/transport/GrpcTransportClient.java
```diff
@@ -0,0 +1,182 @@
...
+        .setIsBatchRequest(true)
+        .build();
+
+    VeniceReadServiceGrpc.VeniceReadServiceStub clientStub = getStub(channel);
+    GrpcTransportClientCallback callback = new GrpcTransportClientCallback(clientStub, request);
+    return callback.post();
+  }
+
+  @Override
+  public void close() {
```
gaojieliu, line 103: It is good to call `r2TransportClient.close()` here even that function is doing nothing now.

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/transport/GrpcTransportClient.java
```diff
@@ -0,0 +1,182 @@
...
+          valueFuture.complete(
+              new TransportClientResponse(
+                  value.getSchemaId(),
+                  CompressionStrategy.valueOf(value.getCompressionStrategy()),
+                  value.getData().toByteArray()));
+        }
+
+        @Override
+        public void onError(Throwable t) {
+          LOGGER.error("Error in gRPC request", t);
```
gaojieliu, line 139: Normally, we don't want to log the exception here since normally application will catch the exception anyway.

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/transport/GrpcTransportClient.java
```diff
@@ -0,0 +1,182 @@
...
+          valueFuture.completeExceptionally(t);
+        }
+
+        @Override
+        public void onCompleted() {
+          LOGGER.debug("Completed gRPC request");
```
gaojieliu, line 145: What is the difference between `onCompleted` and `onNext`?

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/transport/GrpcTransportClient.java
```diff
@@ -0,0 +1,182 @@
...
+        }
+      });
+
+      return valueFuture;
+    }
+
+    public CompletableFuture<TransportClientResponse> post() {
```
gaojieliu, line 152: `post` is sharing most of the code as `get` function, so can we reduce the duplicate codes by creating a common inner class?

internal/venice-common/src/main/proto/VeniceReadService.proto
gaojieliu: Can you cut a ticket track this? or at least leave a TODO in the code?

services/venice-server/src/main/java/com/linkedin/venice/grpc/VeniceGrpcServer.java
gaojieliu: Can you leave a `TODO` here?

gaojieliu (commented):
Left some more comments, and you don't have to address them in this PR.
I am fine if you cut tracking ticket to track these issues and fix them in a following PR.

build.gradle
rkunds: I agree. Currently, our gRPC service implementation is within `/services/venice-server`, we can move this to `venice-common` in order to create a mock implementation, but, the service implementation has a lot of dependencies within `venice-server` such as `RouterRequest` and `ReadResponse`. As such, moving this to the common folder is probably not a good idea due to these dependencies. I am looking into other solutions in order to create unit tests for this.

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/transport/GrpcTransportClient.java
rkunds: We can pass headers in this manner via attaching metadata to the response with a gRPC service interceptor. I will look into this.

sushantmane: We can also consider adding it as part of the API itself

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/transport/GrpcTransportClient.java
rkunds: `onNext` will be invoked every time the server sends/streams a message to the client, and `onCompleted` will be invoked when the `rpc` is completed.

build.gradle
rkunds: Created an internal ticket to track the change.

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/transport/GrpcTransportClient.java
rkunds: I will address in a later PR.

sushantmane (approved):

sushantmane (changes requested):
fix build.gralde

sushantmane (approved):

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/transport/GrpcTransportClient.java
gaojieliu: I see. this will be useful for a true streaming response.

gaojieliu (approved):
Thanks a lot @rkunds !

Let us iterate the action items in a separate PR to make it more robust!

PR #476, opened 2023-06-08
[all] Raise ConnectException only if all hosts in ControllerClient are unreachable

nisargthakkar:
## Raise `ConnectException` only if all hosts in `ControllerClient` are unreachable

To instantiate a URL based ControllerClient, we need to pass in a list of Router or Controller URLs. Since this is a static list that might be configured by operators, it can grow stale over time to include hosts that are no longer running Venice routers or controllers on that port. When querying these endpoints the following cases are possible:
1. A different service is now listening on that port and it is likely to throw a 4XX for route not supported or some 5XX that might trigger some error scenarios in the new service.
2. `java.net.ConnectException` is thrown. This is the most probable option since even though different services might get deployed on that host, the ports are kept mutually exclusive.

In error scenarios, we wish to try our best to provide the user with an actionable error and only return a `ConnectException`
when all hosts throw it. In that case, the `ConnectException` is the most actionable error.

## How was this PR tested?

GH CI

### DIFF
--- a/internal/venice-test-common/src/integrationtest/java/com/linkedin/venice/controllerapi/TestControllerClient.java
+++ b/internal/venice-test-common/src/integrationtest/java/com/linkedin/venice/controllerapi/TestControllerClient.java
@@ -97,8 +97,8 @@ public class TestControllerClient {
     String errorResponseStoreName = Utils.getUniqueString("test-error-store");
     String fakeLeaderControllerUrl = "http://fake.leader.controller.url";
 
-    String nonExistentControllerUrl1 = "http://localhost:22";
-    String nonExistentControllerUrl2 = "http://localhost:23";
+    String nonExistentControllerUrl1 = "http://localhost:1100";
+    String nonExistentControllerUrl2 = "http://localhost:1101";
 
     try (MockHttpServerWrapper mockController = ServiceFactory.getMockHttpServer("mock_controller");
         MockHttpServerWrapper mockLegacyHost = ServiceFactory.getMockHttpServer("mock_legacy_router")) {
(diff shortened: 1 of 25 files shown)


FelixGV (dismissed):

internal/venice-test-common/src/integrationtest/java/com/linkedin/venice/controllerapi/TestControllerClient.java
```diff
@@ -97,8 +97,8 @@ public void testControllerClientWithInvalidUrls() throws IOException {
     String errorResponseStoreName = Utils.getUniqueString("test-error-store");
     String fakeLeaderControllerUrl = "http://fake.leader.controller.url";
 
-    String nonExistentControllerUrl1 = "http://localhost:22";
-    String nonExistentControllerUrl2 = "http://localhost:23";
+    String nonExistentControllerUrl1 = "http://localhost:1100";
+    String nonExistentControllerUrl2 = "http://localhost:1101";
```
FelixGV, line 101: Must it be a free port? That could make the test flaky if some other test runs at the same time on the host, right...?

nisargthakkar: Yes. I assumed that using a free port would automatically make a service listen on it. Tested locally that that was wrong. Updating the PR

FelixGV: Ok... Using `Utils::getFreePort` also has a chance to be flaky BTW, but I guess we could let the test-retry framework take care of that...

On that note... `Utils::getFreePort` doesn't belong in `venice-common`, IMHO. We should probably move it to `TestUtils` in `venice-test-common`...

nisargthakkar: Oh yeah. For sure. I'll move it to `TestUtils`

FelixGV (approved):
Thanks!

PR #477, opened 2023-06-08
[server][client][test] Refactor pub-sub message deserialization

sushantmane:
# [server][client][test] Refactor pub-sub message deserialization

This commit introduces changes to improve the handling of pub-sub message deserialization. The following modifications were made:

- `PubSubMessageDeserializer` has been transformed from an interface into a concrete implementation. This change ensures that all pub-sub clients utilize the same deserializer.

Additionally, the following enhancements were made to the integration tests:

- `VeniceTlsConfiguration` is now available to configure and expose TLS for pub-sub brokers during integration testing.

## How was this PR tested?

GHCI


internal/venice-common/src/main/java/com/linkedin/venice/pubsub/api/PubSubConsumerAdapterFactory.java
```diff
@@ -17,15 +17,18 @@
    *                                    veniceProperties using prefix matching. For example, to construct kafka consumer
    *                                    configs that start with "kafka." prefix will be used.
    * @param isOffsetCollectionEnabled   A flag to enable collection of offset or not.
-   * @param pubSubMessageDeserializer   To deserialize the raw byte records into {@link PubSubMessage}s to process.
    * @param consumerName                Name of the consumer. If not null, it will be used to set the context
    *                                    for consumer thread.
    * @return                            Returns an instance of a consumer adapter
    */
+  default ADAPTER create(VeniceProperties veniceProperties, boolean isOffsetCollectionEnabled, String consumerName) {
```
haoxu07, line 24: Why not we just use this interface and let each specific `PubSubConsumerAdapter` like `ApacheKafkaConsumerAdapter` has its inner `ValueDeserializer`. Here the `PubSubClientsConfigs` just contain the value deserializer, is it just for configuarable deserializer purpose? I feel like for a specific deserializer, we either extend or create the specific `ConsumerAdapter`.

haoxu07 (commented):
I have not finished the SSL part review, just comment about the interface.

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/api/PubSubConsumerAdapterFactory.java
sushantmane: > Why not we just use this interface and let each specific PubSubConsumerAdapter like ApacheKafkaConsumerAdapter has its inner ValueDeserializer.

Because some code paths update the `ValueDeserializer` to use a specified schema. 

> Here the PubSubClientsConfigs just contain the value deserializer, is it just for configuarable deserializer purpose?

`PubSubClientsConfigs` will be used to pass general pub-sub settings to adapters. For now it is used just for `ValueDeserializer`.

> I feel like for a specific deserializer, we either extend or create the specific ConsumerAdapter.

With this approach, each pub-sub client implementation of the adapter will have to provide multiple implementations. Are we gonna do this for each config?

haoxu07: For `KafkaValueSerializer`, we only have one extended class `OptimizedKafkaValueSerializer`. The instantiation of original `KafkaValueSerializer` is only used in: `PartitionOffsetFetcherFactory` and `DictionaryUtils`, which I think we can replace it with `OptimizedKafkaValueSerializer`, and if we want to use another type of `OptimizedKafkaValueSerializer`, we can only override the adapter class or adpater factory class to add this, what do you think?

sushantmane: I agree we can replace `KafkaValueSerializer` with `OptimizedKafkaValueSerializer` but I'm not sure if there is any historic reason behind its usage. 

Besides, even with the replacement we still have to pass schemaReader for `KVS` as we do update it at some places in our code. 
```java
    KafkaValueSerializer kafkaValueSerializer = new OptimizedKafkaValueSerializer();
    kafkaMessageEnvelopeSchemaReader.ifPresent(kafkaValueSerializer::setSchemaReader);
```

sushantmane: >do you think that we should put other configs like boolean isOffsetCollectionEnabled into the PubSubClientConfigs to simplify the interface like builder. 

Yeah, I agree. I'm making this change. 


> We can also let adapter to expose the KafkaValueSerializer to let server or controller logic configure it.

This is not a good idea IMHO

clients/venice-admin-tool/src/main/java/com/linkedin/venice/AdminTool.java
```diff
@@ -2866,7 +2870,12 @@ private static void printErr(String errorMessage, Map<String, String> customMess
   }
 
   private static PubSubConsumerAdapter getConsumer(Properties consumerProps) {
+    PubSubMessageDeserializer pubSubMessageDeserializer = new PubSubMessageDeserializer(
+        new OptimizedKafkaValueSerializer(),
+        new LandFillObjectPool<>(KafkaMessageEnvelope::new),
+        new LandFillObjectPool<>(KafkaMessageEnvelope::new));
     return new ApacheKafkaConsumerAdapterFactory()
-        .create(new VeniceProperties(consumerProps), false, "admin-tool-topic-dumper");
+        .create(new VeniceProperties(consumerProps), false, pubSubMessageDeserializer, "");
```
haoxu07, line 2878: I think keeping the consumer name is better. :-)

haoxu07 (dismissed):
Overall LGTM! Could you add some description into this PR? Better for tracking purpose.

clients/venice-admin-tool/src/main/java/com/linkedin/venice/AdminTool.java
sushantmane: Done

haoxu07 (approved):

sushantmane:
Thanks for the review, @haoxu07!

PR #478, opened 2023-06-09
[admin-tool] Fix an issue in data recovery executor

lluwm:
In data recovery executor, there is an issue in detecting if there is an ongoing push in system. `getFutureVersions` can return NON_EXISTING_VERSION to indicate that none of future version is available to read, but current implementation doesn't check it, thus could cause a false positive in the result. This rb fixes this bug by adding additional check against NON_EXISTING_VERSION value.

## How was this PR tested?
Local manual test against EI.
Internal CI passed.


lyssarzr (approved):
LGTM

PR #480, opened 2023-06-09
[controller] Check store existing A/A config when updating store type

sixpluszero:
## [controller] Check store existing A/A config when updating store type
This PR aims to fix the bug that when changing store type (e.g. batch <-> pure hybrid <-> inc push), we set the store A/A config based on default config only but not looking at store's existing A/A config: If the default A/A config for new store type is false and existing config is true, we should also keep A/A enabled.

## How was this PR tested?

Update existing integration test to reflect the logic.


internal/venice-test-common/src/integrationtest/java/com/linkedin/venice/controller/TestClusterLevelConfigForActiveActiveReplication.java
adamxchen, line 1: I think integ test is a bit heavy for such config-related testing. We will get there eventually...

adamxchen (approved):
Thank you for the change!

internal/venice-test-common/src/integrationtest/java/com/linkedin/venice/controller/TestClusterLevelConfigForActiveActiveReplication.java
sixpluszero: Yeah I agree. Although my feeling is the VeniceHelixAdmin has way to many logics and variables - not sure if it is easy to add unit test to it directly :)

PR #482, opened 2023-06-10
[build] Fix configuration of custom SourceSets with protobuf

nisargthakkar:
## Fix configuration of custom SourceSets with protobuf

Protobuf gradle plugin adds `proto` to either the `main` or `test` `SourceSets`. It appears that when it adds the protobuf support to custom `SourceSets` like `integrationTest` where paths don't match the name, it misconfigures how the IDE parses it. This commit is an attempt to fix the IDE support for proto

## How was this PR tested?

Locally on IntelliJ and GH CI for regression testing


adamxchen (approved):
I've tested this PR and it worked well for me. Great work!

PR #483, opened 2023-06-10
[test] Fix flaky test TestAdminToolEndToEnd.testWipeClusterCommand

lluwm:
On Kafka side, if a topic hasn't been fully deleted before asked to create a new topic with the same name, it can fail. This change fixes the issue by waiting all topics to be fully removed from Kafka service before proceeding to the next steps in the test.

## How was this PR tested?
Internal CI (No testWipeClusterCommand failure in Venice-WcTest: #8046, #8047; Venice-WcTest8 #1225, #1226)

### DIFF
--- a/internal/venice-test-common/src/integrationtest/java/com/linkedin/venice/controller/TestAdminToolEndToEnd.java
+++ b/internal/venice-test-common/src/integrationtest/java/com/linkedin/venice/controller/TestAdminToolEndToEnd.java
@@ -3,6 +3,7 @@ package com.linkedin.venice.controller;
 import static com.linkedin.venice.ConfigKeys.ALLOW_CLUSTER_WIPE;
 import static com.linkedin.venice.ConfigKeys.LOCAL_REGION_NAME;
 import static com.linkedin.venice.ConfigKeys.TOPIC_CLEANUP_DELAY_FACTOR;
+import static org.testng.Assert.assertFalse;
 
 import com.linkedin.venice.AdminTool;
 import com.linkedin.venice.Arg;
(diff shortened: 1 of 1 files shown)


internal/venice-test-common/src/integrationtest/java/com/linkedin/venice/controller/TestAdminToolEndToEnd.java
```diff
@@ -140,6 +145,20 @@ public void testWipeClusterCommand() throws Exception {
       MultiStoreResponse multiStoreResponse = controllerClient.queryStoreList(false);
       Assert.assertEquals(multiStoreResponse.getStores().length, 0);
 
+      // Wait until all topics are indeed removed from Kafka service.
+      PubSubTopicRepository pubSubTopicRepository = new PubSubTopicRepository();
+
+      PubSubTopic testStoreTopic1 = pubSubTopicRepository.getTopic(Version.composeKafkaTopic(testStoreName1, 1));
```
adamxchen, line 151: why not just do `new PubSubTopicImpl(Version.composeKafkaTopic(testStoreName1, 1))`?

adamxchen (approved):
Thanks for the fix! Have one comment left

internal/venice-test-common/src/integrationtest/java/com/linkedin/venice/controller/TestAdminToolEndToEnd.java
lluwm: Thanks for the review Adam. I think that's simpler, will fix it.

lluwm: In fact, if I go with what you suggested, I have to make `PubSubTopicImpl` constructor public , which seems to disagree the author's intention. maybe I will just keep it as it is.

```
  /**
   * N.B. leaving this constructor package-private, as the intent is not necessarily the instantiate this object
   * from anywhere in the code. Most likely, it would be preferable to have some form of singleton topic-registry where
   * i
```nstances can be gotten from, so that the internal state is computed just once, and then reused for the lifetime
   * of the topic.
   */
  PubSubTopicImpl(String name) {
    this.name = Objects.requireNonNull(name, "Topic name cannot be null");
    this.pubSubTopicType = PubSubTopicType.getPubSubTopicType(name);
    this.storeName = Version.parseStoreFromKafkaTopicName(name);
  }

PR #484, opened 2023-06-12
[client] Use HttpClient5Utils for HTTP1 connections

eolivelli:
## Summary
It is better to use the battle tested HttpClient5Utils utility to build the Http Client also for HTTP1.
This generally leads to better numbers because of the configuration of the IO Reactor, it also reduces the tech debt due to duplicate code.

## How was this PR tested?

This is a trivial rework, already covered by existing tests. We are actually removing code.

### DIFF
--- a/clients/venice-thin-client/src/main/java/com/linkedin/venice/client/store/transport/HttpTransportClient.java
+++ b/clients/venice-thin-client/src/main/java/com/linkedin/venice/client/store/transport/HttpTransportClient.java
@@ -64,8 +64,6 @@ public class HttpTransportClient extends TransportClient {
 
     if (requireHHtp2) {
       // HTTP2 only client doesn't allow you to configure the ConnectionManager, 1 connection is enough
-      // this is kind of problem for K8S deployment, where we have multiple pods
-      // but we access the router via a single Service
       LOGGER.info("Creating a TLS HTTP2 only client to {}", routerUrl);
       return new HttpClient5Utils.HttpClient5Builder().setHttp1(false)
           .setSslContext(sslFactory.getSSLContext())


clients/venice-thin-client/src/main/java/com/linkedin/venice/client/store/transport/HttpTransportClient.java
```diff
@@ -67,27 +63,24 @@ protected static CloseableHttpAsyncClient buildClient(
       SSLFactory sslFactory) {
 
     if (requireHHtp2) {
-      // HTTP2 only client doesn't allow you to configure the ConnectionManager
+      // HTTP2 only client doesn't allow you to configure the ConnectionManager, 1 connection is enough
+      // this is kind of problem for K8S deployment, where we have multiple pods
+      // but we access the router via a single Service
```
gaojieliu, line 68: I don't fully follow this javadoc.
Is it a problem to have a conn for route with H2 or not?
If it is a problem, in LinkedIn, we would spin up multiple H2 clients to do round-robin traffic routing.

gaojieliu (commented):
Just left a comment to clarify the javasoc.

clients/venice-thin-client/src/main/java/com/linkedin/venice/client/store/transport/HttpTransportClient.java
eolivelli: I can remove this comment.

The problem with having many routers behind one single dns name (and IP address probably) is that http client will only open one connection and so it will talk with only 1 router service.

On the client side you won't ever know to which router you are connected to and there is now way to ensure that the load is distributed among different instances.

I am not sure that the generic user will try to start multiple clients in order to have multiple chances to get better.

At this point here we cannot do much

Cc @michealmarshal

gaojieliu: Sure.
If it is useful, you can test it out in your env to how it goes with multiple http2 clients.
I assume there will be many client applications talking to the single dns and these connections may be distributed to different instances.

FelixGV: HC4 and HC5 both have this issue. Right?

eolivelli: I don't think that HC4 has support for HTTP2.

Let me remove the comment

FelixGV: I didn't mean to talk about HTTP2 support in HC4, but rather that you pointed out the client would connect to a single router instance.

With the D2 client, there is client-side load balancing across many router instances, so we never paid attention to that for the non-D2 client. But it might be desirable to have that as well...

I just wanted to confirm that this behavior existed in HC4 as well.

internal/venice-client-common/src/main/java/com/linkedin/venice/httpclient5/HttpClient5Utils.java
```diff
@@ -43,6 +45,10 @@ public static class HttpClient5Builder {
     private int ioThreadCount = 48;
     private boolean skipCipherCheck = false;
 
+    private boolean http1 = false;
+    private int http1MaxConnectionsTotal = 0;
```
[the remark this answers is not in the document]
eolivelli: Zero means that HC uses its own defaults.
It is the default configuration of HC.
Behind the hood if you keep 0 then you have the same behavior as before, that is IIUC max 2 connections.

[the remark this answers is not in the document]
gaojieliu: I guess that is fine then for non-critical usage.

gaojieliu (approved):

PR #487, opened 2023-06-13
[compat][common][controller][server][admin-tool] Introduce new store level config for SN read quota

xunyin8:
## Summary
Deployment order: Meta system store schema and admin operation schema updated. Controllers before servers and child controllers before parent controllers.

New store level config storageNodeReadQuotaEnabled to provide dynamic store level control when rolling out storage node read quota. To enable storage node read quota for a store the servers in the corresponding cluster need to have server.quota.enforcement.enabled and the store config storageNodeReadQuotaEnabled set to true

## How was this PR tested?
Existing unit and integration tests. Added new unit tests.


internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/fastclient/utils/AbstractClientEndToEndSetup.java
```diff
@@ -238,6 +239,8 @@ protected void prepareData() throws Exception {
     VersionCreationResponse creationResponse = veniceCluster.getNewStoreVersion(KEY_SCHEMA_STR, VALUE_SCHEMA_STR);
     storeVersionName = creationResponse.getKafkaTopic();
     storeName = Version.parseStoreFromKafkaTopicName(storeVersionName);
+    veniceCluster.useControllerClient(
+        client -> client.updateStore(storeName, new UpdateStoreQueryParams().setStorageNodeReadQuotaEnabled(true)));
```
nisargthakkar, line 243: Instead of having this here, let's put this test in `TestParentControllerWithMultiDataCenter.testUpdateStore`?

services/venice-server/src/main/java/com/linkedin/venice/listener/ReadQuotaEnforcementHandler.java
```diff
@@ -126,15 +126,23 @@ public boolean isInitialized() {
 
   @Override
   public void channelRead0(ChannelHandlerContext ctx, RouterRequest request) {
-    if (!isInitialized()) {
-      // If we haven't completed initialization, allow all requests
+    String storeName = Version.parseStoreFromKafkaTopicName(request.getResourceName());
```
nisargthakkar, line 129: Let's use `request.getStoreName()` directly?

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/fastclient/utils/AbstractClientEndToEndSetup.java
xunyin8: I can add a test in testUpdateStore but we also need to enable it here because FC tests should be executed with SN read quota enforced and there is a test `FastClientServerReadQuotaTest` that explicitly test SN read quota.

nisargthakkar: Oh, I see. Then there is no need to add it in the other place

services/venice-server/src/main/java/com/linkedin/venice/listener/ReadQuotaEnforcementHandler.java
xunyin8: Good idea, will do.

nisargthakkar (approved):

gaojieliu (commented):
So when the store-level quota is enabled, will Router skip the quota enforcement for this specific store?
Based on the code, it seems such store will apply  both Router and Storage node quota, right?

I am wondering whether this would cause any issue or not when the customer adopts fast-client colo by colo.
So maybe we could consider to only apply storage node quota to fast-client traffic as the first step to reduce the blast radius in case some bad happens?

xunyin8:
> So when the store-level quota is enabled, will Router skip the quota enforcement for this specific store? Based on the code, it seems such store will apply both Router and Storage node quota, right?
> 
> I am wondering whether this would cause any issue or not when the customer adopts fast-client colo by colo. So maybe we could consider to only apply storage node quota to fast-client traffic as the first step to reduce the blast radius in case some bad happens?

Router quota will still be enforced regardless of the store level storage read quota config if there is traffic going through router.

gaojieliu (approved):

PR #489, opened 2023-06-13
[davinci][ii] Remove zk usage from II for davinci stores

ZacAttack:
## [davinci][ii] Remove zk usage from II for davinci stores

User libraries should not access zk.  Additional work could be done to remove zk completely from the sub project, but it would seem the way things are architected currently that could be quite a lot of work, and we need to make some incremental progress here.

## How was this PR tested?
Running tests with this now


FelixGV (dismissed):
Thanks for the cleanup!!

sixpluszero (approved):
The change itself LGTM! But the test AvroCompatibility one seems cancelled/timed out, could you please retry it?

sixpluszero (approved):
LGTM, thanks for the fix and the test fix!

PR #491, opened 2023-06-13
[producer] Disable logs for the shell producer

nisargthakkar:
## Disable logs for the shell producer

The shell producer currently prints loads of logs that clutter the terminal. The previous attempt at stopping logs did not stop all logs. This one prevents all logs from being printed to the terminal.

## How was this PR tested?

Tested locally

### DIFF
--- a/clients/venice-producer/src/main/java/com/linkedin/venice/producer/online/ProducerTool.java
+++ b/clients/venice-producer/src/main/java/com/linkedin/venice/producer/online/ProducerTool.java
@@ -4,6 +4,7 @@ import static com.linkedin.venice.CommonConfigKeys.SSL_FACTORY_CLASS_NAME;
 import static com.linkedin.venice.VeniceConstants.DEFAULT_SSL_FACTORY_CLASS_NAME;
 
 import com.linkedin.avroutil1.compatibility.AvroCompatibilityHelper;
+import com.linkedin.venice.LogConfigurator;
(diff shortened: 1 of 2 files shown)


clients/venice-producer/src/main/java/com/linkedin/venice/producer/online/ProducerTool.java
```diff
@@ -76,8 +74,7 @@ public class ProducerTool {
       .addOption(HELP_OPTION);
 
   public static void main(String[] args) throws Exception {
-    // Disable Logging
-    Configurator.setAllLevels(LogManager.getRootLogger().getName(), Level.OFF);
+    LogConfigurator.disableLog();
```
sushantmane, line 77: Can we try to use `log4j2.properties` file to control the logging output instead of switching it through the code

nisargthakkar: I'm not sure how to specify different properties files for the shadow jar vs the regular jar and I'm not too keen on spending much cycles for it. These things generally tend to end up being rabbit holes. 

This way, we can control the log behavior in a quick (and admittedly, dirty) way

sushantmane: We could use a props file for the shadow jar. For regular jar-calling applications can control logging behavior as per their need.

sushantmane (approved):

PR #492, opened 2023-06-15
[controller] Return proper ErrorType in getFutureVersion API when Store doesn't exist

nisargthakkar:
## Return proper `ErrorType` in `getFutureVersion` API when Store doesn't exist

When a store doesn't exist, the `list_future_versions` API on controllers returns a map with 0 as the future version. This commit changes the response to error with a `STORE_NOT_FOUND` ErrorType

## How was this PR tested?

Added unit tests


adamxchen (dismissed):
Thank you for catching this and adding additional unit tests!

PR #493, opened 2023-06-15
[dvc][controller][server][test] Support plugging in `PubSubClientFactory` for integration test and venice components.

haoxu07:

This change supports plugging in `PubSubClientFactory` from server config and controller config. This will enable configuration passing specific combination of PubSubClients. For example, we may have admin factory and producer factory for one type of PubSub Client, consumer factory is in other type. This change will also benefit Ingestion Isolation and DaVinci client with specific PubSub Client. 

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/config/VeniceServerConfig.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/config/VeniceServerConfig.java
@@ -121,6 +121,9 @@ import com.linkedin.venice.exceptions.ConfigurationException;
 import com.linkedin.venice.exceptions.VeniceException;
 import com.linkedin.venice.meta.IngestionMode;
 import com.linkedin.venice.pubsub.adapter.kafka.admin.ApacheKafkaAdminAdapter;
+import com.linkedin.venice.pubsub.adapter.kafka.admin.ApacheKafkaAdminAdapterFactory;
+import com.linkedin.venice.pubsub.adapter.kafka.consumer.ApacheKafkaConsumerAdapterFactory;
+import com.linkedin.venice.pubsub.adapter.kafka.producer.ApacheKafkaProducerAdapterFactory;
 import com.linkedin.venice.pubsub.api.PubSubAdminAdapterFactory;
 import com.linkedin.venice.pubsub.api.PubSubConsumerAdapterFactory;
 import com.linkedin.venice.pubsub.api.PubSubProducerAdapterFactory;
@@ -663,19 +666,16 @@ public class VeniceServerConfig extends VeniceClusterConfig {
     this.divProducerStateMaxAgeMs =
         serverProperties.getLong(DIV_PRODUCER_STATE_MAX_AGE_MS, KafkaDataIntegrityValidator.DISABLED);
     try {
-      String producerFactoryClassName = serverProperties.getString(
-          PUB_SUB_PRODUCER_ADAPTER_FACTORY_CLASS,
-          "com.linkedin.venice.pubsub.adapter.kafka.producer.ApacheKafkaProducerAdapterFactory");
+      String producerFactoryClassName = serverProperties
+          .getString(PUB_SUB_PRODUCER_ADAPTER_FACTORY_CLASS, ApacheKafkaProducerAdapterFactory.class.getName());
       pubSubProducerAdapterFactory =
           (PubSubProducerAdapterFactory) Class.forName(producerFactoryClassName).newInstance();
-      String consumerFactoryClassName = serverProperties.getString(
-          PUB_SUB_CONSUMER_ADAPTER_FACTORY_CLASS,
-          "com.linkedin.venice.pubsub.adapter.kafka.consumer.ApacheKafkaConsumerAdapterFactory");
+      String consumerFactoryClassName = serverProperties
+          .getString(PUB_SUB_CONSUMER_ADAPTER_FACTORY_CLASS, ApacheKafkaConsumerAdapterFactory.class.getName());
       pubSubConsumerAdapterFactory =
           (PubSubConsumerAdapterFactory) Class.forName(consumerFactoryClassName).newInstance();
-      String adminFactoryClassName = serverProperties.getString(
-          PUB_SUB_ADMIN_ADAPTER_FACTORY_CLASS,
-          "com.linkedin.venice.pubsub.adapter.kafka.admin.ApacheKafkaAdminAdapterFactory");
+      String adminFactoryClassName = serverProperties
+          .getString(PUB_SUB_ADMIN_ADAPTER_FACTORY_CLASS, ApacheKafkaAdminAdapterFactory.class.getName());
       pubSubAdminAdapterFactory = (PubSubAdminAdapterFactory) Class.forName(adminFactoryClassName).newInstance();
     } catch (ClassNotFoundException | IllegalAccessException | InstantiationException e) {
       LOGGER.error("Failed to create an instance of pub sub clients factory", e);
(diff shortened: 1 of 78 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaStoreIngestionService.java
```diff
@@ -323,7 +323,7 @@ public KafkaStoreIngestionService(
         .setPubSubTopicRepository(pubSubTopicRepository)
         .setMetricsRepository(metricsRepository)
         .setLocalKafkaBootstrapServers(serverConfig.getKafkaBootstrapServers())
-        .setPubSubConsumerAdapterFactory(new ApacheKafkaConsumerAdapterFactory())
```
sushantmane, line 326: Can you replace all instances of `Apache*AdapterFactory` with the corresponding `pubSubClientsFactory.get*AdapterFactory()`? For example, L423.

sushantmane (changes requested):

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaStoreIngestionService.java
haoxu07: I realized we need to add config for all client adapters setting, so that `DavinciBackend` and `IngestionIsolationServer` could pick up those from config.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/config/VeniceServerConfig.java
```diff
@@ -632,6 +643,26 @@ public VeniceServerConfig(VeniceProperties serverProperties, Map<String, Map<Str
     ingestionMemoryLimit = extractIngestionMemoryLimit(serverProperties, ingestionMode, forkedProcessJvmArgList);
     LOGGER.info("Ingestion memory limit: {} after subtracting other usages", ingestionMemoryLimit);
     ingestionMlockEnabled = serverProperties.getBoolean(INGESTION_MLOCK_ENABLED, false);
+
+    try {
+      String producerFactoryClassName = serverProperties.getString(
+          PUB_SUB_PRODUCER_ADAPTER_FACTORY_CLASS,
+          "com.linkedin.venice.pubsub.adapter.kafka.producer.ApacheKafkaProducerAdapterFactory");
```
FelixGV, line 650: Unless you plan to completely exclude the Kafka dependency from the DVC module (is that the case?) then we should refer to class names programmatically rather than hard coding them, e.g. `ApacheKafkaProducerAdapterFactory.class.getName()`

FelixGV (commented):
Just a minor comment.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/config/VeniceServerConfig.java
haoxu07: Yeah, we should not hardcode them.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/config/VeniceServerConfig.java
```diff
@@ -632,6 +646,23 @@ public VeniceServerConfig(VeniceProperties serverProperties, Map<String, Map<Str
     ingestionMemoryLimit = extractIngestionMemoryLimit(serverProperties, ingestionMode, forkedProcessJvmArgList);
     LOGGER.info("Ingestion memory limit: {} after subtracting other usages", ingestionMemoryLimit);
     ingestionMlockEnabled = serverProperties.getBoolean(INGESTION_MLOCK_ENABLED, false);
+
+    try {
+      String producerFactoryClassName = serverProperties
```
sushantmane, line 651: Any particular reason to pass client factories as configs instead of actual instance of `PubSubClientsFactory` dependency to `venice-server` and `venice-controller`?

haoxu07: The reason is: for `DavinciBackend` and `IngestionIsolationServer`, the configuration information is most from `VeniceServerConfig`. As for `PubSubClientsFactory`, it is not meaningful class name we can put into the configs as string, so I need this concrete facotroies name. This reminds me one thing, for integration test, we also need to add this into configs to be passed to `IngestionIsolationServer` or `DavinciBackend`.

sushantmane: For II and DVC we can pass them as configs but for controller and server, I don't think we need to pass them as configs.

sushantmane: > For II and DVC we can pass them as configs but for controller and server, I don't think we need to pass them as configs.

This will mean we'll have two diff ways to pass pus-sub adapters which may not be ideal. Maybe it's better to just use a config-based approach.

haoxu07: Yeah, this change supports a config-based approach. Inside `VeniceControllerWrapper`  and `VeniceServerWrapper` class, there are some logics to get adapter factory class from `PubSubClientsFactory` and put them into configs. The benefits will be: 1) we will not need to add logic inside our internal project to prepare `PubSubClientsFactory` , 2)For II and DVC, they can leverage these configs naturally.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceController.java
```diff
@@ -128,7 +119,10 @@ public VeniceController(VeniceControllerContext ctx) {
     this.routerClientConfig = Optional.ofNullable(ctx.getRouterClientConfig());
     this.icProvider = Optional.ofNullable(ctx.getIcProvider());
     this.externalSupersetSchemaGenerator = Optional.ofNullable(ctx.getExternalSupersetSchemaGenerator());
-    this.pubSubClientsFactory = Objects.requireNonNull(ctx.getPubSubClientsFactory(), "PubSubClientsFactory is null");
+    this.pubSubClientsFactory = new PubSubClientsFactory(
+        multiClusterConfigs.getPubSubProducerAdapterFactory(),
+        multiClusterConfigs.getPubSubConsumerAdapterFactory(),
+        multiClusterConfigs.getPubSubAdminAdapterFactory());
     createServices();
```
sushantmane, line 126: I think we should just expose `PubSubClientsFactory getPubSubClientsFactory()` in `VeniceControllerMultiClusterConfig` instead of constructing `PubSubClientsFactory` here. Same for server too.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerConfig.java
```diff
@@ -271,6 +280,12 @@ public class VeniceControllerConfig extends VeniceControllerClusterConfig {
 
   private final boolean systemSchemaInitializationAtStartTimeEnabled;
 
+  private final PubSubProducerAdapterFactory pubSubProducerAdapterFactory;
+
+  private final PubSubConsumerAdapterFactory pubSubConsumerAdapterFactory;
+
+  private final PubSubAdminAdapterFactory pubSubAdminAdapterFactory;
```
sushantmane, line 287: let's replaces this with single `PubSubClientsFactory``

services/venice-server/src/main/java/com/linkedin/venice/server/VeniceServer.java
```diff
@@ -182,7 +173,11 @@ public VeniceServer(VeniceServerContext ctx) throws VeniceException {
     this.metricsRepository = ctx.getMetricsRepository();
     this.icProvider = ctx.getIcProvider();
     this.serviceDiscoveryAnnouncers = ctx.getServiceDiscoveryAnnouncers();
-    this.pubSubClientsFactory = Objects.requireNonNull(ctx.getPubSubClientsFactory(), "PubSubClientsFactory is null");
+    VeniceServerConfig veniceServerConfig = ctx.getVeniceConfigLoader().getVeniceServerConfig();
+    this.pubSubClientsFactory = new PubSubClientsFactory(
+        veniceServerConfig.getPubSubProducerAdapterFactory(),
+        veniceServerConfig.getPubSubConsumerAdapterFactory(),
+        veniceServerConfig.getPubSubAdminAdapterFactory());
     this.sslFactory = Optional.ofNullable(ctx.getSslFactory());
```
sushantmane, line 181: let's get `PubSubClientsFactory` from server config instead of individual factories

sushantmane (changes requested):
Overall lgtm. Requested few minor changes. Also please make sure that we are no longer using `new ApacheKafka*AdapterFactory()` anywhere in the code except for integration tests. Thanks!

sushantmane (dismissed):
LGTM. Thanks, @haoxu07!

sushantmane (dismissed):

sushantmane (approved):

PR #494, opened 2023-06-15
[pubsub][test] Add region name to PubSubBrokerWrapper

sushantmane:
## [pubsub][test] Add region name to PubSubBrokerWrapper

This commit introduces the region name property to the PubSubBrokerWrapper class.   

Other changes in this commit include:
- Relocating Kafka producer-specific constants to the `ApacheKafkaProducerConfig` class for better organization.
- Dynamically creating the producer based on the pubsub broker in use, replacing the previous hardcoded approach.


internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/integration/utils/PubSubBrokerWrapper.java
```diff
@@ -25,4 +25,8 @@ public String toString() {
   }
 
   public abstract PubSubClientsFactory getPubSubClientsFactory();
+
+  public abstract String getPubSubClusterName();
+
+  public abstract String getRegionName();
```
haoxu07, line 31: As we treat  `regionName` as one property of `PubSubBrokerWrapper` and we want to Client Adapter Factories to get this information when we create adapters. Do we plan to treat this as one field of properties passed into the client adapter factory `create` interface or you have other ideas?

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/integration/utils/VeniceClusterCreateOptions.java
```diff
@@ -380,8 +381,9 @@ private void verifyAndAddDefaults() {
       if (clusterName == null) {
         clusterName = Utils.getUniqueString("venice-cluster");
       }
-      if (standalone && regionName == null) {
-        regionName = "";
+      if (regionName == null || regionName.isEmpty()) {
+        regionName = STANDALONE_REGION_NAME;
+        // throw new IllegalArgumentException("regionName cannot be null or empty");
```
haoxu07, line 386: Nit: remove this comment line.

haoxu07 (changes requested):
Thanks for preparing this change! Just left comments for the first pass, besides could you add the reason about why we need to use regionName field into the PR description. I believe it is for specific pub sub systems FQN, feel free to correct me if I am wrong.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/integration/utils/PubSubBrokerWrapper.java
sushantmane: > Do we plan to treat this as one field of properties passed into the client adapter factory create interface or you have other ideas?

I think we should pass it via `VeniceProperties` option that we have.

haoxu07: So do you plan to include add the region name into the `VeniceProperties`  in this change?

sushantmane: No, it is not going to be part of this change. This change is to unblock integration testing.

sushantmane:
Thank you, @haoxu07 for the quick reviews! Appreciate it.

PR #495, opened 2023-06-15
Fix router quota calculation bug

majisourav99:
##  Fix router quota calculation bug

The quota calculation done in `ReadRequestThrottler` is diffrent from  `NoopRouterThrottler`. Since both were initialized in the router, later one wins during stat recording. This leads to discrepancies in actual router quota metric vs throttler quota. This PR lazily initializes the router throttler depending on the config (which should be enough to fix the bug) and extracted quota calculation into the base class interface. 
## How was this PR tested?

CI


services/venice-router/src/main/java/com/linkedin/venice/router/throttle/NoopRouterThrottler.java
```diff
@@ -5,25 +5,43 @@
...
 import com.linkedin.venice.meta.Store;
 import com.linkedin.venice.meta.StoreDataChangedListener;
+import com.linkedin.venice.router.VeniceRouterConfig;
 import com.linkedin.venice.router.stats.AggRouterHttpRequestStats;
 import java.util.List;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 
 public class NoopRouterThrottler
```
gaojieliu, line 15: This class seems not designed properly..
If we want to keep the same store-level quota as the regular `ReadRequestThrottler`, why not just introduces a feature flag in `ReadRequestThrottler` to indicate whether the throttler is a pass-through or not, and we don't need this `NoOpRouterThrottler` at all.
By eliminating this class, we can continue to construct a `ReadRequestThrottler` in constructor function and change the feature flag at runtime and this will help address the previous comment about potential leaking of ZK watches as well.

What do you think?

gaojieliu (changes requested):

services/venice-router/src/main/java/com/linkedin/venice/router/RouterServer.java
```diff
@@ -985,7 +970,22 @@ public VeniceRouterConfig getConfig() {
   }
 
   public void setReadRequestThrottling(boolean throttle) {
-    RouterThrottler throttler = throttle ? readRequestThrottler : noopRequestThrottler;
+    RouterThrottler throttler;
```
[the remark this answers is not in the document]
majisourav99: Only time it can be runtime is through admin command which I believe is very rare. Ideally it should be done through router config IMO

services/venice-router/src/main/java/com/linkedin/venice/router/throttle/NoopRouterThrottler.java
majisourav99: yeah I mentioned in approach in the slack. For noop throttler, reusing the read throttler code might be bit heavy when it really not need anything but emitting the metrics.

gaojieliu: That doesn't change the fact that  there could be leaking no matter the total number of times.

I think this throttler is mainly used by Dyno testing.

If it can simplify the structure significantly, I don't see much of overhead since `mayThrottleRead` is the only hot path.

majisourav99: yeah, code could be simpler. I will remove the noop class to consolidate into `ReadRequestThrottler`

FelixGV (commented):
Left a few comments.

services/venice-router/src/main/java/com/linkedin/venice/router/RouterServer.java
```diff
@@ -985,8 +978,8 @@ public VeniceRouterConfig getConfig() {
   }
 
   public void setReadRequestThrottling(boolean throttle) {
-    RouterThrottler throttler = throttle ? readRequestThrottler : noopRequestThrottler;
-    scatterGatherMode.initReadRequestThrottler(throttler);
+    readRequestThrottler.setIsNoopThrottlerEnabled(throttle);
+    scatterGatherMode.initReadRequestThrottler(readRequestThrottler);
```
gaojieliu, line 982: There is no need to call `scatterGatherMode.initReadRequestThrottler` here since the throttler instance never changes, right?

services/venice-router/src/main/java/com/linkedin/venice/router/throttle/ReadRequestThrottler.java
```diff
@@ -72,6 +72,8 @@ public class ReadRequestThrottler implements RouterThrottler, RoutersClusterMana
   private final long storageNodeQuotaCheckTimeWindow;
   private final boolean perStorageNodeThrottlerEnabled;
 
+  private boolean isNoopThrottlerEnabled;
```
gaojieliu, line 75: This flag can be changed in different thread, so shall we use `volatile` here?

gaojieliu (changes requested):

services/venice-router/src/main/java/com/linkedin/venice/router/RouterServer.java
majisourav99: Actually `initReadRequestThrottler` is never called anywhere else, so we could move the init call to `startService` and remove from here, or keep it as is.

services/venice-router/src/main/java/com/linkedin/venice/router/throttle/ReadRequestThrottler.java
majisourav99: Sure, good safety

services/venice-router/src/main/java/com/linkedin/venice/router/RouterServer.java
majisourav99: probably need to add some test here,  it should be `readRequestThrottler.setIsNoopThrottlerEnabled(!throttle);` but surprised no test caught it.

services/venice-router/src/main/java/com/linkedin/venice/router/RouterServer.java
```diff
@@ -985,7 +970,22 @@ public VeniceRouterConfig getConfig() {
...
+          routerStats.getStatsByType(RequestType.SINGLE_GET),
+          config);
+      ;
+    } else {
+      throttler = new NoopRouterThrottler(
+          routersClusterManager,
+          metadataRepository,
+          routerStats.getStatsByType(RequestType.SINGLE_GET),
+          config);
+    }
```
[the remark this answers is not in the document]
majisourav99: I removed `NoopRouterThrottler` class altogether.

services/venice-router/src/main/java/com/linkedin/venice/router/throttle/ReadRequestThrottler.java
```diff
@@ -161,44 +170,15 @@ protected long calculateStoreQuotaPerRouter(long storeQuota) {
...
 
     if (routerCount <= 0) {
-      LOGGER.error("Could not find any live router to serve traffic.");
-    }
-
-    long idealStoreQuotaPerRouter = routerCount > 0
-        ? Math.max(storeQuota / routerCount, 5) // Do not make quota to be 0 when storeQuota < routerCount
-        : 0;
-
-    if (!zkRoutersManager.isMaxCapacityProtectionEnabled() || idealTotalQuotaPerRouter <= maxRouterReadCapacity) {
```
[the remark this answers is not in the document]
majisourav99: Again code is updated to undo the above changes.

gaojieliu (approved):

PR #497, opened 2023-06-15
[controller][da-vinci] Updated PushStatusCollector to allow more chances to poll DaVinci status and fixed several DVC memory limiter issues

gaojieliu:
    This code change makes the following changes:
    1. Keeping polling DaVinci status if it is empty before encountering the following conditions:
      a. DaVinci status becomes non-empty.
      b. Server ingestion status becomes terminal: Complete or Error.
    2. Updated the default DaVinci empty status retry to be 6 (3 mins).
    
    The reason to have more empty retry is as follows:
    1. When firstly create the version, DaVinci may not send ingestion status quickly enough before
       PushStatusCollector polls the push status system store, which might return empty result, which would lead
       Controller think that all DaVinci instances have completed the ingestion.
    2. Right now, there are two logic, which will trigger empty DaVinci status retry to minimize the risk, but it is
       still a best effort:
       a. Before DaVinci status report becomes non-empty and Server reports terminal status, DaVinci status polling
          will be retried.
       b. If Server finishes the ingestion too quick and DaVinci ingestion hasn't started yet, and DaVinci status
          polling will retry up to the configured number.
    
    This code change also fixed several DVC memory limiter related issues:
    1. Fixed the memory limit calculation in VeniceServerConfig.
    2. Make sure `VersionBackend.delete` will delete all the partitions even some of them can't be opened because of memory restriction.
    3. Fixed an issue that Isolated Process won't reset ingestion offset when reporting completed or error.
    4. Added two new metrics for DaVinci memory limiter:
       a. RocksDBMemoryStats--memory_limit.Gauge : Memory Limit
       b. RocksDBMemoryStats--memory_usage.Gauge : Memory Usage
    5. DaVinciBackend won't open the hosted data partitions when II is enabled:
       a. It is a waste of effort to open and close them when II is enabled.
       b. SSTFileManager can't handle reopen well in the main process if the database changes in between. For example, if II process deletes
          some sst files because of compaction, and these deleted files will still be tracked by the SSTFileManager when the main process reopens it.
          Hope we can fix this behavior inside RocksDB in the future.

## How was this PR tested?
unit test

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
@@ -160,7 +160,22 @@ public class DaVinciBackend implements Closeable {
       IsolatedIngestionUtils.destroyLingeringIsolatedIngestionProcess(configLoader);
       /**
        * The constructor of {@link #storageService} will take care of unused store/store version cleanup.
+       *
+       * When Ingestion Isolation is enabled, we don't want to restore data partitions here:
+       * 1. It is a waste of effort since all the opened partitioned will be closed right after.
+       * 2. When DaVinci memory limiter is enabled, currently, SSTFileManager doesn't clean up the entries belonging
+       *    to the closed database, which means when the Isolated process hands the database back to the main process,
+       *    some removed SST files (some SST files can be removed in Isolated Process because of log compaction) will
+       *    remain in SSTFileManager tracked file list.
+       * 3. We still want to open metadata partition, otherwise {@link StorageService} won't scan the local db folder.
+       *    Also opening metadata partition in main process won't cause much side effect from DaVinci memory limiter's
+       *    POV, since metadata partition won't be handed back to main process in the future.
+       * 4. When Ingestion Isolation is enabled with suppressing live update feature, main process needs to open all the
+       *    data partitions since Isolated Process won't re-ingest the existing partitions.
        */
+      boolean whetherToRestoreDataPartitions = !isIsolatedIngestion()
+          || configLoader.getVeniceServerConfig().freezeIngestionIfReadyToServeOrLocalDataExists();
+      LOGGER.info("DaVinci {} restore data partitions.", whetherToRestoreDataPartitions ? "will" : "won't");
       storageService = new StorageService(
           configLoader,
           aggVersionedStorageEngineStats,
@@ -168,7 +183,7 @@ public class DaVinciBackend implements Closeable {
           storeVersionStateSerializer,
           partitionStateSerializer,
           storeRepository,
-          true,
+          whetherToRestoreDataPartitions,
           true,
           functionToCheckWhetherStorageEngineShouldBeKeptOrNot(managedClients));
       storageService.start();
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/storage/StorageService.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/storage/StorageService.java
@@ -383,6 +383,16 @@ public class StorageService extends AbstractVeniceService {
     factory.removeStorageEngine(storageEngine);
   }
 
+  /**
+   * This function is used to forcely clean up all the databases belonging to {@param kafkaTopic}.
+   * This function will only be used when the {@link #removeStorageEngine(String)} function can't
+   * handle some edge case, such as some partitions are lingering, which are not visible to the corresponding
+   * {@link AbstractStorageEngine}
+   */
+  public synchronized void forceStorageEngineCleanup(String kafkaTopic) {
+    persistenceTypeToStorageEngineFactoryMap.values().forEach(factory -> factory.removeStorageEngine(kafkaTopic));
+  }
+
   public synchronized void closeStorageEngine(String kafkaTopic) {
     AbstractStorageEngine<?> storageEngine = getStorageEngineRepository().removeLocalStorageEngine(kafkaTopic);
     if (storageEngine == null) {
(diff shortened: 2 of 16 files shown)


internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/integration/utils/VeniceControllerWrapper.java
```diff
@@ -206,6 +206,7 @@ static StatefulServiceProvider<VeniceControllerWrapper> generateService(VeniceCo
             .put(DAVINCI_PUSH_STATUS_SCAN_INTERVAL_IN_SECONDS, 5)
             .put(CONCURRENT_INIT_ROUTINES_ENABLED, true)
             .put(CLUSTER_DISCOVERY_D2_SERVICE, VeniceRouterWrapper.CLUSTER_DISCOVERY_D2_SERVICE_NAME)
+            .put(DAVINCI_PUSH_STATUS_SCAN_INTERVAL_IN_SECONDS, 1)
```
majisourav99, line 209: repeated ? update the previous entry

services/venice-controller/src/main/java/com/linkedin/venice/pushmonitor/PushStatusCollector.java
```diff
@@ -150,6 +151,7 @@ private void scanDaVinciPushStatus() {
       }
       ExecutionStatusWithDetails daVinciStatus = pushStatus.getDaVinciStatus();
       if (daVinciStatus.isNoDaVinciStatusReport()) {
+        LOGGER.info("Received empty DaVinci status report for topic: {}", pushStatus.topicName);
```
majisourav99, line 154: what if server ingested too quickly before the dvc poll and by next poll DVC also finished and since it stopped emitting heartbeat, then it can return empty report ?

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/integration/utils/VeniceControllerWrapper.java
gaojieliu: Good catch, I should remove the previous one.

services/venice-controller/src/main/java/com/linkedin/venice/pushmonitor/PushStatusCollector.java
gaojieliu: If there are some staled instance report by checking HB timestamp, these instances' status will be ignored, and another config will limit the maximum number of retries and the job will succeed after retries.

majisourav99 (approved):
LGTM

clients/da-vinci-client/src/main/java/com/linkedin/davinci/storage/StorageService.java
```diff
@@ -383,6 +383,16 @@ public synchronized void removeStorageEngine(String kafkaTopic) {
...
   }
 
+  /**
+   * This function is used to forcely clean up all the databases belonging to {@param kafkaTopic}.
+   * This function will only be used when the {@link #removeStorageEngine(String)} function can't
+   * handle some edge case, such as some partitions are lingering, which are not visible to the corresponding
+   * {@link AbstractStorageEngine}
+   */
+  public synchronized void forceStorageEngineCleanup(String kafkaTopic) {
+    persistenceTypeToStorageEngineFactoryMap.values().forEach(factory -> factory.removeStorageEngine(kafkaTopic));
```
sixpluszero, line 393: Do you think it is better to merge this call into removeStorageEngine directly?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerConfig.java
```diff
@@ -438,7 +438,7 @@ public VeniceControllerConfig(VeniceProperties props) {
     this.daVinciPushStatusScanIntervalInSeconds = props.getInt(DAVINCI_PUSH_STATUS_SCAN_INTERVAL_IN_SECONDS, 30);
     this.daVinciPushStatusScanThreadNumber = props.getInt(DAVINCI_PUSH_STATUS_SCAN_THREAD_NUMBER, 4);
     this.daVinciPushStatusScanNoReportRetryMaxAttempt =
-        props.getInt(DAVINCI_PUSH_STATUS_SCAN_NO_REPORT_RETRY_MAX_ATTEMPTS, 0);
+        props.getInt(DAVINCI_PUSH_STATUS_SCAN_NO_REPORT_RETRY_MAX_ATTEMPTS, 6);
```
sixpluszero, line 441: Yes good catch, also found this in heap dump in controller. But why 6? Looks like a specific number to me.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
```diff
@@ -160,15 +160,30 @@ public DaVinciBackend(
...
+      LOGGER.info("DaVinci {} restore data partitions.", whetherToRestoreDataPartitions ? "will" : "won't");
       storageService = new StorageService(
           configLoader,
           aggVersionedStorageEngineStats,
           rocksDBMemoryStats,
           storeVersionStateSerializer,
           partitionStateSerializer,
           storeRepository,
-          true,
+          whetherToRestoreDataPartitions,
```
sixpluszero, line 186: I guess I get what you mean here, also there is another logic at the end of the bootstrap trying to close metadata partition or metadata + data partition.
btw if we don't restore data partition, how can we drop storage engine cleanly? Is it done by forcibly drop whole folder?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/storage/StorageService.java
gaojieliu: No.
For DaVinci use case, it is expected to have some leaked partitions, but for the backend use cases, that might indicate an issue.

And we can only invoke this function after the regular `removeStorageEngine` when II is enabled since both main the process and II process can host a subset of total partitions.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerConfig.java
gaojieliu: This is a random number and I was trying to increase the chance of polling DaVinci status in the worst sceanrio, and 6 * 30s = 3 mins, so at most the batch push job can be delayed by 3 mins if there is no DaVinci push status report.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
gaojieliu: If there is no dvc memory limiter, the data partition will be either opened by the II process or by the main process, and the forcibly drop the database is mainly for the partitions, which are not opened by any process.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerConfig.java
sixpluszero: OK I see.

sixpluszero (approved):
Thanks for the change! LGTM now.

PR #499, opened 2023-06-15
[common] [server] Inlucdes TOPIC_SWITCH_RECEIVED and END_OF_PUSH_RECEI…

lluwm:
…VED during old replica status removal

When updating the statusHistory for a given host on ZK, we take the chance to remove the old replica status when the list length is greater than MAX_HISTORY_LENGTH. However, today, this removal is only applied to COMPLETED or IncrementalPushStatus in the list. In real environment, we saw that TOPIC_SWITCH_RECEIVED and END_OF_PUSH_RECEIVED could also have a very long history, e.g. when Helix rebalances a partition back and forth on a particular host, thus causing the Znode size to increase beyond 1MB for example.

To fix the issue, this change improves the replica status removal logic by also considering and checking EOP and TS updates, and removes them when MAX_HISTORY_LENGTH is violated.

## How was this PR tested?
Tested in the EI environment.

### DIFF
--- a/internal/venice-common/src/main/java/com/linkedin/venice/pushmonitor/ReplicaStatus.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/pushmonitor/ReplicaStatus.java
@@ -131,11 +131,19 @@ public class ReplicaStatus {
       return;
     }
 
-    long completeCount = statusHistory.stream().filter(status -> status.getStatus() == COMPLETED).count();
-    long incPushCount = statusHistory.stream().filter(status -> isIncrementalPushStatus(status.getStatus())).count();
-    long topicSwitchCount =
-        statusHistory.stream().filter(status -> status.getStatus() == TOPIC_SWITCH_RECEIVED).count();
-    long endOfPushCount = statusHistory.stream().filter(status -> status.getStatus() == END_OF_PUSH_RECEIVED).count();
+    long completeCount = 0, incPushCount = 0, topicSwitchCount = 0, endOfPushCount = 0;
+    for (StatusSnapshot snapshot: statusHistory) {
+      ExecutionStatus status = snapshot.getStatus();
+      if (status == COMPLETED) {
+        completeCount++;
+      } else if (isIncrementalPushStatus(status)) {
+        incPushCount++;
+      } else if (status == TOPIC_SWITCH_RECEIVED) {
+        topicSwitchCount++;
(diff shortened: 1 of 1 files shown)


internal/venice-common/src/main/java/com/linkedin/venice/pushmonitor/ReplicaStatus.java
```diff
@@ -133,31 +133,55 @@ private void removeOldStatuses() {
 
     long completeCount = statusHistory.stream().filter(status -> status.getStatus() == COMPLETED).count();
     long incPushCount = statusHistory.stream().filter(status -> isIncrementalPushStatus(status.getStatus())).count();
-    Iterator<StatusSnapshot> itr = statusHistory.iterator();
+    long topicSwitchCount =
+        statusHistory.stream().filter(status -> status.getStatus() == TOPIC_SWITCH_RECEIVED).count();
+    long endOfPushCount = statusHistory.stream().filter(status -> status.getStatus() == END_OF_PUSH_RECEIVED).count();
```
adamxchen, line 138: Can we only scan once to get the count so it improves the performance here?
For example
```
         List<Status> statuses = Arrays.asList(Status.OK, Status.NOT_OK, Status.UNKNOWN, Status.OK, Status.NOT_OK);

        Map<Status, Long> statusCounts = statuses.stream()
                .collect(Collectors.groupingBy(status -> status, Collectors.counting()));

        statusCounts.forEach((status, count) -> System.out.println(status + ": " + count));

```

adamxchen (commented):
Thanks for the fix. Do you have ballpark number of how much size this fix can save us?

internal/venice-common/src/main/java/com/linkedin/venice/pushmonitor/ReplicaStatus.java
lluwm: Can it work easily for `isIncrementalPushStatus`?

sixpluszero: Inc push count will then be checking SOIP count + EOIP count

lluwm: Sure, change to use a single loop to count the numbers.

adamxchen (approved):
LGTM!

lluwm:
Thanks for the review @adamxchen! Proceed to checkin.

PR #500, opened 2023-06-15
[Not For Merge][debug] Add logging for Router unhealthy request during cluster expansion

sixpluszero:

Add logging when retry is aborted due to slow route and try to print out original failure trigger reason.

### DIFF
--- a/internal/alpini/router/alpini-router-base/src/main/java/com/linkedin/alpini/router/ScatterGatherRequestHandlerImpl.java
+++ b/internal/alpini/router/alpini-router-base/src/main/java/com/linkedin/alpini/router/ScatterGatherRequestHandlerImpl.java
@@ -679,7 +679,7 @@ public abstract class ScatterGatherRequestHandlerImpl<H, P extends ResourcePath<
         boolean isRequestRetriable = isRequestRetriable(path, role, retryStatus);
 
         // Indicate the current request is a retry request
-        path.setRetryRequest(retryFuture.isSuccess() ? null : retryFuture.getCause());
+        path.setRetryRequest((HttpResponseStatus) retryStatus);
 
         HostHealthMonitor<H> healthMonitor =
             _scatterGatherHelper.isEnableRetryRequestAlwaysUseADifferentHost() ? (H host, String partName) ->
--- a/internal/alpini/router/alpini-router-base/src/main/java/com/linkedin/alpini/router/ScatterGatherRequestHandlerImpl.java
+++ b/internal/alpini/router/alpini-router-base/src/main/java/com/linkedin/alpini/router/ScatterGatherRequestHandlerImpl.java
@@ -679,7 +679,7 @@ public abstract class ScatterGatherRequestHandlerImpl<H, P extends ResourcePath<
         boolean isRequestRetriable = isRequestRetriable(path, role, retryStatus);
 
         // Indicate the current request is a retry request
-        path.setRetryRequest();
+        path.setRetryRequest(retryFuture.isSuccess() ? null : retryFuture.getCause());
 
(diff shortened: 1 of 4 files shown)


internal/alpini/router/alpini-router-base/src/main/java/com/linkedin/alpini/router/ScatterGatherRequestHandlerImpl.java
```diff
@@ -679,7 +679,7 @@ private void prepareRetry(AsyncFuture<HRS> retryFuture) {
         boolean isRequestRetriable = isRequestRetriable(path, role, retryStatus);
 
         // Indicate the current request is a retry request
-        path.setRetryRequest();
+        path.setRetryRequest(retryFuture.isSuccess() ? null : retryFuture.getCause());
```
gaojieliu, line 682: I think we can use `retryStatus` directly here:
```
path.setRetryRequest(retryStatus)
```

services/venice-router/src/main/java/com/linkedin/venice/router/api/VeniceDelegateMode.java
```diff
@@ -329,6 +332,22 @@ private <H> H selectLeastLoadedHost(List<H> hosts, VenicePath path) throws Route
     }
     if (minHost == null) {
       if (path.isRetryRequest()) {
+        List<String> replicaHostIds = hosts.stream().map(x -> ((Instance) x).getNodeId()).collect(Collectors.toList());
+        LogManager.getLogger()
```
gaojieliu, line 336: Why not create a static `Logger` in this class?

services/venice-router/src/main/java/com/linkedin/venice/router/api/VeniceDelegateMode.java
```diff
@@ -329,6 +332,22 @@ private <H> H selectLeastLoadedHost(List<H> hosts, VenicePath path) throws Route
...
+            .error(
+                "VENG-10611-Debug Retry Request Aborted. Store Name: {}, Resource Name: {} ReplicaHostIds: {}, VenicePath SlowNodeSet: {}",
+                path.getStoreName(),
+                path.getResourceName(),
+                replicaHostIds,
+                path.getSlowStorageNodeSet());
+        if (path.getRetryFutureThrowable() != null) {
+          LogManager.getLogger()
```
gaojieliu, line 344: Can we print the retry reason here instead of in this function?
https://github.com/linkedin/venice/blob/main/services/venice-router/src/main/java/com/linkedin/venice/router/api/VeniceDelegateMode.java#L225

services/venice-router/src/main/java/com/linkedin/venice/router/api/VeniceDelegateMode.java
sixpluszero: Either way is fine, I can change that.

internal/alpini/router/alpini-router-base/src/main/java/com/linkedin/alpini/router/ScatterGatherRequestHandlerImpl.java
sixpluszero: Do you know what exactly the type retryStatus is it is saying HRS but that's generics I believe.

services/venice-router/src/main/java/com/linkedin/venice/router/api/VeniceDelegateMode.java
sixpluszero: Hmm but this will print out on every retry request? Instead of those who failed eventually due to slow route reason we are seeing in the incident?

PR #501, opened 2023-06-15
[router] Make key/value profiling sensor per router only

majisourav99:
## Make key/value profiling sensor per router only

Currently is KV profiling is enabled it generates key value fine grained profiling stats for each store which creates many metrics (num_stores X 20 X 2 X 2). This PR makes KV profiling metrics emit per router total only. 
## How was this PR tested?

GHCI

   Added a new metric named `value_size_in_byte` but kept the per store response_size metrics intact.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -4,5 +4,11 @@ org.gradle.caching=true
 pulsarGroup = org.apache.pulsar
 pulsarVersion = 2.10.3
 
-
+org.gradle.jvmargs= \
+--add-exports=java.base/sun.nio.ch=ALL-UNNAMED \
+--add-opens=java.base/java.lang=ALL-UNNAMED \
+--add-opens=java.base/java.lang.reflect=ALL-UNNAMED \
+--add-opens=java.base/java.io=ALL-UNNAMED \
+--add-opens=java.base/sun.net=ALL-UNNAMED \
+--add-exports=jdk.unsupported/sun.misc=ALL-UNNAMED
 
--- a/gradle.properties
+++ b/gradle.properties
@@ -4,11 +4,4 @@ org.gradle.caching=true
 pulsarGroup = org.apache.pulsar
 pulsarVersion = 2.10.3
 
-org.gradle.jvmargs= \
---add-exports=java.base/sun.nio.ch=ALL-UNNAMED \
---add-opens=java.base/java.lang=ALL-UNNAMED \
---add-opens=java.base/java.lang.reflect=ALL-UNNAMED \
---add-opens=java.base/java.io=ALL-UNNAMED \
---add-opens=java.base/sun.net=ALL-UNNAMED \
---add-exports=jdk.unsupported/sun.misc=ALL-UNNAMED
 
--- a/gradle.properties
+++ b/gradle.properties
@@ -4,4 +4,11 @@ org.gradle.caching=true
 pulsarGroup = org.apache.pulsar
 pulsarVersion = 2.10.3
 
+org.gradle.jvmargs= \
+--add-exports=java.base/sun.nio.ch=ALL-UNNAMED \
+--add-opens=java.base/java.lang=ALL-UNNAMED \
+--add-opens=java.base/java.lang.reflect=ALL-UNNAMED \
+--add-opens=java.base/java.io=ALL-UNNAMED \
+--add-opens=java.base/sun.net=ALL-UNNAMED \
+--add-exports=jdk.unsupported/sun.misc=ALL-UNNAMED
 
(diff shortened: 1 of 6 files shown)


gaojieliu:
Have you checked with Venice SRE to see whether this is what we want?
Do we need the store-level profiling for some specific stores in some occasions?

I was wondering whether this feature should become a store-level feature or not.

majisourav99:
> 

This is the from SRE request, they want total metric only, not per store.

services/venice-router/src/main/java/com/linkedin/venice/router/stats/RouterHttpRequestStats.java
```diff
@@ -156,19 +162,20 @@ public RouterHttpRequestStats(
           new Max(),
           TehutiUtils.getFineGrainedPercentileStat(getName(), getFullMetricName(keySizeSensorName)));
 
-      responseSizeSensor = registerSensor(
-          responseSizeSensorName,
+      String valueSizeSensorName = "value_size_in_byte";
+      valueSizeSensor = registerSensor(
```
gaojieliu, line 166: This looks weird..

Can we get rid of the new sensor but keep using the existing `responseSizeSensor`?
Essentially, when `isKeyValueProfilingEnabled` is `true`, we only use `FineGrainedPercentileStat` for total stats.
It can be done like this:

```
if (isKeyValueProfilingEnabled && storeName.equals(AbstractVeniceAggStats.STORE_NAME_FOR_TOTAL_STAT)) {
      String keySizeSensorName = "key_size_in_byte";
      keySizeSensor = registerSensor(
          keySizeSensorName,
          new Avg(),
          new Max(),
          TehutiUtils.getFineGrainedPercentileStat(getName(), getFullMetricName(keySizeSensorName)));

      responseSizeSensor = registerSensor(
          responseSizeSensorName,
          new Avg(),
          new Max(),
          TehutiUtils.getFineGrainedPercentileStat(getName(), getFullMetricName(responseSizeSensorName)));
    } else {
      responseSizeSensor = registerSensor(
          responseSizeSensorName,
          new Avg(),
          new Max(),
          TehutiUtils.getPercentileStat(getName(), getFullMetricName(responseSizeSensorName)));
    }

```

services/venice-router/src/main/java/com/linkedin/venice/router/stats/RouterHttpRequestStats.java
```diff
@@ -55,6 +55,9 @@ public class RouterHttpRequestStats extends AbstractVeniceHttpStats {
   private final Sensor readQuotaUsageSensor;
   private final Sensor inFlightRequestSensor;
   private Sensor keySizeSensor;
+
```
gaojieliu, line 58: Also `keySizeSensor` will be `null` today if `isKeyValueProfilingEnabled` is `disabled`..
I wonder why today, there is no NPE thrown when calling function: `recordKeySizeInByte`, which doesn't do a null check..

gaojieliu (changes requested):

gradle.properties
```diff
@@ -4,11 +4,5 @@ org.gradle.caching=true
 pulsarGroup = org.apache.pulsar
 pulsarVersion = 2.10.3
 
-org.gradle.jvmargs= \
```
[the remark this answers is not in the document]
majisourav99: yes, its by mistake. my intellij fails if I have them. I will remove it

services/venice-router/src/main/java/com/linkedin/venice/router/stats/RouterHttpRequestStats.java
majisourav99: there is a null check before calling the key size record.

gaojieliu: Can you share the code link?

majisourav99: https://github.com/linkedin/venice/blob/8e0f0ecd0f4c69dbedcd74edc3adb7b1ebef486c/services/venice-router/src/main/java/com/linkedin/venice/router/api/VenicePathParser.java#L146

gaojieliu: I see.
The check in `VenicePathParser` is not great, and we should do a `null` check in `RouterHttpRequestStats` in function of `recordKeySizeInByte`.

gaojieliu (approved):

PR #502, opened 2023-06-16
[controller] Fail pushes due to maintenance mode from target colo push only

majisourav99:
## Fail pushjobs during target colo push only due to maintenance mode.

Restrict fail pushes due to maintenance mode from target colo push only. Otherwise during some outrages too many jobs would fail and create bad user experience.

## How was this PR tested?

CI


lluwm (approved):
LGTM!

huangminchn (approved):
Thanks a lot Sourav!

PR #503, opened 2023-06-20
[server] Fix bug in A/A partial update Map field merge update timestamp tie braking logic bug

sixpluszero:
## [server] Fix bug in A/A partial update Map field merge update timestamp tie braking logic bug

This PR fixes the bug in A/A partial update map collection merge tie breaking logic.

**Problem:**
Existing logic tries to get the map schema and then try to compare the value of two active keys that have same TS.
However, there are issues: We are actually not comparing the map item value but instead we are putting KeyValPair (which contains key and value) to the comparator and in certain case it will lead to exception and SIT ingestion failure. If the map value type is primitive (int/string or so) it will be fine as the logic in comparator won't cast the value, however if it is record, then the logic will try to cast the value and throw.

**Change in this PR:**
Instead of comparing value, let's only compare KV's key string to break tie. This should be more performant considering (1) we don't need to extract map value schema (2) Usually value type is complex record.
In this PR, I also change the logic to add a new private method for sorting logic in map only to differentiate with list type. This method will be applied to both added key-value pairs and deleted keys.

## How was this PR tested?

Added new integration test

### DIFF
--- a/internal/venice-client-common/src/main/java/com/linkedin/venice/schema/merge/SortBasedCollectionFieldOpHandler.java
+++ b/internal/venice-client-common/src/main/java/com/linkedin/venice/schema/merge/SortBasedCollectionFieldOpHandler.java
@@ -15,6 +15,7 @@ import java.util.List;
 import java.util.ListIterator;
 import java.util.Map;
 import java.util.Set;
+import java.util.function.Function;
 import java.util.function.Supplier;
 import javax.annotation.concurrent.ThreadSafe;
 import org.apache.avro.Schema;
@@ -701,6 +702,19 @@ public class SortBasedCollectionFieldOpHandler extends CollectionFieldOperationH
     });
   }
 
+  private void sortElementAndTimestampListInMap(
+      List<ElementAndTimestamp> elementAndTsList,
+      Function<Object, String> keyExtractor) {
+    Comparator<Object> mapKeyComparator = Comparator.comparing(x -> (String) x);
+    elementAndTsList.sort((o1, o2) -> {
+      final int timestampCompareResult = Long.compare(o1.getTimestamp(), o2.getTimestamp());
+      if (timestampCompareResult == 0) {
+        return mapKeyComparator.compare(keyExtractor.apply(o1.getElement()), keyExtractor.apply(o2.getElement()));
+      }
+      return timestampCompareResult;
+    });
+  }
+
   private void removeIntersectionElements(Set<Object> set1, Set<Object> set2) {
     List<Object> intersectionElements = new ArrayList<>();
     for (Object element: set1) {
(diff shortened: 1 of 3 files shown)


internal/venice-client-common/src/main/java/com/linkedin/venice/schema/merge/SortBasedCollectionFieldOpHandler.java
```diff
@@ -701,6 +702,19 @@ private void sortElementAndTimestampList(
     });
   }
 
+  private void sortElementAndTimestampListInMap(
+      List<ElementAndTimestamp> elementAndTsList,
+      Function<Object, String> keyExtractor) {
+    Comparator<Object> mapKeyComparator = Comparator.comparing(x -> (String) x);
+    elementAndTsList.sort((o1, o2) -> {
```
ZacAttack, line 709: I'm not sure I totally get this.  We're only comparing the keys in the map for the conflict resolution, but if theres a conflict, isn it not a given the keys will be the same?

FelixGV: I agree, this seems confusing... shouldn't it be the value which is the tie breaker, rather than the key?

sixpluszero: Hmm what you are talking about is when comparing old / new value for a same key for one update operation. Here it is different case, basically after we handle all these conflicts for the same keys and we need to sort the active elements by TS. Now two KV pairs can have the same TS inside map data structure and we need to break tie here. So for old logic we were using value to break tie, but I think using key is more lightweight.

sixpluszero: Basically like this:
After all the conflict resolution, the map could have
{k1=v1, k4=v4, k3=v3, k2=v2, }
put only part length = 1
activeElementsTimestamps = [3, 3, 2]

The last step of the merge update operation is to sort it out so the active elements are sorted by TS. we expect [2, 3, 3] but tie breaking can be either by key or value.

ZacAttack: ohhhh ok.  We're talking about tie breaking for the sake of SORTING, not conflict resolution?

ZacAttack: Ok.  synced on chat.  I totally misunderstood what was going on here.  Using key should be fine.  This is ONLY for sorting the result, not conflict resolution.

ZacAttack (approved):
thanks for the fix!

internal/venice-client-common/src/main/java/com/linkedin/venice/schema/merge/SortBasedCollectionFieldOpHandler.java
FelixGV: Thanks for clarifying. We should write the DCR spec in the open-source doc.

PR #504, opened 2023-06-21
[controller] Make sure push status handlers check OfflinePushStatus before action

sixpluszero:
## [controller] Make sure push status handlers check OfflinePushStatus before action

If the OfflinePushStatus was dropped right before handling COMPLETED/ERROR report, it will throw NPE in scanner's periodic scanning logic and lead to push status stuck. This PR fixes issue by checking if retrieved OfflinePushStatus is null before action. Also it adds a try catch in scanning logic to make sure if there is any exception happening in one topic, it will not break scanner.


gaojieliu (approved):
Thanks for the quick fix.

PR #505, opened 2023-06-21
[samza][vpj][controller] Make SharedObjectFactory threadsafe

nisargthakkar:
## Make `SharedObjectFactory` threadsafe

Currently, `SharedObjectFactory` is not thread-safe and if there are multiple threads trying to create an object that previously didn't exist, then there can be multiple objects created. While this may be okay in most cases, the issue occurs when releasing these objects, the objects do not have the correct refcount. Hence, there will be an `Reference counter underflow` error. This commit attempts to fix this issue by wrapping all operations using the synchronization offered by `ConcurrentHashMap#compute`

## How was this PR tested?

Added a unit test

### DIFF
--- a/internal/venice-common/src/main/java/com/linkedin/venice/utils/SharedObjectFactory.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/utils/SharedObjectFactory.java
@@ -3,6 +3,8 @@ package com.linkedin.venice.utils;
 import com.linkedin.venice.utils.concurrent.VeniceConcurrentHashMap;
 import java.util.Map;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.Consumer;
 import java.util.function.Supplier;
 
@@ -14,6 +16,11 @@ import java.util.function.Supplier;
  */
 public class SharedObjectFactory<T> {
   private final Map<String, ReferenceCounted<T>> OBJECT_MAP = new VeniceConcurrentHashMap<>();
+  private final Map<String, ReentrantLock> LOCK_MAP = new VeniceConcurrentHashMap<>();
+
+  private Lock getLock(String identifier) {
+    return LOCK_MAP.computeIfAbsent(identifier, id -> new ReentrantLock());
+  }
 
   /**
    * Get a shared object that has the specified {@param identifier}. If an object with the {@param identifier} doesn't
@@ -25,7 +32,8 @@ public class SharedObjectFactory<T> {
    * @return A shared object
    */
   public T get(String identifier, Supplier<T> constructor, Consumer<T> destroyer) {
-    synchronized (OBJECT_MAP) {
+    getLock(identifier).lock();
+    try {
       final AtomicBoolean newObjectCreated = new AtomicBoolean(false);
       ReferenceCounted<T> referenceCounted = OBJECT_MAP.computeIfAbsent(identifier, id -> {
         newObjectCreated.set(true);
@@ -38,6 +46,8 @@ public class SharedObjectFactory<T> {
       }
 
       return referenceCounted.get();
+    } finally {
+      getLock(identifier).unlock();
     }
   }
 
@@ -49,7 +59,8 @@ public class SharedObjectFactory<T> {
    * @return {@code true} if the shared object is no longer being used; {@code false} otherwise
    */
   public boolean release(String identifier) {
-    synchronized (OBJECT_MAP) {
+    getLock(identifier).lock();
+    try {
       ReferenceCounted<T> referenceCounted = OBJECT_MAP.get(identifier);
       if (referenceCounted != null) {
         referenceCounted.release(); // Also removes the object from the objectMap
@@ -60,6 +71,8 @@ public class SharedObjectFactory<T> {
           return false;
         }
       }
+    } finally {
+      getLock(identifier).unlock();
     }
     return true;
   }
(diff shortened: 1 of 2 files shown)


internal/venice-common/src/main/java/com/linkedin/venice/utils/SharedObjectFactory.java
```diff
@@ -24,17 +32,22 @@ public class SharedObjectFactory<T> {
...
+        return new ReferenceCounted<>(object, destroyer);
+      });
+
+      if (!newObjectCreated.get()) {
+        referenceCounted.retain();
+      }
+
       return referenceCounted.get();
+    } finally {
+      getLock(identifier).unlock();
```
FelixGV, line 50: Why lookup the lock twice? Can we look it up once at the beginning of the function, store it in a variable, and keep using it?

internal/venice-common/src/main/java/com/linkedin/venice/utils/SharedObjectFactory.java
```diff
@@ -69,10 +92,15 @@ public int size() {
...
-      return 0;
-    } else {
-      return objectMap.get(identifier).getReferenceCount();
+    getLock(identifier).lock();
+    try {
+      if (!OBJECT_MAP.containsKey(identifier)) {
+        return 0;
+      } else {
+        return OBJECT_MAP.get(identifier).getReferenceCount();
+      }
```
FelixGV, line 101: Can we replace by a `get` followed by a null check, rather than `containsKey` followed by `get`?

internal/venice-common/src/main/java/com/linkedin/venice/utils/SharedObjectFactory.java
```diff
@@ -12,7 +15,12 @@
  * @param <T> Class whose objects need to release resources cleanly.
  */
 public class SharedObjectFactory<T> {
-  private final Map<String, ReferenceCounted<T>> objectMap = new VeniceConcurrentHashMap<>();
+  private final Map<String, ReferenceCounted<T>> OBJECT_MAP = new VeniceConcurrentHashMap<>();
+  private final Map<String, ReentrantLock> LOCK_MAP = new VeniceConcurrentHashMap<>();
```
FelixGV, line 19: Nitpick: I think the Java convention is to put in upper case only the `static final` state, not the instance-level state?

internal/venice-common/src/main/java/com/linkedin/venice/utils/SharedObjectFactory.java
```diff
@@ -24,17 +32,23 @@ public class SharedObjectFactory<T> {
...
+    lock.lock();
+    try {
+      final AtomicBoolean newObjectCreated = new AtomicBoolean(false);
+      ReferenceCounted<T> referenceCounted = objectMap.computeIfAbsent(identifier, id -> {
+        newObjectCreated.set(true);
+        T object = constructor.get();
+        return new ReferenceCounted<>(object, destroyer);
+      });
+
+      if (!newObjectCreated.get()) {
```
majisourav99, line 45: can we check the reference count of the object to find if its newly created or not?

gaojieliu:
Can we use `ConcurrentHashMap#compute` to improve the concurrency?
Based on the javadoc, the offered compute function will be executed atomically:
```
    /**
     * Attempts to compute a mapping for the specified key and its
     * current mapped value (or {@code null} if there is no current
     * mapping). The entire method invocation is performed atomically.
     * Some attempted update operations on this map by other threads
     * may be blocked while computation is in progress, so the
     * computation should be short and simple, and must not attempt to
     * update any other mappings of this Map.
     *
     * @param key key with which the specified value is to be associated
     * @param remappingFunction the function to compute a value
     * @return the new value associated with the specified key, or null if none
     * @throws NullPointerException if the specified key or remappingFunction
     *         is null
     * @throws IllegalStateException if the computation detectably
     *         attempts a recursive update to this map that would
     *         otherwise never complete
     * @throws RuntimeException or Error if the remappingFunction does so,
     *         in which case the mapping is unchanged
     */
    public V compute(K key,
                     BiFunction<? super K, ? super V, ? extends V> remappingFunction) {
```

gaojieliu:
My bad..
Just realized that this PR is using a lock per `identifier`...
Will it create too many locks if there are a lot of objects in this factory?

internal/venice-common/src/main/java/com/linkedin/venice/utils/SharedObjectFactory.java
nisargthakkar: Yeah, I think it works now that I added the lock around this. Previously I hadn't added the lock and needed this. I prefer the solution that Gaojie suggested. If that doesn't work, I'll remove this `AtomicBoolean`

nisargthakkar:
> Can we use ConcurrentHashMap#compute to improve the concurrency?
This is a great suggestion. I will try it out

> Will it create too many locks if there are a lot of objects in this factory?
Yes, there will be 1 lock per identifier and it will not get cleaned up too. So, maybe the `.compute` option will be a better option

internal/venice-common/src/main/java/com/linkedin/venice/utils/SharedObjectFactory.java
nisargthakkar: With `compute`, I was able to get rid of this too

internal/venice-common/src/main/java/com/linkedin/venice/utils/SharedObjectFactory.java
```diff
@@ -69,10 +72,12 @@ public int size() {
    * managing an object with the identifier, returns 0
    */
   public int getReferenceCount(String identifier) {
-    if (!objectMap.containsKey(identifier)) {
+    // It is possible that this method is not thread-safe, but since this is only used in tests currently, it is okay
```
gaojieliu, line 75: This should be fine even it is used by the regular code path since it won't throw CME regardless of whether the compute is being executed or not.

gaojieliu (approved):

majisourav99 (approved):
LGTM

internal/venice-common/src/main/java/com/linkedin/venice/utils/SharedObjectFactory.java
nisargthakkar: Yeah, it won't give CME, but the issue is that if the upstream code is doing some operation based on this reference count and the reference counts are continuously moving, then there would be jumps in the ref counts. It is non-idiomatic to do this kind of operation, though

PR #510, opened 2023-06-26
[server] RocksDB cleanup on bad store/store versions

majisourav99:
##  RocksDB cleanup on bad store/store versions

If server rocksDB directories have resources which are bad and cannot be opened by rocksdb during restart, they are not loaded into storage engine and none of the background resource cleaner kicks for those directories. This leads to massive disk space waste. This PR if rocksDB open fails, it cleans up unused resources.

## How was this PR tested?

GHCI

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -726,6 +726,7 @@ ext.createDiffFile = { ->
         ':!services/venice-standalone/*', // exclude the entire standalone project
         ':!clients/venice-client/src/main/java/com/linkedin/venice/fastclient/factory/ClientFactory.java',
         ':!clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java',
+        ':!clients/da-vinci-client/src/main/java/com/linkedin/davinci/storage/StorageService.java',
         ':!clients/venice-producer/src/main/java/com/linkedin/venice/producer/online/OnlineProducerFactory.java',
         ':!clients/venice-producer/src/main/java/com/linkedin/venice/producer/online/ProducerTool.java',
 
--- a/build.gradle
+++ b/build.gradle
@@ -726,7 +726,6 @@ ext.createDiffFile = { ->
         ':!services/venice-standalone/*', // exclude the entire standalone project
         ':!clients/venice-client/src/main/java/com/linkedin/venice/fastclient/factory/ClientFactory.java',
         ':!clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java',
-        ':!clients/da-vinci-client/src/main/java/com/linkedin/davinci/storage/StorageService.java',
         ':!clients/venice-producer/src/main/java/com/linkedin/venice/producer/online/OnlineProducerFactory.java',
         ':!clients/venice-producer/src/main/java/com/linkedin/venice/producer/online/ProducerTool.java',
 
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/storage/StorageService.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/storage/StorageService.java
@@ -191,7 +191,7 @@ public class StorageService extends AbstractVeniceService {
     }
 
     int versionNumber = Version.parseVersionFromKafkaTopicName(storageEngineName);
-    return store.getVersion(versionNumber).isPresent() || versionNumber < store.getCurrentVersion();
+    return !store.getVersion(versionNumber).isPresent() || versionNumber < store.getCurrentVersion();
   }
 
   private void restoreAllStores(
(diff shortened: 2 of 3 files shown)


build.gradle
```diff
@@ -726,6 +726,7 @@ ext.createDiffFile = { ->
         ':!services/venice-standalone/*', // exclude the entire standalone project
         ':!clients/venice-client/src/main/java/com/linkedin/venice/fastclient/factory/ClientFactory.java',
         ':!clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java',
+        ':!clients/da-vinci-client/src/main/java/com/linkedin/davinci/storage/StorageService.java',
```
gaojieliu, line 729: Why not write some unit test against this class?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/storage/StorageService.java
```diff
@@ -178,6 +179,21 @@ private void initInternalStorageEngineFactories() {
...
+    try {
+      store = storeRepository.getStoreOrThrow(storeName);
+    } catch (VeniceNoStoreException e) {
+      // The store does not exist in Venice anymore, so it will be deleted.
+      LOGGER.warn("Store does not exist, will delete invalid local version: {}", storageEngineName);
+      return true;
+    }
+
+    int versionNumber = Version.parseVersionFromKafkaTopicName(storageEngineName);
+    return !store.getVersion(versionNumber).isPresent() || versionNumber < store.getCurrentVersion();
```
gaojieliu, line 194: So if we remove the backup version, it means Server will re-bootstrap the deleted backup version, right?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/storage/StorageService.java
```diff
@@ -207,6 +223,10 @@ private void restoreAllStores(
             if (ExceptionUtils.recursiveClassEquals(e, RocksDBException.class)) {
               LOGGER.error("Could not load the following store : " + storeName, e);
               aggVersionedStorageEngineStats.recordRocksDBOpenFailure(storeName);
```
gaojieliu, line 225: I think we should only record this metric if the Server receives an exception and can't delete it.
And we might need to put up an alert against it since it is critical.
But the storage engine, which can be deleted, I guess a warn log should be good enough.

gaojieliu (changes requested):

clients/da-vinci-client/src/main/java/com/linkedin/davinci/storage/StorageService.java
majisourav99: yes, but that could help in successful re-ingestion, as this case is only for error path.

build.gradle
majisourav99: sure

majisourav99: actually its bit complex throw the rocksDBException as the rocksdb engine is opened several level under storage service and hard to mock it.

adamxchen: Can we just mock the behavior of `openStore` so we don't need to mock rocksDB's behavior?

majisourav99: We would need to add a surgery method for `persistenceTypeToStorageEngineFactoryMap` to put in a mocked version of `RocksDBStorageEngineFactory` and store engine map to return mock of `RocksDBStorageEngine` which is doable, but will add too many test only design in the main class.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/storage/StorageService.java
```diff
@@ -178,7 +179,21 @@ private void initInternalStorageEngineFactories() {
     persistenceTypeToStorageEngineFactoryMap.put(BLACK_HOLE, new BlackHoleStorageEngineFactory());
   }
 
-  private void restoreAllStores(
+  private boolean deleteStorageEngine(String storageEngineName) {
```
gaojieliu, line 182: For unit test, I think you can turn this function into a `static` method, which will take two params:
```
readOnlyStoreRepository
storageEngineName
```
And you can write unit tests against it.

Also, I think we better to rename it to be something like: `whetherToDeleteStorageEngineWhenEncounterError` to make it less scary.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/storage/StorageService.java
```diff
@@ -205,10 +220,15 @@ private void restoreAllStores(
...
-              LOGGER.error("Could not load the following store : " + storeName, e);
-              aggVersionedStorageEngineStats.recordRocksDBOpenFailure(storeName);
+              LOGGER.warn("Encountered RocksDB error while opening store: {}", storeName, e);
+              if (deleteStorageEngine(storeName)) {
+                LOGGER.info("Store {} does not exist, will delete it.", storeName);
+                factory.removeStorageEngine(storeName);
+              }
               continue;
             }
+            LOGGER.error("Could not load the following store : " + storeName, e);
```
gaojieliu, line 230: Maybe we should consolidate this logic into the above function to make it more testable.

Such as renaming the above method to be: `cleanupStorageEngineWhenEncounteringError` and that static function can take three params:
```
Exception
readOnlyStoreRepository
storageEngineName
```
So you can test almost all the new logic.

majisourav99: good idea!  Let me add some unit test

gaojieliu (approved):

PR #511, opened 2023-06-26
[controller]  Do not lookup RF of existing topics when creating a new topic as it may not be accurate

sushantmane:
##  Do not lookup the replication factor of existing topics when creating a new topic as it may not be accurate
The way we get the replication factor for VTs is not reliable since `partitionInfo.replicas().length != RF` in some  
cases (e.g., offline replicas,  partition reassignment). Besides Kafka doesn't offer an API to get RF at the moment.   
Hence use `kafka.replication.factor` config value as it is a safer option. 

### DIFF
--- a/internal/venice-test-common/src/main/java/com/linkedin/venice/unit/kafka/MockInMemoryAdminAdapter.java
+++ b/internal/venice-test-common/src/main/java/com/linkedin/venice/unit/kafka/MockInMemoryAdminAdapter.java
@@ -52,7 +52,7 @@ public class MockInMemoryAdminAdapter implements PubSubAdminAdapter {
     topicPubSubTopicConfigurationMap.put(topic, topicPubSubTopicConfiguration);
     topicPartitionNumMap.put(topic, new ArrayList<>());
     for (int i = 0; i < numPartitions; i++) {
-      topicPartitionNumMap.get(topic).add(new PubSubTopicPartitionInfo(topic, i, replication, true));
+      topicPartitionNumMap.get(topic).add(new PubSubTopicPartitionInfo(topic, i, true));
     }
   }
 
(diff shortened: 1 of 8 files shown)


internal/venice-test-common/src/main/java/com/linkedin/venice/unit/kafka/MockInMemoryAdminAdapter.java
```diff
@@ -52,7 +52,7 @@ public void createTopic(
     topicPubSubTopicConfigurationMap.put(topic, topicPubSubTopicConfiguration);
     topicPartitionNumMap.put(topic, new ArrayList<>());
     for (int i = 0; i < numPartitions; i++) {
-      topicPartitionNumMap.get(topic).add(new PubSubTopicPartitionInfo(topic, i, replication, true));
```
nisargthakkar, line 55: Can we remove the `replication` arg from the function and everywhere else if it is not being used?

services/venice-controller/src/main/java/com/linkedin/venice/ingestion/control/RealTimeTopicSwitcher.java
```diff
@@ -47,6 +47,7 @@ public class RealTimeTopicSwitcher {
   private final VeniceWriterFactory veniceWriterFactory;
   private final Time timer;
   private final Integer kafkaReplicationFactorForRTTopics;
+  private final Integer kafkaReplicationFactor;
```
nisargthakkar, line 50: Can this be `int` instead of `Integer`? Can you update the one above too if that can be changed?

nisargthakkar (commented):
Mostly LGTM. Thanks Sushant

services/venice-controller/src/main/java/com/linkedin/venice/ingestion/control/RealTimeTopicSwitcher.java
sushantmane: done

haoxu07 (dismissed):

nisargthakkar (approved):
LGTM

sushantmane:
Thank you @nisargthakkar & @haoxu07 for the review!

PR #512, opened 2023-06-26
[vpj] Force collect at least 1 sample per input file for zstd dict

m-nagarajan:
## Summary
In VPJ, `PushJobZstdConfig` calculates number of bytes to be collected per input file baed on the configurable sample size(default `200MB`). If that ends up being less than size of first value in file, no samples get collected from that file potentially leading to skipping zstd dictionary creation. Changing this to add an extra check to add at least 1 sample per input file(until reaching 200MB).

## How was this PR tested?
github ci

### DIFF
--- a/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/InputDataInfoProvider.java
+++ b/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/InputDataInfoProvider.java
@@ -101,13 +101,16 @@ public interface InputDataInfoProvider extends Closeable {
         continue;
       }
 
-      if (fileSampleSize + data.length > pushJobZstdConfig.getMaxBytesPerFile()) {
-        String perFileLimitErrorMsg = String.format(
-            "Read %s to build dictionary. Reached limit per file of %s.",
-            ByteUtils.generateHumanReadableByteCountString(fileSampleSize),
-            ByteUtils.generateHumanReadableByteCountString(pushJobZstdConfig.getMaxBytesPerFile()));
-        LOGGER.debug(perFileLimitErrorMsg);
-        return;
+      // At least 1 sample per file should be added until the max sample size is reached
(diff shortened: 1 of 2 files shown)


clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/InputDataInfoProvider.java
```diff
@@ -101,13 +101,16 @@ static void loadZstdTrainingSamples(AbstractVeniceRecordReader recordReader, Pus
...
-            "Read %s to build dictionary. Reached limit per file of %s.",
-            ByteUtils.generateHumanReadableByteCountString(fileSampleSize),
-            ByteUtils.generateHumanReadableByteCountString(pushJobZstdConfig.getMaxBytesPerFile()));
-        LOGGER.debug(perFileLimitErrorMsg);
-        return;
+      // At least 1 sample per file should be added until the max sample size is reached
+      if (fileSampleSize > 0) {
+        if (fileSampleSize + data.length > pushJobZstdConfig.getMaxBytesPerFile()) {
+          String perFileLimitErrorMsg = String.format(
+              "Read %s to build dictionary. Reached limit per file of %s.",
```
nisargthakkar, line 108: nit: Since we're already touching this piece, can you replace the `String.format` with log interpolation in the `LOGGER.debug` call itself? Also, can you make the same changes for L119?
```
"Read %s to build dictionary. Reached sample limit of %s."
```

nisargthakkar (approved):
LGTM

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/InputDataInfoProvider.java
m-nagarajan: Modified, thanks for noticing this.

nisargthakkar (approved):
LGTM

PR #513, opened 2023-06-26
[VIP-2] Removing Per Record Offset Metadata From Venice-Server Storage With Heartbeats

ZacAttack:
## [VIP-2] Removing Per Record Offset Metadata From Venice-Server Storage With Heartbeats

This VIP explores a strategy for removing the offset metadata stored per record in Venice by utilizing replica heartbeats.

### DIFF
--- a/docs/proposals/VIP-2.md
+++ b/docs/proposals/VIP-2.md
@@ -1,0 +1,117 @@
+---
+layout: default
+title: VIP-1
+parent: Community Guides
+permalink: /docs/proposals/VIP_TEMPLATE.md
+---
+
+# [VIP-2] Removing Per Record Offset Metadata From Venice-Server Storage
+
+* **Status**: Discussion
+* **Author(s)**: Zac Policzer
+* **Pull Request**: _TODO_
+* **Release**: _TODO_
+
+
+
+## Introduction
+
+This VIP explores strategies for removing the offset metadata stored per record by utilizing heartbeats as well as an exploration of what other system benefits can be had by adopting replica heartbeats.
+
+## Problem Statement
+
+Today, Venice stores which have Active/Active replication enabled store N number of offsets per record where N is the number of replicating sites. This is done so as to enable data drift detection via a Venice algorithm called Leap Frog (which we will document in a later section) so that it's possible to detect and flag inconsistencies triggered by outages and bugs with very high granularity.
+
+In addition, these offsets are utilized by Venice Change Capture consumers for filtering out duplicated events across version pushes, so that downstream consumers do not get callbacks triggered for events that they already consumed.
+
+This was a fine approach, however, it presumes that for a given message in an underlying pubsub implementation, that the object used to describe it's position (in Kafka's case, a long) is reasonably small. So long as it's small, the added overhead per record is tolerable.
+
+This is not neceassrily the case. **In at least one PubSub implementation that Venice will need to support, the position object is as large as over 20 Bytes.**
+
+Venice (at least in LinkedIn deploments at time of writing) is a storage bound service, and this high overhead has real cost implications.
+
+## Scope
+
+_Call out explicitly:_
+
+1. **What is in scope?**
+
+   We must be able to shrink down or remove this per record overhead in the venice server, but still be able to retain the functionality that we depend on (diff detection and change capture client filtering).
+
+2. **What is out of scope?**
+
+    We will not seek to reduce the footprint of this data in the underlying PubSub (where it's presumed that disk is cheaper).
+
+## Project Justification
+
+This project seeks to improve the **performance, scalability, and cost to serve of Venice**. By shrinking the per record metadata we'll not only save on disk, but we should be able to improve the ingestion speed slightly (less bytes is less bytes afterall) and reduce the potential increase in storage cost as new replicating sites (henceforth refered to as **colos**) are added.
+
+## Proposed Design
+
+Before we can discuss proposed designs, it's important to clear up some aspects of per record pubsub positions (henceforth referred to as **offsets**) which have heretofore not been documented in Venice open source (but for which there are some mentions of in the code).
+
+### How Are Per Record Offsets Used Today?
+* #### **Leap Frog**
+
+* #### **Venice Change Capture**
+
+_This section must describe the proposed design and implementation in detail, and how it
+addresses the problems outlined in the problem statement. It must provide details on alternative solutions
+that were considered, stating clearly why they lost out to the final design. This latter point is particularly
+important or else we tend to vacillate on old decisions because we cannot remember the original rationale.
+Phrase this section in terms of trade offs since it is rare that one approach is better in all ways; there is
+generally a tradeoff and making that explicit makes it much easier to understand._
+
+_The following aspects of the design must be covered when applicable:_
+
+1. _Changes to APIs or protocols or data format highlighting any compatibility issues and how they will be addressed._
+2. _Major components and sequence diagrams between them._
+3. _Multi-region implications including Parent/Child Controller communication._
+4. _Alternative designs considered, and brief explanation of tradeoffs and decision versus the chosen design,
+and important aspects to consider_:
+   1. Security.
+   2. Performance.
+   3. Maintainability.
+   4. Operability.
+   5. Scalability.
+   6. Compatibility.
+   7. Testability.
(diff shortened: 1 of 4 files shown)


docs/proposals/VIP-2.md
```diff
@@ -0,0 +1,240 @@
...
+>
+>Record 4: {20, _, _}
+
+Originally, we only needed the highwatermark of coloA and the rmd of an individual record in coloB (and vice versa going the other way). However, in this context, we need more information. Each record will only have the offset of the LAST upsream RT message which touched the row.  We can backfill a bit more information based on the last heartbeat. Since everything is produced in serial in increasing offsets we know that the other updates came from an offset previous to the published highwatermark in the last received heartbeat.  So, this looks like:
+
+>**COLO A**
+>
+>HEARTBEAT: {3, 199, 4500}
+>                           
+>Record 1: {<3, 200, <4500}
```
haoxu07, line 187: For the backfill information like `<3`, does that mean we can always gurantee between HEARTBEAT and Record 1, there is no update from other regions. Only region #2 with offset 200 mute that key?

docs/proposals/VIP-2.md
```diff
@@ -0,0 +1,240 @@
...
+If you squint at both of these, they boil down to relying on the following properties.
+
+* Being able to build a high watermark based on events consumed
+* Being able to causally compare a consumed event/record to this highwatermark and take action.
+
+We're able to meet the first requirement by just starting at an arbitrary point in a version topic stream and consuming.  Every single event has an RMD field that can help a consumer of this data build a highwatermark.  And then since every single record we consume has the same information, it's very straightforward to determine the causality of that event relative to a built up high watermark.
+
+That said, it's not actually a requirement to be able to do this on every single event we consume. It's possible to meet the first two requirements at a courser granularity of updates.
+
+### Heartbeat Algorithm
```
haoxu07, line 83: Not sure my understanding is correct or not, just want to confirm: If we have 3 regions, the Heartbeat algorithm can save RMD space from storing 3 regions's offset to 1 region's offset?

ZacAttack: You are correct, but we'll actually store none of the regions offset metadata in rocksdb.  We'll only persist it to PubSubBroker as envelope metadata.  None of which has to go into rocksdb.

docs/proposals/VIP-2.md
ZacAttack: I'll include a case here for the two updates as it's an important one. We have two choices, we can either rely on the heartbeat + previous events to build a running highwatermark from which we can apply this backfill, or, we can compact key updates which occur between two heartbeats.  

My intuition is that key updates are the way to go because if we apply an increasingly growing highwatermark into the offsets of individual records, then we narrow the common window between two colos. I've got a spec half done which models this, I'll post back here once I've determined conclusively that this intuition is right.  But it's a good call out because with two to the same key within the heartbeat interval then backfilling with the less then up to the last heartbeat is an incorrect generalization, and will lead to false positives in some simple cases.

docs/proposals/VIP-2.md
```diff
@@ -0,0 +1,240 @@
+---
+layout: default
+title: [VIP-2] Removing Per Record Offset Metadata From Venice-Server Storage With Heartbeats
+parent: Community Guides
+permalink: /docs/proposals/VIP_TEMPLATE.md
```
ziyaddin, line 5: For generating a reachable URL:

```suggestion
permalink: /docs/proposals/vip-2
```

ziyaddin: Please also update the Proposals [table](https://github.com/ZacAttack/venice/blob/VIP-2/docs/proposals/proposals.md).

haoxu07:
Hi, @ZacAttack , currently we are trying to get back the consistency checker, so I reopened this PR to discuss with you about the proposal.  I remember this proposal's original intention is:  to use heartbeat message to carry a vector of information to track the progress of each region of real-time topic, then we could avoid  the offset vector for all regions previously we had for every record. 

Assume my understanding is correct, so what will be inside the vector like {3, 199, 4500} below is each region's position(offset) or just timestamp? I still think it might be position/offset, but either way we need is sth comparable in one region to achieve comparsion between record and high watermark. 

>HEARTBEAT: {3, 199, 4500}
>                           
>Record 1: {_, 200, _}
>
>Record 2: {_, 201, _}
>
>Record 3: {5, _, _}
>
>Record 4: {_, _, 5001}


Please correct me if I am wrong.

ZacAttack:
> Hi, @ZacAttack , currently we are trying to get back the consistency checker, so I reopened this PR to discuss with you about the proposal. I remember this proposal's original intention is: to use heartbeat message to carry a vector of information to track the progress of each region of real-time topic, then we could avoid the offset vector for all regions previously we had for every record.
> 
> Assume my understanding is correct, so what will be inside the vector like {3, 199, 4500} below is each region's position(offset) or just timestamp? I still think it might be position/offset, but either way we need is sth comparable in one region to achieve comparsion between record and high watermark.
> 
> > HEARTBEAT: {3, 199, 4500}
> > Record 1: {_, 200, _}
> > Record 2: {_, 201, _}
> > Record 3: {5, _, _}
> > Record 4: {_, _, 5001}
> 
> Please correct me if I am wrong.

Yeah, so your only requirement is something which gives you the ability to causally relate two events on this stream. Upstream offset was the original goal, timestamp can work so long as it's a broker side one (the source is RT's which don't have the guarantee of a single writer per topic partition, so producer timestamps from RT's are out of the question).  Honestly timestamp can be tricky in some scenarios, so offset is vastly preferable, but I'm not sure how practical that is these days in NG.

kvargha:
@haoxu07 Is it possible to support consistency checker without offset vector comparison? I was under the impression we won't be supporting offset vectors anymore due to NG.

cc: @sushantmane

ZacAttack:
> @haoxu07 Is it possible to support consistency checker without offset vector comparison? I was under the impression we won't be supporting that anymore due to NG.
> 
> cc: @sushantmane

It should be possible with heartbeats.  If you think about a heartbeat, it has a single writer from a colo for a specific topic partition.  Abstractly thinking about it, let's say from a single colo you have h1 event1, event2, event3, h2, event4

The offset vector for those events from that topic partition can be inferred as:

```
<<event1, <_, _, h1>>
<<event2, <_,_,h1>>
<<event3, <_,_,h1>>
<<event4, <_,_,h2>>
```

ZacAttack:
> > @haoxu07 Is it possible to support consistency checker without offset vector comparison? I was under the impression we won't be supporting that anymore due to NG.
> > cc: @sushantmane
> 
> It should be possible with heartbeats. If you think about a heartbeat, it has a single writer from a colo for a specific topic partition. Abstractly thinking about it, let's say from a single colo you have h1 event1, event2, event3, h2, event4
> 
> The offset vector for those events from that topic partition can be inferred as:
> 
> ```
> <<event1, <_, _, h1>>
> <<event2, <_,_,h1>>
> <<event3, <_,_,h1>>
> <<event4, <_,_,h2>>
> ```

worth calling out though.  It'll potentially be important to make sure we control the compaction interval.  Post compaction, you'll only have the LAST heartbeat.

haoxu07:
> > > @haoxu07 Is it possible to support consistency checker without offset vector comparison? I was under the impression we won't be supporting that anymore due to NG.
> > > cc: @sushantmane
> > 
> > 
> > It should be possible with heartbeats. If you think about a heartbeat, it has a single writer from a colo for a specific topic partition. Abstractly thinking about it, let's say from a single colo you have h1 event1, event2, event3, h2, event4
> > The offset vector for those events from that topic partition can be inferred as:
> > ```
> > <<event1, <_, _, h1>>
> > <<event2, <_,_,h1>>
> > <<event3, <_,_,h1>>
> > <<event4, <_,_,h2>>
> > ```
> 
> worth calling out though. It'll potentially be important to make sure we control the compaction interval. Post compaction, you'll only have the LAST heartbeat.

IIUC, with your example: h1 event1, event2, event3, h2, event4, you suggest to use h1 to represent the one region's progress for all the records between h1 and h2. It is right, if we gurantee h1, h2... h4 is growing, but the gap between h1 and h2 is minute level, so we lose the ability to catch issues sub second level, that is my major concern for Lilypad on top of leap frog.

PR #514, opened 2023-06-26
[vpj][samza][controller][tc][fc] Increase timeout for D2Client startup

nisargthakkar:
## Increase timeout for D2Client startup

Currently, when creating a D2 client via `D2ClientFactory`, the code only waits for 5 seconds. This commit increases the timeout to 60 seconds.

## How was this PR tested?

GH CI

### DIFF
--- a/internal/venice-common/src/main/java/com/linkedin/venice/d2/D2ClientFactory.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/d2/D2ClientFactory.java
@@ -41,7 +41,6 @@ public class D2ClientFactory {
           .setSSLContext(sslFactoryOptional.map(SSLFactory::getSSLContext).orElse(null))
           .setIsSSLEnabled(sslFactoryOptional.isPresent())
           .setSSLParameters(sslFactoryOptional.map(SSLFactory::getSSLParameters).orElse(null))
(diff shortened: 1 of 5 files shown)


internal/venice-common/src/main/java/com/linkedin/venice/d2/D2ClientFactory.java
```diff
@@ -41,6 +41,7 @@ public static D2Client getD2Client(String d2ZkHost, Optional<SSLFactory> sslFact
           .setSSLContext(sslFactoryOptional.map(SSLFactory::getSSLContext).orElse(null))
           .setIsSSLEnabled(sslFactoryOptional.isPresent())
           .setSSLParameters(sslFactoryOptional.map(SSLFactory::getSSLParameters).orElse(null))
+          .setRetry(true)
```
gaojieliu, line 44: Based on the javadoc, this `retry` is used to retry the d2 request instead of initialization.
Is it expected?

nisargthakkar: Removed this to see if just increasing the timeout helps

gaojieliu (approved):

PR #515, opened 2023-06-27
[controller][common] Use ExecutionStatusWithDetails class to replace existing Pair usage in controller

sixpluszero:
## [controller][common] Use ExecutionStatusWithDetails class to replace existing Pair usage in controller
This PR uses existing POJO class ExecutionStatusWithDetails to replace Pair<ExecutionStatus, Optional<String>> and Pair<ExecutionStatus, String> usage in controller push monitoring logics to improve readability 

## How was this PR tested?
Existing unit test.


internal/venice-common/src/main/java/com/linkedin/venice/pushmonitor/PushStatusDecider.java
```diff
@@ -28,15 +27,15 @@
...
+ * Decide the offline push status by checking all replicas statuses under different offline push strategies.
  */
 public abstract class PushStatusDecider {
   private final Logger logger = LogManager.getLogger(PushStatusDecider.class);
   private static final String REASON_NOT_IN_EV = "not yet in EXTERNALVIEW";
   private static final String REASON_NOT_ENOUGH_PARTITIONS_IN_EV = "not enough partitions in EXTERNALVIEW";
   private static final String REASON_UNDER_REPLICATED = "does not have enough replicas";
 
-  private static Map<OfflinePushStrategy, PushStatusDecider> decidersMap = new HashMap<>();
+  private static final Map<OfflinePushStrategy, PushStatusDecider> decidersMap = new HashMap<>();
```
FelixGV, line 38: I don't think this map is needed at all... We can add a `PushStatusDecider` property to the `OfflinePushStrategy` enum, and then get it from there directly, rather than from a map.

FelixGV (commented):
Thanks Jialin! Great to see this `Pair` cleanup! I left a minor comment.

internal/venice-common/src/main/java/com/linkedin/venice/pushmonitor/PushStatusDecider.java
sixpluszero: Thanks for pointing out! That's a good catch. Yeah actually I found that it should also not be hard to merge PushStatusDecider with its corresponding OfflinePushStatusStrategy as it has very tight mapping, and for the common publish static method in PushStatusDecider class it can go to util class for example PushStatusMonitorUtils.
For this PR I only removed the map as you suggested, I plan to make the change as a separate PR and then 3rd PR trying to consolidate push monitor classes and other changes.

services/venice-controller/src/test/java/com/linkedin/venice/pushmonitor/PartitionStatusBasedPushMonitorTest.java
```diff
@@ -162,9 +177,12 @@ public void testLoadRunningPushWhichIsNotUpdateToDateAndDeletionError() {
     Assert.assertEquals(getMonitor().getOfflinePushOrThrow(topic).getCurrentStatus(), ExecutionStatus.ERROR);
 
     // set the push status decider back
+    /*
     PushStatusDecider.updateDecider(
         OfflinePushStrategy.WAIT_N_MINUS_ONE_REPLCIA_PER_PARTITION,
         new WaitNMinusOnePushStatusDecider());
+    
+     */
```
FelixGV, line 185: Can we get rid of this commented code?

FelixGV (dismissed):
Thanks, looks great!

FelixGV (approved):

PR #516, opened 2023-06-27
[controller][test] Fix a race condition between kill job and version …

lluwm:
…swap

Today there's a race around kill job and push completion or version swap. This is because we only write the kill message and do nothing and rely on the servers to receive the kill, report ERROR to mark a version killed, and prevent it from becoming current/active. However, this is not always the case especially when the push strategy only requires n-1 replicas to be completed in every partition. This means if a kill is sent when a push is near completion or almost completed, it can still become ONLINE and current version after the kill was sent and recognized by some SNs. This is an issue because we now have a current version that is also having a kill message in the participant store. No re-balance or re-bootstrap can succeed for this store version. Current mitigation is to re-push or rollback to previous version when this race happens.

A prerequisite rb (https://github.com/linkedin/venice/pull/269) that introduces the new VersionStatus.KILLED has deployed in production allows us to continue with the following fix.

This fix includes:

1. On kill, we take a store lock (since current version swap will also be acquiring this lock) when checking if the version to-be-killed is still in a non-complete state. Update the version status on ZK to KILLED, if it is.
2. On version swap, we check if the version to which we are swapping has been killed already by checking the VersionStatus.
3. integration and unit tests have been added to verify the correctness of the above behavior.

## How was this PR tested?
Passed CI test.
Passed newly added integration tests and unit tests.


internal/venice-common/src/main/java/com/linkedin/venice/meta/AbstractStore.java
```diff
@@ -290,18 +290,19 @@ public List<Version> retrieveVersionsToDelete(int clusterNumVersionsToPreserve)
     List<Version> versionsToDelete = new ArrayList<>();
 
     /**
-     * The current version need not be the last largest version (eg we rolled back to a earlier version).
+     * The current version need not be the last largest version (e.g. we rolled back to an earlier version).
```
ZacAttack, line 293: thanks!

ZacAttack (approved):
Looks good, thanks!

sixpluszero (dismissed):
VersionStatus.canDelete() naming sounds weird, but that's fine. 
I think the change itself makes sense.

lluwm:
Thanks @ZacAttack and @sixpluszero for the review! I will fix the title when checking in. Have to do a minor update to pass all tests and it is updating a stale url in a docker file. Please approve it again, if it makes sense.

sixpluszero (approved):

PR #519, opened 2023-06-28
[vpj] minor logging improvements

majisourav99:
##  Minor logging improvements related to target region push

Added some log to indicate push completion to target regions. 

## How was this PR tested?

GHCI

### DIFF
--- a/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
+++ b/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
@@ -2744,7 +2744,7 @@ public class VenicePushJob implements AutoCloseable {
         if (isTargetedRegionPush) {
           LOGGER.info("Successfully pushed {} to targeted region {}", topicInfo.topic, targetedRegions);
         } else {
-          LOGGER.info("Successfully pushed {} to regions {}", topicInfo.topic, completedDatacenters);
+          LOGGER.info("Successfully pushed {} to all the regions", topicInfo.topic);
         }
         return;
       }
--- a/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
+++ b/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
@@ -1160,7 +1160,8 @@ public class VenicePushJob implements AutoCloseable {
               controllerClient,
               pushJobSetting,
               kafkaTopicInfo,
-              pushJobSetting.targetedRegions);
+              pushJobSetting.targetedRegions,
+              pushJobSetting.isTargetedRegionPushEnabled);
         }
 
         updatePushJobDetailsWithCheckpoint(PushJobCheckpoints.JOB_STATUS_POLLING_COMPLETED);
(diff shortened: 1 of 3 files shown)


clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
```diff
@@ -2733,7 +2739,11 @@ void pollStatusUntilComplete(
         }
 
         // Every known datacenter have successfully reported a completed status at least once.
-        LOGGER.info("Successfully pushed {}", topicInfo.topic);
+        if (isTargetedRegionPush) {
```
adamxchen, line 2742: Can we do one-liner for this?
```
LOGGER.info("Successfully pushed {} to regions {}", topicInfo.topic, isTargetedRegionPush ? targetedRegions : completedDatacenters);
```

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
```diff
@@ -2774,7 +2784,9 @@ private String printJobStatus(
 
     Optional<String> details = response.getOptionalStatusDetails();
     if (details.isPresent() && detailsAreDifferent(previousOverallDetails, details.get())) {
-      LOGGER.info("\t\tNew overall details: {}", details.get());
+      if (!details.get().isEmpty()) {
```
adamxchen, line 2787: nit: could be cleaner if we move this check to `detailsAreDifferent`. Feel free to drop

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
majisourav99: I thought about ternary operator, but it wont be show if its `targeted` or not. we would need another ternary for it.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
majisourav99: logic wont be same as in the next line it returns the `newOverallDetails`

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
adamxchen: We don't really need `targeted` IMO.  If we need to tell if that's a targeted push, we can have a complete log ahead of this line, e.g. `LOGGER.info("Starting to push to targeted region XXX`

majisourav99: yeah, but when it finished after say 1hr, there will be tons of other logs and you may not easily find the starting log

adamxchen: BTW, I think we don't need to pass in the boolean `isTargetedRegionPush`. Just need to check if the `targetedRegions` is empty or not. That would save your work on modifying the method signature in multiple places.

adamxchen: Also, can we just drop `targeted `? I think that's okay w/o it, since we have told what regions are pushed in the same log?

adamxchen: Synced offline. Resolve this comment now

adamxchen (approved):
Thank you!

PR #520, opened 2023-06-28
[admin-tool] Add support for '--cluster' parameter in data recovery t…

lluwm:
…ools

This changes adds support for taking '--cluster' (cluster name) as optional parameter to several data recovery tools. Because of this change, the '--stores' parameter is changed to optional. When both are present, '--stores' can
overwrite '--cluster' values.

## How was this PR tested?
Internal CI test.
newly added unit tests.

Run example with environment info masked.
```
lelu@lelu-mn1 venice % java -jar ./clients/venice-admin-tool/build/libs/venice-admin-tool-all.jar --estimate-data-recovery-time --url xxxxx --cluster xxxxx --dest-fabric xxxx
[WARN] Running admin tool without SSL.
...
2023-06-28 14:21:07 INFO [DataRecoveryEstimator] [main] [store: xxxxxx, recovery time estimate: 10:10:10]
2023-06-28 14:21:07 INFO [DataRecoveryEstimator] [main] [store: xxxxxx, recovery time estimate: 00:02:11]
2023-06-28 14:21:07 INFO [DataRecoveryWorker] [main] Total: 2, Succeeded: 2, Error: 0, Uncompleted: 0
"TOTAL RECOVERY TIME FOR ALL STORES = 10:12:21"
```
when overwrite happens:
```
lelu@lelu-mn1 venice % java -jar ./clients/venice-admin-tool/build/libs/venice-admin-tool-all.jar --estimate-data-recovery-time --url xxxxx --stores AAAA --cluster xxxxx --dest-fabric xxxx
[WARN] Running admin tool without SSL.
[WARN] --stores overwrites --cluster value.
...
2023-06-30 12:35:50 INFO [DataRecoveryEstimator] [main] [store: AAAA, recovery time estimate: 00:02:11]
2023-06-30 12:35:50 INFO [DataRecoveryWorker] [main] Total: 1, Succeeded: 1, Error: 0, Uncompleted: 0
"TOTAL RECOVERY TIME FOR ALL STORES = 00:02:11"
```


clients/venice-admin-tool/src/test/java/com/linkedin/venice/TestAdminTool.java
```diff
@@ -164,4 +165,33 @@ public void testAdminTopicIsAllowedByTopicConfigsRelatedApi() {
...
+    try {
+      AdminTool.main(estimateArgs);
+      AdminTool.main(estimateArgs2);
+      AdminTool.main(executeArgs);
+      AdminTool.main(monitorArgs);
+    } catch (VeniceClientException e) {
+      // Expected exception.
+    } catch (Exception err) {
+      Assert.fail("AdminTool should allow admin topic to be updated by config update API", err);
+    }
```
shuhui-liu, line 195: 1. When will VeniceClientException throw?
2. The assertion fail message does not sound right.

clients/venice-admin-tool/src/main/java/com/linkedin/venice/AdminTool.java
```diff
@@ -725,6 +732,26 @@ private static void monitorDataRecovery(CommandLine cmd) {
     }
   }
 
+  private static Set<String> calculateRecoveryStoreNames(String multiStores, String cluster, String parentUrl) {
+    if (!StringUtils.isEmpty(multiStores)) {
+      return Utils.parseCommaSeparatedStringToSet(multiStores);
```
shuhui-liu, line 737: Do you think it would be helpful to print a message here when both cluster and stores are present, mentioning that '--stores' will overwrite '--cluster' values, or mention that in the data recovery command help text? Because users might think stores in both stores list and cluster will be executed.

lluwm: Yes, I think it is a good point. I will do both, i.e. printing out a warning message and displaying it in the command help text. Btw, do you think the which behavior is more intuitive in this case,  `overwrite` or `append`?

clients/venice-admin-tool/src/test/java/com/linkedin/venice/TestAdminTool.java
lluwm: 1. This is from the new code that I added in this change.
```
    if (!StringUtils.isEmpty(cluster)) {
      try (ControllerClient parentCtrlCli = new ControllerClient(cluster, parentUrl, sslFactory)) {
        MultiStoreResponse response = parentCtrlCli.queryStoreList(false);
        if (response.isError()) {
          throw new VeniceClientException("Error in getting store list for " + cluster + ": " + response.getError());
        }
        if (response.getStores() == null) {
          return null;
        }
        return new HashSet<>(Arrays.asList(response.getStores()));
      }
    }
```
It would throw the following exception if I don't catch it, when running locally.
```
Caused by: com.linkedin.venice.client.exceptions.VeniceClientException: Error in getting store list for venice-1: An error occurred during controller request. controller = null, route = /list_stores, params = [cluster_name=venice-1, hostname=localhost, include_system_stores=false], timeout = 600000. Exception type: class com.linkedin.venice.exceptions.VeniceException. Detailed message: Unable to discover leader controller from [http://localhost:7036]
	at com.linkedin.venice.AdminTool.calculateRecoveryStoreNames(AdminTool.java:744)
	at com.linkedin.venice.AdminTool.estimateDataRecoveryTime(AdminTool.java:674)
	at com.linkedin.venice.AdminTool.main(AdminTool.java:518)
	at com.linkedin.venice.TestAdminTool.testAdminToolDataRecoveryApi(TestAdminTool.java:188)
```

2. Oops, yeah, it's a copy-and-paste error. Will fix it.

clients/venice-admin-tool/src/main/java/com/linkedin/venice/AdminTool.java
shuhui-liu: Personally I think the right behavior is to respect stores parameter as you do in this PR. Because store is a finer grained resource than cluster. Other admin commands also execute on the store, instead of the whole cluster, when both are provided. Printing a message is to make the behavior explicitly clear to users.

shuhui-liu (approved):

PR #521, opened 2023-06-28
[controller] register meta system store schema before initializing helix manager

shuhui-liu:
## Summary

Current behavior: After helix manager initialization, controller finds system schema cluster leader controller url from helix manager, build controller client using the url, register schemas via the controller client.

Problem: During deployment, when leader controller A stops, helix assigns controller B as leader. When A's controller service starts again, helix wants to hand over leadership to A so it transits B from leader to standby. Before transition completes, B is still the leader and all A's requests are sent to B. But since B clears in-memory store map, it throws no store exception to A's requests. A fails to start.

Fix: Before helix manager initialization, controller builds controller client using url or d2 configs and registers system schemas. Since the controller instance is still disabled in helix, helix won't hand over leadership back to A and won't transit B from leader to standby. B will handle A's requests successfully. Also, in case that leadership transits from B to another controller C, increase the number of retries to cover leader->standby period.

## How was this PR tested?

Ran integration tests and tested in EI

### DIFF
--- a/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceController.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceController.java
@@ -202,6 +202,7 @@ public class VeniceController {
           Optional.of(new StoreBackupVersionCleanupService((VeniceHelixAdmin) admin, multiClusterConfigs));
       LOGGER.info("StoreBackupVersionCleanupService is enabled");
     }
+    // Run before enabling controller in helix so leadership won't hand back to this controller during schema requests.
     initializeSystemSchema(controllerService.getVeniceHelixAdmin());
   }
 
--- a/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceController.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceController.java
@@ -202,6 +202,7 @@ public class VeniceController {
           Optional.of(new StoreBackupVersionCleanupService((VeniceHelixAdmin) admin, multiClusterConfigs));
       LOGGER.info("StoreBackupVersionCleanupService is enabled");
     }
+    initializeSystemSchema(controllerService.getVeniceHelixAdmin());
   }
 
   /**
@@ -218,7 +219,6 @@ public class VeniceController {
     if (sslEnabled) {
       secureAdminServer.start();
     }
-    initializeSystemSchema(controllerService.getVeniceHelixAdmin());
     topicCleanupService.start();
     storeBackupVersionCleanupService.ifPresent(AbstractVeniceService::start);
     storeGraveyardCleanupService.ifPresent(AbstractVeniceService::start);
@@ -232,9 +232,9 @@ public class VeniceController {
 
   private void initializeSystemSchema(Admin admin) {
     String systemStoreCluster = multiClusterConfigs.getSystemSchemaClusterName();
+    VeniceControllerConfig systemStoreClusterConfig = multiClusterConfigs.getControllerConfig(systemStoreCluster);
     if (!multiClusterConfigs.isParent() && multiClusterConfigs.isZkSharedMetaSystemSchemaStoreAutoCreationEnabled()
-        && multiClusterConfigs.getControllerConfig(systemStoreCluster)
-            .isSystemSchemaInitializationAtStartTimeEnabled()) {
+        && systemStoreClusterConfig.isSystemSchemaInitializationAtStartTimeEnabled()) {
       ControllerClientBackedSystemSchemaInitializer metaSystemStoreSchemaInitializer =
           new ControllerClientBackedSystemSchemaInitializer(
               AvroProtocolDefinition.METADATA_SYSTEM_SCHEMA_STORE,
@@ -243,7 +243,7 @@ public class VeniceController {
               AvroProtocolDefinition.METADATA_SYSTEM_SCHEMA_STORE_KEY.getCurrentProtocolVersionSchema(),
               VeniceHelixAdmin.DEFAULT_USER_SYSTEM_STORE_UPDATE_QUERY_PARAMS,
               true,
-              multiClusterConfigs.isControllerEnforceSSLOnly());
+              systemStoreClusterConfig);
       metaSystemStoreSchemaInitializer.execute();
     }
   }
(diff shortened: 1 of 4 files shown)


services/venice-controller/src/main/java/com/linkedin/venice/controller/init/ControllerClientBackedSystemSchemaInitializer.java
```diff
@@ -29,16 +31,21 @@
...
   private final AvroProtocolDefinition protocolDefinition;
   private final String clusterName;
   private final VeniceHelixAdmin admin;
   private final Schema keySchema;
   private final UpdateStoreQueryParams storeMetadataUpdate;
   private final boolean autoRegisterPartialUpdateSchema;
-  private final boolean enforceSSLOnly;
-  private ControllerClient controllerClient;
+  private final VeniceControllerConfig controllerConfig;
+  private D2ControllerClient d2ControllerClient;
```
FelixGV, line 48: Is the controller tightly-coupled to D2 with this change? Or can the `D2ControllerClient` connect to non-D2 URLs as well...? We have open source users that wish to not run D2, so it would be better to avoid introducing a tight coupling... can it be made to use one or the other?

shuhui-liu: Thanks for the prompt review. Good call. I didn't know that there are open source users not willing to use D2. Are they planning to configure child.cluster.url instead?

If so, I can make code changes here to be able to construct controller client using child.cluster.url config. If not, there will be two issues:
1. Before initializing helix manager, we can not use VeniceHelixAdmin#getLeaderController to get leader url. Without controller url and d2, the feature that registers system schemas upon controller startup won't work for these users.
2. Parent controller construct child controller clients to query job status, get future version etc. Child controller client are built via url or d2 ([code](https://github.com/linkedin/venice/blob/23d91ddbadeba045764b71391056560a10e0740e/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java#L1510)). If neither url nor d2 is available, parent controller can not query job or store info in child regions.

services/venice-controller/src/main/java/com/linkedin/venice/controller/init/ControllerClientBackedSystemSchemaInitializer.java
```diff
@@ -47,23 +55,53 @@ public ControllerClientBackedSystemSchemaInitializer(
...
+          controllerConfig.getChildControllerD2ServiceName(),
+          clusterName,
+          controllerConfig.getChildControllerD2ZkHost(regionName),
+          controllerConfig.isControllerEnforceSSLOnly() ? admin.getSslFactory() : Optional.empty());
+    } else {
+      LOGGER.info("No child controller url or d2 configs. Skip system schema registration via controller client.");
+      return;
+    }
+
+    for (int attempt = 1; true; attempt++) {
```
adamxchen, line 90: can we use `RetryUtil` for this? I'd be good to reuse some retry mechanism?

services/venice-controller/src/main/java/com/linkedin/venice/controller/init/ControllerClientBackedSystemSchemaInitializer.java
```diff
@@ -29,15 +35,20 @@
 public class ControllerClientBackedSystemSchemaInitializer {
   private static final Logger LOGGER = LogManager.getLogger(ControllerClientBackedSystemSchemaInitializer.class);
   private static final String DEFAULT_KEY_SCHEMA_STR = "\"int\"";
-  private static final int DEFAULT_RETRY_TIMES = 10;
+  /**
+   * Current leader controller might transit from leader to standby, clear cached store repository, and fail to handle
+   * schema requests. We leverage retries to cover the leader->standby period so that a later retry will be sent to the
+   * new leader controller, who can handle schema requests successfully.
```
lluwm, line 41: Maybe it also helps to specify how long the period it can cover with the current setting?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceController.java
```diff
@@ -202,6 +202,7 @@ private void createServices() {
           Optional.of(new StoreBackupVersionCleanupService((VeniceHelixAdmin) admin, multiClusterConfigs));
       LOGGER.info("StoreBackupVersionCleanupService is enabled");
     }
+    initializeSystemSchema(controllerService.getVeniceHelixAdmin());
```
lluwm, line 205: Is this code move corresponding to the following part in the description? If so, can we add some short comments before this line to remind future readers?

> Before helix manager initialization, controller builds controller client using url or d2 configs and registers system schemas. Since the controller instance is still disabled in helix, helix won't hand over leadership back to A and won't transit B from leader to standby.

services/venice-controller/src/main/java/com/linkedin/venice/controller/init/ControllerClientBackedSystemSchemaInitializer.java
```diff
@@ -47,23 +58,52 @@ public ControllerClientBackedSystemSchemaInitializer(
...
+
+    try {
+      RetryUtils.executeWithMaxAttempt(
+          () -> controllerClient.getLeaderControllerUrl(),
+          3,
+          Duration.ofSeconds(2),
+          Collections.singletonList(Exception.class));
+    } catch (Exception e) {
+      LOGGER.info(
+          "Could not find leader controller after retries. This controller is very likely the first one in "
```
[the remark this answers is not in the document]
shuhui-liu: Yes, this means the region does not have any live controller yet (current controller is spinning up but is not live either). Otherwise, helix will assign leadership and there will be a leader. Moved the logic into a function.

services/venice-controller/src/main/java/com/linkedin/venice/controller/init/ControllerClientBackedSystemSchemaInitializer.java
shuhui-liu: Done.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceController.java
shuhui-liu: Yes. Added a comment.

lluwm (approved):
Thanks Shuhui for this change and addressing the comments. LGTM!

PR #522, opened 2023-06-30
[server] Minor improvement of ingestion isolation logging

sixpluszero:
## [server] Minor improvement of ingestion isolation logging
1. Fix a log bug happen during ingestion isolation handover: When the SIT partition is shutdown in time, the old logic still log wrong msg that it does not stop in time and the log is ERROR level. This causes in house deployment tooling to report ERROR logging issues. This PR fixes the logic to only log it when it is actually times out and reduce the log to WARN level.
2. Remove StoreVersionState.toString() logging in handover as this is the underlying SpecificRecord's toString and will print out binary data we store inside SVS. This PR simply remove the logging as we are not interested into the content of the SVS. 

## How was this PR tested?

Existing tests.


majisourav99 (approved):
LGTM

PR #523, opened 2023-06-30
[pubsub][test] Add support for passing pubsub broker configs to Venice components in the test framework

sushantmane:
## Add support for passing pubsub broker configs to Venice components in the test framework
Sometimes to construct pubsub clients we need to pass additional details about the pubsub brokers.   
This PR will let us pass these details to the pubsub clients in integration testing env. 


ZacAttack (approved):
Looks good, lets rerun test suite A (I've noticed it's been flaky lately) and then should be able to commit.

sushantmane:
Thanks, @ZacAttack for the review! Can you please take one more look, please?

ZacAttack (approved):

PR #524, opened 2023-06-30
[dvc][server] Support for clearing expired DIV state

FelixGV:
[dvc][server] Support for clearing expired DIV state

The DIV state could grow unboudedly since it tracks some metadata
for each producer, and never cleared them out. This is problematic
for use cases where a hybrid store is RT-only, meaning it doesn't
rev up store-versions, and as the servers are bounced overtime,
the leader replicas generate new GUIDs overtime, which linger in
the DIV state forever. In extreme cases, this leads to OOM errors.

This change mitigates the unbounded growth by introducing a max
age, beyond which DIV metadata can be evicted. This is controlled 
by a new config, which is disabled by default:

div.producer.state.max.age.ms (default: -1)

The clearing logic is inserted in the updateOffsetRecordForPartition
and function of KafkaDataIntegrityValidator, which is itself called
prior to each sync (both as part of ingestion when a threshold is
reached, and during graceful shutdown). It is also called in the
setPartitionState function of the same class, which is called when
subscribing to a partition, including when the server starts up.

The age of a producer is determined by the most recent timestamp
within a partition (computed by OffsetRecord::getMaxMessageTimeInMs)
minus the most recent timestamp of that producer (which is the 
Segments' lastRecordProducerTimestamp). These timestamps can be 
subject to drift but in practice it probably isn't going to drift 
enough to be a concern. Importantly, because the math is entirely
within a single partition, this means one partition can be lagging
(e.g. as would be the case in a rebootstrap scenario) while another
may be caught up, and each partition will independently clear their 
own states without contaminating each other.

This property of isolating the partitions' expired state clearing
is achieved by inverting the KafkaDataIntegrityValidator's internal 
data structure. Previously, it contained a Map<GUID, ProducerTracker> 
with each PT containing one Segment per tracked partition. Now, it
instead contains a SparseConcurrentList of PartitionTracker (which 
is essentially the same as ProducerTracker, but renamed), each of 
which contains a Map<GUID, Segment>. This is a more natural structure 
that aligns with other parts of the code as well. It also results in
eliminating 2 map lookups from the DIV hot path (called 1/msg in 
followers, and 2/msg in leaders).

Note that currently the state clearing happens only in the "main"
KafkaDataIntegrityValidator, used by drainers, and which persist
their state to disk. There is another KafkaDataIntegrityValidator
used for leader replicas only, which track the RT topic's DIV state,
but that one is transient and not persisted to disk. As such, it is
not expected to grow as large, and bouncing effectively clears it
(which is not the case for the one which gets persisted). Later on,
we could trigger state cleaning for this other validator as well.

Miscellaneous changes:

- Removed the RedundantExceptionFilter::getRedundantExceptionFilter
  with parameters, since that is broken by design... this and the
  non-parameterized functions both returned a singleton object, so
  the actual parameters of the singleton were driven by the timing
  of which code path got called first, which makes no sense.

- Moved ProducerTracker (renamed as PartitionTracker) and its unit 
  test from venice-common to dvc, since it is only used in the latter.

- StoreIngestionTaskTest::testDataValidationCheckPointing no longer
  has invocationCount = 3, as this is a pretty long test, and it 
  doesn't seem useful.

- Introduced a ReentrantLockWithRef, a new utility class which 
  extends ReentrantLock and also carries a ref to another object. 
  This is used in PartitionTracker, where previously we would need
  to query two maps, one to get the lock, and the other to get the
  protected object. Since we were always operating on the same key 
  of both maps, we could instead group the elements into this shared
  container, hosted in a single map.

- Also removed several unnecessary objects from Segment. A server
  heap dump was found containing >3M instances of this class, so 
  making it more efficient should be beneficial (i.e. even when DIV 
  state clearing is not enabled yet).

- Checksum::getInstance does not return an Optional anymore.

- Reduced string concatenation in several places.

## How was this PR tested?
CI (including new unit tests)

A new dvc/server config.


clients/da-vinci-client/src/main/java/com/linkedin/davinci/validation/KafkaDataIntegrityValidator.java
```diff
@@ -24,30 +29,53 @@
...
+  private volatile long oldestRetainedProducerTS = 0;
 
   /** Keeps track of every upstream producer this consumer task has seen so far. */
   protected final Map<GUID, ProducerTracker> producerTrackerMap;
   protected final Function<GUID, ProducerTracker> producerTrackerCreator;
+  private final Time time;
 
-  public KafkaDataIntegrityValidator(String kafkaVersionTopic) {
-    this(kafkaVersionTopic, -1);
+  public KafkaDataIntegrityValidator(String versionTopicName, long kafkaLogCompactionDelayInMs) {
```
ZacAttack, line 45: optional/maybe nit?: Does this class only work for version topics or can it be used for RT?  If not, maybe just remove the version notion so folks don't make that assumption later?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -246,7 +246,10 @@ public LeaderFollowerStoreIngestionTask(
             .build();
     this.veniceWriter = Lazy.of(() -> veniceWriterFactory.createVeniceWriter(writerOptions));
     this.kafkaClusterIdToUrlMap = serverConfig.getKafkaClusterIdToUrlMap();
-    this.kafkaDataIntegrityValidatorForLeaders = new KafkaDataIntegrityValidator(kafkaVersionTopic);
+    this.kafkaDataIntegrityValidatorForLeaders = new KafkaDataIntegrityValidator(
+        kafkaVersionTopic,
+        KafkaDataIntegrityValidator.DISABLED,
```
ZacAttack, line 251: Why is this hardwired to DISABLED?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/validation/KafkaDataIntegrityValidator.java
FelixGV: IIUC, the topic name is there only for logging purposes, and the class ought to work for RT (or any Venice topics, really)... I'll change the variable name. Thanks for pointing it out!

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
FelixGV: So... this 2nd param is used in a code path that's not used by anything in open source... but the proprietary ETL populates that param. It is kind of messy, admittedly, but I fear refactoring this is a bit complex for now.

ZacAttack: Ah I see.  Ok.  Maybe throw a TODO at the interface in the class so no one deletes it later while "cleaning up"

clients/da-vinci-client/src/main/java/com/linkedin/davinci/validation/KafkaDataIntegrityValidator.java
```diff
@@ -97,4 +122,37 @@ public void checkMissingMessage(
...
+              if (currentProducerOldestRetainedTS == Long.MAX_VALUE) {
+                iterator.remove();
+                numberOfClearedGUIDs++;
+              } else {
+                newOldestRetainedProducerTS = Math.min(newOldestRetainedProducerTS, currentProducerOldestRetainedTS);
+              }
+            }
+            if (newOldestRetainedProducerTS == Long.MAX_VALUE) {
+              /** We only update the {@link #oldestRetainedProducerTS} if there is any state we retained. */
+              this.oldestRetainedProducerTS = newOldestRetainedProducerTS;
```
ZacAttack, line 148: How does the state of oldestRetainedProducerTS get reset to something less then Long.MAX_VALUE after this state?  It seems like if this happens, then this function won't run again ever.

ZacAttack (commented):
Looks good overall.  Just one comment above and then I think we can get this in.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/validation/KafkaDataIntegrityValidator.java
FelixGV: Oh that's not right... it's meant to be if !=... I'll try to extend the unit test to demonstrate that this is broken...

FelixGV: The latest diff has the test and the fix. Thanks for catching this bug!

FelixGV:
I just thought of something... there might be a problem with the approach taken here... while bootstrapping old data (e.g. during a rebalance), we may clear the DIV state during the bootstrapping, which doesn't seem right. The ingestion might still succeed, since it needs to be resilient to log compaction, which may result in similar symptoms, but it feels sketchy.

The last producer timestamp of a segment may not be a good enough heuristic... it may be necessary to consider lag as well, and only clear state for partitions which aren't lagging... this would complicate the implementation though.

FelixGV:
> I just thought of something... there might be a problem with the approach taken here... while bootstrapping old data (e.g. during a rebalance), we may clear the DIV state during the bootstrapping, which doesn't seem right. The ingestion might still succeed, since it needs to be resilient to log compaction, which may result in similar symptoms, but it feels sketchy.
> 
> The last producer timestamp of a segment may not be a good enough heuristic... it may be necessary to consider lag as well, and only clear state for partitions which aren't lagging... this would complicate the implementation though.

I let it stew overnight and I have another idea... instead of measuring the age as `current time - last producer timestamp`, we could measure it as `most recent producer timestamp within the partition - last producer timestamp`. I think this would make the algorithm work as desired during rebootstrap scenarios as well. It also seems more elegant since the algorithm becomes a state machine without any non-deterministic IO...

I will think about it some more and try implementing the change later...

gaojieliu:
> > I just thought of something... there might be a problem with the approach taken here... while bootstrapping old data (e.g. during a rebalance), we may clear the DIV state during the bootstrapping, which doesn't seem right. The ingestion might still succeed, since it needs to be resilient to log compaction, which may result in similar symptoms, but it feels sketchy.
> > The last producer timestamp of a segment may not be a good enough heuristic... it may be necessary to consider lag as well, and only clear state for partitions which aren't lagging... this would complicate the implementation though.
> 
> I let it stew overnight and I have another idea... instead of measuring the age as `current time - last producer timestamp`, we could measure it as `most recent producer timestamp within the partition - last producer timestamp`. I think this would make the algorithm work as desired during rebootstrap scenarios as well. It also seems more elegant since the algorithm becomes a state machine without any non-deterministic IO...
> 
> I will think about it some more and try implementing the change later...

I think you most recent idea sounds good since it aligns with the original expectation that the producer, which is not producing in the latest X time period, can be cleared.

sixpluszero:
> I just thought of something... there might be a problem with the approach taken here... while bootstrapping old data (e.g. during a rebalance), we may clear the DIV state during the bootstrapping, which doesn't seem right. The ingestion might still succeed, since it needs to be resilient to log compaction, which may result in similar symptoms, but it feels sketchy.
> 
> The last producer timestamp of a segment may not be a good enough heuristic... it may be necessary to consider lag as well, and only clear state for partitions which aren't lagging... this would complicate the implementation though.

Hi Felix, why clear DIV state during bootstrap is not good?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/validation/KafkaDataIntegrityValidator.java
```diff
@@ -97,4 +130,37 @@ public void checkMissingMessage(
...
+              producerTracker = iterator.next();
+              long currentProducerOldestRetainedTS = producerTracker.clearExpiredState(maxRecordProducerTimestamp);
+              if (currentProducerOldestRetainedTS == Long.MAX_VALUE) {
+                iterator.remove();
+                numberOfClearedGUIDs++;
+              } else {
+                newOldestRetainedProducerTS = Math.min(newOldestRetainedProducerTS, currentProducerOldestRetainedTS);
+              }
+            }
+            if (newOldestRetainedProducerTS != Long.MAX_VALUE) {
```
gaojieliu, line 154: If there is no state retained, shall we update `newOldestRetainedProducerTS` to the current time to avoid the empty while loop?
Maybe it doesn't matter since empty loop is cheap..

clients/da-vinci-client/src/main/java/com/linkedin/davinci/validation/ProducerTracker.java
```diff
@@ -105,49 +103,58 @@ public void setPartitionState(int partition, ProducerPartitionState state) {
...
     }
-    ReentrantLock partitionLock = getPartitionLock(partition);
-    partitionLock.lock();
+    lockAndSegment.lock();
     try {
-      Segment sourceSegment = segments.get(partition);
+      if (!lockAndSegment.containsRef()) {
+        // This producer didn't write anything to requested partition
+        return;
+      }
```
gaojieliu, line 136: Why do we need this check inside the lock scope? Any race condition this logic is trying to guard? If yes, can you leave some javadoc?
Same for the following function.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -246,7 +246,10 @@ public LeaderFollowerStoreIngestionTask(
...
+        getServerConfig().getDivProducerStateMaxAgeMs());
```
gaojieliu, line 252: `kafkaDataIntegrityValidatorForLeaders` should never persist offset, so shall we disable this DIV producer cleanup feature explicitly?
Maybe we should introduce a flag to indicate that `kafkaDataIntegrityValidatorForLeaders` should never call `updateOffsetRecordForPartition`, and if that happens, an exception should be thrown to fail fast.

FelixGV: Hmm... we could hardcode it to disabled, but IDK if that is ideal. The goal is not just to clear old state to be persisted, but also to clear old state which is hanging in memory. Actually, the latter is the main motivation for this change, since unbounded memory growth can lead to OOMing, whereas unbounded growth of persistent storage is not a concern (given that metadata is pretty small overall...).

That being said, maybe there is something special about the usage of this "DIV for leaders" which makes it not prone to leaking old state. I'm not sure about that.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/validation/KafkaDataIntegrityValidator.java
FelixGV: I don't use this field anymore in the latest iteration. Maybe it could be brought back in some way if the state clearing algorithm is feared to be too expensive, but I'm not sure ATM.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/validation/ProducerTracker.java
FelixGV: It is possible that another thread calls `removeRef` between the first (lock-free) check and the lock acquisition.

But if we think this looks too complicated, then I could remove the pre-lock check.

FelixGV:
> Hi Felix, why clear DIV state during bootstrap is not good?

The problem I had in mind was that in my first iteration, I considered all partitions together. This means that if any partition had received a recent record, then any other partition that is lagging behind considerably (e.g. due to a rebootstrap) would have all of its state eligible for cleaning, since all of the messages it's consuming in the early messages of the partition are all older than the most recent timestamp - the maximum age.

With the new algorithm, the check is done within a single partition, rather than across partition, which seems cleaner to me, as partitions should be independent of each other, and not capable of contaminating each other.

LMK if that makes sense...

clients/da-vinci-client/src/main/java/com/linkedin/davinci/validation/KafkaDataIntegrityValidator.java
```diff
@@ -62,12 +119,41 @@ public void validateMessage(
...
+
+  void clearExpiredStateAndUpdateOffsetRecordForPartition(int partition, OffsetRecord offsetRecord) {
+    long minimumRequiredRecordProducerTimestamp = offsetRecord.getMaxMessageTimeInMs() - this.maxAgeInMs;
+    int numberOfClearedGUIDs = 0;
+    Iterator<ProducerTracker> iterator = this.producerTrackerMap.values().iterator();
+    ProducerTracker producerTracker;
+    while (iterator.hasNext()) {
+      producerTracker = iterator.next();
+      if (producerTracker.clearExpiredState(partition, minimumRequiredRecordProducerTimestamp)
+          && producerTracker.isEmpty()) {
```
gaojieliu, line 146: It seems there is a race condition here.
`ProducerTracker` is maintaining a state per partition and multiple partitions can update `ProducerTracker` concurrently via `validateMessage`.

Maybe we need to use some read-write lock to guard this logic.
Such as the clear logic will grab the write lock and all other operations need to grab the read lock.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/validation/ProducerTracker.java
gaojieliu: I guess that is fine to leave it in this way.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/validation/ProducerTracker.java
```diff
@@ -632,10 +670,78 @@ public void checkMissingMessage(
...
+        return true;
+      }
+      return false;
+    } finally {
+      lockAndSegment.unlock();
+    }
+  }
+
+  public void removeState(int partition) {
+    segments.remove(partition);
```
gaojieliu, line 706: Is this one line function necessary?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/validation/KafkaDataIntegrityValidator.java
```diff
@@ -62,12 +119,41 @@ public void validateMessage(
...
+        iterator.remove();
+        offsetRecord.setProducerPartitionState(producerTracker.getProducerGUID(), null);
```
gaojieliu, line 148: if `producerTracker.clearExpiredState(partition, minimumRequiredRecordProducerTimestamp)` returns true, that means this specific guid is stale for this specific partition, we can clean up the corresponding `OffsetRecord`, which is per partition, right?
Also when cleaning up the OffsetRecord, shall we remove that producer guid completely instead of setting it to be `null`?

```
public java.util.Map<java.lang.CharSequence,com.linkedin.venice.kafka.protocol.state.ProducerPartitionState> producerStates;
```
doesn't support `null` value, so I suspect `null` could break the serialization.

gaojieliu (changes requested):
In StoreIngestionTask, there are two instances of `KafkaDataIntegrityValidator`:
1. For drainer to validate the messages consumed from VT.
2. For leader consumer to validate the messages consumed from RT.

If I understand the code change correctly, there is no logic to clean up the producer info in #2 since #2 doesn't persist offset record, such as if the leader replicas run for a very long time and the total number of producer will accumulate.

There is a concern about the potential race condition when clearing some empty `ProducerTracker`.

I am thinking whether we can consider another cleanup approach:
1. StoreIngestionTask has a function called `checkIngestionProgress`, which will be executed periodically.
2. We can execute the clean up logic inside this function and the cleanup logic will cleanup every partition subscribed including both instances of `KafkaDataIntegrityValidator`.
3. Put a read-write lock inside `KafkaDataIntegrityValidator` to avoid race condition:
   a. Cleanup logic will grab a write lock and other functions such as `validateMessage` and `updateOffsetRecordForPartition` and others need to grab a read lock.
4. The cleanup logic can be triggered by time interval, such as clean up the staled producer every 10 mins (configurable), and I assume the read-write lock overhead can be minimal if the  delay is big enough.

If my understanding is correct, what do you think about the above idea?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/validation/KafkaDataIntegrityValidator.java
FelixGV: I think you're right... putting null here seems messy. I'll change it to removing from the map, which seems cleaner IMHO.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/validation/KafkaDataIntegrityValidator.java
FelixGV: Hmm... I'll see what I can do to tweak this. Also, now that I look at this, having the `&& producerTracker.isEmpty()` as part of the same if condition might be wrong... maybe the ideal flow would be along the lines of:

```
if (producerTracker.clearExpiredState(partition, minimumRequiredRecordProducerTimestamp)) {
  // lock
  if (producerTracker.isEmpty()) {
    // remove state
  }
  // unlock
} else {
  producerTracker.updateOffsetRecord(partition, offsetRecord);
}
```

Because the else block should happen when the state of that specific partition is not expired, regardless of the state of the other partitions.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/validation/PartitionTracker.java
```diff
@@ -19,136 +20,178 @@
...
    */
-  public void clearPartition(int partition) {
-    ReentrantLock partitionLock = getPartitionLock(partition);
-    partitionLock.lock();
-    try {
-      segments.remove(partition);
-    } finally {
-      partitionLock.unlock();
-    }
+  private ReentrantLockWithRef<Segment> getLockAndSegment(GUID guid) {
```
gaojieliu, line 101: This potentially doesn't require any more.
The original reason to introduce partition lock is for AA with aggregate mode to avoid race conditions for the same records from the same GUID.
Now, such mode doesn't exist any more, this locking logic can be cleaned up.

Today, records produced by the same GUID shouldn't appear in multiple RT topics.
But this optimization can be done later.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/validation/KafkaDataIntegrityValidator.java
```diff
@@ -55,28 +77,29 @@ public void validateMessage(
...
   }
 
   public void cloneProducerStates(int partition, KafkaDataIntegrityValidator newValidator) {
-    for (Map.Entry<GUID, ProducerTracker> entry: producerTrackerMap.entrySet()) {
-      GUID producerGUID = entry.getKey();
-      ProducerTracker sourceProducerTracker = entry.getValue();
-      ProducerTracker destProducerTracker = newValidator.registerProducer(producerGUID);
-      sourceProducerTracker.cloneProducerStates(partition, destProducerTracker);
-    }
+    PartitionTracker destPartitionTracker = newValidator.registerPartition(partition);
```
gaojieliu, line 101: Leader DIV validator will clone the producer states from the regular DIV validator since the cleanup will happen when first restoring the regular DIV validator state, the producer state copied to Leader DIV validator should be minimal.
This issue can be addressed when we implement the Leader DIV validator periodical cleanup later.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/validation/PartitionTracker.java
```diff
@@ -19,136 +20,178 @@
...
+        segments.remove(producerGuid);
+        iterator.remove();
+      }
+    }
   }
 
-  private void setPartitionState(int partition, Segment segment) {
-    ReentrantLock partitionLock = getPartitionLock(partition);
-    partitionLock.lock();
+  private void setPartitionState(GUID guid, Segment segment) {
```
gaojieliu, line 139: We may want to rename it to be `setProducerState` to reflect the function purpose.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/validation/PartitionTracker.java
```diff
@@ -632,7 +691,43 @@ public void checkMissingMessage(
...
+          numberOfClearedGUIDs++;
+        } else {
+          updateOffsetRecord(entry.getKey(), lockAndSegment, offsetRecord);
+        }
+      } finally {
+        lockAndSegment.unlock();
+      }
+    }
+    if (numberOfClearedGUIDs > 0) {
+      logger.info("Cleared {} expired producer GUID(s).", numberOfClearedGUIDs);
```
gaojieliu, line 730: Can we also put `partition id` and `topic name` into the log message?

gaojieliu (commented):
Code change looks good to me, thanks!

Left some minor comments, feel free to address them later.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/validation/PartitionTracker.java
FelixGV: Hmm... it'd be great to get rid of this locking as it would simplify the code even more. That being said, the previous structure should not have concurrent interactions on the segments either, since the messages coming from different RT topics would have affected different `ProducerTracker` instances, even if they hit the same partition within each `PT`… I wonder if we’re missing something about why this locking was there in the first place… :thinking:

clients/da-vinci-client/src/main/java/com/linkedin/davinci/validation/PartitionTracker.java
FelixGV: Thanks. Good call. I renamed it to `setSegment`.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/validation/PartitionTracker.java
FelixGV: It will be there. The `logger` is a class property, which is prefixed by topic name and partition id. See the constructor.

FelixGV (commented):
Thanks for the review Goajie!

clients/da-vinci-client/src/main/java/com/linkedin/davinci/validation/PartitionTracker.java
FelixGV: Talked offline. It does look like the locking isn't necessary anymore. It was there for a temporary phase of the AA migration, which is over now. We can remove it, but we'll do it in a future PR.

PR #525, opened 2023-07-11
[router][server] Ignore slow node logic when 5XX encountered

adamxchen:
### The problem statement ### 
During rebalancing caused by the cluster expansion, the router sends the multiget requests to a replica that is supposed to be dropped since router can have stale global view. 
Due to LEAST_LOADED_HOST policy along with the retry logic and the slow node logic, the router ended up with no choice to select a storage node hence treated this request as unhealthy and triggered alarms.

We want to make this behavior more resilient so there's less false alarms. Iin this PR, when retry request is set to 5XX error, we ignore the slow node set so nodes can be selected.

## How was this PR tested?

Newly added unit tests and Github CI


internal/alpini/router/alpini-router-base/src/test/java/com/linkedin/alpini/router/TestScatterGatherRequestHandlerImpl.java
```diff
@@ -86,14 +87,35 @@ private interface Response {
     void setKeepAlive(boolean keepAlive);
   }
 
-  private enum Status {
-    OK(200), MULTI_STATUS(207), BAD_REQUEST(400), NOT_FOUND(404), TOO_BUSY(429), INTERNAL_SERVER_ERROR(500),
-    GATEWAY_TIMEOUT(502), SERVICE_UNAVAILABLE(503);
+  private static class Status extends HttpResponseStatus {
+    static final Status OK = new Status(200, "OK");
```
gaojieliu, line 91: Why this change?

services/venice-server/src/main/java/com/linkedin/venice/listener/StorageReadRequestsHandler.java
```diff
@@ -296,7 +296,7 @@ public void channelRead(ChannelHandlerContext context, Object message) throws Ex
           context.writeAndFlush(
               new HttpShortcutResponse("No storage exists for: " + e.getStoreName(), HttpResponseStatus.BAD_REQUEST));
         } catch (VeniceRequestEarlyTerminationException e) {
```
gaojieliu, line 298: Can you just update `VeniceRequestEarlyTerminationException` to use Netty's `HttpResponseStatus`?

gaojieliu (changes requested):

internal/alpini/router/alpini-router-base/src/test/java/com/linkedin/alpini/router/TestScatterGatherRequestHandlerImpl.java
adamxchen: This change is to make the test-only class `Status` to be compatible with `HttpResponseStatus`. Without this, the casting from HRS to `HttpResponseStatus` in [this line](https://github.com/linkedin/venice/blob/ca18d4f7a3fd9bc51128a038b810a9e6860d6034/internal/alpini/router/alpini-router-base/src/main/java/com/linkedin/alpini/router/ScatterGatherRequestHandlerImpl.java#L682) would fail in tests because `Status` didn't respect it.

services/venice-server/src/main/java/com/linkedin/venice/listener/StorageReadRequestsHandler.java
adamxchen: I don't follow. Do you mean remove the `HttpStatus` at [this line](https://github.com/linkedin/venice/blob/fabfe8912719b4a140f5416525c5991cb44a80b4/services/venice-server/src/main/java/com/linkedin/venice/listener/VeniceRequestEarlyTerminationException.java#L15) from `VeniceRequestEarlyTerminationException`?

adamxchen: Updated.

gaojieliu (approved):

PR #526, opened 2023-07-11
[client-common] Fix non-idempotent `LongAdderRateGauge` by adding cached value. 

haoxu07:
Previously, `LongAdderRateGauge`'s `getRate` call is not idempotent, but it will be called twice in `getLeaderBytesConsumed` at the same time. The second call will always get 0. Here we add a cached value to record sum value for previous bucket. 

Resolves #XXX


internal/venice-client-common/src/main/java/com/linkedin/venice/stats/LongAdderRateGauge.java
```diff
@@ -38,14 +43,17 @@ public double getRate() {
...
+   */
   private double getRate(long currentTimeMs) {
-    /**
-     * N.B.: We are not looking for great precision at a very short timescale. The intended use case is for this
-     * function to be queried by the metric system ~1/minute. So the goal here is just to avoid a division by zero.
-     */
-    double elapsedTimeInSeconds =
-        Math.max((double) (currentTimeMs - this.lastMeasurementTime) / Time.MS_PER_SECOND, 0.001);
+    double elapsedTimeInSeconds = (double) (currentTimeMs - this.lastMeasurementTime) / Time.MS_PER_SECOND;
+    if (elapsedTimeInSeconds < Time.SECONDS_PER_MINUTE) {
```
FelixGV, line 52: I don't think this is a reasonable default... because I don't think we have a guarantee that the metric system will query the metric at least 1 minute apart. If the metric is queried 59 seconds apart, it will lead to stale data... How about 30s?

haoxu07: I feel like there is always stale state due to the cache behavior. I feel like our metric plotting system present data in per-min style, then I think a minute is sufficient, for this type metric, will staleness at minute level be a big problem?

gaojieliu: I guess either 30s or 60s doesn't matter much, but our monitoring/alerting system is being affected by this bug (false alert is generating all the time).
@haoxu07 Please work with @FelixGV to address the comment and merge it ASAP since it has been pending for more than 2 weeks.

FelixGV: Sorry, I had not noticed your response @haoxu07...

My concern is that if the metric is queried 59 seconds apart, then we'll return the old value, which means the graph will contain two identical data point. That is not good...

At first, I was thinking a very short cache duration on the order of 1s or 5s would be enough to fix the issue discovered related to this. But if we want a longer cache that is fine too. That is why I went for the middle ground of 30s, which I think is unlikely that our metric system would have such a short gap (though not necessarily impossible if some Gauge metrics running arbitrary code got stuck for a long time, but I guess we could not worry about that for now...).

So, which do you prefer: 1s or 30s? Take your pick and let's merge :D ...

haoxu07: I decide to change it into 30s window.

FelixGV (dismissed):
Thanks!

FelixGV (approved):

PR #531, opened 2023-07-13
[server][controller] Support meta system store schema registeration at server start time

shuhui-liu:
## Summary

When a server starts, if system schema initialization is enabled, it will call controller to check and register new meta system store schemas. Therefore, servers can be deployed before controllers when meta system store schema is upgraded. Controller url (for open source users who do not want to onboard d2) or d2 configs must be provided to enable system schema initialization at start time.

## How was this PR tested?

Unit test and integration test 

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/config/VeniceServerConfig.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/config/VeniceServerConfig.java
@@ -20,6 +20,9 @@ import static com.linkedin.venice.ConfigKeys.KEY_VALUE_PROFILING_ENABLED;
 import static com.linkedin.venice.ConfigKeys.LEADER_FOLLOWER_STATE_TRANSITION_THREAD_POOL_STRATEGY;
 import static com.linkedin.venice.ConfigKeys.LISTENER_HOSTNAME;
 import static com.linkedin.venice.ConfigKeys.LISTENER_PORT;
+import static com.linkedin.venice.ConfigKeys.LOCAL_CONTROLLER_D2_SERVICE_NAME;
+import static com.linkedin.venice.ConfigKeys.LOCAL_CONTROLLER_URL;
+import static com.linkedin.venice.ConfigKeys.LOCAL_D2_ZK_HOST;
 import static com.linkedin.venice.ConfigKeys.MAX_FUTURE_VERSION_LEADER_FOLLOWER_STATE_TRANSITION_THREAD_NUMBER;
 import static com.linkedin.venice.ConfigKeys.MAX_LEADER_FOLLOWER_STATE_TRANSITION_THREAD_NUMBER;
 import static com.linkedin.venice.ConfigKeys.OFFSET_LAG_DELTA_RELAX_FACTOR_FOR_FAST_ONLINE_TRANSITION_IN_RESTART;
@@ -101,6 +104,7 @@ import static com.linkedin.venice.ConfigKeys.STORE_WRITER_BUFFER_MEMORY_CAPACITY
 import static com.linkedin.venice.ConfigKeys.STORE_WRITER_BUFFER_NOTIFY_DELTA;
 import static com.linkedin.venice.ConfigKeys.STORE_WRITER_NUMBER;
 import static com.linkedin.venice.ConfigKeys.SYSTEM_SCHEMA_CLUSTER_NAME;
+import static com.linkedin.venice.ConfigKeys.SYSTEM_SCHEMA_INITIALIZATION_AT_START_TIME_ENABLED;
 import static com.linkedin.venice.ConfigKeys.UNREGISTER_METRIC_FOR_DELETED_STORE_ENABLED;
 import static com.linkedin.venice.ConfigKeys.UNSORTED_INPUT_DRAINER_SIZE;
 
@@ -368,6 +372,10 @@ public class VeniceServerConfig extends VeniceClusterConfig {
   private final long fastClassSchemaWarmupTimeout;
 
   private final boolean schemaPresenceCheckEnabled;
+  private final boolean systemSchemaInitializationAtStartTimeEnabled;
+  private final String localControllerUrl;
+  private final String localControllerD2ServiceName;
+  private final String localD2ZkHost;
 
   private final boolean enableLiveConfigBasedKafkaThrottling;
 
@@ -599,6 +607,11 @@ public class VeniceServerConfig extends VeniceClusterConfig {
     fastClassSchemaWarmupTimeout =
         serverProperties.getLong(SERVER_SCHEMA_FAST_CLASS_WARMUP_TIMEOUT, 2 * Time.MS_PER_MINUTE);
     schemaPresenceCheckEnabled = serverProperties.getBoolean(SERVER_SCHEMA_PRESENCE_CHECK_ENABLED, true);
+    systemSchemaInitializationAtStartTimeEnabled =
+        serverProperties.getBoolean(SYSTEM_SCHEMA_INITIALIZATION_AT_START_TIME_ENABLED, false);
+    localControllerUrl = serverProperties.getString(LOCAL_CONTROLLER_URL, "");
+    localControllerD2ServiceName = serverProperties.getString(LOCAL_CONTROLLER_D2_SERVICE_NAME, "");
+    localD2ZkHost = serverProperties.getString(LOCAL_D2_ZK_HOST, "");
     enableLiveConfigBasedKafkaThrottling =
         serverProperties.getBoolean(SERVER_ENABLE_LIVE_CONFIG_BASED_KAFKA_THROTTLING, false);
     /**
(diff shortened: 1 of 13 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/config/VeniceServerConfig.java
```diff
@@ -599,6 +607,11 @@ public VeniceServerConfig(VeniceProperties serverProperties, Map<String, Map<Str
     fastClassSchemaWarmupTimeout =
         serverProperties.getLong(SERVER_SCHEMA_FAST_CLASS_WARMUP_TIMEOUT, 2 * Time.MS_PER_MINUTE);
     schemaPresenceCheckEnabled = serverProperties.getBoolean(SERVER_SCHEMA_PRESENCE_CHECK_ENABLED, true);
+    systemSchemaInitializationAtStartTimeEnabled =
+        serverProperties.getBoolean(SYSTEM_SCHEMA_INITIALIZATION_AT_START_TIME_ENABLED, false);
+    localControllerUrl = serverProperties.getString(LOCAL_CONTROLLER_URL, "");
+    localControllerD2ServiceName = serverProperties.getString(LOCAL_CONTROLLER_D2_SERVICE_NAME, "");
+    localD2ZkHost = serverProperties.getString(LOCAL_D2_ZK_HOST, "");
```
gaojieliu, line 614: Why do we need both?
Can we just use d2?

internal/venice-common/src/main/java/com/linkedin/venice/common/VeniceSystemStoreUtils.java
```diff
@@ -12,6 +14,12 @@ public class VeniceSystemStoreUtils {
   private static final String PUSH_JOB_DETAILS_STORE_NAME =
       String.format(Store.SYSTEM_STORE_FORMAT, PUSH_JOB_DETAILS_STORE);
   public static final String SEPARATOR = "_";
+  public static final UpdateStoreQueryParams DEFAULT_USER_SYSTEM_STORE_UPDATE_QUERY_PARAMS =
```
gaojieliu, line 17: This static field is the same as the one defined in the following class, so can reuse it?

internal/venice-common/src/main/java/com/linkedin/venice/system/store/ControllerClientBackedSystemSchemaInitializer.java
```diff
@@ -97,22 +106,32 @@ public void execute() {
...
+                + discoveryResponse.getError());
+      }
+    } else {
+      String currSystemStoreCluster = discoveryResponse.getCluster();
       if (!currSystemStoreCluster.equals(clusterName)) {
         LOGGER.warn(
-            "The system store for {} already exists in cluster {}, "
-                + "which is inconsistent with the config {} which specifies that it "
-                + "should be in cluster {}. Will abort the initialization routine.",
+            "The system store for {} already exists in cluster {}, which is inconsistent with the config {} which "
```
gaojieliu, line 127: This should be an `error` message, right?
Also I wonder whether we should return some error code for this case, so that Venice Server can choose to abort the startup when this issue happens.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerConfig.java
```diff
@@ -476,7 +476,7 @@ public VeniceControllerConfig(VeniceProperties props) {
     this.parentExternalSupersetSchemaGenerationEnabled =
         props.getBoolean(CONTROLLER_PARENT_EXTERNAL_SUPERSET_SCHEMA_GENERATION_ENABLED, false);
     this.systemSchemaInitializationAtStartTimeEnabled =
-        props.getBoolean(SYSTEM_SCHEMA_INITIALIZATION_AT_START_TIME_ENABLED, true);
+        props.getBoolean(SYSTEM_SCHEMA_INITIALIZATION_AT_START_TIME_ENABLED, false);
```
gaojieliu, line 479: Why this change?

gaojieliu (changes requested):

clients/da-vinci-client/src/main/java/com/linkedin/davinci/config/VeniceServerConfig.java
shuhui-liu: I explained in the description that controller url is for open source users who do not want to onboard d2. I heard from Felix that some open source users do not want to run d2 https://github.com/linkedin/venice/pull/521#discussion_r1245914950

internal/venice-common/src/main/java/com/linkedin/venice/common/VeniceSystemStoreUtils.java
shuhui-liu: Yes they're duplicate. I'll remove one of them.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerConfig.java
shuhui-liu: I made a change in this commit. When this config is enabled but neither url nor d2 config is provided, throw exception instead of aborting the initialization and continuing that startup of controller.
```
throw new VeniceException(
            "System schema initialization is enabled but neither controller url nor d2 config is provided.");
```
If the default of this config continues to be true, open source users who do not provide controller url or d2 will fail to start controllers. So I changed it to false.

internal/venice-common/src/main/java/com/linkedin/venice/system/store/ControllerClientBackedSystemSchemaInitializer.java
shuhui-liu: Yeah the system store should exist in the system schema cluster. There must be something wrong if it exists in another cluster. I'm ok with throwing an exception to abort server/controller startup when this issue happens.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerConfig.java
gaojieliu: I see.
Let us make sure we have this config enabled in LinkedIn env.

gaojieliu (commented):
Sorry for the late follow up.
Do we have a test to validate the new server behavior when enabling system store schema registration? I didn't see it from this PR.

shuhui-liu:
> Sorry for the late follow up. Do we have a test to validate the new server behavior when enabling system store schema registration? I didn't see it from this PR.

No problem. Since VeniceServerWrapper has the following code
```
.put(SYSTEM_SCHEMA_INITIALIZATION_AT_START_TIME_ENABLED, true)
```
, when any server starts during the integration test, ControllerClientBackedSystemSchemaInitializer will be executed. If something goes wrong during the initialization, server will fail to start and the integration test will fail. Let me know if you think that more testings need to be added.

services/venice-server/src/main/java/com/linkedin/venice/server/VeniceServer.java
```diff
@@ -209,32 +211,64 @@ private List<AbstractVeniceService> createServices() {
     // Create jvm metrics object
     jvmStats = new VeniceJVMStats(metricsRepository, "VeniceJVMStats");
 
+    if (serverConfig.isSystemSchemaInitializationAtStartTimeEnabled()) {
```
lluwm, line 214: What would happen if multiple servers send concurrent `addValueSchema` to the leader controller with the same value schema contents, looks like `VeniceHelixAdmin.addValueSchema` might have an issue dealing with it?

gaojieliu:
> > Sorry for the late follow up. Do we have a test to validate the new server behavior when enabling system store schema registration? I didn't see it from this PR.
> 
> No problem. Since VeniceServerWrapper has the following code
> 
> ```
> .put(SYSTEM_SCHEMA_INITIALIZATION_AT_START_TIME_ENABLED, true)
> ```
> 
> , when any server starts during the integration test, ControllerClientBackedSystemSchemaInitializer will be executed. If something goes wrong during the initialization, server will fail to start and the integration test will fail. Let me know if you think that more testings need to be added.

So this behavior is being tested implicitly, right?
In integration test, normally, we will spin up the Controller first and then other components, that means Venice Server won't execute the schema registration code path (very likely).
Is it possible to explicitly disable system store schema registration in Controller and exercise the Server registration path?

services/venice-server/src/main/java/com/linkedin/venice/server/VeniceServer.java
shuhui-liu: It uses the pre-defined schema id to add value schema. API `addValueSchema` with value schema id is idempotent. For example, even if there are multiple calls to add value schema v15 at the same time, we will end up having only one znode /value-schema/15.

shuhui-liu:
> > > Sorry for the late follow up. Do we have a test to validate the new server behavior when enabling system store schema registration? I didn't see it from this PR.
> > 
> > 
> > No problem. Since VeniceServerWrapper has the following code
> > ```
> > .put(SYSTEM_SCHEMA_INITIALIZATION_AT_START_TIME_ENABLED, true)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > , when any server starts during the integration test, ControllerClientBackedSystemSchemaInitializer will be executed. If something goes wrong during the initialization, server will fail to start and the integration test will fail. Let me know if you think that more testings need to be added.
> 
> So this behavior is being tested implicitly, right? In integration test, normally, we will spin up the Controller first and then other components, that means Venice Server won't execute the schema registration code path (very likely). Is it possible to explicitly disable system store schema registration in Controller and exercise the Server registration path?

I see what you mean. That's a good point. I added a new integ test VeniceServerTest#testStartServerWithSystemSchemaInitialization which disables controller meta system store initialization, lets server execute the code path, verifies that the system store is created and has the right number of schemas.

gaojieliu (approved):
Thanks!

PR #532, opened 2023-07-13
[fc][test] update health data for batch gets and fix metrics issues

m-nagarajan:
## Summary
1. FC batch get uses `instanceHealthMonitor` to check for healthy instances before sending our the post request, but doesn't update the health monitor if an instance ends up being healthy/unhealthy: This change reuses the methods which the single get uses and its counted per route requests, not per key.
2. FC (both single/batch gets) doesn't increment metrics when the instance is not reachable: Currently in such scenarios, a future is completed based on `routingLeakedRequestCleanupThresholdMS` to update the health data, but it doesn't complete the value future and the stats future. Made changes such that we end up completing these futures and added tests. 
3. When both original and retry requests fail and the stats future is completed, only the unhealthy count and latency metric is incremented, not providing any insights into the reasons (eg: noAvailableReplica) and other insights like whether retries happened in the request. Modified that flow to increment a few more metrics for such failure cases. 
4. Even though batchGet uses `streamingBatchGet` (needs to be enabled for now), it still returns `CompletableFuture<Map<K, V>>` and if one of the routes is taking a long time, it still slows down the entire batchGet and currently returns an exception if the returned results is partial results. 

## How was this PR tested?
GH CI


clients/venice-client/src/main/java/com/linkedin/venice/fastclient/RetriableAvroGenericStoreClient.java
```diff
@@ -198,7 +199,8 @@ public void onCompletion(Optional<Exception> exception) {
         if (exception.isPresent()) {
           streamingResponseFuture.completeExceptionally(exception.get());
         } else {
-          streamingResponseFuture.complete(new VeniceResponseMapImpl<>(valueMap, nonExistingKeys, true));
+          streamingResponseFuture
+              .complete(new VeniceResponseMapImpl<>(valueMap, nonExistingKeys, valueMap.size() == keys.size()));
```
m-nagarajan, line 203: Even though batchGet uses `streamingBatchGet` (needs to be enabled for now), `batchGet` API still returns `CompletableFuture<Map<K, V>>` and if one of the routes takes a long time, it still slows down the entire batchGet before being used and currently returns an exception if the returned results is partial. Is my understanding of the API correct here? @xunyin8

xunyin8: You meant server instead of router right? Good find, I think this is just an incomplete implementation of stream batch get in FC. We want to support partial batch get result just like what we do in thin-client.

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/DispatchingAvroGenericStoreClient.java
```diff
@@ -467,6 +481,10 @@ private void transportRequestCompletionHandler(
     if (exception != null) {
```
xunyin8, line 481: This used to only skip the transport client for one route and set the partial response exception flag. The overall request will have an optional exception set in the `whenComplete` callback. 
```
callback.onCompletion(Optional.of(new VeniceClientException("At least one route did not complete", clientException)));
```
Based on this and `BatchGetRequestContext`'s implementation I don't think one route will slowdown the entire request but we can sync up offline to resolve the confusion.

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/StatsAvroGenericStoreClient.java
```diff
@@ -155,24 +155,16 @@ private <R> CompletableFuture<R> recordRequestMetrics(
...
-        if (throwable instanceof VeniceClientException) {
-          throw (VeniceClientException) throwable;
-        } else {
-          throw new VeniceClientException(throwable);
-        }
+        exceptionReceived = true;
       }
 
-      if (latency > TIMEOUT_IN_SECOND * Time.MS_PER_SECOND) {
+      if (exceptionReceived || (latency > TIMEOUT_IN_SECOND * Time.MS_PER_SECOND)) {
```
xunyin8, line 167: Good fix, I'm surprised that we used to not count requests that throws exceptions as unhealthy requests.

xunyin8 (commented):
Thanks for the metric and instance health monitor fix but let's double check the streaming behavior. It's important for partial response to work for batch-get

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/StatsAvroGenericStoreClient.java
m-nagarajan: We were counting that. But just for unhealthy request count and unhealthy latency. We were not counting it any retry related metrics, etc in such scenarios. So the overview of what was happening in the system felt incomplete, so incrementing more metrics in failure cases as well.

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/RetriableAvroGenericStoreClient.java
m-nagarajan: as per the discussion yesterday, batchGet() only returns full response or exception. So its inline.

xunyin8 (dismissed):
Thanks Manoj

xunyin8 (approved):

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/fastclient/utils/AbstractClientEndToEndSetup.java
```diff
@@ -496,29 +498,40 @@ protected AvroSpecificStoreClient<String, TestValueSchema> getSpecificThinClient
...
     double successKeyCount = useStreamingBatchGetAsDefault ? expectedBatchGetKeySizeSuccessMetricsCount : 1;
     Map<String, ? extends Metric> metrics = metricsRepository.metrics();
+
     // counters are incremented in an async manner, so adding non-deterministic wait
-    TestUtils.waitForNonDeterministicAssertion(10, TimeUnit.SECONDS, () -> {
+    TestUtils.waitForNonDeterministicAssertion(5, TimeUnit.SECONDS, () -> {
+      assertTrue(metrics.get(metricPrefix + "request.OccurrenceRate").value() > 0);
+      assertTrue(metrics.get(metricPrefix + "healthy_request.OccurrenceRate").value() > 0);
+      assertTrue(metrics.get(metricPrefix + "healthy_request_latency.Avg").value() > 0);
+      assertFalse(metrics.get(metricPrefix + "unhealthy_request.OccurrenceRate").value() > 0);
```
FelixGV, line 521: I see CI failures on this assertion... not sure what's the right way to go about it... should we expect zero unhealthy requests? Or is it acceptable to have some low rate of unhealthy requests? Is there a retry mechanism where some request could fail, get retried, and then succeed? And if that happens, would it get counted here (even though things worked out in the end)?

PR #533, opened 2023-07-15
[server] Integrated with RocksDB multiget API and benchmark

gaojieliu:
RocksDB offers a set of multiget API besides single-get API, and it claims that it should improve the performance by parallelizing the IO operations with linux kernal 5.1 or later.
We also want to benchmark these APIs for PT format since in theory, it should reduce the cost of JNI because of less JNI calls.

Based on the benchmarking, the performance of multiget API is performing worse than single-get API (which is weird), and we would like to keep these benchmarks and code, so that we can evaluate BT format in the future.

PT benchmark result (subset):
Benchmark                                                   (BATCH_SIZE)  Mode  Cnt      Score      Error  Units
RocksDBLookupApiBenchmark.measureMultiGetAPI                           1  avgt    5  18195.987 ± 1484.887  ms/op
RocksDBLookupApiBenchmark.measureMultiGetAPI                           2  avgt    5  16774.263 ±  212.736  ms/op
RocksDBLookupApiBenchmark.measureMultiGetAPI                          10  avgt    5  18206.964 ± 3331.436  ms/op
RocksDBLookupApiBenchmark.measureMultiGetAPIWithByteBuffer             1  avgt    5  27395.865 ±  688.723  ms/op
RocksDBLookupApiBenchmark.measureMultiGetAPIWithByteBuffer             2  avgt    5  23750.618 ± 4607.772  ms/op
RocksDBLookupApiBenchmark.measureMultiGetAPIWithByteBuffer            10  avgt    5  15060.939 ±  941.485  ms/op
RocksDBLookupApiBenchmark.measureSingleGetAPI                          1  avgt    5   8542.114 ±  378.653  ms/op

## How was this PR tested?

CI

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBStoragePartition.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBStoragePartition.java
@@ -15,7 +15,9 @@ import java.nio.ByteBuffer;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.Iterator;
 import java.util.List;
+import java.util.ListIterator;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
@@ -26,6 +28,7 @@ import org.apache.commons.io.FileUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.rocksdb.BlockBasedTableConfig;
+import org.rocksdb.ByteBufferGetStatus;
 import org.rocksdb.Cache;
 import org.rocksdb.ColumnFamilyDescriptor;
 import org.rocksdb.ColumnFamilyHandle;
@@ -44,6 +47,7 @@ import org.rocksdb.Slice;
 import org.rocksdb.SstFileManager;
 import org.rocksdb.SstFileWriter;
 import org.rocksdb.Statistics;
+import org.rocksdb.Status;
 import org.rocksdb.WriteOptions;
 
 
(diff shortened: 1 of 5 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBStoragePartition.java
```diff
@@ -548,6 +552,79 @@ public byte[] get(ByteBuffer keyBuffer) {
...
+            }
+            ByteBuffer newValueBuffer = ByteBuffer.allocateDirect(newValue.length);
+            newValueBuffer.put(newValue);
+            newValueBuffer.flip();
+            resultList.add(newValueBuffer);
+            valueIter.set(newValueBuffer);
+          }
+        } else if (statusCode.equals(Status.Code.NotFound)) {
+          resultList.add(null);
+        } else {
```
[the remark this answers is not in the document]
gaojieliu: For other codes, we don't know the exact ways to handle them, so it is better to error out to catch all the potential issues.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBStoragePartition.java
```diff
@@ -548,6 +552,79 @@ public byte[] get(ByteBuffer keyBuffer) {
...
+    } finally {
+      readCloseRWLock.readLock().unlock();
+    }
+  }
+
+  public List<ByteBuffer> multiGet(List<ByteBuffer> keys, List<ByteBuffer> values) {
+    readCloseRWLock.readLock().lock();
+    try {
+      makeSureRocksDBIsStillOpen();
+      List<ByteBufferGetStatus> statusList = rocksDB.multiGetByteBuffers(keys, values);
```
majisourav99, line 571: This can ` * @throws IllegalArgumentException thrown if the number of passed keys, passed values and
   * passed column family handles do not match.` Can we add some assert before this call or catch the exception ?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBStoragePartition.java
```diff
@@ -548,6 +552,79 @@ public byte[] get(ByteBuffer keyBuffer) {
...
+
+      List<ByteBuffer> resultList = new ArrayList<>(keys.size());
+      Iterator<ByteBufferGetStatus> statusIter = statusList.iterator();
+      Iterator<ByteBuffer> keyIter = keys.iterator();
+      ListIterator<ByteBuffer> valueIter = values.listIterator();
+      while (keyIter.hasNext()) {
+        if (!statusIter.hasNext() || !valueIter.hasNext()) {
```
anaberezhnov, line 578: This check can be done once outside of the loop.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBStoragePartition.java
```diff
@@ -548,6 +552,79 @@ public byte[] get(ByteBuffer keyBuffer) {
...
+    }
+  }
+
+  public List<ByteBuffer> multiGet(List<ByteBuffer> keys, List<ByteBuffer> values) {
+    readCloseRWLock.readLock().lock();
+    try {
+      makeSureRocksDBIsStillOpen();
+      List<ByteBufferGetStatus> statusList = rocksDB.multiGetByteBuffers(keys, values);
+
+      List<ByteBuffer> resultList = new ArrayList<>(keys.size());
```
anaberezhnov, line 573: It should be ok to reuse `values` and avoid excessive allocation, at least in the scenario when value buffers are sized probably. But even in general case it might be ok to modify `values` list since most likely the caller allocates a new list each time it calls the api.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBStoragePartition.java
```diff
@@ -548,6 +552,79 @@ public byte[] get(ByteBuffer keyBuffer) {
...
+            resultList.add(value);
+          } else {
+            // Need to look it up again since the passed buffer is too small
+            byte[] newValue;
+            if (key.isDirect()) {
+              byte[] keyBytes = new byte[key.remaining()];
+              key.mark();
+              key.get(keyBytes, 0, key.remaining());
+              key.reset();
+              newValue = get(keyBytes);
```
anaberezhnov, line 601: My understanding that the buffer is allocated only because there is no `RocksDB::get` overload that takes `ByteBuffer`? It can be expensive especially if this is not a one off large value. Depending on the number of failed keys it might be faster to retry `multiGetByteBuffers` for them. What do you think?

anaberezhnov: Also since this fallback is expensive, it might be worth having a metric to detect if we're losing performance here.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBStoragePartition.java
gaojieliu: RocksDB internally will do all the checks:
https://github.com/facebook/rocksdb/blob/main/java/src/main/java/org/rocksdb/RocksDB.java#L2347
That is the reason I don't want to perform the same check again here.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBStoragePartition.java
gaojieliu: Fallback is expensive and in practice, it should rarely happen since by default, I think we will allocate 1MB buffer and metrics can be expensive in RocksDB layer.
That is the same behavior in singel-get API.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBStoragePartition.java
gaojieliu: Sure

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBStoragePartition.java
gaojieliu: The passed `values` param is for reuse purpose, so I don't think customer will allocate a new list every time and that was also the reason to construct a new list, which should be cheap since all the elements are referring to the populated buffer.

anaberezhnov (approved):

majisourav99 (approved):

PR #534, opened 2023-07-17
[client][router][server] Add client-side computation support

anaberezhnov:
Read-compute functionality was available to select users and required support from service operators to ensure safety and stability for all users. This change makes the read-compute interface available to all users by leveraging batch-get and local computations by default. This allows service operators to enable/disable remote computations without disrupting user applications.

The order of operations is as follows:
1. Upon start, the client sends a compute request and sets a new HTTP header VENICE_CLIENT_COMPUTE to indicate that client-side computation is supported.
2. The router checks store config, and when remote computation is off, fails the request if client-computation is not available. Otherwise, the router converts the compute request into a batch-get.
3. The client checks HTTP headers to resolve what kind of response it got, and if necessary performs computations locally.
4. The following computation requests are converted to batch-get on the client side to avoid extra work and sending extra data over the network.
5. Batch-get response always carries a header indicating the latest availability of remote computation so that the client can switch back and forth if needed (e.g., when service operator makes changes to store config).

After deployment recommendation:
- Monitor correlation of `request_parse_latency` and `multiget_fallback` metrics on the router.
- Monitor `multiget_fallback` metric on the client to confirm that client-computation is happening.

- Moved common computation logic from server and da-vinci-client to `ComputeUtils` to reduce code duplication
- Merged `com.linkedin.venice.utils.ComputeUtils` and `com.linkedin.venice.compute.ComputeOperationUtils` into
`com.linkedin.venice.compute.ComputeUtils`
- Moved `AbstractAvroStoreClient:: StoreClientStreamingCallback ` out to reduce client complexity and maintain independency between transport and client logic. The new decoder layer is responsible for deserializing records from stream of bytes
- Added `storage_engine_read_compute_efficiency` server metric to help monitor data compression due to read compute
- Added `multiget_fallback` metric to both client and router to monitor client-computation activity

## How was this PR tested?

- Updated  `TestRead`, `TestStreaming`, `TestVeniceResponseAggregator` suites
- Added `TestVeniceComputePath::testToMultiGetPath` test
- Internal CI

`AvroGenericStoreClient::compute` will succeed by default, regardless of whether Read-Compute feature is enabled on the service side or not. Extremely latency-sensitive users who expect Read-Compute to be enabled may disable client-computation completely by calling`ClientConfig.setRemoteComputationOnly(true)`


internal/venice-common/src/main/java/com/linkedin/venice/utils/AvroRecordUtils.java
```diff
@@ -84,4 +85,11 @@ public static <T extends SpecificRecord> T prefillAvroRecordWithDefaultValue(T r
 
     return recordType;
   }
+
+  public static void clearRecord(GenericRecord record) {
+    int fieldCount = record.getSchema().getFields().size();
+    for (int i = 0; i < fieldCount; ++i) {
+      record.put(i, null);
```
huangminchn, line 92: What if one field is primitive type?

clients/venice-thin-client/src/main/java/com/linkedin/venice/client/store/streaming/ClientComputeRecordStreamDecoder.java
```diff
@@ -0,0 +1,78 @@
...
+      StreamingCallback trackingCallback,
+      Executor deserializationExecutor,
+      RecordDeserializer<StreamingFooterRecordV1> streamingFooterRecordDeserializer,
+      Supplier<RecordDeserializer<V>> computeDeserializerProvider,
+      Function<Integer, RecordDeserializer<V>> valueDeserializerProvider,
+      BiFunction<CompressionStrategy, ByteBuffer, ByteBuffer> decompressor) {
+    this.decoderProvider = headers -> {
+      if (headers.containsKey(HttpConstants.VENICE_CLIENT_COMPUTE)) {
+        callback.onRemoteComputeStateChange("0".equals(headers.get(HttpConstants.VENICE_CLIENT_COMPUTE)));
+        return new MultiGetRecordStreamDecoder<>(keyList, new StreamingCallback<K, V>() {
```
huangminchn, line 35: I find this part a bit confusing. It appears that as long as the response has header "VENICE_CLIENT_COMPUTE", we expect the responses are multi-get results, regarding of the header values; but instead of always setting false in "onRemoteComputeStateChange", we check whether it's equal to "0".

clients/venice-thin-client/src/main/java/com/linkedin/venice/client/store/AbstractAvroStoreClient.java
```diff
@@ -657,59 +634,96 @@ public void compute(
...
+        remoteComputationAllowed.set(enabled);
+      }
+
+      @Override
+      public void onCompletion(Optional<Exception> exception) {
+        callback.onCompletion(exception);
+      }
+    };
+
+    TransportClientStreamingCallback decoder = new ClientComputeRecordStreamDecoder<K, GenericRecord>(
```
huangminchn, line 675: Why a callback is named as decoder?

huangminchn (commented):
I want to publish what I have so far to make incremental progress.

internal/venice-common/src/main/java/com/linkedin/venice/utils/AvroRecordUtils.java
anaberezhnov: I borrowed this implementation from `StorageReadRequestsHandler::clearFieldsInReusedRecord`. GenericRecord internally is like a `Map<String, Object>`, so it's correct. It won't work for SpecificRecord however. Btw, I'm adding unit-test to verify that it works as I expect.

clients/venice-thin-client/src/main/java/com/linkedin/venice/client/store/streaming/ClientComputeRecordStreamDecoder.java
anaberezhnov: This is because only multi-get sets `VENICE_CLIENT_COMPUTE` header. When this header is used in the response, it indicates the recommended compute processing mode. This way clients who were doing client-compute using multi-get can detect a new recommendation to switch to remote computation when Venice service operator turns on read compute for the store.

clients/venice-thin-client/src/main/java/com/linkedin/venice/client/store/AbstractAvroStoreClient.java
anaberezhnov: To indicate the function of it. Decoder encapsulates client-router record-based communication protocol, i.e it transforms a stream of bytes into records that client can operate on. Decoder implements callback interface, so that it can be used with transport client.

Another reason is that there are several callbacks already in this method scope and it helps to distinguish between them.

clients/venice-thin-client/src/main/java/com/linkedin/venice/client/stats/ClientStats.java
```diff
@@ -109,7 +111,7 @@ protected ClientStats(MetricsRepository metricsRepository, String storeName, Req
     retryKeySuccessRatioSensor = registerSensor(
         "retry_key_success_ratio",
         new TehutiUtils.SimpleRatioStat(retryRequestSuccessKeyCount, getSuccessRequestKeyCountRate()));
-
+    multiGetFallbackSensor = registerSensor("multiget_fallback", new Total(), new OccurrenceRate());
```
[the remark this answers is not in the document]
anaberezhnov: Will do, thanks!

internal/venice-client-common/src/main/java/com/linkedin/venice/compute/ComputeUtils.java
```diff
@@ -0,0 +1,344 @@
...
+import org.apache.avro.generic.GenericRecord;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+
+/**
+ * This class provides utilities for float-vector operations, and it also handles {@link PrimitiveFloatList}
+ * transparently to the user of this class.
+ */
+public class ComputeUtils {
```
gaojieliu, line 36: So this file is just renamed, no change, right?

gaojieliu (changes requested):
Looks good overall, thanks @anaberezhnov !
Left some minor comments.

huangminchn (commented):
Looks good overall! The offline syncup really helps me with understanding the changes. Could you please also add the same explanation in the PR description? Basically, call out the difference between transport client decode and client compute callback, and how they depend on each other.

Besides, what is our plan for fast client?

services/venice-server/src/main/java/com/linkedin/venice/stats/ServerHttpRequestStats.java
```diff
@@ -250,6 +251,13 @@ public ServerHttpRequestStats(
             getName(),
             getFullMetricName("storage_engine_read_compute_serialization_latency")));
 
+    String readComputeEfficiencySensorName = "storage_engine_read_compute_efficiency";
+    readComputeEfficiencySensor = registerPerStoreAndTotal(
+        readComputeEfficiencySensorName,
+        totalStats,
+        () -> totalStats.readComputeEfficiencySensor,
+        TehutiUtils.getPercentileStatWithAvgAndMax(getName(), getFullMetricName(readComputeEfficiencySensorName)));
```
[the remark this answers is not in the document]
anaberezhnov: I don't think so, average should be enough. Good catch!

services/venice-router/src/main/java/com/linkedin/venice/router/api/path/VeniceComputePath.java
```diff
@@ -150,6 +150,29 @@ public final RequestType getStreamingRequestType() {
     return RequestType.COMPUTE_STREAMING;
   }
 
+  public VeniceMultiGetPath toMultiGetPath() {
```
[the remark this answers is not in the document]
anaberezhnov: Cost wise this is similar to sub-path creation during normal Router operation, i.e not very expensive. Also, this code path is only active during short periods of time when client and service is out of sync regarding the state of Read-Compute. This I expect the overall overhead to be low.

Regarding the possible optimization we can deprecate `ComputeRouterRequestKeyV1` and use `MultiGetRouterRequestKeyV1` instead. It'd even be a backward compatible change since both records are binary compatible. The downside is that such change would tie multi-get and compute transport protocols.

clients/venice-thin-client/src/main/java/com/linkedin/venice/client/store/AbstractAvroStoreClient.java
```diff
@@ -647,7 +624,7 @@ public ComputeRequestBuilder<K> compute(
 
   @Override
   public void compute(
-      ComputeRequestWrapper computeRequestWrapper,
+      ComputeRequestWrapper computeRequest,
       Set<K> keys,
       Schema resultSchema,
       StreamingCallback<K, ComputeGenericRecord> callback,
```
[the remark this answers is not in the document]
anaberezhnov: `compute` is a public api and there is no other callbacks in the parameters, so it unambiguous already.

clients/venice-thin-client/src/main/java/com/linkedin/venice/client/store/AbstractAvroStoreClient.java
```diff
@@ -657,59 +634,96 @@ public void compute(
...
-            },
-            envelope -> envelope.keyIndex,
-            envelope -> streamingFooterRecordDeserializer.deserialize(envelope.value)),
-        keyList.size());
+    int schemaId = getSchemaReader().getValueSchemaId(computeRequest.getValueSchema());
+    headers.put(VENICE_KEY_COUNT, Integer.toString(keyList.size()));
+    headers.put(VENICE_COMPUTE_VALUE_SCHEMA_ID, Integer.toString(schemaId));
+    if (!clientConfig.isRemoteComputationOnly()) {
+      headers.put(VENICE_CLIENT_COMPUTE, "1");
+    }
```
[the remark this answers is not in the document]
anaberezhnov: Correct, the header value of "1" in the request indicates that client supports client-computation and is ready to receive multi-get response if remote-computation cannot be performed for whatever reason (store config, peak load, etc).

In certain cases, the latency difference between remote and client computation can be dramatic. I've added `isRemoteComputationOnly` so that use-cases that enabled remote-computation can detect service operator mistakes quickly.

clients/venice-thin-client/src/main/java/com/linkedin/venice/client/store/AbstractAvroStoreClient.java
```diff
@@ -657,59 +634,96 @@ public void compute(
...
+        keyList,
+        clientComputeCallback,
+        callback,
+        getDeserializationExecutor(),
+        streamingFooterRecordDeserializer,
+        () -> getComputeResultRecordDeserializer(resultSchema),
+        schemaId -> (RecordDeserializer<GenericRecord>) getDataRecordDeserializer(schemaId),
+        this::decompressRecord);
+
+    if (clientConfig.isRemoteComputationOnly() || remoteComputationAllowed.get()) {
```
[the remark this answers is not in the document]
anaberezhnov: In certain cases, the latency difference between remote and client computation can be dramatic. I've added `isRemoteComputationOnly` so that use-cases that enabled remote-computation can detect service operator mistakes quickly.

internal/venice-client-common/src/main/java/com/linkedin/venice/compute/ComputeUtils.java
anaberezhnov: Yes, I just merged `com.linkedin.venice.utils.ComputeUtils` and `com.linkedin.venice.compute.ComputeOperationUtils` plus added `computeResult`, which is based on original implementation in `StorageReadRequestsHandler`.

services/venice-server/src/main/java/com/linkedin/venice/listener/StorageReadRequestsHandler.java
```diff
@@ -133,19 +120,47 @@ public class StorageReadRequestsHandler extends ChannelInboundHandlerAdapter {
...
+  private static class ReusableObjects {
+    /**
+     * When constructing a {@link BinaryDecoder}, we pass in this 16 bytes array because if we pass anything
+     * less than that, it would end up getting discarded by the ByteArrayByteSource's constructor, a new byte
+     * array created, and the content of the one passed in would be copied into the newly constructed one.
+     * Therefore, it seems more efficient, in terms of GC, to statically allocate a 16 bytes array and keep
+     * re-using it to construct decoders. Since we always end up re-configuring the decoder and not actually
+     * using its initial value, it shouldn't cause any issue to share it.
+     */
+    private static final byte[] BINARY_DECODER_PARAM = new byte[16];
```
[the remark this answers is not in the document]
anaberezhnov: To be clear this wasn't changed in the PR. It could be ok to share it across threads according to the comment, but from `BinaryDecoder` interface perspective it's not clear that it will be ignored. Thus I suggest to leave it as is.

clients/venice-thin-client/src/main/java/com/linkedin/venice/client/store/AbstractAvroStoreClient.java
```diff
@@ -657,59 +634,96 @@ public void compute(
...
-                  computeRequestWrapper.getValueSchema());
-            },
-            envelope -> envelope.keyIndex,
-            envelope -> streamingFooterRecordDeserializer.deserialize(envelope.value)),
-        keyList.size());
+    int schemaId = getSchemaReader().getValueSchemaId(computeRequest.getValueSchema());
+    headers.put(VENICE_KEY_COUNT, Integer.toString(keyList.size()));
+    headers.put(VENICE_COMPUTE_VALUE_SCHEMA_ID, Integer.toString(schemaId));
+    if (!clientConfig.isRemoteComputationOnly()) {
+      headers.put(VENICE_CLIENT_COMPUTE, "1");
```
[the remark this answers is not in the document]
anaberezhnov: Yeah, it may reduce GC a little bit, but I prefer to do it as a separate PR, since this one is already substantial.

clients/venice-thin-client/src/main/java/com/linkedin/venice/client/stats/ClientStats.java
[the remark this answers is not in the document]
anaberezhnov: The sensor tracks number of keys handled via multiget-fallback (i.e client-compute) mechanism. Batch sizes can vary greatly, so key-granularity should give more sense about the actual load.

clients/venice-thin-client/src/main/java/com/linkedin/venice/client/store/AbstractAvroStoreClient.java
```diff
@@ -657,59 +634,96 @@ public void compute(
...
-                return null;
-              }
-              return new ComputeGenericRecord(
-                  computeResultRecordDeserializer.deserialize(envelope.value),
-                  computeRequestWrapper.getValueSchema());
-            },
-            envelope -> envelope.keyIndex,
-            envelope -> streamingFooterRecordDeserializer.deserialize(envelope.value)),
-        keyList.size());
+    int schemaId = getSchemaReader().getValueSchemaId(computeRequest.getValueSchema());
```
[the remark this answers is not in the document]
anaberezhnov: Thank you for the suggestion! This is original compute logic. I prefer to address this optimizations in a separate PR, since this one is already substantial.

clients/venice-thin-client/src/main/java/com/linkedin/venice/client/store/StatTrackingStoreClient.java
```diff
@@ -252,21 +254,21 @@ public void compute(
   private static void handleMetricTrackingForStreamingCallback(
       ClientStats clientStats,
       long startTimeInNS,
-      Optional<VeniceClientException> veniceException,
+      Optional<Exception> exception,
```
majisourav99, line 257: Can we just use nullable Exception instead of Optional ?

clients/venice-thin-client/src/main/java/com/linkedin/venice/client/store/streaming/AbstractRecordStreamDecoder.java
```diff
@@ -0,0 +1,239 @@
...
+                .recordRequestSubmissionToResponseHandlingTime(LatencyUtils.getLatencyInMS(preSubmitTimeInNS)));
+    isStreamingResponse = headers.containsKey(HttpConstants.VENICE_STREAMING_RESPONSE);
+    String schemaIdHeader = headers.get(HttpConstants.VENICE_SCHEMA_ID);
+    if (schemaIdHeader != null) {
+      responseSchemaId = Integer.parseInt(schemaIdHeader);
+    }
+    envelopeDeserializer = getEnvelopeDeserializer(responseSchemaId);
+    String compressionHeader = headers.get(HttpConstants.VENICE_COMPRESSION_STRATEGY);
+    if (compressionHeader != null) {
+      compressionStrategy = CompressionStrategy.valueOf(Integer.parseInt(compressionHeader));
```
majisourav99, line 84: can we handle illegal argument exception from `valueOf` here ?

clients/venice-thin-client/src/main/java/com/linkedin/venice/client/store/StatTrackingStoreClient.java
anaberezhnov: `Optional` provides more safety and minimal overhead in this particular case, since Exceptions are rare and `Optional.empty()` is free

clients/venice-thin-client/src/main/java/com/linkedin/venice/client/store/streaming/AbstractRecordStreamDecoder.java
anaberezhnov: The exception is caught in `TransportClient::streamPost`, so there is no gap here. In terms of handling, not much we can do if we get data than cannot be understood besides from failing the request.

majisourav99: I believe failing the request in that case is good idea?

clients/venice-thin-client/src/main/java/com/linkedin/venice/client/store/AbstractAvroStoreClient.java
anaberezhnov: Replaced `TransportClientStreamingCallback` with `RecordSteamDecoder` interface here to reduce confusion.

clients/venice-thin-client/src/main/java/com/linkedin/venice/client/store/AbstractAvroStoreClient.java
```diff
@@ -819,234 +819,47 @@ private ByteBuffer decompressRecord(CompressionStrategy compressionStrategy, Byt
...
-                t -> t.onDeserializationCompletion(finalCompletedException, successfulKeyCnt.get(), duplicateEntryCnt));
-            clientStats.ifPresent(
-                stats -> stats.recordResponseDeserializationTime(
-                    LatencyUtils.convertLatencyFromNSToMS(deserializationTimeInNS.sum())));
-          });
+    List<K> keyList = new ArrayList<>(keys);
+    TransportClientStreamingCallback decoder = new MultiGetRecordStreamDecoder<>(
+        keyList,
+        callback,
+        callback,
```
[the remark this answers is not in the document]
anaberezhnov: It was necessary because TrackingStreamingCallback didn't support callback composition, and not passing caller provided callback would disable metric collection at the decoder level.

Refactored the code in the latest commit to avoid passing callback twice. Thanks for the feedback!

clients/venice-thin-client/src/main/java/com/linkedin/venice/client/store/AbstractAvroStoreClient.java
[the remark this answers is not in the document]
anaberezhnov: I understand now what you've been asking for after checking the code again.  I've addressed this in the latest commit by reducing number of different callbacks used in the method to just two.

services/venice-router/src/main/java/com/linkedin/venice/router/api/path/VeniceComputePath.java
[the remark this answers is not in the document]
anaberezhnov: I'm open to benchmark it with JMH and implement this optimization in a separate PR. @gaojieliu what do you think about it?

clients/venice-thin-client/src/main/java/com/linkedin/venice/client/store/streaming/AbstractRecordStreamDecoder.java
anaberezhnov: That's exactly what happens in this case, any unhandled exception fail the request.

anaberezhnov: One clarification, `AbstractRecordStreamDecoder` is not exactly a new code. I copied it almost as-is from `AbstractAvroStoreClient` as part of refactoring.

services/venice-router/src/main/java/com/linkedin/venice/router/api/path/VeniceComputePath.java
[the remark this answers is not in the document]
gaojieliu: As @anaberezhnov mentioned, this function won't be executed in the hot path, so I don't have a strong preference.

gaojieliu (approved):
Thanks a lot for the change!

Please cut a ticket to track the optimizations we discovered during code review as a follow-up.

PR #535, opened 2023-07-17
[server][dvc] Delete deprecated chunks when processing partial update

sixpluszero:
## [server][dvc] Delete deprecated chunks when processing partial update
This PR aims to clean up deprecated value and RMD chunks to avoid storage engine chunking leaks. When we processed a partial update request, it will also try to look up old value and RMD manifest and will produce DELETE messages to old value and RMD chunks to remove them.
Note:
1. As discussed offline, this solution will be a short term solution to clean up disk space, but it will not clean up Kafka version topic garbage chunks as even after log compression there will be one DELETE message left per deprecated chunk key, but we think this is acceptable at this stage as DELETE message is considered small and we can work on long term complete solution later.
2. Based on current implementation, we could not fully delete a RMD of a key. We can only insert an empty RMD into it. (If this is not acceptable, we will need to add new implementations to execute real RocksDB delete on RMD Column Family.
3. We refactor the chunking get method to store the retrieved value's manifest if it is chunked. That means for AASIT, we will always delete old RMD chunks (as it is always fetched for every operation) and we will do best effort to delete old value's RMD (in theory, we will always do it for UPDATE). For LFSIT, we will always delete value chunks for UPDATE operation.

## How was this PR tested?
Added new integration test logic to validate existing value and RMD chunks are deleted after a new UPDATE request.

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
@@ -3,6 +3,7 @@ package com.linkedin.davinci.kafka.consumer;
 import static com.linkedin.davinci.kafka.consumer.LeaderFollowerStateType.LEADER;
 import static com.linkedin.davinci.kafka.consumer.LeaderFollowerStateType.STANDBY;
 import static com.linkedin.venice.VeniceConstants.REWIND_TIME_DECIDED_BY_SERVER;
+import static com.linkedin.venice.writer.VeniceWriter.APP_DEFAULT_LOGICAL_TS;
 
 import com.linkedin.avroutil1.compatibility.AvroCompatibilityHelper;
 import com.linkedin.davinci.config.VeniceStoreVersionConfig;
@@ -52,7 +53,6 @@ import com.linkedin.venice.writer.ChunkAwareCallback;
 import com.linkedin.venice.writer.DeleteMetadata;
 import com.linkedin.venice.writer.LeaderMetadataWrapper;
 import com.linkedin.venice.writer.PutMetadata;
-import com.linkedin.venice.writer.VeniceWriter;
 import java.nio.ByteBuffer;
 import java.util.ArrayList;
 import java.util.Collections;
@@ -670,7 +670,10 @@ public class ActiveActiveStoreIngestionTask extends LeaderFollowerStoreIngestion
                   key,
                   callback,
                   sourceTopicOffset,
-                  new DeleteMetadata(valueSchemaId, rmdProtocolVersionID, updatedRmdBytes));
+                  APP_DEFAULT_LOGICAL_TS,
+                  new DeleteMetadata(valueSchemaId, rmdProtocolVersionID, updatedRmdBytes),
+                  oldValueManifest,
+                  oldRmdManifest);
       LeaderProducedRecordContext leaderProducedRecordContext =
           LeaderProducedRecordContext.newDeleteRecord(kafkaClusterId, consumerRecord.getOffset(), key, deletePayload);
       produceToLocalKafka(
@@ -1402,7 +1405,7 @@ public class ActiveActiveStoreIngestionTask extends LeaderFollowerStoreIngestion
               valueSchemaId,
               callback,
               leaderMetadataWrapper,
-              VeniceWriter.APP_DEFAULT_LOGICAL_TS,
+              APP_DEFAULT_LOGICAL_TS,
               new PutMetadata(getRmdProtocolVersionID(), updatedRmdBytes),
               oldValueManifest,
               oldRmdManifest);
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
@@ -15,7 +15,6 @@ import com.linkedin.davinci.replication.merge.RmdSerDe;
 import com.linkedin.davinci.replication.merge.StringAnnotatedStoreSchemaCache;
 import com.linkedin.davinci.stats.AggVersionedIngestionStats;
 import com.linkedin.davinci.storage.chunking.ChunkedValueManifestContainer;
-import com.linkedin.davinci.storage.chunking.ChunkingUtils;
 import com.linkedin.davinci.storage.chunking.RawBytesChunkingAdapter;
 import com.linkedin.davinci.storage.chunking.SingleGetChunkingAdapter;
 import com.linkedin.davinci.store.cache.backend.ObjectCacheBackend;
@@ -506,7 +505,6 @@ public class ActiveActiveStoreIngestionTask extends LeaderFollowerStoreIngestion
       producePutOrDeleteToKafka(
           mergeConflictResult,
           partitionConsumptionState,
-          msgType,
           keyBytes,
           consumerRecord,
           subPartition,
@@ -630,7 +628,6 @@ public class ActiveActiveStoreIngestionTask extends LeaderFollowerStoreIngestion
   private void producePutOrDeleteToKafka(
       MergeConflictResult mergeConflictResult,
       PartitionConsumptionState partitionConsumptionState,
-      MessageType msgType,
       byte[] key,
       PubSubMessage<KafkaKey, KafkaMessageEnvelope, Long> consumerRecord,
       int subPartition,
@@ -649,10 +646,6 @@ public class ActiveActiveStoreIngestionTask extends LeaderFollowerStoreIngestion
     GenericRecord rmdRecord = mergeConflictResult.getRmdRecord();
     final ByteBuffer updatedRmdBytes =
         rmdSerDe.serializeRmdRecord(mergeConflictResult.getValueSchemaId(), mergeConflictResult.getRmdRecord());
-    // Since data is not chunked in RT but chunked in VT, creating the key for the small record case or CVM to be
-    // used to persist on disk after producing to Kafka.
-    final byte[] updatedKeyBytes =
-        isChunked ? ChunkingUtils.KEY_WITH_CHUNKING_SUFFIX_SERIALIZER.serializeNonChunkedKey(key) : key;
     // finally produce and update the transient record map.
     if (updatedValueBytes == null) {
       hostLevelIngestionStats.recordTombstoneCreatedDCR();
@@ -722,8 +715,7 @@ public class ActiveActiveStoreIngestionTask extends LeaderFollowerStoreIngestion
       produceToLocalKafka(
           consumerRecord,
           partitionConsumptionState,
-          LeaderProducedRecordContext
-              .newPutRecord(kafkaClusterId, consumerRecord.getOffset(), updatedKeyBytes, updatedPut),
+          LeaderProducedRecordContext.newPutRecord(kafkaClusterId, consumerRecord.getOffset(), key, updatedPut),
           produceToTopicFunction,
           subPartition,
           kafkaUrl,
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
@@ -13,6 +13,7 @@ import com.linkedin.davinci.replication.merge.MergeConflictResult;
 import com.linkedin.davinci.replication.merge.RmdSerDe;
 import com.linkedin.davinci.replication.merge.StringAnnotatedStoreSchemaCache;
 import com.linkedin.davinci.stats.AggVersionedIngestionStats;
+import com.linkedin.davinci.storage.chunking.ChunkedValueManifestContainer;
 import com.linkedin.davinci.storage.chunking.ChunkingUtils;
 import com.linkedin.davinci.storage.chunking.RawBytesChunkingAdapter;
 import com.linkedin.davinci.storage.chunking.SingleGetChunkingAdapter;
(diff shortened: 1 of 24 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
```diff
@@ -615,7 +620,16 @@ private void producePutOrDeleteToKafka(
...
+    // Since data is not chunked in RT but chunked in VT, creating the key for the small record case or CVM to be
+    // used to persist on disk after producing to Kafka.
+    final byte[] updatedKeyBytes =
+        isChunked ? ChunkingUtils.KEY_WITH_CHUNKING_SUFFIX_SERIALIZER.serializeNonChunkedKey(key) : key;
+    final ChunkedValueManifest oldValueManifest = isChunked && msgType.equals(MessageType.UPDATE)
+        ? ChunkingUtils.getChunkValueManifestFromStorage(updatedKeyBytes, subPartition, false, storageEngine)
+        : null;
+    final ChunkedValueManifest oldRmdManifest = isChunked && msgType.equals(MessageType.UPDATE)
+        ? ChunkingUtils.getChunkValueManifestFromStorage(updatedKeyBytes, subPartition, true, storageEngine)
+        : null;
```
gaojieliu, line 632: So for AA use case, RMD value will always be retrieved before executing DCR, I think we should reuse the value retrieved previously.
For UPDATE operation, if RMD permits, the old value will be looked up when executing DCR, so there is a high chance the value manifest is retrieved before this function.
Today, `ChunkingUtils` doesn't expose the details of manifest, so I guess we should update the function interface to include a placeholder, so that internally, `ChunkingUtils` can return the manifest if it is chunked.
This will require a refactoring of `ChunkingUtils`, but I think it is worthwhile considering RocksDB lookup is deemed to be an expensive operation.

gaojieliu: Also based on the comment description, we should only do these lookups when the message type is `UPDATE`, right?
Right now, it will be executed for operation types.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
```diff
@@ -686,6 +694,56 @@ private void producePutOrDeleteToKafka(
...
+              consumerRecord,
+              partitionConsumptionState,
+              rmdChunkDeleteRecordContext,
+              (callback, leaderMetadataWrapper) -> veniceWriter.get()
+                  .deleteDeprecatedChunk(chunkKeyBytes, subPartition, callback, leaderMetadataWrapper, deleteMetadata),
+              subPartition,
+              kafkaUrl,
+              kafkaClusterId,
+              beforeProcessingRecordTimestampNs);
+        }
```
gaojieliu, line 744: I am wondering whether this logic belongs to `VeniceWriter` and `LeaderProducerCallback` or not.
For example, we can add a new param called `byte[] previousChunks` in `VeniceWriter#put` method and VeniceWriter will send these deletion messages to Kafka internally, and in the callback, StoreIngestionTask will produce the chunk deletion messages into buffer drainer.
IMO, this structure makes more sense because of modular structure and reusable, so that we don't need to repeat the same logic in `LeaderFollowerStoreIngestionTask`.
What do you think?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -2816,8 +2817,16 @@ private void handleUpdateRequest(
...
         consumerRecord.getTopicPartition());
 
-    // Apply Write Compute.
     final byte[] updatedValueBytes;
+    // Samza VeniceWriter doesn't handle chunking config properly. It reads chunking config
+    // from user's input instead of getting it from store's metadata repo. This causes SN
+    // to der-se of keys a couple of times.
+    final byte[] updatedKeyBytes =
+        isChunked ? ChunkingUtils.KEY_WITH_CHUNKING_SUFFIX_SERIALIZER.serializeNonChunkedKey(keyBytes) : keyBytes;
+    final ChunkedValueManifest oldValueManifest = isChunked
```
gaojieliu, line 2826: Same comment as `ActiveActiveStoreIngestionTask` since the previous manifest should be already available.

gaojieliu (changes requested):
Regarding the performance concern of the ingestion path, I guess it is fine to keep the current scope, but we should cut tickets to track the follow-up tasks.
From my understanding, large messages in chunk-enabled store are rare, so maybe we can create a RocksDB CF to keep the keys with large messages, and in the ingestion path, Ingestion Task can lookup this small CF to check whether a clean up is required or not for full put/delete.
Lookup in this small CF should be very fast if some specific partition only contains 10s of large records.

Please let me know if you want to chat about the comments.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
sixpluszero: Thanks for the comment, looks like I missed it yesterday. Sure, I was kind of hesitant about this change, but you are right, this should save some more look up.

sixpluszero: But there is one issue - for those value/RMD that's cached in the transient record, we will have to do the RocksDB get then

gaojieliu: I wonder whether we can cache the manifest in the transient record or not.

sixpluszero: Updated the PR, but not yet update the transient record part. Will do later.

sixpluszero: The latest version should trim value/rmd manifest look up for both cached and not-cached code path for AASIT and LFSIT.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
sixpluszero: Sure, let me see if I can do that. 

Previously I was struggling to produce multiple chunk delete within a single VW call as it somehow has issue with Leader produce context and such...but let me try to re-org the code...

sixpluszero: Hmm, in my latest changes, I actually move the for loop logic into LFSIT#produceToLocalKafka. 
But I think what you said makes more sense...let me see how I can move it into VW.
(And I will have to change VW a bit, as it does not have information of rmdProtocolVersionId, which makes it pretty ugly to pass along some static information all the value in the call chain).

Maybe I will change PUT API to carry old manifests, and change ChunkAwareCallback#setChunkingInfo to add two args so that in the LPC it can iterate over them and produce them to Leader drainer as well.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
```diff
@@ -615,13 +639,22 @@ private void producePutOrDeleteToKafka(
     GenericRecord rmdRecord = mergeConflictResult.getRmdRecord();
     final ByteBuffer updatedRmdBytes =
         rmdSerDe.serializeRmdRecord(mergeConflictResult.getValueSchemaId(), mergeConflictResult.getRmdRecord());
-
+    // Since data is not chunked in RT but chunked in VT, creating the key for the small record case or CVM to be
+    // used to persist on disk after producing to Kafka.
+    final byte[] updatedKeyBytes =
+        isChunked ? ChunkingUtils.KEY_WITH_CHUNKING_SUFFIX_SERIALIZER.serializeNonChunkedKey(key) : key;
```
gaojieliu, line 650: This logic seems to be correct, but the implementation is very tricky..
For deletion, we don't use the `updatedKeyBytes` when calling `LeaderProducedRecordContext.newDeleteRecord`, and it works because of this piece of logic in `LeaderProducerCallback`:
https://github.com/linkedin/venice/blob/main/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderProducerCallback.java#L148

For `put` record, this function uses `updatedKeyBytes` when calling `newPutRecord` since the logic in `LeaderProducerCallback` won't update key when `chunkedValueManifest` is `null`, that means for large message, it won't update and for regular-size value, it will update it.
This is very confusing..
I guess we should update it in the following way:
1. In `ActiveActiveStoreIngestionTask`, we always use raw `key`.
2. In `LeaderProducerCallback`, we will always update the `keyBytes` field of `LeaderProducedRecordContext`  if `key` is not `null` (which will be updated if chunking is enabled).


What do you think?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/replication/merge/RmdSerDe.java
```diff
@@ -48,7 +48,9 @@ public RmdSerDe(StringAnnotatedStoreSchemaCache annotatedStoreSchemaCache, int r
    * @return A {@link RmdWithValueSchemaId} object composed by extracting the value schema ID from the
    *    * header of the replication metadata.
    */
-  public RmdWithValueSchemaId deserializeValueSchemaIdPrependedRmdBytes(byte[] valueSchemaIdPrependedBytes) {
+  public void deserializeValueSchemaIdPrependedRmdBytes(
```
gaojieliu, line 51: Why change the signature here?
We are not trying to reuse the passed `RmdWithValueSchemaId`, are we?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/storage/chunking/ChunkedValueManifestContainer.java
```diff
@@ -0,0 +1,19 @@
+package com.linkedin.davinci.storage.chunking;
+
+import com.linkedin.venice.storage.protocol.ChunkedValueManifest;
+
+
+public class ChunkedValueManifestContainer {
```
gaojieliu, line 6: This can be a generic container to keep any type of object, right?
Such as:
```
public class ObjectContainer<T>
```

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
```diff
@@ -615,13 +639,22 @@ private void producePutOrDeleteToKafka(
...
-      partitionConsumptionState
-          .setTransientRecord(kafkaClusterId, consumerRecord.getOffset(), key, valueSchemaId, rmdRecord);
+      partitionConsumptionState.setTransientRecord(
+          kafkaClusterId,
+          consumerRecord.getOffset(),
+          key,
+          valueSchemaId,
+          rmdRecord,
+          oldValueManifest,
+          oldRmdManifest);
```
gaojieliu, line 662: For any operation type, AASIT will alway lookup RMD CF to decide whether the incoming request should be rejected or not, so that means, we can reuse the retrieved rmd manifest for any operation type to do the clean up in RMD CF, right?

gaojieliu (commented):
Looks good overall, left some more minor comments.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/replication/merge/RmdSerDe.java
sixpluszero: Not reused, but use it in this single call. This method is called by getReplicationMetadataAndSchemaId() in which it already is able to get the old rmd manifest. and this particular inner method also need to set rmd record and other info into RmdWithValueSchemaId class, so I passed it into this method.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
sixpluszero: Yes..for now PUT and UPDATE should be handled..We could also do it for DELETE.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/storage/chunking/ChunkedValueManifestContainer.java
sixpluszero: Hmm...why? Here we are always targeting Manifest type right? No matter it is value or RMD

gaojieliu: I was trying to see whether we can create this class as a utility class, which can be used by other places.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
sixpluszero: Yes, this also confused me for a whole day. Today I have the same basic understanding as what you described, but I was not 100% sure though. But I am open to this change, will do.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/replication/merge/RmdSerDe.java
gaojieliu: I see.

Can we update the javadoc to reflect the change? and also remove unused code?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/storage/chunking/ChunkedValueManifestContainer.java
sixpluszero: I think let's leave it as it is for now. We could potentially refactor it if we do see more general use case there...

sixpluszero:
> Regarding the performance concern of the ingestion path, I guess it is fine to keep the current scope, but we should cut tickets to track the follow-up tasks. From my understanding, large messages in chunk-enabled store are rare, so maybe we can create a RocksDB CF to keep the keys with large messages, and in the ingestion path, Ingestion Task can lookup this small CF to check whether a clean up is required or not for full put/delete. Lookup in this small CF should be very fast if some specific partition only contains 10s of large records.
> 
> Please let me know if you want to chat about the comments.

I think I get what you were trying to indicate here. But is it true that small CF has much faster performance? And I am even wondering if only 10s of keys are chunked, we can maybe cache it in memory, just like TransientRecord

gaojieliu (approved):
Thanks!

gaojieliu:
> 

There is no guarantee that the large entry count will be 10s, but I think that is a common pattern, so we can't keep them in memory since the memory usage will explode if the large entry count is high.
For a small CF, that is my guess about performance implication, and we can always try it in a local benchmark. Another thing we can consider is to have a dedicated block cache for this purpose to make sure other lookups won't pollute the cache, and if the entry count is small, I guess all the entries will be cached.

PR #536, opened 2023-07-18
[controller] Cleanup Helix disabled partition map

majisourav99:
## Cleanup Helix disabled partition map during dropping resources

During ingestion error in leader, we add  disabled partition map config in helix. But even if those resources are dropped later, the disabled map not cleared, which later might cause znode to blow up. This PR will cleanup disabled partition map (by calling enable, which will clear the node) during dropping of resources.

## How was this PR tested?

unit test

### DIFF
--- a/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
@@ -4700,7 +4700,19 @@ public class VeniceHelixAdmin implements Admin, StoreCleaner {
   @Override
   public void deleteHelixResource(String clusterName, String kafkaTopic) {
     checkControllerLeadershipFor(clusterName);
-    helixAdminClient.dropResource(clusterName, kafkaTopic);
+    List<String> instances = getStorageNodes(clusterName);
+    int partitionCount = getHelixVeniceClusterResources(clusterName).getRoutingDataRepository()
+        .getPartitionAssignments(kafkaTopic)
+        .getExpectedNumberOfPartitions();
+    for (String instance: instances) {
+      List<String> partitions = new ArrayList<>(partitionCount);
+      for (int partitionId = 0; partitionId < partitionCount; partitionId++) {
+        partitions.add(HelixUtils.getPartitionName(kafkaTopic, partitionId));
+      }
+      // clean up disabled partition map, so that it does not grow indefinitely with dropped resources
+      getHelixAdminClient().enablePartition(true, clusterName, instance, kafkaTopic, partitions);
+    }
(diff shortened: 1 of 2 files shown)


services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -4700,7 +4700,21 @@ public boolean containsHelixResource(String clusterName, String kafkaTopic) {
...
+      for (Map.Entry<String, List<String>> entry: disabledPartitions.entrySet()) {
+        if (entry.getKey().equals(kafkaTopic)) {
+          // clean up disabled partition map, so that it does not grow indefinitely with dropped resources
+          getHelixAdminClient().enablePartition(true, clusterName, instance, kafkaTopic, entry.getValue());
+          LOGGER.info("Cleaning up disabled replica of resource {}, partitions {}", entry.getKey(), entry.getValue());
+        }
+      }
+    }
+
+    getHelixAdminClient().dropResource(clusterName, kafkaTopic);
```
gaojieliu, line 4717: Does the order matter?
Such as dropping resource first before re-enabling these disabled partition?
If we switch the order, will re-enabling disabled partition operation still work?
I am worried that the re-enabling before dropping could trigger unnecessary rebalancing.

anaberezhnov: I find the change and the description confusing. Can you please clarify how enabling the partition helps to clean up (remove?) the map? Also wouldn't it make sense to make it part of HelixAdminClient rather than VeniceHelixAdmin?

majisourav99: The disabled partitions are kept in a config map for a resource on a znode.  If we enable those map, it clears those map. The API call to helix is part of `HelixAdminClient`, but the rest of the logic should stay in VeniceHelixAdmin

gaojieliu:
@anaberezhnov brought up a valid point regarding whether helix resource drop API should take care of the disabled partition znode cleanup or not.
@majisourav99 Can you follow up with the Helix team on this?
We can adopt this temp solution for now and once Helix agrees and delivers such logic, we can get rid of the Venice logic.

gaojieliu (approved):

majisourav99:
> @anaberezhnov brought up a valid point regarding whether helix resource drop API should take care of the disabled partition znode cleanup or not. @majisourav99 Can you follow up with the Helix team on this? We can adopt this temp solution for now and once Helix agrees and delivers such logic, we can get rid of the Venice logic.

yes, let me raise a ticket with Helix team on this.

PR #537, opened 2023-07-18
[controller] Update superset schema ID after all value schema related operations

xunyin8:
## [controller] Update superset schema ID after all value schema related operations

Existing implementation updates the superset schema id as soon as we add a new value or superset schema. This is problematic because in parent controller we perform schema addition in the following order:
1. Add new value schema (superset schema id is also updated).
2. Check if A/A is enabled and update RMD schema.
3. Check if WC is enabled and update derived schema.

This execution order allows race to happen when superset schema id is updated but the corresponding RMD schema and derived schema are not available yet. Resulting in failure in downstream consumers like the Venice storage node during ingestion.

The fix is to perform superset schema id as the last step and separate it into its own update store admin message.

## How was this PR tested?
Existing unit and integration tests.

### DIFF
--- a/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
@@ -4843,7 +4843,7 @@ public class VeniceHelixAdmin implements Admin, StoreCleaner {
   }
 
   /**
-   * @see #addValueSchema(String, String, String, int, DirectionalSchemaCompatibilityType, boolean)
+   * @see #addValueSchema(String, String, String, int, DirectionalSchemaCompatibilityType)
    */
   @Override
   public SchemaEntry addValueSchema(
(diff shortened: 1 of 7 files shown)


services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -4899,15 +4892,6 @@ public SchemaEntry addValueSchema(
...
               + valueSchemaStr);
     }
-
-    if (doUpdateSupersetSchemaID) {
-      LOGGER.info(
-          "For store: {} in cluster: {}, value schema is the same as superset schema. Update superset schema ID to {}.",
-          storeName,
-          clusterName,
-          schemaId);
-      updateSupersetSchemaForStore(storeName, clusterName, schemaId);
```
majisourav99, line 4909: Looks like you can delete this `updateSupersetSchemaForStore` method.

xunyin8: Good catch, will remove it.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
```diff
@@ -2996,6 +2996,10 @@ private SchemaEntry addValueAndSupersetSchemaEntries(
           newSupersetSchemaEntry.getId(),
           newSuperSetWriteComputeSchema.toString());
     }
+    updateStore(
```
sixpluszero, line 2999: I guess this is the main changes. Do you have any existing test to cover the changes here?

sixpluszero: Actually, I am wondering in the incident, is it Server ZK listener failed to catch up schema update? If this is the case, even you tweak the superset schema update order, this will still happen? (As superset schema listener is store repo's listener but schema is another one... 
Or maybe my concern is already covered by some Slack discussion that I am aware of.

xunyin8: I don't have a good test in mind to reproduce the race in an integration test deterministically. Since anything we write might still pass even without the fix. Regarding to the zk watch listener issue you are correct that the internal cache updates are still separate callbacks however that should be covered by the forced schema refresh we currently have in our code (when we encounter a new superset schema ID). The problem that this PR is trying to solve is that in some scenarios the superset schema ID IS UPDATED but the corresponding RMD is not available on zk yet so even with an ad-hoc zk update we still can't find the corresponding schema.

sixpluszero (approved):

PR #539, opened 2023-07-20
[controller] Support adding value schema with id and compat type via child controller client

shuhui-liu:
## Summary

To resolve KME upgrade deployment order, one action is to make child controllers and servers register KME from the code into local system store at startup time. However, KME is only backward compatible. Adding schema that is not fully compatible via controller client is not supported. This commit supports adding value schema with id and compat type via child controller client.

## How was this PR tested?

New unit test
CI


lluwm (approved):
Thanks Shuhui and it looks good to me!

anaberezhnov (dismissed):

shuhui-liu (commented):
Thanks @lluwm and @anaberezhnov for the review. AddValueSchema parameters are changed by another commit. I rebased and resolved conflicts. Need another approval to merge. Please take a look when you have time. Thanks.

lluwm (approved):

PR #541, opened 2023-07-21
[dvc][server] Removed locking from PartitionTracker

FelixGV:
This was flagged when reviewing 6c8b672 and left as a followup.

The historical context is that there was a temporary phase of the A/A migration where RT producers would be replicated to all regions and so servers would consume duplicate events concurrently, even some written by the same producer GUID. This led to race conditions in the DIV code, and so this locking was introduced to guard against it. The locking was retained as part of the aforementioned commit, but is not actually necessary anymore, so we're cleaning it up. This is expected to provide a small boost to ingestion performance as well as the memory efficiency of DIV, especially in cases where there are lots of producer GUIDs involved.

Miscellaneous:

- Tweaked the VeniceSystemFactoryTest::testSerializationCastParams data provider so that it returns parameters with a deterministic toString output, which is necessary for the test-retry plugin to work well.

## How was this PR tested?
CI


gaojieliu (approved):

FelixGV:
Thanks Gaojie!

PR #542, opened 2023-07-21
[vpj] Reuse KafkaConsumer when building dict from an existing topic

gaojieliu:
Previously, `KafkaInputDictTrainer` would initiate a new KafkaConsumer per partition, which results in ton of KafkaConsumer related logs printed in VPJ log, which would make VPJ log analyzing less efficient. This code change will reuse the same consumer when extracting data messages from different partitions to reduce the total amount of logs produced by KafkaInputDictTrainer.

## How was this PR tested?
CI


clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/input/kafka/KafkaInputDictTrainer.java
```diff
@@ -119,6 +125,9 @@ public Param build() {
   private final CompressionStrategy sourceVersionCompressionStrategy;
   private final CompressorBuilder compressorBuilder;
 
+  // For testing purpose
```
FelixGV, line 128: Nitpick... it seems to me that it's the setter for this variable that is for testing purposes, but not the variable itself. Is that right?

FelixGV (dismissed):
LGTM. Just left a minor nitpick/question. I can give another ship it if necessary.

PR #545, opened 2023-07-25
[test] Fix flaky FastClientServerReadQuotaTest

xunyin8:
[test] Fix flaky FastClientServerReadQuotaTest
The previous test behavior would stop making requests as soon as it receives one
VeniceClientRateExceededException while checking the metric of only one server.
This is problematic when the test cluster has multiple servers since the rejected
request might not be handled by the server metric that we picked arbitrarily. 

The fix is to refactor the test to take multiple servers into consideration when
validating their metrics.

## How was this PR tested?
Test only fix and verified by running the test itself multiple times.

### DIFF
--- a/internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/fastclient/FastClientServerReadQuotaTest.java
+++ b/internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/fastclient/FastClientServerReadQuotaTest.java
@@ -54,14 +54,13 @@ public class FastClientServerReadQuotaTest extends AbstractClientEndToEndSetup {
         assertNotNull(serverMetric.getMetric(readQuotaUsageRatio));
       }
     });
-    int averageQuotaRequested = 500 / serverMetrics.size();
+    int quotaRequestedSum = 0;
     for (MetricsRepository serverMetric: serverMetrics) {
-      assertTrue(
-          serverMetric.getMetric(readQuotaRequestedString).value() >= averageQuotaRequested,
-          "Metric value: " + serverMetric.getMetric(readQuotaRequestedString).value());
+      quotaRequestedSum += serverMetric.getMetric(readQuotaRequestedString).value();
       assertEquals(serverMetric.getMetric(readQuotaRejectedString).value(), 0d);
       assertTrue(serverMetric.getMetric(readQuotaUsageRatio).value() > 0);
     }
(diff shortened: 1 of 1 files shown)


FelixGV (commented):
Thanks for working on this flaky test!

I don't understand though... if one request got a quota exception, why isn't it reflected in the metric? I don't understand the part about there being a small window?

xunyin8:
> Thanks for working on this flaky test!
> 
> I don't understand though... if one request got a quota exception, why isn't it reflected in the metric? I don't understand the part about there being a small window?

Based on trial and error in IDE debug mode there are two servers in the test cluster. With only one rejected request (on a arbitrary key and replica) the test always check the metrics on one of the server. If we don't get rejected by the correct server then the test will fail. Still a good point that you raised, let me actually fix it in another way that's deterministic since this way we can technically still get unlucky with 100 requests all going to the wrong replica.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/fastclient/FastClientServerReadQuotaTest.java
```diff
@@ -39,22 +40,30 @@ public void testServerReadQuota() throws Exception {
...
     });
-    assertTrue(
-        serverMetric.getMetric(readQuotaRequestedString).value() >= 250,
-        "Metric value: " + serverMetric.getMetric(readQuotaRequestedString).value());
-    assertEquals(serverMetric.getMetric(readQuotaRejectedString).value(), 0d);
-    assertTrue(serverMetric.getMetric(readQuotaUsageRatio).value() > 0);
+    int averageQuotaRequested = 500 / serverMetrics.size();
+    for (MetricsRepository serverMetric: serverMetrics) {
+      assertTrue(
+          serverMetric.getMetric(readQuotaRequestedString).value() >= averageQuotaRequested,
```
FelixGV, line 60: This assumes the traffic is distributed uniformly, right? Is the assumption too tight? Should we check that the sum of that metric across all servers is >= 500 instead?

xunyin8: It's pretty uniform based on the current routing strategy but sure if it's just for the purpose of verifying quota we can do the sum instead.

FelixGV (approved):
Thanks a lot Xun!

PR #546, opened 2023-07-25
[controller] Always get leader controller from EV of the controller cluster

shuhui-liu:
## Summary

Problem: There is a race condition during HaaS transition, which causes the flakiness of testTransitionToHAASControllerAsStorageClusterLeader. The test case stops 1 old controller that disables HaaS and starts 1 new controller that enables HaaS 3 times. When the 2nd new controller starts, if the get leader controller request is sent to an old controller, since the old controller disables HaaS, it returns storage cluster controller leader property. However, since a controller that enables HaaS is added, the property becomes a Helix controller instead of Venice controller. Helix controller could not handle system schema requests at controller start time. The new controller could not start and the test fails.

Fix: Always get leader controller from EV of the controller cluster because for both Helix as a library and HaaS, the leader in the controller cluster EV is the Venice controller leader that we want.

## How was this PR tested?

Ran CI a few times and testTransitionToHAASControllerAsStorageClusterLeader passed.

### DIFF
--- a/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
@@ -5903,52 +5903,15 @@ public class VeniceHelixAdmin implements Admin, StoreCleaner {
 
   /**
    * @see Admin#getLeaderController(String)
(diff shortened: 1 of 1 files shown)


xunyin8 (approved):
Thanks Shuhui!

PR #547, opened 2023-07-25
[pubsub][kafka] Skip callback processing when the producer is closed forcefully

sushantmane:
## Skip callback processing on forceful producer closure

When the producer is closed forcefully, we skip the callback processing. Nevertheless, we  
ensure that any future returned by the producer to the caller is completed, either normally  
or exceptionally.


sixpluszero:
If callback is triggered, then it means it is successfully produced, why shouldn't we process the record, like in LEADER we drain it to the drainer?

sushantmane:
> If callback is triggered, then it means it is successfully produced

Not necessarily. Callback is also triggered if producer fails to send records due to any reasons, e.g., network issues, producer closed, etc.

>why shouldn't we process the record, like in LEADER we drain it to the drainer?

Sometimes we do close producer (killing ingestion, for example) and don't really care what happens to the records that were added in the batch but weren't sent out to the brokers yet. In such cases, it is ok to skip user provided callback processing (we still complete future as it may be used by caller to make sync/blocking calls).  

When producer is closed forcefully sometimes (when there are unsent records in buffer) it messes up stats and triggers alerts unnecessarily. This fix will eliminate that issue.

sixpluszero (approved):
Thanks for the changes

sushantmane:
Thanks a lot, @sixpluszero for the review!

PR #551, opened 2023-07-26
[controller] Partial update only allows nullable union pair

sixpluszero:
## [controller] Partial update only allows nullable union pair

This PR adds restriction on the UNION field for partial update schema generation.
Previously, any union type can be enabled partial update, except for multiple collection field scenario.
A/A requires null-value pair of union field:
1. [null, regular field] => long RMD timestamp
2. [null, collection field] => complex record RMD timestamp tracking each element in the collection
3. other => not supported, throw exception during schema generation.

In the future, all new hybrid store will by default enable A/A replication, it is good to align the requirement for UNION here so that all partial update stores can transition into A/A replication smoothly.

## How was this PR tested?
new unit test

User won't be able to enable partial update on store schema that contains union field that is not a null-value pair.
This is also the requirement from A/A replication and we are aligning it to make sure they will have a smooth transition into A/A replication.

### DIFF
--- a/internal/venice-client-common/src/main/java/com/linkedin/venice/schema/writecompute/WriteComputeSchemaConverter.java
+++ b/internal/venice-client-common/src/main/java/com/linkedin/venice/schema/writecompute/WriteComputeSchemaConverter.java
@@ -1,5 +1,6 @@
 package com.linkedin.venice.schema.writecompute;
 
+import static com.linkedin.venice.schema.SchemaUtils.isCollectionUnionSchema;
 import static com.linkedin.venice.schema.writecompute.WriteComputeConstants.WRITE_COMPUTE_RECORD_SCHEMA_SUFFIX;
 import static com.linkedin.venice.schema.writecompute.WriteComputeOperation.LIST_OPS;
 import static com.linkedin.venice.schema.writecompute.WriteComputeOperation.MAP_OPS;
@@ -27,7 +28,7 @@ import org.apache.commons.lang.Validate;
 
 
 /**
- * This class converts a Avro {@link Type#RECORD} schema to its write compute schema. Note that the input schema must
+ * This class converts an Avro {@link Type#RECORD} schema to its write compute schema. Note that the input schema must
  * be of type {@link Type#RECORD}.
  *
  * Currently, it supports record partial update, collection merging and deletion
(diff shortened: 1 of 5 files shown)


internal/venice-client-common/src/main/java/com/linkedin/venice/schema/writecompute/WriteComputeSchemaConverter.java
```diff
@@ -318,91 +319,24 @@ private Schema convertMap(Schema mapSchema, String name, String namespace) {
...
-   *     }, {
-   *       "type" : "array",
-   *       "items" : "simpleRecord"
-   *     } ],
-   *     "default" : { }
-   *   } ]
-   * }
+   * Wrap up a union schema with possible partial update operations.
+   * Only two branches union with one being NULL type is allowed. This will make partial update logic compatible with
+   * Active/Active replication schema requirement.
```
adamxchen, line 324: Is it possible to add what's the Active/Active replication schema requirement with a sentence or two? It'd be easier for people later to understand this method.

internal/venice-client-common/src/main/java/com/linkedin/venice/schema/writecompute/WriteComputeSchemaConverter.java
```diff
@@ -318,91 +319,24 @@ private Schema convertMap(Schema mapSchema, String name, String namespace) {
...
    */
   private Schema convertUnion(Schema unionSchema, String name, String namespace) {
```
adamxchen, line 326: Do we have existing tests for `convertUnion` somewhere?

adamxchen (commented):
LGTM. Have some minor comments

internal/venice-client-common/src/main/java/com/linkedin/venice/schema/writecompute/WriteComputeSchemaConverter.java
```diff
@@ -318,91 +319,24 @@ private Schema convertMap(Schema mapSchema, String name, String namespace) {
...
     }
-    SchemaUtils.containsOnlyOneCollection(unionSchema);
-    return SchemaUtils.createFlattenedUnionSchema(unionSchema.getTypes().stream().sequential().map(schema -> {
-      Schema.Type type = schema.getType();
-      if (type == RECORD) {
-        return schema;
-      }
-      return convert(schema, name, namespace);
-    }).collect(Collectors.toList()));
+    if (isCollectionUnionSchema(unionSchema)) {
```
gaojieliu, line 330: I have a question about this requirement.
For union with multiple types (1. not nullable union, 2. contains more than 2 types), for write compute, I assume we can still allow the full field update, right? And I assume for RMD schema generation, we can generate a field-level timestamp for such union field, right?

sixpluszero: What you said might make sense, let me think about it.

sixpluszero:
Yeah, technically yes. 
I don't have a context about why these are not supported in A/A though, I suspect there might be some earlier discussion for A/A implementation that leads to the decision to block it (maybe for simplicity in implementation?)
So in theory, there are all these types:
    SCHEMA TYPE        CURRENT RMD TYPE
1. primitive - long
2. collection - [long, complex record]
3. [primitive] - long (wrong, should be [long, complex record])
4. [collection] - long (wrong, should be [long, complex record])
5. [null, primitive] - long
6. [null, collection] - [long, complex record]
7. [primitive1, primitive2] - long
8. [primitive, collection] - Not supported
9. [collection1, collection2] - Not supported

Union with 3 branches must have same judgement call as type 5-9.
For today, all of them are supported by WC converter, except 9. 

Are you suggesting we should support all these, except 9 which leads to ambiguious RMD data for value?
My current understanding is, technically we can support that, but I feel like in real use case people usually have 2 branches union, and for all Fedex-backed schema, it is always [null, something]...

internal/venice-client-common/src/main/java/com/linkedin/venice/schema/writecompute/WriteComputeSchemaConverter.java
sixpluszero: Yeah, technically yes. 
I don't have a context about why these are not supported in A/A though, I suspect there might be some earlier discussion for A/A implementation that leads to the decision to block it (maybe for simplicity in implementation?)
So in theory, there are all these types:
    SCHEMA TYPE        CURRENT RMD TYPE
1. primitive - long
2. collection - [long, complex record]
3. [primitive] - long (wrong, should be [long, complex record])
4. [collection] - long (wrong, should be [long, complex record])
5. [null, primitive] - long
6. [null, collection] - [long, complex record]
7. [primitive1, primitive2] - long
8. [primitive, collection] - Not supported
9. [collection1, collection2] - Not supported

Union with 3 branches must have same judgement call as type 5-9.
For today, all of them are supported by WC converter, except 9. 

Are you suggesting we should support all these, except 9 which leads to ambiguious RMD data for value?
My current understanding is, technically we can support that, but I feel like in real use case people usually have 2 branches union, and for all Fedex-backed schema, it is always [null, something]...

majisourav99: Yeah, most used usecase for union is nullable union. If someone designs [collection1,collection2] as union, they should  redesign the schema and rejecting such schemas for WC should be fine.

sixpluszero:
Let me close this for now. For what types of WC / AA collection schema to be support in Venice, I might need more time to think about it. Since this PR is stale and it is not top prioritized issue, let's close it and create one in the future.

PR #552, opened 2023-07-27
[server][dvc][tc] Avro fixes, optimizations and 1.10.2 upgrade

FelixGV:
The upgrade of Avro from 1.9 to 1.10 carries one major change which is that GenericRecords no longer return null for fields they do not have, but rather throw AvroRuntimeException. This makes the read compute code fail.

This commit also optimizes various usages of Avro in the read compute hot path, including:

- The ComputeRequestWrapper now carries a list of operation result fields, which is looked up once at the start of the query, rather than redundantly during each processed key. This reduces map lookups from 1/op/key to 1/op/query.

- The getNullableFieldValueAsList function removes 1/op/key map lookups.

- The validateNullableFieldAndGetErrorMsg function removes 1 to 2/op/key map lookups.

- The ComputeRouterRequestWrapper now uses a static final deserializer rather than looking it up on each query.

And other non-Avro optimizations:

- The stats for each read compute op are now incremented once at the end of the query by the amount of keys that found a hit in the storage engine, rather than piecemeal, while processing each key.

- The validateNullableFieldAndGetErrorMsg does not return an Optional anymore.

- ComputeOperationType::valueOf now uses EnumUtils to remove a map lookup. This removes 2/op/key map lookups.

## How was this PR tested?
CI

The user does not have to bump Avro, and the client libraries will continue to work with older Avro versions. However, if the user does adopt Avro 1.10 in its client app, then the change of semantics mentioned at the beginning will apply to the GenericRecords returned by the client libs.

### DIFF
--- a/internal/venice-avro-compatibility-test/src/test/java/com/linkedin/venice/utils/ClassPathSupplierForVeniceCluster.java
+++ b/internal/venice-avro-compatibility-test/src/test/java/com/linkedin/venice/utils/ClassPathSupplierForVeniceCluster.java
@@ -23,7 +23,7 @@ import org.apache.logging.log4j.Logger;
 public class ClassPathSupplierForVeniceCluster implements Supplier<String> {
   private static final Logger LOGGER = LogManager.getLogger(ClassPathSupplierForVeniceCluster.class);
 
-  private static final String AVRO_192_JAR_FILE = "avro-1.9.2.jar";
+  private static final String AVRO_JAR_FILE = "avro-1.10.2.jar";
 
   @Override
   public String get() {
@@ -67,7 +67,7 @@ public class ClassPathSupplierForVeniceCluster implements Supplier<String> {
 
   private File extractAvro192JarFileBasedOnExistingAvroJarFile(File existingAvroJarFile) {
     LOGGER.info("Existing avro jar file: {}", existingAvroJarFile.getAbsolutePath());
-    if (existingAvroJarFile.getName().equals(AVRO_192_JAR_FILE)) {
+    if (existingAvroJarFile.getName().equals(AVRO_JAR_FILE)) {
       return existingAvroJarFile;
     }
     /**
@@ -79,11 +79,11 @@ public class ClassPathSupplierForVeniceCluster implements Supplier<String> {
     File avroRootDir = existingAvroJarFile.getParentFile().getParentFile().getParentFile();
     Collection<File> jarFiles = FileUtils.listFiles(avroRootDir, new String[] { "jar" }, true);
     for (File jarFile: jarFiles) {
-      if (jarFile.getName().equals(AVRO_192_JAR_FILE)) {
-        LOGGER.info("Found the jar file: {} for {}", jarFile.getAbsolutePath(), AVRO_192_JAR_FILE);
+      if (jarFile.getName().equals(AVRO_JAR_FILE)) {
+        LOGGER.info("Found the jar file: {} for {}", jarFile.getAbsolutePath(), AVRO_JAR_FILE);
         return jarFile;
       }
     }
-    throw new VeniceException("Failed to find out " + AVRO_192_JAR_FILE + " in the existing class path");
+    throw new VeniceException("Failed to find out " + AVRO_JAR_FILE + " in the existing class path");
   }
 }
--- a/internal/venice-avro-compatibility-test/src/test/java/com/linkedin/venice/utils/ClassPathSupplierForVeniceCluster.java
+++ b/internal/venice-avro-compatibility-test/src/test/java/com/linkedin/venice/utils/ClassPathSupplierForVeniceCluster.java
@@ -17,8 +17,8 @@ import org.apache.logging.log4j.Logger;
 
 
 /**
- * This utility is used to remove the existing Avro jars and prepend avro-1.9 to make sure the Venice Cluster
- * is always using avro-1.9.
+ * This utility is used to remove the existing Avro jars and prepend a specific Avro jar version to the classpath,
+ * to control which version the Venice Cluster will use. The version used is controlled by {@link #AVRO_JAR_FILE}.
  */
 public class ClassPathSupplierForVeniceCluster implements Supplier<String> {
   private static final Logger LOGGER = LogManager.getLogger(ClassPathSupplierForVeniceCluster.class);
@@ -54,10 +54,7 @@ public class ClassPathSupplierForVeniceCluster implements Supplier<String> {
         if (existingAvroJarFile == null) {
           throw new VeniceException("There should be some existing Avro lib in the class path");
         }
-        /**
-         * Append avro-1.9 jar to the classpath, which is the one being used by the backend.
-         */
-        paths.add(extractAvro192JarFileBasedOnExistingAvroJarFile(existingAvroJarFile));
+        paths.add(extractAvroJarFileBasedOnExistingAvroJarFile(existingAvroJarFile));
       } catch (Exception e) {
         throw new VeniceException("Failed to compose class path", e);
       }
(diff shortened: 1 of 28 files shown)


majisourav99 (commented):
LGTM. Thanks a lot for the changes.  I was thinking if  `record.get(i)` could ever throw `ArrayIndexOutOfBoundsException` if field.pos() is returns such indices ?

internal/venice-avro-compatibility-test/src/test/java/com/linkedin/venice/utils/ClassPathSupplierForVeniceCluster.java
```diff
@@ -67,7 +67,7 @@ public String get() {
 
   private File extractAvro192JarFileBasedOnExistingAvroJarFile(File existingAvroJarFile) {
```
gaojieliu, line 68: Rename this function to drop `192`?

internal/venice-client-common/src/main/java/com/linkedin/venice/compute/ComputeRequestWrapper.java
```diff
@@ -153,4 +156,20 @@ public void setOperations(List<ComputeOperation> operations) {
         throw new VeniceException("Compute request version " + version + " is not support yet.");
     }
   }
+
+  public void initializeOperationResultFields(Schema resultSchema) {
+    List<ComputeOperation> operations = getOperations();
+    this.operationResultFields = new ArrayList<>(operations.size());
+    ComputeOperation computeOperation;
+    ReadComputeOperator operator;
+    for (int i = 0; i < operations.size(); i++) {
```
gaojieliu, line 165: Shall we use an iterator here in case the `operations` is using a LinkedList-like structure?

internal/venice-client-common/src/main/java/com/linkedin/venice/compute/ComputeRequestWrapper.java
```diff
@@ -153,4 +156,20 @@ public void setOperations(List<ComputeOperation> operations) {
...
+      computeOperation = operations.get(i);
+      operator = ComputeOperationType.valueOf(computeOperation).getOperator();
+      this.operationResultFields.add(resultSchema.getField(operator.getResultFieldName(computeOperation)));
+    }
+  }
+
+  public List<Schema.Field> getOperationResultFields() {
```
gaojieliu, line 172: Shall we combine `getOperationResultFields` and `initializeOperationResultFields`?
For example, we can keep a reference of `resultSchema` in `ComputeRequestWrapper` and when calling function: `getOperationResultFields`, it can check whether the cached `operationResultFields` is null or not, and if it is `null`, do the initialization.

With two functions, we need to make sure the caller will call `initializeOperationResultFields` before `getOperationResultFields`.

internal/venice-client-common/src/main/java/com/linkedin/venice/compute/ComputeUtils.java
```diff
@@ -245,48 +244,47 @@ public static float squaredL2Norm(List<Float> list) {
...
     if (!(value instanceof List)) {
       throw new IllegalArgumentException(
-          String.format("Field %s in the record is not of the type list. Value: %s", fieldName, record));
+          String.format("Field %s in the record is not of the type list. Value: %s", field.name(), record));
     }
     return (List<T>) value;
   }
 
   /**
    * @return Error message if the nullable field validation failed or Optional.empty() otherwise.
```
gaojieliu, line 266: Update javadoc to reflect the change?

internal/venice-client-common/src/main/java/com/linkedin/venice/compute/ComputeUtils.java
```diff
@@ -315,17 +321,32 @@ public static GenericRecord computeResult(
...
-    for (ComputeOperation computeOperation: operations) {
-      ReadComputeOperator operator = ComputeOperationType.valueOf(computeOperation).getOperator();
-      String errorMessage =
-          validateNullableFieldAndGetErrorMsg(operator, inputRecord, operator.getOperatorFieldName(computeOperation))
-              .orElse(null);
+    ComputeOperation computeOperation;
+    ReadComputeOperator operator;
+    String operatorFieldName, errorMessage;
+    Schema.Field operatorField, resultField;
+    for (int i = 0; i < operations.size(); i++) {
```
gaojieliu, line 328: What is the reason to change it to a counter-based for loop?

gaojieliu (commented):
Left some minor comments.

internal/venice-client-common/src/main/java/com/linkedin/venice/compute/ComputeUtils.java
majisourav99: I believe it was done for performance reason. But not sure how much extra these micro-optimization would yield.
https://stackoverflow.com/questions/256859/is-there-a-performance-difference-between-a-for-loop-and-a-for-each-loop

FelixGV: I'm accessing two lists so I need the `i`.

FelixGV:
> LGTM. Thanks a lot for the changes. I was thinking if `record.get(i)` could ever throw `ArrayIndexOutOfBoundsException` if field.pos() is returns such indices ?

It can if we get the position from the fields of the wrong schema. I believe the schema/records are all well-matched here, but I invite you to double-check.

internal/venice-client-common/src/main/java/com/linkedin/venice/compute/ComputeUtils.java
gaojieliu: So using two list iterator would work?

internal/venice-client-common/src/main/java/com/linkedin/venice/compute/ComputeRequestWrapper.java
FelixGV: Hmm... should we worry about that? My understanding is that LL almost always underperforms ArrayList...

internal/venice-client-common/src/main/java/com/linkedin/venice/compute/ComputeRequestWrapper.java
FelixGV: If I always had the result schema at construction-time, then I could pre-init at that time, but the code using this class does not make it very easy, IIRC. I'll look into it again and see if I can improve this.

internal/venice-client-common/src/main/java/com/linkedin/venice/compute/ComputeUtils.java
FelixGV: Yes it would, but does it matter...? Is two iterators that much cleaner? I'm wary of the garbage cost of allocating iterators, since this is one of our hottest paths.

gaojieliu: The iterator comment is mainly for defensive coding, not a big concern.
I was considering if someone changes the implementation of `operations` to use LinkedList, the for-loop will perform worse than an iterator.
In practice, it doesn't matter with the current logic.

internal/venice-client-common/src/main/java/com/linkedin/venice/compute/ComputeRequestWrapper.java
gaojieliu: Correct, the iterator is mainly for defensive purpose in case someone changes the implementation to use a LinkedList.

internal/venice-client-common/src/main/java/com/linkedin/venice/compute/ComputeRequestWrapper.java
FelixGV: FYI... I've been working on this this morning, and I'm finding a LOT of cruft. There is definitely room to simplify many things and make the code less error-prone, but it's going to be a big refactoring. Not sure that it should be clubbed into this change. I can keep going for now, but LMK if you prefer that we just go ahead with the smaller change, since it seems kind of critical. I could always do the bigger refactoring afterwards...

gaojieliu (approved):

PR #553, opened 2023-07-27
[controller][server] Pass all properties from server and controller to pub sub adatpers to include special properties we need.

haoxu07:

Previous server and controller prepare specific properties map with only kafka related properties from server config and controller config, respectively. To support other type of pub sub system, we may need to pass additionally properties to pub sub adapters. Instead of creating new empty properties, we just use copied properties of server configs and controller configs and the overriding logic is still the same. 

Besides, fix some additional configs merge logic issues inside test utils. 

## How was this PR tested?

internal CI

### DIFF
--- a/clients/da-vinci-client/src/test/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTaskTest.java
+++ b/clients/da-vinci-client/src/test/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTaskTest.java
@@ -2982,7 +2982,7 @@ public abstract class StoreIngestionTaskTest {
     doReturn("localhost").when(mockKafkaConsumerProperties).getProperty(eq(KAFKA_BOOTSTRAP_SERVERS));
 
     VeniceServerConfig mockVeniceServerConfig = mock(VeniceServerConfig.class);
-    doReturn(new VeniceProperties(new Properties())).when(mockVeniceServerConfig).getClusterProperties();
+    doReturn(VeniceProperties.empty()).when(mockVeniceServerConfig).getClusterProperties();
     VeniceProperties mockVeniceProperties = mock(VeniceProperties.class);
     doReturn(true).when(mockVeniceProperties).isEmpty();
     doReturn(mockVeniceProperties).when(mockVeniceServerConfig).getKafkaConsumerConfigsForLocalConsumption();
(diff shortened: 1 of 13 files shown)


clients/da-vinci-client/src/test/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTaskTest.java
```diff
@@ -3084,6 +3085,7 @@ public void testLeaderShouldSubscribeToCorrectVTOffset() {
         .getLocalStorageEngine(anyString());
     doReturn(mockStorageEngineRepository).when(builder).getStorageEngineRepository();
     VeniceServerConfig veniceServerConfig = mock(VeniceServerConfig.class);
+    doReturn(new VeniceProperties(new Properties())).when(veniceServerConfig).getClusterProperties();
```
nisargthakkar, line 3088: Instead of `new VeniceProperties(new Properties())`, let's use `VeniceProperties.empty()`

nisargthakkar, line 3088: Instead of `new VeniceProperties(new Properties())`, let's use `VeniceProperties.empty()`

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/adapter/kafka/admin/ApacheKafkaAdminConfig.java
```diff
@@ -20,7 +20,7 @@ public class ApacheKafkaAdminConfig {
   private final String brokerAddress;
 
   public ApacheKafkaAdminConfig(VeniceProperties veniceProperties) {
-    this.brokerAddress = ApacheKafkaProducerConfig.getPubsubBrokerAddress(veniceProperties);
+    this.brokerAddress = veniceProperties.getString(ApacheKafkaProducerConfig.KAFKA_BOOTSTRAP_SERVERS);
```
nisargthakkar, line 23: Why do we need this change? The function also checks for Kafka using SSL too

haoxu07: This config creation will be triggered by server and controller. For server and controller, before we pass the properties we will have some logic override `ApacheKafkaProducerConfig.KAFKA_BOOTSTRAP_SERVERS` field. And this logic is a bit different between server and controller. And the original goal of these logics for adpaters is: only respect this `ApacheKafkaProducerConfig.KAFKA_BOOTSTRAP_SERVERS` field. With this change, we pass all the configs instead of creating a new one, so `ApacheKafkaProducerConfig.getPubsubBrokerAddress` does not fit this, then it will cause a lot troubles.

clients/da-vinci-client/src/test/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTaskTest.java
haoxu07: Good catch! I will replace all occurances.

nisargthakkar (approved):
LGTM

PR #554, opened 2023-07-27
[compat][controller][admintool] Added store-level min compaction lag support

gaojieliu:
## Summary
Today, the default log compaction lag for the version topics of hybrid stores is 24h, and for high write throughput stores, such as several hundreds of thousands of write QPS, the long compaction delay will accumulate a huge number of messages in VT, which will drag down the rebalance and fresh bootstrapping in DaVinci instances. We should evaluate the high write throughput stores case by case to make a decision about what compaction delay it should use to fulfill the rebalance and fresh bootstrapping requirement.
Also the compaction delay policy change is not free since it will increase the cpu usage of Kafka broker.

In this code change, I also deprecated `StoreMetaValueWriteOpRecord` by using `UpdateBuilder` to make it less error prone.

## How was this PR tested?

CI


nisargthakkar (approved):
LGTM. Thanks Gaojie

PR #555, opened 2023-07-28
[router] Skip registering router metrics for system stores

majisourav99:
## Do not register router metrics for system stores

The system stores are mostly read from controller and their router metrics are hardly useful for debugging system store issues as most of the time system store issues stems from write path to them. But nonetheless for each user store router emits 2x the number of system store metrics, one for its DVC push status and one for meta store system stores. This PR will stop emitting system store router metrics thus cutting down router metric count by 1/3.

## How was this PR tested?

GHCI

### DIFF
--- a/services/venice-router/src/main/java/com/linkedin/venice/router/stats/RouterHttpRequestStats.java
+++ b/services/venice-router/src/main/java/com/linkedin/venice/router/stats/RouterHttpRequestStats.java
@@ -67,8 +67,6 @@ public class RouterHttpRequestStats extends AbstractVeniceHttpStats {
   private final Sensor multiGetFallbackSensor;
   private final Sensor metaStoreShadowReadSensor;
   private Sensor keySizeSensor;
-  private final boolean isSystemStore;
-
   private Sensor systemStoreSensor = null;
 
   // QPS metrics
@@ -79,8 +77,7 @@ public class RouterHttpRequestStats extends AbstractVeniceHttpStats {
       ScatterGatherStats scatterGatherStats,
       boolean isKeyValueProfilingEnabled) {
     super(metricsRepository, storeName, requestType);
-    this.isSystemStore = VeniceSystemStoreUtils.isSystemStore(storeName);
-    if (isSystemStore) {
+    if (VeniceSystemStoreUtils.isSystemStore(storeName)) {
       MetricsRepository dummyMetricRepo = new MetricsRepository();
       this.systemStoreSensor = dummyMetricRepo.sensor("dummy_system_store_sensor");
     }
@@ -374,7 +371,7 @@ public class RouterHttpRequestStats extends AbstractVeniceHttpStats {
 
   @Override
   protected Sensor registerSensor(String sensorName, MeasurableStat... stats) {
-    if (isSystemStore) {
+    if (systemStoreSensor != null) {
       return systemStoreSensor;
     }
     return super.registerSensor(getFullMetricName(sensorName), null, stats);
--- a/services/venice-router/src/main/java/com/linkedin/venice/router/stats/RouterHttpRequestStats.java
+++ b/services/venice-router/src/main/java/com/linkedin/venice/router/stats/RouterHttpRequestStats.java
@@ -3,6 +3,7 @@ package com.linkedin.venice.router.stats;
 import static com.linkedin.venice.stats.AbstractVeniceAggStats.*;
 
 import com.linkedin.alpini.router.monitoring.ScatterGatherStats;
+import com.linkedin.venice.common.VeniceSystemStoreUtils;
 import com.linkedin.venice.read.RequestType;
 import com.linkedin.venice.stats.AbstractVeniceHttpStats;
 import com.linkedin.venice.stats.LambdaStat;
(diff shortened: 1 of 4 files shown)


FelixGV (commented):
Hi Sourav,

I feel like this change is a bit heavy, and I wonder if we could do it a bit differently. It would be nice if we could maintain immutability of the state, and also not need all these boilerplate null checks in all the record functions. How about the following idea:

1. We add a bit of static state which is just a `Sensor` instance coming out of a throwaway `MetricsRepository` (this is important, since using the main `MR` results in the metrics getting published, which is what this PR wants to avoid...).
2. At the start of the constructor, we populate `this.isSystemStore` as you're doing here.
3. We override the `registerSensor` function along the lines of:

```
  @Override
  protected Sensor registerSensor(String sensorName, MeasurableStat... stats) {
    if (this.isSystemStore) {
      return // the statically defined Sensor mentioned in 1.
    }
    return super.registerSensor(sensorName, stats);
  }
```

In this way, I think the constructor need not have the big if, the state can all remain final, and the record functions would be unchanged. It might also help the `diffCoverage` check ;) ...

majisourav99:
> Hi Sourav,
> 
> I feel like this change is a bit heavy, and I wonder if we could do it a bit differently. It would be nice if we could maintain immutability of the state, and also not need all these boilerplate null checks in all the record functions. How about the following idea:
> 
> 1. We add a bit of static state which is just a `Sensor` instance coming out of a throwaway `MetricsRepository` (this is important, since using the main `MR` results in the metrics getting published, which is what this PR wants to avoid...).
> 2. At the start of the constructor, we populate `this.isSystemStore` as you're doing here.
> 3. We override the `registerSensor` function along the lines of:
> 
> ```
>   @Override
>   protected Sensor registerSensor(String sensorName, MeasurableStat... stats) {
>     if (this.isSystemStore) {
>       return // the statically defined Sensor mentioned in 1.
>     }
>     return super.registerSensor(sensorName, stats);
>   }
> ```
> 
> In this way, I think the constructor need not have the big if, the state can all remain final, and the record functions would be unchanged. It might also help the `diffCoverage` check ;) ...

Thanks for the suggestion, I thought of something similar, but balked at the idea of creating a dummy sensor. But using a dummy MR to register the sensor to the metrics are not published is a good idea, I will update the PR.

FelixGV:
> I thought of something similar, but balked at the idea of creating a dummy sensor.

Another variation would be to register not a dummy sensor, but to have the overridden function exclude the user store name, and rather use just the system store name, when it encounters a system store. This way, all the individual system stores would funnel their metrics into one consolidated metric per system store type. WDYT of this?

majisourav99:
> > I thought of something similar, but balked at the idea of creating a dummy sensor.
> 
> Another variation would be to register not a dummy sensor, but to have the overridden function exclude the user store name, and rather use just the system store name, when it encounters a system store. This way, all the individual system stores would funnel their metrics into one consolidated metric per system store type. WDYT of this?

Having consolidated metrics make sense for some time type of sensors like count/number, but for various latency metric will it be useful ? Even if see high or aberrant value of some consolidated metric, can we debug them ?

FelixGV:
> Having consolidated metrics make sense for some time type of sensors like count/number, but for various latency metric will it be useful ? Even if see high or aberrant value of some consolidated metric, can we debug them ?

I guess that would depend on a case by case basis. I am not sure how often it would be useful or not.

majisourav99:
> > Having consolidated metrics make sense for some time type of sensors like count/number, but for various latency metric will it be useful ? Even if see high or aberrant value of some consolidated metric, can we debug them ?
> 
> I guess that would depend on a case by case basis. I am not sure how often it would be useful or not.

Right, I would rather skip it altogether. I am ok with keeping per system store type metrics as well, just not sure it will be useful or not. Let me know whats your preference

services/venice-router/src/main/java/com/linkedin/venice/router/stats/RouterHttpRequestStats.java
```diff
@@ -65,6 +67,7 @@ public class RouterHttpRequestStats extends AbstractVeniceHttpStats {
   private final Sensor multiGetFallbackSensor;
   private final Sensor metaStoreShadowReadSensor;
   private Sensor keySizeSensor;
+  private Sensor systemStoreSensor = null;
```
FelixGV, line 70: The approach of the singleton sensor only works if it's going to be a dummy sensor, but in order to consolidate all system stores of a given type together, we can't do it with a singleton sensor. Each metric still needs a different sensor, even though it's going to be the same sensor as for other system stores of the same type.

What I would suggest is this:

1. Instead of a singleton sensor, let's have a `private final String systemStoreName` here.
2. In the constructor, let's do something like: `this.systemStoreName = VeniceSystemStoreUtils.isSystemStore(storeName) ? VeniceSystemStoreUtils.extractSystemStoreType(storeName) : null;`
3. Then in `registerSensor`, let's do something like `return super.registerSensor(getFullMetricName(this.systemStoreName == null ? sensorName : this.systemStoreName), null, stats);`

In this way, each metric will have its own sensor, but it will always be the same sensor for a given system store type.

FelixGV (commented):
Thanks Sourav. I would prefer the approach of consolidating the metric per system store type. I think there's a bug in the latest commit though, and I am proposing a fix to it. Please take a look.

services/venice-router/src/main/java/com/linkedin/venice/router/stats/RouterHttpRequestStats.java
majisourav99: Thanks for the review., there another bug, it was using `super.registerSensor(getFullMetricName(this.systemStoreName == null ? ....), null, stats)` as `super.registerSensor` will again call `getFullMetricName` on the already full metric name.

FelixGV (approved):
Great, thanks Sourav!

PR #556, opened 2023-07-28
[admin-tool][vpj] Log Chunk metadata for chunked records in Admin tool topic dumper

nisargthakkar:
## Log Chunk metadata for chunked records in Admin tool topic dumper

This commit adds chunk metadata in the output of the `KafkaTopicDumper` in admin tool 

## How was this PR tested?

Tested manually. GH CI


clients/venice-admin-tool/src/main/java/com/linkedin/venice/KafkaTopicDumper.java
```diff
@@ -42,35 +52,42 @@
...
+  private final boolean isChunkingEnabled;
   private final String parentDirectory;
   private final PubSubConsumerAdapter consumer;
   private final long messageCount;
   private final long endOffset;
   private final int maxConsumeAttempts;
   private final boolean logMetadataOnly;
 
+  private final ChunkKeyValueTransformer chunkKeyValueTransformer;
+  private final SpecificDatumReader<ChunkedKeySuffix> specificDatumReader;
```
gaojieliu, line 87: This one can be a local variable instead of a class-level and also the naming is confusing when sitting in class level.

clients/venice-admin-tool/src/main/java/com/linkedin/venice/KafkaTopicDumper.java
```diff
@@ -309,6 +345,79 @@ private void writeToFile(PubSubMessage<KafkaKey, KafkaMessageEnvelope, Long> rec
...
+    final ByteBuffer chunkedKeySuffixBytes = rawKeyBytesAndChunkedKeySuffix.getChunkedKeySuffixBytes();
+    final ChunkedKeySuffix chunkedKeySuffix = chunkedKeySuffixDeserializer.deserialize(chunkedKeySuffixBytes);
+
+    switch (keyType) {
+      case WITH_VALUE_CHUNK:
+        return chunkedKeySuffix.chunkId;
+      case WITH_CHUNK_MANIFEST:
+        ChunkedValueManifest chunkedValueManifest = manifestSerializer.deserialize(
+            ByteUtils.extractByteArray(((Put) kafkaMessageEnvelope.payloadUnion).putValue),
+            AvroProtocolDefinition.CHUNKED_VALUE_MANIFEST.getCurrentProtocolVersion());
```
gaojieliu, line 404: I know we haven't evolve chunk value manifest yet, but it seems to be safer to use the schema id embedded in KME rather than the current version for decoding, right?

gaojieliu (commented):
Left some minor comments.

clients/venice-admin-tool/src/main/java/com/linkedin/venice/KafkaTopicDumper.java
nisargthakkar: Yeah, I agree

PR #558, opened 2023-07-31
[tc][dvc][router][server] Cleaned up some read compute tech debt

FelixGV:
Prior to this commit, the ComputeRequestWrapper class was used across client, router and server, but these components all had different needs and the end result was a bit awkward, as it led to the class needing to be constructed in a way that was not fully initialized, only to have the rest of the init happen later by mutating its internal state. This seemed error-prone.

Furthermore, the class attempted to serve as a way to evolve the read compute wire protocol, and although we did evolve read compute, the way we evolved it was already compatible anyway, and if we had evolved it in less compatible ways, it's not clear that the original structure of this class would have been good enough to support such evolution.

In this commit, we are cleaning all of that up and simplifying many aspects of the code in the process. Notably:

- ComputeRequestWrapper is now only used in the client, and it is fully initialized at construction-time, with all internal state final.

- The read compute protocol version passed in a header is still checked when parsing an incoming request, but then it's not carried into deeper code paths, where it was essentially ignored and useless.

- Created a SchemaAndToString container class to replace some usages of Pair<Schema, String>.

- Added a ComputeRequest specific record, which is the same as ComputeRequestV[1-4] except that the items inside the list of operations are strongly-typed, rather than objects. We do still need to keep (at least one of) the other schemas around since those are the ones the clients use on the wire, but we can cheaply convert between those and the new one, and so the server code need not do as much casting as before.

- In the router, the VeniceComputePath does not hang on to a ComputeRequestWrapper anymore, which it didn't need to. It merely skips over the bytes represented by the compute request, with as little allocation as possible. It also hangs on to the raw String headers it needs rather than convert them to int and then back to strings moments later. Made all state final.

### DIFF
--- a/internal/venice-client-common/src/main/java/com/linkedin/venice/compute/ComputeUtils.java
+++ b/internal/venice-client-common/src/main/java/com/linkedin/venice/compute/ComputeUtils.java
@@ -1,17 +1,23 @@
 package com.linkedin.venice.compute;
 
+import static com.linkedin.venice.serializer.FastSerializerDeserializerFactory.*;
+
 import com.linkedin.avro.api.PrimitiveFloatList;
 import com.linkedin.venice.VeniceConstants;
 import com.linkedin.venice.compute.protocol.request.ComputeOperation;
+import com.linkedin.venice.compute.protocol.request.ComputeRequest;
+import com.linkedin.venice.compute.protocol.request.ComputeRequestV3;
 import com.linkedin.venice.compute.protocol.request.CosineSimilarity;
 import com.linkedin.venice.compute.protocol.request.Count;
 import com.linkedin.venice.compute.protocol.request.DotProduct;
 import com.linkedin.venice.compute.protocol.request.HadamardProduct;
 import com.linkedin.venice.compute.protocol.request.enums.ComputeOperationType;
 import com.linkedin.venice.exceptions.VeniceException;
+import com.linkedin.venice.serializer.RecordDeserializer;
 import com.linkedin.venice.utils.CollectionUtils;
 import com.linkedin.venice.utils.Pair;
 import com.linkedin.venice.utils.RedundantExceptionFilter;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -24,6 +30,7 @@ import java.util.stream.Collectors;
 import org.apache.avro.Schema;
 import org.apache.avro.generic.GenericData;
 import org.apache.avro.generic.GenericRecord;
+import org.apache.avro.io.BinaryDecoder;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
@@ -41,11 +48,22 @@ public class ComputeUtils {
   private static final RedundantExceptionFilter REDUNDANT_EXCEPTION_FILTER =
       RedundantExceptionFilter.getRedundantExceptionFilter();
 
-  public static void checkResultSchema(
-      Schema resultSchema,
-      Schema valueSchema,
-      int version,
-      List<ComputeOperation> operations) {
+  /**
+   * N.B.: This deserializer performs an evolution from the schema of {@link ComputeRequestV3} to that of
+   * {@link ComputeRequest}, with the only difference between the two being that the items of the operations list
+   * in the former are defined as a union of one type, which unfortunately results in the SpecificRecord typing this
+   * as a {@link List<Object>}. This is a design shortcoming, but which we cannot easily fix, since there are already
+   * clients using this protocol. On the server-side, however, we wish to use proper types without lots of casting,
+   * which we can achieve by letting Avro do the evolution.
+   */
+  private static final RecordDeserializer<ComputeRequest> DESERIALIZER =
+      getFastAvroSpecificDeserializer(ComputeRequestV3.SCHEMA$, ComputeRequest.class);
+
+  public static ComputeRequest deserializeComputeRequest(BinaryDecoder decoder, ComputeRequest reuse) {
+    return DESERIALIZER.deserialize(reuse, decoder);
+  }
+
+  public static void checkResultSchema(Schema resultSchema, Schema valueSchema, List<ComputeOperation> operations) {
     if (resultSchema.getType() != Schema.Type.RECORD || valueSchema.getType() != Schema.Type.RECORD) {
       throw new VeniceException("Compute result schema and value schema must be RECORD type");
     }
@@ -173,6 +191,18 @@ public class ComputeUtils {
     }
   }
 
+  public static List<Schema.Field> getOperationResultFields(List<ComputeOperation> operations, Schema resultSchema) {
+    List<Schema.Field> operationResultFields = new ArrayList<>(operations.size());
+    ComputeOperation computeOperation;
+    ReadComputeOperator operator;
+    for (int i = 0; i < operations.size(); i++) {
+      computeOperation = operations.get(i);
+      operator = ComputeOperationType.valueOf(computeOperation).getOperator();
+      operationResultFields.add(resultSchema.getField(operator.getResultFieldName(computeOperation)));
+    }
+    return operationResultFields;
+  }
+
   private interface FloatSupplierByIndex {
     float get(int index);
   }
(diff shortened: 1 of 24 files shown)


gaojieliu:
Hi @FelixGV 
I haven't reviewed your PR yet.
Based on the PR description, I would like to share my understanding of the current structure.
Read compute is versioned for read compute logic evolution besides schema evolution.
For example, the original version of read compute doesn't support hadamard product, and we added such support in a following read compute version.
The way we evolved in the past:
1. Added a new read compute version to include new read compute operators.
2. Added the new version support in the backend.
3. Rolling out the backend support first.
4. Rolling out the new client version.

In case we rollback backend because of the bug/regression, the backend can fail fast because of unsupported read compute version.
These are the historical discussions I remembered.
In the future, the situation can change considering we are planning to add client read compute support across all the clients.
If we want to add new read compute operators, we can always add the client side support regardless of whether the backend supports it or not. For example, if the backend doesn't support the new operator, the client compute can be the fallback.

Is this idea similar to the one in your mind with this PR?

FelixGV:
> Hi @FelixGV
> I haven't reviewed your PR yet.
> Based on the PR description, I would like to share my understanding of the current structure.
> Read compute is versioned for read compute logic evolution besides schema evolution.
> For example, the original version of read compute doesn't support hadamard product, and we added such support in a following read compute version.
> The way we evolved in the past:
> 
> Added a new read compute version to include new read compute operators.
> Added the new version support in the backend.
> Rolling out the backend support first.
> Rolling out the new client version.
> In case we rollback backend because of the bug/regression, the backend can fail fast because of unsupported read compute version.
> These are the historical discussions I remembered.
> In the future, the situation can change considering we are planning to add client read compute support across all the clients.
> If we want to add new read compute operators, we can always add the client side support regardless of whether the backend supports it or not. For example, if the backend doesn't support the new operator, the client compute can be the fallback.
> 
> Is this idea similar to the one in your mind with this PR?

My understanding of the historical context is in line with what you described. The thing is, given the changes we did, which were to add new operation types to the union of operations, there is actually no need to maintain the fork in the server-side code, since the latest schema (v3) has a union which is a superset of the previous unions, and thus it can deserialize all previous versions too (it is backwards compatible).

Furthermore... at some point (for reasons that aren't completely clear...) we made all versions of the compute request schema exactly the same. So the current fork in the code just makes the code a bit confusing, but without adding any value, IIUC.

gaojieliu:
From the schema evolution perspective, it is backward compatible, and introducing a new compute version is to guard the following scenario: backend is using a lower version of compute than the client, and when that happens, backend will fail fast based on the compute version in the request.
If it is always guaranteed that backend will always be using a higher version of compute than client and backward compatible, there is no need to to maintain different compute versions.

FelixGV:
> From the schema evolution perspective, it is backward compatible, and introducing a new compute version is to guard the following scenario: backend is using a lower version of compute than the client, and when that happens, backend will fail fast based on the compute version in the request.
> If it is always guaranteed that backend will always be using a higher version of compute than client and backward compatible, there is no need to to maintain different compute versions.

Yes, I think we are aligned. To clarify further:

1. The backend needs to support the minimum functionality the client needs, which can be ensured via the API version request header.
2. Assuming the check above succeeds, then the operation declarations need to be deserialized and for this, the latest operations schema will do (at least in the cases we have so far) so there is no need to keep various schemas nor their `SpecificRecord`.

Makes sense?

services/venice-router/src/main/java/com/linkedin/venice/router/api/path/VeniceComputePath.java
```diff
@@ -22,30 +22,44 @@
...
+import org.apache.avro.generic.GenericRecord;
 import org.apache.avro.io.BinaryDecoder;
 import org.apache.avro.io.OptimizedBinaryDecoderFactory;
 
 
 public class VeniceComputePath extends VeniceMultiKeyPath<ComputeRouterRequestKeyV1> {
+  private static final Schema EMPTY_RECORD_SCHEMA = Schema.createRecord(
+      ComputeRequestV3.class.getSimpleName(),
+      "no-op",
+      ComputeRequestV3.class.getPackage().getName(),
```
gaojieliu, line 42: With this logic, if v4 becomes a wire protocol or we add a v5, which contains some new optional field, we need to change `ComputeRequestV3` to the latest protocol, right?
If v5 adds a new optional field instead of adding a branch in the union types, we will need to use the wire protocol version to deserialize it rather than the latest compute version: v5, otherwise, the deserialization won't work from my understanding.
Can you clarify?
Can you clarify?

internal/venice-client-common/src/main/java/com/linkedin/venice/compute/ComputeUtils.java
```diff
@@ -41,11 +48,22 @@ public class ComputeUtils {
...
+   * {@link ComputeRequest}, with the only difference between the two being that the items of the operations list
+   * in the former are defined as a union of one type, which unfortunately results in the SpecificRecord typing this
+   * as a {@link List<Object>}. This is a design shortcoming, but which we cannot easily fix, since there are already
+   * clients using this protocol. On the server-side, however, we wish to use proper types without lots of casting,
+   * which we can achieve by letting Avro do the evolution.
+   */
+  private static final RecordDeserializer<ComputeRequest> DESERIALIZER =
+      getFastAvroSpecificDeserializer(ComputeRequestV3.SCHEMA$, ComputeRequest.class);
+
+  public static ComputeRequest deserializeComputeRequest(BinaryDecoder decoder, ComputeRequest reuse) {
```
gaojieliu, line 62: Same question here.
If v5 adds a new optional field in the top-level, we have to use the right write schema to decode it, right?

gaojieliu (commented):
Left some clarification comments.

services/venice-router/src/main/java/com/linkedin/venice/router/api/path/VeniceComputePath.java
FelixGV: We must always know the correct writer schema (which is the one the client used to encode its request). Previously and in this commit too we have the assumption that the backend knows all protocol versions the client might use, and if that assumption is violated, it's ok to fail the request.

Making new clients compatible with older backends is not supported (neither before nor in this commit).

gaojieliu: I may have a misunderstanding here.. 
```
  private static final RecordDeserializer<GenericRecord> COMPUTE_REQUEST_NO_OP_DESERIALIZER =
      FastSerializerDeserializerFactory.getFastAvroGenericDeserializer(ComputeRequestV3.SCHEMA$, EMPTY_RECORD_SCHEMA);
```
So the writer schema is always v3, and it will certainly work now since all other versions are not being used as wire protocol.
In the future, I assume we will change this logic when we add more wire protocol versions, right?

internal/venice-client-common/src/main/java/com/linkedin/venice/compute/ComputeUtils.java
FelixGV: Yes that's correct.

services/venice-router/src/main/java/com/linkedin/venice/router/api/path/VeniceComputePath.java
FelixGV: Yes, that would need to change. Depending on how the protocol is evolved, the change here could differ, e.g.:

1. If we add more branches to the same union we kept evolving so far, then this writer schema can be bumped to the latest and that's it. A single schema could still deserialize all versions.

2. If we we add fields elsewhere in the schema, then we would need two deserializers, this old one we already have plus a new one, and we would need to pick the right one based on the API version header.

gaojieliu: SG.
Can you add some javadoc about ^?
It is a little confusing when first read this piece of code, and also, the dev of read compute needs to be aware of this when evolving read compute schemas.

FelixGV: Will do. I can't get to it today but I'll update the PR on Monday. Thanks for the reviews!

FelixGV: Done

PR #559, opened 2023-08-01
[server] Adopted RocksDB log rotation and enabled fast-avro in LF write-compute path

gaojieliu:
For a long-running stream-only stores, RocksDB can accumulated a lot of old logs files in RocksDB data folder, which is a waste of resources. RocksDB has log rotation, and Venice Server exposes the following new configs: rocksdb.max.log.file.num: default 3
rocksdb.max.log.file.size: default 10MB

This code change also adopted fast-avro in LF write-compute path. The original code is using map-order-preserved serde, which is not necessary for non-AA write-compute stores since DCR validation workflow won't run against such stores.
When fast-avro is enabled, the map-order is not preserved in serde.

This code change also reduces two map lookups in the write compute code path of LF model.

TODO: evaluate how to enable fast-avro in AA WC stores.

## How was this PR tested?
CI


clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreWriteComputeProcessor.java
```diff
@@ -118,12 +120,34 @@ private SchemaAndUniqueId getSchemaAndUniqueId(int valueSchemaId, int writeCompu
...
+
+  private RecordSerializer<GenericRecord> getValueSerializer(int valueSchemaId) {
     return serializerIndexedBySchemaId.computeIfAbsent(valueSchemaId, this::generateValueSerializer);
   }
 
-  private AvroSerializer<GenericRecord> generateValueSerializer(int valueSchemaId) {
+  RecordSerializer<GenericRecord> generateValueSerializer(int valueSchemaId) {
     Schema valueSchema = getValueSchema(valueSchemaId);
+    /**
+     * TODO: why do we need to preserve the map-order since this class is only used by {@link LeaderFollowerStoreIngestionTask}.
```
FelixGV, line 145: So you are certain that this class is not used in AA, right?

sixpluszero: The only public API "applyWriteCompute" is only used by LFSIT#handleUpdateRequest, which should only be used by non-AA store.

gaojieliu: yeah, I am certain based on the call stack, and AA is using this method:
https://github.com/linkedin/venice/blob/main/internal/venice-client-common/src/main/java/com/linkedin/venice/schema/writecompute/WriteComputeProcessor.java#L58

FelixGV: Ok I see. It makes sense that LF and AA have separate code paths, since their logic needs to be a bit different. On the other hand, it seems a bit hard to maintain, and the class names are very similar (of course, since they do almost the same thing...). Hopefully, we could find a better way to structure this code in the future.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBServerConfig.java
```diff
@@ -372,6 +377,9 @@ public RocksDBServerConfig(VeniceProperties props) {
     this.separateRMDCacheEnabled = props.getBoolean(ROCKSDB_SEPARATE_RMD_CACHE_ENABLED, false);
 
     this.blockBaseFormatVersion = props.getInt(ROCKSDB_BLOCK_BASE_FORMAT_VERSION, 2);
+
+    this.maxLogFileNum = props.getInt(ROCKSDB_MAX_LOG_FILE_NUM, 3);
+    this.maxLogFileSize = props.getSizeInBytes(ROCKSDB_MAX_LOG_FILE_SIZE, 10 * 1024 * 1024); // 10MB;
```
FelixGV, line 382: What is the granularity of this 3 * 10 MB limit? Is it per Venice partition? Per store-version? Per host? Let's add JavaDoc that mentions this please.

FelixGV (approved):
Thanks, LGTM!

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBServerConfig.java
gaojieliu: Per partition, and updated the javadoc.

FelixGV (approved):

PR #562, opened 2023-08-01
[test] Support additional properties passing for test utils.

haoxu07:

Previously, integration test will create some adhoc topic manager, venice writer or producer by passing raw properties without additional properties needed for other pub sub system type. This will cause integration test failure with other pub sub system. To resolve this, we changed some interfaces of utility function of `TestUtils` function by passing these additional configs or adding these additional configs directly inside the utility function if `PubSubBroker` is available. 

### DIFF
--- a/internal/venice-test-common/src/main/java/com/linkedin/venice/utils/TestUtils.java
+++ b/internal/venice-test-common/src/main/java/com/linkedin/venice/utils/TestUtils.java
@@ -264,7 +264,8 @@ public class TestUtils {
       String keySchema,
       String valueSchema,
       Stream<Map.Entry> batchData,
-      PubSubProducerAdapterFactory pubSubProducerAdapterFactory) {
+      PubSubProducerAdapterFactory pubSubProducerAdapterFactory,
+      Map<String, String> additionalProperties) {
     return createVersionWithBatchData(
         controllerClient,
         storeName,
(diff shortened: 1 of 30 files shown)


FelixGV (dismissed):
Thanks Hao!

internal/venice-test-common/src/main/java/com/linkedin/venice/utils/TestUtils.java
```diff
@@ -327,11 +339,13 @@ public static void writeBatchData(
...
       CompressionStrategy compressionStrategy,
       Function<String, ByteBuffer> compressionDictionaryGenerator,
-      PubSubProducerAdapterFactory pubSubProducerAdapterFactory) {
+      PubSubProducerAdapterFactory pubSubProducerAdapterFactory,
+      Map<String, String> additionalProperties) {
     Properties props = new Properties();
     props.put(KAFKA_BOOTSTRAP_SERVERS, response.getKafkaBootstrapServers());
     props.setProperty(PARTITIONER_CLASS, response.getPartitionerClass());
     props.putAll(response.getPartitionerParams());
+    props.putAll(additionalProperties);
```
sixpluszero, line 348: nit: for the priority of the additional property you are adding, maybe let's move it one line down to become the last props it is adding to prevent it being overwritten by below configs.

sixpluszero (approved):
Thanks Hao, overall it is good, left a nit and should be resolved easily.

internal/venice-test-common/src/main/java/com/linkedin/venice/utils/TestUtils.java
haoxu07: Good point, will change it.

sixpluszero (approved):

PR #565, opened 2023-08-02
[server][fc] Integrate server side handlers with gRPC service + mTLS

rkunds:
## [server][fc] Integrate server side handlers with gRPC service + mTLS

- Adapts Netty Channel pipeline and corresponding handler classes to be compatible with gRPC requests
- Can collect server side statistics and enforce read quota for gRPC requests
- Introduce mutual TLS support on gRPC enabled servers and fast client, when an SSLFactory is present
- Update unit tests and integration tests 

## How was this PR tested?

CI/Unit Tests

If SSL is enabled for a gRPC enabled Venice Server, we start the gRPC server with SSL active and Client Auth required. If this is the case, we have to pass in an SSLFactory instance in the Client Configs when creating a gRPC enabled FC. If this isn't the case, then we would not be able to connect to a Venice Server instance.


internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/fastclient/AvroStoreClientEndToEndTest.java
```diff
@@ -228,12 +228,13 @@ protected void runTest(
     }
   }
 
-  @Test(dataProvider = "FastClient-Four-Boolean-A-Number-Store-Metadata-Fetch-Mode", timeOut = TIME_OUT)
+  @Test(dataProvider = "FastClient-Five-Boolean-A-Number-Store-Metadata-Fetch-Mode", timeOut = TIME_OUT)
```
rkunds, line 231: We use this to test the server-side stats collected by gRPC enabled server with fast clients. This increases the total amount of time that integration tests take to run by a significant amount D: Would separating these tests be a good idea in this case?

FelixGV: Thanks for flagging this, but I think it's fine... this class is part of `IntegrationTestsI`, and also extended by a couple of other classes which are part of `IntegrationTestsG`. Based on your latest CI run, these two tasks took 20m and 15m respectively, which is less than the longest running task, so it should not increase the total time the test suite takes.

internal/venice-common/src/main/java/com/linkedin/venice/acl/handler/StoreAclHandler.java
```diff
@@ -201,4 +205,15 @@ public void channelRead0(ChannelHandlerContext ctx, HttpRequest req) throws SSLP
...
     }
   }
+
+  @Override
+  public <ReqT, RespT> ServerCall.Listener<ReqT> interceptCall(
+      ServerCall<ReqT, RespT> call,
+      Metadata headers,
+      ServerCallHandler<ReqT, RespT> next) {
+    // TODO: Have to implement StoreAclHandler for gRPC
+    // - need to implement client-side interceptor to include information such as request path and method in
```
FelixGV, line 215: Hmm... why would the server ACL depend on client-side information? If today the server is already capable of serving the gRPC request (without ACL of course) then it must already know which store is being queried, right? So why would it need to receive an extra request path and method?

services/venice-server/src/main/java/com/linkedin/venice/listener/OutboundHttpWrapperHandler.java
```diff
@@ -162,4 +169,65 @@ public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise)
...
+    ByteBuf body;
+    CompressionStrategy compressionStrategy = CompressionStrategy.NO_OP;
+
+    ReadResponse obj = ctx.getReadResponse();
+    GrpcStatsContext statsContext = ctx.getGrpcStatsContext();
+    VeniceServerResponse.Builder veniceServerResponseBuilder = ctx.getVeniceServerResponseBuilder();
+    if (!ctx.hasError()) {
+      compressionStrategy = obj.getCompressionStrategy();
+
+      statsContext.setDatabaseLookupLatency(obj.getDatabaseLookupLatency());
```
FelixGV, line 190: Hmm... the stuff here seems copy-pasted... can you extract common pieces of code into functions that will be called from both paths?

services/venice-server/src/main/java/com/linkedin/venice/listener/ReadQuotaEnforcementHandler.java
```diff
@@ -190,6 +194,86 @@ public void channelRead0(ChannelHandlerContext ctx, RouterRequest request) {
     stats.recordReadQuotaUsage(storeName, storageNodeBucket.getStaleUsageRatio());
   }
 
+  @Override
+  public void grpcRead(GrpcHandlerContext ctx, GrpcHandlerPipeline pipeline) {
+    RouterRequest request = ctx.getRouterRequest();
+    String storeName = request.getStoreName();
```
FelixGV, line 200: This again looks quite duplicated. Can we consolidate it please?

services/venice-server/src/main/java/com/linkedin/venice/listener/StatsHandler.java
```diff
@@ -245,6 +250,62 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) {
...
+    }
+
+    if (ctx.getGrpcStatsContext() == null) {
+      // new request
+      GrpcStatsContext statsContext = new GrpcStatsContext(singleGetStats, multiGetStats, computeStats);
+
+      ctx.setGrpcStatsContext(statsContext);
+      pipeline.processRequest(ctx);
+    } else {
+      // existing request
```
FelixGV, line 267: Does this mean it's a multi-part request? In the old implementation, we track the request part count, as we have found it to have a big influence on performance. Maybe this cannot happen in gRPC, but I suppose it probably can, so it may be good to track it as well.

services/venice-server/src/main/java/com/linkedin/venice/listener/StatsHandler.java
```diff
@@ -245,6 +250,62 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) {
...
+      pipeline.processRequest(ctx);
```
FelixGV, line 268: If this is the same statement at the end of both branches of the if, then we can move it out of the if, and always execute it after.

services/venice-server/src/main/java/com/linkedin/venice/listener/StorageReadRequestHandler.java
```diff
@@ -349,6 +353,68 @@ public void channelRead(ChannelHandlerContext context, Object message) throws Ex
     }
   }
 
+  @Override
+  public void grpcRead(GrpcHandlerContext ctx, GrpcHandlerPipeline pipeline) {
```
FelixGV, line 357: Is it the right understanding that this code executes the request inline within gRPC's (Netty?) worker thread? In the old code, we delegate the request execution to another thread pool. I wonder if we should do the same in gRPC (and if it's even possible at all).

services/venice-server/src/main/java/com/linkedin/venice/listener/grpc/GrpcStatsContext.java
```diff
@@ -0,0 +1,363 @@
...
+
+public class GrpcStatsContext {
+  /**
+   * We need to be able to record server side statistics for gRPC requests. The current StatsHandler in the Netty
+   * Pipeline maintains instance variables per channel, and guarantees that each request will be handled by the same
+   * thread, thus we cannot augment StatsHandler in the same way as other handlers for gRPC requests. We create a
+   * new StatsContext for each gRPC request, so we can maintain per request stats which will be aggregated on request
+   * completion using references to the proper Metrics Repository in AggServerHttpRequestStats. This class is almost a
+   * direct copy of StatsHandler, without Netty Channel Read/Write logic.
+   */
```
FelixGV, line 21: Can we call this class `ServerStatsContext` and have the `StatsHandler` hold an instance of this new class, rather than holding each individual stat as a class property? This way both paths could use the same code without duplication.

FelixGV (changes requested):
Thanks a lot Rohit! Let's iterate a bit more to consolidate the common code and avoid duplication. I pointed out a few places but may have missed some... In general, we should always keep in the back of our mind that cutting and pasting is fine, but copy and pasting usually leads to maintenance issues later down the road... Of course, the best is straight up deleting without repasting anywhere else 😆

internal/venice-common/src/main/java/com/linkedin/venice/acl/handler/StoreAclHandler.java
rkunds: When using interceptors, it is a little tricky to access information about the message that is being sent (where information such as store/resource is being held). We would be able to add additional metadata from the client side in order to accommodate for this, but after taking another look, we should be able to access the message in this context as well.

services/venice-server/src/main/java/com/linkedin/venice/listener/StatsHandler.java
rkunds: As of right now, we are only sending Unary requests to client/server. In this case, the entire request will be in one part. When message streaming is enabled for the gRPC service, we might want to keep track of stats for each message that is sent. For now, I will remove the second part -- it will never be invoked in this case.

services/venice-server/src/main/java/com/linkedin/venice/listener/StorageReadRequestHandler.java
rkunds: That is correct, it will be executed with a gRPC worker thread. We can define the executor that gRPC uses to better suit our needs, but at the same time we should be able to reuse or define another thread pool to handle storage requests for grpc.

internal/venice-common/src/main/java/com/linkedin/venice/acl/handler/StoreAclHandler.java
FelixGV: The issue I am concerned about is... if the client needs to specify store twice in the request, with one of them being the store that will be queried, and the other one being the store the ACL will be evaluated against, then a rogue client could specify different values for these and read data the ACL would not otherwise allow. So I would prefer that the ACL be evaluated strictly against the resource that is going to be read, not some side-channel that is equal to it merely by convention.

services/venice-server/src/main/java/com/linkedin/venice/listener/StatsHandler.java
```diff
@@ -245,6 +250,62 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) {
...
+
+    double elapsedTime = LatencyUtils.getLatencyInMS(statsContext.getRequestStartTimeInNS());
+
+    if (!ctx.hasError() && !responseStatus.equals(OK) || responseStatus.equals(NOT_FOUND)) {
+      statsContext.successRequest(serverHttpRequestStats, elapsedTime);
+    } else {
+      statsContext.errorRequest(serverHttpRequestStats, elapsedTime);
+    }
+
+    pipeline.processResponse(ctx);
```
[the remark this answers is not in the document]
rkunds: I think we should be able to do this (?) `ctx.writeAndFlush` are done async which isn't exactly possible with my implementation of the gRPC handler pipeline. We can send the response back to the client here, and collect stats once that has been completed.

sushantmane (changes requested):
I'm yet to finish full review. Will try to finish it today

services/venice-server/src/main/java/com/linkedin/venice/listener/grpc/GrpcHandlerPipeline.java
```diff
@@ -0,0 +1,93 @@
...
+  private int inboundIdx;
+  private int outboundIdx;
+  private boolean completed = false;
+
+  public GrpcHandlerPipeline() {
+    inboundHandlers = new ArrayList<>();
+    outboundHandlers = new ArrayList<>();
+
+    inboundIdx = 0;
+    outboundIdx = 0;
```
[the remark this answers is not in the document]
rkunds: We can't make them final, they are being changed when the next handler is being called.

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/transport/GrpcTransportClient.java
```diff
@@ -176,5 +217,29 @@ public void onCompleted() {
 
       return valueFuture;
     }
+
+    private void handleError(VeniceServerResponse response) {
```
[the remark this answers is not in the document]
rkunds: should be resolved in latest commit.

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/transport/GrpcTransportClient.java
```diff
@@ -155,17 +200,20 @@ public CompletableFuture<TransportClientResponse> post() {
       clientStub.batchGet(request, new StreamObserver<VeniceServerResponse>() {
         @Override
         public void onNext(VeniceServerResponse value) {
+          LOGGER.debug("Performing BatchGet in gRPC");
```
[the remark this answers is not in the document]
rkunds: i think we should just remove this log.

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/transport/GrpcTransportClient.java
```diff
@@ -23,17 +35,46 @@
...
+    this.clientConfig = clientConfig;
+    this.grpcClientConfig = clientConfig.getGrpcClientConfig();
+
     serverGrpcChannels = new VeniceConcurrentHashMap<>();
     stubCache = new VeniceConcurrentHashMap<>();
-    nettyAddressToGrpcAddressMap = clientConfig.getNettyServerToGrpcAddressMap();
-    this.r2TransportClient = new R2TransportClient(clientConfig.getR2Client());
+
+    nettyAddressToGrpcAddressMap = grpcClientConfig.getNettyServerToGrpcAddressMap();
+    r2TransportClient = new R2TransportClient(grpcClientConfig.getR2Client());
```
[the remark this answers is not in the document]
rkunds: We can't cast R2Client to R2TransportClient. At this layer within FC, we need the transport client itself in order to be able to send requests. In order to instantiate an R2TransportClient, we require construction in this manner.

internal/venice-common/src/main/java/com/linkedin/venice/listener/VerifySslHandler.java
```diff
@@ -47,4 +54,30 @@ public void channelRead0(ChannelHandlerContext ctx, HttpRequest req) {
...
     }
   }
+
+  @Override
+  public <ReqT, RespT> ServerCall.Listener<ReqT> interceptCall(
+      ServerCall<ReqT, RespT> call,
+      Metadata headers,
+      ServerCallHandler<ReqT, RespT> next) {
+    // check if SSL is enabled for this call
+    if (call.getAttributes().get(Grpc.TRANSPORT_ATTR_SSL_SESSION) == null) {
```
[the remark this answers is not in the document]
rkunds: it should not, if it is null, then the request was not made over SSL. However, I think this condition should be inverted.

services/venice-server/src/main/java/com/linkedin/venice/grpc/VeniceGrpcServerConfig.java
```diff
@@ -82,15 +114,18 @@ private void verifyAndAddDefaults() {
...
-        credentials = InsecureServerCredentials.create();
-      }
       if (service == null) {
         throw new IllegalArgumentException("Service must be set");
       }
       if (interceptors == null) {
         interceptors = Collections.emptyList();
       }
+      if (numThreads == 0) {
+        numThreads = 20;
```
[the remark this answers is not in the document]
rkunds: do we have access to that field in ListenerService?

sushantmane (changes requested):
I have a few more comments left. 

I believe there's room for further simplification in `GrpcHandlerPipeline` and `VeniceGrpcHandler`.  For example, we should create just one instance of `GrpcHandlerPipeline` and use it for all the requests.

services/venice-server/src/main/java/com/linkedin/venice/listener/grpc/VeniceReadServiceImpl.java
```diff
@@ -1,81 +1,39 @@
...
-
-    return VeniceServerResponse.newBuilder()
-        .setData(ByteString.copyFrom(data.array()))
-        .setSchemaId(schemaId)
-        .setCompressionStrategy(compressionStrategy.getValue())
-        .build();
+  private void handleRequest(VeniceClientRequest request, StreamObserver<VeniceServerResponse> responseObserver) {
+    VeniceServerResponse.Builder responseBuilder = VeniceServerResponse.newBuilder().setErrorCode(200);
+    GrpcHandlerContext ctx = new GrpcHandlerContext(request, responseBuilder, responseObserver);
+    GrpcHandlerPipeline requestPipeline = handlerPipeline.getNewPipeline();
```
sushantmane, line 34: This doesn't feel right, we should not create pipeline for each request

services/venice-server/src/main/java/com/linkedin/venice/listener/grpc/GrpcHandlerPipeline.java
sushantmane, line 1: I think this class can be written in a way that does not require one instance of it per request

sushantmane (changes requested):

internal/venice-common/src/main/java/com/linkedin/venice/listener/VerifySslHandler.java
```diff
@@ -47,4 +54,29 @@ public void channelRead0(ChannelHandlerContext ctx, HttpRequest req) {
...
+  @Override
+  public <ReqT, RespT> ServerCall.Listener<ReqT> interceptCall(
+      ServerCall<ReqT, RespT> call,
+      Metadata headers,
+      ServerCallHandler<ReqT, RespT> next) {
+    // check if SSL is enabled for this call
+    if (call.getAttributes().get(Grpc.TRANSPORT_ATTR_SSL_SESSION) != null) {
+      return next.startCall(call, headers);
+    }
+    LOGGER.info("test {}", call.getAttributes().get(Grpc.TRANSPORT_ATTR_SSL_SESSION));
```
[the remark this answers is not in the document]
rkunds: this log should have been removed 😓

services/venice-server/src/main/java/com/linkedin/venice/listener/grpc/handlers/GrpcOutboundStatsHandler.java
```diff
@@ -0,0 +1,42 @@
...
+
+public class GrpcOutboundStatsHandler extends VeniceServerGrpcHandler {
+  @Override
+  public void processNext(GrpcHandlerContext ctx) {
+    writeResponse(ctx);
+
+    // do not exit early if there is an error, we want to record stats for error cases as well
+    ServerStatsContext statsContext = ctx.getGrpcStatsContext();
+    HttpResponseStatus responseStatus = statsContext.getResponseStatus();
+    if (statsContext.getResponseStatus() == null) {
```
sushantmane, line 21: IMO we should return error response instead of throwing an exception. We could additionally add a log statement

services/venice-server/src/main/java/com/linkedin/venice/listener/grpc/handlers/VeniceServerGrpcHandler.java
```diff
@@ -0,0 +1,41 @@
...
+
+public abstract class VeniceServerGrpcHandler {
+  protected VeniceServerGrpcHandler next;
+
+  public VeniceServerGrpcHandler() {
+  }
+
+  public void addNextHandler(VeniceServerGrpcHandler next) {
+    VeniceServerGrpcHandler current = this;
+    while (current.next != null) {
```
sushantmane, line 16: Imo we should loop in `VSGRP::addHandler`. Here we should just to `this.next = next`

services/venice-server/src/main/java/com/linkedin/venice/listener/HttpChannelInitializer.java
```diff
@@ -247,6 +254,27 @@ public GrpcHandlerPipeline initGrpcHandlers() {
     return pipeline;
   }
 
+  public VeniceServerGrpcRequestProcessor initGrpcServerHandler() {
+    VeniceServerGrpcRequestProcessor grpcServerHandler = new VeniceServerGrpcRequestProcessor();
```
sushantmane, line 258: change var name to `grpcServerRequestProcessor`

services/venice-server/src/main/java/com/linkedin/venice/listener/ReadQuotaEnforcementHandler.java
```diff
@@ -247,26 +250,30 @@ public void grpcRead(GrpcHandlerContext ctx, GrpcHandlerPipeline pipeline) {
...
     if (!storageNodeBucket.tryConsume(rcu)) {
       if (handleServerOverCapacity(null, ctx, pipeline, storeName, rcu, true))
-        return;
+        pipeline.processRequest(ctx);
+      return;
     }
 
     handleEpilogue(null, ctx, pipeline, request, storeName, rcu, true);
+
+    pipeline.processRequest(ctx);
```
sushantmane, line 268: Let's delete old code

services/venice-server/src/main/java/com/linkedin/venice/listener/StatsHandler.java
```diff
@@ -37,6 +36,10 @@ public StatsHandler(
     serverStatsContext = new ServerStatsContext(singleGetStats, multiGetStats, computeStats);
   }
 
+  public ServerStatsContext constructStatsContext() {
```
sushantmane, line 39: Why do we need to build stats context here and out of `StatsHandler`? Why not pass it as dependency ?

services/venice-server/src/main/java/com/linkedin/venice/listener/grpc/handlers/GrpcOutboundResponseHandler.java
```diff
@@ -0,0 +1,59 @@
...
+      statsContext.setResponseStatus(HttpResponseStatus.BAD_REQUEST);
+      body = Unpooled.EMPTY_BUFFER;
+    }
+
+    byte[] array = new byte[body.readableBytes()];
+    body.getBytes(body.readerIndex(), array);
+    veniceServerResponseBuilder.setData(ByteString.copyFrom(array))
+        .setCompressionStrategy(compressionStrategy.getValue());
+
+    if (ctx.hasError())
```
sushantmane, line 54: let's use braces. For an empty response can we create a static byte array and use it everytime we want to send error response. Something like
```java 
public static byte[] EMPTY_BUFFER = new byte[0]
```

services/venice-server/src/main/java/com/linkedin/venice/listener/grpc/handlers/VeniceServerGrpcHandler.java
```diff
@@ -0,0 +1,41 @@
...
+      current = current.next;
+    }
+
+    current.next = next;
+  }
+
+  public abstract void processNext(GrpcHandlerContext ctx);
+
+  public void sendError(GrpcHandlerContext ctx) {
```
sushantmane, line 25: I think we can get rid of sendError since in each processNext we are checking ctx.hasError

services/venice-server/src/main/java/com/linkedin/venice/listener/grpc/handlers/VeniceServerGrpcRequestProcessor.java
```diff
@@ -0,0 +1,28 @@
...
+
+  public void addHandler(VeniceServerGrpcHandler handler) {
+    if (head == null) {
+      head = handler;
+    } else {
+      head.addNextHandler(handler);
+    }
+  }
+
+  public boolean hasNext() {
```
sushantmane, line 21: do we need this method?

sushantmane (changes requested):

services/venice-server/src/main/java/com/linkedin/venice/listener/StatsHandler.java
rkunds: i guess we can... if we did so we would have to construct the context within listener service as we need the three aggregate stats. We also provide a method to create the stats context in order to be able to use this for gRPC functionality.

services/venice-server/src/main/java/com/linkedin/venice/listener/grpc/handlers/GrpcOutboundResponseHandler.java
rkunds: Is there a benefit to this? `Unpooled.EMPTY_BUFFER` would achieve the same functionality as using the above method. .

services/venice-server/src/main/java/com/linkedin/venice/listener/grpc/handlers/VeniceServerGrpcHandler.java
rkunds: i think we should not do the check. We can pass the request along via sendError and send the request back in the last handler (outbound stats). This way we can still achieve the same functionality where we record metrics for failed/errored requests.

services/venice-server/src/main/java/com/linkedin/venice/listener/grpc/handlers/VeniceServerGrpcRequestProcessor.java
rkunds: no 😅

services/venice-server/src/main/java/com/linkedin/venice/listener/grpc/handlers/VeniceServerGrpcHandler.java
sushantmane: Yeah let's keep do one of the these: check ctx has error set or not or just use next.sendError

internal/venice-common/src/main/java/com/linkedin/venice/acl/handler/StoreAclHandler.java
sushantmane, line 1: let's keep old code as is as the newly added method is not reducing duplication much.

internal/venice-common/src/test/java/com/linkedin/venice/acl/StoreAclHandlerTest.java
sushantmane, line 1: let's delete dead code

sushantmane (changes requested):

sushantmane (approved):
LGTM! Thanks a lot, @rkunds for great work on this PR!

FelixGV (approved):

FelixGV:
🚀

PR #566, opened 2023-08-02
[server] Make sure TransientRecord saves correct value and RMD manifest

sixpluszero:
## [server] Make sure TransientRecord saves correct value and RMD manifest

This PR fixes the bug that TransientRecord is not saving correct value/RMD manifest when updating a key. In the current code, it will try to save the old value/RMD manifest and it will be fetching the same value/RMD manifest until cache expires. This mean chunk is still leaking. 
This PR fixes it by updating the TransientRecord cache's manifests when VeniceWriter invokes ChunkAwareCallback#setChunkingInfo() API. This is prior to the next value record processing so it will be processed correctly.

## How was this PR tested?

Will add unit test to make sure LPC API is called.

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderProducerCallback.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderProducerCallback.java
@@ -250,6 +250,7 @@ public class LeaderProducerCallback implements ChunkAwareCallback {
     this.oldValueManifest = oldValueManifest;
     this.oldRmdManifest = oldRmdManifest;
     if (getPartitionConsumptionState() == null) {
+      LOGGER.error("PartitionConsumptionState is missing in chunk producer callback");
       return;
     }
     // TransientRecord map is indexed by non-chunked key.
@@ -258,6 +259,8 @@ public class LeaderProducerCallback implements ChunkAwareCallback {
     if (record != null) {
       record.setValueManifest(chunkedValueManifest);
       record.setRmdManifest(chunkedRmdManifest);
+    } else {
+      LOGGER.error("Transient record is missing when trying to update value/RMD manifest.");
     }
   }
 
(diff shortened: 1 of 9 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/PartitionConsumptionState.java
```diff
@@ -603,10 +604,18 @@ public ChunkedValueManifest getRmdManifest() {
       return rmdManifest;
     }
 
+    public void setRmdManifest(ChunkedValueManifest rmdManifest) {
```
gaojieliu, line 607: Shall we remove the following params from the constructor since they are always wrong?
```
ChunkedValueManifest valueManifest,
      ChunkedValueManifest rmdManifest
```

sixpluszero: Yes, thanks for pointing out.  Will do in the next update.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderProducerCallback.java
```diff
@@ -249,6 +249,16 @@ public void setChunkingInfo(
     this.rmdChunks = rmdChunks;
     this.oldValueManifest = oldValueManifest;
     this.oldRmdManifest = oldRmdManifest;
+    if (getPartitionConsumptionState() == null) {
```
gaojieliu, line 252: If this happens, shall we log a error message?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderProducerCallback.java
```diff
@@ -249,6 +249,16 @@ public void setChunkingInfo(
...
+      return;
+    }
+    // TransientRecord map is indexed by non-chunked key.
+    PartitionConsumptionState.TransientRecord record =
+        getPartitionConsumptionState().getTransientRecord(getSourceConsumerRecord().getKey().getKey());
+    if (record != null) {
```
gaojieliu, line 258: If `record` is `null`, shall we log an error message?

gaojieliu (commented):
Also can we write an E2E integration tests to validate the issue is actually fixed?

gaojieliu (approved):

PR #567, opened 2023-08-02
[server][test] Fix log statements, delete static config files, and replace Kafka with PubSub at some places

sushantmane:
## Fix log statements, delete static config files, and replace Kafka with PubSub at some places
- Fixed log statement in `LeaderFollowerStoreIngestionTask` to log correct name of the incoming message
- Renamed some classes which are not Kafka specific to use `PubSub` instead of `Kafka` in their name
- Removed century old config files: `venice-server/config/server.properties`, `venice-server/config/cluster.properties`, and `venice-controller/config/controller.properties`.  These were used in few tests and had very no longer supported configs. Replaced it explicit properties in respective tests. 


internal/venice-common/src/main/java/com/linkedin/venice/pubsub/adapter/kafka/admin/InstrumentedPubSubAdminAdapter.java
```diff
@@ -0,0 +1,165 @@
+package com.linkedin.venice.pubsub.adapter.kafka.admin;
```
haoxu07, line 1: Minor: this is might be previously left problem. I think we may need to move this to higher module kafka, since other pub sub adpaters could also use this.

haoxu07 (dismissed):
Thanks for these cleanups!

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/adapter/kafka/admin/InstrumentedPubSubAdminAdapter.java
sushantmane: yeah that's right. Moved it to pubsub api module. Thanks!

haoxu07 (approved):
LGTM!

PR #568, opened 2023-08-03
[controller] Store Migration Fix: destination cluster will replica version configs from source cluster

huangminchn:
## Summary
Bug fixed in this PR: During store migration, versions created in the destination cluster only replicates the hybrid configs, partition count from the version configs in the source cluster; other configs values are from the store configs. However, store configs might change after current version is created; if destination cluster uses those updated store configs, it could conflict with the current version resources like Kafka topics, Helix resource, etc.

Incident due to the above bug: a store didn't have compression enabled when current version was created, and later compression was enabled. After store migration, destination cluster thought compression was enabled and thus trying to leverage ZSTD dictionary in the topic which didn't exist.
Chunking feature would have the same problem, like destination cluster though topic is not chunked because chunking is disable after current version is created.

This PR fixed the bug by replicating version configs from the source cluster - while adding version:
Get version configs from the source cluster in the same region (must copy from the same region) if
1) The store is migrating;
2) the cluster processing the add version message is the destination cluster;
3) the version number being added is smaller or equal to the current version in the source cluster,
   meaning that the immutable version level resources are already created.
Otherwise, apply the store configs to the new version.

The fix requires updating the child.cluster.allowlist config for the child controller too, child controller needs this config in order to construct the region to child controller D2 address.

## How was this PR tested?
An integration test is added to verify that version configs in the destination cluster would be exactly the same as the version configs from the source cluster; configs would be changed before starting the store migration, simulating the incident mentioned above.
Confirmed that the new test can't pass without the fix.

### DIFF
--- a/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
@@ -2268,6 +2268,55 @@ public class VeniceHelixAdmin implements Admin, StoreCleaner {
         null);
   }
 
+  private Optional<Version> getVersionFromSourceCluster(
+      ReadWriteStoreRepository repository,
+      String destinationCluster,
+      String storeName,
+      int versionNumber) {
+    Store store = repository.getStore(storeName);
+    if (store == null) {
+      throwStoreDoesNotExist(destinationCluster, storeName);
+    }
+    if (!store.isMigrating()) {
+      return Optional.empty();
+    }
+    ZkStoreConfigAccessor storeConfigAccessor = getStoreConfigAccessor(destinationCluster);
+    StoreConfig storeConfig = storeConfigAccessor.getStoreConfig(storeName);
+    if (!destinationCluster.equals(storeConfig.getMigrationDestCluster())) {
+      // destinationCluster is passed from the add version task, so if the task is not in the destination cluster but in
+      // the source cluster, do not replicate anything from the "source" since this is the source cluster
+      return Optional.empty();
+    }
+    String migrationSourceCluster = storeConfig.getMigrationSrcCluster();
+    ControllerClient srcControllerClient = getControllerClientMap(migrationSourceCluster).get(getRegionName());
+    if (srcControllerClient == null) {
+      throw new VeniceException(
+          "Failed to constructed controller client for cluster " + migrationSourceCluster + " and region "
+              + getRegionName());
+    }
+    StoreResponse srcStoreResponse = srcControllerClient.getStore(storeName);
+    if (srcStoreResponse.isError()) {
+      throw new VeniceException(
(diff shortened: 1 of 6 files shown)


services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -2268,6 +2268,55 @@ private Pair<Boolean, Version> addVersion(
...
+    }
+    StoreResponse srcStoreResponse = srcControllerClient.getStore(storeName);
+    if (srcStoreResponse.isError()) {
+      throw new VeniceException(
+          "Failed to get store " + storeName + " from cluster " + migrationSourceCluster + " and region "
+              + getRegionName() + " with error " + srcStoreResponse.getError());
+    }
+    StoreInfo srcStoreInfo = srcStoreResponse.getStore();
+    // For ongoing new pushes, destination cluster does not need to replicate the exact version configs from the source
+    // cluster; destination cluster can apply its own store configs to the new version.
```
FelixGV, line 2305: Is there any advantage to handling this case differently? Why not always copy from the source cluster?

FelixGV (commented):
Thanks Min, I'm sure you thought about it more than me and it's probably fine, but I have some questions. Maybe it's just a matter of clarifying the comments, or maybe there are issues... IDK. Please take a look.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -2352,199 +2401,226 @@ private Pair<Boolean, Version> addVersion(
...
+          // cluster;
+          // it's safest to replicate the version configs from the child controller in the same region, because parent
+          // region or other regions could have their own configs.
+          Optional<Version> sourceVersion = (store.isMigrating() && versionNumber != VERSION_ID_UNSET)
+              ? getVersionFromSourceCluster(repository, clusterName, storeName, versionNumber)
+              : Optional.empty();
+          if (sourceVersion.isPresent()) {
+            // Adding an existing version to the destination cluster whose version level resources are already created,
+            // including Kafka topics with data ready, and Helix resource, so skip the steps of recreating these
+            // resources.
```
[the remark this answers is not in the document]
huangminchn: You are right; Helix resource should be recreated in the destination cluster.

And the code does that indeed, I keep all the codes which check "startIngestion" shared in both store-migration and regular cases, which invokes "helixAdminClient.createVeniceStorageClusterResources" and creates Helix resource.

Will update the comment.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
huangminchn: The new version might not be there in the source cluster yet, so we may or may not get it from the source; the behavior is somewhat non-deterministic.

FelixGV (approved):

PR #571, opened 2023-08-07
[server] Remove excessive logging in leader producer callback when store is not A/A or partial update enabled

sixpluszero:
## [server] Remove excessive logging in leader producer callback when store is not A/A or partial update enabled
Resolves excessive logging found in internal certification flow. This newly added log line will be logged for store that does not have A/A or partial update enabled. Since it is not easy to check store property inside this LPC callback, this PR removes it.

## How was this PR tested?

Existing unit test.


gaojieliu (approved):
Thanks

gaojieliu (approved):

PR #572, opened 2023-08-07
[controller][test] Fix issues in data recovery API

lluwm:
This rb targets several issues in the data recovery API:

1\. Today, during data recovery, only when the version number from the source equals to the current version in the dest fabric, we bump a store's version to LargestUsedVersionNumber+1 in parent and use the increased version for data recovery. So, if the version number from the source is less than the current version in the dest, we simply use it in the dest. As a result, data recovery can not complete successfully, because backup version cleanup thread can kick in, consider it as an expired backup version, and delete it.

As a fix to it, we should extends current condition, and increase the store version for such cases as well.

On the other hand, if the version number from the source fabric is larger than current version in dest fabric, we can use it for the dest fabric without increasing it. This is fine because it is larger than the current, and resources for this version will be deleted during prepareDataRecovery. So, even though the version number is used (created and then deleted) in dest fabric before, it can still be reused.

2\. Users sometimes can put the same fabric name for src and dest which is considered as an error operation. This rb adds guardrail to protect such cases and fails early.

Integration tests are also improved to cover the above two issues to verify the correctness of this fix.

## How was this PR tested?
Improved integration tests.

### DIFF
--- a/internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/DataRecoveryTest.java
+++ b/internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/DataRecoveryTest.java
@@ -56,6 +56,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Properties;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.IntStream;
 import org.apache.samza.config.MapConfig;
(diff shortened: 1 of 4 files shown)


services/venice-controller/src/test/java/com/linkedin/venice/controller/datarecovery/TestDataRecoveryManager.java
```diff
@@ -7,7 +7,7 @@
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
-import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.*;
```
shuhui-liu, line 10: Nit: let's try to avoid the * import

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/DataRecoveryTest.java
```diff
@@ -234,27 +268,26 @@ public void testBatchOnlyDataRecovery() throws Exception {
...
-        }
-      }
+
+      // Perform multiple rounds of dc-0 --> dc-1 data recovery.
+      performDataRecoveryTest(storeName, parentControllerClient, dc1Client, "dc-0", "dc-1", 2, 2);
+
+      // Perform multiple rounds of dc-1 --> dc-0 data recovery. Now, since src colo's (dc-1) current version is 3, we
+      // should expect
+      // dc-0 starts with dc-1's current version = 3;
+      performDataRecoveryTest(storeName, parentControllerClient, dc0Client, "dc-1", "dc-0", 2, 3);
```
shuhui-liu, line 290: It took me some time to understand the test. Please check if my understanding is correct:

Before data recovery, current version in dc-0 and dc-0 is 1. With two rounds of dc-0 -> dc-1 data recovery, current version in dc-1 changes to 2 and then 3. Then with two rounds of dc-1 -> dc-0 data recovery, current version in dc-0 becomes 3 and then 4.

services/venice-controller/src/test/java/com/linkedin/venice/controller/datarecovery/TestDataRecoveryManager.java
lluwm: Sure, will fix it.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/DataRecoveryTest.java
lluwm: Yes, your understanding is exactly right. If you find it not easy to understand, I should add more comments to make it clearer. In fact your writeup is actually a good one I think.

shuhui-liu (approved):
Thanks!

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
```diff
@@ -3622,10 +3622,16 @@ private boolean whetherToCreateNewDataRecoveryVersion(
...
-    // Currently new version data recovery is only supported for batch-only store.
-    // For existing data centers, current store version might be serving read requests. Need to create a new version.
-    // For new data centers or non-current version, it's ok to delete and recreate it. No need to create a new version.
-    return destStore.getHybridStoreConfig() == null && versionNumber == destStore.getCurrentVersion()
+    /**
+     * Creating a new data recovery version on the destination colo when satisfying:
+     * 1. New version data recovery is only supported for batch-only store.
+     * 2. For the existing destination data center, a new version is needed if
+     *    2.1. srcVersionNumber equals to the current version in dest colo, as current version is serving read requests.
+     *    2.2. srcVersionNumber is less than the current version in dest colo, because Venice normally assumes that a
```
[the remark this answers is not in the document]
lluwm: Just to make sure that I understand the question accurately, let me give a concrete example for the ease of discussion, assuming src current ver = 19, dest current ver = 21 (but has data loss issue), and we want to recover the dest from data loss by using the src. What would happen after this change is that we create a new version in the dest, so dest cur version = 22. 

So, with this example, is the question about why not use ver = 21 in the dest for recovery (e.g. delete it and then recreate it)? To that, I think there was concern about deleting the current version in dest because it may still serve read requests.

PR #573, opened 2023-08-08
[test] Resolve potential port conflict issue in VeniceRouterWrapper

sushantmane:
## Resolve Port Conflict in VeniceRouterWrapper

To start a router on a specific port, passing `LISTENER_PORT` configuration via 
`extraProperties` could accidentally cause it to be utilized by the VeniceServer  
which could lead to port conflicts. To resolve this, introduced  a separate config  
for starting a router on a designated port. This fix ensures that the port  
assignment remains distinct and does not interfere with other components.

Additional improvements:
- Removed unused classes related to Zk client.
- Removed a log statement that was polluting the logs.


internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/integration/utils/VeniceRouterWrapper.java
```diff
@@ -112,9 +116,12 @@ static StatefulServiceProvider<VeniceRouterWrapper> generateService(
...
     return (serviceName, dataDirectory) -> {
-      String listenerPort =
-          properties.getOrDefault(ConfigKeys.LISTENER_PORT, Integer.toString(TestUtils.getFreePort())).toString();
-      int port = Integer.parseInt(listenerPort);
+      int port;
+      if (properties.containsKey(ROUTER_PORT_TO_USE_IN_VENICE_ROUTER_WRAPPER)) {
+        port = Integer.parseInt(properties.getProperty(ROUTER_PORT_TO_USE_IN_VENICE_ROUTER_WRAPPER));
+      } else {
+        port = TestUtils.getFreePort();
+      }
```
FelixGV, line 124: Why does it help to define a new constant, rather than using `LISTENER_PORT`?

FelixGV (commented):
Thanks for cleaning up the unused classes... I am wondering how does the fix help though. Can you take a look at my question below, or provide extra details if that is not the significant part?

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/integration/utils/VeniceRouterWrapper.java
sushantmane: `LISTENER_PORT` property is also used by VeniceServers. 

In VeniceServerWrapper we do explicitly add `LISTENER_PORT` property but this could be overridden by `LISTENER_PORT` from `configProperties`. `configProperties` is essentially `extraProperties` passed to all the components. 

So it's possible that both server and router will try to use this property

FelixGV: I see, thanks! Maybe we should add a comment for this?

sushantmane: Done!

FelixGV (approved):

sushantmane:
Thanks, @FelixGV!

PR #574, opened 2023-08-08
[server][fc] Enable mutual TLS on gRPC enabled Server & FC

rkunds:
## [server][fc] Enable mutual TLS on gRPC enabled Server & FC

- Introduce mutual TLS support on gRPC enabled servers and fast client, when an SSLFactory is present
- Update unit tests and integration tests to enable TLS

## How was this PR tested?

- CI/Unit Tests

If SSL is enabled for a gRPC enabled Venice Server, we start the gRPC server with SSL active and Client Auth required. If this is the case, we have to pass in an SSLFactory instance in the Client Configs when creating a gRPC enabled FC. If this isn't the case, then we would not be able to connect to a Venice Server instance.


rkunds:
messed up with git 🤪 see #565 for updated PR

PR #576, opened 2023-08-08
[producer] Allow online producer users to specify underlying VeniceWriter configs

nisargthakkar:
## Allow online producer users to specify underlying VeniceWriter configs

This commit allows the users of online producer to specify arbitrary configs for the underlying `VeniceWriter`

## How was this PR tested?

GH CI. Tested manually 

### DIFF
--- a/clients/venice-producer/src/main/java/com/linkedin/venice/producer/AbstractVeniceProducer.java
+++ b/clients/venice-producer/src/main/java/com/linkedin/venice/producer/AbstractVeniceProducer.java
@@ -3,7 +3,6 @@ package com.linkedin.venice.producer;
 import static com.linkedin.venice.ConfigKeys.CLIENT_PRODUCER_THREAD_NUM;
 import static com.linkedin.venice.ConfigKeys.KAFKA_BOOTSTRAP_SERVERS;
 import static com.linkedin.venice.ConfigKeys.KAFKA_OVER_SSL;
-import static com.linkedin.venice.ConfigKeys.KAFKA_SECURITY_PROTOCOL;
 import static com.linkedin.venice.ConfigKeys.SSL_KAFKA_BOOTSTRAP_SERVERS;
 import static com.linkedin.venice.writer.VeniceWriter.APP_DEFAULT_LOGICAL_TS;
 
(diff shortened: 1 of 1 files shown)


clients/venice-producer/src/main/java/com/linkedin/venice/producer/AbstractVeniceProducer.java
```diff
@@ -111,12 +110,11 @@ protected void configure(
...
 
   private VeniceWriter<byte[], byte[], byte[]> getVeniceWriter(VersionCreationResponse versionCreationResponse) {
-    Properties writerProps = new Properties();
+    Properties writerProps = producerConfigs.getPropertiesCopy();
 
     if (versionCreationResponse.isEnableSSL()) {
       writerProps.put(KAFKA_OVER_SSL, "true");
       writerProps.put(SSL_KAFKA_BOOTSTRAP_SERVERS, versionCreationResponse.getKafkaBootstrapServers());
-      writerProps.put(KAFKA_SECURITY_PROTOCOL, producerConfigs.getString(KAFKA_SECURITY_PROTOCOL));
       writerProps.putAll(new SSLConfig(producerConfigs).getKafkaSSLConfig());
```
xunyin8, line 118: I'm still a little confused to how the change is related to the PR title. If `writePros = producerConfigs.getPropertiesCopy();` then we can technically remove ` writerProps.putAll(new SSLConfig(producerConfigs).getKafkaSSLConfig());` too right?

xunyin8 (commented):
I'm still a little confused to how the change is related to the PR title. I see more of a passing along rather than override. Maybe you can point me in the right direction.

clients/venice-producer/src/main/java/com/linkedin/venice/producer/AbstractVeniceProducer.java
nisargthakkar: Yep. We can remove that too. Thanks for catching it.

The way it overrides it is via passing it through. Say we want to edit the default `delivery.ms`, we specify `kafka.delivery.ms` it in online `producerConfigs` object and it gets passed to the writer via `writerProps`. In that way, we can configure the pubsub system.

I can change the PR title and commit message to specify that we're only configuring the `VeniceWriter` instead of the pubsub producer.

xunyin8: I see, now it makes sense.

xunyin8 (approved):
Thanks Nisarg

PR #577, opened 2023-08-08
[controller][server] Reduce logging in controller, server

majisourav99:
## Reduce exception logging in push status collector in controller and SST file ingestion in Server

For stores without pushstatus system store, controller logs get flooded with full excception stacktrace. Also in server most of the time RMD file sst writer are empty, so do not print any warn log for it.


sixpluszero:
But today, every user store should have push status system store. If it is not, then it is not expected.

majisourav99:
> But today, every user store should have push status system store. If it is not, then it is not expected.

For oms stores it was not created and was flooding the controller logs. Also it will print the exception message instead of full stacktrace

sixpluszero:
> > But today, every user store should have push status system store. If it is not, then it is not expected.
> 
> For oms stores it was not created and was flooding the controller logs. Also it will print the exception message instead of full stacktrace

Ok, as long as e.getMessage() itself won't throw it should be fine..

sixpluszero (approved):

PR #578, opened 2023-08-08
[server][da-vinci] Added a configurable delay from OFFLINE to DROPPED transition

huangminchn:
## Summary
Previously in O/B/O mode, there is an explicit delay from OFFLINE to DROPPED transition; we missed this delay when building L/F mode. This PR added the delay back.

## How was this PR tested?
CI ongoing.

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/AbstractPartitionStateModel.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/AbstractPartitionStateModel.java
@@ -248,10 +248,8 @@ public abstract class AbstractPartitionStateModel extends StateModel {
     // Since this removes the storageEngine from the map not doing an un-subscribe and dropping a partition could
     // lead to NPE and other issues.
     // Adding a topic unsubscribe call for those race conditions as a safeguard before dropping the partition.
-    ingestionBackend.dropStoragePartitionGracefully(
-        storeConfig,
-        partition,
-        getStoreConfig().getPartitionGracefulDropDelaySeconds());
+    ingestionBackend
+        .dropStoragePartitionGracefully(storeConfig, partition, getStoreConfig().getStopConsumptionWaitRetriesNumber());
     removeCustomizedState();
   }
 
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/AbstractPartitionStateModel.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/AbstractPartitionStateModel.java
@@ -249,7 +249,7 @@ public abstract class AbstractPartitionStateModel extends StateModel {
     // lead to NPE and other issues.
     // Adding a topic unsubscribe call for those race conditions as a safeguard before dropping the partition.
     ingestionBackend
-        .dropStoragePartitionGracefully(storeConfig, partition, getStoreConfig().getStopConsumptionWaitRetriesNumber());
+        .dropStoragePartitionGracefully(storeConfig, partition, getStoreConfig().getStopConsumptionTimeoutInSeconds());
     removeCustomizedState();
   }
 
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/AbstractPartitionStateModel.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/AbstractPartitionStateModel.java
@@ -51,7 +51,7 @@ public abstract class AbstractPartitionStateModel extends StateModel {
 
   private final VeniceIngestionBackend ingestionBackend;
   private final ReadOnlyStoreRepository storeRepository;
-  private final VeniceStoreVersionConfig storeConfig;
+  private final VeniceStoreVersionConfig storeAndServerConfigs;
   private final int partition;
   private final String storePartitionDescription;
   private final CompletableFuture<HelixPartitionStatusAccessor> partitionStatusAccessorFuture;
@@ -62,16 +62,16 @@ public abstract class AbstractPartitionStateModel extends StateModel {
   public AbstractPartitionStateModel(
       VeniceIngestionBackend ingestionBackend,
       ReadOnlyStoreRepository storeRepository,
-      VeniceStoreVersionConfig storeConfig,
+      VeniceStoreVersionConfig storeAndServerConfigs,
       int partition,
       CompletableFuture<HelixPartitionStatusAccessor> accessorFuture,
       String instanceName) {
     this.ingestionBackend = ingestionBackend;
     this.storeRepository = storeRepository;
-    this.storeConfig = storeConfig;
+    this.storeAndServerConfigs = storeAndServerConfigs;
     this.partition = partition;
     this.storePartitionDescription =
-        String.format(STORE_PARTITION_DESCRIPTION_FORMAT, storeConfig.getStoreVersionName(), partition);
+        String.format(STORE_PARTITION_DESCRIPTION_FORMAT, storeAndServerConfigs.getStoreVersionName(), partition);
     /**
      * We cannot block here because helix manager connection depends on the state model constructing in helix logic.
      * If we block here in the constructor, it will cause deadlocks.
@@ -110,7 +110,7 @@ public abstract class AbstractPartitionStateModel extends StateModel {
         rollback ? "rolling back" : "initiating",
         from,
         to,
-        getStoreConfig().getStoreVersionName(),
+        getStoreAndServerConfigs().getStoreVersionName(),
         partition,
         message,
         context);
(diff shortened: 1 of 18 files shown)


FelixGV:
What is the purpose of the delay? And will this delay increase risk of resource leakage?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/LeaderFollowerPartitionStateModel.java
```diff
@@ -127,7 +129,15 @@ public void onBecomeOfflineFromStandby(Message message, NotificationContext cont
 
   @Transition(to = HelixState.DROPPED_STATE, from = HelixState.OFFLINE_STATE)
   public void onBecomeDroppedFromOffline(Message message, NotificationContext context) {
-    executeStateTransition(message, context, this::removePartitionFromStoreGracefully);
+    executeStateTransition(message, context, () -> {
+      try {
+        // Gracefully drop partition to drain the requests to this partition
+        Thread.sleep(TimeUnit.SECONDS.toMillis(getStoreConfig().getPartitionGracefulDropDelaySeconds()));
```
asaxena76, line 135: Might be clearer to move this inside removePartitionFromStoreGacefully and explain that this is to give time for the router to stop sending requests

gaojieliu: I think we should remove or change the waiting time in this function:
https://github.com/linkedin/venice/blob/main/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaStoreIngestionService.java#L873

The wait in this function is mainly to try to drain the ingestion task, which shouldn't need 10mins.

Putting the wait here is fine with enough comments and the function in `KafkaStoreIngestionService` shouldn't leverage the same config for waiting and it should be much shorter.

huangminchn:
> What is the purpose of the delay? And will this delay increase risk of resource leakage?

@FelixGV It is to achieve graceful partition drop. When a partition is dropped, there will be a delay on callback to router and thus router would keep sending requests to the node that no longer has the data. Adding a delay in OFFLINE to DROPPED transition can help cover this delay.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/LeaderFollowerPartitionStateModel.java
huangminchn: @gaojieliu That function doesn't wait for the entire 10 minutes actually; it's a loop with 10 iteration with 1 minute sleep time only if the draining is not completed; and we break the loop early if draining is completed. That function looks okay to me.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/AbstractStateModelFactory.java
```diff
@@ -53,6 +57,10 @@ public ExecutorService getExecutorService(String resourceName) {
     return executorService;
   }
 
+  public ParticipantStateTransitionStats getStateTransitionStats(String resourceName) {
```
gaojieliu, line 60: Why do you need `resourceName` param since it is not being used?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/LeaderFollowerPartitionStateModel.java
gaojieliu: If that is the case, I don't think we should use the same config: 
```
getStoreConfig().getPartitionGracefulDropDelaySeconds()
```
in function: `removePartitionFromStoreGracefully`.
Either we use some hard-coded number in `removePartitionFromStoreGracefully` or use a different config.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/AbstractStateModelFactory.java
gaojieliu: I realized that this is an overrided function, so can you put the right annotation here?

huangminchn: This is the root of this function actually; I can't add annotation here; but the annotation was added in `LeaderFollowerPartitionStateModelDualPoolFactory` which indeed override it.

I added a javaDoc in the latest diff to explain the input.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/LeaderFollowerPartitionStateModel.java
huangminchn: Updated in the latest diff; I reused an existing config called "server.stop.consumption.wait.retries.num"; Da-Vinci client does use this config "server.stop.consumption.wait.retries.num" for "removePartitionFromStoreGracefully"; and backend doesn't use it currently.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/AbstractStateModelFactory.java
gaojieliu: I see.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/LeaderFollowerPartitionStateModel.java
```diff
@@ -127,7 +133,18 @@ public void onBecomeOfflineFromStandby(Message message, NotificationContext cont
 
   @Transition(to = HelixState.DROPPED_STATE, from = HelixState.OFFLINE_STATE)
   public void onBecomeDroppedFromOffline(Message message, NotificationContext context) {
-    executeStateTransition(message, context, this::removePartitionFromStoreGracefully);
+    executeStateTransition(message, context, () -> {
+      try {
+        this.threadPoolStats.incrementThreadBlockedOnOfflineToDroppedTransitionCount();
+        // Gracefully drop partition to drain the requests to this partition
+        Thread.sleep(TimeUnit.SECONDS.toMillis(getStoreConfig().getPartitionGracefulDropDelaySeconds()));
```
gaojieliu, line 140: As discussed in the slack channel, this sleep is used to drain the read request.
So the sleep can be skipped if the resource is not for the current version, right?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/AbstractPartitionStateModel.java
```diff
@@ -248,10 +248,8 @@ protected void removePartitionFromStoreGracefully() {
     // Since this removes the storageEngine from the map not doing an un-subscribe and dropping a partition could
     // lead to NPE and other issues.
     // Adding a topic unsubscribe call for those race conditions as a safeguard before dropping the partition.
-    ingestionBackend.dropStoragePartitionGracefully(
-        storeConfig,
-        partition,
-        getStoreConfig().getPartitionGracefulDropDelaySeconds());
+    ingestionBackend
+        .dropStoragePartitionGracefully(storeConfig, partition, getStoreConfig().getStopConsumptionWaitRetriesNumber());
```
gaojieliu, line 252: This is a little confusing since the function param is actual timeout, not retry:
```
public void dropStoragePartitionGracefully(
      VeniceStoreVersionConfig storeConfig,
      int partition,
      int timeoutInSeconds,
      boolean removeEmptyStorageEngine) 
```

Shall we update the config name to be a timeout instead of retry?
If you think it is reasonable, we should update DaVinci as well and I assume no DaVinci customer is overriding that retry config, so it should be fine to rename the config.
`retry` config doesn't make sense if the `interval` is not configurable or visible.

gaojieliu (changes requested):

clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/LeaderFollowerPartitionStateModel.java
huangminchn: Right; I forgot about this. Will add the check

clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/AbstractPartitionStateModel.java
huangminchn: It makes sense. I did a code search, no DVC user configures it. If I change the config name to timeout, we can reuse the graceful drop timeout as its default now, what do you think? Or keep the default as 180 seconds?

gaojieliu: I think we should use a different default from the graceful drop config, otherwise, it might be confusing.

gaojieliu (approved):
Thanks for the fix!

asaxena76 (approved):
LGTM

PR #580, opened 2023-08-09
[dvc][server] WC classes move from client-common to DVC

FelixGV:
Many classes were located in venice-client-common, yet were only used within DVC. This commit moves these classes to DVC. Ideally, they should eventually end up in server, but since DVC itself is tangled with server code, that is difficult to refactor in one shot. At least for now venice-client-common will be less bloated, and we'll reduce the likelihood of future tangling to continue.

Miscellaneous:

- Renamed several ID to Id in variables and method names, at the request of Spotbugs.

- Removed the --stacktrace argument from the CI, as it makes it harder to scroll to the root cause of a build failure, and typically does not contain any useful information (but rather, only Gradle internal details).

- Also enabled MergeGenericRecordTest::testUpdate which was previously disabled. Had to fix it up a bit so it could run.

- Introduced a deserializer cache in VeniceRmdTTLFilter to get rid of some code which was instantiating MapOrderPreservingDeserializer instances on the hot path. Map order preservation is only necessary within the server when we will deserialize and then reserialize the payloads, but that was not the case in the TTL filter.

Module-level test branch coverage changes:

- venice-client-common: 37.76% to 44.33%
- da-vinci-client: 48.69% to 51.7%

## How was this PR tested?

CI


sixpluszero (dismissed):
Thanks for the changes!

clients/da-vinci-client/src/main/java/com/linkedin/davinci/schema/writecompute/WriteComputeHandlerV2.java
```diff
@@ -166,10 +169,20 @@ private GenericRecord assertAndGetTsRecordFieldIsRecord(GenericRecord timestampR
...
-              fieldTimestamp,
+              "Expect field '%s' in the timestamp record to be a generic record. Got timestamp record: %s",
+              fieldName,
               timestampRecord));
     }
     return (GenericRecord) fieldTimestamp;
   }
+
+  // protected access for testing
+  public Object updateArray(Schema arraySchema, List originalArray, Object writeComputeArray) {
```
nisargthakkar, line 180: Should this be `protected`?

internal/venice-client-common/src/main/java/com/linkedin/venice/schema/writecompute/WriteComputeHandlerV1.java
```diff
@@ -84,8 +84,8 @@ private Object updateFieldValue(Schema valueFieldSchema, Object originalFieldVal
     }
   }
 
-  // Visible for testing
-  Object updateArray(Schema arraySchema, List originalArray, Object writeComputeArray) {
+  // protected access for testing
+  public Object updateArray(Schema arraySchema, List originalArray, Object writeComputeArray) {
```
nisargthakkar, line 88: Same nit: `protected`?

internal/venice-client-common/src/main/java/com/linkedin/venice/utils/AvroSchemaUtils.java
```diff
@@ -1,6 +1,6 @@
 package com.linkedin.venice.utils;
 
-import static org.apache.avro.Schema.Type.RECORD;
+import static org.apache.avro.Schema.Type.*;
```
nisargthakkar, line 3: Avoid wildcard imports?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/schema/writecompute/WriteComputeHandlerV2.java
```diff
@@ -166,10 +169,20 @@ private GenericRecord assertAndGetTsRecordFieldIsRecord(GenericRecord timestampR
...
+    return super.updateArray(arraySchema, originalArray, writeComputeArray);
+  }
+
+  // protected access for testing
+  protected Object updateMap(Map originalMap, Object writeComputeMap) {
```
gaojieliu, line 185: Since these functions are calling `super` function directly without additional logic, so why override them at all?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/schema/writecompute/WriteComputeHandlerV2.java
sixpluszero: Yeah, this is used in client-common's unit test but I think that test class it moved to DVC package now, so it can be "protected" or even default (no modifier)

FelixGV: Thanks for catching this... I did this in an earlier iteration of the refactoring, but since then it's not needed anymore. I reverted this change.

internal/venice-client-common/src/main/java/com/linkedin/venice/utils/AvroSchemaUtils.java
FelixGV: Thanks for catching it. Done.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/schema/writecompute/WriteComputeHandlerV2.java
FelixGV: Thanks for catching this... I did this in an earlier iteration of the refactoring, but since then it's not needed anymore. I reverted this change.

FelixGV:
Thanks for the reviews!

PR #581, opened 2023-08-09
[da-vinci] Added one more config to simplify the DaVinci memory limiter test

gaojieliu:
## Summary
For some DaVinci customers, they want to validate DaVinci memory limiter feature without affecting prod in canary instance, and currently, it is not possible since DaVinci memory limiter will apply globally.
To simplify the test, this code change introduces a new config: 
```
ingestion.memory.limit.store.list : default empty
```
When it is specified, only the stores configured will enforce memory limiter, and if it is not specified (default), all the stores in the same DaVinci instance will enforce memory limiter.

With this way, the customer can specify a test store to test DaVinci memory limiter feature, and when the bootstrapping of test store fails because of memory exhaust, only the push job to the test store will fail.

## How was this PR tested?
CI

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBStorageEngine.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBStorageEngine.java
@@ -258,6 +258,10 @@ public class RocksDBStorageEngine extends AbstractStorageEngine<RocksDBStoragePa
   @Override
   public boolean hasMemorySpaceLeft() {
     SstFileManager sstFileManager = factory.getSstFileManagerForMemoryLimiter();
+    if (sstFileManager == null) {
+      // Memory limiter is disabled.
+      return true;
+    }
     if (sstFileManager.isMaxAllowedSpaceReached() || sstFileManager.isMaxAllowedSpaceReachedIncludingCompactions()) {
       return false;
     }
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBStorageEngine.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBStorageEngine.java
@@ -257,7 +257,7 @@ public class RocksDBStorageEngine extends AbstractStorageEngine<RocksDBStoragePa
 
   @Override
   public boolean hasMemorySpaceLeft() {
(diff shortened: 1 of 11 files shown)


clients/da-vinci-client/src/test/java/com/linkedin/davinci/store/rocksdb/RocksDBStorageEngineFactoryTest.java
```diff
@@ -20,7 +21,7 @@ public void testRocksDBCreation() {
 
     RocksDBStorageEngineFactory factory = new RocksDBStorageEngineFactory(serverConfig);
 
-    final String testStore = Utils.getUniqueString("test_store");
+    final String testStore = Version.composeKafkaTopic(Utils.getUniqueString("test_store"), 1);
```
majisourav99, line 24: So we have seen using wrong resource names in the tests! may be a util validator method should check these resource name and fail early.

majisourav99 (dismissed):
LGTM. Thanks Gaojie!

clients/da-vinci-client/src/test/java/com/linkedin/davinci/store/rocksdb/RocksDBStorageEngineFactoryTest.java
gaojieliu: These are unit tests and previously it can work with any database name, and with the new change, it has to be valid topic resource name.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBStorageEngine.java
```diff
@@ -257,7 +257,7 @@ private String getRocksDbEngineConfigPath() {
 
   @Override
   public boolean hasMemorySpaceLeft() {
-    SstFileManager sstFileManager = factory.getSstFileManager();
+    SstFileManager sstFileManager = factory.getSstFileManagerForMemoryLimiter();
     if (sstFileManager.isMaxAllowedSpaceReached() || sstFileManager.isMaxAllowedSpaceReachedIncludingCompactions()) {
```
majisourav99, line 261: should we add a null check here?

gaojieliu: Sure.

majisourav99 (approved):
Thanks Gaojie

PR #583, opened 2023-08-11
[server] Make store ingestion task idle counter configurable

sixpluszero:
This PR makes change to resolve race condition between:
(1) new partition ingestion
(2) store ingestion task idle timeout

This PR makes max idle counter configurable and set the default from 100 to 10000, so total timeout would be 1s -> 100s. This should make the race condition very very unlikely to happen. 
Also, to make the logic reliable, this PR makes a change that makes sure the retrieved SIT for startConsumption() is **active** SIT. It will try to reset the idle counter with synchronization guarantee.

## How was this PR tested?

New unit test

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaStoreIngestionService.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaStoreIngestionService.java
@@ -519,7 +519,9 @@ public class KafkaStoreIngestionService extends AbstractVeniceService implements
     return true;
   }
 
-  private StoreIngestionTask createConsumerTask(VeniceStoreVersionConfig veniceStoreVersionConfig, int partitionId) {
+  private StoreIngestionTask createStoreIngestionTask(
+      VeniceStoreVersionConfig veniceStoreVersionConfig,
+      int partitionId) {
     String storeName = Version.parseStoreFromKafkaTopicName(veniceStoreVersionConfig.getStoreVersionName());
     int versionNumber = Version.parseVersionFromKafkaTopicName(veniceStoreVersionConfig.getStoreVersionName());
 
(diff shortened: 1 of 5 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaStoreIngestionService.java
```diff
@@ -620,20 +622,23 @@ public void startConsumption(
...
 
     final String topic = veniceStore.getStoreVersionName();
+
     try (AutoCloseableLock ignore = topicLockManager.getLockForResource(topic)) {
-      StoreIngestionTask consumerTask = topicNameToIngestionTaskMap.get(topic);
-      if (consumerTask == null || !consumerTask.isRunning()) {
-        consumerTask = createConsumerTask(veniceStore, partitionId);
-        topicNameToIngestionTaskMap.put(topic, consumerTask);
-        versionedIngestionStats.setIngestionTask(topic, consumerTask);
+      StoreIngestionTask storeIngestionTask = topicNameToIngestionTaskMap.get(topic);
```
xunyin8, line 627: `topicNameToIngestionTaskMap` is a concurrent map and technically this kind of get() + null check is not thread safe. Is it possible to replace with `compute` or `computeIfAbsent`.

sixpluszero: In theory, yes, but it is also guarded by the external topic lock though, so in theory it is already ok..

sixpluszero: Although I can still do the update, which has no more harm

sixpluszero: Let me see if it is easy to change it. Will update the PR

sixpluszero: Fixed.

xunyin8 (approved):

PR #584, opened 2023-08-11
[server][controller] WC field handling optimizations

FelixGV:
Replaced several lookups by field name to use field position instead in write compute code paths. There are probably more such opportunities left. This is just one pass for cases where the safety of the usage is more obvious.

Also moved CollectionTimestampBuilder to tests since it was meant from the start to be a test-only class.

## How was this PR tested?

CI


sixpluszero (approved):
Thanks for the change! 
I think it is good, but just want to confirm: This optimization can only be used for in place update right? My concern is if it is used in a different record or different version of schema, position might not match, is that right?

FelixGV:
> Thanks for the change!
> I think it is good, but just want to confirm: This optimization can only be used for in place update right? My concern is if it is used in a different record or different version of schema, position might not match, is that right?

Thanks for the review!

The rule for using this technique correctly is this:

You can only a field's position if that field was extracted from the schema of the record you're using it on. So for example, in many places, we operate on two records: the value and the timestamp (RMD) record; if the field came out of the schema of the value record, then the field's position can be used to call get and set on the value, but when getting and setting on the RMD we still have to do it by name (or else extract a different field instance from the RMD schema).

PR #585, opened 2023-08-11
[server][admin-tool] Improve Server Ingestion State Dump

gaojieliu:
This code change adds a new command to dump ingestion state, also Venice Server will skip the ACL check for admin request endpoint since there is no PII data.
In the meantime, this code change will skip compression dict in the response since it is not readable and string-format the `Fixed` type.

## How was this PR tested?

CI

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/listener/response/AdminResponse.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/listener/response/AdminResponse.java
@@ -28,6 +28,7 @@ import org.apache.avro.io.Encoder;
  * This class stores all the information required for answering a server admin request.
  */
 public class AdminResponse {
+  public static final ByteBuffer IGNORED_COMPRESSION_DICT = ByteBuffer.wrap("ignored".getBytes());
   private boolean isError;
   private String message;
   private final AdminResponseRecord responseRecord;
@@ -60,6 +61,22 @@ public class AdminResponse {
     responseRecord.partitionConsumptionStates.add(snapshot);
   }
 
+  static StoreVersionState suppressCompressionDict(StoreVersionState storeVersionState) {
+    StoreVersionState updatedState = storeVersionState;
+    if (storeVersionState.compressionDictionary != null && storeVersionState.compressionDictionary.hasRemaining()) {
+      // We don't want to dump compression dictionary since it is not readable
+      updatedState = new StoreVersionState();
+      for (Schema.Field field: StoreVersionState.SCHEMA$.getFields()) {
+        if (field.name().equals("compressionDictionary")) {
+          updatedState.put(field.pos(), IGNORED_COMPRESSION_DICT);
+        } else {
+          updatedState.put(field.pos(), storeVersionState.get(field.pos()));
+        }
+      }
+    }
+    return updatedState;
+  }
+
   /**
    * Add store version state metadata into the admin response record
    */
@@ -67,19 +84,7 @@ public class AdminResponse {
     try (ByteArrayOutputStream output = new ByteArrayOutputStream()) {
       GenericDatumWriter<Object> avroDatumWriter = new GenericDatumWriter<>(StoreVersionState.SCHEMA$);
       Encoder byteToHexJsonEncoder = new ByteBufferToHexFormatJsonEncoder(StoreVersionState.SCHEMA$, output);
-      StoreVersionState updatedState = storeVersionState;
-      if (storeVersionState.compressionDictionary != null && storeVersionState.compressionDictionary.hasRemaining()) {
-        // We don't want to dump compression dictionary since it is not readable
-        updatedState = new StoreVersionState();
-        for (Schema.Field field: StoreVersionState.SCHEMA$.getFields()) {
-          if (field.name().equals("compressionDictionary")) {
-            updatedState.put(field.pos(), ByteBuffer.wrap("ignored".getBytes()));
-          } else {
-            updatedState.put(field.pos(), storeVersionState.get(field.pos()));
-          }
-        }
-      }
-      avroDatumWriter.write(updatedState, byteToHexJsonEncoder);
+      avroDatumWriter.write(suppressCompressionDict(storeVersionState), byteToHexJsonEncoder);
       byteToHexJsonEncoder.flush();
       output.flush();
       responseRecord.storeVersionState = new String(output.toByteArray());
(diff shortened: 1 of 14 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/listener/response/AdminResponse.java
```diff
@@ -65,7 +67,19 @@ public void addStoreVersionState(StoreVersionState storeVersionState) {
     try (ByteArrayOutputStream output = new ByteArrayOutputStream()) {
       GenericDatumWriter<Object> avroDatumWriter = new GenericDatumWriter<>(StoreVersionState.SCHEMA$);
       Encoder byteToHexJsonEncoder = new ByteBufferToHexFormatJsonEncoder(StoreVersionState.SCHEMA$, output);
-      avroDatumWriter.write(storeVersionState, byteToHexJsonEncoder);
+      StoreVersionState updatedState = storeVersionState;
+      if (storeVersionState.compressionDictionary != null && storeVersionState.compressionDictionary.hasRemaining()) {
+        // We don't want to dump compression dictionary since it is not readable
+        updatedState = new StoreVersionState();
+        for (Schema.Field field: StoreVersionState.SCHEMA$.getFields()) {
```
FelixGV, line 74: Nitpick: IMHO it looks cleaner to use `getClassSchema()` than `SCHEMA$`.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/listener/response/AdminResponse.java
```diff
@@ -65,7 +67,19 @@ public void addStoreVersionState(StoreVersionState storeVersionState) {
     try (ByteArrayOutputStream output = new ByteArrayOutputStream()) {
       GenericDatumWriter<Object> avroDatumWriter = new GenericDatumWriter<>(StoreVersionState.SCHEMA$);
       Encoder byteToHexJsonEncoder = new ByteBufferToHexFormatJsonEncoder(StoreVersionState.SCHEMA$, output);
-      avroDatumWriter.write(storeVersionState, byteToHexJsonEncoder);
+      StoreVersionState updatedState = storeVersionState;
```
FelixGV, line 70: What is the intent of this? Creating a new reference to the same object on the heap? Modifying `updatedState` will also modify `storeVersionState`, is it intentional? If you do not wish to modify the input param, then we would need to do a deep copy. Later on in the code you are seemingly doing a deep copy, but it ends up being a copy onto itself. Maybe this line should be `StoreVersionState updatedState = new StoreVersionState();` instead? Or alternatively, something like this:

```
StoreVersionState updatedState = GenericData.get().deepCopy(
    storeVersionState.getSchema(), 
    storeVersionState);
updatedState.set("compressionDictionary", ByteBuffer.wrap("ignored".getBytes()));
```

And I think you would not need to loop manually on the fields at all.

FelixGV: Oh, sorry, I missed that you are already doing `updatedState = new StoreVersionState();` inside the if below. So then the code is fine, although you might still want to use the built-in deep copier function, if you'd like...

FelixGV (commented):
Code change LGTM, but there unit test module is failing... let's take a look.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/listener/response/AdminResponse.java
gaojieliu: yeah, the code is doing a shallow copy and only update the `compressionDictionary` field, which should be more lightweight than a deep copy without modifying the original param.

internal/venice-common/src/main/java/com/linkedin/venice/serialization/avro/AvroProtocolDefinition.java
```diff
@@ -154,9 +154,9 @@ public enum AvroProtocolDefinition {
   PARTICIPANT_MESSAGE_SYSTEM_STORE_VALUE(1, ParticipantMessageValue.class),
 
   /**
-   * Response record for admin request v1.
+   * Response record for admin request.
    */
-  SERVER_ADMIN_RESPONSE_V1(1, AdminResponseRecord.class),
+  SERVER_ADMIN_RESPONSE(2, AdminResponseRecord.class),
```
huangminchn, line 159: I don't get this; where is version 2?

internal/venice-client-common/src/main/java/org/apache/avro/io/ByteBufferToHexFormatJsonEncoder.java
```diff
@@ -38,6 +39,19 @@ public void writeBytes(byte[] bytes, int start, int len) throws IOException {
     writeString(ByteUtils.toHexString(bytes, start, len));
   }
 
+  @Override
+  public void writeFixed(byte[] bytes, int start, int len) throws IOException {
```
huangminchn, line 43: Just curious, what schema fields are fixed type?

clients/venice-admin-tool/src/main/java/com/linkedin/venice/AdminTool.java
```diff
@@ -528,6 +532,9 @@ public static void main(String args[]) throws Exception {
         case REQUEST_BASED_METADATA:
           getRequestBasedMetadata(cmd);
           break;
+        case DUMP_INGESTION_STATE:
```
huangminchn, line 535: Thanks for adding the command :D

huangminchn (commented):
Thanks a lot for improving this command, Gaojie!

internal/venice-common/src/main/java/com/linkedin/venice/serialization/avro/AvroProtocolDefinition.java
gaojieliu: https://github.com/linkedin/venice/blob/main/internal/venice-common/src/main/resources/avro/AdminResponseRecord/v2/AdminResponseRecord.avsc

internal/venice-client-common/src/main/java/org/apache/avro/io/ByteBufferToHexFormatJsonEncoder.java
gaojieliu: `PartitionState` contains GUID.
https://github.com/linkedin/venice/blob/main/internal/venice-common/src/main/java/com/linkedin/venice/offsets/OffsetRecord.java#L303

huangminchn (approved):
LGTM! Thanks Gaojie!

clients/da-vinci-client/src/main/java/com/linkedin/davinci/listener/response/AdminResponse.java
gaojieliu: Fixed.

huangminchn (dismissed):

huangminchn (approved):

PR #589, opened 2023-08-14
[controller] Catch HelixException in deleteHelixResource

majisourav99:
##  Catch HelixException in deleteHelixResource

We see some host contained in  `/CONFIG/PARTICIPANT` path but not in `/INSTANCE` of a cluster .  If queried a list of instances using `ZKHelixAdmin::getInstancesInCluster` and called `getDisabledPartitionsMap` which throws HelixException. This PR tries to skip such exception to continue cleaning the helix resources. Also bumps the Helix dependency to 1.1.0 which would resolve the exception log in zk node deletion code path.

## How was this PR tested?

GHCI

### DIFF
--- a/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
@@ -237,6 +237,7 @@ import org.apache.commons.lang.StringUtils;
 import org.apache.commons.lang.Validate;
 import org.apache.helix.AccessOption;
 import org.apache.helix.HelixAdmin;
+import org.apache.helix.HelixException;
 import org.apache.helix.HelixManagerProperty;
 import org.apache.helix.HelixPropertyFactory;
 import org.apache.helix.InstanceType;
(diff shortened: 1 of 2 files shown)


services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -4759,9 +4760,18 @@ public void deleteHelixResource(String clusterName, String kafkaTopic) {
...
 
     List<String> instances = getStorageNodes(clusterName);
+    Map<String, List<String>> disabledPartitions;
     for (String instance: instances) {
-      Map<String, List<String>> disabledPartitions =
-          getHelixAdminClient().getDisabledPartitionsMap(clusterName, instance);
+      try {
+        disabledPartitions = getHelixAdminClient().getDisabledPartitionsMap(clusterName, instance);
+      } catch (HelixException helixException) {
+        LOGGER.warn(
```
nisargthakkar, line 4768: Won't this flood the logs? Can we rate-limit this?

majisourav99: Its should not be that many,  it can happen only during deleting old version if some instance configs are not in sync.

majisourav99: Yes for each new push it will print once for each out of sync configs.

nisargthakkar (approved):
LGTM

PR #591, opened 2023-08-15
[server] Code change for server connection count metrics

xunyin8:
## Add Venice server connection and misrouted metrics
Two gauges (and its max) to show server's live connections count from routers and servers.
- router_connection_count (Gauge and OccurrenceRate)
- client_connection_count (Gauge and OccurrenceRate)
These metrics are collected by a new `ChannelInboundHandlerAdapter`, the `ServerConnectionStatsHandler` that will collect stats accordingly on `channelRegistered` and `channelUnregistered.

Additional per store and total metric for misrouted requests based on invalid store version.
- misrouted_store_version_request_count
The check only occurs on error requests, the corresponding flag is set on the `HttpShortcutResponse` if it's determined that the request had a misrouted store version. `StatsHandler` will emit the corresponding metric on its. `ServerStatsContext.errorRequest`

## How was this PR tested?
TODO will have unit and integration test to verify the metrics
-->

### DIFF
--- a/services/venice-server/src/main/java/com/linkedin/venice/listener/ServerConnectionStatsHandler.java
+++ b/services/venice-server/src/main/java/com/linkedin/venice/listener/ServerConnectionStatsHandler.java
@@ -1,0 +1,61 @@
+package com.linkedin.venice.listener;
+
+import com.linkedin.venice.stats.ServerConnectionStats;
+import com.linkedin.venice.utils.SslUtils;
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.channel.ChannelInboundHandlerAdapter;
+import io.netty.handler.ssl.SslHandler;
+import java.security.cert.X509Certificate;
+import javax.net.ssl.SSLPeerUnverifiedException;
+
+
+public class ServerConnectionStatsHandler extends ChannelInboundHandlerAdapter {
+  private final ServerConnectionStats serverConnectionStats;
+  private final String routerPrincialName;
+
+  public ServerConnectionStatsHandler(ServerConnectionStats serverConnectionStats, String routerPrincipalName) {
+    this.serverConnectionStats = serverConnectionStats;
+    this.routerPrincialName = routerPrincipalName;
+  }
+
+  @Override
+  public void channelRegistered(ChannelHandlerContext ctx) throws Exception {
+    SslHandler sslHandler = extractSslHandler(ctx);
+    if (sslHandler == null) {
+      // No ssl enabled, record all connections as client connections
+      serverConnectionStats.incrementClientConnectionCount();
+      return;
+    }
+    String principalName = getPrincipal(sslHandler);
+    if (principalName.contains(routerPrincialName)) {
+      serverConnectionStats.incrementRouterConnectionCount();
+    } else {
+      serverConnectionStats.incrementClientConnectionCount();
+    }
+  }
+
+  @Override
+  public void channelUnregistered(ChannelHandlerContext ctx) throws Exception {
+    SslHandler sslHandler = extractSslHandler(ctx);
+    if (sslHandler == null) {
+      // No ssl enabled, record all connections as client connections
+      serverConnectionStats.decrementClientConnectionCount();
+      return;
+    }
+    String principalName = getPrincipal(sslHandler);
+    if (principalName.contains(routerPrincialName)) {
+      serverConnectionStats.incrementRouterConnectionCount();
+    } else {
+      serverConnectionStats.incrementClientConnectionCount();
+    }
+  }
+
+  protected SslHandler extractSslHandler(ChannelHandlerContext ctx) {
+    return ServerHandlerUtils.extractSslHandler(ctx);
+  }
+
+  private String getPrincipal(SslHandler sslHandler) throws SSLPeerUnverifiedException {
+    X509Certificate clientCert = SslUtils.getX509Certificate(sslHandler.engine().getSession().getPeerCertificates()[0]);
+    return clientCert.getSubjectX500Principal().getName();
+  }
+}
--- a/services/venice-server/src/main/java/com/linkedin/venice/listener/ServerConnectionStatsHandler.java
+++ b/services/venice-server/src/main/java/com/linkedin/venice/listener/ServerConnectionStatsHandler.java
@@ -11,11 +11,11 @@ import javax.net.ssl.SSLPeerUnverifiedException;
 
 public class ServerConnectionStatsHandler extends ChannelInboundHandlerAdapter {
   private final ServerConnectionStats serverConnectionStats;
-  private final String routerPrincialName;
+  private final String routerPrincipalName;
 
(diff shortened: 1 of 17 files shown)


services/venice-server/src/main/java/com/linkedin/venice/listener/ServerConnectionStatsHandler.java
```diff
@@ -0,0 +1,61 @@
...
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.channel.ChannelInboundHandlerAdapter;
+import io.netty.handler.ssl.SslHandler;
+import java.security.cert.X509Certificate;
+import javax.net.ssl.SSLPeerUnverifiedException;
+
+
+public class ServerConnectionStatsHandler extends ChannelInboundHandlerAdapter {
+  private final ServerConnectionStats serverConnectionStats;
+  private final String routerPrincialName;
```
majisourav99, line 14: typo :)

services/venice-server/src/main/java/com/linkedin/venice/listener/ServerConnectionStatsHandler.java
```diff
@@ -0,0 +1,61 @@
...
+  @Override
+  public void channelUnregistered(ChannelHandlerContext ctx) throws Exception {
+    SslHandler sslHandler = extractSslHandler(ctx);
+    if (sslHandler == null) {
+      // No ssl enabled, record all connections as client connections
+      serverConnectionStats.decrementClientConnectionCount();
+      return;
+    }
+    String principalName = getPrincipal(sslHandler);
+    if (principalName.contains(routerPrincialName)) {
```
majisourav99, line 46: shouldnt these call decrement count ?

xunyin8: Ops, copy paste errors will fix.

services/venice-server/src/main/java/com/linkedin/venice/listener/StorageReadRequestHandler.java
```diff
@@ -349,6 +354,21 @@ public void channelRead(ChannelHandlerContext context, Object message) throws Ex
     }
   }
 
+  /**
+   * Best effort check for the purpose of reporting misrouted store version metric when the request errors.
+   */
+  private boolean checkMisroutedStoreVersionRequest(RouterRequest request) {
```
gaojieliu, line 360: Do we need to check partition?

services/venice-server/src/main/java/com/linkedin/venice/stats/ServerConnectionStats.java
```diff
@@ -0,0 +1,43 @@
+package com.linkedin.venice.stats;
+
+import io.tehuti.metrics.MetricsRepository;
+import io.tehuti.metrics.Sensor;
+import io.tehuti.metrics.stats.Max;
+
+
+public class ServerConnectionStats extends AbstractVeniceStats {
+  public static final String ROUTER_CONNECTION_COUNT_GAUGE = "router_connection_count_gauge";
```
gaojieliu, line 9: We don't need to put `gauge` into the metric name since `Gauge` type will be added to the metric name automatically.

services/venice-server/src/main/java/com/linkedin/venice/listener/StatsHandler.java
```diff
@@ -194,10 +199,10 @@ public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise)
        * multiple times for a single request
        */
       if (!serverStatsContext.isStatCallBackExecuted()) {
-        ServerHttpRequestStats serverHttpRequestStats =
-            serverStatsContext.getCurrentStats().getStoreStats(serverStatsContext.getStoreName());
+        ServerHttpRequestStats serverHttpRequestStats = serverStatsContext.getStoreName() == null
```
gaojieliu, line 202: Can `serverStatsContext.getStoreName()` be null?

gaojieliu (commented):

services/venice-server/src/main/java/com/linkedin/venice/stats/ServerConnectionStats.java
```diff
@@ -0,0 +1,43 @@
...
+
+  private final Sensor routerConnectionCountSensor;
+  private final Sensor clientConnectionCountSensor;
+
+  private int routerConnectionCount;
+  private int clientConnectionCount;
+
+  public ServerConnectionStats(MetricsRepository metricsRepository, String name) {
+    super(metricsRepository, name);
+    routerConnectionCountSensor = registerSensorIfAbsent(ROUTER_CONNECTION_COUNT_GAUGE, new Gauge(), new Max());
```
gaojieliu, line 20: Also I wonder whether the default constructor for `Gauge` works at all.
Normally, we will offer a `Measurable` callback when constructing a `Gauge` metric and for this use case, it will be:
```
new Gauge(() -> routerConnectionCount)
```

Also `Max` is a window-based stat, and if there is no update of the sensor in the past 1 min, the `max` value will be `0`, which will be confusing.

If we would like to use `Max` to capture the conn storm issue, we can potentially use `OccurrenceRate` stat to achieve the same goal.

services/venice-server/src/main/java/com/linkedin/venice/listener/StorageReadRequestHandler.java
xunyin8: Felix and i talked about this in a slack thread and agreed that initially let’s just have a metric that detects misrouted requests due to wrong version. Since checking partition will require a lot more work

services/venice-server/src/main/java/com/linkedin/venice/listener/StatsHandler.java
xunyin8: I think so since prior to the code change we do the null check within serverStatsContext. I think it could be null if the request somehow fail super early before the storeName is set?

services/venice-server/src/main/java/com/linkedin/venice/stats/ServerConnectionStats.java
xunyin8: Right, let me fix that.

services/venice-server/src/main/java/com/linkedin/venice/stats/ServerConnectionStats.java
xunyin8: Oh, maybe that's why the gauge is returning 0 in the integration test. Initially I thought it's due to connections closing already. Let me fix this.

xunyin8: Update: the gauge is indeed working and it's returning 0 in the integration test because the channel is actually unregistered. Verified this by commenting out the decrement methods.

gaojieliu (approved):
Thanks for the quick fix!

PR #592, opened 2023-08-15
[vpj][controller][router][server][test] create dummy zstd dictionary for empty push and suppress redundant dictionary fetch logs

m-nagarajan:
## Summary
1. Dictionary not created:
- When there is an empty push, controller creates a dictionary using synthetic data and sends it via `SOP`. But when `send.control.messages.directly` is enabled, VPJ driver sends the `SOP` and not the controller, so no dictionary gets created. Moving this flow from controller to the VPJ driver code such that the dummy dictionary gets created for both the flows. 
- corrected some misleading logs in VPJ driver about dictionary creation for empty push scenario
- Added exceptions in server's `processStartOfPush` to check for dictionary if `ZSTD_WITH_DICT` is configured.

2. Too many logs/retries to fetch dictionary in router
- Added redundant log filter to the router logs for fetching dictionary. And a test to satisfy diff coverage.
- exponential backoff for router retries (min 100ms, max 5s)
- Cancelling the scheduled dictionary fetch retries when removing the version/store: potentially preventing further retires or exceptions. 

## How was this PR tested?
GH CI


services/venice-router/src/main/java/com/linkedin/venice/router/api/DictionaryRetrievalService.java
```diff
@@ -137,6 +139,27 @@ public void handleStoreChanged(Store store) {
     }
   };
 
+  enum LogLevel {
+    INFO, WARN
+  };
+
+  private final RedundantExceptionFilter redundantExceptionFilter = new RedundantExceptionFilter(
+      RedundantExceptionFilter.DEFAULT_BITSET_SIZE,
```
gaojieliu, line 147: Shall we create a new constructor for `RedundantExceptionFilter`, which will take these default params instead of the users of this class to do that?

services/venice-router/src/main/java/com/linkedin/venice/router/api/DictionaryRetrievalService.java
```diff
@@ -137,6 +139,27 @@ public void handleStoreChanged(Store store) {
     }
   };
 
+  enum LogLevel {
```
gaojieliu, line 142: Shall we use `Level` from log4j instead?

gaojieliu (commented):
Looks good, left some minor comments.

services/venice-router/src/main/java/com/linkedin/venice/router/api/DictionaryRetrievalService.java
m-nagarajan: Modified to use log4j.

services/venice-router/src/main/java/com/linkedin/venice/router/api/DictionaryRetrievalService.java
m-nagarajan: Cleaner suggestion. Done.

services/venice-router/src/main/java/com/linkedin/venice/router/api/DictionaryRetrievalService.java
```diff
@@ -424,5 +453,18 @@ public void stopInner() throws IOException {
...
+    redundantExceptionFilter.shutdown();
+  }
+
+  // Visible for testing
+  public StoreDataChangedListener getStoreChangeListener() {
+    return storeChangeListener;
+  }
+
+  // Visible for testing
+  public RedundantExceptionFilter getRedundantExceptionFilter() {
```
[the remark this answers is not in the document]
m-nagarajan: Yes, done.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
```diff
@@ -1869,49 +1880,62 @@ private Optional<ByteBuffer> getCompressionDictionary() throws VeniceException {
...
 
+          /**
+           * ZSTD_WITH_DICT needs a dictionary even if there is no input data, so we generate a basic one based \
+           * on synthetic data.
+           *
+           * TODO: It would be smarter to query it from the previous version and pass it along. However,
+           *  the 'previous' version can mean different things in different colos, and ideally we'd want
+           *  a consistent compressed result in all colos so as to make sure we don't confuse our consistency
+           *  checking mechanisms. So this needs some (maybe) complicated reworking
+           */
```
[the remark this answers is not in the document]
m-nagarajan: Thanks, corrected these. I moved these comments from controller code, but we did change the behavior after the original comment.

m-nagarajan (commented):
Thanks @gaojieliu for the review. Fixed the review comments and also added
1. exponential backoff for router retries.
2. Cancelling the scheduled dictionary fetch retries when removing the version/store: potentially preventing further retires or exceptions. 
3. validating presence of dict in SIT during SOP processing

gaojieliu (approved):
Thanks for the fix!

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -2585,6 +2569,15 @@ private Pair<Boolean, Version> addVersion(
             }
 
             if (sendStartOfPush) {
+              ByteBuffer compressionDictionaryBuffer = null;
+              if (compressionDictionary != null) {
+                compressionDictionaryBuffer =
+                    ByteBuffer.wrap(EncodingUtils.base64DecodeFromString(compressionDictionary));
+              } else if (store.getCompressionStrategy().equals(CompressionStrategy.ZSTD_WITH_DICT)) {
+                throw new VeniceException(
+                    "compression Dictionary should not be empty if CompressionStrategy is ZSTD_WITH_DICT");
```
nisargthakkar, line 2578: This will break for empty-push when it doesn't go via the KIF mode - which is still new. Can we keep the controller sending an dummy dictionary logic? Are there any issues?

gaojieliu: yeah, this code change will require some collaboration between VPJ and Controller because of this.
Maybe we can make it backward compatible in Controller  to check whether the request param contains dict or not and if it doesn't for empty push, the Controller will continue to use this dummy dict, otherwise throw exception.
Once VPJ is fully deployed, we can cleanup this logic from Controller.

m-nagarajan: As sending via controller is introduced for samza batch push, I think this flow is still needed even if we add this logic in vpj as well. So added it back with proper comments.

gaojieliu (approved):
Thanks for the fix!

m-nagarajan:
Thanks @gaojieliu and @nisargthakkar for the review.

PR #593, opened 2023-08-16
[controller] Add rollForwardFuture Version api to controller

ZacAttack:
## Add rollForwardFuture Version api to controller
The venice parent controller doesn't allow setting a version to current.  This makes sense as version numbers don't necessarily align across datacenters. We have a need to be able to mark a future version to be a serving version after necesary checks have been performed by it (usually by the customer in the context of reprocessing version push). We could augment the setCurrentVersion api in parent, but that seems dicey in terms of being clear about the intention of the operator.  So instead we add a new api to make it very clear. In Venice we will keep at most a current Version, a backup version or a future version.  We today have a rollback api, so we mirror that to do the opposite and rollforward.  This way the numbers don't need to match up for the versions, we just always take the next version in each child fabric and promote it to current.

## How was this PR tested?
Added integration test case, need to probably improve code coverage


FelixGV:
Hmm... I may be missing something but... isn't the future version guaranteed to be the same across datacenters? I understand how the current version can be different, if a push errored out in one DC but not another... but the future, if it exists, should be the same, no?

ZacAttack:
> Hmm... I may be missing something but... isn't the future version guaranteed to be the same across datacenters? I understand how the current version can be different, if a push errored out in one DC but not another... but the future, if it exists, should be the same, no?

Huh...  That actually could be if we're reserving the version numbers across all colos?  Is that the case?

FelixGV:
> > Hmm... I may be missing something but... isn't the future version guaranteed to be the same across datacenters? I understand how the current version can be different, if a push errored out in one DC but not another... but the future, if it exists, should be the same, no?
> 
> Huh... That actually could be if we're reserving the version numbers across all colos? Is that the case?

AFAIK it is the parent controller that decides the version number, and child controllers honor that (even if it means skipping over a number and creating a gap in the sequence, from their point of view).

ZacAttack:
> > > Hmm... I may be missing something but... isn't the future version guaranteed to be the same across datacenters? I understand how the current version can be different, if a push errored out in one DC but not another... but the future, if it exists, should be the same, no?
> > 
> > 
> > Huh... That actually could be if we're reserving the version numbers across all colos? Is that the case?
> 
> AFAIK it is the parent controller that decides the version number, and child controllers honor that (even if it means skipping over a number and creating a gap in the sequence, from their point of view).

Ok.  That makes sense.  Does that change the implementation we should go for you think?  Like, armed with this information, does it make sense to pad this into setCurrentVersion in parent?  The interface today takes a version argument which triggered my hesitation.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
```diff
@@ -1865,6 +1865,61 @@ public void setStoreCurrentVersion(String clusterName, String storeName, int ver
...
+      controllerClientMap.forEach((region, cc) -> tasks.add(() -> {
+        StoreResponse storeResponse = cc.getStore(storeName, waitingTimeForConsumptionMs);
+        if (storeResponse.isError()) {
+          throw new VeniceException(storeResponse.getError() + " in region " + region);
+        }
+        StoreInfo store = storeResponse.getStore();
+        if (!store.isEnableStoreWrites()) {
+          throw new VeniceException("Unable to rollback since store does not enable write in region " + region);
+        }
+        int futureVersion = getVeniceHelixAdmin().getFutureVersion(clusterName, storeName);
```
FelixGV, line 1885: Is it the right understanding that this line is the only one that's different between this new function and the older `rollbackToBackupVersion` function? Can we extract all of this into a common function, perhaps one which takes a `Function<Admin, Integer>` parameter?

We should update the various comments which talk about backups since this code would be used for both rolling forward and back.

And one more thing... can you double-check that the implementation of `Admin::getFutureVersion` is correct? In the parent, it returns `Store.NON_EXISTING_VERSION`, which does not seem right in the code path here... whereas in the child, it returns:

```
monitor.getTopicsOfOngoingOfflinePushes()
        .stream()
        .filter(topic -> Version.parseStoreFromKafkaTopicName(topic).equals(storeName))
        .findFirst();
```

Which I'm not sure is right either? Does this rely on there being only one future push monitored? Is that guaranteed? In normal operations that should be the case, certainly... but is it not possible that there could be some lingering push that failed or is due to fail but hasn't been cleaned up yet?

Currently, the `Admin::getFutureVersion` API seems to be called only from `StoresRoute::getFutureVersion`, which AFAICT is a read-only API, and so we may or may not have noticed whether it was wonky or not... In any case, we should double-check, since with this PR, it will now be used in a more significant manner as it could mutate the state of the system.

services/venice-controller/src/main/java/com/linkedin/venice/controller/server/AdminSparkServer.java
```diff
@@ -1,104 +1,7 @@
...
-import static com.linkedin.venice.controllerapi.ControllerRoute.UPDATE_ADMIN_TOPIC_METADATA;
-import static com.linkedin.venice.controllerapi.ControllerRoute.UPDATE_CLUSTER_CONFIG;
-import static com.linkedin.venice.controllerapi.ControllerRoute.UPDATE_KAFKA_TOPIC_LOG_COMPACTION;
-import static com.linkedin.venice.controllerapi.ControllerRoute.UPDATE_KAFKA_TOPIC_MIN_IN_SYNC_REPLICA;
-import static com.linkedin.venice.controllerapi.ControllerRoute.UPDATE_KAFKA_TOPIC_RETENTION;
-import static com.linkedin.venice.controllerapi.ControllerRoute.UPDATE_STORAGE_PERSONA;
-import static com.linkedin.venice.controllerapi.ControllerRoute.UPDATE_STORE;
-import static com.linkedin.venice.controllerapi.ControllerRoute.UPLOAD_PUSH_JOB_STATUS;
-import static com.linkedin.venice.controllerapi.ControllerRoute.WIPE_CLUSTER;
+import static com.linkedin.venice.controllerapi.ControllerRoute.*;
```
FelixGV, line 4: Let's undo this...

services/venice-controller/src/main/java/com/linkedin/venice/controller/server/StoresRoutes.java
```diff
@@ -23,33 +23,7 @@
...
-import static com.linkedin.venice.controllerapi.ControllerRoute.MIGRATE_STORE;
-import static com.linkedin.venice.controllerapi.ControllerRoute.REMOVE_STORE_FROM_GRAVEYARD;
-import static com.linkedin.venice.controllerapi.ControllerRoute.ROLLBACK_TO_BACKUP_VERSION;
-import static com.linkedin.venice.controllerapi.ControllerRoute.SET_OWNER;
-import static com.linkedin.venice.controllerapi.ControllerRoute.SET_TOPIC_COMPACTION;
-import static com.linkedin.venice.controllerapi.ControllerRoute.SET_VERSION;
-import static com.linkedin.venice.controllerapi.ControllerRoute.STORAGE_ENGINE_OVERHEAD_RATIO;
-import static com.linkedin.venice.controllerapi.ControllerRoute.STORE;
-import static com.linkedin.venice.controllerapi.ControllerRoute.UPDATE_STORE;
+import static com.linkedin.venice.controllerapi.ControllerRoute.*;
```
FelixGV, line 26: Ditto

FelixGV (commented):
Some comments on the actual code. It seems there the unit test task is failing, so we should look into that. It might be good to add some unit test specifically for this new functionality...

FelixGV:
> > > > Hmm... I may be missing something but... isn't the future version guaranteed to be the same across datacenters? I understand how the current version can be different, if a push errored out in one DC but not another... but the future, if it exists, should be the same, no?
> > > 
> > > 
> > > Huh... That actually could be if we're reserving the version numbers across all colos? Is that the case?
> > 
> > 
> > AFAIK it is the parent controller that decides the version number, and child controllers honor that (even if it means skipping over a number and creating a gap in the sequence, from their point of view).
> 
> Ok. That makes sense. Does that change the implementation we should go for you think? Like, armed with this information, does it make sense to pad this into setCurrentVersion in parent? The interface today takes a version argument which triggered my hesitation.

I guess it doesn't change much, we can still have a roll forward API. I left some more comments on the code itself.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
ZacAttack: Yeah, it is a little wonky.  I have backed out from using it here and now just default to using the most 'future' version in the store list (the biggest numbered one).

Also condensed the code down by using a functional interface so now no more repeated logic.

lluwm (approved):
Thanks Zac and LGTM!

PR #594, opened 2023-08-17
[vpj] Support user-friendly input data format for partial update + incremental push

sixpluszero:
## [vpj] Support user-friendly input data format for partial update + incremental push
This PR adds a new input format for partial update when running incremental push.
For field update, user will only need to provide the subset dataset they need to update and VPJ will convert each record into partial update record.

For example: 
Store value schema: {F1: String, F2: Int, F3: Array<Long>}
User can create a dataset with:
[ {key: k1, value: {f1: "foo", f3: [1, 2]}}, {key: k2, value: {f1: "bar", f3: [3,4]}}]

And after the run, the f1, f3 field of key k1, k2 will be updated and other fields will remain untouched.

## How was this PR tested?
New Integration Test added
Added new unit test for Avro record reader logic

### DIFF
--- a/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VeniceAvroRecordReader.java
+++ b/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VeniceAvroRecordReader.java
@@ -23,7 +23,6 @@ import com.linkedin.venice.writer.update.UpdateBuilder;
 import com.linkedin.venice.writer.update.UpdateBuilderImpl;
 import java.io.IOException;
 import java.io.InputStream;
-import java.util.Arrays;
 import java.util.Collections;
 import java.util.Iterator;
 import java.util.LinkedList;
@@ -96,8 +95,6 @@ public class VeniceAvroRecordReader extends AbstractVeniceRecordReader<AvroWrapp
 
   public VeniceAvroRecordReader(VeniceProperties props) {
     super(props.getString(TOPIC_PROP));
-    LOGGER.info("DEBUGGING PROPS: {}", props.toString(true));
-    LOGGER.info("DEBUGGING STACKTRACE: {}", Arrays.toString(Thread.currentThread().getStackTrace()));
     this.fileSchema = AvroSchemaParseUtils.parseSchemaFromJSON(
         props.getString(SCHEMA_STRING_PROP),
         props.getBoolean(EXTENDED_SCHEMA_VALIDITY_CHECK_ENABLED, DEFAULT_EXTENDED_SCHEMA_VALIDITY_CHECK_ENABLED));
@@ -109,11 +106,7 @@ public class VeniceAvroRecordReader extends AbstractVeniceRecordReader<AvroWrapp
       this.updateSchema =
           AvroSchemaParseUtils.parseSchemaFromJSONLooseValidation(props.getString(UPDATE_SCHEMA_STRING_PROP));
     }
-    if (updateSchema != null) {
-      LOGGER.info("DEBUGGING: UPDATE SCHEMA" + updateSchema.toString(true));
-    }
     setupSchema(props.getString(KEY_FIELD_PROP), props.getString(VALUE_FIELD_PROP));
-    LOGGER.info("DEBUGGING PROPS: " + props.toString(true));
   }
 
   @Override
--- a/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VeniceAvroRecordReader.java
+++ b/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VeniceAvroRecordReader.java
@@ -30,7 +30,6 @@ import java.util.List;
 import org.apache.avro.Schema;
 import org.apache.avro.file.DataFileStream;
 import org.apache.avro.generic.GenericDatumReader;
-import org.apache.avro.generic.GenericRecord;
 import org.apache.avro.generic.IndexedRecord;
 import org.apache.avro.mapred.AvroWrapper;
 import org.apache.hadoop.fs.FileSystem;
@@ -128,13 +127,13 @@ public class VeniceAvroRecordReader extends AbstractVeniceRecordReader<AvroWrapp
     if (!generatePartialUpdateRecordFromInput) {
       return valueObject;
     }
-    if (!(valueObject instanceof GenericRecord)) {
-      throw new VeniceException("Retrieved record is not a Avro record");
+    if (!(valueObject instanceof IndexedRecord)) {
+      throw new VeniceException("Retrieved record is not a Avro indexed record");
     }
     UpdateBuilder updateBuilder = new UpdateBuilderImpl(updateSchema);
-    GenericRecord genericRecordValue = (GenericRecord) valueObject;
-    for (Schema.Field field: genericRecordValue.getSchema().getFields()) {
-      updateBuilder.setNewFieldValue(field.name(), genericRecordValue.get(field.pos()));
+    IndexedRecord indexedRecordValue = (IndexedRecord) valueObject;
+    for (Schema.Field field: indexedRecordValue.getSchema().getFields()) {
+      updateBuilder.setNewFieldValue(field.name(), indexedRecordValue.get(field.pos()));
     }
     return updateBuilder.build();
   }
--- a/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VeniceAvroRecordReader.java
+++ b/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VeniceAvroRecordReader.java
@@ -139,14 +139,11 @@ public class VeniceAvroRecordReader extends AbstractVeniceRecordReader<AvroWrapp
       throw new VeniceException("Retrieved record is not a Avro record");
     }
     UpdateBuilder updateBuilder = new UpdateBuilderImpl(updateSchema);
-    // TODO: Is it going to be GenericRecord?
     GenericRecord genericRecordValue = (GenericRecord) valueObject;
     for (Schema.Field field: genericRecordValue.getSchema().getFields()) {
       updateBuilder.setNewFieldValue(field.name(), genericRecordValue.get(field.pos()));
     }
-    GenericRecord result = updateBuilder.build();
-    LOGGER.info("DEBUGGING GET AVRO UPDATE: {}", result);
-    return result;
+    return updateBuilder.build();
   }
 
   public Schema getFileSchema() {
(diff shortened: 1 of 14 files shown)


clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
```diff
@@ -145,13 +146,15 @@ public class VenicePushJob implements AutoCloseable {
...
   public static final boolean DEFAULT_EXTENDED_SCHEMA_VALIDITY_CHECK_ENABLED = true;
+  public static final String UPDATE_SCHEMA_STRING_PROP = "update.schema";
 
   // Vson input configs
   // Vson files store key/value schema on file header. key / value fields are optional
   // and should be specified only when key / value schema is the partial of the files.
   public static final String FILE_KEY_SCHEMA = "key.schema";
   public static final String FILE_VALUE_SCHEMA = "value.schema";
   public static final String INCREMENTAL_PUSH = "incremental.push";
+  public static final String PARTIAL_UPDATE_WITH_NEW_INPUT_FORMAT = "partial.update.with.new.input.format";
```
gaojieliu, line 157: The naming is a little confusing since `new.input.format` is too generic and it might be clear now, but it won't be a `new` format after a few years' evolution..

Can we name it to be `generate.partial.update.schema.from.input` as the one you used for variable name in `VeniceAvroRecordReader`?

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
```diff
@@ -2307,12 +2311,26 @@ void validateValueSchema(
...
-          setting.controllerRetries,
-          c -> c.getValueOrDerivedSchemaId(setting.storeName, pushJobSchemaInfo.getValueSchemaString()));
+      if (!isUpdateSchema(pushJobSchemaInfo.getValueSchemaString())) {
+        // With new input format, we will need to use the latest update schema to generate partial update record.
+        getValueSchemaIdResponse = ControllerClient.retryableRequest(
+            controllerClient,
+            setting.controllerRetries,
+            c -> c.getLatestUpdateSchema(setting.storeName));
+        // Update the schema to be the update schema of the superset schema.
+        LOGGER.info(
```
gaojieliu, line 2322: Shall we validate the input schema align with the latest update schema?
For example, we need to make sure the following:
1. The fields present in the value schema must exist in the latest update schema.
2. The field schema in the value schema must exactly match the corresponding field in the latest update schema.
If the above conditions don't meet, the push job should fail, otherwise the partial update record passing to the backend can be wrong.

Also, let us try to put these validation in some other class, such as `AvroSchemaUtils` to reduce the size of `VenicePushJob`.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
```diff
@@ -2364,6 +2381,10 @@ void validateValueSchema(
         setting.storeName);
   }
 
+  boolean isUpdateSchema(String schemaString) {
```
gaojieliu, line 2384: This check seems to be too simple.
Shall we send this update schema to the backend to check whether it matches any existing update schema or not?

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
```diff
@@ -2307,12 +2311,26 @@ void validateValueSchema(
         .info("Validating value schema: {} for store: {}", pushJobSchemaInfo.getValueSchemaString(), setting.storeName);
 
     SchemaResponse getValueSchemaIdResponse;
-
     if (setting.enableWriteCompute) {
```
gaojieliu, line 2314: So the user has to specify `venice.write.compute.enable` to enable write compute in inc push job, it is fine for now, and we need to let the new users be aware of this config.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VeniceAvroRecordReader.java
```diff
@@ -82,36 +88,78 @@ public VeniceAvroRecordReader(
...
+  @Override
+  protected Object getAvroValue(AvroWrapper<IndexedRecord> record, NullWritable nullValue) {
+    Object valueObject = record.datum().get(valueFieldPos);
+    if (!generatePartialUpdateRecordFromInput) {
+      return valueObject;
+    }
+    if (!(valueObject instanceof IndexedRecord)) {
+      throw new VeniceException("Retrieved record is not a Avro indexed record");
+    }
+    UpdateBuilder updateBuilder = new UpdateBuilderImpl(updateSchema);
```
gaojieliu, line 133: We can evaluate reusing `UpdateBuilder` in the future since every instance creation will do the repetitive work, such as schema validation, retrieving fast serde class (which is proved to be slow because of schema comparison).

gaojieliu (changes requested):
Let us also highlight that the Controller needs to be deployed before the push job before leveraging this new feature.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
sixpluszero: Yeah, it is the current behavior. I don't plan to change it inside this PR as it could be very big change but we could migrate this out in a separate PR.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
sixpluszero: I'd like to double check about this: For now I added getLatestUpdateSchema endpoint which retrieves partial update schema of the superset schema. We don't have the endpoint to get superset schema yet and I don't plan to add it as this seems a bit duplicated in functionality.

Since update schema contains set field op for each field, I plan to iterate the field in the input (subset) value schema and see if each field type is in update schema and name matches. For union field, we will check if each sub type matches. I think that should work and can be added in to the util class you mentioned, but let me know if you think otherwise.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
sixpluszero: Ok, I can also add it, this is existing logic and we did not check it.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
gaojieliu: I think we should fetch the latest superset schema and the corresponding update schema, and the validation should happen against the value schema instead of update schema since iterating the fields in update schema will result in pulling wc schema generation related logic to Venice Push Job, which is bad.
```
getAllValueAndDerivedSchema
```
can return all the value schemas, right? I assume it contains the latest superset schema and the corresponding update schema.
Can you leverage this endpoint instead?

sixpluszero: It is not inside VPJ logic though, but I do agree, validating value field in update schema is a bit weird, even though I can do it in util class.
This endpoint alone does not tell me which schema version is the superset schema. An alternative is use this endpoint and get_store_in_cluster endpoint (and get the superset schema id from store object), so this can get the value schema and corresponding update schema without adding the new get latest update schema I added in this PR.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
sixpluszero: Actually, after reading the code again I think we did validate the schema if the input schema is already a partial update schema. Existing logic try to locate the schema id from controller client call.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
sixpluszero: Updated the code and remove the new Controller API

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VeniceAvroRecordReader.java
sixpluszero: Sure, we can offer that, this should not be too difficult to achieve in my opinion.

gaojieliu (commented):
Looks good overall, and left one minor comment.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
```diff
@@ -2307,12 +2313,56 @@ void validateValueSchema(
...
+        StoreResponse storeResponse = ControllerClient
+            .retryableRequest(controllerClient, setting.controllerRetries, c -> c.getStore(setting.storeName));
+
+        if (storeResponse.isError()) {
+          throw new VeniceException("Store does not exist: " + setting.storeName);
+        }
+        if (multiSchemaResponse.isError()) {
+          throw new VeniceException("Unable to retrieve schemas for store: " + setting.storeName);
+        }
+        int supersetSchemaId = storeResponse.getStore().getLatestSuperSetValueSchemaId();
```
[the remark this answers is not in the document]
sixpluszero: I know this, but seems like it is not set in current API code path... I am kind of hesitated about this -> If I changed the implementation that's fine, but this change will be depending on controller deployment (that's ......kind of OK but becomes slower to rollout) What do you think?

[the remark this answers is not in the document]
gaojieliu: I see..
Let us add some comments to mention this behavior since the current logic is a little confusing.
It is good to update this field in the existing endpoint in the future, and let us cut a ticket to track this.
Once this endpoint is ready, VPJ can migrate to this single endpoint.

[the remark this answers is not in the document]
sixpluszero: Sure, let me add some doc and create a ticket.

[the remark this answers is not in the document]
sixpluszero: Updated the java doc and add a ticket in WC epic as it is not very suitable for BB queue.

gaojieliu (approved):
You need to fix the unit test coverage failure.

sixpluszero:
> You need to fix the unit test coverage failure.

Thanks! That's weird, all I changed is a comment... let me retry it.

PR #597, opened 2023-08-20
[fc][test]make streamingBatchGet return partial response

m-nagarajan:
## Summary
`streamingBatchGet(keys)` was returning `AppTimeOutTrackingCompletableFuture`, so couldn't return partial results. Modified it to return `VeniceResponseCompletableFuture` and added some unit tests.

## How was this PR tested?
GH CI


clients/venice-client/src/test/java/com/linkedin/venice/fastclient/DispatchingAvroGenericStoreClientTest.java
```diff
@@ -174,7 +188,9 @@ private void setUpClient(
         batchGetValueFuture0.complete(batchGetResponse0);
         doReturn(batchGetValueFuture1).when(mockedTransportClient)
             .post(eq(REPLICA2_NAME + "/storage/test_store_v1"), any(), any());
-        batchGetValueFuture1.completeExceptionally(new VeniceClientException("Exception for client to return 503"));
+        if (!transportClientPartialIncomplete) {
```
xunyin8, line 191: Does this mean if `transportClientPartialIncomplete` is set to true then future1 will never complete and we should still be able to read the result of future0?

clients/venice-client/src/test/java/com/linkedin/venice/fastclient/DispatchingAvroGenericStoreClientTest.java
```diff
@@ -494,10 +510,175 @@ public void testBatchGetWithExceptionFromTransportLayerForOneRoute(boolean useSt
     }
   }
 
+  /**
+   * Condition to test: batchGet API either returns full results or exception, but no partial results.
```
xunyin8, line 514: Very helpful comments!

clients/venice-client/src/test/java/com/linkedin/venice/fastclient/DispatchingAvroGenericStoreClientTest.java
```diff
@@ -494,10 +510,175 @@ public void testBatchGetWithExceptionFromTransportLayerForOneRoute(boolean useSt
...
+   * In this test:
+   * setup: 1 key returns valid value and the other key doesn't return anything.
+   * Behavior: routingLeakedRequestCleanupThresholdMS times out, so returns exception with
+   *           "At least one route did not complete".
+   */
+  @Test(timeOut = TEST_TIMEOUT, expectedExceptions = ExecutionException.class, expectedExceptionsMessageRegExp = ".*VeniceClientException: At least one route did not complete")
+  public void testBatchGetWithTimeoutV1() throws IOException, ExecutionException, InterruptedException {
+    long routingLeakedRequestCleanupThresholdMS = TimeUnit.SECONDS.toMillis(3);
```
xunyin8, line 522: Since this is an unit test ideally we should be able to control the time and not having to actually wait. That's probably too much effort for this PR but at least we can make the timeout much smaller. 3 seconds is too long and will add unnecessary wait time.

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/StatsAvroGenericStoreClient.java
```diff
@@ -126,8 +126,9 @@ protected CompletableFuture<VeniceResponseMap<K, V>> streamingBatchGet(
       BatchGetRequestContext<K, V> requestContext,
       Set<K> keys) {
     long startTimeInNS = System.nanoTime();
-    CompletableFuture<VeniceResponseMap<K, V>> innerFuture = super.streamingBatchGet(requestContext, keys);
-    return recordMetrics(requestContext, keys.size(), innerFuture, startTimeInNS, clientStatsForBatchGet);
+    CompletableFuture<VeniceResponseMap<K, V>> streamingBatchGetFuture = super.streamingBatchGet(requestContext, keys);
+    recordMetrics(requestContext, keys.size(), streamingBatchGetFuture, startTimeInNS, clientStatsForBatchGet);
```
xunyin8, line 130: Not an issue, just a question. If we pass the `VeniceResponseCompletableFuture` as the innerFuture for `recordMetrics`. Does it mean in the case of a partial response scenario, metrics are reported when the future times out?

xunyin8 (commented):
Thanks Manoj, the change looks good just some questions.

clients/venice-client/src/test/java/com/linkedin/venice/fastclient/DispatchingAvroGenericStoreClientTest.java
m-nagarajan: Makes sense. Will reduce the timeouts.

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/StatsAvroGenericStoreClient.java
m-nagarajan: Yes actually.
If `streamingBatchGetFuture.get(userTimeout)` times out before all routes results are back but if all results are back within `routingLeakedRequestCleanupThresholdMS` which is more than `userTimeout`, metrics will reflect full success.

For `batchGet(keys)`, even though there is no partial response, the metrics will reflect that there was partial response.

Both are inconsistent with what the end user sees. Will have to check how to control this.

xunyin8 (approved):
Thanks!

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/StatsAvroGenericStoreClient.java
m-nagarajan: As discussed offline, for now, even if future.get(timeout) timeouts, metrics will get incremented when all routers reply back (with proper response or exception) or when `routingLeakedRequestCleanupThresholdMS` times out.

clients/venice-client/src/test/java/com/linkedin/venice/fastclient/DispatchingAvroGenericStoreClientTest.java
m-nagarajan: Fixed this.

PR #598, opened 2023-08-21
[server][da-vinci-client] Restore storage engine if not found

majisourav99:
##  Restore storage engine if not found while executing delete storage engine

When bootstrapping server, if II is not enabled, by default all data partitions are restored. So even if there is offline->standby msg sent, storage engine will be created during restore, and if Helix think this resource should be dropped, it will send out a offline->dropped. This should be fine, as it can locate the partition in storage engine and drop it. 
For II, during server bootstrap, we do not do any partition restore at all. When there is any offline > standby message coming in, we will restore it first in II process, ingest it until COMPLETED and then handover to main process. If Helix does not even send offline>standby but send offline->dropped directly, our logic will try to execute it in main process. Since storage engine is not opened, drop will be ignored which leads to version leaks on servers. This PR will try to directly delete DB files if storage engine is not found.

## How was this PR tested?

GHCI

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/IsolatedIngestionBackend.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/IsolatedIngestionBackend.java
@@ -347,17 +347,24 @@ public class IsolatedIngestionBackend extends DefaultIngestionBackend
       String topicName,
       int partition,
       IngestionCommandType command,
-      Supplier<Boolean> remoteCommandSupplier,
-      Runnable localCommandRunnable) {
+      Supplier<Boolean> isolatedProcessCommandSupplier,
+      Runnable mainProcessCommandRunnable) {
+    boolean isTopicPartitionHosted = isTopicPartitionHosted(topicName, partition);
+
+    // Start consumption for new partition happens in isolated process
+    if (command == START_CONSUMPTION && !isTopicPartitionHosted) {
+      isolatedProcessCommandSupplier.get();
+      return;
+    } else if (command == REMOVE_PARTITION && !isTopicPartitionHosted) {
+      mainProcessCommandRunnable.run();
+      isolatedProcessCommandSupplier.get();
+      return;
+    }
+
     do {
-      if (isTopicPartitionHostedInMainProcess(topicName, partition)
-          || (!isTopicPartitionHosted(topicName, partition) && command != START_CONSUMPTION)) {
-        LOGGER.info(
-            "Executing command {} of topic: {}, partition: {} in main process process.",
-            command,
-            topicName,
-            partition);
-        localCommandRunnable.run();
+      if (isTopicPartitionHostedInMainProcess(topicName, partition) || !isTopicPartitionHosted) {
+        LOGGER.info("Executing command {} of topic: {}, partition: {} in main process.", command, topicName, partition);
+        mainProcessCommandRunnable.run();
         return;
       }
       LOGGER.info("Sending command {} of topic: {}, partition: {} to fork process.", command, topicName, partition);
@@ -376,7 +383,7 @@ public class IsolatedIngestionBackend extends DefaultIngestionBackend
       try (AutoCloseableLock ignored =
           AutoCloseableSingleLock.of(getMainIngestionMonitorService().getForkProcessActionLock().readLock())) {
         try {
-          if (remoteCommandSupplier.get()) {
+          if (isolatedProcessCommandSupplier.get()) {
             return;
           }
         } catch (Exception e) {
@@ -403,7 +410,7 @@ public class IsolatedIngestionBackend extends DefaultIngestionBackend
             topicName,
             partition,
             command);
-        localCommandRunnable.run();
+        mainProcessCommandRunnable.run();
         return;
       }
       /**
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/IsolatedIngestionBackend.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/IsolatedIngestionBackend.java
@@ -351,18 +351,15 @@ public class IsolatedIngestionBackend extends DefaultIngestionBackend
       Runnable mainProcessCommandRunnable) {
     boolean isTopicPartitionHosted = isTopicPartitionHosted(topicName, partition);
 
-    // Start consumption for new partition happens in isolated process
-    if (command == START_CONSUMPTION && !isTopicPartitionHosted) {
-      isolatedProcessCommandSupplier.get();
-      return;
-    } else if (command == REMOVE_PARTITION && !isTopicPartitionHosted) {
+    if (command == REMOVE_PARTITION && !isTopicPartitionHosted) {
       mainProcessCommandRunnable.run();
       isolatedProcessCommandSupplier.get();
       return;
     }
 
     do {
-      if (isTopicPartitionHostedInMainProcess(topicName, partition) || !isTopicPartitionHosted) {
+      if (isTopicPartitionHostedInMainProcess(topicName, partition)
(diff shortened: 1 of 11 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/storage/StorageService.java
```diff
@@ -397,8 +397,12 @@ public synchronized void closeStorePartition(VeniceStoreVersionConfig storeConfi
   public synchronized void removeStorageEngine(String kafkaTopic) {
     AbstractStorageEngine<?> storageEngine = getStorageEngineRepository().removeLocalStorageEngine(kafkaTopic);
     if (storageEngine == null) {
-      LOGGER.warn("Storage engine {} does not exist, ignoring remove request.", kafkaTopic);
-      return;
+      LOGGER.warn("Storage engine {} does not exist, trying to reopen storage for .", kafkaTopic);
+      storageEngine = openStorageEngine(kafkaTopic);
```
m-nagarajan, line 401: This also calls `addLocalStorageEngine` which needs to be removed similar to line 398.

m-nagarajan (changes requested):
Thanks for the fix.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/storage/StorageService.java
```diff
@@ -409,6 +414,20 @@ public synchronized void removeStorageEngine(String kafkaTopic) {
     factory.removeStorageEngine(storageEngine);
   }
 
+  AbstractStorageEngine openStorageEngine(String storeName) {
+    VeniceStoreVersionConfig storeConfig = configLoader.getStoreConfig(storeName);
+    storeConfig.setRestoreDataPartitions(false);
```
sixpluszero, line 419: If we want to drop all data partition in II case, then we need to setRestoreDataPartition to true. 

Also, looks like the test does not cover this.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/storage/StorageService.java
```diff
@@ -409,6 +414,20 @@ public synchronized void removeStorageEngine(String kafkaTopic) {
     factory.removeStorageEngine(storageEngine);
   }
 
+  AbstractStorageEngine openStorageEngine(String storeName) {
```
sixpluszero, line 417: And I am struggling to think if it is better to restore everything inside II process. It looks like a better option in general (for non-DVC case). And we can tweak the logic to claim if a resource is not found in II's metadata management flow, it should by default sent to forked process. (It cannot be in the handover situation). 
For II's metadata management, feel free to check IsolatedIngesitonBackend and let me know if you think otherwise.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/isolated/IsolatedIngestionServer.java
```diff
@@ -722,7 +722,7 @@ private void initializeIsolatedIngestionServer() {
         storeVersionStateSerializer,
         partitionStateSerializer,
         storeRepository,
-        false,
+        true,
```
sixpluszero, line 725: Can you double check if DVC logic is compatible with this?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/storage/StorageService.java
```diff
@@ -397,7 +397,7 @@ public synchronized void closeStorePartition(VeniceStoreVersionConfig storeConfi
   public synchronized void removeStorageEngine(String kafkaTopic) {
     AbstractStorageEngine<?> storageEngine = getStorageEngineRepository().removeLocalStorageEngine(kafkaTopic);
     if (storageEngine == null) {
-      LOGGER.warn("Storage engine {} does not exist, ignoring remove request.", kafkaTopic);
```
sixpluszero, line 400: Let'e revert the comment

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBServerConfig.java
```diff
@@ -564,6 +564,10 @@ public void setBlockBaseFormatVersion(int version) {
     this.blockBaseFormatVersion = version;
   }
 
+  public void setRocksdbOptionsCompressionType(String compressionType) {
```
sixpluszero, line 567: Is it part of the change?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/IsolatedIngestionBackend.java
```diff
@@ -347,17 +347,21 @@ void executeCommandWithRetry(
...
       IngestionCommandType command,
-      Supplier<Boolean> remoteCommandSupplier,
-      Runnable localCommandRunnable) {
+      Supplier<Boolean> isolatedProcessCommandSupplier,
+      Runnable mainProcessCommandRunnable) {
+    boolean isTopicPartitionHosted = isTopicPartitionHosted(topicName, partition);
+
+    if (command == REMOVE_PARTITION && !isTopicPartitionHosted) {
+      mainProcessCommandRunnable.run();
+      isolatedProcessCommandSupplier.get();
```
sixpluszero, line 356: Could you double check if in forked process side, if the process will accept drop in this situation?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBServerConfig.java
majisourav99: no, but I needed it some to benchmark, I believe its good to have this setter

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBStorageEngineFactory.java
```diff
@@ -332,23 +333,36 @@ public synchronized void removeStorageEngine(AbstractStorageEngine engine) {
...
     if (storageEngineMap.containsKey(storeName)) {
       throw new VeniceException(
           "Storage engine has already been opened previously, and please use #removeStorageEngine(AbstractStorageEngine) for deletion");
     }
-    File storeDir = new File(rocksDBPath, storeName);
-    if (storeDir.exists()) {
-      LOGGER.info("Started removing RocksDB database folder for store: {}", storeName);
+    File dbDir;
 
+    if (partitionId == -1) {
```
sixpluszero, line 351: In which occasion this partition will be -1?

sixpluszero: Oh I see. it is the new impl being used in removeStorageEngine..

sixpluszero (approved):
Thanks for the fix!

clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/IsolatedIngestionBackend.java
```diff
@@ -347,17 +347,21 @@ void executeCommandWithRetry(
...
     do {
       if (isTopicPartitionHostedInMainProcess(topicName, partition)
-          || (!isTopicPartitionHosted(topicName, partition) && command != START_CONSUMPTION)) {
-        LOGGER.info(
-            "Executing command {} of topic: {}, partition: {} in main process process.",
-            command,
-            topicName,
-            partition);
-        localCommandRunnable.run();
+          || !isTopicPartitionHosted && command != START_CONSUMPTION) {
```
sixpluszero, line 362: nit: I'd suggest having a bracket here || ( condition && condition). Even though we know && has higher priority, this is somewhat not straightforward to people and might lead to some confusion.

PR #600, opened 2023-08-22
[controller][server][test] Support cluster by cluster pub sub clients for controller.

haoxu07:

Previously, different venice cluster for controller share the same pub sub admin, consumer and producer clients, especially for `TopicManager`. As venice is going to support different pub sub clients, we need to add support for different pub sub clients for different cluster running in the same controller host. 

In current controller's pub sub client usage in common class (e.g `TopicManager`) used by controller, there is no cluster name information in API. To avoid changing the interface, a cluster discovery supplier function is added to `TopicManagerRepository` and other utility class. For controller, this cluster discovery logic will rely on `HelixReadOnlyStoreConfigRepository` to resolve cluster name from topic. For server, it just uses the default cluster name for that host. 

It is possible that `HelixReadOnlyStoreConfigRepository` based logic cannot resolve cluster name from topic and some pub sub client usage is not for specific cluster (e.g. listing all topics inside `TopicManager`), we will rely on default pub sub clients factory to do this. 

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
@@ -186,7 +186,11 @@ public class DaVinciBackend implements Closeable {
           new VeniceWriterFactory(backendProps.toProperties(), pubSubClientsFactory.getProducerAdapterFactory(), null);
       String instanceName = Utils.getHostName() + "_" + Utils.getPid();
       int derivedSchemaID = backendProps.getInt(PUSH_STATUS_STORE_DERIVED_SCHEMA_ID, 1);
-      pushStatusStoreWriter = new PushStatusStoreWriter(writerFactory, instanceName, derivedSchemaID);
+      Map<String, VeniceWriterFactory> writerFactoryMap = new HashMap<>(1);
+      String clusterName = backendConfig.getClusterName();
+      writerFactoryMap.put(clusterName, writerFactory);
+      pushStatusStoreWriter =
+          new PushStatusStoreWriter(writerFactoryMap, instanceName, derivedSchemaID, s -> clusterName);
 
       SchemaReader kafkaMessageEnvelopeSchemaReader = ClientFactory.getSchemaReader(
           ClientConfig.cloneConfig(clientConfig)
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
@@ -190,7 +190,7 @@ public class DaVinciBackend implements Closeable {
       String clusterName = backendConfig.getClusterName();
       writerFactoryMap.put(clusterName, writerFactory);
       pushStatusStoreWriter =
-          new PushStatusStoreWriter(writerFactoryMap, instanceName, derivedSchemaID, s -> clusterName);
+          new PushStatusStoreWriter(writerFactoryMap, instanceName, derivedSchemaID, s -> Optional.of(clusterName));
 
       SchemaReader kafkaMessageEnvelopeSchemaReader = ClientFactory.getSchemaReader(
           ClientConfig.cloneConfig(clientConfig)
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
@@ -308,6 +308,8 @@ public abstract class StoreIngestionTask implements Runnable, Closeable {
   private final String[] msgForLagMeasurement;
   private final Runnable runnableForKillIngestionTasksForNonCurrentVersions;
 
+  protected final String clusterName;
+
   public StoreIngestionTask(
       StoreIngestionTaskFactory.Builder builder,
       Store store,
@@ -321,6 +323,7 @@ public abstract class StoreIngestionTask implements Runnable, Closeable {
       Queue<VeniceNotifier> notifiers) {
     this.readCycleDelayMs = storeConfig.getKafkaReadCycleDelayMs();
     this.emptyPollSleepMs = storeConfig.getKafkaEmptyPollSleepMs();
+    this.clusterName = storeConfig.getClusterName();
     this.databaseSyncBytesIntervalForTransactionalMode = storeConfig.getDatabaseSyncBytesIntervalForTransactionalMode();
     this.databaseSyncBytesIntervalForDeferredWriteMode = storeConfig.getDatabaseSyncBytesIntervalForDeferredWriteMode();
     this.kafkaProps = kafkaConsumerProperties;
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
@@ -308,8 +308,6 @@ public abstract class StoreIngestionTask implements Runnable, Closeable {
   private final String[] msgForLagMeasurement;
   private final Runnable runnableForKillIngestionTasksForNonCurrentVersions;
 
-  protected final String clusterName;
-
   public StoreIngestionTask(
       StoreIngestionTaskFactory.Builder builder,
       Store store,
@@ -323,7 +321,6 @@ public abstract class StoreIngestionTask implements Runnable, Closeable {
       Queue<VeniceNotifier> notifiers) {
     this.readCycleDelayMs = storeConfig.getKafkaReadCycleDelayMs();
     this.emptyPollSleepMs = storeConfig.getKafkaEmptyPollSleepMs();
-    this.clusterName = storeConfig.getClusterName();
     this.databaseSyncBytesIntervalForTransactionalMode = storeConfig.getDatabaseSyncBytesIntervalForTransactionalMode();
     this.databaseSyncBytesIntervalForDeferredWriteMode = storeConfig.getDatabaseSyncBytesIntervalForDeferredWriteMode();
     this.kafkaProps = kafkaConsumerProperties;
--- a/clients/da-vinci-client/src/test/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTaskTest.java
+++ b/clients/da-vinci-client/src/test/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTaskTest.java
@@ -3522,6 +3522,7 @@ public abstract class StoreIngestionTaskTest {
     doReturn(databaseSyncBytesIntervalForDeferredWriteMode).when(storeConfig)
         .getDatabaseSyncBytesIntervalForDeferredWriteMode();
     doReturn(false).when(storeConfig).isReadOnlyForBatchOnlyStoreEnabled();
+    // doReturn(clusterName).when(storeConfig).getClusterName();
     storeVersionConfigOverride.accept(storeConfig);
     return storeConfig;
   }
--- a/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/input/kafka/KafkaInputFormat.java
+++ b/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/input/kafka/KafkaInputFormat.java
@@ -8,12 +8,15 @@ import com.linkedin.venice.hadoop.input.kafka.avro.KafkaInputMapperKey;
 import com.linkedin.venice.hadoop.input.kafka.avro.KafkaInputMapperValue;
 import com.linkedin.venice.kafka.TopicManager;
 import com.linkedin.venice.kafka.TopicManagerRepository;
+import com.linkedin.venice.pubsub.PubSubClientsFactory;
 import com.linkedin.venice.pubsub.PubSubTopicRepository;
 import com.linkedin.venice.pubsub.adapter.kafka.admin.ApacheKafkaAdminAdapterFactory;
 import com.linkedin.venice.pubsub.adapter.kafka.consumer.ApacheKafkaConsumerAdapterFactory;
+import com.linkedin.venice.pubsub.adapter.kafka.producer.ApacheKafkaProducerAdapterFactory;
 import com.linkedin.venice.pubsub.api.PubSubConsumerAdapter;
 import com.linkedin.venice.utils.VeniceProperties;
 import java.io.IOException;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.List;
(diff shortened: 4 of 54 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
```diff
@@ -186,7 +186,11 @@ public DaVinciBackend(
           new VeniceWriterFactory(backendProps.toProperties(), pubSubClientsFactory.getProducerAdapterFactory(), null);
       String instanceName = Utils.getHostName() + "_" + Utils.getPid();
       int derivedSchemaID = backendProps.getInt(PUSH_STATUS_STORE_DERIVED_SCHEMA_ID, 1);
-      pushStatusStoreWriter = new PushStatusStoreWriter(writerFactory, instanceName, derivedSchemaID);
+      Map<String, VeniceWriterFactory> writerFactoryMap = new HashMap<>(1);
+      String clusterName = backendConfig.getClusterName();
```
xunyin8, line 190: DVC is cluster agnostic and can discover cluster using its NativeMetadataRepository since a DVC can be subscribed to multiple stores belonging to different Venice clusters. You can use the `ClusterInfoProvider` interface to get cluster name given a store name.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -674,6 +697,28 @@ public void handleDeletedInstances(Set<Instance> deletedInstances) {
     }
   }
 
+  private Optional<String> discoverClusterFromStore(PubSubTopic pubSubTopic) {
+    if (pubSubTopic.getPubSubTopicType().equals(PubSubTopicType.ADMIN_TOPIC)) {
+      return Optional.of(AdminTopicUtils.getClusterNameFromTopicName(pubSubTopic.getName()));
+    }
+    String storeName = pubSubTopic.getStoreName();
+    if (VeniceSystemStoreUtils.isSystemStore(storeName)) {
+      if (storeName.equals(AvroProtocolDefinition.METADATA_SYSTEM_SCHEMA_STORE.getSystemStoreName())
```
lluwm, line 706: Can we combine this two if conditions for simplicity?

internal/venice-common/src/main/java/com/linkedin/venice/kafka/TopicManager.java
```diff
@@ -77,64 +85,118 @@ public class TopicManager implements Closeable {
...
 
+  private PubSubConsumerAdapter getConsumerAdapter(PubSubTopic pubSubTopic) {
+    if (pubSubTopic == null) {
+      return pubSubConsumerAdapter.get();
+    }
+    Optional<String> clusterName = clusterNameSupplier.get(pubSubTopic);
+    if (!clusterName.isPresent()) {
+      return pubSubConsumerAdapter.get();
+    }
+    return pubSubConsumerAdapterMap.computeIfAbsent(
```
lluwm, line 116: Is it possible that `getConsumerAdapter` being called concurrently from different threads?  `pubSubConsumerAdapterMap` probably needs to be a `VeniceConcurrentHashMap`, if it is.

internal/venice-common/src/main/java/com/linkedin/venice/kafka/TopicManager.java
```diff
@@ -77,64 +85,118 @@ public class TopicManager implements Closeable {
...
+      return pubSubDefaultAdminAdapter.get();
+    }
+    Optional<String> clusterName = clusterNameSupplier.get(pubSubTopic);
+    if (!clusterName.isPresent() || pubSubClientsFactoryMap == null) {
+      return pubSubDefaultAdminAdapter.get();
+    }
+    if (!pubSubClientsFactoryMap.containsKey(clusterName.get())) {
+      throw new PubSubClientException("No PubSubClientsFactory found for cluster: " + clusterName.get());
+    }
+    return pubSubAdminAdapterMap.computeIfAbsent(clusterName.get(), c -> {
```
lluwm, line 138: Ditto for `pubSubAdminAdapterMap`.

internal/venice-common/src/main/java/com/linkedin/venice/kafka/TopicManager.java
```diff
@@ -77,64 +85,118 @@ public class TopicManager implements Closeable {
   private final String pubSubBootstrapServers;
   private final long kafkaOperationTimeoutMs;
   private final long topicMinLogCompactionLagMs;
-  private final PubSubAdminAdapterFactory<PubSubAdminAdapter> pubSubAdminAdapterFactory;
-  // TODO: Use single PubSubAdminAdapter for both read and write operations
-  private final Lazy<PubSubAdminAdapter> pubSubWriteOnlyAdminAdapter;
-  private final Lazy<PubSubAdminAdapter> pubSubReadOnlyAdminAdapter;
+  private final Lazy<PubSubAdminAdapter> pubSubDefaultAdminAdapter;
+  private final Lazy<PubSubConsumerAdapter> pubSubConsumerAdapter;
+  private final Map<String, PubSubAdminAdapter> pubSubAdminAdapterMap = new HashMap<>();
```
lluwm, line 90: Reading the type declaration of `pubSubDefaultAdminAdapter` raises a question that if it's necessary to use the `Lazy<PubSubAdminAdapter>` as the value in the map? Same for the below `pubSubConsumerAdapterMap`.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
haoxu07: For DVC and server side, we do not need clusterName discovery logic. As each host naturally belong to a specific cluster.

xunyin8: That assumption is true for server but how is it true for DVC? What if the DVC user subscribe to multiple stores that span across multiple venice clusters?

haoxu07: That is a valid concern.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -672,6 +687,28 @@ public void handleDeletedInstances(Set<Instance> deletedInstances) {
...
+          || storeName.equals(AvroProtocolDefinition.PUSH_STATUS_SYSTEM_SCHEMA_STORE.getSystemStoreName())) {
+        return Optional.of(multiClusterConfigs.getSystemSchemaClusterName());
+      }
+    }
+
+    try {
+      String clusterName = discoverCluster(pubSubTopic.getStoreName()).getFirst();
+      LOGGER.debug("Discovered cluster {} from store for topic {}", clusterName, pubSubTopic);
```
sixpluszero, line 704: debug level logging is not printed by default

haoxu07: As this path will be called very frequently, better to keep it as debug level.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
```diff
@@ -186,7 +186,11 @@ public DaVinciBackend(
           new VeniceWriterFactory(backendProps.toProperties(), pubSubClientsFactory.getProducerAdapterFactory(), null);
       String instanceName = Utils.getHostName() + "_" + Utils.getPid();
       int derivedSchemaID = backendProps.getInt(PUSH_STATUS_STORE_DERIVED_SCHEMA_ID, 1);
-      pushStatusStoreWriter = new PushStatusStoreWriter(writerFactory, instanceName, derivedSchemaID);
+      Map<String, VeniceWriterFactory> writerFactoryMap = new HashMap<>(1);
```
ZacAttack, line 189: Why (1)?  It seems we don't do this in other similar spots in this PR, but why here?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
ZacAttack: This comment is kind of old, but I think it's pretty valid.  Are we going to address this?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -300,6 +300,8 @@ public abstract class StoreIngestionTask implements Runnable, Closeable {
   private final String[] msgForLagMeasurement;
   private final Runnable runnableForKillIngestionTasksForNonCurrentVersions;
 
+  protected final String clusterName;
```
ZacAttack, line 303: Why is this needed?

clients/da-vinci-client/src/test/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTaskTest.java
```diff
@@ -3498,6 +3498,7 @@ private VeniceStoreVersionConfig getDefaultMockVeniceStoreVersionConfig(
     doReturn(databaseSyncBytesIntervalForDeferredWriteMode).when(storeConfig)
         .getDatabaseSyncBytesIntervalForDeferredWriteMode();
     doReturn(false).when(storeConfig).isReadOnlyForBatchOnlyStoreEnabled();
+    // doReturn(clusterName).when(storeConfig).getClusterName();
```
ZacAttack, line 3501: Do we still need this?

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/input/kafka/KafkaInputFormat.java
```diff
@@ -44,12 +47,17 @@ public class KafkaInputFormat implements InputFormat<KafkaInputMapperKey, KafkaI
...
+        new ApacheKafkaAdminAdapterFactory());
+
     try (TopicManagerRepository topicManagerRepository = TopicManagerRepository.builder()
         .setPubSubProperties(k -> consumerProperties)
         .setLocalKafkaBootstrapServers(config.get(KAFKA_INPUT_BROKER_URL))
         .setPubSubTopicRepository(pubSubTopicRepository)
-        .setPubSubAdminAdapterFactory(new ApacheKafkaAdminAdapterFactory())
-        .setPubSubConsumerAdapterFactory(new ApacheKafkaConsumerAdapterFactory())
+        .setDefaultPubSubClientsFactory(pubSubClientsFactory)
+        .setClusterNameSupplier(s -> Optional.empty())
```
ZacAttack, line 60: Maybe have a default name supplier that's s->Optional.empty() in topicManagerRepository?

ZacAttack:
I think I need some explanation here.  I understand that the intent is that we want to change the pubsub implementation under the hood cluster by cluster.  Thats just an artifact of how we usually roll out features.  However, it seems unfortunate and potentially brittle looking forward to leak the cluster->store association into all these different components.  Particularly since up to this point, the architecture supports materializing a given store in a different locations (store move, davinci, etc.).  

Also, are there no gotcha's here for store move?  Like, if we move a store from one cluster with a certain pubsub implementation to another cluster, will it continue to work?

I think I need to hear a compelling justification for why we couldn't wrap this up in a different way.  One way that springs to mind is that instead of coupling this to clusters we couple it to stores?  Maybe that's naive.

haoxu07:
> I think I need some explanation here. I understand that the intent is that we want to change the pubsub implementation under the hood cluster by cluster. Thats just an artifact of how we usually roll out features. However, it seems unfortunate and potentially brittle looking forward to leak the cluster->store association into all these different components. Particularly since up to this point, the architecture supports materializing a given store in a different locations (store move, davinci, etc.).
> 
> Also, are there no gotcha's here for store move? Like, if we move a store from one cluster with a certain pubsub implementation to another cluster, will it continue to work?
> 
> I think I need to hear a compelling justification for why we couldn't wrap this up in a different way. One way that springs to mind is that instead of coupling this to clusters we couple it to stores? Maybe that's naive.
> 
> At the very least, it seems most of the logic provided discerns the clusterName adjacently from the topic name. It seems like we can pull that logic into the adapter implementations and discern it centrally there as opposed to breeching all these interfaces.



> I think I need some explanation here. I understand that the intent is that we want to change the pubsub implementation under the hood cluster by cluster. Thats just an artifact of how we usually roll out features. However, it seems unfortunate and potentially brittle looking forward to leak the cluster->store association into all these different components. Particularly since up to this point, the architecture supports materializing a given store in a different locations (store move, davinci, etc.).
> 
> Also, are there no gotcha's here for store move? Like, if we move a store from one cluster with a certain pubsub implementation to another cluster, will it continue to work?
> 
> I think I need to hear a compelling justification for why we couldn't wrap this up in a different way. One way that springs to mind is that instead of coupling this to clusters we couple it to stores? Maybe that's naive.
> 
> At the very least, it seems most of the logic provided discerns the clusterName adjacently from the topic name. It seems like we can pull that logic into the adapter implementations and discern it centrally there as opposed to breeching all these interfaces.

Currently clusterName is still not a property for store and this cause trouble discovering cluster for store. Regarding store move, as long as we have pub sub client implementation for source cluster and destination cluster, then this still work, it is only up to which cluster this store belongs at the time point we want to use pub sub client to do sth. 

I think let adapter implementation getting aware of cluster name is not good, clusterName is venice definition. Maybe we can put clusterName into Store or PubSubTopic, that is the question I am thinking. I feel that would also be a bigger change than this.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
haoxu07: Since this cluster by cluster is mainly used for controller only. I do not plan to do it in this PR, but the concern is valid, we can do it in future.

sixpluszero: Hmm, sorry I am not sure if I am fully follow your intention of this PR, if this is only for controller, I think you could skip DVC for now? As for seas-careers, their DVC use cases are spanning across different clusters already. This might not be correct for them. 
(Some of the config in backend config does not apply to all DVC stores)

internal/venice-common/src/main/java/com/linkedin/venice/kafka/TopicManager.java
sixpluszero: I think if it is optional lazy is good, otherwise no need to have lazy here.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
ZacAttack: Yeah I don't see a good reason here to introduce a potential bug into a client library that even if harmless we'll have to force everyone to upgrade off of at some point.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
haoxu07: Using (1) here is for having fixed `clusterName` for all stores in this `DavinciBackend`, as I changed the constructor of `PushStatusStoreWriter` for using `clusterName` to `VeniceWriterFactory` map and cluster discovery supplier function. For controller, it will use proper cluster discover supplier function. As we want to have a PubSubClientFactory for all stores regardless which cluster a specific store belongs to, we just using a placeholder clusterName here as only one entry for this map, and each store discovery logic will return that clusterName to help it find that specific VencieWriterFactory for every store.

PR #603, opened 2023-08-22
[vpj] Fixed KIF repush to handle chunk cleanup messages

gaojieliu:
The behavior of the regression is as follows:
1. The chunk delete message will use `-10` as schema id.
2. When KIF encounters a delete message, it would ignore the schema id and treat the key as a top-level key and it would use non-chunk-suffix to decide the key portion from the composite key, which is wrong.
3. During shuffle and sort phase, if there is no collsion between the regular keys and the wrong key extracted from (2), no issue, and if there is a collision (which is rare), this entry will be removed from repush, which would result in data loss.

This code change fixes the regression in the following way:
1. Use delete.schema_id to decide how to handle delete key.
2. Skip the chunk cleanup message when assembling large message in repush. There is an existing E2E test for this scenario:
PartialUpdateTest#testReplicationMetadataChunkingE2E And the reason it didn't fail is that the exact condition as described above didn't trigger.

## How was this PR tested?
CI

### DIFF
--- a/internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
@@ -190,9 +190,9 @@ public class VeniceWriter<K, V, U> extends AbstractVeniceWriter<K, V, U> {
    */
   public static final int VENICE_DEFAULT_VALUE_SCHEMA_ID = -1;
 
-  private static final byte[] EMPTY_BYTE_ARRAY = new byte[0];
+  public static final byte[] EMPTY_BYTE_ARRAY = new byte[0];
 
-  private static final ByteBuffer EMPTY_BYTE_BUFFER = ByteBuffer.wrap(EMPTY_BYTE_ARRAY);
+  public static final ByteBuffer EMPTY_BYTE_BUFFER = ByteBuffer.wrap(EMPTY_BYTE_ARRAY);
 
   public static final LeaderMetadataWrapper DEFAULT_LEADER_METADATA_WRAPPER =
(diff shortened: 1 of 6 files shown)


internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
```diff
@@ -190,9 +190,9 @@ public class VeniceWriter<K, V, U> extends AbstractVeniceWriter<K, V, U> {
    */
   public static final int VENICE_DEFAULT_VALUE_SCHEMA_ID = -1;
 
-  private static final byte[] EMPTY_BYTE_ARRAY = new byte[0];
```
ZacAttack, line 193: Not protected?

ZacAttack (approved):
Looks good, optional comment.

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
gaojieliu: I am using it in a separate unit test and since it is `static final`, I guess it is fine.

majisourav99: you may use `EMPTY_BYTE_ARRAY` from io commons.

sixpluszero:
Does it mean when it encounter a DELETE chunk, the key split is wrong for now and it could not recover the original key?

gaojieliu:
> Does it mean when it encounter a DELETE chunk, the key split is wrong for now and it could not recover the original key?

That is correct.

sixpluszero (approved):
Thanks for the fix and get to the root cause that current integration test won't trigger such issue. This is really tricky...

PR #604, opened 2023-08-23
VIP-3: Rust Server Read Path

FelixGV:
Also tweaked the VIP template, main VIP landing page and VIP-1 in order to improve the navigation.

Added guidance on how to embed images in the doc_guide.

The rendered page for this VIP can be seen here: https://felixgv.github.io/venice/docs/proposals/vip-3


docs/assets/images/vip_3_read_path.drawio.xml
```diff
@@ -0,0 +1,212 @@
+<mxfile host="app.diagrams.net" modified="2023-08-23T13:47:02.064Z" agent="Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/16.6 Safari/605.1.15" etag="jDuoTWvUTu5oXmrRuByY" version="21.6.9" type="device">
```
manujose0, line 1: Confused on why there is two paths from FC -> RustServer via gRPC Handler? 

Why do we need the gPRC Request Handler - JNI call -> Rust Server while we have FC directly reaching out to RUST server which embeds a gRPC Request handler ?

FelixGV: This is a transitory stage, where we would support both the legacy and new paths at the same time. The end state would be that there's no Java code left to handle read requests (though that is probably far out...).

manujose0: Ahh I see, while its getting tested. Thanks for clarification.

ZacAttack:
So, I won't belabor the rust vs. C++ conversation, I don't think I'm really qualified to die on either hill.  

But I do wonder,  Rocksdb is written in C++.  Today, the Venice team is far more java oriented.  If we go for a native read path, we're going to have to build a competency in some non managed language in the team.  It strikes me that if we build that competency in the same language that rocksdb is implemented, then we'd gain more introspection within the team in that layer.  Admittedly, the situations where we're digging through the guts of rocksdb are few and far in between, but maybe that would start to change?

I kind of liken this to our Kafka clients.  We're not doing big changes in the clients, but by being able to read the code and understanding how the clients/Kafka works, we've been able to make design decisions more informed based on that.

docs/proposals/vip-3.md
```diff
@@ -0,0 +1,226 @@
...
+   their traffic across the two ports, they could get up to twice their configured quota.
+
+Since the migration is expected to be controlled by the operator, and should only be activated for users of the Fast
+Client (i.e. not for users that run a mix of Thin Clients and Fast Clients), the edge case solved by the more 
+complicated solution 1 does not seem likely to come up. Therefore, the proposal is to pick the simpler solution 2.
+
+Also, it can be considered whether quota should be part of the MVP at all. If the Rust server mode will be tried in 
+dedicated clusters at first, then quota may not even be a must-have (though of course it will be needed eventually).
+
+### Cluster Discovery
```
ZacAttack, line 151: ACL's also needs some retweak.

FelixGV: Yeah. I mentioned it in passing in the Development Milestones section. I don't have much thoughts on how to flesh it out further for the time being, besides saying that it's needed. If you have any inputs, please share!

ZacAttack: Weeeeelllll, since you asked :)

Something I've always lamented about the current implementation is that for every single request we do a lookup against the validation library to determine if an action is valid.  This is a reflection on how the interface for our internal authorizer works.

In my view, there are a few problems with this performance wise.  The first one is Venice specific, the router only allows 1 kind of operation to a store, so it seems like a pointless lookup for every single request.  The second reason, is that it's not as if the principal changes on a connection.  It's always the same actor on that connection, but we still interrogate it's permissions on every request on that connection.  Add to it, the LinkedIn representation of an Ace is Resource->[Permissions], so every time we're walking a (albeit usually short) list of principals to match up to the principal making the request.  There can be multiple aces for the same principal in this array, so every time we walk the entire array.  Ideally it should be the opposite at query time, being indexed by the principal as opposed to the resource (basically Principal->listOfResources)

So, what might be doable, is that we negotiate the permission once per connection, at time of establishing the connection.  We could just call into the java interface we have today with 'hasAccess', check back on the result, and use that for weather or not we accept the connection or not, and then let the client do whatever they want.  We'd need to tweak the protocol a bit to pass along at time of establishing the store(s) they want to talk to. Also today the transport client doing connection management can be shared between multiple Venice clients, so we'd need to do some wiring in here.

Other complexity gotcha, for internal auditing we might need to tweak this behavior for LDAP typed certificates so it gets audited correctly, but thats an internal detail.

It seems to me that this would be the sweet spot in terms of performance and not have to do anything too drastic in terms of writing equivalent interface/implementations between the native and java layer, and we could even incrementally approach going full native provided that at some point down the road we get a good Rust implementation of the LinkedIn authorizer.

FelixGV:
> So, I won't belabor the rust vs. C++ conversation, I don't think I'm really qualified to die on either hill.  
> 
> But I do wonder,  Rocksdb is written in C++.  Today, the Venice team is far more java oriented.  If we go for a native read path, we're going to have to build a competency in some non managed language in the team.  It strikes me that if we build that competency in the same language that rocksdb is implemented, then we'd gain more introspection within the team in that layer.  Admittedly, the situations where we're digging through the guts of rocksdb are few and far in between, but maybe that would start to change?
> 
> I kind of liken this to our Kafka clients.  We're not doing big changes in the clients, but by being able to read the code and understanding how the clients/Kafka works, we've been able to make design decisions more informed based on that.

Hmm... it is true that if the team gains more hands-on experience with C++, we could better understand the RocksDB codebase than today. But I am not sure that this incremental benefit (meaning: we're not starting from zero understanding today) is sufficiently important to warrant the downsides. The downsides being that the learning curve and productivity of C++ are generally considered to be worse than Rust's.

docs/proposals/vip-3.md
```diff
@@ -0,0 +1,226 @@
...
+
+### Quota
+
+The read quota will need to be re-implemented in Rust. There are essentially two main approaches we can consider:
+
+1. Let the Rust server and Java server share a quota amount between them, so that no matter if the traffic comes in via
+   the Java port or the Rust port, it will be accounted for and the configured limit will be respected. This requires
+   having some kind of synchronization between the two sides (either at the granularity of each request, which is likely
+   too expensive, or by reserving some allotment periodically, which is complicated).
+2. Let the Rust server and Java server each have the full quota amount and not need any interaction between each other
```
gaojieliu, line 140: I assume the initial version of Rust server will only support single-get/batch-get, which doesn't require complex handling logic.
Maybe in the first phase, we only allow the users, which are not using read compute to access Rust Server, so that Rust Server will take all the traffic.
Once Rust Server supports read compute, we can migrate these leftover use cases.

docs/proposals/vip-3.md
FelixGV: Thanks! I think these are really good ideas!

So let me recap to see if I got it:

The Rust server would maintain a mapping of `{principal, resource} -> {isAuthorized, lastCheckTime}`. If the mapping is missing, it would call into the Java server in a blocking way to find out whether this principal is allowed to access that resource, and then update the mapping accordingly. Some background thread could refresh the mapping when `lastCheckTime` is beyond some threshold, which also serves as a heartbeat into the auditing on the Java side. In this way, a Venice operator only needs to implement the Authorizer API once, in Java, regardless of whether the Rust server is enabled or not.

docs/proposals/vip-3.md
```diff
@@ -0,0 +1,226 @@
...
+   migration.
+3. Instead of announcing per-cluster resources, we could announce per-store resources, and then have some admin tooling
+   to enable and disable the announcement for Rust and Java servers on a per-store basis. This could give the most
+   flexibility, since the client would not remain stuck with whatever resource name was returned by the cluster 
+   discovery call done at startup time (and periodically thereafter), but rather could adapt rapidly.
+
+These options provide different levels of granularity of control and speed of change, with finer granularity and faster
+change carrying greater complexity cost.
+
+## Development Milestones
```
gaojieliu, line 176: Venice Server is also using a lot of dependencies, such as Helix, ZK, ACL (maybe some others I don't remember), and I am not sure whether we have rust libraries for all these dependencies.

One strategy I was thinking is to use a push model from the Java process to Rust Server.
It means, Java Process/Server will have a thread to explicitly detect/listen to the changes in Helix/ZK/ACL and push these information in Venice controlled format asynchronously (ASAP) to the rust server, so that Venice could develop these functionalities even Rust Helix lib is not there yet. Once the native Rust support for these dependencies are available, we can switch to them.

gaojieliu:
> Hmm... it is true that if the team gains more hands-on experience with C++, we could better understand the RocksDB codebase than today. But I am not sure that this incremental benefit (meaning: we're not starting from zero understanding today) is sufficiently important to warrant the downsides. The downsides being that the learning curve and productivity of C++ are generally considered to be worse than Rust's.

When I read RocksDB code, my feeling is that I can certainly understand the code even written in C++, and I think the difficult part is the design principle and architecture of RocksDB.
We didn't get involved in any RocksDB development and we don't read RocksDB that often since it is quite stable, and I think that might be the main reason we don't understand RocksDB that much.
Language is not the biggest barrier from my understanding. RocksDB has been developed by many people in the past 10+ years,  understanding the design principle and architecture and techs being used is most challenging part IMHO.

docs/proposals/vip-3.md
FelixGV: Yes, it would be convenient to run only one implementation at a time... I think it is a good idea.

docs/proposals/vip-3.md
FelixGV: Regarding ACLs, there is a similar discussion going on here: https://github.com/linkedin/venice/pull/604#discussion_r1303656727

Regarding Helix, I am thinking that ideally the Rust server doesn't need to know about it at all, but maybe that is not quite possible. When the Java side does the JNI call to create/open a RocksDV store-version-partition, the Rust wrapper can remember the fact that this partition exists, and therefore serve it when requested. Likewise when the JNI call comes in to drop a RocksDB partition, the Rust wrapper needs to respond with an error if it receives a query for that. These RocksDB calls would be side-effects of the Java-side Helix state transitions. If that is sufficient, then no extra exchange protocol is needed as far as Helix is concerned...

Where I think we will definitely need to explore the idea you're bringing up is for Read Compute, where we'll need the Rust side to know about schemas, and those can be added over time.

There might be other situations too, such as quota changes, where Java needs to pass Venice-specific info to Rust.

docs/proposals/vip-3.md
ZacAttack: Thats about right.

ahmadbeirkdar:
> So, I won't belabor the rust vs. C++ conversation, I don't think I'm really qualified to die on either hill.
> 
> But I do wonder, Rocksdb is written in C++. Today, the Venice team is far more java oriented. If we go for a native read path, we're going to have to build a competency in some non managed language in the team. It strikes me that if we build that competency in the same language that rocksdb is implemented, then we'd gain more introspection within the team in that layer. Admittedly, the situations where we're digging through the guts of rocksdb are few and far in between, but maybe that would start to change?

This could hold true, but not sure how deep you folks want to dig into RocksDB, I doubt y'all will do any development on RocksDB. Even so, if the plan using C++ here so the team will have a more hands on experience in C++ because that is what RocksDB is written in, then in my opinion Rust is still a better option as Rust programmers tend to treat and write C++ code in a Rust way vs the old style C++ way. Where as onboarding engineers who previously only used Java or Python, they will write C++ code the Java way which will lead into problems w/ memory leaks and safety. The Rust -> Modern C++ transition is quicker and the safer one. But that is even IF the team decides to work on RocksDB code

ahmadbeirkdar:
Love the proposal, don't have much to say from a architecture, justification and a higher level aspect it looks promising.

Would love to see how things plan to be implemented, from the choice of the gRPC server to choice of how you folks decide to build the rust server, lots of promising things in that space, especially for clean, robust and explicit external APIs that insure idiomatic Rust code

docs/proposals/vip-3.md
```diff
@@ -0,0 +1,226 @@
...
+Since the migration is expected to be controlled by the operator, and should only be activated for users of the Fast
+Client (i.e. not for users that run a mix of Thin Clients and Fast Clients), the edge case solved by the more 
+complicated solution 1 does not seem likely to come up. Therefore, the proposal is to pick the simpler solution 2.
+
+Also, it can be considered whether quota should be part of the MVP at all. If the Rust server mode will be tried in 
+dedicated clusters at first, then quota may not even be a must-have (though of course it will be needed eventually).
+
+### Cluster Discovery
+
+The general idea is that when the Rust mode is activated in a server, that server would announce that it has a gRPC
```
sixpluszero, line 153: For metrics (especially read metrics), now it is inside Rust Server then? (Since for the FC->Rust path no Java is touched) 
Just confirming are we going to do metrics aggregation just like what II is doing today?

docs/proposals/vip-3.md
FelixGV: I've included this within the VIP. Thanks!

docs/proposals/vip-3.md
FelixGV: I am thinking to do it similarly to II, yes. If you have feedback on that approach (either for or against), please share!

docs/proposals/vip-3.md
```diff
@@ -0,0 +1,246 @@
...
+   that JNI via Rust is no worse than JNI via C++.
+3. Integrate the chosen crate for adding gRPC support, and do a "hello world" of the RPC.
+4. Add server configs to enable the Rust gRPC module (including basic service disco announcement).
+5. Implement the batch get support.
+6. Implement support for auth/auth.
+7. Multi-architecture (Linux + Mac) support.
+8. (Possibly optional) implement quota.
+9. (Possibly optional) more advanced/granular service disco control.
+
+Beyond the MVP, the following scope expansions are natural followups:
```
sixpluszero, line 212: If I understand this idea correctly, this should achieve:
(1) full ingestion isolation without hybrid store continuous ingestion concern.
(2) Simplify write path and state transition (back to non-II state without maintaining state machine between main and forked)
The concern might mostly be:
(1) Effort to develop the Rust impl (Well we are all young so should be easy to learn, but for site maintenance perspective I am not sure if SRE will have concern on this. My understanding is if keep providing enough metrics/loggings it should be fine).
(2) I know other ppl chat about C++ vs Rust about pure performance, I am not super concern about that...as we were fine with Java today. The one minor concern is https://github.com/rust-rocksdb/rust-rocksdb I think it would be the wrapper of RocksDB to use in Rust Read Process. It is not maintained by RocksDB main contributor, but rather a community support, as it is yet another wrapper on top of original C++ implementation. I am not sure if the support is up to date and there is any issue / future issue with that. 

Also, since we have limited dev resources today, if we are determined to prioritize this in either R/W team, we should pivot our plan from exploring II to this. (Which is partially done, but not yet)

docs/proposals/vip-3.md
sixpluszero: I think it depends on if LI sidecar container support is fully there or not, for now it is not there, so we have to consolidate metrics into single Java process for our InGraph to consume. 
For OSS Venice, either way is fine, we can expose endpoints to let external consumer to consume metrics from Rust Process.

docs/proposals/vip-3.md
sixpluszero: Although this project is orthogonal to blob ingestion idea, which is great. 
I am seeing stronger need to have blob ingestion in our Venice support -> for both Da Vinci and Server

FelixGV: You bring up several important points here, so let me try to address each of them...

Regarding II, the way I see it, if we choose to go forward with the Rust server proposal, then we should cancel the development of hybrid II. As for batch II, we need to make a judgment call... batch II has proven useful in some of our workloads, and if it is close to being stable, then it would be great to leverage it for the time being, since the Rust server may take 6-12 months to land in production. If batch II is far from being stable and we need another 6 months to stabilize it, then of course it's not as attractive... we don't know that for sure, but we'll need to make a judgment call. Besides, even if we made the call that batch II in servers is too troublesome to stabilize and we want to abandon it, we are still using batch II for DVC, and that still needs to be supported going forward (unless/until blob ingestion comes around, I guess). In the end state, when the Rust server is available, at that point it seems logical that we may want to stop using batch II for server, but I'm not 100% sure... there may be a reason to keep it around (e.g. batch vs hybrid isolation perhaps).

I am hoping to be able to use the crate coming of the rust-rocksdb repo you linked. It seems to have gotten a fair amount of development, with the latest commit being not that long ago, so it looks promising. If something is missing or unsatisfactory, we can consider contributing to it, or other alternatives...

docs/proposals/vip-3.md
FelixGV: I added a metrics section to the doc.

docs/proposals/vip-3.md
```diff
@@ -0,0 +1,257 @@
...
+capacity as well.
+
+The specific language proposed to replace Java is [Rust](https://www.rust-lang.org), which has become mature and well 
+accepted in the industry over the past years. It promises performance similar to C++, without some of its pitfalls, 
+thanks to its borrow checker.
+
+## Problem Statement 
+
+The Venice read path should be completely free of garbage collection. Venice should be able to serve single get read
+requests, end-to-end (i.e. as measured by the client), at a p99.99 latency under 1 millisecond.
```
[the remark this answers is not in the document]
FelixGV: We don't know yet whether we will achieve this goal, but the goal can be defined either way :D ...

That being said, the number is not completely pulled out of a hat. Based on performance numbers achieved by other systems with a thick client and GC-free server implementation, I believe it is realistic.

docs/proposals/vip-3.md
```diff
@@ -0,0 +1,257 @@
...
+    (or possibly a mix, where only some servers enable the Rust implementation).
+
+From the user's perspective, as long as they are using the Fast Client with the gRPC mode enabled, they should be
+eligible to benefiting from the Rust server, if it happens to be enabled by the operator.
+
+## Proposed Design
+
+In a nutshell, the idea is to create a Rust program which wraps RocksDB, and presents to our server JVM the same JNI
+APIs which RocksJava has today (or at least the subset we use). In this way, the server continues to use RocksDB as
+normal, for both the ingestion path and the legacy read path. However, this Rust program would also open up a port and
```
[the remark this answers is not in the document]
FelixGV: No, I do not intend to leverage RocksDB secondary instance.

adamxchen (approved):
LGTM

PR #606, opened 2023-08-24
[controller] Add system store health check service in child controller

sixpluszero:
## [controller] Add system store health check service in child controller

Add system store health check service in child controller, it is controlled by newly added configs: 
"controller.system.store.health.check.enabled" and "controller.system.store.health.check.interval" by default 1h.
This PR is the first part of the change, the follow up change is to add parent controller support to query and repair system store.

## How was this PR tested?
New unit tests for sub-components


services/venice-controller/src/main/java/com/linkedin/venice/controller/systemstore/SystemStoreHealthCheckService.java
```diff
@@ -0,0 +1,164 @@
...
+
+  @Override
+  public void stopInner() {
+    isRunning.set(false);
+    checkServiceExecutor.shutdownNow();
+  }
+
+  private class SystemStoreHealthCheckTask implements Runnable {
+    @Override
+    public void run() {
```
lluwm, line 91: Nit, could we add one line of log to indicate the service states when it is started/stopped.

services/venice-controller/src/main/java/com/linkedin/venice/controller/systemstore/SystemStoreHealthCheckService.java
```diff
@@ -0,0 +1,164 @@
...
+    this.metaStoreReader = metaStoreReader;
+    this.pushStatusStoreWriter = pushStatusStoreWriter;
+    this.pushStatusStoreReader = pushStatusStoreReader;
+    this.checkPeriodInSeconds = systemStoreCheckPeriodInSeconds;
+  }
+
+  /**
+   * Return unhealthy system store name set. This API is expected to be called by parent controller.
+   */
+  public Set<String> getUnhealthySystemStoreSet() {
```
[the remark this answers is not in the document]
sixpluszero: I am wondering if strict visibility is needed here:
The getter will be used in the followup PR for getting the unhealthy system stores from parent controller call. This will be periodic (1h maybe) and this child periodic job will also be something like once per hour. The set will be updated, but it is OK so see the old copy. Are you saying that without volatile, it may never see the update in memory from another thread?

services/venice-controller/src/main/java/com/linkedin/venice/controller/HelixVeniceClusterResources.java
```diff
@@ -206,6 +208,23 @@ public HelixVeniceClusterResources(
...
+      }
+      this.systemStoreHealthCheckService = new SystemStoreHealthCheckService(
+          storeMetadataRepository,
+          admin.getMetaStoreReader(),
+          admin.getMetaStoreWriter(),
+          admin.getPushStatusStoreReader().get(),
+          admin.getPushStatusStoreWriter().get(),
+          config.getDaVinciPushStatusScanIntervalInSeconds());
+    } else {
+      this.systemStoreHealthCheckService = null;
```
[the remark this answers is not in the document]
sixpluszero: Yeah that's true.

services/venice-controller/src/main/java/com/linkedin/venice/controller/systemstore/SystemStoreHealthCheckService.java
```diff
@@ -0,0 +1,164 @@
...
+        checkPeriodInSeconds,
+        checkPeriodInSeconds,
+        TimeUnit.SECONDS);
+    return true;
+  }
+
+  @Override
+  public void stopInner() {
+    isRunning.set(false);
+    checkServiceExecutor.shutdownNow();
```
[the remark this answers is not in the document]
sixpluszero: Hmm, actually I use this intentionally as in this case (controller is shutting down or it is no longer the leader controller) we should not care ongoing data collection task...

services/venice-controller/src/main/java/com/linkedin/venice/controller/systemstore/SystemStoreHealthCheckService.java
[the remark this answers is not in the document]
lluwm: My understanding is that without volatile/AtomicReference, there is no guarantee in when the updates (cause we are doing reference re-assignment periodically) can be seen by another thread. The thread is free to cache this value locally and I probably won't be surprised if it cached it for a very long time.

services/venice-controller/src/main/java/com/linkedin/venice/controller/HelixVeniceClusterResources.java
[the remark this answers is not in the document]
sixpluszero: Actually no, the variable is final, so in each branch we need to set a value for once.

services/venice-controller/src/main/java/com/linkedin/venice/controller/systemstore/SystemStoreHealthCheckService.java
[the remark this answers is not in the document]
lluwm: If I am not wrong, shutting down a controller and its JVM requires all non-daemon threads to shutdown first https://docs.oracle.com/javase/8/docs/api/java/lang/Thread.html. So, in the case of controller shutting down, if this executor is not fully down, it could actually prevent the controller JVM from shutting down.

services/venice-controller/src/main/java/com/linkedin/venice/controller/HelixVeniceClusterResources.java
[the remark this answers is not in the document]
lluwm: Sure, keep it if you prefer, just a nit point as mentioned.

services/venice-controller/src/main/java/com/linkedin/venice/controller/systemstore/SystemStoreHealthCheckService.java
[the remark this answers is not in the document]
sixpluszero: For this, I think the current implementation every step we check the isRunning atomic boolean to make early termination. 

I know for other implementations in Venice, we have such pattern:
executorService.shutdown();
executorService.awaitTermination(x, seconds)
if (!executorService.isTerminated()) {
  executorService.shutdownNow();
} 

So my understanding of this pattern is this is still best effort to wait for current submitted task to complete. If the running task does not have early termination logic, it will have the same effect. 
I don't mind changing to above pattern to make it looks more gracefully, but I feel like it is the same. It is basically up to the task logic, unless there is an API to force cancel the task.

services/venice-controller/src/main/java/com/linkedin/venice/controller/systemstore/SystemStoreHealthCheckService.java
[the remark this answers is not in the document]
sixpluszero: Sure, I see your concern, I can update that.

services/venice-controller/src/main/java/com/linkedin/venice/controller/HelixVeniceClusterResources.java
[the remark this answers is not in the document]
sixpluszero: Hmm, just want to clarify, this should not be a nit, it is required by Java compiler that a final variable is assigned on every path inside the constructor. Since the variable does not need to be re-assigned, I use final modifier here so it requires declaring the value on the "else" branch...

services/venice-controller/src/main/java/com/linkedin/venice/controller/systemstore/SystemStoreHealthCheckService.java
sixpluszero: Yeah, sure, looks like current impl does not have any log, I will add some for critical places for debugging purpose.

services/venice-controller/src/main/java/com/linkedin/venice/controller/systemstore/SystemStoreHealthCheckService.java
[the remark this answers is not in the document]
lluwm: Reading the [example code](https://docs.oracle.com/javase%2F7%2Fdocs%2Fapi%2F%2F/java/util/concurrent/ExecutorService.html#awaitTermination(long,%20java.util.concurrent.TimeUnit)) from official Javadoc, `shutdown` call seems not very necessary given that this is a single thread pool with just one ever-running Task. I agree with you that it is not guaranteed to always cancel the task, but that part that I like is that, once the executor can not be terminated within the given timeout period, there can have a log line to let us to know the issue. Maybe, the following lines would be good enough in this case.

```
       pool.shutdownNow(); // Cancel currently executing tasks
       // Wait a while for tasks to respond to being cancelled
       if (!pool.awaitTermination(60, TimeUnit.SECONDS))
           System.err.println("Pool did not terminate");
```

internal/venice-common/src/main/java/com/linkedin/venice/system/store/MetaStoreReader.java
```diff
@@ -0,0 +1,71 @@
...
+    this.d2Client = d2Client;
+    this.clusterDiscoveryD2ServiceName = clusterDiscoveryD2ServiceName;
+  }
+
+  public long getHeartbeat(String storeName) {
+    AvroSpecificStoreClient<StoreMetaKey, StoreMetaValue> client = getVeniceClient(storeName);
+    StoreMetaKey key =
+        MetaStoreDataType.HEARTBEAT.getStoreMetaKey(Collections.singletonMap(KEY_STRING_STORE_NAME, storeName));
+    try {
+      StoreMetaValue value = client.get(key).get();
```
gaojieliu, line 38: I think we need to specify a timeout here to avoid waiting forever.

services/venice-controller/src/main/java/com/linkedin/venice/controller/HelixVeniceClusterResources.java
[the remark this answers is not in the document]
lluwm: Sure, understood. Alternative is to remove `final` and the `else` branch. But, declaring it as `final` also have benefit as well, so it's trade-off  :-)

services/venice-controller/src/main/java/com/linkedin/venice/controller/systemstore/SystemStoreHealthCheckService.java
```diff
@@ -0,0 +1,179 @@
...
+      PushStatusStoreReader pushStatusStoreReader,
+      PushStatusStoreWriter pushStatusStoreWriter,
+      int systemStoreCheckPeriodInSeconds) {
+    this.storeRepository = storeRepository;
+    this.metaStoreWriter = metaStoreWriter;
+    this.metaStoreReader = metaStoreReader;
+    this.pushStatusStoreWriter = pushStatusStoreWriter;
+    this.pushStatusStoreReader = pushStatusStoreReader;
+    this.checkPeriodInSeconds = systemStoreCheckPeriodInSeconds;
+    this.unhealthySystemStoreSet.set(new HashSet<>());
```
lluwm, line 65: We don't really need to new the HashSet from memory, do we? We just periodically update `unhealthySystemStoreSet` to point to the most recent data.

services/venice-controller/src/main/java/com/linkedin/venice/controller/HelixVeniceClusterResources.java
```diff
@@ -206,6 +208,23 @@ public HelixVeniceClusterResources(
...
+      if (!admin.getPushStatusStoreWriter().isPresent()) {
+        throw new VeniceException("Da Vinci push status writer is not enabled.");
+      }
+      this.systemStoreHealthCheckService = new SystemStoreHealthCheckService(
+          storeMetadataRepository,
+          admin.getMetaStoreReader(),
+          admin.getMetaStoreWriter(),
+          admin.getPushStatusStoreReader().get(),
+          admin.getPushStatusStoreWriter().get(),
+          config.getDaVinciPushStatusScanIntervalInSeconds());
```
gaojieliu, line 224: This param doesn't look right, and it should be using the newly added interval config.

services/venice-controller/src/main/java/com/linkedin/venice/controller/systemstore/SystemStoreHealthCheckService.java
```diff
@@ -0,0 +1,179 @@
...
+      Map<String, Long> systemStoreToHeartbeatTimestampMap = new VeniceConcurrentHashMap<>();
+      for (Store store: storeRepository.getAllStores()) {
+        if (!isRunning.get()) {
+          return;
+        }
+        if (!VeniceSystemStoreUtils.isUserSystemStore(store.getName())) {
+          continue;
+        }
+        // It is also possible that the store is a new store and is being empty pushed.
+        if (store.getCurrentVersion() == 0) {
```
gaojieliu, line 121: So there are several error scenarios:
1. Store doesn't have system store.
2. Store has system store without a current version.
3. Store has an active current version, but it is not ingesting.

I see 1) is missing, and is it expected?

services/venice-controller/src/main/java/com/linkedin/venice/controller/systemstore/SystemStoreHealthCheckService.java
```diff
@@ -0,0 +1,179 @@
...
+          retrievedTimestamp = getPushStatusStoreReader()
+              .getHeartbeat(userStoreName, PushStatusStoreUtils.CONTROLLER_HEARTBEAT_INSTANCE_NAME);
+        } else {
+          retrievedTimestamp = getMetaStoreReader().getHeartbeat(userStoreName);
+        }
+        if (retrievedTimestamp < heartbeatTimestamp) {
+          throw new VeniceException("Heartbeat not refreshed.");
+        }
+        return true;
+      }, 3, Duration.ofSeconds(1), Collections.singletonList(VeniceException.class));
```
gaojieliu, line 166: So the healthiness of system store criteria depends on the ordering of the iteration.
For the first store, the heart-beat message needs to be consumed in 3 seconds, otherwise, we would mark the system store as unhealthy, and for the following system stores, the write delay tolerance would be higher.

This may result in some false alarms.

I think we need to define some E2E write latency SLA for system store, and 3 seconds might be too aggressive.
I would like us to make it configurable and maybe we should wait for at least 1 min before querying to avoid false alarms as much as possible.

gaojieliu (changes requested):

gaojieliu:
I would like us to emit a metric to track the unhealthy system store set to provide some visibilty.

services/venice-controller/src/main/java/com/linkedin/venice/controller/systemstore/SystemStoreHealthCheckService.java
sixpluszero: It is expected. System store is always be created as part of store creation. It is not an independent store object but it is inside user store's ZK node. I don't think we need to check it here. 
The truly important action here should be (2) which is empty push failure (3) it goes back silently

services/venice-controller/src/main/java/com/linkedin/venice/controller/systemstore/SystemStoreHealthCheckService.java
sixpluszero: It is either here, or in the get method. If the parent query comes first before first iteration, you will need to have one to return.

services/venice-controller/src/main/java/com/linkedin/venice/controller/systemstore/SystemStoreHealthCheckService.java
sixpluszero: I was thinking that when writing to the list of the system stores I am implicitly giving them enough time to consume the message. 
But if you think this is not enough, then we can wait for 1 minutes. Thread.sleep(1min) should be good enough for the thread. 

But if we sleep for 1 minutes, I don't think we need to have another config to tune the retry 3 seconds? Otherwise, the 1minutes does not make sense.

internal/venice-common/src/main/java/com/linkedin/venice/system/store/MetaStoreReader.java
sixpluszero: OK I copied the class from PushStatusStoreReader. 

It looks like all the usage in that class also don't have timeout, I am thinking to add a default 1s (that's even too much for this small data) as timeout. What do you think? This is a very specific usage, and I don't think it needs extra config for that.

services/venice-controller/src/main/java/com/linkedin/venice/controller/systemstore/SystemStoreHealthCheckService.java
lluwm: Hmm... why can't return a null pointer if parent query comes before the 1st iteration?

sixpluszero: Nullpointer in my opinion is dangerous, we will have to handle null pointer anyway, some time some where. I prefer we do proper initialization in child controller to avoid this handling. I think an alternative way is to check in get method, but  I think the effect is the same.

sixpluszero:
> I would like us to emit a metric to track the unhealthy system store set to provide some visibilty.

But if we are going to have auto repair and alert for failure to fix the store, do you think it is redundant here?

gaojieliu:
> > I would like us to emit a metric to track the unhealthy system store set to provide some visibilty.
> 
> But if we are going to have auto repair and alert for failure to fix the store, do you think it is redundant here?

There are two reasons:
1. Understand the current state of the system, and if there are always a high number of unhealthy sys stores, which means there are two possiblities:
a. System store healthy check is problematic since it could be too sensitive.
b. Venice ingestion pipeline is too fragile, which might indicate some more serious issue, which we should fix them explicitly rather than auto fix since such issue can apply to regular Venice store as well.
2. Auto-repair may not work for every scenario, and if it can't repair after several rounds of re-push, alert should be triggered.

services/venice-controller/src/main/java/com/linkedin/venice/controller/systemstore/SystemStoreHealthCheckService.java
gaojieliu: What if the system store creation is problematic? For example, empty push didn't hit child controller at all, and sys stores won't be auto created.

services/venice-controller/src/main/java/com/linkedin/venice/controller/systemstore/SystemStoreHealthCheckService.java
gaojieliu: Well, the waiting time is non deterministic.
For example, if the cluster only contains 2 stores, the heart-beat message write won't take more than 2 seconds, that means the gap between sending heart-beat messages and checking the latest heartbeat timestamp will be very short.
That will make it very hard to reason with different types of setup.

I was thinking retry should only be used when client request fails, such as timeout or other errors.

internal/venice-common/src/main/java/com/linkedin/venice/system/store/MetaStoreReader.java
gaojieliu: Let us use a higher timeout, such as 3 seconds since the first several requests can be slow because connection setup and warming.

services/venice-controller/src/main/java/com/linkedin/venice/controller/systemstore/SystemStoreHealthCheckService.java
sixpluszero: >     
    // Get all the regular Venice stores
    List<Store> regularVeniceStores = regularStoreRepository.getAllStores();
    List<Store> allStores = new ArrayList<>(regularVeniceStores);
    // So far, only consider meta system store.
    Store zkSharedStoreForMetaSystemStore =
        systemStoreRepository.getStore(VeniceSystemStoreType.META_STORE.getZkSharedStoreName());
    Store zkSharedStoreForDaVinciPushStatusSystemStore =
        systemStoreRepository.getStore(VeniceSystemStoreType.DAVINCI_PUSH_STATUS_STORE.getZkSharedStoreName());
    // Populate all the systems stores for the regular Venice stores if they are enabled.
    regularVeniceStores.forEach(store -> {
      if (store.isStoreMetaSystemStoreEnabled()) {
        allStores.add(new SystemStore(zkSharedStoreForMetaSystemStore, VeniceSystemStoreType.META_STORE, store));
      }
      if (store.isDaVinciPushStatusStoreEnabled()) {
        allStores.add(
            new SystemStore(
                zkSharedStoreForDaVinciPushStatusSystemStore,
                VeniceSystemStoreType.DAVINCI_PUSH_STATUS_STORE,
                store));
      }
    });
    return allStores;


This is the impl of getAllStores() in store repo. My understanding is for a regular store, two system stores always exist as the flag will be true for them. 

I see addVersionOnly tries to update the store to have two system stores. Are you suggesting that this method might not be hit for some region?

The reason that I did not check it in the current implementation is if I am going to check it I might need to iterate the store list twice. (Although it is affordable).

sixpluszero: Let me resolve all other issues first. I think I am 50/50 towards adding this check or not, if you think this could actually happen, I think I can add it, probably tweak the structure a bit to make unit test coverage happy.

sixpluszero:
> > > I would like us to emit a metric to track the unhealthy system store set to provide some visibilty.
> > 
> > 
> > But if we are going to have auto repair and alert for failure to fix the store, do you think it is redundant here?
> 
> There are two reasons:
> 
> 1. Understand the current state of the system, and if there are always a high number of unhealthy sys stores, which means there are two possiblities:
>    a. System store healthy check is problematic since it could be too sensitive.
>    b. Venice ingestion pipeline is too fragile, which might indicate some more serious issue, which we should fix them explicitly rather than auto fix since such issue can apply to regular Venice store as well.
> 2. Auto-repair may not work for every scenario, and if it can't repair after several rounds of re-push, alert should be triggered.

Sure. I've also added the Stats class to emit bad system store count per cluster per region in the latest changes.

services/venice-controller/src/main/java/com/linkedin/venice/controller/systemstore/SystemStoreHealthCheckService.java
sixpluszero: Synced offline, I've updated the PR to reflected the changes.

services/venice-controller/src/main/java/com/linkedin/venice/controller/systemstore/SystemStoreHealthCheckService.java
lluwm: Not quite sure about the exact concern, but maybe we can live with it. It might have some performance trade-offs though, e.g. memory allocation/deallocation on the heap which itself is not cheap and extra call to constructor.

I think the change looks okay to me, please address other reviews, thanks Jialin!

gaojieliu (commented):
Left some minor comments and overall looks good.

services/venice-controller/src/main/java/com/linkedin/venice/controller/stats/SystemStoreCheckStats.java
```diff
@@ -0,0 +1,21 @@
...
+import io.tehuti.metrics.stats.Count;
+
+
+public class SystemStoreCheckStats extends AbstractVeniceStats {
+  private final Sensor badSystemStoreCount;
+
+  public SystemStoreCheckStats(MetricsRepository metricsRepository, String name) {
+    super(metricsRepository, name);
+
+    badSystemStoreCount = registerSensorIfAbsent("bad_system_store_count", new Count());
```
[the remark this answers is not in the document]
sixpluszero: I expect only a few system stores are going to be bad. If we are going to do that it will be two queues though. And we do have logs printing out the bad store names.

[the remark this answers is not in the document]
gaojieliu: There is no need to maintain two queues.
We just need to scan the unhealthy store sets and break them down into two categories before reporting metrics.

services/venice-controller/src/main/java/com/linkedin/venice/controller/stats/SystemStoreCheckStats.java
```diff
@@ -1,21 +1,26 @@
...
 
 
 public class SystemStoreCheckStats extends AbstractVeniceStats {
-  private final Sensor badSystemStoreCount;
+  private final Sensor badMetaSystemStoreCount;
+  private final Sensor badPushStatusSystemStoreCount;
 
   public SystemStoreCheckStats(MetricsRepository metricsRepository, String name) {
     super(metricsRepository, name);
+    badMetaSystemStoreCount = registerSensorIfAbsent("bad_meta_system_store_count", new Gauge());
```
gaojieliu, line 15: This is not the right way to use `Gauge` metric and maybe we should just remove the default constructor of `Gauge`.
We need to pass a metric value provider when constructing the Gauge metric and you can see the existing usage.

sixpluszero: Oh right... I think I remember that.

gaojieliu (approved):

PR #608, opened 2023-08-25
[producer] Make online producer release resources cleanly if exceptions occur in the constructor

nisargthakkar:
## Make online producer release resources cleanly if exceptions occur in the constructor

In the constructor of `OnlineVeniceProducer`, an exception is thrown if the store cannot accept writes from online applications. Due to this, the producer object is never created - and hence, the resources created inside it are not cleaned up.

For D2 clients, these resources include some executors that create non-daemon threads. When we use D2 client with the `ProducerTool` and the producer throws exceptions in the constructor, the main thread is unable to exit because non-daemon threads are still running.

This commit ensures that these resources are cleaned on such exceptions. This commit also modifies the initialization arguments of the `OnlineVeniceProducer` to reduce some leaking abstractions into `OnlineProducerFactory`.

## How was this PR tested?

Tested locally via `ProducerTool`. Modified unit tests to verify the behavior of the online producer after the constructor argument change


clients/venice-producer/src/main/java/com/linkedin/venice/producer/online/OnlineVeniceProducer.java
```diff
@@ -42,26 +44,50 @@ public class OnlineVeniceProducer<K, V> extends AbstractVeniceProducer<K, V> {
...
+        .setStoreName(AvroProtocolDefinition.KAFKA_MESSAGE_ENVELOPE.getSystemStoreName())
+        .setSpecificValueClass(KafkaMessageEnvelope.class);
+
+    try (SchemaReader kmeSchemaReader = ClientFactory.getSchemaReader(kmeClientConfig, icProvider)) {
+      configure(storeName, producerConfigs, metricsRepository, schemaReader, kmeSchemaReader);
+    } catch (Throwable e) {
+      // When using D2TransportClient, the threads that D2 creates are non-daemon threads which hold up graceful
+      // termination of the "main" thread. This is especially problematic for the shell producer.
+      Utils.closeQuietlyWithErrorLogged(this);
+      if (e instanceof RuntimeException) {
```
majisourav99, line 85: VeniceException is extended from RTE, so this will hit true to all VeniceException which might be thrown from getSchemaReader.  do you want to cast those?

clients/venice-producer/src/main/java/com/linkedin/venice/producer/online/OnlineVeniceProducer.java
```diff
@@ -42,26 +44,50 @@ public class OnlineVeniceProducer<K, V> extends AbstractVeniceProducer<K, V> {
...
+        ClientConfig.cloneConfig(storeClientConfig).setSchemaRefreshPeriod(schemaRefreshPeriod);
+
+    this.schemaReader = ClientFactory.getSchemaReader(clientConfigForSchemaReader, icProvider);
+    ClientConfig<KafkaMessageEnvelope> kmeClientConfig = ClientConfig.cloneConfig(storeClientConfig)
+        .setStoreName(AvroProtocolDefinition.KAFKA_MESSAGE_ENVELOPE.getSystemStoreName())
+        .setSpecificValueClass(KafkaMessageEnvelope.class);
+
+    try (SchemaReader kmeSchemaReader = ClientFactory.getSchemaReader(kmeClientConfig, icProvider)) {
+      configure(storeName, producerConfigs, metricsRepository, schemaReader, kmeSchemaReader);
+    } catch (Throwable e) {
```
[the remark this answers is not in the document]
nisargthakkar: Yeah, I can do that

clients/venice-producer/src/main/java/com/linkedin/venice/producer/online/OnlineVeniceProducer.java
nisargthakkar: Yes. The idea here is to throw all runtime exceptions directly. Casting VeniceException to RuntimeException doesn't change it's type. Only how the compiler sees it. Ideally, I wanted to do:
```
if (e instanceof RuntimeException) {
    throw e;
}
```

But the compiler doesn't like it as it needs to be explicitly told the type. Anything that is not a `RuntimeException` cannot be thrown unchecked. So, we wrap it in a `VeniceException` to make it a `RuntimeException`

gaojieliu (approved):

nisargthakkar:
Thanks, Gaojie and Sourav

PR #613, opened 2023-08-31
[controller] Check system store health and issue auto repair in Venice parent controller

sixpluszero:
## [controller] Check system store health and issue auto repair in Venice parent controller

This PR changes previous design and execute the system store health check and repair in Venice parent controller only.
It will issue controller client requests to send and retrieve heartbeats in child controller in different regions.
For repair, it will utilize existing helper method to empty push to system store. 

## How was this PR tested?
Added new unit test for functionalities 


services/venice-controller/src/main/java/com/linkedin/venice/controller/UserSystemStoreLifeCycleHelper.java
```diff
@@ -69,39 +69,49 @@ public List<VeniceSystemStoreType> materializeSystemStoresForUserStore(String cl
...
-            null);
-      }
-      parentAdmin.writeEndOfPush(clusterName, systemStoreName, version.getNumber(), true);
+      materializeSystemStore(parentAdmin, clusterName, systemStoreName, pushJobId);
       createdSystemStoreTypes.add(systemStoreType);
     }
     return createdSystemStoreTypes;
   }
 
+  public static Version materializeSystemStore(
```
sixpluszero, line 78: @gaojieliu this is the class that requires VeniceParentHelixAdmin

services/venice-controller/src/main/java/com/linkedin/venice/controller/systemstore/SystemStoreRepairTask.java
```diff
@@ -0,0 +1,236 @@
...
+      AtomicBoolean isRunning) {
+    this.parentAdmin = parentAdmin;
+    this.maxRepairRetry = maxRepairRetry;
+    this.heartbeatWaitTimeSeconds = heartbeatWaitTimeSeconds;
+    this.isRunning = isRunning;
+  }
+
+  @Override
+  public void run() {
+    for (String clusterName: parentAdmin.getClustersLeaderOf()) {
```
sixpluszero, line 51: @gaojieliu And here we are iterating through cluster. Even though we will check if it is the leader controller in the internal task, I still think it will be great to separate them into one task per cluster. What do you think?

services/venice-controller/src/main/java/com/linkedin/venice/controller/HelixVeniceClusterResources.java
```diff
@@ -208,26 +206,6 @@ public HelixVeniceClusterResources(
     veniceAdminStats = new VeniceAdminStats(metricsRepository, "venice-admin-" + clusterName);
     this.storagePersonaRepository =
         new StoragePersonaRepository(clusterName, this.storeMetadataRepository, adapterSerializer, zkClient);
-    if (!config.isParent() && config.isSystemStoreHealthCheckEnabled()) {
```
sixpluszero, line 211: @gaojieliu Originally I thought we can make the change here, but considering the class is spinned up inside VeniceHelixAdmin, instead of VeniceParentHelixAdmin, repair task is unlikely to be created here, which requires parent admin

sixpluszero (commented):
Left some pointer for the issue I was talking about in Slack thread.

services/venice-controller/src/main/java/com/linkedin/venice/controller/systemstore/SystemStoreRepairTask.java
sixpluszero: StoreBackupVersionCleanupService but I took another look at other service in controller. Looks like it is OK to for loop over cluster.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerConfig.java
```diff
@@ -280,9 +280,9 @@ public class VeniceControllerConfig extends VeniceControllerClusterConfig {
 
   private final int storeGraveyardCleanupSleepIntervalBetweenListFetchMinutes;
 
-  private final boolean systemStoreHealthCheckEnabled;
+  private final boolean systemStoreRepairServiceEnabled;
 
-  private final int systemStoreHealthCheckIntervalSeconds;
+  private final int systemStoreRepairCheckIntervalHours;
```
gaojieliu, line 285: I think `hour` as the config unit is too large, and we should continue to use `seconds`.
For example, if there is a problem with meta system store, repairing it after 6 hours will still trigger a GCN.
DaVinci is still using meta system store today.

services/venice-controller/src/main/java/com/linkedin/venice/controller/systemstore/SystemStoreRepairService.java
```diff
@@ -0,0 +1,71 @@
...
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicLong;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+
+public class SystemStoreRepairService extends AbstractVeniceService {
```
gaojieliu, line 16: Let us leave some javadoc to describe the mechanism.

services/venice-controller/src/main/java/com/linkedin/venice/controller/HelixVeniceClusterResources.java
gaojieliu: An alternative is to leverage `ClusterLeaderInitializationRoutine` to register a callback from parent admin.
But I think the way this PR chooses is fine as well since `VeniceParentHelixAdmin#isLeaderControllerFor` is a cheap operation.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -7824,6 +7826,38 @@ public Optional<PushStatusStoreWriter> getPushStatusStoreWriter() {
     return pushStatusStoreWriter;
   }
 
+  @Override
+  public void sendHeartbeatToSystemStore(String clusterName, String storeName, long heartbeatTimeStamp) {
+    VeniceSystemStoreType systemStoreType = VeniceSystemStoreType.getSystemStoreType(storeName);
```
gaojieliu, line 7831: `VeniceSystemStoreType.getSystemStoreType` can return `BATCH_JOB_HEARTBEAT_STORE` or `null` if there is no match.
To guard these cases, we should do a `null` check, and only send heartbeat for `DAVINCI_PUSH_STATUS_STORE` and `META_STORE`, and for `  BATCH_JOB_HEARTBEAT_STORE`, this function should fail fast.

This comment applies to function: `getHeartbeatFromSystemStore` as well.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -7824,6 +7826,38 @@ public Optional<PushStatusStoreWriter> getPushStatusStoreWriter() {
...
+        if (systemStoreType == VeniceSystemStoreType.DAVINCI_PUSH_STATUS_STORE) {
+          retrievedTimestamp = getPushStatusStoreReader().get()
+              .getHeartbeat(userStoreName, PushStatusStoreUtils.CONTROLLER_HEARTBEAT_INSTANCE_NAME);
+        } else {
+          retrievedTimestamp = getMetaStoreReader().getHeartbeat(userStoreName);
+        }
+        return retrievedTimestamp;
+      }, 3, Duration.ofSeconds(1), Collections.singletonList(VeniceException.class));
+    } catch (VeniceException e) {
+      return 0;
```
gaojieliu, line 7857: Should we return the error response to Parent?
For example, we should return `-1` when error happens, and Parent Controller shouldn't try to repair on error response, but log it and maybe emit a metric in Parent Controller.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -7824,6 +7826,38 @@ public Optional<PushStatusStoreWriter> getPushStatusStoreWriter() {
     return pushStatusStoreWriter;
   }
 
+  @Override
+  public void sendHeartbeatToSystemStore(String clusterName, String storeName, long heartbeatTimeStamp) {
```
gaojieliu, line 7830: Do we need to check whether the push status store is enabled or not in these functions?

services/venice-controller/src/main/java/com/linkedin/venice/controller/server/StoresRoutes.java
```diff
@@ -1001,4 +1003,61 @@ public Route removeStoreFromGraveyard(Admin admin) {
...
+          responseObject.setError(ACL_CHECK_FAILURE_WARN_MESSAGE_PREFIX + request.url());
+          responseObject.setErrorType(ErrorType.BAD_REQUEST);
+          return AdminSparkServer.OBJECT_MAPPER.writeValueAsString(responseObject);
+        }
+        AdminSparkServer.validateParams(request, SEND_HEARTBEAT_TIMESTAMP.getParams(), admin);
+        String clusterName = request.queryParams(CLUSTER);
+        String storeName = request.queryParams(NAME);
+        responseObject.setCluster(clusterName);
+        responseObject.setName(storeName);
+        // admin.sendHeartbeatToSystemStore(clusterName, storeName);
```
gaojieliu, line 1023: Why comment out this line?

services/venice-controller/src/main/java/com/linkedin/venice/controller/server/StoresRoutes.java
```diff
@@ -1001,4 +1003,61 @@ public Route removeStoreFromGraveyard(Admin admin) {
...
```
gaojieliu, line 1051: This is not correct, right?

services/venice-controller/src/main/java/com/linkedin/venice/controller/server/StoresRoutes.java
```diff
@@ -1001,4 +1003,61 @@ public Route removeStoreFromGraveyard(Admin admin) {
...
+      HeartbeatResponse responseObject = new HeartbeatResponse();
+      response.type(HttpConstants.JSON);
+      try {
+        if (!isAllowListUser(request)) {
+          response.status(HttpStatus.SC_FORBIDDEN);
+          responseObject.setError(ACL_CHECK_FAILURE_WARN_MESSAGE_PREFIX + request.url());
+          responseObject.setErrorType(ErrorType.BAD_REQUEST);
+          return AdminSparkServer.OBJECT_MAPPER.writeValueAsString(responseObject);
+        }
+        AdminSparkServer.validateParams(request, SEND_HEARTBEAT_TIMESTAMP.getParams(), admin);
```
gaojieliu, line 1046: `SEND_HEARTBEAT_TIMESTAMP` is not right.

services/venice-controller/src/main/java/com/linkedin/venice/controller/systemstore/SystemStoreRepairTask.java
```diff
@@ -0,0 +1,255 @@
...
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicLong;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+
+public class SystemStoreRepairTask implements Runnable {
```
gaojieliu, line 24: Let us add some javadoc for this class.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
```diff
@@ -508,6 +510,16 @@ public VeniceParentHelixAdmin(
         initRoutineForHeartbeatSystemStore.setAllowEmptyDelegateInitializationToSucceed();
       }
     }
+    if (multiClusterConfigs.getCommonConfig().isSystemStoreRepairServiceEnabled()) {
+      this.systemStoreRepairService = new SystemStoreRepairService(
+          this,
+          multiClusterConfigs.getCommonConfig().getSystemStoreRepairCheckIntervalHours(),
+          1, // Set it to 1 for now. It could also be controlled by a config if necessary.
```
gaojieliu, line 517: Let us make it configurable.

services/venice-controller/src/main/java/com/linkedin/venice/controller/systemstore/SystemStoreRepairTask.java
```diff
@@ -0,0 +1,255 @@
...
+      String clusterName,
+      Set<String> newUnhealthySystemStoreSet,
+      Map<String, Long> systemStoreToHeartbeatTimestampMap) {
+    for (Store store: getParentAdmin().getAllStores(clusterName)) {
+      if (!shouldContinue(clusterName)) {
+        return;
+      }
+
+      // For user store, if corresponding system store flag is not true, it indicates system store is not created.
+      if (!VeniceSystemStoreUtils.isUserSystemStore(store.getName())) {
```
gaojieliu, line 117: For the regular Venice store creation, system stores will create right after, which means there is a gap in between.
Shall we check the store creation time before repairing the system stores since it is possible that the system store creation itself would take some time.

services/venice-controller/src/main/java/com/linkedin/venice/controller/systemstore/SystemStoreRepairTask.java
```diff
@@ -0,0 +1,255 @@
...
+    }
+  }
+
+  boolean isHeartbeatReceivedBySystemStore(
+      String clusterName,
+      String systemStoreName,
+      long expectedHeartbeatTimestamp) {
+    for (Map.Entry<String, ControllerClient> entry: getControllerClientMap(clusterName).entrySet()) {
+      long heartbeatFromChildController =
+          entry.getValue().getHeartbeatFromSystemStore(systemStoreName).getHeartbeatTimestamp();
```
gaojieliu, line 198: I think we should check whether the response contains any error or not.
If the response is errored, we shouldn't mark heart beat failure since it might indicate some serious issue, which we need to fix manually.
Let us also emit a metric to track error responses.

gaojieliu (commented):
It seems this PR is WIP, and I will review it once it is finished.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerConfig.java
sixpluszero: But repairing for once could take up to 1h if the push is stuck. 
If the previous run is not complete, we should wait until next run right? So I think in theory it should be at least in hour level...

gaojieliu: Were you referring to push job? 
The issue can happen to both DVC push system store and meta system store.

Here is an example:
1. User adds a new value schema.
2. User starts using this schema in. the nearline job.
3. DaVinci could receive a message encoded by the newly registered schema.
4. If meta system store is broken, DaVinci app won't be able to decode the payload.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
sixpluszero: I think this comment relates to the below comment right? So maybe just return -1 and check it and emit the error metrics.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
sixpluszero: I think instead of checking them here, let me add another safety check in the 
checkAndSendHeartbeatToSystemStores() which has iterations over stores. 

I refactored that method a bit to contain the check. So it will first check if it is a system store, and then check if it is a user system store.

services/venice-controller/src/main/java/com/linkedin/venice/controller/systemstore/SystemStoreRepairTask.java
sixpluszero: Added to track unreachableSystemStore.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
sixpluszero: Let me leave a comment there indicating that PushStatusStore is fully rolled out. We should clean up these configs and move optional to required arguments.

services/venice-controller/src/main/java/com/linkedin/venice/controller/systemstore/SystemStoreRepairTask.java
```diff
@@ -0,0 +1,355 @@
...
+  }
+
+  /**
+   * Poll the system store push status until it reaches terminal status.
+   * Based on system store setup, stuck job should time out in 1h.
+   * If push job completes, it will return true, otherwise return false.
+   */
+  boolean pollSystemStorePushStatusUntilCompleted(String clusterName, String systemStoreName, int version) {
+    String kafkaTopic = Version.composeKafkaTopic(systemStoreName, version);
+    while (true) {
```
gaojieliu, line 301: Shall we put a upper limit here otherwise this poll function can stuck forever in some scenario?

gaojieliu (changes requested):
Overall looks good, and we need to careful monitor the test env once this feature is enabled.

services/venice-controller/src/main/java/com/linkedin/venice/controller/server/StoresRoutes.java
```diff
@@ -1021,4 +1025,62 @@ public Route removeStoreFromGraveyard(Admin admin) {
...
+          return AdminSparkServer.OBJECT_MAPPER.writeValueAsString(responseObject);
+        }
+        AdminSparkServer.validateParams(request, SEND_HEARTBEAT_TIMESTAMP_TO_SYSTEM_STORE.getParams(), admin);
+        String clusterName = request.queryParams(CLUSTER);
+        String storeName = request.queryParams(NAME);
+        long heartbeatTimestamp = Long.parseLong(request.queryParams(HEARTBEAT_TIMESTAMP));
+        responseObject.setCluster(clusterName);
+        responseObject.setName(storeName);
+        admin.sendHeartbeatToSystemStore(clusterName, storeName, heartbeatTimestamp);
+      } catch (ResourceStillExistsException exception) {
```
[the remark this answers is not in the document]
sixpluszero: Hmm it might be some wrong auto complete. Let me check...

services/venice-controller/src/main/java/com/linkedin/venice/controller/systemstore/SystemStoreRepairTask.java
```diff
@@ -0,0 +1,355 @@
...
+          LOGGER.info("System store: {} in cluster: {} has been fixed by repair job.", entry.getKey(), clusterName);
+        }
+      }
+    }
+    Set<String> notRepairableSystemStoreSet = new HashSet<>(unhealthySystemStoreSet);
+    notRepairableSystemStoreSet.removeAll(unreachableSystemStoreSet);
+    // After repairing system stores, update stats again.
+    updateBadSystemStoreCount(clusterName, unhealthySystemStoreSet);
+    updateUnreachableSystemStoreCount(clusterName, unreachableSystemStoreSet);
+    updateNotRepairableSystemStoreCount(clusterName, notRepairableSystemStoreSet);
```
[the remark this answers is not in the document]
sixpluszero: I thought we see the bad system store going down should be suffice?

services/venice-controller/src/main/java/com/linkedin/venice/controller/server/StoresRoutes.java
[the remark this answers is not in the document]
sixpluszero: Oh apparently I copied the method from another method which does check this. Let me fix it.

services/venice-controller/src/main/java/com/linkedin/venice/controller/systemstore/SystemStoreRepairTask.java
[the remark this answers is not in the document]
sixpluszero: I think it is also ok to add one, based on the agreement that we don't have too many metric in controller and it is not promotional to store number.

services/venice-controller/src/main/java/com/linkedin/venice/controller/systemstore/SystemStoreRepairTask.java
```diff
@@ -0,0 +1,355 @@
...
+      repairBadSystemStore(clusterName, unhealthySystemStoreSet, unreachableSystemStoreSet, maxRepairRetry);
+    }
+  }
+
+  void repairBadSystemStore(
+      String clusterName,
+      Set<String> unhealthySystemStoreSet,
+      Set<String> unreachableSystemStoreSet,
+      int maxRepairRetry) {
+    for (int i = 0; i < maxRepairRetry; i++) {
```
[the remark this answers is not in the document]
sixpluszero: Sure, in one iteration if a store cannot be fixed within maxRetry, we should look into it (oncall) as it must have some problem in the cluster.

services/venice-controller/src/main/java/com/linkedin/venice/controller/systemstore/SystemStoreRepairTask.java
sixpluszero: My understanding is if it is stuck, after one hour it should timeout right? Or are you suggesting we add extra safeguard here?

services/venice-controller/src/main/java/com/linkedin/venice/controller/systemstore/SystemStoreRepairTask.java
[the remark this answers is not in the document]
sixpluszero: But, I think a bit more, how do you represent the number in the metric?
Let's say we have 5 bad stores, and we fixed 3 stores, and 2 are not fixable. I think we will see 5->2 in bad system store metrics(for meta / push status respectively), we will see 2 in not_fixable metric, but the 3 should not be a long standing number in the metric right? It is not a persistent state.

services/venice-controller/src/main/java/com/linkedin/venice/controller/systemstore/SystemStoreRepairTask.java
gaojieliu: Let us keep this for now.

services/venice-controller/src/main/java/com/linkedin/venice/controller/systemstore/SystemStoreRepairTask.java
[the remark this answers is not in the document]
gaojieliu: Yeah, that is fine then.

gaojieliu (approved):
Thanks for the fix!

PR #615, opened 2023-08-31
[fast-client][server] Bump io.grpc down to 1.49.2

vshwnth2:
## Bump io.grpc down to 1.49.2

Due to backwards incompatible changes, only versions up to 1.49.2 will be supported for now

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -40,7 +40,7 @@ if (project.hasProperty('overrideBuildEnvironment')) {
 
 def avroVersion = '1.10.2'
 def avroUtilVersion = '0.2.150'
-def grpcVersion = '1.54.1'
+def grpcVersion = '1.49.2'
 def kafkaGroup = 'com.linkedin.kafka'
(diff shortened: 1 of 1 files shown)


ZacAttack (commented):
Is there some information somewhere that explains the compatibility issue?  It'd be a shame to have to be pinned to a version in perpetuity, maybe if we understand the gotcha we can take steps to insure compatibility (kind of like what we do with the avro versions).

FelixGV:
Some extra context on this...

>Although Google has the version policy to not make incompatible changes with minor version upgrades, we have found [breaking changes](https://github.com/protocolbuffers/protobuf/commit/c440da9e1389db520b79acb19cb55e5b3266dfba) starting with com.google.protobuf:protobuf-java:3.23.*.

nisargthakkar (approved):
Thanks for the contribution @vshwnth2 . When you merge, can you ensure that you merge with the updated PR title as the main commit message?

PR #616, opened 2023-08-31
[server] Update stalled leader metric to only emit after EOP

ZacAttack:
## [server] Update stalled leader metric to only emit after EOP
Metric is spurious when an alert is ongoing, tweaking the method call assures that we only monitor on those partitions which have received EOP

Resolves #XXX


services/venice-server/src/test/java/com/linkedin/venice/stats/IngestionStatsTest.java
adamxchen, line 1: Why do we still keep this file?

clients/da-vinci-client/src/test/java/com/linkedin/davinci/utils/IngestionStatsTest.java
```diff
@@ -0,0 +1,38 @@
+package com.linkedin.davinci.utils;
+
+import static com.linkedin.venice.ConfigKeys.*;
```
adamxchen, line 3: nit: not import all.

adamxchen (approved):
Thank you for catching and fixing it!

PR #617, opened 2023-08-31
[server][da-vinci-client] Add leaked resource cleaner thread to II server

majisourav99:
## Add leaked resource cleaner thread to isolated ingestion server

We recently added code to delete leaked version SST files in II mode. But it will not delete metadata partitions lingering behind. There could be other edge cases when venice leave leaked version in II. This PR adds `LeakedResourceCleaner` background thread in II to cleanup left over resources.

## How was this PR tested?

GHCI


clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/isolated/IsolatedIngestionServer.java
```diff
@@ -769,9 +774,17 @@ private void initializeIsolatedIngestionServer() {
...
-        configLoader.getVeniceServerConfig().getIngestionApplicationPort());
+    if (serverConfig.isLeakedResourceCleanupEnabled()) {
+      LeakedResourceCleaner leakedResourceCleaner = new LeakedResourceCleaner(
+          storageService.getStorageEngineRepository(),
+          serverConfig.getLeakedResourceCleanUpIntervalInMS(),
+          storeRepository,
+          storeIngestionService,
+          storageService,
+          metricsRepository);
+      leakedResourceCleaner.start();
```
sixpluszero, line 785: We will also need to shut it down properly

clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/isolated/IsolatedIngestionServer.java
```diff
@@ -769,9 +774,17 @@ private void initializeIsolatedIngestionServer() {
     storeIngestionService.addIngestionNotifier(new IsolatedIngestionNotifier(this));
     ingestionBackend = new DefaultIngestionBackend(storageMetadataService, storeIngestionService, storageService);
 
-    LOGGER.info(
-        "Starting report client with target application port: {}",
-        configLoader.getVeniceServerConfig().getIngestionApplicationPort());
+    if (serverConfig.isLeakedResourceCleanupEnabled()) {
```
sixpluszero, line 777: Also, it suddenly comes to me that -> Is Helix guaranteed to issue DROP message for old resources if a node is down for enough time? For now we can assume that, but if this is not the case then the fix in main process to drop resource if NOT_EXIST is not enough I guess..(then we will need to restore here, and try to rely on LeakedResourceCleaner to clean old storage engine). It is my concern, but I might be wrong though.

sixpluszero: Just my thinking. If Helix is guaranteed to send DROP as I mentioned above, then data partition is guaranteed to be dropped and this PR will guarantee metadata partition will also be deleted, that's good enough. 

The only lingering would be existing leaks. My take is it is OK for now.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/isolated/IsolatedIngestionServer.java
sixpluszero: And you will unfortunately have to play with unit test coverage, which I don't know what to add to... you might want to game the system for this case.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/isolated/IsolatedIngestionServer.java
majisourav99: Thanks for the review. I have verified with Helix team, if a node goes down for a period longer than rebalance delay period, Helix will select a new replica and when the node comes up online later it will see drop messages. So our main process deletion PR should work.

sixpluszero: Okay. I see. if this is true, then it should be good. If we found out that it does not work as Helix said, we should let them know and try to go back to original proposal (restore in II and rely on cleaner to clean up).

sixpluszero (approved):
Thanks for catching this issue and get the fix done!

PR #618, opened 2023-09-01
[vpj] Fix regression in partial update + incremental push input schema detection

sixpluszero:
## [vpj] Fix regression in partial update + incremental push input schema detection
Fix the regression in checking whether input schema is a update schema or not.
Previously it is using NO_OP_ON_FIELD.name(), while it should use .getName() instead to return "NoOp"

## How was this PR tested?
New unit test and new integration test.


gaojieliu:
We need to have an integration test to test inc push + write compute with the new format and the existing format (AKA, using write compute schema).

sixpluszero:
> We need to have an integration test to test inc push + write compute with the new format and the existing format (AKA, using write compute schema).

They are both here now. 
This PR adds the integration test for old format, in the previous PR that adds new format support, I already added integration test for new input format.

FelixGV (approved):
Thank you!

gaojieliu (approved):
Thanks!

PR #619, opened 2023-09-01
[da-vinci][server] Fixed a small perf regression when handling compute request

gaojieliu:
In a previous schema optimization, Venice Server/DaVinci could pass an inconsistent schema id from the reused record when decoding the binary payload from the database.
When mismatch happens, object reuse won't work, and Avro deserializer will create a new object for every call.
For Server, this could happen in the following setup:
1. The user evolves the value schema, which would also trigger a new superset schema.
2. The new value/superset schema won't be used by any data entries.
3. Client will continue to use the locally-known latest value schema.
4. The mismatch will happen in Venice Server when handling such compute request.

For DaVinci, the chance of mismatch is very low since the request will be executed locally, and that is the reason this PR uses a lightweight change to make schema & schema id consistent when serving compute request.

This code change also fixed a regression introduced by client compute support, and
it will handle the scenario if an error happens before handling any header/data received
from the backend. Currently, this seems only happening in integration test, which is
using `HttpTransportClient`, where the streaming impl is different from `D2TransportClient`.

This code change also enhances a test utility: `ErrorCountAppender` to handle an error log without a valid throwable or a throwable without a valid message.

## How was this PR tested?
CI

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -415,7 +415,7 @@ subprojects {
     }
 
     retry {
-      maxRetries = 1 // 5 attempts in total
+      maxRetries = 4 // 5 attempts in total
       maxFailures = 100
       failOnPassedAfterRetry = false
     }
--- a/build.gradle
+++ b/build.gradle
@@ -415,7 +415,7 @@ subprojects {
     }
 
     retry {
-      maxRetries = 4 // 5 attempts in total
+      maxRetries = 1 // 5 attempts in total
       maxFailures = 100
       failOnPassedAfterRetry = false
     }
--- a/services/venice-server/src/test/java/com/linkedin/venice/listener/ErrorCountAppender.java
+++ b/services/venice-server/src/test/java/com/linkedin/venice/listener/ErrorCountAppender.java
@@ -57,7 +57,9 @@ public class ErrorCountAppender extends AbstractAppender {
 
   @Override
   public void append(LogEvent event) {
-    if (event.getLevel().equals(Level.ERROR) && event.getThrown().getLocalizedMessage().contains(exceptionMessage)) {
+    if (event.getLevel().equals(Level.ERROR) && event.getThrown() != null
+        && event.getThrown().getLocalizedMessage() != null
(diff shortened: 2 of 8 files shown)


services/venice-server/src/test/java/com/linkedin/venice/listener/ErrorCountAppender.java
```diff
@@ -57,7 +57,9 @@ public ErrorCountAppender build() {
 
   @Override
   public void append(LogEvent event) {
-    if (event.getLevel().equals(Level.ERROR) && event.getThrown().getLocalizedMessage().contains(exceptionMessage)) {
+    if (event.getLevel().equals(Level.ERROR) && event.getThrown() != null
```
FelixGV, line 60: Could you mention this change also in the commit message?

FelixGV (commented):
Thanks Gaojie! The change LGTM... but there is an integration test failure, so we should look at that first...

gaojieliu:
> Thanks Gaojie! The change LGTM... but there is an integration test failure, so we should look at that first...
Yes, I found a regression in client compute, which seems not serious in prod, but troublesome in integration test since integration test is normally using http client instead of d2 and http client's streaming impl is different from d2 client.
Fixed it in the latest code .

services/venice-server/src/test/java/com/linkedin/venice/listener/ErrorCountAppender.java
gaojieliu: Updated.

build.gradle
```diff
@@ -415,7 +415,7 @@ subprojects {
     }
 
     retry {
-      maxRetries = 4 // 5 attempts in total
```
majisourav99, line 418: not part of the change?

gaojieliu: yeah, I missed it, and I will revert it.

FelixGV (approved):
Thanks!

majisourav99 (approved):
Thanks Gaojie. LGTM

PR #620, opened 2023-09-06
[WIP][compat][server][common] Leader hybrid offset lag fix

xunyin8:
## [WIP][compat][server][common] Leader hybrid offset lag fix
Introduced new control message SyncOffset. The purpose of the new CM is to allow leader replica update its latest processed upstream RT offset and calculate lag correctly when the tail of the RT is filled with SOS and EOS CMs. 

In addition, this will unblock the rollout of A/A lag calculation since it will address the edge case where all the ending
messages of a remote RT topic are ignored due to DCR. Currently this will produce a positive lag when the consumption
of that remote fabric is actually up-to-date.

Here is the new flow, upon consuming SOS, EOS CMs or skipped records from upstream RT as leader the following will
happen:
1. Leader will update the in-memory latest consumed upstream RT offset.
2. The leader will periodically produce the SyncOffset CM containing a map of KafkaURLs and its corresponding latest consumed upstream RT offset (works for both L/F and A/A). 
3. Upon processing the CM in the drainer the consumer will only update its latest processed leader upstream RT offset
if the offset in the CM is greater.

## How was this PR tested?
TODO

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
@@ -167,6 +167,8 @@ public class LeaderFollowerStoreIngestionTask extends StoreIngestionTask {
 
   protected final AvroStoreDeserializerCache storeDeserializerCache;
 
+  private long lastSyncOffsetTimestamp = 0;
+
   public LeaderFollowerStoreIngestionTask(
       StoreIngestionTaskFactory.Builder builder,
       Store store,
@@ -2077,40 +2079,13 @@ public class LeaderFollowerStoreIngestionTask extends StoreIngestionTask {
                *   work here.
                *
                * To address #2 we will periodically produce {@link com.linkedin.venice.kafka.protocol.SyncOffset}
-               * control message to VT and upon consumption of those messages in the drainer we will update the offset.
+               * control message to VT and upon consumption of those control messages in the drainer we will update the
+               * offset accordingly.
                *
-               * So in summary NO further processing is needed SOS/EOS received from RT topics. Just silently drop the message here.
+               * So in summary NO further processing is needed SOS/EOS received from RT topics. Just silently drop the
+               * message here.
                * We should not return false here.
                */
-              if (consumerRecord.getTopicPartition().getPubSubTopic().isRealTime()) {
-                long timestamp = System.currentTimeMillis();
-                long previousSyncOffsetTimestamp = partitionConsumptionState.getLatestRTSyncOffsetTimestamp(kafkaUrl);
-                if (timestamp > previousSyncOffsetTimestamp + getServerConfig().getSyncOffsetMinIntervalMs()) {
-                  ControlMessage syncOffsetCM = new ControlMessage();
-                  syncOffsetCM.controlMessageType = ControlMessageType.SYNC_OFFSET.getValue();
-                  SyncOffset syncOffset = new SyncOffset();
-                  syncOffset.upstreamOffset = consumerRecord.getOffset();
-                  syncOffset.kafkaURL = kafkaUrl;
-                  syncOffsetCM.controlMessageUnion = syncOffset;
-                  int versionTopicPartition =
-                      consumerRecord.getTopicPartition().getPartitionNumber() * amplificationFactor;
-                  produceToLocalKafka(
-                      consumerRecord,
-                      partitionConsumptionState,
-                      leaderProducedRecordContext,
-                      (callback, leaderMetadataWrapper) -> veniceWriter.get()
-                          .asyncSendControlMessage(
-                              syncOffsetCM,
-                              versionTopicPartition,
-                              new HashMap<>(),
-                              callback,
-                              leaderMetadataWrapper),
-                      subPartition,
-                      kafkaUrl,
-                      kafkaClusterId,
-                      beforeProcessingRecordTimestampNs);
-                }
-              }
               producedFinally = false;
             }
             break;
@@ -3110,20 +3085,13 @@ public class LeaderFollowerStoreIngestionTask extends StoreIngestionTask {
     }
     long lag = lastOffsetInRealTimeTopic - latestLeaderOffset;
 
-    // Here we handle the case where the topic is actually empty,
-    // if consumerLag is a positive number then that means there is an existing offset that we've consumed to and a
-    // bigger offset out there somewhere. Meaning that there are at least two messages in the topic and it's not empty
-    long consumerLag = getPartitionOffsetLag(sourceRealTimeTopicKafkaURL, leaderTopic, partitionToGetLatestOffsetFor);
-    if (consumerLag <= 0) {
-      // We don't have a positive consumer lag, but this could be because we haven't polled.
-      // So as a final check to determine if the topic is empty, we check
-      // if the end offset is the same as the beginning
-      long earliestOffset = cachedPubSubMetadataGetter.getEarliestOffset(
-          getTopicManager(sourceRealTimeTopicKafkaURL),
-          new PubSubTopicPartitionImpl(leaderTopic, partitionToGetLatestOffsetFor));
-      if (earliestOffset == lastOffsetInRealTimeTopic - 1) {
-        lag = 0;
-      }
+    // Here we handle the case where the topic is actually empty, we check
+    // if the end offset is the same as the beginning
+    long earliestOffset = cachedPubSubMetadataGetter.getEarliestOffset(
+        getTopicManager(sourceRealTimeTopicKafkaURL),
+        new PubSubTopicPartitionImpl(leaderTopic, partitionToGetLatestOffsetFor));
+    if (earliestOffset == lastOffsetInRealTimeTopic - 1) {
+      lag = 0;
     }
 
     if (shouldLog) {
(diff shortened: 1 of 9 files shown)


xunyin8:
For protocol upgrade like this that introduces new CM. I think we need to have a config to control when to start producing the new CM since once the new CM is in the VT. Components with older code cannot process it since our default behavior right now for UNKNOWN CM type is to throw exception.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -2075,9 +2076,41 @@ protected DelegateConsumerRecordResult delegateConsumerRecord(
                * For #3 stat counter update will not happen for SOS/EOS message. This should not be a big issue. If needed we can copy some of the stats maintenance
                *   work here.
                *
+               * To address #2 we will periodically produce {@link com.linkedin.venice.kafka.protocol.SyncOffset}
+               * control message to VT and upon consumption of those messages in the drainer we will update the offset.
+               *
                * So in summary NO further processing is needed SOS/EOS received from RT topics. Just silently drop the message here.
                * We should not return false here.
                */
+              if (consumerRecord.getTopicPartition().getPubSubTopic().isRealTime()) {
```
huangminchn, line 2085: This approach is still subject to the requirements that there is data in RT, or we happen to produce SyncOffset at the end of RT.

I wonder if we can find a way to produce SyncOffset regularly without depending on RT, and without introducing race condition (or the readers of this CM is robust to such race).

Could you please check if producer is thread safe? My hunch is yes. If so, can we produce this CM regularly in this function: `checkIngestionProgress`? One potential issue is that the upstream offset produced in this function could be stale; one potential case:
1. share consumer thread produce a data message with upstream 100;
2. ingestion thread get the latest consumed RT offset in function `checkIngestionProgress` and think the current value is 100;
3. before the ingestion thread is able to send out SyncOffset, share consumer thread sends out a new data message with upstream offset 101
4. ingestion thread send out SyncOffset with upstream offset 100
When it happens, the reader of SyncOffset just need to ignore offset 100 since it's smaller than the checkpoint upstream offset 101 which is correct.

Please help double check if it makes sense. The drawback of keeping the regularly producing logic here is that - let's say the end offset of RT is 102, the last 2 messages are EOS, and consumer threads already produce a SyncOffset CM at upstream offset 100, then we will end up in the same situation, since it wouldn't take longer than 15 minutes to process the rest 2 messages.

@sushantmane Could you please also double check here? Thanks!

huangminchn (changes requested):
Thanks a lot Xun!

Could you please also mention in the PR description that this change can fix an edge case in AA that when all the ending messages of RT from a colo are ignored, lag would stay high for that colo; and this edge case is the blocker for us to roll out the ACTIVE_ACTIVE ready-to-serve algorithm which checks all RT lags.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
xunyin8: I think this is doable, will try to implement it this way and control the behavior via a server config for smooth rollout.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -3118,6 +3120,76 @@ protected void processVersionSwapMessage(
...
 
+  @Override
+  protected void processSyncOffsetMessage(
+      ControlMessage controlMessage,
+      int partition,
+      PartitionConsumptionState partitionConsumptionState) {
+    // Update the latest processed upstream offset in memory.
+    OffsetRecord offsetRecord = partitionConsumptionState.getOffsetRecord();
+    PubSubTopic upstreamTopic = offsetRecord.getLeaderTopic(pubSubTopicRepository);
+    if (upstreamTopic != null && upstreamTopic.isRealTime()) {
```
xunyin8, line 3131: Hi @huangminchn, I'm wondering if this upstream topic check is necessary for consumers of the SyncOffset CM.

FelixGV (commented):
Thanks for getting started on this Xun! Cool stuff...

A few comments:

1. This is similar to another idea we had earlier, which is to have a CM carry the leader's RT DIV state to the follower. Essentially, serialize the `PartitionState` (or some parts of it, at least...) and pack that into the CM. This would allow the follower to keep track of where the leader is at in the RT consumption, and if the follower becomes leader, it could restore that `PartitionState` it got from the last such CM, subscribe to the RT at that offset, and begin consuming from there. This would mean the follower to leader transition has no DIV gap at all. The `PartitionState` contains the `upstreamOffsetMap`, which I think is the same content that you intend to pass into the new CM.
2. Because the VT is consumed by DVC, adding a new CM type could be a bit risky (it'll throw a `UnsupportedMessageTypeException`...). So I think it would be ideal to NOT introduce a new CM here, actually, but to add a new field to the EOS CM containing the above info. Then the leader server could periodically do a EOS/SOS, and the DVC instances and non-upgraded followers that don't care about the new EOS field could safely ignore it, simply leveraging the already-existing KME forward compat.
3. If the EOS start packing a lot more content, and we periodically do a EOS/SOS, then they could add up to a non-trivial amount amount of Kafka space. I think that's not ideal... and we should consider changing the key generation logic for these CMs, so that they are always the same (e.g. containing just the CM type and producer GUID, perhaps). This way, they could be log-compacted... That may have some side-effects so it's potentially a bit finicky, but anyway...

xunyin8:
> Thanks for getting started on this Xun! Cool stuff...
> 
> A few comments:
> 
> 1. This is similar to another idea we had earlier, which is to have a CM carry the leader's RT DIV state to the follower. Essentially, serialize the `PartitionState` (or some parts of it, at least...) and pack that into the CM. This would allow the follower to keep track of where the leader is at in the RT consumption, and if the follower becomes leader, it could restore that `PartitionState` it got from the last such CM, subscribe to the RT at that offset, and begin consuming from there. This would mean the follower to leader transition has no DIV gap at all. The `PartitionState` contains the `upstreamOffsetMap`, which I think is the same content that you intend to pass into the new CM.
> 2. Because the VT is consumed by DVC, adding a new CM type could be a bit risky (it'll throw a `UnsupportedMessageTypeException`...). So I think it would be ideal to NOT introduce a new CM here, actually, but to add a new field to the EOS CM containing the above info. Then the leader server could periodically do a EOS/SOS, and the DVC instances and non-upgraded followers that don't care about the new EOS field could safely ignore it, simply leveraging the already-existing KME forward compat.
> 3. If the EOS start packing a lot more content, and we periodically do a EOS/SOS, then they could add up to a non-trivial amount amount of Kafka space. I think that's not ideal... and we should consider changing the key generation logic for these CMs, so that they are always the same (e.g. containing just the CM type and producer GUID, perhaps). This way, they could be log-compacted... That may have some side-effects so it's potentially a bit finicky, but anyway...

2. If we were to do this we’d hide the feature behind the config and deprecate older DVC version and wait for everyone to be on the new version before enabling it. In addition I think we should change the code to ignore unknown CMs instead of throwing exceptions.
3. Using SOS and EOS to carry the information we needed here is an option however it’s not going to be reliable to be used as a heartbeat in the future for time based lag. Since upstream writer could go idle or have issue writing while the Venice pipeline could be doing fine.

FelixGV:
> If we were to do this we’d hide the feature behind the config and deprecate older DVC version and wait for everyone to be on the new version before enabling it.

Adding a field to EOS is an alternative that doesn't require upgrading DVC. This makes it easier to roll out, IMHO.

> Using SOS and EOS to carry the information we needed here is an option however it’s not going to be reliable to be used as a heartbeat in the future for time based lag. Since upstream writer could go idle or have issue writing while the Venice pipeline could be doing fine.

I don't understand... why would it be less reliable? We could still periodically send EOS+SOS even while the upstream writers are idle.

xunyin8:
> > If we were to do this we’d hide the feature behind the config and deprecate older DVC version and wait for everyone to be on the new version before enabling it.
> 
> Adding a field to EOS is an alternative that doesn't require upgrading DVC. This makes it easier to roll out, IMHO.
> 
> > Using SOS and EOS to carry the information we needed here is an option however it’s not going to be reliable to be used as a heartbeat in the future for time based lag. Since upstream writer could go idle or have issue writing while the Venice pipeline could be doing fine.
> 
> I don't understand... why would it be less reliable? We could still periodically send EOS+SOS even while the upstream writers are idle.

Oh, I think I misunderstood your suggestion then. Are you recommending the following:

Leader will periodically emit a separate SOS and EOS containing additional information to either RT or VT. We will persist it to VT upon processing if we are writing it to RT. This way the rollout is much easier.

FelixGV:
> Oh, I think I misunderstood your suggestion then. Are you recommending the following:
> 
> Leader will periodically emit a separate SOS and EOS containing additional information to either RT or VT. We will persist it to VT upon processing if we are writing it to RT. This way the rollout is much easier.

The point of extending the EOS with an extra field is indeed to make the rollout easier, though I was thinking to use the existing SOS/EOS that wrap the data, not use separate ones. Using separate ones is viable too, and might provide interesting tradeoffs (not sure, need to think it through a bit more).

gaojieliu:
> Could you please also mention in the PR description that this change can fix an edge case in AA that when all the ending messages of RT from a colo are ignored, lag would stay high for that colo; and this edge case is the blocker for us to roll out the ACTIVE_ACTIVE ready-to-serve algorithm which checks all RT lags.
So, we still have customers, which are running this mode in prod? I thought agg mode is gone.

xunyin8:
Will create a new PR after revising the design. Decided to go with the reusing SOS CM to generate heartbeat and write through RT.

PR #621, opened 2023-09-06
[server] Support compression with Active/Active + Partial Update

sixpluszero:
## [server] Support compression with Active/Active + Partial Update
This PR fixes compression feature incompatibility with A/A Partial Update:
(1) Add decompression logic for old value fetched from TransientCache.
(2) Make sure the decompressed data ByteBuffer from storage engine contains schema header in front so when updating a key that's written during batch phase, it will be able to infer the schema from the byte array's first 4 bytes.
**Important**: Also this PR fixes a bug in repush in AAWC: Inside repush logic, the chunk assembler for RMD is using value byte's total length to initialize byte array, which will fail when compression is enabled as value size could be less than rmd size in this case.

## How was this PR tested?
New unit test / Enable compression on existing integration test.
For ZSTD_WITH_DICT it is not fully validated after changing to produce float array instead of string map, as the dictionary size is still too big when requesting topic during repush. This should be fixed in a separated PR.


internal/venice-client-common/src/main/java/com/linkedin/venice/compression/GzipCompressor.java
```diff
@@ -95,6 +95,22 @@ public ByteBuffer decompress(byte[] data, int offset, int length) throws IOExcep
...
+  public ByteBuffer decompressAndPrependSchemaHeader(byte[] data, int length) throws IOException {
+    int valueSchema = ByteUtils.readInt(data, 0);
+    byte[] decompressedByteArray;
+    try (InputStream gis =
+        decompress(new ByteArrayInputStream(data, SCHEMA_HEADER_LENGTH, length - SCHEMA_HEADER_LENGTH))) {
+      decompressedByteArray = IOUtils.toByteArray(gis);
+      ByteBuffer byteBufferWithHeader = ByteBuffer.allocate(SCHEMA_HEADER_LENGTH + decompressedByteArray.length)
+          .putInt(valueSchema)
+          .put(decompressedByteArray);
+      byteBufferWithHeader.flip();
```
FelixGV, line 108: Nitpick: `flip()` will do:

```
    public Buffer flip() {
        limit = position;
        position = 0;
        mark = -1;
        return this;
    }
```

But it seems like we only care about the `limit = position` part of that. I think it's more clear to do just the operation we need, than to do this package of operations, e.g.:

```
byteBufferWithHeader.limit(byteBufferWithHeader.position());
byteBufferWithHeader.position(SCHEMA_HEADER_LENGTH);
```

FelixGV (commented):
Thanks Jialin! Just a few minor comments...

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
```diff
@@ -605,13 +606,27 @@ private ByteBuffer getValueBytesForKey(
...
     return originalValue;
   }
 
+  ByteBuffer getCurrentValueFromTransientRecord(PartitionConsumptionState.TransientRecord transientRecord) {
+    ByteBuffer compressedOriginalValue =
+        ByteBuffer.wrap(transientRecord.getValue(), transientRecord.getValueOffset(), transientRecord.getValueLen());
+    try {
+      return getCompressionStrategy().isCompressionEnabled()
+          ? getCompressor().get()
+              .decompressAndPrependSchemaHeader(
```
gaojieliu, line 621: This function assumes that the backed array must have a schema header in the front, but `TransientRecord` doesn't have such condition check, which makes it hard to debug if this does happen in the future.
I still recommend we pass the position and length explicitly, and you can prepend the schema header in the returned ByteBuffer, which is fine.

gaojieliu: It is fine to let the returned ByteBuffer has schema id prepended, but it is risky to assume the backed array of input value has schema id prepended and the position of data is always 4.

gaojieliu (changes requested):
As we discussed, you can use a list of random float as the value, and zstd dict compression ratio for such value will be very low, so that you can test the new logic against zstd dict compression without timing out.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
sixpluszero: Sure, chatted offline and fixed with updated API.

internal/venice-client-common/src/main/java/com/linkedin/venice/compression/GzipCompressor.java
sixpluszero: Sure, thanks for pointing out, I can update this.

internal/venice-client-common/src/main/java/com/linkedin/venice/compression/GzipCompressor.java
```diff
@@ -96,16 +97,22 @@ public ByteBuffer decompress(byte[] data, int offset, int length) throws IOExcep
...
-    int valueSchema = ByteUtils.readInt(data, 0);
+  public ByteBuffer decompressAndMaybePrependSchemaHeader(
+      byte[] data,
+      int offset,
+      int length,
+      boolean prependSchemaHeader) throws IOException {
+    if (prependSchemaHeader && offset < SCHEMA_HEADER_LENGTH) {
+      throw new VeniceException("Start offset does not have enough room for schema header.");
+    }
+    int schemaHeader = prependSchemaHeader ? ByteUtils.readInt(data, offset - SCHEMA_HEADER_LENGTH) : 0;
```
gaojieliu, line 108: What is the reason that we can safely assume there will be a valid schema id before the data when the offset is larger than `SCHEMA_HEADER_LENGTH`?

sixpluszero: Hmm, do you mean we should throw when the offset != SCHEMA_HEADER_LENGTH? Are you suggesting this must be a regression during fetching/decompression logic?

gaojieliu: The previous logic doesn't assume there is a valid schema id before the data and this PR changed this assumption.
Previously, this function just leaves 4 bytes before the decompressed data in the returned ByteBuffer, so that the user of this function can write the schema id into it to avoid another data copy.

sixpluszero: Synced offline, we agreed to use a more straightforward API signature to make sure no confusion.

internal/venice-client-common/src/main/java/com/linkedin/venice/compression/NoopCompressor.java
```diff
@@ -40,14 +39,8 @@ public ByteBuffer decompress(byte[] data, int offset, int length) throws IOExcep
...
   @Override
-  public ByteBuffer decompressAndMaybePrependSchemaHeader(
-      byte[] data,
-      int offset,
-      int length,
-      boolean prependSchemaHeader) throws IOException {
-    if (prependSchemaHeader && offset < SCHEMA_HEADER_LENGTH) {
-      throw new VeniceException("Start offset does not have enough room for schema header.");
-    }
+  public ByteBuffer decompressAndPrependSchemaHeader(byte[] data, int offset, int length, int schemaHeader)
```
gaojieliu, line 42: So this function is really prepending `schemaHeader`..
Is it expected?

sixpluszero: As we discussed offline, we are passing the schema header retrieved into the method, so it should be OK to prepend inside the method right?

gaojieliu (approved):
Thanks for the fix!

PR #622, opened 2023-09-07
[server][da-vinci-client] Kill SIT if missing SOP for 1 hr

majisourav99:
## Kill push if SOP is not seen for over 1hr

For topic which are missing START_OF_PUSH control message, the store ingestion task keeps waiting on it thus blocking ST threads till it barfs out after 24hrs. This PR fails the push if it does not see SOP for 1 hour.

## How was this PR tested?

GHCI

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
@@ -111,6 +111,8 @@ import java.util.Properties;
 import java.util.Queue;
 import java.util.Set;
 import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 import java.util.concurrent.PriorityBlockingQueue;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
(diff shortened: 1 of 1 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -406,6 +413,8 @@ public StoreIngestionTask(
         new IngestionNotificationDispatcher(notifiers, kafkaVersionTopic, isCurrentVersion),
         amplificationFactorAdapter);
 
+    this.runnableForKillIngestionTasksForMissingSOP = () -> waitForStateVersion(kafkaVersionTopic);
```
ZacAttack, line 416: Can we verify that this triggers ERROR or an exception in OFFLINE->STANDBY?  The original bug hit was because we got stuck on waiting for someone to call StateModelIngestionProgressNotifier.stopConsumption(), a quick glance implies we only call this if we get an exception in that transition.  Looks like kill() errors the resource, but I'm trying to find where/if it triggers that exception case.

ZacAttack: Walked though in IDE with Sourav, looks good.  We'll call reportKilled in run() if the consumerAction is processed.

ZacAttack (approved):
Looks good.

PR #623, opened 2023-09-07
[vpj][controller][common] Add KME schema reader into repush

shuhui-liu:
## Summary

Today KIF repush does not have KME schema reader. During KME upgrades, if controllers deploy first, controller will produce control messages with new KME, and repush will fail because it could not find the new KME in code to deserialize these messages. The commit adds schema reader support into KIF repush so that repush can retrieve unknown KME from source colo system store.

## How was this PR tested?

New integ test


clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
```diff
@@ -168,6 +168,9 @@ public class VenicePushJob implements AutoCloseable {
   public static final String SEND_CONTROL_MESSAGES_DIRECTLY = "send.control.messages.directly";
   public static final String SOURCE_ETL = "source.etl";
   public static final String ETL_VALUE_SCHEMA_TRANSFORMATION = "etl.value.schema.transformation";
+  public static final String SCHEMA_READER_ENABLED = "schema.reader.enabled";
+  public static final String SCHEMA_SYSTEM_STORE_D2_ZK_HOST = "schema.system.store.d2.zk.host";
+  public static final String SCHEMA_SYSTEM_STORE_D2_SERVICE_NAME = "schema.system.store.d2.service.name";
```
nisargthakkar, line 173: Instead of having d2 zk host differentiated by feature, let's use something region based. VPJ already has `d2.zk.hosts.<region name>` configs. So, we can have a config that defines the parent region, or we could reuse `source.grid.fabric`

nisargthakkar (commented):
Not read through the whole PR, but wanted to leave one comment

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
shuhui-liu: It is a good idea to reuse existing configs. But neither `d2.zk.hosts.<parent.controller.region.name>` or `d2.zk.hosts.<source.grid.fabric>` works for the following case. Please correct me if I'm wrong.

When we use prod-lor1 data to repush the store, the push job will consume prod-lor1 kakfa topic, which may have records with new KME so we need to retrieve KME from prod-lor1 routers. But the push job will run in WAR, whose `source.grid.fabric` is prod-lva1. Using `d2.zk.hosts.prod-lva1` will connect to prod-lva1 routers and may not be able to retrieve the new KME as KME is not registered in all fabrics at the same time.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/input/kafka/KafkaInputUtils.java
```diff
@@ -55,6 +72,35 @@ public static VeniceProperties getConsumerProperties(JobConf config) {
     return new VeniceProperties(consumerFactoryProperties);
   }
 
+  public static KafkaValueSerializer getKafkaValueSerializer(JobConf config) {
+    KafkaValueSerializer kafkaValueSerializer = new OptimizedKafkaValueSerializer();
+    boolean isSchemaReaderEnabled = Boolean.parseBoolean(config.get(SCHEMA_READER_ENABLED, "false"));
```
gaojieliu, line 77: I saw this is mainly for KIF repush, which means the d2 host/service needs to align with the repush source colo, right?
Having them as hard-coded config could introduce some maintenance burden, and if we need to configure a different zk host when switching repush source, that would introduce some operation toils.

If the above assumption is true, I would like to automate this and here is some idea:
1. Talk to parent controller to retrieve the zk host for a particular region.
2. We can pass the region param in the call to Parent Controller.
3. Parent Controller will talk to the corresponding Child Controller to retrieve these setting.

Or an even simpler way is to return the ZK host/service in the `RepushInfoResponse`, which is region specific:
https://github.com/linkedin/venice/blob/main/internal/venice-common/src/main/java/com/linkedin/venice/controllerapi/ControllerClient.java#L242

The `RepushInfoResponse` already contains the Kafka url for the source region and we can add source ZK into it, so that there is no ZK config for repush and it is fully automated.

What do you think?

clients/venice-push-job/src/test/java/com/linkedin/venice/hadoop/VenicePushJobTest.java
```diff
@@ -682,6 +684,14 @@ public void testTargetedRegionPushPostValidationFailedForValidation() throws Exc
     verify(pushJob, never()).postValidationConsumption(any());
   }
 
+  @Test(expectedExceptions = VeniceException.class, expectedExceptionsMessageRegExp = "D2 zk host and service name must be provided when schema reader is enabled")
```
gaojieliu, line 687: Is it possible to write an integration test for this?

I know it is hard, and if not, let us try to write more comprehensive unit tests since the current unit test seems to be dummy.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/input/kafka/KafkaInputUtils.java
shuhui-liu: This is a great idea to reduce maintenance overhead. Thanks. I made controller changes to return system schema cluster d2 service name and zk host of the source olo in `RepushInfoResponse`.

clients/venice-push-job/src/test/java/com/linkedin/venice/hadoop/VenicePushJobTest.java
shuhui-liu: Sure. After taking another look, I find that adding an integ test for this is not as complex as I thought. There is an existing test class `ConsumerIntegrationTest` I can leverage. This class can create a venice writer which produces records with new KME into a kafka topic.

Therefore, I added `testKIFRepushForwardCompatibility` which produces records with new KME into VT and then runs repush job. The test passes when schema reader is enabled and fails otherwise with exception:
```
java.lang.Exception: com.linkedin.venice.exceptions.VeniceMessageException: Received Protocol Version '12' which is not supported by OptimizedKafkaValueSerializer. Protocol forward compatibility is not enabled. The only supported Protocol Versions are: {1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11}.
```

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
```diff
@@ -715,6 +722,7 @@ private PushJobSetting getPushJobSetting(VeniceProperties props) {
     pushJobSettingToReturn.repushTTLStartTimeMs = props.getLong(REPUSH_TTL_START_TIMESTAMP, System.currentTimeMillis());
     pushJobSettingToReturn.isTargetedRegionPushEnabled = props.getBoolean(TARGETED_REGION_PUSH_ENABLED, false);
     pushJobSettingToReturn.postValidationConsumption = props.getBoolean(POST_VALIDATION_CONSUMPTION_ENABLED, true);
+    pushJobSettingToReturn.isSystemSchemaReaderEnabled = props.getBoolean(SYSTEM_SCHEMA_READER_ENABLED, false);
```
gaojieliu, line 725: I guess we can only start using this feature after deploying the related code changes to Controller first.
Let us add such deployment order requirement in the commit messages.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/input/kafka/KafkaInputUtils.java
```diff
@@ -55,6 +72,35 @@ public static VeniceProperties getConsumerProperties(JobConf config) {
     return new VeniceProperties(consumerFactoryProperties);
   }
 
+  public static KafkaValueSerializer getKafkaValueSerializer(JobConf config) {
+    KafkaValueSerializer kafkaValueSerializer = new OptimizedKafkaValueSerializer();
+    boolean isSchemaReaderEnabled = Boolean.parseBoolean(config.get(SYSTEM_SCHEMA_READER_ENABLED, "false"));
+    if (isSchemaReaderEnabled) {
```
gaojieliu, line 78: Let us make sure the Grid/AZ has access to prod Router to retrieve system schemas before starting using this feature.

gaojieliu (approved):
LGTM, thanks!

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
shuhui-liu: VPJ can be deployed before controller because the feature is disabled by default. But to enable the feature, this commit needs to be deployed to controller. Otherwise, the repush job will fail with `D2 service name and zk host must be provided when system schema reader is enabled` exception. I'll mention this in the commit message.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/input/kafka/KafkaInputUtils.java
shuhui-liu: Yeah I have checked with grid team. It has network acl to prod router.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
shuhui-liu: In the latest change of the PR, we do not need to specify d2 zk hosts or service name before running repush job. Those information will be retrieved from controller by `getRepushInfo` api.

PR #624, opened 2023-09-08
[all] Bumped up avro-util dep to pick up a fix in fast serializer generation

gaojieliu:
The latest version of fast-avro added the support of splitting large methods in fast serializer.

## How was this PR tested?
CI

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -39,7 +39,7 @@ if (project.hasProperty('overrideBuildEnvironment')) {
 }
 
 def avroVersion = '1.10.2'
-def avroUtilVersion = '0.2.150'
+def avroUtilVersion = '0.3.17'
 def grpcVersion = '1.49.2'
 def kafkaGroup = 'com.linkedin.kafka'
 def kafkaVersion = '2.4.1.65'


nisargthakkar (approved):

build.gradle
```diff
@@ -39,7 +39,7 @@ if (project.hasProperty('overrideBuildEnvironment')) {
 }
 
 def avroVersion = '1.10.2'
-def avroUtilVersion = '0.2.150'
+def avroUtilVersion = '0.3.17'
```
FelixGV, line 42: That looks like a big bump! We should look carefully to get confidence that there are no regressions....

FelixGV (approved):
Thanks!

build.gradle
gaojieliu: The CI passed, and we will certainly bake this in EI.
Fast-Avro hasn't updated much, and mostly the avro compatiblity layer.

nisargthakkar: It's not really too big of a jump in production, actually. Our internal dependencies were bumping up `avro-util` transitively to a much more recent version (`0.3.14`)

nisargthakkar: There is one regression related to `RandomRecordGenerator` that we've identified - [This PR](https://github.com/linkedin/avro-util/pull/495/files) breaks random object generation for non-record schemas

PR #625, opened 2023-09-08
[producer] Add log when extract value object fails

majisourav99:
## Add log when extract value object fails

It will help debugging online producer CLI issues.


clients/venice-producer/src/main/java/com/linkedin/venice/producer/online/ProducerTool.java
```diff
@@ -198,7 +198,8 @@ private static Object getValueObject(String valueString, RouterBasedStoreSchemaF
         value = adaptDataToSchema(valueString, valueSchema);
         break;
       } catch (Exception e) {
-        // Nothing to do. Try the next schema
```
nisargthakkar, line 201: This is too verbose as it will print exceptions even if a schema could be matched. I have another PR (#626) for this. Can we use that instead?

PR #626, opened 2023-09-08
[producer] Log details if input doesn't match any value schema in ProducerTool

nisargthakkar:
## Log details if input doesn't match any value schema in ProducerTool

It will give users more context about the errors. New output format:
```
[ERROR] Value schema not found. Is a schema that is compatible with the specified data already registered?
Exception messages for each schema:
	Schema Id: 1. Exception: org.apache.avro.AvroTypeException: Expected start-union. Got START_ARRAY
	Schema Id: 2. Exception: org.apache.avro.AvroTypeException: Expected start-union. Got START_ARRAY
```

## How was this PR tested?

Tested locally. GH CI for the rest

### DIFF
--- a/clients/venice-producer/src/main/java/com/linkedin/venice/producer/online/ProducerTool.java
+++ b/clients/venice-producer/src/main/java/com/linkedin/venice/producer/online/ProducerTool.java
@@ -19,8 +19,8 @@ import com.linkedin.venice.utils.SslUtils;
 import com.linkedin.venice.utils.VeniceProperties;
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
-import java.util.ArrayList;
-import java.util.List;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.Properties;
 import java.util.stream.Collectors;
 import org.apache.avro.Schema;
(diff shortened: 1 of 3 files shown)


clients/venice-producer/src/main/java/com/linkedin/venice/producer/online/ProducerTool.java
```diff
@@ -193,17 +196,23 @@ private static void writeToStore(ProducerContext producerContext) throws Excepti
 
   private static Object getValueObject(String valueString, RouterBasedStoreSchemaFetcher schemaFetcher) {
     Object value = null;
+    List<Exception> exceptionList = new ArrayList<>();
     for (Schema valueSchema: schemaFetcher.getAllValueSchemas()) {
```
majisourav99, line 200: not path of this change, can we match the schemas in reverse order ? hope is that it will reduce older schema mismatches.

clients/venice-producer/src/main/java/com/linkedin/venice/producer/online/ProducerTool.java
```diff
@@ -193,17 +196,23 @@ private static void writeToStore(ProducerContext producerContext) throws Excepti
...
       try {
         value = adaptDataToSchema(valueString, valueSchema);
         break;
       } catch (Exception e) {
-        // Nothing to do. Try the next schema
+        exceptionList.add(e);
```
majisourav99, line 205: can we add only the top most or top few stacks ? rest of them may not provide any valuable info.

clients/venice-producer/src/main/java/com/linkedin/venice/producer/online/ProducerTool.java
```diff
@@ -193,17 +196,23 @@ private static void writeToStore(ProducerContext producerContext) throws Excepti
...
+        // Try the next schema
       }
     }
     if (value == null) {
+      String exceptionDelimiter = "\n\t";
+      String exceptionDetails =
+          exceptionList.stream().map(Throwable::getMessage).collect(Collectors.joining(exceptionDelimiter));
       throw new VeniceException(
-          "Value schema not found. Is a schema that is compatible with the specified data already registered?");
+          "Value schema not found. Is a schema that is compatible with the specified data already registered?\nException messages for each schema (not in order):"
```
majisourav99, line 214: possible to add schema id to corresponding stack trace ? (probably will need a different API to fetch)

clients/venice-producer/src/main/java/com/linkedin/venice/producer/online/ProducerTool.java
nisargthakkar: `schemaFetcher.getAllValueSchemas()` returns a `Set` and not a `List`. So, we don't have info about the schema id there

clients/venice-producer/src/main/java/com/linkedin/venice/producer/online/ProducerTool.java
nisargthakkar: Yeah. We don't have that info currently. We'll need to add a new API

clients/venice-producer/src/main/java/com/linkedin/venice/producer/online/ProducerTool.java
nisargthakkar: This messaging is for the user and the stack trace doesn't provide any valuable information for them. We can add a debug mode if we want more details for debugging.

clients/venice-producer/src/main/java/com/linkedin/venice/producer/online/ProducerTool.java
```diff
@@ -193,17 +196,23 @@ private static void writeToStore(ProducerContext producerContext) throws Excepti
...
       } catch (Exception e) {
-        // Nothing to do. Try the next schema
+        exceptionList.add(e);
+        // Try the next schema
       }
     }
     if (value == null) {
+      String exceptionDelimiter = "\n\t";
+      String exceptionDetails =
+          exceptionList.stream().map(Throwable::getMessage).collect(Collectors.joining(exceptionDelimiter));
```
majisourav99, line 212: IIRC getMessage could be null for certain types of exceptions eg NPE.

nisargthakkar: Changed the output format to:
```
[ERROR] Value schema not found. Is a schema that is compatible with the specified data already registered?
Exception messages for each schema:
	Schema Id: 1. Exception: org.apache.avro.AvroTypeException: Expected start-union. Got START_ARRAY
	Schema Id: 2. Exception: org.apache.avro.AvroTypeException: Expected start-union. Got START_ARRAY
```

majisourav99 (approved):
LGTM. Thanks for the change

PR #627, opened 2023-09-08
[build] Fix JDK8 builds and set up CI

nisargthakkar:
## Fix JDK8 builds and set up CI

Previously, builds using JDK8 would fail. Also, cleaned up the `alpn-boot` loading process for HTTP/2 on JDK 8 tests.

## How was this PR tested?

Running GH CI

### DIFF
--- a/.github/rawWorkflows/gh-ci-alpini-parametrized-flow.txt
+++ b/.github/rawWorkflows/gh-ci-alpini-parametrized-flow.txt
@@ -2,7 +2,7 @@
     strategy:
       fail-fast: false
       matrix:
-        jdk: [11, 17]
+        jdk: [8, 11, 17]
     runs-on: ubuntu-latest
     timeout-minutes: $TimeOut
     outputs:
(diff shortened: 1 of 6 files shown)


.github/rawWorkflows/gh-ci-alpini-parametrized-flow.txt
```diff
@@ -28,7 +28,7 @@
         if: steps.check_alpini_files_changed.outputs.alpini == 'true'
         with:
           java-version: ${{ matrix.jdk }}
-          distribution: 'microsoft'
+          distribution: 'temurin'
```
FelixGV, line 31: Microsoft is not available for jdk8?

FelixGV (approved):
Nice, thanks for the cleanups!

.github/rawWorkflows/gh-ci-alpini-parametrized-flow.txt
nisargthakkar: Yeah. Microsoft's docs suggest to use `temurin` instead: https://learn.microsoft.com/en-us/java/openjdk/download#openjdk-8. We could split the matrix to use `microsoft` for jdk 11 and 17, and `temurin` for jdk 8 if we wish to

FelixGV: Hmm... either way is fine I guess. I can give another ship it if you prefer changing it. Your call.

nisargthakkar: I think we can keep it this way and change it later if there are problems

PR #628, opened 2023-09-11
[controller][test] Message header triggered new KME schema registrati…

lluwm:
…on in child controller

As part of the solution to remove deployment order for KafkaMessageEnvelope (KME) value schema, for controllers, when they find an unknown KME schema from the message header, they need to (talk to the system cluster leader controller to) register the new schema into local region system store. It helps controllers to deal with unknown KME during admin message consumption.

This change enables the child controllers to register unknown KME schemas when discovering them from the message's header. It mainly leverages the existing functionalities from the ControllerClientBackedSystemSchemaInitializer class for the implementation. A new config is introduced to enable this feature and a new integration test is added to verify the correctness of the new feature.

## How was this PR tested?

Added a new integration test.
Passed CI tests.

### DIFF
--- a/internal/venice-common/src/main/java/com/linkedin/venice/helix/HelixReadWriteSchemaRepository.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/helix/HelixReadWriteSchemaRepository.java
@@ -560,7 +560,7 @@ public class HelixReadWriteSchemaRepository implements ReadWriteSchemaRepository
   }
 
   // For testing purpose only.
-  public void removeValueSchema(String storeName, int schemaId) {
+  void removeValueSchema(String storeName, int schemaId) {
     preCheckStoreCondition(storeName);
     accessor.removeValueSchema(storeName, String.valueOf(schemaId));
   }
--- a/internal/venice-common/src/main/java/com/linkedin/venice/helix/HelixReadWriteSchemaRepository.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/helix/HelixReadWriteSchemaRepository.java
@@ -558,4 +558,10 @@ public class HelixReadWriteSchemaRepository implements ReadWriteSchemaRepository
       throw new VeniceNoStoreException(storeName);
     }
   }
+
+  // For testing purpose only.
+  public void removeValueSchema(String storeName, int schemaId) {
+    preCheckStoreCondition(storeName);
+    accessor.removeValueSchema(storeName, String.valueOf(schemaId));
+  }
 }
--- a/internal/venice-common/src/main/java/com/linkedin/venice/pubsub/api/PubSubMessageDeserializer.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/pubsub/api/PubSubMessageDeserializer.java
@@ -1,6 +1,7 @@
 package com.linkedin.venice.pubsub.api;
 
 import com.linkedin.avroutil1.compatibility.AvroCompatibilityHelper;
+import com.linkedin.venice.exceptions.VeniceException;
 import com.linkedin.venice.kafka.protocol.KafkaMessageEnvelope;
 import com.linkedin.venice.kafka.protocol.enums.MessageType;
 import com.linkedin.venice.message.KafkaKey;
@@ -63,6 +64,15 @@ public class PubSubMessageDeserializer {
             Schema providedProtocolSchema = AvroCompatibilityHelper.parse(new String(header.value()));
             value =
                 valueSerializer.deserialize(valueBytes, providedProtocolSchema, getEnvelope(key.getKeyHeaderByte()));
+          } catch (VeniceException e) {
+            /*
+             * If a Venice exception is caught when registering a new discovered schema, we don't need any special
+             * management here, delegate it to the follower deserialize to handle such case.
+             */
+            LOGGER.warn(
+                "Venice exception in registering new encountered schema in protocol header: ",
+                VENICE_TRANSPORT_PROTOCOL_HEADER,
+                e);
           } catch (Exception e) {
             // Improper header... will ignore.
             LOGGER.warn("Received unparsable schema in protocol header: " + VENICE_TRANSPORT_PROTOCOL_HEADER, e);
@@ -96,5 +106,10 @@ public class PubSubMessageDeserializer {
       default:
         throw new IllegalStateException("Illegal key header byte: " + keyHeaderByte);
     }
+  }
+
+  // For testing only.
+  public KafkaValueSerializer getValueSerializer() {
+    return valueSerializer;
   }
 }
--- a/internal/venice-common/src/main/java/com/linkedin/venice/pubsub/api/PubSubMessageDeserializer.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/pubsub/api/PubSubMessageDeserializer.java
@@ -1,7 +1,6 @@
 package com.linkedin.venice.pubsub.api;
 
 import com.linkedin.avroutil1.compatibility.AvroCompatibilityHelper;
-import com.linkedin.venice.exceptions.VeniceException;
 import com.linkedin.venice.kafka.protocol.KafkaMessageEnvelope;
 import com.linkedin.venice.kafka.protocol.enums.MessageType;
 import com.linkedin.venice.message.KafkaKey;
@@ -64,18 +63,12 @@ public class PubSubMessageDeserializer {
             Schema providedProtocolSchema = AvroCompatibilityHelper.parse(new String(header.value()));
             value =
                 valueSerializer.deserialize(valueBytes, providedProtocolSchema, getEnvelope(key.getKeyHeaderByte()));
-          } catch (VeniceException e) {
-            /*
-             * If a Venice exception is caught when registering a new discovered schema, we don't need any special
-             * management here, delegate it to the follower deserialize to handle such case.
-             */
-            LOGGER.warn(
-                "Venice exception in registering new encountered schema in protocol header: ",
-                VENICE_TRANSPORT_PROTOCOL_HEADER,
-                e);
           } catch (Exception e) {
             // Improper header... will ignore.
-            LOGGER.warn("Received unparsable schema in protocol header: " + VENICE_TRANSPORT_PROTOCOL_HEADER, e);
+            LOGGER.warn(
+                "Received unparsable schema or encountered schema registration issue in protocol header: "
+                    + VENICE_TRANSPORT_PROTOCOL_HEADER,
+                e);
           }
           break; // We don't look at other headers
         }
--- a/internal/venice-common/src/main/java/com/linkedin/venice/system/store/ControllerClientBackedSystemSchemaInitializer.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/system/store/ControllerClientBackedSystemSchemaInitializer.java
@@ -15,6 +15,7 @@ import com.linkedin.venice.controllerapi.StoreResponse;
 import com.linkedin.venice.controllerapi.UpdateStoreQueryParams;
 import com.linkedin.venice.exceptions.ErrorType;
 import com.linkedin.venice.exceptions.VeniceException;
+import com.linkedin.venice.schema.avro.DirectionalSchemaCompatibilityType;
 import com.linkedin.venice.schema.writecompute.WriteComputeSchemaConverter;
 import com.linkedin.venice.security.SSLFactory;
 import com.linkedin.venice.serialization.avro.AvroProtocolDefinition;
@@ -76,7 +77,7 @@ public class ControllerClientBackedSystemSchemaInitializer {
     this.enforceSslOnly = enforceSslOnly;
   }
 
-  public void execute() {
+  public void execute(Map<Integer, Schema> inputSchemas) {
     if (controllerClient == null) {
       if (!controllerUrl.isEmpty()) {
         controllerClient = ControllerClient.constructClusterControllerClient(clusterName, controllerUrl, sslFactory);
@@ -100,7 +101,9 @@ public class ControllerClientBackedSystemSchemaInitializer {
     }
 
     String storeName = protocolDefinition.getSystemStoreName();
-    Map<Integer, Schema> schemasInLocalResources = Utils.getAllSchemasFromResources(protocolDefinition);
+
+    Map<Integer, Schema> schemasInLocalResources =
+        inputSchemas != null ? inputSchemas : Utils.getAllSchemasFromResources(protocolDefinition);
     D2ServiceDiscoveryResponse discoveryResponse = controllerClient.retryableRequest(
         DEFAULT_RETRY_TIMES,
         c -> c.discoverCluster(storeName),
@@ -142,17 +145,29 @@ public class ControllerClientBackedSystemSchemaInitializer {
     for (int version = 1; version <= protocolDefinition.getCurrentProtocolVersion(); version++) {
       Schema schemaInLocalResources = schemasInLocalResources.get(version);
       if (schemaInLocalResources == null) {
-        throw new VeniceException(
-            "Invalid protocol definition: " + protocolDefinition.name() + " does not have a version " + version
-                + " even though it is less than or equal to the current version ("
-                + protocolDefinition.getCurrentProtocolVersion() + ").");
-      }
-      checkAndMayRegisterValueSchema(storeName, version, schemasInZk.get(version), schemaInLocalResources);
+        if (inputSchemas == null) {
+          throw new VeniceException(
+              "Invalid protocol definition: " + protocolDefinition.name() + " does not have a version " + version
+                  + " even though it is less than or equal to the current version ("
+                  + protocolDefinition.getCurrentProtocolVersion() + ").");
+        }
+      } else {
+        checkAndMayRegisterValueSchema(
+            storeName,
+            version,
+            schemasInZk.get(version),
+            schemaInLocalResources,
+            inputSchemas != null);
 
-      if (autoRegisterPartialUpdateSchema) {
-        checkAndMayRegisterPartialUpdateSchema(storeName, version, schemaInLocalResources);
+        if (autoRegisterPartialUpdateSchema) {
+          checkAndMayRegisterPartialUpdateSchema(storeName, version, schemaInLocalResources);
+        }
       }
     }
+  }
+
+  public void execute() {
+    execute(null);
   }
 
   private boolean hasLeaderController() {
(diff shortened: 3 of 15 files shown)


internal/venice-common/src/main/java/com/linkedin/venice/pubsub/api/PubSubMessageDeserializer.java
```diff
@@ -63,6 +64,15 @@ public PubSubMessage<KafkaKey, KafkaMessageEnvelope, Long> deserialize(
...
+          } catch (VeniceException e) {
+            /*
+             * If a Venice exception is caught when registering a new discovered schema, we don't need any special
+             * management here, delegate it to the follower deserialize to handle such case.
+             */
+            LOGGER.warn(
+                "Venice exception in registering new encountered schema in protocol header: ",
+                VENICE_TRANSPORT_PROTOCOL_HEADER,
+                e);
           } catch (Exception e) {
```
ZacAttack, line 76: These two exception cases look very similar to me.  What is their difference?

internal/venice-common/src/main/java/com/linkedin/venice/helix/HelixReadWriteSchemaRepository.java
```diff
@@ -558,4 +558,10 @@ private void preCheckStoreCondition(String storeName) {
       throw new VeniceNoStoreException(storeName);
     }
   }
+
+  // For testing purpose only.
+  public void removeValueSchema(String storeName, int schemaId) {
```
ZacAttack, line 563: Package private maybe?  This method sounds scary

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerService.java
```diff
@@ -120,17 +125,45 @@ public VeniceControllerService(
...
+    /**
+     * Register a callback function to handle the case when a new KME value schema is encountered when the child controller
+     * consumes the admin topics.
+     */
+    BiConsumer<Integer, Schema> newSchemaEncountered = (schemaId, schema) -> {
+      LOGGER.info("Encountered a new KME value schema (id = {}), proceed to register", schemaId);
+      String systemClusterName = multiClusterConfigs.getSystemSchemaClusterName();
+      VeniceControllerConfig systemStoreClusterConfig = multiClusterConfigs.getControllerConfig(systemClusterName);
+      try {
+        ControllerClientBackedSystemSchemaInitializer schemaInitializer =
```
ZacAttack, line 138: Can we maybe make this a member variable?  Might make testing easier.

internal/venice-common/src/main/java/com/linkedin/venice/helix/HelixReadWriteSchemaRepository.java
lluwm: Yeah, a scary api and I was hesitating and trying to find alternative ways of doing it, but I didn't have a good one.

Do we have a convention to put all integration tests under `venice/internal/venice-test-common/src/integrationTest`?  If not, can move it to a different folder. Tried to remove the public keyword with the current folder structure, but compiler complained.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/controller/TestControllerKMERegistrationFromMessageHeader.java
```diff
@@ -0,0 +1,131 @@
...
+    String storeName = Utils.getUniqueString("store");
+
+    VeniceControllerWrapper pController =
+        multiRegionMultiClusterWrapper.getLeaderParentControllerWithRetries(clusterName);
+
+    VeniceMultiClusterWrapper child = childDatacenters.get(0);
+
+    // Remove the latest schema from child controller's local value serializer and remove it from child colo's schema
+    // repository (ZK).
+    VeniceControllerWrapper leaderController = child.getLeaderController(clusterName);
```
ZacAttack, line 82: This is an interesting approach...  But I worry it's too interesting lol.  I wonder, how hard would it be to just read the latest schema and make a doc string change on some field and publish it?  I only ask this as a future proof method of testing this (in case some schema is cached somewhere and the removing/adding post service start up is a NoOp here, but in the production scenario where we add a new schema this may do something else.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/controller/TestControllerKMERegistrationFromMessageHeader.java
```diff
@@ -0,0 +1,131 @@
...
+     */
+
+    pController.getVeniceAdmin().createStore(clusterName, storeName, "", "\"string\"", "\"string\"", false);
+
+    // Verify that schema is registered in the child colo's schema repo.
+    TestUtils.waitForNonDeterministicAssertion(30, TimeUnit.SECONDS, true, () -> {
+      Assert.assertTrue(
+          repo.getValueSchema(
+              AvroProtocolDefinition.KAFKA_MESSAGE_ENVELOPE.getSystemStoreName(),
+              AvroProtocolDefinition.KAFKA_MESSAGE_ENVELOPE.getCurrentProtocolVersion()) != null);
```
ZacAttack, line 123: This doesn't seem strict enough?  Could we get the protocol version that we expect at the top of the test?

lluwm: Yeah, we can add the same check in the beginning of the test.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/controller/TestControllerKMERegistrationFromMessageHeader.java
lluwm: Yeah, it's a bit involved to develop this testing code. The high level idea here is that I want to create a inconsistent scenario where parent controller has a newer KME version, but a child controller hasn't got it yet. This is to mimic the case where parent controller deploy a new schema before the child and verify that new code can handle it.

With that idea in mind, say if we just update and publish a new schema through e.g. parent controller's `addValueSchema` api, the parent and child controller (through admin messages) are always in sync with their schema and we cannot trigger the new code path anymore. Maybe I misunderstood you comments?

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/api/PubSubMessageDeserializer.java
lluwm: You are right, I don't have to add the new VeniceException catch and it can still work. The only reason was to add a  bit precise warning message. So, if it doesn't give us enough benefits, we can simply remove it.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/controller/TestControllerKMERegistrationFromMessageHeader.java
ZacAttack: I think my concern is that the child controller started up and THEN we removed the zk.  Maybe we bounce the child controller, or, remove the schema from child zk and then start the controller, or, after this step start a new child controller and kill the old one.

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/api/PubSubMessageDeserializer.java
ZacAttack: I'm trying to even understand the difference between the error messages

internal/venice-common/src/main/java/com/linkedin/venice/helix/HelixReadWriteSchemaRepository.java
ZacAttack: If it's only used in an integration test and and not a unit test I think we could put the test case where it would work.  Unit tests are the ones that get finnicky as then they won't attribute coverage properly.

lluwm: Sounds good. I will move it.

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/api/PubSubMessageDeserializer.java
lluwm: Okay, let me generalize the message and remove the new exception catch block

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/controller/TestControllerKMERegistrationFromMessageHeader.java
lluwm: I think there might be a catch with the proposal. Today, controller, when transitioning from standby->leader, has a logic of loading all schemas from local KafkaMessageEvelop.avsc (`internal/venice-common/src/main/resources/avro/KafkaMessageEnvelope`) file, looping through them and registering in the local cluster, if there is any missing ones in the ZK. So, it means that even if we bounce a child controller after removing it from zk, the newly started controller would always fix it after taking the leadership.

 The current testing approach might be okay, because `HelixReadWriteSchemaRepository` doesn't cache any schemas locally, so even if we remove after the starting up, it doesn't have stale schema.

`KafkaValueSerializer` does cache schemas in memory, thus in the test, we make sure to remove all cached data by `valueSerializer.removeAllSchemas(); `

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerService.java
lluwm: Do you mean `newSchemaEncountered` ? yeah, I think I can make it a class member.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/controller/TestControllerKMERegistrationFromMessageHeader.java
ZacAttack: != null part, I think we should check the version number

lluwm: okay, sure, will check that.

internal/venice-common/src/main/java/com/linkedin/venice/system/store/ControllerClientBackedSystemSchemaInitializer.java
```diff
@@ -142,19 +145,31 @@ public void execute() {
     for (int version = 1; version <= protocolDefinition.getCurrentProtocolVersion(); version++) {
       Schema schemaInLocalResources = schemasInLocalResources.get(version);
       if (schemaInLocalResources == null) {
-        throw new VeniceException(
-            "Invalid protocol definition: " + protocolDefinition.name() + " does not have a version " + version
-                + " even though it is less than or equal to the current version ("
-                + protocolDefinition.getCurrentProtocolVersion() + ").");
-      }
-      checkAndMayRegisterValueSchema(storeName, version, schemasInZk.get(version), schemaInLocalResources);
+        if (inputSchemas == null) {
```
shuhui-liu, line 148: Curious, why do we need this if branch? From above logic:
```
Map<Integer, Schema> schemasInLocalResources =
        inputSchemas != null ? inputSchemas : Utils.getAllSchemasFromResources(protocolDefinition);
```
If `schemaInLocalResources` is null, `inputSchemas` must be null, right?

internal/venice-common/src/main/java/com/linkedin/venice/system/store/ControllerClientBackedSystemSchemaInitializer.java
```diff
@@ -236,11 +251,20 @@ private void checkAndMayRegisterValueSchema(
...
+      SchemaResponse addValueSchemaResponse = controllerClient.retryableRequest(DEFAULT_RETRY_TIMES, c -> {
+        if (skipCompatibilityCheck) {
+          return c.addValueSchema(
+              storeName,
+              schemaInLocalResources.toString(),
+              valueSchemaId,
+              DirectionalSchemaCompatibilityType.NONE);
+        }
+        return c.addValueSchema(storeName, schemaInLocalResources.toString(), valueSchemaId);
+      });
```
shuhui-liu, line 266: KME schemas are backward compatible. We'd better enforce backward compatibility `DirectionalSchemaCompatibilityType.BACKWARD` when registering a new KME. Note it varies for different schemas. For example, meta system store schema needs to be registered with ` DirectionalSchemaCompatibilityType.FULL`. You may need to add ` DirectionalSchemaCompatibilityType` as a parameter of the function or a member of the class.

shuhui-liu (commented):
Thanks for making this change. I took a look at all code changes except for the new test class (will review the test after meetings).

Great work overall! Left two comments. Also, regarding the following commit message:
```
Otherwise, as KME is not embedded in every Kafka message, if servers restart and resume the consumption from a non-SOS message with new KME (for example, lor1 servers could consume TS sent by lva1 controller), servers can not find the right schema to deserialize it and keep failing.
```
I think this PR could not help with servers that encounter unknown KME after restart because this PR does not include any server change. It actually helps controllers which may encounter unknown KME after restarting and resuming admin message consumption. You may need to update the commit message to avoid confusion.

internal/venice-common/src/main/java/com/linkedin/venice/system/store/ControllerClientBackedSystemSchemaInitializer.java
lluwm: I know it's a bit confusing, but `schemaInLocalResources` is defined one line above (not the same as `schemasInLocalResources`).

So, in its usage, e.g. `schemaInitializer.execute(ImmutableMap.of(schemaId, schema));`, the passed in map contains only 1 entry, so we could have schemaInLocalResources == null for some versions, which is a valid case, and we not throw exception in this case.

internal/venice-common/src/main/java/com/linkedin/venice/system/store/ControllerClientBackedSystemSchemaInitializer.java
lluwm: Sounds good. I will fix this.

lluwm:
> Thanks for making this change. I took a look at all code changes except for the new test class (will review the test after meetings).
> 
> Great work overall! Left two comments. Also, regarding the following commit message:
> 
> ```
> Otherwise, as KME is not embedded in every Kafka message, if servers restart and resume the consumption from a non-SOS message with new KME (for example, lor1 servers could consume TS sent by lva1 controller), servers can not find the right schema to deserialize it and keep failing.
> ```
> 
> I think this PR could not help with servers that encounter unknown KME after restart because this PR does not include any server change. It actually helps controllers which may encounter unknown KME after restarting and resuming admin message consumption. You may need to update the commit message to avoid confusion.

Thanks for review, good catch, will fix the description.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/helix/TestControllerKMERegistrationFromMessageHeader.java
```diff
@@ -0,0 +1,130 @@
...
+
+    // Remove the latest schema from child controller's local value serializer and remove it from child colo's schema
+    // repository (ZK).
+    VeniceControllerWrapper leaderController = child.getLeaderController(clusterName);
+    KafkaValueSerializer valueSerializer =
+        leaderController.getController().getVeniceControllerService().getKafkaValueSerializer();
+    TestUtils.waitForNonDeterministicAssertion(30, TimeUnit.SECONDS, true, () -> {
+      Assert.assertTrue(
+          valueSerializer.getProtocolVersionSize() >= AvroProtocolDefinition.KAFKA_MESSAGE_ENVELOPE
+              .getCurrentProtocolVersion());
```
shuhui-liu, line 86: Just curious, why do we need this assertion?

internal/venice-common/src/main/java/com/linkedin/venice/system/store/ControllerClientBackedSystemSchemaInitializer.java
shuhui-liu: I see. I missed `s` in the variable name. Thanks for the explanation. It brings up another question: if a controller or server finds a new unknown KME schema in the message header, the version of the new schema will be larger than `protocolDefinition.getCurrentProtocolVersion()`. In this case, `schemaInLocalResources` will be null and the code here won't register the new KME schema. Please correct me if I missed something.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/helix/TestControllerKMERegistrationFromMessageHeader.java
shuhui-liu: Sure, it does no harm to add additional assertion if you think it will reduce the chance of race. Though I thought local protocol loading (i.e. add schemas from local resource into `protocolVersionToReader`) happens in `InternalAvroSpecificSerializer` constructor. Does `ClusterLeaderInitializationManager.execute` also trigger local protocol loading?

lluwm: Oh, I see the problem. I actually wanted to wait the schema to appear in the ZK, not in `protocolVersionToReader`, because we have a fall back mechanism to read from ZK if local `protocolVersionToReader` doesn't have it, I should have waited for it to appear in HelixReadWriteSchemaRepository instead of  `protocolVersionToReader`, does that make sense?

shuhui-liu: IIRC, you're referring to the following logic. After reading from ZK, it also puts the schema into `protocolVersionToReader` in `cacheDatumReader`. But I didn't get why waiting this fallback ensures that local protocol loading completes. I think the loading of **local** protocols finishes during the construction time. The fallback is triggered when one consumer does not have the schema that another producer has, which is not the case right after the setup of this integ test.
```
Schema newProtocolSchema = schemaReader.getValueSchema(protocolVersion);
if (newProtocolSchema == null) {
    throw new VeniceMessageException(
        "Received Protocol Version '" + protocolVersion + "' which is not currently known by " +
        this.getClass().getSimpleName() + ". A remote fetch was attempted, but the " +
        SchemaReader.class.getSimpleName() + " returned null" +
        ". The currently known Protocol Versions are: " + getCurrentlyLoadedProtocolVersions() + ".");
}
specificDatumReader = cacheDatumReader(protocolVersion, newProtocolSchema);
```

Waiting for the schema to appear in ZK is still beneficial though, which could help avoid the race that the latest schema is removed but then added back by leader controller initializer.

shuhui-liu (approved):

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/helix/TestControllerKMERegistrationFromMessageHeader.java
lluwm: Yes, you're right and it doesn't make sense to wait for the local ones stored in `protocolVersionToReader` and it doesn't make sense.
> I think the loading of local protocols finishes during the construction time.

This is exactly what I wanted to wait for and why. It's a mistake in the previous version to wait for a different one. :-) 
> Waiting for the schema to appear in ZK is still beneficial though, which could help avoid the race that the latest schema is removed but then added back by leader controller initializer.

PR #630, opened 2023-09-13
[server] Fix storage node read quota enforcement

xunyin8:
## [server] Fix storage node read quota enforcement
The PR fixed several issues in the `ReadQuotaEnforcementHandler.java`:

1. The `ReadQuotaEnforcementHandler` used to only keep the token bucket of the current version. As a result when FC metadata is updated asynchronously it's possible that it will query the server for a different "current" version. During this transition period we don't have the corresponding token bucket and quota is not enforced.

2. In the old `channelRead0` there is a bug when handling retry requests. Retry request will not enter the if block but enter the else if unintentionally and trigger `handleEnforcingAndNoBucket` on the request.

3. `updateQuota` used to blindly replace existing token bucket with a new one even if they are identical. Now it checks if it's different and only replace it if it's different to avoid unnecessary initialization of new token buckets.

4. Modified the old usage ratio metric since it's not very useful due to the bucket's large headroom for spike handling and we can get a more useful ratio by looking at the token requested and amortized refill per second. Added a new `quota_rcu_allowed_unintentionally` metric to track requests allowed due to missing token bucket.

5. Initialize `ServerQuotaTokenBucketStats` for the server node's token bucket to gain visibility on the storage level quota enforcement. 

## How was this PR tested?
Existing and new tests

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/config/VeniceServerConfig.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/config/VeniceServerConfig.java
@@ -508,7 +508,7 @@ public class VeniceServerConfig extends VeniceClusterConfig {
     serverCalculateQuotaUsageBasedOnPartitionsAssignmentEnabled =
         serverProperties.getBoolean(SEVER_CALCULATE_QUOTA_USAGE_BASED_ON_PARTITIONS_ASSIGNMENT_ENABLED, true);
 
-    nodeCapacityInRcu = serverProperties.getLong(SERVER_NODE_CAPACITY_RCU, 50000);
+    nodeCapacityInRcu = serverProperties.getLong(SERVER_NODE_CAPACITY_RCU, 100000);
     kafkaMaxPollRecords = serverProperties.getInt(SERVER_KAFKA_MAX_POLL_RECORDS, 100);
     kafkaPollRetryTimes = serverProperties.getInt(SERVER_KAFKA_POLL_RETRY_TIMES, 100);
     kafkaPollRetryBackoffMs = serverProperties.getInt(SERVER_KAFKA_POLL_RETRY_BACKOFF_MS, 0);
--- a/internal/venice-common/src/main/java/com/linkedin/venice/throttle/TokenBucket.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/throttle/TokenBucket.java
@@ -17,9 +17,10 @@ public class TokenBucket {
   private final float refillPerSecond;// only used for logging
   private final Clock clock;
   private final AtomicLong tokens;
-  private final AtomicLong tokensCountAfterLastRefill;
-  private final AtomicLong tokensConsumedSinceLastRefill;
+  private final AtomicLong tokensRequestedSinceLastRefill;
+  private volatile long previousRefillTime = 0;
   private volatile long nextUpdateTime;
+  private volatile double staleUsageRatio = 0;
 
   /**
    * This constructor should only be used by tests.  Application should not specify it's own instance of Clock
@@ -50,9 +51,9 @@ public class TokenBucket {
     this.clock = clock;
 
     tokens = new AtomicLong(capacity);
-    tokensCountAfterLastRefill = new AtomicLong(tokens.get());
-    tokensConsumedSinceLastRefill = new AtomicLong(0);
-    nextUpdateTime = clock.millis() + refillIntervalMs;
+    tokensRequestedSinceLastRefill = new AtomicLong(0);
+    previousRefillTime = clock.millis();
+    nextUpdateTime = previousRefillTime + refillIntervalMs;
 
     refillPerSecond = refillAmount / (float) refillUnit.toSeconds(refillInterval);
   }
@@ -76,8 +77,9 @@ public class TokenBucket {
   private boolean update() {
     if (clock.millis() > nextUpdateTime) {
       synchronized (this) {
-        if (clock.millis() > nextUpdateTime) {
-          long refillCount = (clock.millis() - nextUpdateTime) / refillIntervalMs + 1;
+        long timeNow = clock.millis();
+        if (timeNow > nextUpdateTime) {
+          long refillCount = (timeNow - nextUpdateTime) / refillIntervalMs + 1;
           long totalRefillAmount = refillCount * refillAmount;
           tokens.getAndAccumulate(totalRefillAmount, (existing, toAdd) -> {
             long newTokens = existing + toAdd;
@@ -87,9 +89,14 @@ public class TokenBucket {
               return newTokens;
             }
           });
-          tokensCountAfterLastRefill.set(tokens.get());
-          tokensConsumedSinceLastRefill.set(0);
-          nextUpdateTime += refillCount * refillIntervalMs;
+          if (previousRefillTime > 0) {
+            long timeSinceLastRefill = TimeUnit.MILLISECONDS.toSeconds(timeNow - previousRefillTime);
+            staleUsageRatio =
+                ((double) tokensRequestedSinceLastRefill.get() / (double) timeSinceLastRefill) / refillPerSecond;
+          }
+          previousRefillTime = timeNow;
+          tokensRequestedSinceLastRefill.set(0);
+          nextUpdateTime = timeNow + refillIntervalMs;
         }
       }
       return true;
@@ -112,9 +119,9 @@ public class TokenBucket {
   }
 
   private boolean noRetryTryConsume(long tokensToConsume) {
+    tokensRequestedSinceLastRefill.getAndAdd(tokensToConsume);
     long tokensThatWereAvailable = tokens.getAndAccumulate(tokensToConsume, (existing, toConsume) -> {
       if (toConsume <= existing) { // there are sufficient tokens
-        tokensConsumedSinceLastRefill.getAndAdd(toConsume);
         return existing - toConsume;
       } else {
         return existing; // insufficient tokens, do not consume any
@@ -129,5 +136,9 @@ public class TokenBucket {
 
   public float getAmortizedRefillPerSecond() {
     return refillPerSecond;
+  }
+
+  public double getStaleUsageRatio() {
+    return staleUsageRatio;
   }
 }
--- a/internal/venice-common/src/main/java/com/linkedin/venice/throttle/TokenBucket.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/throttle/TokenBucket.java
@@ -54,8 +54,7 @@ public class TokenBucket {
     tokensConsumedSinceLastRefill = new AtomicLong(0);
     nextUpdateTime = clock.millis() + refillIntervalMs;
 
-    float refillIntervalSeconds = refillIntervalMs / (float) 1000;
-    refillPerSecond = refillAmount / refillIntervalSeconds;
+    refillPerSecond = refillAmount / (float) refillInterval;
   }
 
   /**
@@ -106,14 +105,6 @@ public class TokenBucket {
   public long getStaleTokenCount() {
     // TODO: maybe update the token after getting the stale token count
     return tokens.get();
-  }
-
-  /**
-   * This method does not call #update(), so it is only accurate as of the last time #tryConsume() and update() was called
-   * @return ratio between number of tokens consumed since last refill over the total token count after the last refill
-   */
-  public double getStaleUsageRatio() {
-    return (double) tokensConsumedSinceLastRefill.get() / (double) tokensCountAfterLastRefill.get();
   }
 
   public boolean tryConsume(long tokensToConsume) {
--- a/internal/venice-common/src/main/java/com/linkedin/venice/throttle/TokenBucket.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/throttle/TokenBucket.java
@@ -54,7 +54,7 @@ public class TokenBucket {
     tokensConsumedSinceLastRefill = new AtomicLong(0);
     nextUpdateTime = clock.millis() + refillIntervalMs;
 
-    refillPerSecond = refillAmount / (float) refillInterval;
+    refillPerSecond = refillAmount / (float) refillUnit.toSeconds(refillInterval);
   }
 
   /**
--- a/services/venice-server/src/main/java/com/linkedin/venice/listener/ReadQuotaEnforcementHandler.java
+++ b/services/venice-server/src/main/java/com/linkedin/venice/listener/ReadQuotaEnforcementHandler.java
@@ -388,7 +388,7 @@ public class ReadQuotaEnforcementHandler extends SimpleChannelInboundHandler<Rou
 
   @Override
   public void onExternalViewChange(PartitionAssignment partitionAssignment) {
-    updateQuota(partitionAssignment);
+    // Ignore this event since the partition assignment triggered by EV won't contain the right states.
   }
 
   @Override
--- a/services/venice-server/src/main/java/com/linkedin/venice/listener/ReadQuotaEnforcementHandler.java
+++ b/services/venice-server/src/main/java/com/linkedin/venice/listener/ReadQuotaEnforcementHandler.java
@@ -22,6 +22,7 @@ import com.linkedin.venice.pushmonitor.ReadOnlyPartitionStatus;
 import com.linkedin.venice.routerapi.ReplicaState;
 import com.linkedin.venice.stats.AbstractVeniceAggStats;
 import com.linkedin.venice.stats.AggServerQuotaUsageStats;
+import com.linkedin.venice.stats.ServerQuotaTokenBucketStats;
 import com.linkedin.venice.throttle.TokenBucket;
 import com.linkedin.venice.utils.ExpiringSet;
 import com.linkedin.venice.utils.concurrent.VeniceConcurrentHashMap;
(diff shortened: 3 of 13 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/config/VeniceServerConfig.java
```diff
@@ -508,7 +508,7 @@ public VeniceServerConfig(VeniceProperties serverProperties, Map<String, Map<Str
     serverCalculateQuotaUsageBasedOnPartitionsAssignmentEnabled =
         serverProperties.getBoolean(SEVER_CALCULATE_QUOTA_USAGE_BASED_ON_PARTITIONS_ASSIGNMENT_ENABLED, true);
 
-    nodeCapacityInRcu = serverProperties.getLong(SERVER_NODE_CAPACITY_RCU, 50000);
+    nodeCapacityInRcu = serverProperties.getLong(SERVER_NODE_CAPACITY_RCU, 100000);
```
sixpluszero, line 511: Is the default value change intentional?

services/venice-server/src/main/java/com/linkedin/venice/listener/ReadQuotaEnforcementHandler.java
```diff
@@ -100,9 +115,10 @@ public static int getRcu(RouterRequest request) {
       case SINGLE_GET:
         return 1;
       case MULTI_GET:
-        return request.getKeyCount();
+      case MULTI_GET_STREAMING:
```
sixpluszero, line 118: Hmm so previously STREAMING request  will return Int.MAX?

services/venice-server/src/main/java/com/linkedin/venice/listener/ReadQuotaEnforcementHandler.java
```diff
@@ -188,30 +204,34 @@ public void channelRead0(ChannelHandlerContext ctx, RouterRequest request) {
     String storeName = request.getStoreName();
     Store store = storeRepository.getStore(storeName);
 
-    if (checkStoreNull(ctx, request, null, false, store)) {
+    if (!enforcing || checkStoreNull(ctx, request, null, false, store)) {
```
sixpluszero, line 207: Just double check - what does this enforcing mean? I see the setEnforcing() method is not used, which means it is always true?

services/venice-server/src/main/java/com/linkedin/venice/listener/ReadQuotaEnforcementHandler.java
```diff
@@ -407,8 +427,20 @@ private void updateQuota(PartitionAssignment partitionAssignment) {
...
-    storeVersionBuckets.put(topic, newStoreBucket); // put is atomic, so this method is thread-safe
+    storeVersionBuckets.compute(topic, (k, v) -> {
+      long newRefillAmount = calculateRefillAmount(quotaInRcu, thisNodeQuotaResponsibility);
+      if (v == null || v.getAmortizedRefillPerSecond() * enforcementIntervalSeconds != newRefillAmount) {
+        // only replace the existing bucket if the difference is greater than 1
+        return tokenBucketfromRcuPerSecond(quotaInRcu, thisNodeQuotaResponsibility);
+      } else {
+        return v;
+      }
+    }); // put is atomic, so this method is thread-safe
```
sixpluszero, line 438: nit: comment is outdated "put"->"compute". Or just remove it as it is kind of obvious

services/venice-server/src/main/java/com/linkedin/venice/listener/ReadQuotaEnforcementHandler.java
sixpluszero: Would like to get some explanation for this as it is basically blocking me understanding the change :)

gaojieliu:
>> Removed the old usage ratio metric since it's not very useful due to the bucket's large headroom for spike handling and we can get a more useful ratio by looking at the QPS and bucket stats. Instead, added quota_rcu_allowed_unintentionally metric to track requests allowed due to missing token bucket.

So new metrics will cover the qps beyond quota limit?

I feel the usage ratio is still useful even the current quota mechanism can tolerate spike since this behavior is not deterministic since it is a best effort.
If the existing stores use up the per-server quota, these excessive requests will be rejected.

services/venice-server/src/main/java/com/linkedin/venice/listener/ReadQuotaEnforcementHandler.java
```diff
@@ -188,30 +204,34 @@ public void channelRead0(ChannelHandlerContext ctx, RouterRequest request) {
...
         // TODO: check if extra node capacity and can still process this request out of quota
-        if (handleTooManyRequests(ctx, request, null, store, rcu, false))
-          return;
+        return;
+      }
+    } else {
+      // If this happens it is probably due to a short-lived race condition where the resource is being accessed before
+      // the bucket is allocated. The request will be allowed based on node/server capacity so emit metrics accordingly.
+      stats.recordAllowedUnintentionally(storeName, rcu);
+      if (!noBucketStores.contains(request.getResourceName())) {
```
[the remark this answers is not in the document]
xunyin8: If tokenBucket is null nothing is enforced other than the server node level quota. Hence we don't differentiate retry request from regular request when enforcing server node level quota.

services/venice-server/src/main/java/com/linkedin/venice/listener/ReadQuotaEnforcementHandler.java
```diff
@@ -188,30 +204,34 @@ public void channelRead0(ChannelHandlerContext ctx, RouterRequest request) {
...
+          && handleTooManyRequests(ctx, request, null, store, rcu, false)) {
+        // Enforce store version quota for non-retry requests.
         // TODO: check if extra node capacity and can still process this request out of quota
-        if (handleTooManyRequests(ctx, request, null, store, rcu, false))
-          return;
+        return;
+      }
+    } else {
+      // If this happens it is probably due to a short-lived race condition where the resource is being accessed before
+      // the bucket is allocated. The request will be allowed based on node/server capacity so emit metrics accordingly.
```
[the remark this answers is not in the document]
xunyin8: Right, so we will maintain multiple version's token buckets (all versions that's still visible in the store's zk). FC shouldn't be and can't afford to be stuck in version switch forever.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/config/VeniceServerConfig.java
xunyin8: Yes, 50k is too low for server node level qps as default IMO.

services/venice-server/src/main/java/com/linkedin/venice/listener/ReadQuotaEnforcementHandler.java
xunyin8: Right... it looks like the default is to reject quota for unknown request types since it may not know how to evaluate its cost properly and we forgot to update here when we introduced new request types like `STREAMING`. Let me know if we should change the default.

services/venice-server/src/main/java/com/linkedin/venice/listener/ReadQuotaEnforcementHandler.java
xunyin8: Yes, it should always be enforced. Let me remove it. I think this is legacy code where initially the author wanted to rollout in phases where there is a mode to log/report quota but not actually reject them.

internal/venice-common/src/main/java/com/linkedin/venice/throttle/TokenBucket.java
```diff
@@ -54,8 +54,7 @@ public TokenBucket(long capacity, long refillAmount, long refillInterval, TimeUn
     tokensConsumedSinceLastRefill = new AtomicLong(0);
     nextUpdateTime = clock.millis() + refillIntervalMs;
 
-    float refillIntervalSeconds = refillIntervalMs / (float) 1000;
-    refillPerSecond = refillAmount / refillIntervalSeconds;
+    refillPerSecond = refillAmount / (float) refillInterval;
```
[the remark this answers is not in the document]
xunyin8: Good catch, let me fix this.

xunyin8:
> > > Removed the old usage ratio metric since it's not very useful due to the bucket's large headroom for spike handling and we can get a more useful ratio by looking at the QPS and bucket stats. Instead, added quota_rcu_allowed_unintentionally metric to track requests allowed due to missing token bucket.
> 
> So new metrics will cover the qps beyond quota limit?
> 
> I feel the usage ratio is still useful even the current quota mechanism can tolerate spike since this behavior is not deterministic since it is a best effort. If the existing stores use up the per-server quota, these excessive requests will be rejected.

But we can also arrive at a ratio by looking at the corresponding token bucket stats and qps. Do we really need a ratio that's usually low during steady state but can quickly go to 1 or even exceed 1 during spike scenarios? To me it's only worth it if we plan to put some alerting on it, maybe if we return a simpler ratio. qps / Refill per second allocated to this node, what do you think?

services/venice-server/src/main/java/com/linkedin/venice/listener/ReadQuotaEnforcementHandler.java
```diff
@@ -300,12 +315,13 @@ public boolean handleTooManyRequests(
...
 
-  public void handleEnforcingAndNoBucket(RouterRequest request) {
-    LOGGER.warn(
-        "Request for resource: {} but no TokenBucket for that resource. Not yet enforcing quota",
-        request.getResourceName());
+  /**
+   * This method and the expiring set noBucketStores is only used to throttle the logging of such event
+   */
+  public void handleEnforcingAndNoBucket(String resourceName) {
+    LOGGER.warn("Request for resource: {} but no TokenBucket for that resource. Not yet enforcing quota", resourceName);
```
majisourav99, line 322: could we add a rate limiter log for this? during the edge case there could be many such cases.

services/venice-server/src/main/java/com/linkedin/venice/listener/ReadQuotaEnforcementHandler.java
```diff
@@ -192,26 +207,30 @@ public void channelRead0(ChannelHandlerContext ctx, RouterRequest request) {
...
         // TODO: check if extra node capacity and can still process this request out of quota
-        if (handleTooManyRequests(ctx, request, null, store, rcu, false))
-          return;
+        return;
+      }
+    } else {
+      // If this happens it is probably due to a short-lived race condition where the resource is being accessed before
+      // the bucket is allocated. The request will be allowed based on node/server capacity so emit metrics accordingly.
+      stats.recordAllowedUnintentionally(storeName, rcu);
+      if (!noBucketStores.contains(request.getResourceName())) {
```
majisourav99, line 231: may use `isEnforcingAndNoBucketStoreContainsResource` ?

services/venice-server/src/main/java/com/linkedin/venice/listener/ReadQuotaEnforcementHandler.java
xunyin8: Good call, will do.

services/venice-server/src/main/java/com/linkedin/venice/listener/ReadQuotaEnforcementHandler.java
xunyin8: Will use `isNoBucketStoreContainsResource`, renamed the method since we removed the enforcing flag to avoid confusion.

services/venice-server/src/main/java/com/linkedin/venice/listener/ReadQuotaEnforcementHandler.java
xunyin8: Actually this is already rate limited to 30 seconds per resource with the `noBucketStores` expiring set.

gaojieliu:
> But we can also arrive at a ratio by looking at the corresponding token bucket stats and qps. Do we really need a ratio that's usually low during steady state but can quickly go to 1 or even exceed 1 during spike scenarios? To me it's only worth it if we plan to put some alerting on it, maybe if we return a simpler ratio. qps / Refill per second allocated to this node, what do you think?

Sorry for the late reply.
So what will be the new alerting strategy? 
Will we still come up with a ratio metric based on the qps and token bucket stats and setup alert against it as today?

xunyin8:
> > But we can also arrive at a ratio by looking at the corresponding token bucket stats and qps. Do we really need a ratio that's usually low during steady state but can quickly go to 1 or even exceed 1 during spike scenarios? To me it's only worth it if we plan to put some alerting on it, maybe if we return a simpler ratio. qps / Refill per second allocated to this node, what do you think?
> 
> Sorry for the late reply. So what will be the new alerting strategy? Will we still come up with a ratio metric based on the qps and token bucket stats and setup alert against it as today?

Yes, unless we prefer to have an explicit metric instead of a derived one. I can also emit one directly from the bucket stats. Taking the ratio of token requested/refill amount over the refill interval.

gaojieliu:
> > > But we can also arrive at a ratio by looking at the corresponding token bucket stats and qps. Do we really need a ratio that's usually low during steady state but can quickly go to 1 or even exceed 1 during spike scenarios? To me it's only worth it if we plan to put some alerting on it, maybe if we return a simpler ratio. qps / Refill per second allocated to this node, what do you think?
> > 
> > 
> > Sorry for the late reply. So what will be the new alerting strategy? Will we still come up with a ratio metric based on the qps and token bucket stats and setup alert against it as today?
> 
> Yes, unless we prefer to have an explicit metric instead of a derived one. I can also emit one directly from the bucket stats. Taking the ratio of token requested/refill amount over the refill interval.

If we can use the derived metric to achieve the same result of direct metric, that would be fine.

services/venice-server/src/main/java/com/linkedin/venice/listener/ReadQuotaEnforcementHandler.java
```diff
@@ -159,8 +174,8 @@ public final void init() {
       LOGGER.error(
           "Null resource assignment from HelixCustomizedViewOfflinePushRepository in ReadQuotaEnforcementHandler");
     } else {
-      for (String resource: customizedViewRepository.getResourceAssignment().getAssignedResources()) {
-        this.onExternalViewChange(customizedViewRepository.getPartitionAssignments(resource));
+      for (String resource: resourceAssignment.getAssignedResources()) {
+        this.onExternalViewChange(resourceAssignment.getPartitionAssignment(resource));
```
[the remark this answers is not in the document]
xunyin8: I think that's because the code originally was written before CV existed. Let me see if it's feasible to completely rely on CV as trigger and not both.

services/venice-server/src/main/java/com/linkedin/venice/listener/ReadQuotaEnforcementHandler.java
```diff
@@ -192,26 +207,30 @@ public void channelRead0(ChannelHandlerContext ctx, RouterRequest request) {
...
 
     /**
-     * First check store bucket for capacity; don't throttle retried request at store version level
+     * First check store bucket for capacity don't throttle retried request at store version level
      */
     TokenBucket tokenBucket = storeVersionBuckets.get(request.getResourceName());
-    if (tokenBucket != null && !request.isRetryRequest()) {
-      if (!tokenBucket.tryConsume(rcu)) {
+    if (tokenBucket != null) {
+      if (!request.isRetryRequest() && !tokenBucket.tryConsume(rcu)
```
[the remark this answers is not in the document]
xunyin8: I can try to change it to something that's closer to the original style but the semantic of the old code is incorrect.

services/venice-server/src/main/java/com/linkedin/venice/listener/ReadQuotaEnforcementHandler.java
```diff
@@ -481,7 +483,7 @@ public void handleStoreChanged(Store store) {
          * a the future version will fail in most cases, because the new topic is not in the external view at all.
          *
          */
-        this.onExternalViewChange(customizedViewRepository.getPartitionAssignments(topic));
+        this.onCustomizedViewChange(customizedViewRepository.getPartitionAssignments(topic));
```
gaojieliu, line 486: Is it possible to remove the function of `onExternalViewChange` or just throw exception or do nothing? Since the partition assignment triggered by EV won't contain the right state.

services/venice-server/src/main/java/com/linkedin/venice/listener/ReadQuotaEnforcementHandler.java
```diff
@@ -218,6 +218,8 @@ public void channelRead0(ChannelHandlerContext ctx, RouterRequest request) {
      */
     TokenBucket tokenBucket = storeVersionBuckets.get(request.getResourceName());
     if (tokenBucket != null) {
+      // Can always emit the stale usage ratio first because it's using numbers from the previous refill
+      stats.recordReadQuotaUsageRatio(storeName, tokenBucket.getStaleUsageRatio());
```
gaojieliu, line 222: Why called it `stale usage ratio`? Can you elaborate?

xunyin8: It's stale because it was calculated using previous refill interval's amortized requested QPS divided by refill amount. Depending on the refill interval it could be 30 seconds or a minute behind the current usage ratio.

services/venice-server/src/main/java/com/linkedin/venice/listener/ReadQuotaEnforcementHandler.java
xunyin8: Will make it do nothing, we can't remove it unless we remove it from the interface `RoutingDataChangedListener` or make it abstract.

gaojieliu (approved):
Thanks for the fix!

PR #634, opened 2023-09-18
[controller] Return null instead of Collections.emptySet()

majisourav99:
## Return null instead of Collections.emptySet() when there is no disable replica found.

In `isReplicaDisabled` returning emptySet will cause `java.lang.UnsupportedOperationException: null` exception to be thrown if we try to add any element to it
Instead return null so that it will later create a new HashSet when needed.

## How was this PR tested?

GHCI

### DIFF
--- a/services/venice-controller/src/main/java/com/linkedin/venice/pushmonitor/AbstractPushMonitor.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/pushmonitor/AbstractPushMonitor.java
@@ -765,11 +765,10 @@ public abstract class AbstractPushMonitor
           if (helixMap.containsKey(kafkaTopic)) {
             return helixMap.get(kafkaTopic).stream().map(HelixUtils::getPartitionId).collect(Collectors.toSet());
           } else {
-            return Collections.emptySet();
+            return null;
           }
         });
-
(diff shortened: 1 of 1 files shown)


services/venice-controller/src/main/java/com/linkedin/venice/pushmonitor/AbstractPushMonitor.java
```diff
@@ -765,11 +765,10 @@ public boolean isReplicaDisabled(String instance, int partitionId) {
           if (helixMap.containsKey(kafkaTopic)) {
             return helixMap.get(kafkaTopic).stream().map(HelixUtils::getPartitionId).collect(Collectors.toSet());
```
FelixGV, line 766: Would you mind also changing this into a single lookup? e.g.

```
List<String> partitionsList = helixMap.get(kafkaTopic);
return partitionsList == null
    ? null
    : partitionsList.stream().map(HelixUtils::getPartitionId).collect(Collectors.toSet());
```

FelixGV (commented):
Thanks for the quick fix! Left a minor comment...

FelixGV (approved):
Thanks Sourav!

PR #635, opened 2023-09-18
Changing ingestion consumption check avoiding race condition.

haoxu07:

Previously, when OFFLINE => DROP transition happen, we will check if we can remove the partition from disk only based on partition consumption state exist or not. However, it is also possible that there are some pending state transitions in consume action queue of the `StoreIngestionTask`.  If we drop the partition and there is a pending SUBSCRIBE action in consumption queue, then it will cause data persistence failure due to partition not existed in disk. To avoid this, we add pending queue checking for the partition about to be deleted during the partition consumption check. 

## How was this PR tested?

internal CI

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
@@ -3321,10 +3321,10 @@ public abstract class StoreIngestionTask implements Runnable, Closeable {
   /**
    * To check whether the given partition is still consuming message from Kafka
    */
-  public boolean isPartitionConsuming(int userPartition) {
-    Boolean subPartitionConsumptionStateExist =
+  public boolean isPartitionConsumingOrHasPendingIngestionAction(int userPartition) {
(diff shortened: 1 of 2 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -3325,7 +3324,10 @@ public void disableMetricsEmission() {
    * To check whether the given partition is still consuming message from Kafka
    */
   public boolean isPartitionConsuming(int userPartition) {
-    return amplificationFactorAdapter.meetsAny(userPartition, partitionConsumptionStateMap::containsKey);
+    Boolean subPartitionConsumptionStateExist =
+        amplificationFactorAdapter.meetsAny(userPartition, partitionConsumptionStateMap::containsKey);
+    Boolean pendingPartitionIngestionAction = hasPendingPartitionIngestionAction(userPartition);
```
lluwm, line 3329: LGTM! nit, Boolean -> boolean? and same for L3327

lluwm (approved):

PR #636, opened 2023-09-19
[fc] Log exception when FC increment unhealthy request metric

m-nagarajan:
## Summary
Log exception with a redundant filter when fast client increments unhealthy request metric

## How was this PR tested?
GH CI

### DIFF
--- a/clients/venice-client/src/main/java/com/linkedin/venice/fastclient/StatsAvroGenericStoreClient.java
+++ b/clients/venice-client/src/main/java/com/linkedin/venice/fastclient/StatsAvroGenericStoreClient.java
@@ -10,6 +10,7 @@ import com.linkedin.venice.fastclient.stats.ClusterStats;
 import com.linkedin.venice.fastclient.stats.FastClientStats;
 import com.linkedin.venice.read.RequestType;
 import com.linkedin.venice.utils.LatencyUtils;
+import com.linkedin.venice.utils.RedundantExceptionFilter;
 import com.linkedin.venice.utils.Time;
 import java.util.Collections;
 import java.util.HashMap;
@@ -17,6 +18,7 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.TimeUnit;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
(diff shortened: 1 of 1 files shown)


clients/venice-client/src/main/java/com/linkedin/venice/fastclient/StatsAvroGenericStoreClient.java
```diff
@@ -34,6 +36,8 @@ public class StatsAvroGenericStoreClient<K, V> extends DelegatingAvroStoreClient
 
   private final int maxAllowedKeyCntInBatchGetReq;
   private final boolean useStreamingBatchGetAsDefault;
+  private final RedundantExceptionFilter redundantExceptionFilter =
+      new RedundantExceptionFilter(RedundantExceptionFilter.DEFAULT_BITSET_SIZE, TimeUnit.SECONDS.toMillis(1));
```
huangminchn, line 40: Can we change it to one log per minute?

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/StatsAvroGenericStoreClient.java
```diff
@@ -163,6 +167,13 @@ private <R> CompletableFuture<R> recordRequestMetrics(
          * considered an unhealthy request: so only incrementing a subset of metrics.
          */
         exceptionReceived = true;
+        if (!redundantExceptionFilter.isRedundantException(throwable.getMessage())) {
```
huangminchn, line 170: I am not sure if this message content is consistent. Can we just use the store name, or does it support passing nothing and always print one log per minute?

huangminchn (changes requested):

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/StatsAvroGenericStoreClient.java
m-nagarajan: changed it to default which is one log per minute.

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/StatsAvroGenericStoreClient.java
m-nagarajan: That is a good point. To start with, we can get all unique logs per minute and once validated and fixed, we can remove this or make it a stricter one log per minute.

PR #637, opened 2023-09-19
[router] Fetch dictionary in response decompressor

majisourav99:
## Fetch dictionary in response decompressor

During router startup all current store version's compression dictionary are fetched asynchronously, meanwhile if there is a read request that needs to served it will throw the following 

 `Caused by: com.linkedin.venice.client.exceptions.VeniceClientHttpException: http status: 500, 
   com.linkedin.venice.exceptions.VeniceException: Compressor not available for resource. Dictionary not downloaded.`

This PR will try to fetch the dictionary for the given resource in blocking manner to serve the request in such cases. 
 Many CI tests should also get hardened by this.

## How was this PR tested?

CI

### DIFF
--- a/services/venice-router/src/main/java/com/linkedin/venice/router/api/DictionaryRetrievalService.java
+++ b/services/venice-router/src/main/java/com/linkedin/venice/router/api/DictionaryRetrievalService.java
@@ -231,9 +231,9 @@ public class DictionaryRetrievalService extends AbstractVeniceService {
     Instance instance = getOnlineInstance(kafkaTopic);
 
     if (instance == null) {
-      CompletableFuture<byte[]> future = new CompletableFuture<>();
-      future.completeExceptionally(new VeniceException("No online storage instance for resource: " + kafkaTopic));
-      return future;
+      return CompletableFuture.supplyAsync(() -> {
+        throw new VeniceException("No online storage instance for resource: " + kafkaTopic);
+      }, executor);
     }
 
     String instanceUrl = instance.getUrl(sslFactory.isPresent());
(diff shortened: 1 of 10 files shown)


services/venice-router/src/main/java/com/linkedin/venice/router/api/DictionaryRetrievalService.java
```diff
@@ -216,6 +216,16 @@ public DictionaryRetrievalService(
     this.dictionaryRetrieverThread = new Thread(runnable);
   }
 
+  public byte[] getDictionaryForKafkaTopic(String kafkaTopic) {
```
FelixGV, line 219: I wonder if this function should check whether there is already a future for downloading this dictionary in `downloadingDictionaryFutures`, and if there is, perhaps wait on that instead of kicking off a new one?

I wonder If we may end up with a thundering herd problem otherwise...

On the other hand, if the one future that's in flight is unlucky and hits an unhealthy server that hangs a while before responding, then we could end up having many requests tied up behind the same future, which may be not much better than the thundering herd.

I'm not sure what is better. What do you think?

FelixGV (commented):
Thanks Sourav! I left a question...

Also, another thought is that I wonder if the `DictionaryRetrievalService` might have too much asynchrony in it... Let's consider the times when we need to fetch dictionaries: at startup time and at version swap time... shall we make the dictionary fetching in both of these occasions a blocking operation? IOW, why not delay the startup and the version swap until required dictionaries are fetched? Does the asynchrony really gets us anything?

services/venice-router/src/main/java/com/linkedin/venice/router/api/DictionaryRetrievalService.java
majisourav99: Given we select random host on each request (`getOnlineInstance`) the chances for hitting a same server would be low. 
We should definitely check the existing future queue and if found wait that future. If that future times out we can kick off another future (which hopefully will select another host)?

PR #638, opened 2023-09-19
[all] Fix race condition in RouterBackedSchemaReader when there are concurrent calls to check schema freshness

nisargthakkar:
## Fix race condition in `RouterBackedSchemaReader` when there are concurrent calls to check schema freshness

There is a race condition in the schema refresh logic in `RouterBackedSchemaReader` where if there are concurrent calls to fetch update schemas, then the `loadUpdateSchemas` flag is flipped earlier than it should have - causing future calls to assume that the schemas have already been fetched and to not try to take a lock to fetch schemas. However, since the schemas themselves have not yet been fetched, it incorrectly returns a `null` response.

## How was this PR tested?

GH CI

### DIFF
--- a/clients/venice-thin-client/src/main/java/com/linkedin/venice/client/schema/RouterBackedSchemaReader.java
+++ b/clients/venice-thin-client/src/main/java/com/linkedin/venice/client/schema/RouterBackedSchemaReader.java
@@ -137,8 +137,12 @@ public class RouterBackedSchemaReader implements SchemaReader {
     if (valueSchemaRefreshPeriod.toMillis() > 0) {
       this.refreshSchemaExecutor =
           Executors.newSingleThreadScheduledExecutor(new DaemonThreadFactory("schema-refresh"));
+      Runnable schemaRefresher = () -> this.ensureSchemasAreRefreshed(loadUpdateSchemas.get(), true);
+      // Fetch schemas once on start up
+      schemaRefresher.run();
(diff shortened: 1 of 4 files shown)


clients/venice-thin-client/src/main/java/com/linkedin/venice/client/schema/RouterBackedSchemaReader.java
```diff
@@ -268,8 +272,11 @@ public void close() throws IOException {
    */
   private void ensureSchemasAreRefreshed(boolean needsDerivedSchemas, boolean forceSchemaRefresh) {
     ConcurrencyUtils.executeUnderConditionalLock(() -> {
-      loadUpdateSchemas.compareAndSet(false, needsDerivedSchemas);
-      this.refreshAllSchemas();
+      boolean shouldFetchUpdateSchemas = loadUpdateSchemas.get() || needsDerivedSchemas;
+      this.refreshAllSchemas(shouldFetchUpdateSchemas);
+
+      // Only update this at the end to prevent the conditional check to incorrectly pass for other concurrent threads
+      loadUpdateSchemas.compareAndSet(false, shouldFetchUpdateSchemas);
```
FelixGV, line 279: So this is the crux of the fix, then, huh?

FelixGV (approved):
LGTM! I think an extra comment on the new test would be good though.

FelixGV (approved):

PR #640, opened 2023-09-20
[server] Filter out junk metric values for nearline latency sensors 

asaxena76:
## Filter out junk metric values for nearline latency sensors  

Nearline latency sensors will emit -INF values when no data is present for a partition. This is a problem when aggregating data across time and instane dimensions. Emitting zero is safe for this metric as long as it is not used to detect inactivity.

## How was this PR tested?

Unit tests

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/IngestionStats.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/IngestionStats.java
@@ -488,19 +488,19 @@ public class IngestionStats {
   }
 
   public double getNearlineProducerToLocalBrokerLatencyAvg() {
-    return nearlineProducerToLocalBrokerLatencySensor.getAvg();
+    return unAvailableToZero(nearlineProducerToLocalBrokerLatencySensor.getAvg());
   }
 
   public double getNearlineProducerToLocalBrokerLatencyMax() {
-    return nearlineProducerToLocalBrokerLatencySensor.getMax();
+    return unAvailableToZero(nearlineProducerToLocalBrokerLatencySensor.getMax());
   }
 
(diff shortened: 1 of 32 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/IngestionStats.java
```diff
@@ -511,4 +511,10 @@ public void recordNearlineLocalBrokerToReadyToServeLatency(double value, long cu
     nearlineLocalBrokerToReadyToServeLatencySensor.record(value, currentTimeMs);
   }
 
+  public static double unAvailableToZero(double value) {
+    /* When data is unavailable, return 0 instead of NaN or Infinity. Some metrics are initialized to -INF.
+     This can cause problems when metrics are aggregated. Use only when zero makes semantic sense.
+    */
```
huangminchn, line 517: What if we initialize these metrics to 0? Do you think we would still see infinite in this case?

asaxena76:
I think the default is coming from the Max class inside WritePathLatencySensor . https://github.com/linkedin/venice/blob/9eff266e5443ea4633e56eabbbf899fde6c49b44/clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/WritePathLatencySensor.java#L15 . That class inits this  to -INF. Keep in mind that we don't always see this problem . Only in certain scenarios. In any case I don't think it will be wise to change either the WritePathLatencySensor . I would rather sanitize the data at retrieval time to make sure it is how we want rather than depend on the metric class

huangminchn (approved):
Thanks Amit!

asaxena76:
Superceded by **https://github.com/linkedin/venice/pull/674**

PR #641, opened 2023-09-21
[controller] Add new controller cluster config to enable partial update by default when converting new store from batch to hybrid

sixpluszero:
## [controller] Add new controller cluster config to enable partial update by default when converting new store from batch to hybrid
This PR adds two new config to enable partial update by default when converting a store from batch to hybrid:
"enable.partial.update.for.hybrid.active.active.user.stores"
"enable.partial.update.for.hybrid.non.active.active.user.stores"

If the update request does not explicitly enable/disable partial update, then it will rely on the above 2 configs to determine if a store should enable partial update. If the resolution determines it needs partial update, it will also enable chunking (and RMD chunking, if A/A is enabled).

## How was this PR tested?

Add a new integration test to validate the cluster level config.

### DIFF
--- a/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
@@ -2491,7 +2491,7 @@ public class VeniceParentHelixAdmin implements Admin {
           this,
           clusterName,
           storeName,
-          params.getWriteComputationEnabled(),
+          writeComputationEnabled,
           setStore,
           storeBeingConvertedToHybrid);
       if (partialUpdateConfigUpdated) {
@@ -2501,27 +2501,24 @@ public class VeniceParentHelixAdmin implements Admin {
 
       // Update Chunking config.
       boolean chunkingConfigUpdated = ParentControllerConfigUpdateUtils
-          .checkAndMaybeApplyChunkingConfigChange(this, clusterName, storeName, params.getChunkingEnabled(), setStore);
+          .checkAndMaybeApplyChunkingConfigChange(this, clusterName, storeName, chunkingEnabled, setStore);
       if (chunkingConfigUpdated) {
         updatedConfigsList.add(CHUNKING_ENABLED);
       }
 
       // Update RMD Chunking config.
-      boolean rmdChunkingConfigUpdated = ParentControllerConfigUpdateUtils.checkAndMaybeApplyRmdChunkingConfigChange(
-          this,
-          clusterName,
-          storeName,
-          params.getRmdChunkingEnabled(),
-          setStore);
+      boolean rmdChunkingConfigUpdated = ParentControllerConfigUpdateUtils
+          .checkAndMaybeApplyRmdChunkingConfigChange(this, clusterName, storeName, rmdChunkingEnabled, setStore);
       if (rmdChunkingConfigUpdated) {
         updatedConfigsList.add(RMD_CHUNKING_ENABLED);
       }
 
+      // Validate Amplification Factor config based on latest A/A and partial update status.
       if ((setStore.getActiveActiveReplicationEnabled() || setStore.getWriteComputationEnabled())
           && updatedPartitionerConfig.getAmplificationFactor() > 1) {
         throw new VeniceHttpException(
             HttpStatus.SC_BAD_REQUEST,
-            "Non-default amplification factor is not compatible with active-active replication and/or write compute.",
+            "Non-default amplification factor is not compatible with active-active replication and/or partial update.",
             ErrorType.BAD_REQUEST);
       }
 
--- a/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
@@ -2482,24 +2482,6 @@ public class VeniceParentHelixAdmin implements Admin {
       }
 
       /**
-       * By default, parent controllers will not try to replicate the unchanged store configs to child controllers;
-       * an updatedConfigSet will be used to represent which configs are updated by users.
-       */
-      setStore.replicateAllConfigs = replicateAllConfigs;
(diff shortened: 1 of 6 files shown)


services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
```diff
@@ -2513,12 +2494,48 @@ public void updateStore(String clusterName, String storeName, UpdateStoreQueryPa
...
         // Compute schema generation may fail due to some reasons. If that happens, abort the store update process.
         addWriteComputeSchemaForStore(clusterName, storeName, true);
+      } else {
+        /**
+         * If a store: (1) Is being converted to hybrid (2) Is not partial update enabled for now. (3) Does not change
+         * partial update config in this request.
+         * It means partial update is not enabled, and there is no explict intention to change it. In this case, we will
+         * look up cluster config and based on the replication policy to enable partial update implicitly.
+         */
+        final boolean shouldEnablePartialUpdateBasedOnClusterConfig =
```
gaojieliu, line 2511: Currently, the logic is a little weird and I think it is also buggy..
For example, if someone disables partial update, and it will execute the `else` branch, which may re-enable partial update..

We should have a section to infer whether we should enable partial update or not.
```
If partial update param is explicitly set, Controller should follow it.
If not, Controller will use the cluster config to decide whether to enable partial update or not when `storeBeingConvertedToHybrid` is `true`.

When partial update needs to be enabled for the first time because of the above logic, Controller will try to generate update schemas.
```
This might be more clear.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
```diff
@@ -2513,12 +2494,48 @@ public void updateStore(String clusterName, String storeName, UpdateStoreQueryPa
...
+            storeBeingConvertedToHybrid && (setStore.activeActiveReplicationEnabled
+                ? clusterConfig.isEnablePartialUpdateForHybridActiveActiveUserStores()
+                : clusterConfig.isEnablePartialUpdateForHybridNonActiveActiveUserStores());
+        if (!writeComputationEnabled.isPresent() && !currStore.isWriteComputationEnabled()
+            && shouldEnablePartialUpdateBasedOnClusterConfig) {
+          try {
+            addWriteComputeSchemaForStore(clusterName, storeName, true);
+            partialUpdateJustEnabled = true;
+            setStore.writeComputationEnabled = true;
+            updatedConfigSet.add(WRITE_COMPUTATION_ENABLED);
```
gaojieliu, line 2521: What is the reason to change from `list` to `set` for `updatedConfig`?

sixpluszero: Eh. Because in my updated logic, some feature might get enabled twice (I think it is chunking, which got enabled by my new logic again maybe, and the list should contain two same params.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
sixpluszero: Yeah, I do agree about this, and I have a vague idea to refactor the parent controller update store logic to scatter all the complicated param into pieces and execute them in topological order. Let me do it for partial update first to clean this up incrementally...

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
sixpluszero: I can try see if we actually don't need this change...after I make the above change to address the comment..

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
sixpluszero: I refactored a bit and move it into util class and add java docs to document the logics. Please feel free to take a 2nd look.

gaojieliu (approved):

services/venice-controller/src/main/java/com/linkedin/venice/controller/util/ParentControllerConfigUpdateUtils.java
```diff
@@ -0,0 +1,171 @@
...
+      Optional<Boolean> partialUpdateRequest,
+      UpdateStore setStore,
+      boolean storeBeingConvertedToHybrid) {
+    Store currentStore = parentHelixAdmin.getVeniceHelixAdmin().getStore(clusterName, storeName);
+    VeniceControllerClusterConfig clusterConfig =
+        parentHelixAdmin.getVeniceHelixAdmin().getHelixVeniceClusterResources(clusterName).getConfig();
+    boolean partialUpdateConfigChanged = false;
+    setStore.writeComputationEnabled = currentStore.isWriteComputationEnabled();
+    if (partialUpdateRequest.isPresent()) {
+      setStore.writeComputationEnabled = partialUpdateRequest.get();
```
gaojieliu, line 53: So even the update param is same as the current setup, we want to pass it to child controller, right?
This is fine and just wondering the reason, and is it considering that child controller may be using a different config value for the same field in the same store?

sixpluszero: Honestly, I'm not sure what's the exact original reason to do this. This is actually the original config update logic that exists before my change in this PR, and we even have a check at the end to make sure every update store must have something to change, otherwise it is considered invalid and will throw.

I think, as you pointed out, it does give a chance to uniformly override a config value that might be inconsistent between parent and child region and between different child regions. We can consider evolving this logic in the future if we think it is not necessary, but to change this there will be some extra changes, I think it could be done in future PR though.

gaojieliu: SG.

gaojieliu (approved):

PR #643, opened 2023-09-21
[server] Apply default consumer receive buffer setting only if not provided via configs

sushantmane:
## How was this PR tested?

CI


nisargthakkar (approved):

sushantmane:
Thanks, @nisargthakkar for the review!

PR #644, opened 2023-09-21
[server][router][tc][fc] Minor Avro-related optimizations

FelixGV:
There were many cases of FastSerializerDeserializerFactory.getAvro[...] rather than getFastAvro[...]. This is confusing since both the slow and fast methods are available in the same factory class. Since it is not possible to override static methods, it is not straightforward to make the code less error-prone. For now, this commit does not attempt to refactor the class structure but merely fixes the individual call sites that called a slow avro factory method on the fast factory.

Separately, the RecordDeserializer interface was also tweaked to return a List rather than a Iterable in its two deserializeObjects functions. This makes it more efficient to get the size of the list without iterating. The reason for having an Iterable was from an earlier version of the code where we attempted to do lazy deserialization, but this approach turned out to not provide any performance benefit in benchmarks and was eventually discarded, so there is no point in hanging on to the more constrained interface at this time.

Test changes:

- Fixed use of wrong serializer in DispatchingAvroGenericStoreClientTest.

- Increased timeout on TestWritePathComputation#testFeatureFlagMultipleDC
  since it has difficulty completing in time on JDK8.

### DIFF
--- a/clients/venice-client/src/main/java/com/linkedin/venice/fastclient/DispatchingAvroGenericStoreClient.java
+++ b/clients/venice-client/src/main/java/com/linkedin/venice/fastclient/DispatchingAvroGenericStoreClient.java
@@ -644,11 +644,11 @@ public class DispatchingAvroGenericStoreClient<K, V> extends InternalAvroStoreCl
     metadata.start();
 
     this.multiGetSerializer =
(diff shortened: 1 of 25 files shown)


nisargthakkar (approved):
LGTM. Thanks

PR #646, opened 2023-09-21
[producer] Allow deleting records via the shell producer

nisargthakkar:
## Allow deleting records via the shell producer

The shell producer can allow deleting records by accepting the value `null` regardless of the fact that it is compatible with the store schema or not. Note that other forms of `null` like `"null"`, `'null'`, etc are not allowed.

## How was this PR tested?

Tested locally


majisourav99 (approved):
Thanks Nisarg

PR #648, opened 2023-09-21
[fc][test] Honor max batch get limit and return proper response for nonExistingKeys

m-nagarajan:
## Summary
1. streaming `batchGet` was not honoring max keys limit in a batch get. Returning `VeniceKeyCountLimitException` now.
2. setting the default max number of keys in a batch get in FC to 150. It was 2 before.
3. When there is a non-existing key in batch get, an exception was returned `Response was not complete`: Changed it to consider non-existing keys as part of response key count. 
4. Added unit tests/integration tests for the above cases.

## How was this PR tested?
GH CI

### DIFF
--- a/clients/venice-client/src/main/java/com/linkedin/venice/fastclient/ClientConfig.java
+++ b/clients/venice-client/src/main/java/com/linkedin/venice/fastclient/ClientConfig.java
@@ -11,7 +11,6 @@ import com.linkedin.venice.fastclient.meta.StoreMetadataFetchMode;
 import com.linkedin.venice.fastclient.stats.ClusterStats;
 import com.linkedin.venice.fastclient.stats.FastClientStats;
 import com.linkedin.venice.read.RequestType;
-import com.linkedin.venice.router.api.VenicePathParser;
 import com.linkedin.venice.systemstore.schemas.StoreMetaKey;
 import com.linkedin.venice.systemstore.schemas.StoreMetaValue;
 import com.linkedin.venice.utils.concurrent.VeniceConcurrentHashMap;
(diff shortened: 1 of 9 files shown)


clients/venice-client/src/main/java/com/linkedin/venice/fastclient/ClientConfig.java
```diff
@@ -51,13 +50,9 @@ public class ClientConfig<K, V, T extends SpecificRecord> {
...
 
-  /**
-   * The max allowed key count in batch-get request.
-   * Right now, the batch-get implementation will leverage single-get, which is inefficient, when there
-   * are many keys, since the requests to the same storage node won't be reused.
-   * But to temporarily unblock the first customer, we will only allow at most two keys in a batch-get request.
-   */
+  // Max allowed key count in batch-get request
   private final int maxAllowedKeyCntInBatchGetReq;
+  protected static final int MAX_ALLOWED_KEY_COUNT_IN_BATCHGET = 150;
```
majisourav99, line 55: Curious regular client and fast client should have same default ? probably FC can handle bigger batch ?

m-nagarajan: I also think it would be able to handle a bigger batch but making it to 150(default in thin client) from 2 is a step in the right direction and once we explore further, we can increase this number.

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/RetriableAvroGenericStoreClient.java
```diff
@@ -199,8 +199,8 @@ public void onCompletion(Optional<Exception> exception) {
         if (exception.isPresent()) {
           streamingResponseFuture.completeExceptionally(exception.get());
         } else {
-          streamingResponseFuture
-              .complete(new VeniceResponseMapImpl<>(valueMap, nonExistingKeys, valueMap.size() == keys.size()));
+          boolean isFullResponse = ((valueMap.size() + nonExistingKeys.size()) == keys.size());
```
majisourav99, line 202: nit: do we need the outer parenthesis?

m-nagarajan: fixed.

majisourav99 (approved):
LGTM. Thanks Manoj

m-nagarajan:
Thanks for the review @majisourav99

PR #652, opened 2023-09-21
[router] Check for compressor readiness in version finder

majisourav99:
## Check for compressor readiness in version finder

During startup of the router if there is no valid store versions, `DictionaryRetrievalService::startInner` will not block the router startup.  During initial reads, `VeniceVersionFinder` might return an online version which still does not have a compressor yet downloaded. That will lead to `Compressor not available for resource. Dictionary not downloaded.` exception. 
This PR will keep returning `NON_EXISTING_VERSION` till the compressor is ready and then add to the current version map. 

## How was this PR tested?

CI

### DIFF
--- a/services/venice-router/src/main/java/com/linkedin/venice/router/api/VeniceVersionFinder.java
+++ b/services/venice-router/src/main/java/com/linkedin/venice/router/api/VeniceVersionFinder.java
@@ -88,8 +88,7 @@ public class VeniceVersionFinder {
 
     int metadataCurrentVersion = store.getCurrentVersion();
     if (!lastCurrentVersionMap.containsKey(storeName)) {
-      if (metadataCurrentVersion != Store.NON_EXISTING_VERSION && store.getVersion(metadataCurrentVersion).isPresent()
-          && !isDecompressorReady(store, metadataCurrentVersion)) {
+      if (metadataCurrentVersion != Store.NON_EXISTING_VERSION && !isDecompressorReady(store, metadataCurrentVersion)) {
         return Store.NON_EXISTING_VERSION;
       } else {
         lastCurrentVersionMap.put(storeName, metadataCurrentVersion);
--- a/services/venice-router/src/main/java/com/linkedin/venice/router/api/VeniceVersionFinder.java
+++ b/services/venice-router/src/main/java/com/linkedin/venice/router/api/VeniceVersionFinder.java
@@ -88,9 +88,13 @@ public class VeniceVersionFinder {
 
     int metadataCurrentVersion = store.getCurrentVersion();
     if (!lastCurrentVersionMap.containsKey(storeName)) {
-      lastCurrentVersionMap.put(storeName, metadataCurrentVersion);
+      if (metadataCurrentVersion != Store.NON_EXISTING_VERSION && !isDecompressorReady(store, metadataCurrentVersion)) {
+        return Store.NON_EXISTING_VERSION;
+      } else {
+        lastCurrentVersionMap.put(storeName, metadataCurrentVersion);
+      }
       if (metadataCurrentVersion == Store.NON_EXISTING_VERSION) {
-        /** This should happen at most once per store, since we are adding the mapping to {@link lastCurrentVersionMap} */
+        // This should happen at most once per store, since we are adding the mapping to {@link lastCurrentVersionMap}
         store = metadataRepository.refreshOneStore(storeName);
         metadataCurrentVersion = store.getCurrentVersion();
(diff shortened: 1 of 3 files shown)


services/venice-router/src/main/java/com/linkedin/venice/router/api/VeniceVersionFinder.java
```diff
@@ -88,9 +88,13 @@ public int getVersion(String storeName, BasicFullHttpRequest request) throws Ven
 
     int metadataCurrentVersion = store.getCurrentVersion();
     if (!lastCurrentVersionMap.containsKey(storeName)) {
-      lastCurrentVersionMap.put(storeName, metadataCurrentVersion);
+      if (metadataCurrentVersion != Store.NON_EXISTING_VERSION && !isDecompressorReady(store, metadataCurrentVersion)) {
+        return Store.NON_EXISTING_VERSION;
```
m-nagarajan, line 92: There is some logic handling decompressor not ready case in `maybeServeNewCurrentVersion` which will be dead code moving forward. can we clean that up as well? Thanks.

services/venice-router/src/test/java/com/linkedin/venice/router/api/TestVeniceVersionFinder.java
```diff
@@ -275,7 +277,7 @@ public void returnsCurrentVersionWhenTheDictionaryExists() {
     }
   }
 
-  @Test
+  @Test(enabled = false)
```
adamxchen, line 280: why do we disable it? If we don't need, we can remove it. Is there an issue that keeps failing this test?

services/venice-router/src/test/java/com/linkedin/venice/router/api/TestVeniceVersionFinder.java
adamxchen, line 1: Can we also have a new unit test for the new case you added?

majisourav99: Sure

services/venice-router/src/test/java/com/linkedin/venice/router/api/TestVeniceVersionFinder.java
majisourav99: enabled it

services/venice-router/src/test/java/com/linkedin/venice/router/api/TestVeniceVersionFinder.java
```diff
@@ -318,8 +320,11 @@ public void returnsCurrentVersionWhenItIsTheOnlyOption() {
 
     String firstVersionKafkaTopic = Version.composeKafkaTopic(storeName, firstVersion);
 
-    Assert.assertEquals(versionFinder.getVersion(storeName, request), firstVersion);
+    Assert.assertEquals(versionFinder.getVersion(storeName, request), Store.NON_EXISTING_VERSION);
     Assert.assertNull(compressorFactory.getVersionSpecificCompressor(firstVersionKafkaTopic));
+
+    doReturn(true).when(compressorFactory).versionSpecificCompressorExists(firstVersionKafkaTopic);
+    Assert.assertEquals(versionFinder.getVersion(storeName, request), firstVersion);
```
nisargthakkar, line 327: Now that this test behavior has changed, can you also update the comment (and test name if needed) ?
```
    // When the router doesn't know of any other versions, it will return that version even if dictionary is not
    // downloaded.
    // If the dictionary is not downloaded by the time the records needs to be decompressed, then the router will return
    // an error response.
```

services/venice-router/src/main/java/com/linkedin/venice/router/api/VeniceVersionFinder.java
```diff
@@ -88,9 +88,13 @@ public int getVersion(String storeName, BasicFullHttpRequest request) throws Ven
 
     int metadataCurrentVersion = store.getCurrentVersion();
     if (!lastCurrentVersionMap.containsKey(storeName)) {
```
lluwm, line 90: Not in this change, but it looks like there could have a race condition from L90 - L102, i.e., if multiple threads discover that `lastCurrentVersionMap` doesn't contain the key at same time and, all of them go ahead to refresh and update the K-V at the same time. This is because the test and set operations here is not atomic, however, I don't know if this could be a benign one or not, but just a heads up.

In case if we want to fix the race and get a peaceful mind, we can probably wrap them into a `computeIfAbsent` block.

PR #662, opened 2023-09-22
[server] Refresh the PerStoreVersionState mapping for each request in StorageReadRequestHandler

gaojieliu:
There is a regression in a recent code refactoring, the following mapping in StorageReadRequestHandler can stale during rebalance: Map<String, PerStoreVersionState> perStoreVersionStateMap

In the following rebalance scenario, the storage engine reference in the above map will be stale:
1. There is only one assigned partition for this resource, and let us say the storage engine instance is 'se1'.
2. When Helix rebalance decides to move this partition to another node, there is no partition left, so this storage engine: 'se1' will be closed and removed from the storage engine factory.
3. Later, Helix decides to move back another partition to this node, Storage engine factory will create a new instance of storage engine: 'se2'.
4. This particular node will bootstrap the rebalanced partition in 'se2'.
5. When the bootstrapping is done, Router will start sending request to the rebalanced partition, but 'StorageReadRequestHandler' is referring to a stale storage engine instance: 'se1', which is empty.
6. The read request to this rebalanced partition will be rejected even this particular node has the partition on disk and Helix EV/CV.

The fix is to refresh the storage engine reference for every request.

## How was this PR tested?

CI

### DIFF
--- a/services/venice-server/src/main/java/com/linkedin/venice/listener/StorageReadRequestHandler.java
+++ b/services/venice-server/src/main/java/com/linkedin/venice/listener/StorageReadRequestHandler.java
@@ -124,7 +124,7 @@ public class StorageReadRequestHandler extends ChannelInboundHandlerAdapter {
   private static class PerStoreVersionState {
     final PartitionerConfig partitionerConfig;
     final VenicePartitioner partitioner;
-    final AbstractStorageEngine storageEngine;
+    AbstractStorageEngine storageEngine;
     final StoreDeserializerCache<GenericRecord> storeDeserializerCache;
 
     public PerStoreVersionState(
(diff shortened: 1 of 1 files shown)


sixpluszero (approved):

adamxchen:
Override the check to merge this hot fix.

services/venice-server/src/main/java/com/linkedin/venice/listener/StorageReadRequestHandler.java
```diff
@@ -383,7 +383,18 @@ private int getSubPartitionId(
...
-    return perStoreVersionStateMap.computeIfAbsent(storeVersion, this::generatePerStoreVersionState);
+    return perStoreVersionStateMap.compute(storeVersion, (ignored, versionState) -> {
+      if (versionState == null) {
+        return this.generatePerStoreVersionState(storeVersion);
+      } else {
+        /**
+         * Try to refresh the storage engine reference for every request in case {@link com.linkedin.davinci.store.StorageEngineFactory}
+         * recreates a new instance during re-balance, which may drop the storage engine and recreate a new one.
+         */
+        versionState.storageEngine = storageEngineRepository.getLocalStorageEngine(storeVersion);
```
FelixGV, line 394: If we're going to look this up every time on the hot path, then I don't think it makes sense to have it in the cached mapping... let's just remove it entirely from the `PerStoreVersionState` object and look it up separately.

Or alternatively, we don't look it up in the hot path, but the Helix state transition handling code can call a hook to update and/or clear this mapping when dealing with the first/last partition... that is likely more complicated though.

FelixGV (commented):
Thanks for catching this regression... left a comment about how to solve it.

services/venice-server/src/main/java/com/linkedin/venice/listener/StorageReadRequestHandler.java
gaojieliu: There are certainly ways to make it more efficient.
Yesterday, there was GCN pressure, so we made a simple solution.

In theory, we can check whether the storage engine is empty or not and based on the current logic, we only need to refresh the reference when the storage engine instance kept in `PerStoreVersionState` is empty.

FelixGV: Yeah, I understand the need for a tactical fix. Thanks a lot for putting it together, and my apologies for introducing the regression.

My only concern with the fix in this PR is that the closure passed into the `CHM`'s `compute` function is guaranteed to be executed atomically, which seems to me like it could be a point of high contention, especially in dedicated clusters containing one or few stores that all have a high throughput...

I suggest an alternative approach in this PR, which I believe may be more efficient: https://github.com/linkedin/venice/pull/663

Thanks again for the late night hot fix.

PR #663, opened 2023-09-22
[server] Fixed stale storage engine reference in server

FelixGV:
The issue could happen in a scenario where the server lost its only partition of a store, the storage engine got closed, and then regained a partition for the store. See the JavaDoc of the new unit test for more details.

## How was this PR tested?
New unit test.

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/AbstractStorageEngine.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/AbstractStorageEngine.java
@@ -79,6 +79,8 @@ public abstract class AbstractStorageEngine<Partition extends AbstractStoragePar
    */
   private final List<ReadWriteLock> rwLockForStoragePartitionAdjustmentList = new SparseConcurrentList<>();
 
+  private volatile boolean closed = false;
+
   public AbstractStorageEngine(
       String storeName,
       InternalAvroSpecificSerializer<StoreVersionState> storeVersionStateSerializer,
@@ -340,6 +342,7 @@ public abstract class AbstractStorageEngine<Partition extends AbstractStoragePar
 
   @Override
   public synchronized void close() throws VeniceException {
+    this.closed = true;
     long startTime = System.currentTimeMillis();
     List<Partition> tmpList = new ArrayList<>();
     // SparseConcurrentList does not support parallelStream, copy to a tmp list.
@@ -352,6 +355,10 @@ public abstract class AbstractStorageEngine<Partition extends AbstractStoragePar
         LatencyUtils.getElapsedTimeInMs(startTime));
     partitionList.clear();
     closeMetadataPartition();
+  }
+
+  public boolean isClosed() {
+    return this.closed;
   }
 
   /**
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/AbstractStorageEngine.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/AbstractStorageEngine.java
@@ -20,7 +20,6 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
-import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.Callable;
@@ -60,7 +59,7 @@ public abstract class AbstractStorageEngine<Partition extends AbstractStoragePar
   public static final int METADATA_PARTITION_ID = 1000_000_000;
 
   private final String storeName;
-  private final List<Partition> partitionList = new SparseConcurrentList<>();
+  private final SparseConcurrentList<Partition> partitionList = new SparseConcurrentList<>();
   private Partition metadataPartition;
   private final AtomicReference<StoreVersionState> versionStateCache = new AtomicReference<>();
   private final InternalAvroSpecificSerializer<StoreVersionState> storeVersionStateSerializer;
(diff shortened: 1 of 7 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/AbstractStorageEngine.java
```diff
@@ -340,6 +342,7 @@ public synchronized Map<String, String> sync(int partitionId) {
 
   @Override
   public synchronized void close() throws VeniceException {
+    this.closed = true;
```
FelixGV, line 345: This may not be enough... for this strategy to work, the flag probably needs to be made true in `drop()` as well. I wonder though... is it possible for a storage engine to be dropped, and then to re-add more partitions into it again? If that is the case, then maybe there needs to be a code path that sets `closed` back to false also 🤔 ...

gaojieliu: Maybe we should check whether `partitionList` is empty or not instead of introducing a new flag?

FelixGV: Good idea... I did that in the latest commit. LMK how it looks.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/AbstractStorageEngine.java
```diff
@@ -354,6 +354,10 @@ public synchronized void close() throws VeniceException {
     closeMetadataPartition();
   }
 
+  public boolean isClosed() {
+    return this.partitionList.isEmpty();
```
majisourav99, line 358: Can you verify for sparse list `isEmpty` will actually return the correct value you want.

FelixGV: Good catch... it will not do what I want... that is kind of a problem with unit tests and mocking... the mocked behavior is not guaranteed to match with reality. I will come up with another way...

FelixGV: I added a `nonNullSize` counter in the `SparseConcurrentList` to deal with this...

gaojieliu: @FelixGV 
Can you write some unit tests against this utility class?

internal/venice-client-common/src/main/java/com/linkedin/venice/utils/SparseConcurrentList.java
```diff
@@ -110,55 +118,117 @@ public void forEach(Consumer<? super E> itemConsumer) {
...
+  @Override
   public synchronized void clear() {
+    this.nonNullSize = 0;
     super.clear();
   }
 
+  @Override
   public synchronized boolean addAll(Collection<? extends E> c) {
-    return super.addAll(c);
+    boolean modified = super.addAll(c);
```
majisourav99, line 201: Are we assuming addAll will always add non-null values?

FelixGV: No we're not assuming that... that's why I'm doing this in the line below:

```
this.nonNullSize = values().size();
```

That will iterate through the entire list to figure out the number of non-null entries.

BTW, the various operations listed under `// Boilerplate code just to add synchronization to mutation operations:` are not used today... they're just added for completeness sake.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/AbstractStorageEngine.java
FelixGV: Yep, I did that in the latest commit.

internal/venice-client-common/src/main/java/com/linkedin/venice/utils/SparseConcurrentList.java
majisourav99: I see, `size()` does iterate on non-null elements.  Good to have complete implementation of this util class, in case in future if someone else plans to use all the methods.

FelixGV:
With the latest commits, `logDiffCoverage` is 90% and the module's branch coverage as a whole is 45.54% (up from 44.81%). Seems like the Pulsar/Docker tests are broken though... does not seem related to this PR.

internal/venice-client-common/src/main/java/com/linkedin/venice/utils/SparseConcurrentList.java
```diff
@@ -110,55 +118,117 @@ public void forEach(Consumer<? super E> itemConsumer) {
...
+  @Override
   public synchronized boolean addAll(int index, Collection<? extends E> c) {
-    return super.addAll(index, c);
+    boolean modified = super.addAll(index, c);
+    this.nonNullSize = values().size();
+    return modified;
   }
 
+  @Override
   public synchronized boolean removeIf(Predicate<? super E> filter) {
```
gaojieliu, line 214: I saw this class doesn't override:
```
public boolean remove(Object o)
``
So if someone uses this method, the internal `nonNullSize` will be wrong, right?
Ideally, this class should override all the update functions.

FelixGV: That is correct, it should override all mutation methods... I'll take another look...

gaojieliu (approved):
Thanks!

majisourav99 (commented):
Thanks for the change, if you are up for it can you also update the body for `getNumberOfPartitions` to check for non-nulls size instead, should give bit of perf as well.

majisourav99 (approved):
LGTM! Thank you Felix.

PR #665, opened 2023-09-25
[server] Add metric for poll count only if it returns non-zero record count

majisourav99:
## Add metric for poll count only if it returns non-zero record count

If some consumer poll returns empty record count, the metric will record 0 which will affect the avg count for consumers which are actually polling non-zero records. For a clearer view of the poll result, this PR will add another metric which will record only when there is a non-zero results from the poll

## How was this PR tested?

CI

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ConsumptionTask.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ConsumptionTask.java
@@ -119,11 +119,11 @@ class ConsumptionTask implements Runnable {
         polledPubSubMessages = pollFunction.get();
         lastSuccessfulPollTimestamp = System.currentTimeMillis();
         stats.recordPollRequestLatency(lastSuccessfulPollTimestamp - beforePollingTimeStamp);
+        stats.recordPollResultNum(polledPubSubMessagesCount);
         payloadBytesConsumedInOnePoll = 0;
         polledPubSubMessagesCount = 0;
         if (!polledPubSubMessages.isEmpty()) {
           beforeProducingToWriteBufferTimestamp = System.currentTimeMillis();
-          stats.recordPollResultNum(polledPubSubMessagesCount);
           for (Map.Entry<PubSubTopicPartition, List<PubSubMessage<KafkaKey, KafkaMessageEnvelope, Long>>> entry: polledPubSubMessages
               .entrySet()) {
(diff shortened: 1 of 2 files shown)


adamxchen:
What's the use case of this non-zero metric? I think we have lots of metrics already so I'd hope we don't add more metrics if we have alternatives.

majisourav99:
> What's the use case of this non-zero metric? I think we have lots of metrics already so I'd hope we don't add more metrics if we have alternatives.

Say we have 10 consumer with only 1 is polling non-empty record list and other are getting empty records, that will skew the avg heavily and would appear none of consumers polling any records.

adamxchen:
> > What's the use case of this non-zero metric? I think we have lots of metrics already so I'd hope we don't add more metrics if we have alternatives.
> 
> Say we have 10 consumer with only 1 is polling non-empty record list and other are getting empty records, that will skew the avg heavily and would appear none of consumers polling any records.

I feel that's the expected behavior of metrics right? It's normal to have metrics emitting zero so the avg would drop as expected. 
I guess my question is why we need to see non-zero metric for this situation. For this particular case, if we want to see the behavior of each consumer, we can add per-consumer metrics (I think that'd be too much metrics tho). If we want to know the total count, we can do some math to figure it out

huangminchn:
> > > What's the use case of this non-zero metric? I think we have lots of metrics already so I'd hope we don't add more metrics if we have alternatives.
> > 
> > 
> > Say we have 10 consumer with only 1 is polling non-empty record list and other are getting empty records, that will skew the avg heavily and would appear none of consumers polling any records.
> 
> I feel that's the expected behavior of metrics right? It's normal to have metrics emitting zero so the avg would drop as expected. I guess my question is why we need to see non-zero metric for this situation. For this particular case, if we want to see the behavior of each consumer, we can add per-consumer metrics (I think that'd be too much metrics tho). If we want to know the total count, we can do some math to figure it out

Later I noticed that there are two cases:
1. there are still some remaining records to consume, but poll returns 0 record
2. there are no remaining record to consume, since consumer is still subscribe to the partition, poll returns 0 record, and pollute the avg poll record num metric.

We wanted to fix (2), but later, after testing in EI, we found that (1) could happen. For (1) case, we should report to the avg metric.

And in this diff, adding new metric is fine, since we didn't touch the original one. When we know only one few consumers have remaining data to consume, we can refer to the new metric.

huangminchn (approved):
Thanks Sourav! Looks good to me.

PR #666, opened 2023-09-25
[controller] Create meta store RT topic before writing meta store up…

lluwm:
…dates

In function `setUpMetaStoreAndMayProduceSnapshot`, one of the steps is to update a store flag and the store update operation needs to write to the meta store's rt topic. However, the RT topic is created later in the same function, thus causing errors in writing the update.

To fix the issue, this change move the RT topic creation before writing meta store updates.

## How was this PR tested?
CI

### DIFF
--- a/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
@@ -7416,6 +7416,10 @@ public class VeniceHelixAdmin implements Admin, StoreCleaner {
       throwStoreDoesNotExist(clusterName, regularStoreName);
     }
 
+    // Make sure RT topic exists before producing. There's no write to parent region meta store RT, but we still create
+    // the RT topic to be consistent in case it was not auto-materialized
+    getRealTimeTopic(clusterName, VeniceSystemStoreType.META_STORE.getSystemStoreName(regularStoreName));
+
     // Update the store flag to enable meta system store.
     if (!store.isStoreMetaSystemStoreEnabled()) {
       storeMetadataUpdate(clusterName, regularStoreName, (s) -> {
@@ -7423,10 +7427,6 @@ public class VeniceHelixAdmin implements Admin, StoreCleaner {
         return s;
       });
     }
-
-    // Make sure RT topic exists before producing. There's no write to parent region meta store RT, but we still create
-    // the RT topic to be consistent in case it was not auto-materialized
-    getRealTimeTopic(clusterName, VeniceSystemStoreType.META_STORE.getSystemStoreName(regularStoreName));
 
     Optional<MetaStoreWriter> metaStoreWriter = getHelixVeniceClusterResources(clusterName).getMetaStoreWriter();
     if (!metaStoreWriter.isPresent()) {
--- a/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
@@ -7408,7 +7408,7 @@ public class VeniceHelixAdmin implements Admin, StoreCleaner {
     }
   }
 
-  private void setUpMetaStoreAndMayProduceSnapshot(String clusterName, String regularStoreName) {
+  void setUpMetaStoreAndMayProduceSnapshot(String clusterName, String regularStoreName) {
     checkControllerLeadershipFor(clusterName);
     ReadWriteStoreRepository repository = getHelixVeniceClusterResources(clusterName).getStoreMetadataRepository();
     Store store = repository.getStore(regularStoreName);
--- a/services/venice-controller/src/test/java/com/linkedin/venice/controller/TestVeniceHelixAdmin.java
+++ b/services/venice-controller/src/test/java/com/linkedin/venice/controller/TestVeniceHelixAdmin.java
@@ -5,12 +5,15 @@ import static org.mockito.Mockito.*;
 
 import com.linkedin.venice.helix.HelixExternalViewRepository;
 import com.linkedin.venice.meta.PartitionAssignment;
+import com.linkedin.venice.meta.ReadWriteStoreRepository;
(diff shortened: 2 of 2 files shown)


services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -7416,6 +7416,10 @@ private void setUpMetaStoreAndMayProduceSnapshot(String clusterName, String regu
       throwStoreDoesNotExist(clusterName, regularStoreName);
     }
 
+    // Make sure RT topic exists before producing. There's no write to parent region meta store RT, but we still create
+    // the RT topic to be consistent in case it was not auto-materialized
+    getRealTimeTopic(clusterName, VeniceSystemStoreType.META_STORE.getSystemStoreName(regularStoreName));
+
```
sushantmane, line 7422: Can you add a check here to make sure that RT exists with expected partition count? I thinks API is containsTopicAndPartitionWithRetries.

lluwm: Okay, do you think `getRealTimeTopic` already check that internally? 
getRealTimeTopic -> createTopic -> waitUntilTopicCreated -> containsTopicAndAllPartitionsAreOnline

sushantmane: Oh yeah that check is already there. Thanks for checking it.

sushantmane (approved):
Thanks, @lluwm!

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
adamxchen: @lluwm  Do you think you can have a unit test here to enforce the order? Just want to ensure future refactoring would not break this.

If that's very hard to write a unit test for this, I think that's okay to drop this comment.

sushantmane: @majisourav has used clever way to write UR for VHA class: https://github.com/linkedin/venice/blob/main/services/venice-controller/src/test/java/com/linkedin/venice/controller/TestVeniceHelixAdmin.java

lluwm: To be honest, I don't have a good idea in writing a unit test for this case and I am not very sure if this particular scenario is a good fit for unit test. Too many components other than controller are involved, e.g. ZK, Kafka and then are involved very deeply. It is not easy to mock up all of them, isn't it? Let me know if you have some good ideas.

adamxchen: For this particular case, it's not hard actually. When it comes to unit tests, we don't have mock every dependency if that's test-friendly codes.

I've tried it myself and it looks like this
```
  @Test
  public void testRealTimeTopicCreationOrder() {
    VeniceHelixAdmin veniceHelixAdmin = mock(VeniceHelixAdmin.class);
    doReturn("test_rt").when(veniceHelixAdmin).getRealTimeTopic(anyString(), anyString());
    # note 1: You need to make this method protected instead of private
    doCallRealMethod().when(veniceHelixAdmin).setUpMetaStoreAndMayProduceSnapshot(anyString(), anyString());

    InOrder inorder = inOrder(veniceHelixAdmin);

    HelixVeniceClusterResources veniceClusterResources = mock(HelixVeniceClusterResources.class);
    ReadWriteStoreRepository repo = mock(ReadWriteStoreRepository.class);
    Store store = mock(Store.class);

    doReturn(veniceClusterResources).when(veniceHelixAdmin).getHelixVeniceClusterResources(anyString());
    doReturn(repo).when(veniceClusterResources).getStoreMetadataRepository();
    doReturn(store).when(repo).getStore(anyString());
    doReturn(Boolean.FALSE).when(store).isDaVinciPushStatusStoreEnabled();

    veniceHelixAdmin.setUpMetaStoreAndMayProduceSnapshot(anyString(), anyString());
     
    # note 2: in the main branch, this is order but you probably need to switch them in your branch
    inorder.verify(veniceHelixAdmin).storeMetadataUpdate(anyString(), anyString(), any());
    inorder.verify(veniceHelixAdmin).getRealTimeTopic(anyString(), anyString());
  }
```

lluwm: Thanks @adamxchen for the example.

The reason I think it is hard is because, in my mind, the real order that we want to establish/verify for this function is not only the statement order, but probably something more close to the root cause of the issue, i.e. the invariant that the meta store RT topic has to be created before the Venice writer writing to it. Any changes could break such invariant in this function should be errored out. I think that part is probably hard with unit test and an integration test might be more proper. Statement level order enforcement might be good for now but might not work well when functions change their meaning during refactoring and we probably want to revisit test when the function meaning changes. If you prefer to add the unit test to enforce the statement level order, I think I am okay with that as well.

adamxchen: > invariant that the meta store RT topic has to be created before the Venice writer writing to it.

Yes, I agree that the true behavior we want to test is the topic must have been created before writing to it. `getRealTimeTopic` is a blocking call so if there's no issue, the topic should have been created after moving to the next line. Integ test is heavy for just verifying the topic creation so that's why I think unit test is more appropriate but I also agree enforcing statement ordering is probably not the best way to verify this either in the long term.

I'm fine either way so It's up to you and feel free to dismiss the comment.

lluwm: Sounds good, I think I will add this unit test and improve a little bit on the comments part to make it clear what we want to verify fundamentally by this test and also clarify that in case later if functions change their semantics, it is okay to relax on the ordering enforcement (or delete the unit test if necessary).

adamxchen (approved):
Thank you!

PR #668, opened 2023-09-27
[server] CV state cleanup during standby -> offline and pre-start

gaojieliu:
In the past, CV state will remain 'Completed' when the node is going through a rebalance/restart/ungraceful shutdown, which would lead Venice Router continue to send traffic to unavailable replicas, and this hasn't caused any unhealthy requests yet because of the error retry mechanism in Router.

To fix this issue cleanly, this code change introduces the following strategies:
1. Remove CV states when replica is transitioning from standby to offline.
2. Remove CV states for all the hosted partitions when stopping Venice Server instances.
3. Remove CV states for all the hosted partitions before starting the Venice Server instances to handle ungraceful shutdown.
This code change also adds the handling of missing notification types in main process when II is enabled.

## How was this PR tested?
CI


clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/AbstractPartitionStateModel.java
```diff
@@ -144,7 +144,7 @@ public void rollbackOnError(Message message, NotificationContext context, StateT
        * When state transition fails, we shouldn't remove the corresponding database here since the database could
        * be either recovered by bounce/Helix Reset or completely dropped after going through 'ERROR' to 'DROPPED' state transition.
        */
-      stopConsumption();
+      stopConsumption(false);
```
FelixGV, line 147: What happens to the CV when the replica is in `ERROR` state? Does CV represent `ERROR`, or does it continue carrying the previous state? If it's the latter, then should we also clear CV states here...? Either way, it seems a little tricky, and we might want to leave a comment to that effect?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/AbstractPartitionStateModel.java
```diff
@@ -168,7 +168,7 @@ public void onBecomeDroppedFromError(Message message, NotificationContext contex
   @Override
   public void reset() {
     try {
-      stopConsumption();
+      stopConsumption(false);
```
FelixGV, line 171: What is the purpose of `reset`, again? I forgot when does Helix call this, so I've difficulty assessing whether updating or not updating the CV is ideal in this case...

clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/AbstractPartitionStateModel.java
```diff
@@ -352,8 +352,22 @@ private void initializePartitionPushStatus() {
     }
   }
 
-  protected void stopConsumption() {
-    ingestionBackend.stopConsumption(storeAndServerConfigs, partition);
+  protected void stopConsumption(boolean dropCVState) {
+    CompletableFuture<Void> future = ingestionBackend.stopConsumption(storeAndServerConfigs, partition);
+    if (dropCVState) {
+      future.whenComplete((ignored, throwable) -> {
```
FelixGV, line 358: Why wait until after the consumption is stopped? Why not remove the CV first, then initiate the stop of consumption? Would we not want to notify routers about the CV drop ASAP? In the end, it'll be a race condition anyway whether the router finds out about it before the consumption is stopped or not, but it seems more natural to me to begin notifying routers first since we don't intend them to keep sending traffic here...

clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/HelixParticipationService.java
```diff
@@ -360,6 +369,27 @@ private void asyncStart() {
...
+      StorageService storageService,
+      Logger currentLogger) {
+    // Get all hosted stores
+    currentLogger.info("Started resetting all instance CV states");
+    Map<String, Set<Integer>> storePartitionMapping = storageService.getStoreAndUserPartitionsMapping();
+    storePartitionMapping.forEach((storeName, partitionIds) -> {
+      partitionIds.forEach(partitionId -> {
+        try {
+          accessor.deleteReplicaStatus(storeName, partitionId);
+          currentLogger.info("Deleting CV state for resource: {} and partition id: {}", storeName, partitionId);
```
FelixGV, line 383: Nitpick: Can we do this once for all partitions of the store (could log the `partitionIds ` set in a single line if you want to know the exact partition numbers) rather than one log line per partition? Even at one log per store it could still be quite noisy on the multi-tenant clusters...

FelixGV (commented):
Thanks Gaojie! I left a few minor comments...

clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/AbstractPartitionStateModel.java
gaojieliu: So once a replica reports `COMPLETED` in CV, it won't go back to `ERROR` state.
And if the replica becomes `ERROR` during bootstrapping, `ERROR` state will be reported in CV as well.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/AbstractPartitionStateModel.java
gaojieliu: The reason to have such logic is that `stopConsumption` is async today.
If we update CV state first here, the CV state can be changed again when executing `stopConsumption` logic in `SIT`.
That was the main reason to have such callback.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/HelixParticipationService.java
gaojieliu: Sure, I can do that.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/AbstractPartitionStateModel.java
FelixGV: Why will the CV state be changed again? Is it because of this:

```
statusReportAdapter.reportStopped(consumptionState);
```

In `SIT::processCommonConsumerAction`?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/AbstractPartitionStateModel.java
FelixGV: Ok, so if the CV contains `ERROR`, that is fine, no need to remove it from the CV then.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/AbstractPartitionStateModel.java
gaojieliu: Well, when trying to stop consumption, `SIT` will try to drain the queued messages, and when drainer processing these messages, it can change the replica state, such as changing it to be `completed` or `errored` , which means CV state will be updated after removal.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/AbstractPartitionStateModel.java
gaojieliu: I see.

The PR will explicitly remove all the CV states during graceful shutdown.
I was hesitate to delete CV state in `reset` function since `reset` function can be invoked in different scenarios.

lluwm: I agree. It could have other scenarios calling reset in the HelixStateMachineEngine code and it would be safer to remove CVs only when we are sure to shutdown the server.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/AbstractPartitionStateModel.java
FelixGV: Hmm... this is a bigger scope of change and it is fine to ignore it in this PR, but I wonder... what is the point of this draining? Why not just stop instantly, and later on skip any message belonging to an unsubscribed partition, rather than do this draining?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/HelixParticipationService.java
```diff
@@ -360,6 +369,28 @@ private void asyncStart() {
...
 
+  static void resetAllInstanceCVStates(
+      HelixPartitionStatusAccessor accessor,
+      StorageService storageService,
+      Logger currentLogger) {
+    // Get all hosted stores
+    currentLogger.info("Started resetting all instance CV states");
+    Map<String, Set<Integer>> storePartitionMapping = storageService.getStoreAndUserPartitionsMapping();
+    storePartitionMapping.forEach((storeName, partitionIds) -> {
+      currentLogger.info("Started deleting replica status for resource: {}, partitions: {}", storeName, partitionIds);
```
FelixGV, line 380: I think we can get rid of this... no need for logs before and after. We are a bit sensitive to excessive logging these days so it's better not to pile on too much more, IMHO. In a cluster with 500 stores this line will come up a lot during startup and shutdown...

clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/AbstractPartitionStateModel.java
FelixGV: Shall we add this conclusion to the comment or do you prefer not to?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/HelixParticipationService.java
gaojieliu: I added this logging intentionally to monitor the performance of CV state cleanup since it will cleanup a CV state per partition.

If you think per-store logging is too much, I can only keep the log for the beginning and end.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/AbstractPartitionStateModel.java
gaojieliu: Sure, will do.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/AbstractPartitionStateModel.java
gaojieliu: Yeah, we had such conversion about making ST sync to make logic more straightforward.
Today, handling these consumer actions are completely async.
It will be a much bigger change and I would like to keep that as a separate PR.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/HelixParticipationService.java
FelixGV: If you want the per-store timing you could take the time before and after and include that in the log?

FelixGV (approved):
Thanks!

PR #669, opened 2023-09-27
[server][test] Message header triggered new KME schema registration i…

lluwm:
…n venice server

This is a follow-up change to 0b5af23 and it is also part of the solution to remove deployment order for KafkaMessageEnvelope (KME) value schema. For Venice servers, by design, we apply the same strategy as controllers, i.e., when servers find an unknown KME schema from the message header, they need to request the system cluster leader controller to register the new schema into local region system store.

This change enables Venice servers to perform the unknown KME schemas registration work when it discovers. At its core, it leverages the existing functionalities from the ControllerClientBackedSystemSchemaInitializer class for the implementation. A new server config is introduced and a new integration test is added to verify its correctness.

## How was this PR tested?

CI passed.
Add a new integration test.

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaStoreIngestionService.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaStoreIngestionService.java
@@ -18,6 +18,7 @@ import static com.linkedin.venice.kafka.TopicManager.DEFAULT_KAFKA_OPERATION_TIM
 import static java.lang.Thread.currentThread;
 import static java.lang.Thread.sleep;
 
+import com.google.common.collect.ImmutableMap;
 import com.linkedin.davinci.compression.StorageEngineBackedCompressorFactory;
 import com.linkedin.davinci.config.VeniceConfigLoader;
 import com.linkedin.davinci.config.VeniceServerConfig;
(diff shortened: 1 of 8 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaStoreIngestionService.java
```diff
@@ -399,8 +410,42 @@ public void handleStoreDeleted(Store store) {
...
-    // TODO: Wire configs into these params
-    KafkaValueSerializer kafkaValueSerializer = new OptimizedKafkaValueSerializer();
+    /**
+     * Register a callback function to handle the case when a new KME value schema is encountered when the server
+     * consumes messages from Kafka.
+     */
+    BiConsumer<Integer, Schema> newSchemaEncountered = (schemaId, schema) -> {
+      LOGGER.info("Encountered a new KME value schema (id = {}), proceed to register", schemaId);
+      try {
+        ControllerClientBackedSystemSchemaInitializer schemaInitializer =
```
gaojieliu, line 420: So. this logic will be triggered when encountering an unknown schema, and I see some minor issue:
 The resource is not reclaimed after each usage. Since this logic will create a new instance of `ControllerClientBackedSystemSchemaInitializer`, and it never be reused, so shall we close it after the call?

lluwm: Thanks for the review Gaojie.

That's correct. 
> So. this logic will be triggered when encountering an unknown schema.

Let me understand your concern better, `schemaInitializer` is a local variable and its memory would be good for GC, once it is out of the scope, right?

> The resource is not reclaimed after each usage. Since this logic will create a new instance of ControllerClientBackedSystemSchemaInitializer, and it never be reused, so shall we close it after the call?

gaojieliu: The `schemaInitializer` instance will allocate a ControllerClient to register system schema, so if we don't want to reuse this instance, shall we close the Controller client created internally?

lluwm: Oh, I see, the internal `controllerClient`. right, we should close it.

gaojieliu (approved):
Thanks!

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaStoreIngestionService.java
```diff
@@ -18,6 +18,7 @@
 import static java.lang.Thread.currentThread;
 import static java.lang.Thread.sleep;
 
+import com.google.common.collect.ImmutableMap;
```
FelixGV, line 21: I would prefer that we don't pull in guava classes. They tend to be unstable across versions and cause compatibility headaches.

lluwm: Got it. It has been merged but will fix it in my next change.

PR #670, opened 2023-09-28
[compat][controller][admintool] Added store-level max compaction lag support

gaojieliu:
`max.compaction.lag.ms` is used to control the maximum time a message will remain ineligible for compaction. This is the config, which will trigger frequent compaction when there are a lot of duplicated messages in version topic. This config has been proved to help reduce the DaVinci bootstrapping time when the version topic contains many duplicate messages with high write throughput. So far, we will manually adjust this setting for the high-write throughput DaVinci stores.

## How was this PR tested?
CI

### DIFF
--- a/internal/venice-common/src/main/resources/avro/StoreMetaValue/v16/StoreMetaValue.avsc
+++ b/internal/venice-common/src/main/resources/avro/StoreMetaValue/v16/StoreMetaValue.avsc
@@ -157,6 +157,7 @@
             {"name": "activeActiveReplicationEnabled", "type": "boolean", "default": false, "doc": "Whether or not active/active replication is enabled for hybrid stores; eventually this config will replace native replication flag, when all stores are on A/A"},
             {"name": "applyTargetVersionFilterForIncPush", "type": "boolean", "default": false, "doc": "Whether or not the target version field in Kafka messages will be used in increment push to RT policy"},
             {"name": "minCompactionLagSeconds", "type": "long", "default": -1, "doc": "Store level min compaction lag config and if not specified, it will use the global config for version topics"},
+            {"name": "maxCompactionLagSeconds", "type": "long", "default": -1, "doc": "Store level max compaction lag config and if not specified, 'max.compaction.lag.ms' config won't be setup in the corresponding version topics"},
             {
               "name": "versions",
               "doc": "List of non-retired versions. It's currently sorted and there is code run under the assumption that the last element in the list is the largest. Check out {VeniceHelixAdmin#getIncrementalPushVersion}, and please make it in mind if you want to change this logic",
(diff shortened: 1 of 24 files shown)


services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/consumer/AdminExecutionTask.java
```diff
@@ -519,6 +519,8 @@ private void handleSetStore(UpdateStore message) {
 
     params.setStorageNodeReadQuotaEnabled(message.storageNodeReadQuotaEnabled);
     params.setMinCompactionLagSeconds(message.minCompactionLagSeconds);
+    params.setMaxCompactionLagSeconds(message.maxCompactionLagSeconds);
+    LOGGER.info("max compaction lag ms in AdminExecutionTask: " + message.maxCompactionLagSeconds);
```
sixpluszero, line 523: This is debug log? Should we remove this?

internal/venice-common/src/main/resources/avro/StoreMetaValue/v16/StoreMetaValue.avsc
```diff
@@ -157,6 +157,7 @@
             {"name": "activeActiveReplicationEnabled", "type": "boolean", "default": false, "doc": "Whether or not active/active replication is enabled for hybrid stores; eventually this config will replace native replication flag, when all stores are on A/A"},
             {"name": "applyTargetVersionFilterForIncPush", "type": "boolean", "default": false, "doc": "Whether or not the target version field in Kafka messages will be used in increment push to RT policy"},
             {"name": "minCompactionLagSeconds", "type": "long", "default": -1, "doc": "Store level min compaction lag config and if not specified, it will use the global config for version topics"},
+            {"name": "maxCompactionLagSeconds", "type": "long", "default": -1, "doc": "Store level max compaction lag config and if not specified, 'max.compaction.lag.ms' config won't be setup in the corresponding version topics"},
```
sixpluszero, line 160: Hmm, can we do this in-place protocol upgrade here?

sixpluszero (commented):
Overall looks good, just the protocol part make me a bit concerned as commented below.

internal/venice-common/src/main/resources/avro/StoreMetaValue/v16/StoreMetaValue.avsc
gaojieliu: Good catch, and this is a mistake.

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/consumer/AdminExecutionTask.java
gaojieliu: Fixed.

shuhui-liu (approved):

PR #672, opened 2023-09-28
[server][controller] KME Schema Registration To ZK On Startup

lyssarzr:
## Summary
Today, only `venice-0` leader controller is allowed to register new KME schema into ZK. This change is intended to allow every other controller and server to register KME schema during startup.

Resolves VENG-10678

-->

### DIFF
--- a/services/venice-server/src/main/java/com/linkedin/venice/server/VeniceServer.java
+++ b/services/venice-server/src/main/java/com/linkedin/venice/server/VeniceServer.java
@@ -207,11 +207,11 @@ public class VeniceServer {
       String localControllerUrl = serverConfig.getLocalControllerUrl();
       String d2ServiceName = serverConfig.getLocalControllerD2ServiceName();
       String d2ZkHost = serverConfig.getLocalD2ZkHost();
-      String cluster = serverConfig.getSystemSchemaClusterName();
+      String systemStoreCluster = serverConfig.getSystemSchemaClusterName();
       ControllerClientBackedSystemSchemaInitializer metaSystemStoreSchemaInitializer =
           new ControllerClientBackedSystemSchemaInitializer(
               AvroProtocolDefinition.METADATA_SYSTEM_SCHEMA_STORE,
-              cluster,
+              systemStoreCluster,
               AvroProtocolDefinition.METADATA_SYSTEM_SCHEMA_STORE_KEY.getCurrentProtocolVersionSchema(),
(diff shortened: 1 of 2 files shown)


services/venice-server/src/main/java/com/linkedin/venice/server/VeniceServer.java
```diff
@@ -204,19 +204,36 @@ private List<AbstractVeniceService> createServices() {
     jvmStats = new VeniceJVMStats(metricsRepository, "VeniceJVMStats");
 
     if (serverConfig.isSystemSchemaInitializationAtStartTimeEnabled()) {
+      String localControllerUrl = serverConfig.getLocalControllerUrl();
+      String d2ServiceName = serverConfig.getLocalControllerD2ServiceName();
+      String d2ZkHost = serverConfig.getLocalD2ZkHost();
+      String cluster = serverConfig.getSystemSchemaClusterName();
```
shuhui-liu, line 210: Let's rename `cluster` to `systemStoreCluster` to avoid the confusion between the cluster that the server belongs to and the cluster that system schema locates in.

services/venice-server/src/main/java/com/linkedin/venice/server/VeniceServer.java
```diff
@@ -204,19 +204,36 @@ private List<AbstractVeniceService> createServices() {
...
+              null,
+              null,
+              false,
+              sslFactory,
+              localControllerUrl,
+              d2ServiceName,
+              d2ZkHost,
               false);
       metaSystemStoreSchemaInitializer.execute();
+      kmeSchemaInitializer.execute();
```
shuhui-liu, line 236: Since we've initialized KME schema here, it's not necessary to verify the existence of KME schema later at line 259:
```
kafkaMessageEnvelopeSchemaReader.ifPresent(
          schemaReader -> new SchemaPresenceChecker(schemaReader, AvroProtocolDefinition.KAFKA_MESSAGE_ENVELOPE)
              .verifySchemaVersionPresentOrExit());
```
Please move above code to the following branch:
```
// For system schemas initialized via controller client, no need to verify again because they should already exist
      if (!serverConfig.isSystemSchemaInitializationAtStartTimeEnabled()) {
```

shuhui-liu (changes requested):

shuhui-liu (approved):

PR #673, opened 2023-09-28
[server] Don't block SIT on writing to a view

ZacAttack:
## [server] Don't block SIT on writing to a view
The logic that we'll employ is that only after all view writers have ACK'd back, we'll then invoke the normal logic.

Also added a fix to squelch the spammy logs from abstract storage engine in the producer config, as well as fixing the linger time on the writer in the changecapture view writer.
Resolves #XXX

## How was this PR tested?
Integration tests

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
@@ -497,6 +497,7 @@ public class ActiveActiveStoreIngestionTask extends LeaderFollowerStoreIngestion
       // Get Futures and
       // Pass callback with on completion which decrements, once all decrement call producePutOrDeleteToKafka
       if (this.viewWriters.size() > 0) {
+        Long preprocessingTime = System.currentTimeMillis();
         CompletableFuture
             .allOf(
                 this.viewWriters.values()
@@ -513,6 +514,7 @@ public class ActiveActiveStoreIngestionTask extends LeaderFollowerStoreIngestion
                     .collect(Collectors.toList())
                     .toArray(new CompletableFuture[this.viewWriters.size()]))
             .whenCompleteAsync((value, exception) -> {
+              hostLevelIngestionStats.recordViewProducerLatency(LatencyUtils.getLatencyInMS(preprocessingTime));
               if (exception == null) {
                 producePutOrDeleteToKafka(
                     mergeConflictResult,
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
@@ -19,7 +19,6 @@ import com.linkedin.davinci.storage.chunking.RawBytesChunkingAdapter;
 import com.linkedin.davinci.storage.chunking.SingleGetChunkingAdapter;
 import com.linkedin.davinci.store.cache.backend.ObjectCacheBackend;
 import com.linkedin.davinci.store.record.ValueRecord;
-import com.linkedin.davinci.store.view.ViewWriterUtils;
 import com.linkedin.davinci.utils.ByteArrayKey;
 import com.linkedin.venice.exceptions.PersistenceFailureException;
 import com.linkedin.venice.exceptions.VeniceException;
@@ -40,7 +39,6 @@ import com.linkedin.venice.meta.Version;
 import com.linkedin.venice.offsets.OffsetRecord;
 import com.linkedin.venice.pubsub.PubSubTopicPartitionImpl;
 import com.linkedin.venice.pubsub.api.PubSubMessage;
-import com.linkedin.venice.pubsub.api.PubSubProducerCallback;
 import com.linkedin.venice.pubsub.api.PubSubTopic;
 import com.linkedin.venice.pubsub.api.PubSubTopicPartition;
 import com.linkedin.venice.schema.rmd.RmdUtils;
@@ -65,10 +63,12 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.Properties;
 import java.util.Set;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.BiConsumer;
 import java.util.function.BooleanSupplier;
+import java.util.stream.Collectors;
 import org.apache.avro.generic.GenericRecord;
 import org.apache.avro.io.BinaryDecoder;
 import org.apache.logging.log4j.LogManager;
@@ -362,7 +362,7 @@ public class ActiveActiveStoreIngestionTask extends LeaderFollowerStoreIngestion
       long beforeProcessingRecordTimestampNs,
       long currentTimeForMetricsMs) {
     /**
-     * With {@link com.linkedin.davinci.replication.BatchConflictResolutionPolicy.BATCH_WRITE_LOSES} there is no need
+     * With {@link BatchConflictResolutionPolicy.BATCH_WRITE_LOSES} there is no need
      * to perform DCR before EOP and L/F DIV passthrough mode should be used. If the version is going through data
      * recovery then there is no need to perform DCR until we completed data recovery and switched to consume from RT.
      * TODO. We need to refactor this logic when we support other batch conflict resolution policy.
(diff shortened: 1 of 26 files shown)


nisargthakkar (dismissed):
2 nit comments. Don't matter much, I guess

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/view/ChangeCaptureViewWriter.java
```diff
@@ -196,6 +188,8 @@ synchronized private void initializeVeniceWriter(int version) {
     if (veniceWriter != null) {
       return;
     }
+    // We don't want any linger time for this producer
+    props.put(KAFKA_LINGER_MS, 0);
```
nisargthakkar, line 192: Can we also allow overriding arbitrary pubsub configs?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
```diff
@@ -491,30 +493,62 @@ protected void processMessageAndMaybeProduceToKafka(
...
+                        (v) -> v.processRecord(
+                            mergeConflictResult.getNewValue(),
+                            oldValueProvider.get(),
+                            keyBytes,
+                            versionNumber,
+                            incomingValueSchemaId,
+                            valueSchemaId,
+                            mergeConflictResult.getRmdRecord()))
+                    .collect(Collectors.toList())
+                    .toArray(new CompletableFuture[this.viewWriters.size()]))
```
FelixGV, line 515: Since this function is called for each write, I prefer that we avoid the Java stream operators which are known to generate a lot of garbage. A simple loop will do, e.g.:

```
      CompletableFuture[] viewWriterFutures = new CompletableFuture[this.viewWriters.size()];
      int index = 0;
      for (VeniceViewWriter viewWriter: this.viewWriters.values()) {
        viewWriterFutures[index++] = viewWriter.processRecord(
            mergeConflictResult.getNewValue(),
            oldValueProvider.get(),
            keyBytes,
            versionNumber,
            incomingValueSchemaId,
            valueSchemaId,
            mergeConflictResult.getRmdRecord());
      }
      CompletableFuture.allOf(viewWriterFutures)...
```

It's not only more efficient but also shorter and less indented to do it the old school way... what's not to love 😅 ?

FelixGV (commented):
Minor nitpick which should be easy to address...

There are also test failures. We should take a look and see if they're related to the change or just flaky...

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
ZacAttack: I feel so validated by this.  My brain always wants to think in loops and writing streams is always my least favorite thing to code, but I've had an overwhelming sense that it's what cool coders do, and I should do it too.  Will convert to loop.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/view/ChangeCaptureViewWriter.java
ZacAttack: This is added in the VeniceView logic (basically adding viewParameters to props).  But, I think I should maybe remove this config in case we get cold feet.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
```diff
@@ -491,30 +494,65 @@ protected void processMessageAndMaybeProduceToKafka(
...
+                  mergeConflictResult.getNewValue(),
+                  oldValueProvider.get(),
+                  keyBytes,
+                  versionNumber,
+                  incomingValueSchemaId,
+                  valueSchemaId,
+                  mergeConflictResult.getRmdRecord()));
+        }
+        viewWriteCallbacks.add(partitionConsumptionState.getLastVTProduceCallFuture());
+        CompletableFuture.allOf(viewWriteCallbacks.toArray(new CompletableFuture[0]))
```
FelixGV, line 516: Since `CompletableFuture.allOf` expects an array, can we just create that direct directly, instead of passing by an `ArrayList` first? We can initialize the array at the size of the views map + 1. That would save one object and one array copy from each write...

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
```diff
@@ -491,30 +494,65 @@ protected void processMessageAndMaybeProduceToKafka(
...
+            .whenCompleteAsync((value, exception) -> {
+              hostLevelIngestionStats.recordViewProducerLatency(LatencyUtils.getLatencyInMS(preprocessingTime));
+              if (exception == null) {
+                partitionConsumptionState.setLastVTProduceCallFuture(CompletableFuture.runAsync(() -> {
```
FelixGV, line 520: Hmm... IDK if the design we discussed works if this CF is declared inside the `whenCompleteAsync` 🤔 ...

There could be many writes in a row that use the same "previous VT write future" as we're waiting for this `whenCompleteAsync` closure to kick in, no?

FelixGV (commented):
Thanks for iterating on this... I'm still not sure if the latest code structure is right... will writes to VT be guaranteed to be ordered the sale way as the RT writes?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
ZacAttack: I think this works, but I think theres one last bit of tedium here.  If we set the future stored on the PCS as one which has completed exceptionally, how do we clear it?  If we take the last one and it had an exception and then set the stored future as a 'allOf' then I think it'll be exceptional every time?  I think we'll hit a doom loop.  We need a mechanism to clear it I think....

ZacAttack: Maybe at the top check if it already completed exceptionally and bubble up?

FelixGV: Hmm... yeah, that's a doom loop if I've ever seen one. If we fail to write to VT, then I guess it's not right to move on... we should either stall and retry, or fail outright. And either way, we should record that we failed via a metric... today we have error replica metrics, but after this swallowing  future refactor, I guess that metric won't have a chance of firing... We should find a way to bubble it up, somehow, but I'm not sure what is a good code structure to get there.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/view/ChangeCaptureViewWriter.java
```diff
@@ -83,7 +85,15 @@ public Future<PubSubProduceResult> processRecord(
       initializeVeniceWriter(version);
     }
     // TODO: RecordChangeEvent isn't versioned today.
-    return veniceWriter.put(key, recordChangeEvent, 1);
+    return CompletableFuture.supplyAsync(() -> {
+      try {
+        return (PubSubProduceResult) veniceWriter.put(key, recordChangeEvent, 1).get();
```
FelixGV, line 90: Hmm... why create a new CF which only does `.get()` of a `Future`? Is it to convert a `Future` into a `CF`? What is ultimately returned by this thing is `ApacheKafkaProducerCallback::getProduceResultFuture`, which is a function whose signature returns a `Future`, but whose implementation returns its `private final CompletableFuture<PubSubProduceResult> produceResultFuture` property. Maybe we should just change the signatures to return `CF` instead of `Future`, so this wrapping here isn't necessary?

ZacAttack: The signatures in veniceWriter?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
ZacAttack: `          } else {
            VeniceException veniceException = new VeniceException(exception);
            this.setIngestionException(partitionConsumptionState.getPartition(), veniceException);
            currentVersionTopicWrite.completeExceptionally(veniceException);
          }`

FelixGV: Sure, but if we call `currentVersionTopicWrite.completeExceptionally` then we're in a doom loop, right?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/view/ChangeCaptureViewWriter.java
ZacAttack: Yeah, this purely exists to just get the right object type :/

I just tweaked the veniceWriter interface.  Might make pulling it in internally a bit of a pain?  Not sure.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
ZacAttack: We will be in a doom loop for a while, but my understanding of the ingestionException is that if theres an exception in there the task will fail eventually...  I think.

FelixGV: SG.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/view/ChangeCaptureViewWriter.java
FelixGV: We've gotta do what we've gotta do 🤷

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/view/VeniceViewWriter.java
```diff
@@ -43,14 +46,15 @@ public VeniceViewWriter(
    * @param oldValueSchemaId the schemaId of the old record
    * @param replicationMetadataRecord the associated RMD for the incoming record.
    */
-  public void processRecord(
+  public Future<PubSubProduceResult> processRecord(
       ByteBuffer newValue,
       ByteBuffer oldValue,
```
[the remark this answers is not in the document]
ZacAttack: This is a good idea, but I'm going to defer tweaking it until it's a bit more useful.  Truthfully we should probably do some broader tweaking here.  We have some latent desire to have views that exist in followers (davinci view!), but the code right now basically mandates that this has to live in Leader... idk if we can excise this property, so we might need to split the concept into "leader" and "follower" views (push vs. pull?). idk.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/view/ChangeCaptureViewWriter.java
```diff
@@ -196,6 +188,7 @@ synchronized private void initializeVeniceWriter(int version) {
     if (veniceWriter != null) {
       return;
     }
+    // We don't want any linger time for this producer
```
FelixGV, line 191: This looks like a dangling comment... is it from an earlier iteration of the change?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/view/VeniceViewWriter.java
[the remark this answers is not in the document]
FelixGV: Yeah, good point... we should do this refactor separately.

FelixGV (commented):
LGTM, just a minor comment about a (probably stale?) comment... I think this is good to go.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/view/ChangeCaptureViewWriter.java
ZacAttack: good catch, removing

FelixGV (approved):
Thanks Zac! LGTM!

FelixGV (dismissed):
Thanks Zac, the CI says `:services:venice-controller:test'` also failed, but this test is only about dvc. Is there a fix needed in the controller module as well?

FelixGV (approved):
Hopefully that's the last mock to fix 😅 !

PR #674, opened 2023-09-29
[server] Ignore invalid time stamps for leaderProducerRecordContext.

asaxena76:

Refactored nearline metrics recording code. Ignore invalid timestamps and don't calculate latency for messages with no upstream.
Filtered out junk metric data and emit zero when latency values are unavailable. This avoids junk values for max latency when they are aggregated across time dimension.

Resolves #XXX

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
@@ -2639,24 +2639,12 @@ public abstract class StoreIngestionTask implements Runnable, Closeable {
             partitionConsumptionState,
             leaderProducedRecordContext,
             currentTimeMs);
-        if (!isUserSystemStore() && isHybridMode() && partitionConsumptionState.hasLagCaughtUp()) {
-          long afterProcessingRecordTimestampMs = System.currentTimeMillis();
-          long brokerProducedTimeStamp = (leaderProducedRecordContext == null)
-              ? kafkaValue.producerMetadata.messageTimestamp
-              : leaderProducedRecordContext.getProducedTimestampMs();
-          /** This metric is used in combination with the nearlineProducerToLocalBrokerLatency metric to measure the
-           * end to end latency of a record from the time it is produced to the time it is ready to serve. If we add
-           * them together we get the approx end to end latency of a message. Though the time taken to receive broker
-           * acknowledgement will be double counted for followers. This is estimated to be very small (1%)
-           * TODO: We plan to upgrade the KME to include the producer timestamp after the KME prrotocol upgrade
-           * enhancements and that will give us more accurate E2E latency.
-           */
-          versionedIngestionStats.recordNearlineLocalBrokerToReadyToServeLatency(
-              storeName,
-              versionNumber,
-              afterProcessingRecordTimestampMs - brokerProducedTimeStamp,
-              afterProcessingRecordTimestampMs);
-        }
+        recordNearlineLocalBrokerToReadyToServerLatency(
+            storeName,
+            versionNumber,
+            partitionConsumptionState,
+            kafkaValue,
+            leaderProducedRecordContext);
       }
       versionedIngestionStats.recordConsumedRecordEndToEndProcessingLatency(
           storeName,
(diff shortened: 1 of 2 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -3569,6 +3557,69 @@ protected enum DelegateConsumerRecordResult {
...
   }
 
+  private void recordNearlineLocalBrokerToReadyToServerLatency(
+      String storeName,
+      int versionNumber,
+      PartitionConsumptionState partitionConsumptionState,
+      KafkaMessageEnvelope kafkaMessageEnvelope,
+      LeaderProducedRecordContext leaderProducedRecordContext) {
+    long afterProcessingRecordTimestampMs = System.currentTimeMillis();
+    if (!isUserSystemStore() && isHybridMode() && partitionConsumptionState.hasLagCaughtUp()
```
adamxchen, line 3567: Haven't checked the logic in details yet but I think we can make line 3567 - 3579 a function, that's to determine whether we need to record the latency, so it improves the readability. Some javadoc would be helpful too

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -3569,6 +3557,69 @@ protected enum DelegateConsumerRecordResult {
     SKIPPED_MESSAGE
   }
 
+  private void recordNearlineLocalBrokerToReadyToServerLatency(
```
huangminchn, line 3560: Could you please wrap the entire function within a try-catch block, and tolerates all exception except InterruptException?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -3569,6 +3557,69 @@ protected enum DelegateConsumerRecordResult {
...
+            "LeaderMetadataFooter is null when PCS state is STANDBY for storeName: {},"
+                + " versionNumber: {}, partition: {}",
+            storeName,
+            versionNumber,
+            partitionConsumptionState.getPartition());
+        return;
+      }
+      boolean hasCorrespondingUpstreamMessage = false; // No point in recording E2E latency if there is no upstream
+      if (partitionConsumptionState.getLeaderFollowerState() == LeaderFollowerStateType.LEADER) {
+        hasCorrespondingUpstreamMessage = leaderProducedRecordContext.hasCorrespondingUpstreamMessage();
```
huangminchn, line 3592: "internalProcessConsumerRecord" is called inside drainer thread, while the state transition happens in the ingestion thread. I would recommend using the "leaderProducerRecordContext" as source of truth instead of using the leader/follower state, like this:
"""
if (leaderProducerdRecordContext != null) {
    hasCorrespondingUpstreamMessage = leaderProducedRecordContext.hasCorrespondingUpstreamMessage();
} else if (kafkaMessageEnvelope.getLeaderMetadataFooter() != null) {
    kafkaMessageEnvelope.getLeaderMetadataFooter().getUpstreamOffset() >= 0;
}
"""

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -3569,6 +3557,69 @@ protected enum DelegateConsumerRecordResult {
...
+          LOGGER.warn(
+              "Illegal timestamp for storeName: {}, versionNumber: {}, partition: {}, "
+                  + "leaderProducedRecordContext: {}, leaderMetadataFooter: {}",
+              storeName,
+              versionNumber,
+              partitionConsumptionState.getPartition(),
+              leaderProducedRecordContext == null ? "NA" : leaderProducedRecordContext,
+              kafkaMessageEnvelope.getLeaderMetadataFooter() == null
+                  ? "NA"
+                  : kafkaMessageEnvelope.getLeaderMetadataFooter());
```
huangminchn, line 3611: Same here please - please use redundant log filtering; and this time we cannot simply use the message since offset is in the message; you can break the message into two parts, where the first part only has store name, version and partition, then use the first part as filtering input.

I am worried that the new logs would end up flooding the log file.

huangminchn (changes requested):
Thanks Amit!

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -3569,6 +3557,69 @@ protected enum DelegateConsumerRecordResult {
...
+  private void recordNearlineLocalBrokerToReadyToServerLatency(
+      String storeName,
+      int versionNumber,
+      PartitionConsumptionState partitionConsumptionState,
+      KafkaMessageEnvelope kafkaMessageEnvelope,
+      LeaderProducedRecordContext leaderProducedRecordContext) {
+    long afterProcessingRecordTimestampMs = System.currentTimeMillis();
+    if (!isUserSystemStore() && isHybridMode() && partitionConsumptionState.hasLagCaughtUp()
+        && (partitionConsumptionState.getLeaderFollowerState() == LeaderFollowerStateType.LEADER
+            || partitionConsumptionState.getLeaderFollowerState() == LeaderFollowerStateType.STANDBY)) {
```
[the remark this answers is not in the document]
asaxena76: So I wanted to make sure that the ingestion task is in an online state when we are calculating the E2E latency. We already have the condition pcs.isLagCaughtUp but I have a suspicion that that we might come here when we are transitioning between states . We shouldn't be ingesting if we are offline or dropped unless we are transitioning. Here I was trying to assert the replicas is live .

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
asaxena76: Are you suggesting that the PCS state type is not a reliable indicator of liveness? What I am trying to avoid is measuring latency where we are going from offline to standby / leader but aren't ready to serve yet. The pcs.hasLagCaughtUp should  exclude that situation but I thought additional check on the state would be good. In other words the check is (partitionConsumptionState.getLeaderFollowerState() == LeaderFollowerStateType.LEADER
          && leaderProducedRecordContext == null) which is stronger than just leaderProducedRecordContext == null . Any disadvantage if we test for both?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
asaxena76: ok. will use the filtering

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
asaxena76: Is this necessary ? The calling method already does this

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
asaxena76: Shortened the condition . since it fits in 1 line don't think adding a method will increase readability. added javadoc as well

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
asaxena76: As discussed. Will not be using leader follower state since that can be outdated here.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
huangminchn: It was necessary; the previous diff was much more complicated; if we bubble the exception to the callers, we will end up with failing the ingestion.

The latest diff is fine, since we just added a new log.

huangminchn (approved):
Thanks Amit!

PR #676, opened 2023-10-02
[server][dvc] Fix to RocksDB atomic flush usage

FelixGV:
Even though we use the DB Options to enable atomic flush, we were not passing the list of column family handles into the flush call, and since our WAL is disabled we did not implicitly get the atomic flush behavior either.

This commit fixes the usage to ensure atomic flushing of the data and RMD column families.

More details: https://github.com/facebook/rocksdb/wiki/Atomic-flush


majisourav99 (approved):
Thanks Felix for the quick fix.

FelixGV:
> Thanks Felix for the quick fix.

I am but the messenger 😆 ... most of the hard work of this investigation was done by @sixpluszero, and some more by you and @nisargthakkar too 🤗 

Thanks for shipping it! I'll merge it if/when the CI passes...

PR #677, opened 2023-10-02
[controller] Handle NPE for incremental push reporting

adamxchen:
## Summary

We recently find out that NPE would be thrown when polling the incremental push status. This PR adds the NPE handling so we would have more information on which partition couldn't be found. Please note that this PR is NOT for resolving the real problem but attempts to provide more debugging information

## How was this PR tested?

unit tests

### DIFF
--- a/internal/venice-common/src/main/java/com/linkedin/venice/pushmonitor/OfflinePushStatus.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/pushmonitor/OfflinePushStatus.java
@@ -208,6 +208,10 @@ public class OfflinePushStatus {
     for (PartitionStatus partitionStatus: getPartitionStatuses()) {
       Map<CharSequence, Integer> partitionPushStatus = new HashMap<>();
       Partition partition = partitionAssignment.getPartition(partitionStatus.getPartitionId());
+      if (partition == null) {
+        throw new VeniceException(
+            "Partition " + partitionStatus.getPartitionId() + " not found in partition assignment");
+      }
       Set<String> workingInstances =
           partition.getWorkingInstances().stream().map(Instance::getNodeId).collect(Collectors.toSet());
       for (ReplicaStatus replicaStatus: partitionStatus.getReplicaStatuses()) {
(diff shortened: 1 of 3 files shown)


internal/venice-common/src/main/java/com/linkedin/venice/pushmonitor/OfflinePushStatus.java
```diff
@@ -208,6 +208,10 @@ public Map<Integer, Map<CharSequence, Integer>> getIncrementalPushStatus(
     for (PartitionStatus partitionStatus: getPartitionStatuses()) {
       Map<CharSequence, Integer> partitionPushStatus = new HashMap<>();
       Partition partition = partitionAssignment.getPartition(partitionStatus.getPartitionId());
+      if (partition == null) {
```
majisourav99, line 211: when this could happen?

adamxchen: It happened a few times

sixpluszero: Do you happen to have the error message for this issue, I am wondering if InLogs is still available to reveal this. But I guess the guard here is still valid.

sixpluszero: And also, if we handle NPE by throw Exception here, then it is still throw Exception in controller? I am wondering at which level do we catch the exception and do some proper logging... Otherwise it does not change much, we just wrap the NPE with another exception (not sure if it is what you want though)

adamxchen: Right. This change is not trying to resolve the true cause behind it but attempts to reveal more information otherwise it's very difficult to debug unless we have the heap dump when the incident happens.

majisourav99: was wondering if there is some transient issue to update the routing repo. we could put NOT_CREATED status for that null partition in the map and log error ? ie    
   incPushStatusMap.put(partitionId, NOT_CREATED);

adamxchen: > was wondering if there is some transient issue to update the routing repo. we could put NOT_CREATED status for that null partition in the map and log error ?

It could be a good idea but let's confirm if that's indeed some transient issue to update the routing repos otherwise the suggested change may hide the real NPE issue

sixpluszero (approved):

PR #678, opened 2023-10-03
[admin-tool] Add command to set up change capture view for a store

ZacAttack:
## [admin-tool] Add command to set up change capture view for a store
Still some pending work to make update store work for command line json inputs, but this is probably an easier command to work with in any case.  This will set up needed features like AA and chunking and then enable the view

Resolves #XXX

## How was this PR tested?
Add new unit and integration tests


clients/venice-admin-tool/src/main/java/com/linkedin/venice/Command.java
```diff
@@ -1,117 +1,6 @@
...
-import static com.linkedin.venice.Arg.VALUE_SCHEMA_ID;
-import static com.linkedin.venice.Arg.VENICE_CLIENT_SSL_CONFIG_FILE;
-import static com.linkedin.venice.Arg.VENICE_ZOOKEEPER_URL;
-import static com.linkedin.venice.Arg.VERSION;
-import static com.linkedin.venice.Arg.VOLDEMORT_STORE;
-import static com.linkedin.venice.Arg.VSON_STORE;
-import static com.linkedin.venice.Arg.WRITEABILITY;
-import static com.linkedin.venice.Arg.WRITE_COMPUTATION_ENABLED;
-import static com.linkedin.venice.Arg.ZK_SSL_CONFIG_FILE;
+import static com.linkedin.venice.Arg.*;
```
FelixGV, line 3: Can you revert that please?

clients/venice-admin-tool/src/main/java/com/linkedin/venice/Command.java
```diff
@@ -474,6 +363,11 @@ public enum Command {
       "dump-ingestion-state",
       "Dump the real-time ingestion state for a certain store version in a certain storage node",
       new Arg[] { SERVER_URL, STORE, VERSION }, new Arg[] { PARTITION }
+  ),
+  ADD_CHANGE_CAPTURE_VIEW(
+      "add-change-capture-view",
+      "Add change capture to a given store. This will enable all required configs including AA replication and chunking. View name argument is optional, default view name is changeCaptureView",
+      new Arg[] { URL, STORE, CLUSTER }, new Arg[] { VIEW_NAME }
```
FelixGV, line 370: What is the name for, BTW...?

clients/venice-admin-tool/src/main/java/com/linkedin/venice/AdminTool.java
```diff
@@ -2881,6 +2885,31 @@ private static void dumpIngestionState(CommandLine cmd) throws Exception {
...
+        .setRmdChunkingEnabled(true);
+    ControllerResponse response = client.updateStore(storeName, params);
+    if (response.isError()) {
+      printErrAndExit(response.getError());
+    }
+    Map<String, String> views = new HashMap<>();
+
+    views.put(
+        getOptionalArgument(cmd, Arg.VIEW_NAME, ChangeCaptureView.DEFAULT_CHANGE_CAPTURE_VIEW_NAME),
+        "{\"viewClassName\" : \"" + ChangeCaptureView.class.getCanonicalName() + "\", \"viewParameters\" : {}}");
```
FelixGV, line 2907: Should we support passing view params, or not necessary for this view...?

clients/venice-admin-tool/src/main/java/com/linkedin/venice/Command.java
```diff
@@ -474,6 +363,11 @@ public enum Command {
       "dump-ingestion-state",
       "Dump the real-time ingestion state for a certain store version in a certain storage node",
       new Arg[] { SERVER_URL, STORE, VERSION }, new Arg[] { PARTITION }
+  ),
+  ADD_CHANGE_CAPTURE_VIEW(
```
FelixGV, line 367: Should we also have a command for adding a view of any name? Or maybe it doesn't matter for now since we have no other view implementations...

FelixGV (commented):
Thanks Zac! A few minor comments...

clients/venice-admin-tool/src/main/java/com/linkedin/venice/Command.java
ZacAttack: So the way views are configured for a store are a map of Map<String, ViewConfig>.  The intention of the key is to be able to have a handle for which view you want to edit.  This way, it's conceivable that a store could have multiple views of the same type, but perhaps with different configurations (admittedly for the change capture view, this wouldn't work, or at least, wouldn't work in a useful way).  The optional argument is if you wanted to override the default name for some reason.

....Though... maybe this doesn't make sense now that I type it out.  Maybe it'd be better to just not have it.

clients/venice-admin-tool/src/main/java/com/linkedin/venice/AdminTool.java
ZacAttack: I think we could/should add that if we can puzzle out better why passing json args on the command line isn't working as intended.  I didn't add it here because I was in a hurry.  Best option would be to figure out why json passing isn't working (and thereby, fix other commands which are currently broken) and then add that here as well.

clients/venice-admin-tool/src/main/java/com/linkedin/venice/Command.java
ZacAttack: Yeah can add that.

We also need to add a patch to the controller for removing a view.  That currently doesn't work as intended :(

sixpluszero:
Closing this PR, as #695 addresses the issue already.

PR #680, opened 2023-10-04
[all] Minor Utils cleanups

FelixGV:
- The debug info map inserted in SOS control messages now omits entries for which it could not find the info. Log statements related to the info gathering are also lowered, as they were noisy.

- Deleted dead code.

- Addressed various IDE hints in the Utils class.

## How was this PR tested?
CI


lluwm (approved):
Thanks for the cleanup and it looks good to me!

FelixGV:
Thanks for the review @lluwm!

PR #684, opened 2023-10-06
[controller] add cluster name to the 'Venice-Admin-Execution-Task' th…

lluwm:
…reads

In controller, each cluster has one corresponding AdminConsumptionTask and multiple 'Venice-Admin-Execution-Task' to consume the admin messages and then execute the admin operations. Today, it is not easy to differentiate which cluster one execution task belongs to, when analyzing controller logs or Java heap dump. This change appends the cluster name to the existing thread name prefix to make it distinguishable.

## How was this PR tested?
CI


m-nagarajan (approved):
LGTM, thanks @lluwm

PR #689, opened 2023-10-10
[controller] Do not delete backup versions for repush

majisourav99:
## Do not delete backup versions for repush

At the start of a push, Venice deletes backup versions to save disk space. But for repush, since the current version is same as repush version, the backup version would be the actual backup version and deleting it could cause issues when users would want to rollback. This PR blocks deletion of backup versions for repushes. 
## How was this PR tested?

GHCI

### DIFF
--- a/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
@@ -16,6 +16,7 @@ import static com.linkedin.venice.meta.HybridStoreConfigImpl.DEFAULT_HYBRID_TIME
 import static com.linkedin.venice.meta.HybridStoreConfigImpl.DEFAULT_REWIND_TIME_IN_SECONDS;
 import static com.linkedin.venice.meta.Store.NON_EXISTING_VERSION;
 import static com.linkedin.venice.meta.Version.PushType;
+import static com.linkedin.venice.meta.Version.VENICE_RE_PUSH_PUSH_ID_PREFIX;
 import static com.linkedin.venice.meta.VersionStatus.ERROR;
 import static com.linkedin.venice.meta.VersionStatus.NOT_CREATED;
 import static com.linkedin.venice.meta.VersionStatus.ONLINE;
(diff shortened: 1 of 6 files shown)


services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -2642,9 +2643,10 @@ private Pair<Boolean, Version> addVersion(
         }
 
         if (startIngestion) {
-          // Early delete backup version on start of a push, controlled by store config earlyDeleteBackupEnabled
+          // Early delete backup version on start of a push, except for repush.
           if (backupStrategy == BackupStrategy.DELETE_ON_NEW_PUSH_START
-              && multiClusterConfigs.getControllerConfig(clusterName).isEarlyDeleteBackUpEnabled()) {
+              && multiClusterConfigs.getControllerConfig(clusterName).isEarlyDeleteBackUpEnabled()
+              && !pushJobId.startsWith(VENICE_RE_PUSH_PUSH_ID_PREFIX)) {
```
adamxchen, line 2649: Thanks for the quick fix! Can we add one unit test and integ test to verify it?

gaojieliu (changes requested):
With this change, when will the backup version be deleted?

let us say, currently, there are 2 versions: v1, v2 and v3 is a repushed version, so does it mean the end state will be: v1, v2, v3?
We should delete v2 at some point after v3 becomes the new current version.

majisourav99:
> With this change, when will the backup version be deleted?
> 
> let us say, currently, there are 2 versions: v1, v2 and v3 is a repushed version, so does it mean the end state will be: v1, v2, v3? We should delete v2 at some point after v3 becomes the new current version.

We have backup resource deleter task which will delete them after some time. We can expedite that to delete the middle version. But for target region push, it wont be that straightforward as some colo, v2 will be missing

gaojieliu:
> > With this change, when will the backup version be deleted?
> > let us say, currently, there are 2 versions: v1, v2 and v3 is a repushed version, so does it mean the end state will be: v1, v2, v3? We should delete v2 at some point after v3 becomes the new current version.
> 
> We have backup resource deleter task which will delete them after some time. We can expedite that to delete the middle version. But for target region push, it wont be that straightforward as some colo, v2 will be missing

If we do more repushes, such as v4, v5, what will happen?
We should try to delete the duplicate versions since they will consume extra disk space.
For regions with missing v2, they can just ignore the deletion of v2.

majisourav99:
> > > With this change, when will the backup version be deleted?
> > > let us say, currently, there are 2 versions: v1, v2 and v3 is a repushed version, so does it mean the end state will be: v1, v2, v3? We should delete v2 at some point after v3 becomes the new current version.
> > 
> > 
> > We have backup resource deleter task which will delete them after some time. We can expedite that to delete the middle version. But for target region push, it wont be that straightforward as some colo, v2 will be missing
> 
> If we do more repushes, such as v4, v5, what will happen? We should try to delete the duplicate versions since they will consume extra disk space. For regions with missing v2, they can just ignore the deletion of v2.

We will delete all but `minNumberOfStoreVersionsToPreserve` veresions in `handleCompletedPush`. Also for repush when we try to repair bad versions, its not clear which version to keep and which to delete. We can try to delete the source version in repush source colo, but that would just save resource in one colo.

PR #690, opened 2023-10-10
[all][server][dvc] Fast-avro upgrade, fix meta system store schema update issue and partition size race condition

gaojieliu:
This code change includes several fixes:
1. Bump up fast-avro lib tot pick up the recent serde optimizations.
2. Fix the meta system store schema race condition issue in Venice Server. Previously, the update schema may not be available for the latest superset schema since there is a delay between registering superset schema and its derived schema. The fix will always retrieve a latest value schema with a valid derived schema.
3. Use RocksDB API to fetch partition size to avoid the following race condition: RocksDB log compaction can delete file in the middle of measuring the directory size.
4. Let 'fast.avro.field.limit.per.method' to control fast serializer size as well.

## How was this PR tested?
CI

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBStoragePartition.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBStoragePartition.java
@@ -25,7 +25,6 @@ import java.util.Set;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
 import java.util.function.Supplier;
 import javax.annotation.concurrent.NotThreadSafe;
-import org.apache.commons.io.FileUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.rocksdb.BlockBasedTableConfig;
@@ -923,21 +922,16 @@ public class RocksDBStoragePartition extends AbstractStoragePartition {
         && writeOnly == partitionConfig.isWriteOnlyConfig();
   }
 
-  /**
-   * This method calculates the file size by adding all subdirectories size
-   * @return the partition db size in bytes
-   */
   @Override
   public long getPartitionSizeInBytes() {
-    File partitionDbDir = new File(fullPathForPartitionDB);
-    if (partitionDbDir.exists()) {
-      /**
-       * {@link FileUtils#sizeOf(File)} will throw {@link IllegalArgumentException} if the file/dir doesn't exist.
-       */
-      return FileUtils.sizeOf(partitionDbDir);
-    } else {
-      return 0;
+    readCloseRWLock.readLock().lock();
+    try {
+      makeSureRocksDBIsStillOpen();
+      return getRocksDBStatValue("rocksdb.live-sst-files-size");
+    } finally {
+      readCloseRWLock.readLock().unlock();
     }
+
   }
 
   protected Options getOptions() {
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBStoragePartition.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBStoragePartition.java
@@ -931,7 +931,6 @@ public class RocksDBStoragePartition extends AbstractStoragePartition {
     } finally {
       readCloseRWLock.readLock().unlock();
     }
-
   }
 
   protected Options getOptions() {
--- a/internal/venice-common/src/main/java/com/linkedin/venice/meta/ReadOnlySchemaRepository.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/meta/ReadOnlySchemaRepository.java
@@ -83,8 +83,8 @@ public interface ReadOnlySchemaRepository extends VeniceResource {
       DerivedSchemaEntry latestDerivedSchemaEntry = null;
       for (DerivedSchemaEntry entry: derivedSchemaEntries) {
         if (latestDerivedSchemaEntry == null || entry.getValueSchemaID() > latestDerivedSchemaEntry.getValueSchemaID()
-            || entry.getValueSchemaID() == latestDerivedSchemaEntry.getValueSchemaID()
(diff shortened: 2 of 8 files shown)


internal/venice-common/src/main/java/com/linkedin/venice/meta/ReadOnlySchemaRepository.java
```diff
@@ -68,4 +69,29 @@ public interface ReadOnlySchemaRepository extends VeniceResource {
...
+      /**
+       * Can't find the derived schema for the latest value schema, so it will fall back to find out the latest
+       * value schema, which has a valid derived schema.
+       */
+      Collection<DerivedSchemaEntry> derivedSchemaEntries = getDerivedSchemas(storeName);
+      DerivedSchemaEntry latestDerivedSchemaEntry = null;
+      for (DerivedSchemaEntry entry: derivedSchemaEntries) {
+        if (latestDerivedSchemaEntry == null || entry.getValueSchemaID() > latestDerivedSchemaEntry.getValueSchemaID()
+            || entry.getValueSchemaID() == latestDerivedSchemaEntry.getValueSchemaID()
+                && entry.getId() > latestDerivedSchemaEntry.getId()) {
```
FelixGV, line 87: Minor nitpick on readability... I find it a bit hard to parse a sequence of mixed boolean operators... IIUC, `&&` has higher precedence than `||`, so the if condition `A || B || C && D` is equivalent to `A || B || (C && D)`. Although they are equivalent, I find the latter easier to parse visually. Or alternatively, the whole expression might be more concise and clear if using an extra function, e.g. `A || B || func(C, D)`, where `func(C, D) { return C && D; }`. What do you think?

services/venice-server/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGeneratorAccessor.java
```diff
@@ -6,5 +6,6 @@
 public class FastDeserializerGeneratorAccessor {
   public static void setFieldsPerPopulationMethod(int limit) {
     FastDeserializerGenerator.setFieldsPerPopulationMethod(limit);
+    FastSerializerGenerator.setFieldsPerRecordSerializationMethod(limit);
```
FelixGV, line 9: We should also mention this in the commit message.

internal/venice-common/src/main/java/com/linkedin/venice/meta/ReadOnlySchemaRepository.java
```diff
@@ -68,4 +69,29 @@ public interface ReadOnlySchemaRepository extends VeniceResource {
...
+  default DerivedSchemaEntry getLatestDerivedSchema(String storeName) {
+    SchemaEntry valueSchemaEntry = getSupersetOrLatestValueSchema(storeName);
+    try {
+      return getLatestDerivedSchema(storeName, valueSchemaEntry.getId());
+    } catch (Exception e) {
+      /**
+       * Can't find the derived schema for the latest value schema, so it will fall back to find out the latest
+       * value schema, which has a valid derived schema.
+       */
+      Collection<DerivedSchemaEntry> derivedSchemaEntries = getDerivedSchemas(storeName);
```
sixpluszero, line 82: I am a bit confused about this, why don't we do below logic directly? Given that it should always return the known latest one.

gaojieliu: The latest superset schema might not be the one with the max schema id.

FelixGV (commented):
Thanks! LGTM, I just have some minor comments...

internal/venice-common/src/main/java/com/linkedin/venice/meta/ReadOnlySchemaRepository.java
sixpluszero: Oh...right. This has to be the fallback logic. 

But just wondering if we evolve schema with field reduction, latest value schema and update schema is still not correct right? Even if we use the fall back logic to find a latest derived schema, it might not work? In this case we should still bounce the server to mitigate but SIT will fail still I believe?

gaojieliu: No, `SIT` will use the returned `DerivedSchemaEntry` to extract the corresponding value schema.

services/venice-server/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGeneratorAccessor.java
gaojieliu: Sure.

internal/venice-common/src/main/java/com/linkedin/venice/meta/ReadOnlySchemaRepository.java
gaojieliu: Sure, I will tweak the if condition.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBStoragePartition.java
```diff
@@ -923,20 +922,14 @@ public boolean verifyConfig(StoragePartitionConfig partitionConfig) {
         && writeOnly == partitionConfig.isWriteOnlyConfig();
   }
 
-  /**
-   * This method calculates the file size by adding all subdirectories size
-   * @return the partition db size in bytes
-   */
   @Override
   public long getPartitionSizeInBytes() {
```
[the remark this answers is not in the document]
FelixGV: There are lots of `synchronized` functions in `AbstractStorageEngine`, but none of them look to me like hot path functions...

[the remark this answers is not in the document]
sushantmane: Oh ok.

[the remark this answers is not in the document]
gaojieliu: Yeah, also all the writes to the same partitions are sequential, so the perf impact should be minimal.

PR #691, opened 2023-10-10
[WIP][fast-client] Add a flag to StoreMetadata to denote if read-computation is enabled for a store

nisargthakkar:
## Add a flag to StoreMetadata to denote if read-computation is enabled for a store

This commit enables `StoreMetadata` used by fast-client to offer an API to denote if read computation is enabled for a store or not.

## How was this PR tested?

GH CI


gaojieliu:
Looks good to me, and I guess you want to merge this PR first, so that the other PR to support read compute in fast-client can leverage this flag to guard read compute API, right?
Also it is good to have a server-side check as well, such as if the read compute is disabled from Server POV, it can reject read compute request directly.

PR #692, opened 2023-10-11
[client] Reset user token in httpclient5

majisourav99:
## Reset user token in httpclient5

When a connection lease attempt is made the state requested is null. However when the connection is put back into the pool the state is set to the X500Principal object of the SSL peer. This happens in the DefaultUserTokenHandler class. This PR overrides while creating the HttpClient class so that it will reuse the connections.

Before using httpclient5 using SSL.

client->router delay ~300ms from local mac to ltx1 due to SSL/TLS handshake while creating conn on every requests.

After this change

client->router delay ~100ms from local mac to ltx1.

## How was this PR tested?

testing using venice-thin-client.

### DIFF
--- a/internal/venice-client-common/src/main/java/com/linkedin/venice/httpclient5/HttpClient5Utils.java
+++ b/internal/venice-client-common/src/main/java/com/linkedin/venice/httpclient5/HttpClient5Utils.java
@@ -128,6 +128,7 @@ public class HttpClient5Utils {
                 RequestConfig.custom()
                     .setResponseTimeout(Timeout.ofMilliseconds(requestTimeOutInMilliseconds))
                     .setConnectionRequestTimeout(CONNECT_TIMEOUT_IN_MILLISECONDS)
+                    .setDefaultKeepAlive(1, TimeUnit.HOURS)
                     .build())
             .setUserTokenHandler((route, context) -> null)
             .build();
--- a/internal/venice-client-common/src/main/java/com/linkedin/venice/httpclient5/HttpClient5Utils.java
+++ b/internal/venice-client-common/src/main/java/com/linkedin/venice/httpclient5/HttpClient5Utils.java
@@ -8,6 +8,7 @@ import org.apache.hc.client5.http.impl.async.CloseableHttpAsyncClient;
 import org.apache.hc.client5.http.impl.async.HttpAsyncClients;
 import org.apache.hc.client5.http.impl.nio.PoolingAsyncClientConnectionManagerBuilder;
 import org.apache.hc.client5.http.ssl.ClientTlsStrategyBuilder;
+import org.apache.hc.core5.http.impl.DefaultConnectionReuseStrategy;
 import org.apache.hc.core5.http.nio.ssl.TlsStrategy;
 import org.apache.hc.core5.http.ssl.TLS;
 import org.apache.hc.core5.http2.HttpVersionPolicy;
@@ -111,6 +112,7 @@ public class HttpClient5Utils {
         return HttpAsyncClients.custom()
             .setVersionPolicy(HttpVersionPolicy.FORCE_HTTP_1)
             .setIOReactorConfig(ioReactorConfig)
+            .setConnectionReuseStrategy(new DefaultConnectionReuseStrategy())
             .setConnectionManager(
                 PoolingAsyncClientConnectionManagerBuilder.create()
                     .setMaxConnTotal(http1MaxConnectionsTotal)
@@ -127,6 +129,7 @@ public class HttpClient5Utils {
                     .setResponseTimeout(Timeout.ofMilliseconds(requestTimeOutInMilliseconds))
                     .setConnectionRequestTimeout(CONNECT_TIMEOUT_IN_MILLISECONDS)
                     .build())
+            .setUserTokenHandler((route, context) -> null)
             .build();
       } else {
         return HttpAsyncClients.customHttp2()


internal/venice-client-common/src/main/java/com/linkedin/venice/httpclient5/HttpClient5Utils.java
```diff
@@ -127,6 +129,7 @@ public CloseableHttpAsyncClient build() {
                     .setResponseTimeout(Timeout.ofMilliseconds(requestTimeOutInMilliseconds))
                     .setConnectionRequestTimeout(CONNECT_TIMEOUT_IN_MILLISECONDS)
                     .build())
+            .setUserTokenHandler((route, context) -> null)
```
FelixGV, line 132: This is a very interesting finding... so does it mean that without adding this line, the TLS handshake would be redone on every request?!

FelixGV (dismissed):
Cool stuff! Just left a few questions...

internal/venice-client-common/src/main/java/com/linkedin/venice/httpclient5/HttpClient5Utils.java
```diff
@@ -111,6 +112,7 @@ public CloseableHttpAsyncClient build() {
         return HttpAsyncClients.custom()
```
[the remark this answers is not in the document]
majisourav99: Thats correct, we got extremely lucky in FC as it uses http2 in prod.

internal/venice-client-common/src/main/java/com/linkedin/venice/httpclient5/HttpClient5Utils.java
majisourav99: Right, thats what the experiments show.

internal/venice-client-common/src/main/java/com/linkedin/venice/httpclient5/HttpClient5Utils.java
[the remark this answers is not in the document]
majisourav99: well it will help query.sh users to get faster results 😄

internal/venice-client-common/src/main/java/com/linkedin/venice/httpclient5/HttpClient5Utils.java
gaojieliu: On every request???
This shouldn't be possible, otherwise FC latency will be very bad, and it can't achieve 2ms at p99 for single key lookup.

gaojieliu: Add this is for HTTP/1.1, not HTTP/2, right?
I was overly- aranoid..

internal/venice-client-common/src/main/java/com/linkedin/venice/httpclient5/HttpClient5Utils.java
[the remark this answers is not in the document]
gaojieliu: Not pure lucky...
We did the benchmark test before rolling out to prod.
Lucky didn't play a role in http/2 prod rollout..

internal/venice-client-common/src/main/java/com/linkedin/venice/httpclient5/HttpClient5Utils.java
```diff
@@ -111,6 +112,7 @@ public CloseableHttpAsyncClient build() {
...
             .setVersionPolicy(HttpVersionPolicy.FORCE_HTTP_1)
             .setIOReactorConfig(ioReactorConfig)
+            .setConnectionReuseStrategy(new DefaultConnectionReuseStrategy())
```
gaojieliu, line 115: We may want to use `setKeepAliveStrategy` as well to setup keep-alive, otherwise the connection can be closed quickly when idle.

internal/venice-client-common/src/main/java/com/linkedin/venice/httpclient5/HttpClient5Utils.java
majisourav99: Right its only for HTTP/1, not 2.  I verified for H2 we do only 1 handshake.

gaojieliu: And is this line useful at all?
Maybe the `connectionReuseStrategy` is good enough.

internal/venice-client-common/src/main/java/com/linkedin/venice/httpclient5/HttpClient5Utils.java
majisourav99: If we dont specify it would use `DefaultConnectionKeepAliveStrategy` which had default value of 2 or 3 mins of idle time.

gaojieliu: In practice, we normally would like to keep the connection alive for at least several hours, not mins when idle.

internal/venice-client-common/src/main/java/com/linkedin/venice/httpclient5/HttpClient5Utils.java
majisourav99: hmm.. I could try that, but doubt thats the case.

internal/venice-client-common/src/main/java/com/linkedin/venice/httpclient5/HttpClient5Utils.java
majisourav99: So you want to bump the default to few 1hrs ?

gaojieliu:
```
When a connection lease attempt is made the state requested is null. However when the connection is put back into the pool the state is set to the X500Principal object of the SSL peer. This happens in the DefaultUserTokenHandler class. This PR overrides while creating the HttpClient class so that it will reuse the connections.
```
Why should we share connections among different SSL principals??

internal/venice-client-common/src/main/java/com/linkedin/venice/httpclient5/HttpClient5Utils.java
gaojieliu: Yeah, at least that is the behavior of httpasyncclient.

majisourav99: No its not enough I verified that, the main issue as I said is `However when the connection is put back into the pool the state is set to the X500Principal object of the SSL peer.`.  we need to reset the token

majisourav99:
> ```
> When a connection lease attempt is made the state requested is null. However when the connection is put back into the pool the state is set to the X500Principal object of the SSL peer. This happens in the DefaultUserTokenHandler class. This PR overrides while creating the HttpClient class so that it will reuse the connections.
> ```
> 
> Why should we share connections among different SSL principals??



> ```
> When a connection lease attempt is made the state requested is null. However when the connection is put back into the pool the state is set to the X500Principal object of the SSL peer. This happens in the DefaultUserTokenHandler class. This PR overrides while creating the HttpClient class so that it will reuse the connections.
> ```
> 
> Why should we share connections among different SSL principals??

We dont, thats why it cannot reuse the same connection unless we clear the state.

internal/venice-client-common/src/main/java/com/linkedin/venice/httpclient5/HttpClient5Utils.java
gaojieliu: For the same SSL peer, the `X500Principal` should be same, right?
Or will each request generate a new `X500Principal` every time? Doesn't sounds reasonable.

gaojieliu:
> > ```
> > When a connection lease attempt is made the state requested is null. However when the connection is put back into the pool the state is set to the X500Principal object of the SSL peer. This happens in the DefaultUserTokenHandler class. This PR overrides while creating the HttpClient class so that it will reuse the connections.
> > ```
> > 
> > 
> > Why should we share connections among different SSL principals??
> 
> > ```
> > When a connection lease attempt is made the state requested is null. However when the connection is put back into the pool the state is set to the X500Principal object of the SSL peer. This happens in the DefaultUserTokenHandler class. This PR overrides while creating the HttpClient class so that it will reuse the connections.
> > ```
> > 
> > 
> > Why should we share connections among different SSL principals??
> 
> We dont, thats why it cannot reuse the same connection unless we clear the state.

gaojieliu:
Having one more question:
SSL principal is per conn from my understanding, right?
But from what you described, it is unique per request.
Can you elaborate?

majisourav99:
> Having one more question: SSL principal is per conn from my understanding, right? But from what you described, it is unique per request. Can you elaborate?

For conn pool to reuse the same conn for next request it must be in previous state, but the SSL principal is set on the object stopping it from reusing the same conn.

gaojieliu (approved):

PR #693, opened 2023-10-12
[controller] Merge field-level props when generating superset schema

gaojieliu:
Merge field-level props when generating superset schema

## How was this PR tested?
CI

### DIFF
--- a/internal/venice-client-common/src/main/java/com/linkedin/venice/utils/AvroSupersetSchemaUtils.java
+++ b/internal/venice-client-common/src/main/java/com/linkedin/venice/utils/AvroSupersetSchemaUtils.java
@@ -107,18 +107,22 @@ public class AvroSupersetSchemaUtils {
     return Schema.createUnion(combinedSchema);
   }
 
+  private static void copyFieldProperties(FieldBuilder fieldBuilder, Schema.Field field) {
+    AvroCompatibilityHelper.getAllPropNames(field).forEach(k -> {
+      String propValue = AvroCompatibilityHelper.getFieldPropAsJsonString(field, k);
+      if (propValue != null) {
+        fieldBuilder.addProp(k, propValue);
+      }
+    });
+  }
+
   private static FieldBuilder deepCopySchemaField(Schema.Field field) {
     FieldBuilder fieldBuilder = AvroCompatibilityHelper.newField(null)
         .setName(field.name())
         .setSchema(field.schema())
         .setDoc(field.doc())
         .setOrder(field.order());
-    field.getObjectProps().forEach((k, ignored) -> {
-      String propValue = field.getProp(k);
-      if (propValue != null) {
-        fieldBuilder.addProp(k, propValue);
-      }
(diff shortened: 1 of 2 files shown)


internal/venice-client-common/src/main/java/com/linkedin/venice/utils/AvroSupersetSchemaUtils.java
```diff
@@ -107,32 +107,50 @@ private static Schema unionSchema(Schema s1, Schema s2) {
...
+
+  private static FieldBuilder deepCopySchemaField(Schema.Field field) {
+    FieldBuilder fieldBuilder = AvroCompatibilityHelper.newField(null)
+        .setName(field.name())
+        .setSchema(field.schema())
+        .setDoc(field.doc())
+        .setOrder(field.order());
+    copyFieldProperties(fieldBuilder, field);
+
+    // set default as AvroCompatibilityHelper builder might drop defaults if there is type mismatch
```
sixpluszero, line 127: Could you explain a bit about this comment? 
If type mismatch, then the field schema is invalid right? Shouldn't we throw exception instead?

internal/venice-client-common/src/test/java/com/linkedin/venice/schema/TestAvroSupersetSchemaUtils.java
```diff
@@ -411,6 +411,30 @@ public void testIsSupersetSchemaNestedRecordSchema() {
     Assert.assertFalse(AvroSupersetSchemaUtils.isSupersetSchema(schema1, schema2));
   }
 
+  @Test
+  public void testSupersetSchemaContainsMergeFieldProps() {
+    String valueSchemaStr1 = "{\n" + "  \"name\": \"TestRecord\",\n" + "  \"type\": \"record\",\n" + "  \"fields\": [\n"
```
sixpluszero, line 416: nit: Not need to happen in this PR, but I might eventually want to make a new way to create schema programmatically to make it more readable in legacy test classes. Probably built some simply value schema builder on top of the FieldBuilder.

sixpluszero: I made one for VPJ schema, but not general enough. But anyway, can evolve later.

sixpluszero (approved):
Would like to clarify on this: Is any existing use case using field level props and we are missing it already? I guess Frame is going to move from file level props to field level props, but not sure if it is why you make this change.

internal/venice-client-common/src/main/java/com/linkedin/venice/utils/AvroSupersetSchemaUtils.java
gaojieliu: Actually, I am not sure about this comment and I copied part of the code/comment from the existing function.
I guess it might be fine since f1 default may not be good as the default value for the merged schema.

PR #694, opened 2023-10-12
[server] A few WCAA performance improvements

gaojieliu:
1. Adopt fast-avro in the AAWC code path.
2. Use a schema-id based serde cache to avoid expensive schema comparison.
3. Dynamically choose the right structure for different logics, such as
    LinkedList vs ArrayList.

## How was this PR tested?
CI

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/replication/merge/MergeConflictResolver.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/replication/merge/MergeConflictResolver.java
@@ -11,7 +11,8 @@ import static com.linkedin.venice.schema.writecompute.WriteComputeSchemaConverte
 
 import com.linkedin.davinci.replication.RmdWithValueSchemaId;
 import com.linkedin.davinci.schema.merge.ValueAndRmd;
-import com.linkedin.davinci.serializer.avro.MapOrderingPreservingSerDeFactory;
+import com.linkedin.davinci.serializer.avro.MapOrderPreservingSerDeFactory;
+import com.linkedin.davinci.serializer.avro.fast.MapOrderPreservingFastSerDeFactory;
 import com.linkedin.davinci.store.record.ValueRecord;
 import com.linkedin.venice.annotation.Threadsafe;
 import com.linkedin.venice.exceptions.VeniceException;
@@ -20,7 +21,13 @@ import com.linkedin.venice.schema.SchemaEntry;
 import com.linkedin.venice.schema.rmd.RmdTimestampType;
 import com.linkedin.venice.schema.rmd.RmdUtils;
 import com.linkedin.venice.schema.writecompute.WriteComputeOperation;
+import com.linkedin.venice.serializer.AvroGenericDeserializer;
+import com.linkedin.venice.serializer.RecordDeserializer;
+import com.linkedin.venice.serializer.RecordSerializer;
 import com.linkedin.venice.utils.AvroSchemaUtils;
+import com.linkedin.venice.utils.SparseConcurrentList;
+import com.linkedin.venice.utils.collections.BiIntKeyCache;
+import com.linkedin.venice.utils.concurrent.VeniceConcurrentHashMap;
 import com.linkedin.venice.utils.lazy.Lazy;
 import java.nio.ByteBuffer;
 import java.util.ArrayList;
@@ -52,6 +59,11 @@ public class MergeConflictResolver {
   private final MergeResultValueSchemaResolver mergeResultValueSchemaResolver;
   private final RmdSerDe rmdSerde;
   private final boolean useFieldLevelTimestamp;
+  private final boolean fastAvroEnabled;
+
+  private final SparseConcurrentList<RecordSerializer<GenericRecord>> serializerIndexedByValueSchemaId;
+  private final BiIntKeyCache<RecordDeserializer<GenericRecord>> deserializerCacheForFullValue;
+  private final Map<String, RecordDeserializer<GenericRecord>> deserializerCacheForUpdateValue;
 
   MergeConflictResolver(
       StringAnnotatedStoreSchemaCache storeSchemaCache,
@@ -61,7 +73,8 @@ public class MergeConflictResolver {
       MergeByteBuffer mergeByteBuffer,
       MergeResultValueSchemaResolver mergeResultValueSchemaResolver,
       RmdSerDe rmdSerde,
-      boolean useFieldLevelTimestamp) {
+      boolean useFieldLevelTimestamp,
+      boolean fastAvroEnabled) {
     this.storeSchemaCache = Validate.notNull(storeSchemaCache);
     this.storeName = Validate.notNull(storeName);
     this.newRmdCreator = Validate.notNull(newRmdCreator);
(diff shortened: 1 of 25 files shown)


FelixGV:
I recall that one of the WC code paths has a comment explaining why LL is used here... can you search for it and make sure there is no unintended effect? If it still looks good, then let's update or delete that comment...

FelixGV:
Check here: https://github.com/search?q=repo%3Alinkedin%2Fvenice+%22linked+list%22&type=code

clients/da-vinci-client/src/main/java/com/linkedin/davinci/schema/merge/SortBasedCollectionFieldOpHandler.java
```diff
@@ -51,6 +51,12 @@ public UpdateResultStatus handlePutList(
       toPutList = Collections.emptyList();
     } else {
       toPutList = newFieldValue;
+      /**
+       * LinkedList is more efficient for the following add/remove operations.
+       */
+      if (!(toPutList instanceof LinkedList)) {
+        toPutList = new LinkedList<>((toPutList));
+      }
```
FelixGV, line 59: Is it possible to deserialize directly into a `LL` rather than copying here?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/serializer/avro/MapOrderPreservingDatumReader.java
```diff
@@ -18,15 +16,6 @@ public MapOrderPreservingDatumReader(Schema writer, Schema reader) {
     super(writer, reader);
   }
 
-  @Override
-  protected Object newArray(Object old, int size, Schema schema) {
```
FelixGV, line 22: Do we still need this class at all?

FelixGV (commented):
Thanks Gaojie. I left a few minor comments. Looking forward to merging this in!

clients/da-vinci-client/src/main/java/com/linkedin/davinci/replication/merge/MergeConflictResolver.java
```diff
@@ -582,9 +601,21 @@ private GenericRecord deserializeWriteComputeBytes(
...
-    Schema readerSchema = getWriteComputeSchema(readerValueSchemaId, updateProtocolVersion);
-    return deserializeValue(updateBytes, writerSchema, readerSchema);
+    String combinedKey = new StringBuilder().append(writerValueSchemaId)
+        .append("-")
+        .append(readerValueSchemaId)
+        .append("-")
+        .append(updateProtocolVersion)
+        .toString();
+    RecordDeserializer<GenericRecord> deserializer =
+        deserializerCacheForUpdateValue.computeIfAbsent(combinedKey, ignored -> {
```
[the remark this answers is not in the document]
gaojieliu: Yeah, I thought about it, and I was too lazy to build a `TriIntKeyCache`.. 
If the map is small, I assume the map lookup is quite fast.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/serializer/avro/MapOrderPreservingDatumReader.java
gaojieliu: This class was kept in case the fast-avro adoption has some regression. 
After some stabilization period, we can completely remove these classes.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/replication/merge/MergeConflictResolver.java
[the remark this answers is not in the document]
FelixGV: Yeah a map lookup is fast but I find it unfortunate to allocate a concatenated string on the hot path... 🤷 

`BiIntKeyCache<SparseConcurrentList<...>>` is fairly easy to pull off, right...?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/schema/merge/SortBasedCollectionFieldOpHandler.java
gaojieliu: List optimization is quite complex.
I will update the code to only copy `LL` when the collection is not in a `put_only` state to reduce unnecessary `LL` copy operations.

Today, for `list` type, there are following requirements:
1. For the list type in `value`, it requires `LL` related ops, such as `remove`.
2. For the list type in `rmd`, it requires fast lookup.
3. In fast-avro, for the primitive list, we are using an optimized structure to store them to reduce deserialization/serialization effort.

Optimizing every scenario is quite hard, and I would like to merge this PR and run another round of profiling before further List optimization because of the complexity.
What do you think?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/replication/merge/MergeConflictResolver.java
```diff
@@ -418,13 +444,13 @@ private ValueAndRmd<GenericRecord> createOldValueAndRmd(
 
   private GenericRecord createValueRecordFromByteBuffer(
       Schema readerValueSchema,
+      int readerValueSchemaID,
       int oldValueWriterSchemaID,
       ByteBuffer oldValueBytes) {
     if (oldValueBytes == null) {
       return AvroSchemaUtils.createGenericRecord(readerValueSchema);
```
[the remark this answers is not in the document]
gaojieliu: Yes, it looks quite expensive, but I didn't see this function takes much cpu resources while profiling AAWC code path, maybe this function doesn't trigger frequently.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/replication/merge/MergeConflictResolver.java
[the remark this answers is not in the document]
gaojieliu: Alright, let me give it a try.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/replication/merge/MergeConflictResolver.java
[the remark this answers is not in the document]
gaojieliu: Yes, it is needed when the old value is `null`, but it isn't been triggered frequently in prod.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/replication/merge/MergeConflictResolver.java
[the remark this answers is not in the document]
gaojieliu: Fixed

clients/da-vinci-client/src/main/java/com/linkedin/davinci/schema/merge/SortBasedCollectionFieldOpHandler.java
FelixGV: IIUC, 1 and 2 are different payloads, which can be deserialized with different deserializers, each of which we can configure to use a different type of list thanks to the recent fast-avro improvement, right?

Regarding 3, it is the `rmd` that will have a list of `long`, but the value could be a list of anything, not just primitives, right? So it should be fine to use fast-avro's default primitive lists for use case 2.

I don't mind us doing it incrementally across several PRs. I'm just trying to understand what is feasible.

FelixGV (approved):
LGTM! Thanks!

clients/da-vinci-client/src/main/java/com/linkedin/davinci/schema/merge/SortBasedCollectionFieldOpHandler.java
gaojieliu: yes, but for 1, it can contains primitive arrays, such as embeddings, and if we blindly deserialize them into LinkedList, we will lose the benefit of #3, which will save cpu and GC effort.

Let us profile it again after deploying this change to prod.

FelixGV: I see.

PR #695, opened 2023-10-13
[admin-tool][controller] Add new admin tool command to configure/remove particular view for store

sixpluszero:
## [admin-tool][controller] Add new admin tool command to configure/remove particular view for store

This PR introduces a new AdminTool command to configure (add/update/remove) a view of a store.
The command name is **--configure-store-view**. It takes the following parameter:

1. **--view-name**: Required at all time. Value: Name of the view to act on.
2. **--remove-view**: If present, this action is to remove an existing view from the store config. No Value.
3. **--view-class**: Required when it is not removing a view. Value: View class name of the view.
4. **--view-params**: Optional. Value: Additional parameters for the view in the format of string to string map. If not specified, meaning it is going to use **empty map**, instead of inheriting existing value.

The actual logic is inside update store logic. The reason to separate it into a new command is to make it more user friendly to specify the intention, instead of putting everything into a Json map. 
For the existing setView() API in UpdateStoreQueryParams, this PR **changes the behavior** from merging the input into existing view config to **completely overwrite the old config**, as it is overlapping with the functionality with the new command to some extent.
 
## How was this PR tested?
Added unit tests and integration tests.


ZacAttack (approved):
LGTM!  Thanks!!

PR #696, opened 2023-10-13
[server][da-vinci-client] Check for cancelled compaction 

majisourav99:
## Check for cancelled compaction 

During compaction if rocksDB cannot find enough disk space it just logs `Cancelled compaction because not enough room` and pause the compaction and later other issues arise from low disk space. This PR will fail the ingestion when memory limiter is enabled and compaction cancelled metric is 1. 

## How was this PR tested?

GHCI


gaojieliu:
Let us not commit such behavior change easily since compaction is mostly happening to hybrid store. 
Sometime, compaction failure can recover on its own if some other store's ingestion fails.

PR #697, opened 2023-10-13
[fc][test] FC warmup part 1: delay fc startup until metadata is updated

m-nagarajan:
## Summary
* currently fast client `start()` is a non blocking call and when metadata update fails for the first time it schedules the update and returns leading to the read requests throwing exception until the metadata becomes ready in the background. This change makes it a blocking start until the metadata is ready. In case of `InterruptedException`, it will fall back to the old ways. 
* `testFastClientWithoutServers` as an integration test doesn't make sense anymore as it will keep on waiting, so removed it. The newly added unit test mimics this (a more general version where metadata update fails for any reason). Need to add more unit tests for metadata. Not in scope for this PR.

## How was this PR tested?
Added a unit test

fast client start() is a blocking call now which waits till the initial metadata is updated.


lluwm (approved):
Thanks Manoj and it makes sense to me!

PR #698, opened 2023-10-14
[producer] Allow top-level union-schema in the shell producer

nisargthakkar:
## Allow top-level union-schema in the shell producer

In `ProducerTool`, some logic to strip top-level unions has been inherited from the Query tool. From the commit history, it seems that the logic was added for VSON stores. We do not need to continue building systems that support legacy stores at the expense of stores that use Avro. Hence, removed the special handling of top-level unions in the shell producer.

## How was this PR tested?

Tested locally on a store


majisourav99 (approved):
LGTM.

PR #700, opened 2023-10-17
[producers][server][dvc] Dedupe DIV debug info

FelixGV:
In some use cases, the debug info in servers could take a large amout of heap. This commit optimizes this by deduping the instances used in the various debug maps, since they are very repetitive, and by using a more compact map from the fastutil library, rather than a CHM.

Miscellaneous:

- Deleted the ArrayBasedPrimitiveIntegerSet, which is unused.

- Added back the test permutation number to build logs.

- Added two commits containing just big directory moves to git-blame-ignore-revs

## How was this PR tested?
CI, and one new unit test.


internal/venice-common/src/main/java/com/linkedin/venice/kafka/validation/Segment.java
```diff
@@ -39,6 +42,16 @@
  */
 @NotThreadsafe
 public class Segment {
+  /**
+   * This cache is to reduce the size on heap of debug info, which is very repetitive in nature. Using this cache, each
```
adamxchen, line 46: Do you have a few examples handy to show how the debugging info look like?
I checked a few places where `debugInfo` is written and most of places are either an empty map or null.

FelixGV: Sure, you can see the handful of properties we include in this map in the `Utils::getDebugInfo` function. It includes the pid/path/hostname/user of the producing process, the Java version and (if available on the classpath...) the Venice jar version.

FelixGV: I included all debug info properties in the unit test.

adamxchen (dismissed):
Thank you!

FelixGV:
Thanks for the ship it @adamxchen but I had a compilation failure 😓 ... just fixed it in the latest commit...

adamxchen (approved):

PR #704, opened 2023-10-18
[server][controller][vpj] Support TTL repush for A/A + partial update

sixpluszero:
## [server][controller][vpj] Support TTL repush for A/A + partial update
This PR aims to add TTL repush support for Active/Active with partial update enabled case.
TTL repush filters record by looking at the timestamp in the RMD metadata. For A/A only store which should be using value level TS, it will directly compare long value. For A/A + partial update (AAWC) store, it is supposed to be suing field level TS, it will rely on MergeRecordHelper#deleteRecord API to perform a dummy deletion to filter out stale parts (or drop the record if every field is more stale than the threshold).

This PR also did some small code cleanup and refactor of the related functionalities and fixed some JavaDoc.

## How was this PR tested?

Existing unit tests and integration tests, will also add new unit test and integration tests.


clients/venice-push-job/build.gradle
```diff
@@ -3,6 +3,7 @@ plugins {
 }
 
 dependencies {
+  implementation project(':clients:da-vinci-client')
```
adamxchen, line 6: Do you think we can move these util classes to a common package so we don't need `da-vinci-client` dependency here?

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/input/kafka/ttl/VeniceRmdTTLFilter.java
```diff
@@ -27,68 +33,119 @@
...
+        rmdDeserializerCache.computeIfAbsent(rmdVersionId, this::generateRmdDeserializer).deserialize(rmdPayload);
+    Object rmdTimestampObject = rmdRecord.get(TIMESTAMP_FIELD_POS);
+    RmdTimestampType rmdTimestampType = RmdUtils.getRmdTimestampType(rmdTimestampObject);
+    if (rmdTimestampType.equals(RmdTimestampType.VALUE_LEVEL_TIMESTAMP)) {
+      return (long) rmdTimestampObject > filterTimestamp;
+    }
+    RecordDeserializer<GenericRecord> valueDeserializer =
+        valueDeserializerCache.computeIfAbsent(valueSchemaId, this::generateValueDeserializer);
+    GenericRecord valueRecord = valueDeserializer.deserialize(getValuePayload(value));
+    UpdateResultStatus updateResultStatus = mergeRecordHelper.deleteRecord(valueRecord, rmdRecord, filterTimestamp, 0);
```
majisourav99, line 98: should we not use currentTs instead of `filterTimestamp` ?

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/input/kafka/ttl/VeniceRmdTTLFilter.java
```diff
@@ -27,68 +33,119 @@
...
+    if (updateResultStatus.equals(UpdateResultStatus.COMPLETELY_UPDATED)) {
+      // This means the record is fully stale, we should drop it.
+      return false;
```
majisourav99, line 101: the filter class `apply` method name is kinda confusing. apply returning true meaning we should filter it, ie drop it which is the reverse of whats returned here. should we update the name to something like `shouldFilter/Drop`  ?

clients/venice-push-job/build.gradle
sixpluszero: Yeah...I agree this is not good looking. @nisargthakkar also pointed out this. Let me see how much effort is needed to pull them out. And if it is too big, it might need to happen in a follow up PR.

sixpluszero: For now, I think it would be a fair amount of work to move it and make this PR looks more complicated. I think it will happen in a follow up PR. I can do the Java class shuffling there.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/input/kafka/ttl/VeniceRmdTTLFilter.java
sixpluszero: OK, probably let me update with a name.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
```diff
@@ -1062,15 +1063,13 @@ public void run() {
...
-            StringBuilder schemaDirBuilder = new StringBuilder();
-            schemaDirBuilder.append(TEMP_DIR_PREFIX)
-                .append(pushJobSetting.storeName)
-                .append("/")
-                .append(System.currentTimeMillis());
-            try (HDFSRmdSchemaSource rmdSchemaSource =
-                new HDFSRmdSchemaSource(schemaDirBuilder.toString(), pushJobSetting.storeName)) {
-              rmdSchemaSource.loadRmdSchemasOnDisk(controllerClient);
-              pushJobSetting.rmdSchemaDir = rmdSchemaSource.getPath();
+            String rmdSchemaDir = TEMP_DIR_PREFIX + pushJobSetting.storeName + "/rmd_" + System.currentTimeMillis();
```
majisourav99, line 1066: why this change?

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/input/kafka/ttl/VeniceRmdTTLFilter.java
```diff
@@ -27,68 +32,121 @@
...
+    RmdTimestampType rmdTimestampType = RmdUtils.getRmdTimestampType(rmdTimestampObject);
+    // For value-level RMD timestamp, just compare the value with the filter TS.
+    if (rmdTimestampType.equals(RmdTimestampType.VALUE_LEVEL_TIMESTAMP)) {
+      return (long) rmdTimestampObject < filterTimestamp;
+    }
+    RecordDeserializer<GenericRecord> valueDeserializer =
+        valueDeserializerCache.computeIfAbsent(valueSchemaId, this::generateValueDeserializer);
+    GenericRecord valueRecord = valueDeserializer.deserialize(getValuePayload(value));
+    UpdateResultStatus updateResultStatus =
+        mergeRecordHelper.deleteRecord(valueRecord, (GenericRecord) rmdTimestampObject, filterTimestamp - 1, 0);
```
majisourav99, line 99: may be define `filterTimestamp` as filterTimestamp -1 and update the line 93 logic as such ?

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
sixpluszero: We need to store both RMD schema and value schema, and better in a different folder to make the logic simpler.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/input/kafka/ttl/VeniceRmdTTLFilter.java
sixpluszero: OK

sixpluszero: Yeah. Fixed it.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/input/kafka/ttl/VeniceRmdTTLFilter.java
```diff
@@ -27,68 +32,121 @@
...
+    throw new UnsupportedOperationException(ttlPolicy + " policy is not supported.");
   }
 
   @Override
   public void close() {
     schemaSource.close();
   }
 
-  public long getTimeStampFromRmdRecord(final INPUT_VALUE value) {
+  boolean validateByTTLandMaybeUpdateValue(final INPUT_VALUE value) {
```
majisourav99, line 78: nit: rename to something like filterByTTLandMaybeUpdateValue

sixpluszero: Sure, fixed. 
Just to note that: the update value means updating the input value container. I think about this in the original apply() API name change and decided not to add "update value" there, as for partially stale value, the API actually "filters out" the old updates, and keeps the fresh ones.

majisourav99 (approved):
LGTM. Thanks Jialin

PR #707, opened 2023-10-19
[server] Avoid creating separate consumer services for broker alias addresses

sushantmane:
## [server] Avoid creating separate consumer services for broker alias addresses

In the current setup, we may inadvertently create a separate KafkaConsumerService for each alias address. 
This change prevents such unnecessary duplication.

## How was this PR tested?

CI

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/AggKafkaConsumerService.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/AggKafkaConsumerService.java
@@ -25,6 +25,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Properties;
 import java.util.Set;
+import java.util.function.Function;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
@@ -53,6 +54,7 @@ public class AggKafkaConsumerService extends AbstractVeniceService {
   private final Object2IntMap<String> kafkaClusterUrlToIdMap;
   private final PubSubMessageDeserializer pubSubDeserializer;
   private final TopicManagerRepository.SSLPropertiesSupplier sslPropertiesSupplier;
+  private final Function<String, String> kafkaClusterUrlResolver;
 
   public AggKafkaConsumerService(
       final PubSubConsumerAdapterFactory consumerFactory,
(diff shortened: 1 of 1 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/AggKafkaConsumerService.java
```diff
@@ -115,20 +122,21 @@ private KafkaConsumerService getKafkaConsumerService(final String kafkaURL) {
...
-    final String kafkaUrl = consumerProperties.getProperty(KAFKA_BOOTSTRAP_SERVERS);
+    String kafkaUrl = consumerProperties.getProperty(KAFKA_BOOTSTRAP_SERVERS);
     Properties properties = sslPropertiesSupplier.get(kafkaUrl).toProperties();
     consumerProperties.putAll(properties);
     if (kafkaUrl == null || kafkaUrl.isEmpty()) {
       throw new IllegalArgumentException("Kafka URL must be set in the consumer properties config. Got: " + kafkaUrl);
     }
-    final KafkaConsumerService alreadyCreatedConsumerService = kafkaServerToConsumerServiceMap.get(kafkaUrl);
+    String resolvedKafkaUrl = kafkaClusterUrlResolver == null ? kafkaUrl : kafkaClusterUrlResolver.apply(kafkaUrl);
+    final KafkaConsumerService alreadyCreatedConsumerService = getKafkaConsumerService(resolvedKafkaUrl);
```
huangminchn, line 132: Looks good overall. Just one minor concern:
Does the resolver return our preferred url? Which is the IPv6 url?
I hope our codes can be defensive to this case - let's say IPv6 url is the one that can be resolved in future, and one day the IPv4 url doesn't work anymore; I hope the system will keep the consumers that use the new url.

sushantmane: Yes resolver will resolve to a primary url (`url` field in `kafkaClusterMap`).

huangminchn: Sounds good! Let's make sure IPv6 is the primary url.

huangminchn (approved):
Thanks for the quick turnaround Sushant!

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/AggKafkaConsumerService.java
sushantmane: Ack.

sushantmane:
Thanks a lot, @huangminchn for the review!

PR #708, opened 2023-10-19
[venice-thin-client][da-vinci-client] Do not emit system store client metrics

majisourav99:
##  Do not emit system store client metrics

The venice_system stores are internal to venice architecture and clients should not be looking into them. Also for DaVinci client each store per instance emits around 400 metrics, and 120 of them are for system-stores. This PR stops emission of system store client metrics which would reduce ~1/3 metrics in clients

## How was this PR tested?

GHCI

### DIFF
--- a/clients/venice-thin-client/src/main/java/com/linkedin/venice/client/stats/BasicClientStats.java
+++ b/clients/venice-thin-client/src/main/java/com/linkedin/venice/client/stats/BasicClientStats.java
@@ -4,6 +4,7 @@ import com.linkedin.venice.client.store.ClientConfig;
 import com.linkedin.venice.read.RequestType;
 import com.linkedin.venice.stats.AbstractVeniceHttpStats;
 import com.linkedin.venice.stats.TehutiUtils;
+import io.tehuti.metrics.MeasurableStat;
 import io.tehuti.metrics.MetricsRepository;
 import io.tehuti.metrics.Sensor;
 import io.tehuti.metrics.stats.Avg;
@@ -16,6 +17,8 @@ import io.tehuti.metrics.stats.Rate;
  * This class offers very basic metrics for client, and right now, it is directly used by DaVinci.
  */
 public class BasicClientStats extends AbstractVeniceHttpStats {
+  String SYSTEM_STORE_NAME_PREFIX = "venice_system_store_";
+
   private final Sensor requestSensor;
   private final Sensor healthySensor;
   private final Sensor unhealthySensor;
@@ -24,8 +27,11 @@ public class BasicClientStats extends AbstractVeniceHttpStats {
   private final Sensor successRequestKeyCountSensor;
   private final Sensor successRequestRatioSensor;
   private final Sensor successRequestKeyRatioSensor;
+
   private final Rate requestRate = new OccurrenceRate();
   private final Rate successRequestKeyCountRate = new Rate();
+
+  private final boolean isMetaSystemStore;
 
   public static BasicClientStats getClientStats(
       MetricsRepository metricsRepository,
@@ -39,6 +45,7 @@ public class BasicClientStats extends AbstractVeniceHttpStats {
 
   protected BasicClientStats(MetricsRepository metricsRepository, String storeName, RequestType requestType) {
     super(metricsRepository, storeName, requestType);
+    this.isMetaSystemStore = storeName.startsWith(SYSTEM_STORE_NAME_PREFIX);
     requestSensor = registerSensor("request", requestRate);
     Rate healthyRequestRate = new OccurrenceRate();
     healthySensor = registerSensor("healthy_request", healthyRequestRate);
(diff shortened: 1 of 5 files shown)


clients/venice-thin-client/src/main/java/com/linkedin/venice/client/stats/BasicClientStats.java
```diff
@@ -39,6 +45,7 @@ public static BasicClientStats getClientStats(
 
   protected BasicClientStats(MetricsRepository metricsRepository, String storeName, RequestType requestType) {
     super(metricsRepository, storeName, requestType);
+    this.isMetaSystemStore = storeName.startsWith(SYSTEM_STORE_NAME_PREFIX);
     requestSensor = registerSensor("request", requestRate);
```
sixpluszero, line 49: We are still registering sensor for sys store here?

clients/venice-thin-client/src/main/java/com/linkedin/venice/client/stats/BasicClientStats.java
```diff
@@ -39,6 +45,7 @@ public static BasicClientStats getClientStats(
...
     Rate healthyRequestRate = new OccurrenceRate();
     healthySensor = registerSensor("healthy_request", healthyRequestRate);
```
sixpluszero, line 51: Same as above

clients/venice-thin-client/src/test/java/com/linkedin/venice/client/stats/BasicClientStatsTest.java
```diff
@@ -1,6 +1,6 @@
 package com.linkedin.venice.client.stats;
 
-import static org.testng.Assert.assertTrue;
+import static org.testng.Assert.*;
```
sixpluszero, line 3: Let's remove the * import.

clients/venice-thin-client/src/main/java/com/linkedin/venice/client/stats/BasicClientStats.java
```diff
@@ -56,37 +63,71 @@ protected BasicClientStats(MetricsRepository metricsRepository, String storeName
...
 
   protected final Rate getRequestRate() {
     return requestRate;
   }
 
-  protected final Rate getSuccessRequestKeyCountRate() {
-    return successRequestKeyCountRate;
+  @Override
+  protected Sensor registerSensor(String sensorName, MeasurableStat... stats) {
+    if (!isMetaSystemStore() || sensorName.equals("request")) {
```
sixpluszero, line 120: Why do we care about system store request? Is it for reading meta system store in DVC?

clients/venice-thin-client/src/main/java/com/linkedin/venice/client/stats/ClientStats.java
```diff
@@ -110,87 +106,141 @@ protected ClientStats(MetricsRepository metricsRepository, String storeName, Req
     Rate retryRequestSuccessKeyCount = new Rate();
     retryRequestSuccessKeyCountSensor =
         registerSensor("retry_request_success_key_count", retryRequestSuccessKeyCount, new Avg(), new Max());
-    retryKeySuccessRatioSensor = registerSensor(
```
sixpluszero, line 113: Is it removed by intention for all stores?

clients/venice-thin-client/src/main/java/com/linkedin/venice/client/stats/ClientStats.java
```diff
@@ -110,87 +106,141 @@ protected ClientStats(MetricsRepository metricsRepository, String storeName, Req
...
-        "retry_key_success_ratio",
-        new TehutiUtils.SimpleRatioStat(retryRequestSuccessKeyCount, getSuccessRequestKeyCountRate()));
     multiGetFallbackSensor = registerSensor("multiget_fallback", new OccurrenceRate());
   }
 
   public void recordHttpRequest(int httpStatus) {
+    if (isMetaSystemStore()) {
```
sixpluszero, line 113: I feel like this is okay... probably there might be a better way of writing this, like creating a dummy blackhole sensor/metrics repo and in some internal implementation, check whether to direct to this blackhole sensor or not.
But if you think this is going to be more complicated, I think we can defer this effort. It is just that, for every new metrics getter setter in this class, we are going to repeat this if-else check for system store, and that does not super great to me.

clients/venice-thin-client/src/main/java/com/linkedin/venice/client/stats/BasicClientStats.java
majisourav99: this code has been reverted

clients/venice-thin-client/src/main/java/com/linkedin/venice/client/stats/BasicClientStats.java
majisourav99: stale code, reverted.

clients/venice-thin-client/src/main/java/com/linkedin/venice/client/stats/ClientStats.java
majisourav99: same

sixpluszero (approved):
Thanks for the changes!

PR #710, opened 2023-10-20
[server][controller] Close VeniceWriters in MetaStoreWriter concurrently and in bounded time

sushantmane:
## Close VeniceWriters in MetaStoreWriter concurrently and within bounded time

This commit addresses the issue in the current `MetaStoreWriter::close` implementation,  
which uses `parallelStream` to close Venice writers. Since no thread pool is specified for   
running `parallelStream`, it uses threads from `ForkJoin.commonPool`. This can lead to   
issues where other parts of the system monopolize the FJ thread pool, causing a lack of   
guaranteed concurrency during the close operation. In other cases, the `MetaStoreWriter`   
itself may monopolize the FJ thread pool, preventing other system components' access to it.  

This PR addresses the issue by eliminating the dependency on the FJ common pool during  
the MetaStore shutdown. It introduces the `VeniceWriter::closeAsync` API to close Venice  
 writers asynchronously. This async operation runs in the Venice writer's dedicated elastic  
 thread pool, which ensures concurrent and predictable behavior during the close process.  

## How was this PR tested?
GHCI

### DIFF
--- a/internal/venice-common/src/main/java/com/linkedin/venice/system/store/MetaStoreWriter.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/system/store/MetaStoreWriter.java
@@ -41,6 +41,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.Semaphore;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.Consumer;
@@ -77,6 +78,7 @@ public class MetaStoreWriter implements Closeable {
 
   private final PubSubTopicRepository pubSubTopicRepository;
   private final long closeTimeoutMs;
+  private final int numOfConcurrentVwCloseOps;
 
   public MetaStoreWriter(
       TopicManager topicManager,
@@ -96,6 +98,7 @@ public class MetaStoreWriter implements Closeable {
 
     // TODO: make this configurable
     this.closeTimeoutMs = 300000; // 5 minutes
+    this.numOfConcurrentVwCloseOps = -1; // no limit
   }
 
   /**
@@ -507,8 +510,21 @@ public class MetaStoreWriter implements Closeable {
     List<CompletableFuture<VeniceResourceCloseResult>> closeFutures = new ArrayList<>(metaStoreWriterMap.size());
     List<VeniceWriter> writersToClose = new ArrayList<>(metaStoreWriterMap.values());
     metaStoreWriterMap.clear();
+
+    // permit for the VeniceWriters to be closed asynchronously
+    int permits = numOfConcurrentVwCloseOps == -1 ? writersToClose.size() : numOfConcurrentVwCloseOps;
+    Semaphore semaphore = new Semaphore(permits);
+
     for (VeniceWriter veniceWriter: writersToClose) {
-      closeFutures.add(veniceWriter.closeAsync(true));
+      try {
+        semaphore.acquire();
+      } catch (InterruptedException e) {
+        Thread.currentThread().interrupt();
+        LOGGER.warn("Interrupted while acquiring semaphore", e);
+      }
+      CompletableFuture<VeniceResourceCloseResult> closeFuture = veniceWriter.closeAsync(true);
+      closeFuture.whenComplete((result, throwable) -> semaphore.release());
+      closeFutures.add(closeFuture);
     }
 
     // wait for all the VeniceWriters to be closed with a timeout
--- a/internal/venice-common/src/main/java/com/linkedin/venice/system/store/MetaStoreWriter.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/system/store/MetaStoreWriter.java
@@ -78,27 +78,43 @@ public class MetaStoreWriter implements Closeable {
 
   private final PubSubTopicRepository pubSubTopicRepository;
   private final long closeTimeoutMs;
-  private final int numOfConcurrentVwCloseOps;
+  private int numOfConcurrentVwCloseOps;
 
   public MetaStoreWriter(
       TopicManager topicManager,
       VeniceWriterFactory writerFactory,
       HelixReadOnlyZKSharedSchemaRepository schemaRepo,
       PubSubTopicRepository pubSubTopicRepository) {
-    this.topicManager = topicManager;
-    this.writerFactory = writerFactory;
     /**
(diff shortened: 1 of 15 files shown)


internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
```diff
@@ -357,23 +378,68 @@ public VeniceWriter(
...
+        VENICE_WRITER_CLOSE_FAILED_COUNT.incrementAndGet();
+      }
+      isClosed = true;
+      logger.info("Closed VeniceWriter for topic: {} in {} ms", topicName, System.currentTimeMillis() - startTime);
     }
   }
 
+  public CompletableFuture<VeniceResourceCloseResult> closeAsync(boolean gracefulClose) {
+    return CompletableFuture.supplyAsync(() -> {
+      synchronized (closeLock) {
```
lluwm, line 409: I think maybe I don't quite understand the code here, if all futures sync on the `closeLock`, would that make the close back to sequential close ?

sushantmane: CloseLock is per VenieWriter hence there won't be contention on it during close.

internal/venice-common/src/main/java/com/linkedin/venice/system/store/MetaStoreWriter.java
```diff
@@ -494,15 +500,31 @@ private void closeVeniceWriter(String metaStoreName, VeniceWriter veniceWriter,
...
-      metaStoreWriterMap.entrySet()
-          .parallelStream()
-          .forEach(entry -> closeVeniceWriter(entry.getKey(), entry.getValue(), false));
+    long startTime = System.currentTimeMillis();
+    LOGGER.info("Closing MetaStoreWriter - numOfVeniceWriters: {}", metaStoreWriterMap.size());
+    // iterate through the map and close all the VeniceWriters
+    List<CompletableFuture<VeniceResourceCloseResult>> closeFutures = new ArrayList<>(metaStoreWriterMap.size());
+
+    for (Map.Entry<String, VeniceWriter> entry: metaStoreWriterMap.entrySet()) {
+      closeFutures.add(entry.getValue().closeAsync(true));
```
sushantmane, line 509: One drawback of this is that it will create one thread per `VW::closeAsync` and in some cases it could overwhelm system for a short period of time. Maybe we should update this to have more control over number of inflight close operations.

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
```diff
@@ -315,6 +324,16 @@ public VeniceWriter(
...
+    // Create a thread pool which can have max 2 threads.
+    // Except during VW start and close we expect it to have zero threads to avoid unnecessary resource usage.
+    this.threadPoolExecutor = new ThreadPoolExecutor(
+        2,
+        2,
+        30,
+        TimeUnit.SECONDS,
+        new LinkedBlockingQueue<>(),
+        new DaemonThreadFactory("VW-" + topicName));
+    this.threadPoolExecutor.allowCoreThreadTimeOut(true); // allow core threads to timeout
```
lluwm, line 336: Do we need to shutdown the thread pool in all flavors of `close` even for daemon threads?

internal/venice-common/src/main/java/com/linkedin/venice/system/store/MetaStoreWriter.java
lluwm: Doesn't it actually need 2 threads per `closeAsync` (or per VeniceWriter)? Yeah, it sounds like a good idea to bound the max number of threads for closing writers.

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceResourceCloseResult.java
```diff
@@ -0,0 +1,15 @@
+package com.linkedin.venice.writer;
+
+public enum VeniceResourceCloseResult {
+  SUCCESS(0), FAILURE(1), TIMEOUT(2), ALREADY_CLOSED(3);
```
lluwm, line 4: Don't find the places where we populate TIMEOUT or FAILURE.

internal/venice-common/src/main/java/com/linkedin/venice/system/store/MetaStoreWriter.java
```diff
@@ -494,15 +500,31 @@ private void closeVeniceWriter(String metaStoreName, VeniceWriter veniceWriter,
...
     }
+    // count unclosed VeniceWriters
+    int unclosedVeniceWriterCount = 0;
+    for (CompletableFuture<VeniceResourceCloseResult> future: closeFutures) {
+      if (!future.isDone()) {
+        unclosedVeniceWriterCount++;
+      }
+    }
+
+    LOGGER.info(
```
lluwm, line 525: For completed futures, it might be useful to log the stats for each `VeniceResourceCloseResult` type, if we could.

sushantmane: Yeah good point

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
sushantmane: Good point. 

Done

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceResourceCloseResult.java
sushantmane: It was one of those 'will come in handy in the future' things. I have removed them for now.

internal/venice-common/src/main/java/com/linkedin/venice/system/store/MetaStoreWriter.java
sushantmane: done

lluwm (approved):
Thanks @sushantmane! This change looks very good to me.

sixpluszero:
Just to confirm: This change not only applies to MetaStoreWriter in Server right? It should also apply to every LEADER?

sushantmane:
> Just to confirm: This change not only applies to MetaStoreWriter in Server right? It should also apply to every LEADER?

It will be applicable to all controllers and servers that create `MetaStoreWriter`

sushantmane:
Thanks a lot for the review and suggestions, @lluwm!

Issue #705, opened 2023-10-19
[Feature] Add S3 support to push job


mijicd:
### Willingness to contribute

Yes. I can contribute a fix for this bug independently.

### Feature Request Proposal

Considering the S3's popularity and the existing library support in the Hadoop ecosystem, it would be nice to add support for S3 sources to the Venice push job.

### Motivation

> #### What is the use case for this feature?

Users not using Hadoop in their systems could have another batch data source for their Venice clusters.

### What component(s) does this bug affect?

- [ ] `Controller`: This is the control-plane for Venice. Used to create/update/query stores and their metadata.
- [ ] `Router`: This is the stateless query-routing layer for serving read requests.
- [ ] `Server`: This is the component that persists all the store data.
- [X] `VenicePushJob`: This is the component that pushes derived data from Hadoop to Venice backend.
- [ ] `VenicePulsarSink`: This is a Sink connector for Apache Pulsar that pushes data from Pulsar into Venice.
- [ ] `Thin Client`: This is a stateless client users use to query Venice Router for reading store data.
- [ ] `Fast Client`: This is a stateful client users use to query Venice Server for reading store data.
- [ ] `Da Vinci Client`: This is an embedded, stateful client that materializes store data locally.
- [ ] `Samza`: This is the library users use to make nearline updates to store data.
- [ ] `Admin Tool`: This is the stand-alone client used for ad-hoc operations on Venice.

mijicd:
cc @modithah

PR #711, opened 2023-10-23
[vpj] Enable VenicePushJob to read from any File System

modithah:
## [vpj]Enable VenicePushJob to read from any File System
This PR enables the VenicePushJob to read data from any File System using the Hadoop file system abstraction.

- Bump hadoop to `2.10.2`
- Replace all FileSystem creations to get from the desired path instead of the default one
- The relevant fileSystem-specific jars should be available in the classpath
- Tested with the sample job and following hadoop-overrides and `hadoop-aws:2.10.2` and `aws-java-sdk-bundle-1.11.271.jar` provided in the classpath. Default AWS authentication chain was used. Credentials were available as environment variables.

> ```
> input.path=s3a://<PATH_TO_DATA>
> ```

- Theoretically this should enable reading data from any file system supported by hadoop

Resolves https://github.com/linkedin/venice/issues/705

## How was this PR tested?
- Existing unit and integration tests
- Tested with S3


clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/utils/HadoopUtils.java
```diff
@@ -52,6 +54,22 @@ public static void cleanUpHDFSPath(String path, boolean recursive) {
...
 
+  /**
+   * Set Hadoop configuration from Venice properties.
+   * @param conf Configuration to be updated
+   * @param props Venice properties
+   */
+  public static void setHadoopConfigurationFromProperties(Configuration conf, VeniceProperties props) {
+    for (String key: props.keySet()) {
+      String lowerCase = key.toLowerCase();
+      if (lowerCase.startsWith(HADOOP_PREFIX)) {
```
nisargthakkar, line 65: I'm not sure if we should use `hadoop-conf.` prefix for this as it is the prefix used for the main MR job. There is another MR job in VPJ that validates schemas of all files and builds the dictionaries. That MR job will not be able to read from S3.

Can we create a new one (maybe `hadoop-overrides.` or something better) and in the code either:
1. Extract these general overrides and copy them to `hadoop-conf.` and `hadoop-dict-build-conf.`; or
2. Where we extract configs with these two existing prefixes, we also parse extract from the general overrides prefix

modithah: Removed the usage. These configuration overrides are no longer needed. They should be picked up from the hadoop environment.

modithah:
We should always be getting the FileSystem from the path in all new usages. Otherwise those places won't be able to use other FileSystems. I replaced all such existing usages

FelixGV (approved):
Thanks a lot! LGTM! Let's wait until CI finishes and I think it'll be ready to merge...

PR #712, opened 2023-10-25
[server][fast-client] Reject read-compute requests on server when the store is not enabled with the config

nisargthakkar:
## Reject read-compute requests on server when the store is not enabled with the config

Currently, servers will perform read-compute on compute requests for any store. This commit adds a check on servers to reject such requests so that servers don't spend resources for stores that aren't allowed to use read-compute

## How was this PR tested?

Added UT, Integration tests. GHCI for regression testing

    - Servers will reject read-compute requests for stores that do not have read-compute enabled

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -745,7 +745,7 @@ ext.createDiffFile = { ->
   def file = Files.createTempFile(URLEncoder.encode(project.name, 'UTF-8'), '.diff').toFile()
   def diffBase = "${git.getUpstreamRemote()}/main"
   def command = [
-      'git', 'diff', diffBase, '--no-color', '--minimal', '--', '.'
+      'git', 'diff', "${diffBase}...", '--no-color', '--minimal', '--', '.'
   ]
   command.addAll(exclusionFilter)
   file.withOutputStream { out ->
--- a/clients/venice-client/src/main/java/com/linkedin/venice/fastclient/DispatchingAvroGenericStoreClient.java
+++ b/clients/venice-client/src/main/java/com/linkedin/venice/fastclient/DispatchingAvroGenericStoreClient.java
@@ -710,7 +710,13 @@ public class DispatchingAvroGenericStoreClient<K, V> extends InternalAvroStoreCl
       if (!REDUNDANT_LOGGING_FILTER.isRedundantException(COMPUTE_TRANSPORT_EXCEPTION_FILTER_MESSAGE)) {
         LOGGER.error("Exception received from transport. ExMsg: {}", exception.getMessage());
       }
-      decoder.onCompletion(Optional.of(new VeniceClientException("Exception received from transport", exception)));
+      VeniceClientException clientException;
+      if (exception instanceof VeniceClientException) {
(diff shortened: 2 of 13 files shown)


xunyin8 (dismissed):
The change looks good and is pretty simple and safe IMO. A little odd that some integration tests are failing.

nisargthakkar:
> The change looks good and is pretty simple and safe IMO. A little odd that some integration tests are failing.

I know why it fails. After the FC read-compute PR got merged, it added some compute test cases. The `setReadComputationEnabled` is set in `AbstractClientEndToEndSetup#prepareData()`. This function gets overridden in the other test classes. So, previously FC was allowed even if the store doesn't support it. It is one more point that demonstrates this change 😅

xunyin8 (dismissed):
LGTM, thanks Nisarg for fixing the server d2 restart issue in integration tests.

xunyin8 (approved):

build.gradle
```diff
@@ -745,7 +745,7 @@ ext.createDiffFile = { ->
   def file = Files.createTempFile(URLEncoder.encode(project.name, 'UTF-8'), '.diff').toFile()
   def diffBase = "${git.getUpstreamRemote()}/main"
   def command = [
-      'git', 'diff', diffBase, '--no-color', '--minimal', '--', '.'
+      'git', 'diff', "${diffBase}...", '--no-color', '--minimal', '--', '.'
```
adamxchen, line 748: What's this change, as well as change in `git.gradle` for? Is that because of git version upgrade or something?

nisargthakkar: It should make the coverage checks on our forks a bit more reliable. Currently, we take a diff between `upstream/main` and the branch being worked on. This means if `upstream/main` has PRs merged, then unless we rebase our branch, the diff will have many changes that are not touched by the code change in the branch.

The change in the patch above will only make the diff between the place where the branch branched off of till the head of the current working branch. So, it would mean only the changes that made by the current branch.

I think Github does some magic when running actions on PRs so that the diff is taken properly, but the same doesn't happen when we push the changes to our forks. So, it is possible that diff coverage generally passes in the PR, but fails on our forks

nisargthakkar: This StackOverflow has more details: https://stackoverflow.com/a/40119935/1989520

FelixGV:
Hi @nisargthakkar, I think rejecting read compute requests is different from the expected behavior... the client is able to perform client-side compute now, so there should be no rejection, but rather the server should return the full values back (as if it were a batch get) along with a response header indicating that the computations ought to be performed client-side.

The only scenario where compute requests should be rejected would be if the client does not support client-side compute (which the server should be able to determine by examining request headers). In this case, there is nothing to do but to reject, of course...

gaojieliu:
@FelixGV 
Rejecting read compute for fast-client is valid since Fast-Client doesn't support client-side compute yet.

For thin-client, Router will do all the works, so if the store doesn't support read compute, Router will convert the read compute request into batch-get request and ask client to perform client-side compute.
So in theory, this logic won't break the client-side compute for thin-client.

PR #224, opened 2023-02-13
[vpj][dvc][server][controller] Move kafka producer behind generic producer abstraction

sushantmane:
1. Introduced a generic producer abstraction, `PubSubProducerAdapter`.  All `VeniceWriter` APIs interact with the `PubSubProducerAdapter` interface only. This interface is currently implemented for Apache Kafka: `ApacheKafkaProducerAdapter`. 

2. Introduced `PubSubProducerAdapterFactory` which is passed to`VeniceWriterFactory` to create `PubSubProducerAdapter`s. 

3. To add support for writing data to a new pub-sub system we will implement 
a) `PubSubProducerAdapterFactory` - a factory to create producer instances.
b) `PubSubProducerAdapter`  - a concrete wrapper over a system-specific producer.
In addition to this implementation should convert generic inputs to system-specific inputs and system-specific results to generic results. For example, `ApacheKafkaProducerAdapter` will convert `RecordMetadata` to generic `PubSubProduceResult` and `PubSubProducerCallback` and to Kafka-specific callback.

4. Moved configs that control Kafka producer behavior to `ApacheKafkaProducerConfig`. The remaining configs, e.g. SSL configs, used to construct Kafka producer should be moved to this class in the future. 

5. Converted shared producer service into shared producer factory to adopt a consistent style and make code structure straightforward.

6. Miscellaneous
a) Code cleanup in `VeniceWriterFactory`. This class will no longer extract/interpret component-specific configs. It should either happen in `VeniceWriter` or a specific producer's config class that handles and controls the producer behavior. 
b) Separated Venice writer stats from shared producer stats. 
c) Removed unused classes `GetRequestObject` and `GetResponseObject`.

## Summary, imperative, start upper case, don't end with a period

Resolves #XXX

## How was this PR tested?

Internal CI

## Does this PR introduce any user-facing changes?

- [x] No. You can skip the rest of this section.
- [ ] Yes. Make sure to explain your proposed changes and call out the behavior change.

PR #713, opened 2023-10-25
[controller][server][vpj][samza][producer] Use lazy initialization of producer in VeniceWriter

sushantmane:
## Use lazy initialization of producer in VeniceWriter
My old PR https://github.com/linkedin/venice/pull/224 removed lazy producer init in VWFactory. 
Bringing it back.

## How was this PR tested?

CI

### DIFF
--- a/internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
@@ -53,6 +53,7 @@ import com.linkedin.venice.utils.Time;
 import com.linkedin.venice.utils.Utils;
 import com.linkedin.venice.utils.VeniceProperties;
 import com.linkedin.venice.utils.VeniceResourceCloseResult;
+import com.linkedin.venice.utils.lazy.Lazy;
 import java.nio.ByteBuffer;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
@@ -87,7 +88,7 @@ public class VeniceWriter<K, V, U> extends AbstractVeniceWriter<K, V, U> {
       new ChunkedPayloadAndManifest(null, null);
 
   // use for running async close and to fetch number of partitions with timeout from producer
-  private final ThreadPoolExecutor threadPoolExecutor;
+  private final Lazy<ThreadPoolExecutor> threadPoolExecutorLazy = Lazy.of(this::createThreadPoolExecutor);
 
   // log4j logger
   private final Logger logger;
@@ -213,7 +214,7 @@ public class VeniceWriter<K, V, U> extends AbstractVeniceWriter<K, V, U> {
   private final VeniceKafkaSerializer<K> keySerializer;
   private final VeniceKafkaSerializer<V> valueSerializer;
   private final VeniceKafkaSerializer<U> writeComputeSerializer;
-  private final PubSubProducerAdapter producerAdapter;
+  private final Lazy<PubSubProducerAdapter> producerAdapterLazy;
   private final GUID producerGUID;
   private final Time time;
   private final VenicePartitioner partitioner;
@@ -263,7 +264,7 @@ public class VeniceWriter<K, V, U> extends AbstractVeniceWriter<K, V, U> {
   private final boolean isRmdChunkingEnabled;
 
   public VeniceWriter(VeniceWriterOptions params, VeniceProperties props, PubSubProducerAdapter producerAdapter) {
-    this(params, props, producerAdapter, KafkaMessageEnvelope.SCHEMA$);
+    this(params, props, Lazy.of(() -> producerAdapter), KafkaMessageEnvelope.SCHEMA$);
   }
 
   /**
@@ -275,7 +276,7 @@ public class VeniceWriter<K, V, U> extends AbstractVeniceWriter<K, V, U> {
   public VeniceWriter(
       VeniceWriterOptions params,
       VeniceProperties props,
-      PubSubProducerAdapter producerAdapter,
+      Lazy<PubSubProducerAdapter> producerAdapterLazy,
       Schema overrideProtocolSchema) {
     super(params.getTopicName());
     this.keySerializer = params.getKeySerializer();
@@ -325,16 +326,6 @@ public class VeniceWriter<K, V, U> extends AbstractVeniceWriter<K, V, U> {
     }
     this.producerGUID = GuidUtils.getGUID(props);
     this.logger = LogManager.getLogger("VeniceWriter [" + GuidUtils.getHexFromGuid(producerGUID) + "]");
-    // Create a thread pool which can have max 2 threads.
-    // Except during VW start and close we expect it to have zero threads to avoid unnecessary resource usage.
-    this.threadPoolExecutor = new ThreadPoolExecutor(
-        2,
-        2,
-        5,
-        TimeUnit.SECONDS,
-        new LinkedBlockingQueue<>(),
-        new DaemonThreadFactory("VW-" + topicName));
-    this.threadPoolExecutor.allowCoreThreadTimeOut(true); // allow core threads to timeout
 
     this.protocolSchemaHeaders = overrideProtocolSchema == null
         ? EMPTY_MSG_HEADERS
@@ -342,16 +333,16 @@ public class VeniceWriter<K, V, U> extends AbstractVeniceWriter<K, V, U> {
             PubSubMessageDeserializer.VENICE_TRANSPORT_PROTOCOL_HEADER,
             overrideProtocolSchema.toString().getBytes(StandardCharsets.UTF_8));
     try {
-      this.producerAdapter = producerAdapter;
+      this.producerAdapterLazy = producerAdapterLazy;
       // We cache the number of partitions, as it is expected to be immutable, and the call to Kafka is expensive.
       // Also avoiding a metadata call to kafka here as the partitionsFor() call sometimes may get blocked indefinitely
       // if the kafka broker is overloaded and does not respond in timely manner.
       if (params.getPartitionCount() != null) {
         this.numberOfPartitions = params.getPartitionCount();
       } else {
-        this.numberOfPartitions =
-            CompletableFuture.supplyAsync(() -> producerAdapter.getNumberOfPartitions(topicName), threadPoolExecutor)
-                .get(30, TimeUnit.SECONDS);
+        this.numberOfPartitions = CompletableFuture
+            .supplyAsync(() -> getProducerAdapter().getNumberOfPartitions(topicName), getThreadPoolExecutor())
+            .get(30, TimeUnit.SECONDS);
       }
       if (this.numberOfPartitions <= 0) {
         throw new VeniceException("Invalid number of partitions: " + this.numberOfPartitions);
@@ -370,6 +361,28 @@ public class VeniceWriter<K, V, U> extends AbstractVeniceWriter<K, V, U> {
       logger.error("VeniceWriter cannot be constructed with the props: {}", props);
       throw new VeniceException("Error while constructing VeniceWriter for store name: " + topicName, e);
     }
+  }
+
+  private PubSubProducerAdapter getProducerAdapter() {
+    return producerAdapterLazy.get();
+  }
+
+  private ThreadPoolExecutor getThreadPoolExecutor() {
+    return threadPoolExecutorLazy.get();
+  }
+
+  // Create a thread pool which can have max 2 threads.
+  // Except during VW start and close we expect it to have zero threads to avoid unnecessary resource usage.
+  private ThreadPoolExecutor createThreadPoolExecutor() {
+    ThreadPoolExecutor threadPoolExecutor = new ThreadPoolExecutor(
+        2,
+        2,
+        5,
+        TimeUnit.SECONDS,
+        new LinkedBlockingQueue<>(),
+        new DaemonThreadFactory("VeniceWriter-" + topicName));
+    threadPoolExecutor.allowCoreThreadTimeOut(true); // allow core threads to timeout
+    return threadPoolExecutor;
   }
 
   /**
@@ -394,13 +407,13 @@ public class VeniceWriter<K, V, U> extends AbstractVeniceWriter<K, V, U> {
         // DO NOT call the {@link #PubSubProducerAdapter.close(int) version from here.}
         // For non-shared producer mode gracefulClose will flush the producer
 
-        producerAdapter.close(topicName, closeTimeOutInMs, gracefulClose);
+        getProducerAdapter().close(topicName, closeTimeOutInMs, gracefulClose);
         OPEN_VENICE_WRITER_COUNT.decrementAndGet();
       } catch (Exception e) {
         logger.warn("Swallowed an exception while trying to close the VeniceWriter for topic: {}", topicName, e);
         VENICE_WRITER_CLOSE_FAILED_COUNT.incrementAndGet();
       }
-      threadPoolExecutor.shutdown();
+      getThreadPoolExecutor().shutdown();
       isClosed = true;
       logger.info("Closed VeniceWriter for topic: {} in {} ms", topicName, System.currentTimeMillis() - startTime);
     }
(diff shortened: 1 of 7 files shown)


internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
```diff
@@ -372,6 +363,28 @@ public VeniceWriter(
...
+  private PubSubProducerAdapter getProducerAdapter() {
+    return producerAdapterLazy.get();
+  }
+
+  private ThreadPoolExecutor getThreadPoolExecutor() {
+    return threadPoolExecutorLazy.get();
+  }
+
+  // Create a thread pool which can have max 2 threads.
+  // Except during VW start and close we expect it to have zero threads to avoid unnecessary resource usage.
```
sixpluszero, line 375: What does this mean? Do you mean when VW is really in use it will have at max 2 threads? Otherwise it is 0 thread because of lazy init?

sixpluszero (dismissed):
LGTM except above clarification.

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
```diff
@@ -394,13 +407,13 @@ public void close(boolean gracefulClose) {
...
 
-        producerAdapter.close(topicName, closeTimeOutInMs, gracefulClose);
+        getProducerAdapter().close(topicName, closeTimeOutInMs, gracefulClose);
         OPEN_VENICE_WRITER_COUNT.decrementAndGet();
       } catch (Exception e) {
         logger.warn("Swallowed an exception while trying to close the VeniceWriter for topic: {}", topicName, e);
         VENICE_WRITER_CLOSE_FAILED_COUNT.incrementAndGet();
       }
-      threadPoolExecutor.shutdown();
+      getThreadPoolExecutor().shutdown();
```
lluwm, line 416: This might need a bit improvement, `shutdown` only makes sure no new tasks will be accepted. Take a look at our conventions for close a thread pool, e.g., `VeniceParentHelixAdmin.close` or somewhere else.

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
```diff
@@ -429,36 +442,31 @@ public CompletableFuture<VeniceResourceCloseResult> closeAsync(boolean gracefulC
...
           }
-          producerAdapter.close(topicName, closeTimeOutInMs, gracefulClose);
+          getProducerAdapter().close(topicName, closeTimeOutInMs, gracefulClose);
           OPEN_VENICE_WRITER_COUNT.decrementAndGet();
         } catch (Exception e) {
           logger.warn("Swallowed an exception while trying to close the VeniceWriter for topic: {}", topicName, e);
           VENICE_WRITER_CLOSE_FAILED_COUNT.incrementAndGet();
         }
-        threadPoolExecutor.shutdown();
+        getThreadPoolExecutor().shutdown();
```
lluwm, line 451: Ditto.

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
```diff
@@ -87,7 +88,7 @@ public class VeniceWriter<K, V, U> extends AbstractVeniceWriter<K, V, U> {
       new ChunkedPayloadAndManifest(null, null);
 
   // use for running async close and to fetch number of partitions with timeout from producer
-  private final ThreadPoolExecutor threadPoolExecutor;
+  private final Lazy<ThreadPoolExecutor> threadPoolExecutorLazy = Lazy.of(this::createThreadPoolExecutor);
```
FelixGV, line 91: Should we make this static? Do we really want one thread pool per VW? We can have a lot of VW per server...

sixpluszero: But with single pool, you won't be able to limit the thread count per VW? Or is it still possible?

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
sushantmane: Yeah that's right. Unless needed there will be 0 threads. And these aren't created until we really need them.

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
FelixGV: The limit would need to be higher, but do we really need some per-VW threads? If we have 1K VWs in a server, do we really want 2K threads spread across 1K TPs? Would it be enough to have a single static TP with, e.g., max 100 threads, and other tasks get queued if we tap out all threads?

sixpluszero: Hmm, yeah that's probably more reasonable. 
I thought the 2 per VW is required for XInfra purpose. Maybe @sushantmane knows better about this...

sushantmane: Originally I was thinking about having just one ThreadPool for all VWs.  But I wanted let callers's have some control over the concurrency of close operations which is where this thread pool is primarily used.

sushantmane: >I thought the 2 per VW is required for XInfra purpose. Maybe @sushantmane knows better about this..
Threads here perform close operation (end all segments + close producer) which is blocking call.

sushantmane: We could potentially make both endAllSegments and producer::close async and use static pool for VWs.

sushantmane: Will do some experiments with the above changes.

FelixGV: Maybe another way is to set the TP to have a `corePoolSize` of zero. Then with the `keepAliveTime` of 5 seconds, those threads will die out quickly after being used.

sushantmane: Good point. Setting `this.threadPoolExecutor.allowCoreThreadTimeOut(true);` achieves the same effect which is done inside `VW::createThreadPoolExecutor` method.

FelixGV: Ah I see... so I guess it could be fine then...

sushantmane: @FelixGV  - Should we merge this PR then?

sushantmane: I'm changing the scope of this PR slightly:
1) Keep Per ThreadPoolExecutor for now
2) Keep lazy init of PubSubProducer and ThreadPool
3) Close segments concurrently (VW::endAllSegments)

Do let me know if there are other things that I should address. Thanks.

nisargthakkar:
Since this touches VW, can you update the components affected to include everything that uses VW? (or use all) since probably only routers, tc and fc are excluded

controller, server, dvc, vpj, samza, producer, ...

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
sushantmane: Do you mean use shutdownNow to interrupt the running threads?

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
sushantmane: Update PR with the above changes.

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriterFactory.java
```diff
@@ -41,6 +43,7 @@ public <K, V, U> VeniceWriter<K, V, U> createVeniceWriter(VeniceWriterOptions op
     return new VeniceWriter<>(
         options,
         props,
-        producerAdapterFactory.create(props, options.getTopicName(), options.getBrokerAddress()));
+        Lazy.of(() -> producerAdapterFactory.create(props, options.getTopicName(), options.getBrokerAddress())),
```
FelixGV, line 46: I'm curious about this... have you found specific code paths that were instantiating a `VW` but then not actually using it to produce anything?

In many places, we are already wrapping the `VW` into `Lazy`, or else use other patterns of lazy init (e.g. `computeIfAbsent`). So now we'll be doing double-lazy wrapping in many places. In theory it's fine but it seems a bit dirty to me... maybe the clean thing to do is to remove the "outer" lazy wrapping and just keep this inner one, or perhaps to make the remaining unwrapped `VW` into `Lazy<VW>`. I don't have a strong opinion either way but I feel like it's something we should consider.

Anyway, I don't mean to say you should embark on a big refactoring in this PR, but I am interested in hearing your opinion on the above... I imagine you looked at this code more recently than I did if you're doing this, so maybe you caught some interesting context IDK about along the way.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/MetaSystemStoreTest.java
```diff
@@ -1,8 +1,20 @@
 package com.linkedin.venice.endToEnd;
 
-import static com.linkedin.venice.ConfigKeys.*;
-import static com.linkedin.venice.system.store.MetaStoreWriter.*;
-import static org.testng.Assert.*;
```
FelixGV, line 5: Thanks.

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
```diff
@@ -372,6 +364,28 @@ public VeniceWriter(
     }
   }
 
+  private PubSubProducerAdapter getProducerAdapter() {
+    return producerAdapterLazy.get();
+  }
+
+  private ThreadPoolExecutor getThreadPoolExecutor() {
+    return threadPoolExecutorLazy.get();
+  }
```
FelixGV, line 373: I think personally I would prefer not hiding usages of `Lazy` in this way. I prefer letting the `producerAdapterLazy.get()` be present at the call site, as it gives maintainers a nudge to think about whether they really want to initialize the `Lazy` instance. In some cases, it may be desirable to only use the lazy if it's already initialized, by calling `ifPresent(instance -> { ... })` on it.

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
```diff
@@ -386,79 +400,56 @@ public void close(boolean gracefulClose) {
...
-
-        producerAdapter.close(topicName, closeTimeOutInMs, gracefulClose);
+        getProducerAdapter().close(topicName, closeTimeOutInMs, gracefulClose);
         OPEN_VENICE_WRITER_COUNT.decrementAndGet();
       } catch (Exception e) {
         logger.warn("Swallowed an exception while trying to close the VeniceWriter for topic: {}", topicName, e);
         VENICE_WRITER_CLOSE_FAILED_COUNT.incrementAndGet();
       }
-      threadPoolExecutor.shutdown();
+      getThreadPoolExecutor().shutdown();
```
FelixGV, line 417: Wrap inside `ifPresent` rather than doing a late initialization just for the sake of tearing it down right after...

FelixGV (commented):
Left some minor comments. Thanks for working on this.

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
FelixGV: Thanks, I left another review.

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
```diff
@@ -1824,37 +1837,52 @@ private Segment startSegment(int partition) {
     }
   }
 
-  private void endAllSegments(boolean finalSegment) {
-    Segment segment;
+  private CompletableFuture<Void> endAllSegments() {
```
nisargthakkar, line 1840: The `finalSegment` always being `true`, is actually a bug. It should only be `true` when VW is closed. Can we keep the old interface? Fixing the bug may be out of scope, so, feel free to skip it (but it is low touch if you want to do it any way).

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriterFactory.java
sushantmane: You're right. I just double-checked the code, and, except for two place (and none of them are in server startup path), we're indeed initializing `VW` only when we need it. One trigger to add lazy initialization for VW's producer was based on hazy log reading, where I saw producers being one after another during startup . I should have properly checked the code before adding this code.

Will update the PR to remove lazy init. Thanks!

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
sushantmane: That makes sense. Will update.

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
```diff
@@ -1512,15 +1510,18 @@ private void sendStartOfSegment(int partition, Map<String, String> debugInfo) {
...
+      boolean finalSegment) {
     ControlMessage controlMessage = new ControlMessage();
     controlMessage.controlMessageType = ControlMessageType.END_OF_SEGMENT.getValue();
     EndOfSegment endOfSegment = new EndOfSegment();
     endOfSegment.checksumValue = ByteBuffer.wrap(segments[partition].getFinalCheckSum());
     endOfSegment.computedAggregates = new ArrayList<>(); // TODO Add extra aggregates
     endOfSegment.finalSegment = finalSegment;
     controlMessage.controlMessageUnion = endOfSegment;
-    sendControlMessage(controlMessage, partition, debugInfo, null, DEFAULT_LEADER_METADATA_WRAPPER);
+    return sendControlMessageAsync(controlMessage, partition, debugInfo, null, DEFAULT_LEADER_METADATA_WRAPPER);
```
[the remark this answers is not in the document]
sushantmane: Yes, it will throw an exception when the record enqueue fails within the defined timeout window.

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
sushantmane: Sure will add it back and fix the bug. Thanks

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
sushantmane: Good catch! Thanks

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
sushantmane: Done

sushantmane:
I'm gonna put it in draft mode because I want to take some time to think it over and make the PR better.

sushantmane:
> I'm gonna put it in draft mode because I want to take some time to think it over and make the PR better.

The pull request at https://github.com/linkedin/venice/pull/720 has substantially reduced the number of MetaStore VW's per host. I'm considering updating the PR based on Felix's suggestion to use a single thread pool for all VW's instead of a thread pool per VW. 

I would also like consider sending CMs concurrently.

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
lluwm: yes, shutdownnow + awaitTermination for a limited timeout.  does it make any sense?

sushantmane:
Closing this PR for now. I would like to complete it at some point..

PR #714, opened 2023-10-25
[controller][test] Fix race condition during controller startup and fix flakies

sushantmane:
## Fix race condition during controller startup and fix flakies/reduce noise

- Setup admin topic for a cluster before running cluster initialization routines
- Close cluster wrapper at the end of test in `PartialUpdateClusterConfigTest`
- Close local broker created during `TestActiveActiveIngestion::testAAIngestionWithStoreView`
- Use `volatile` `PubSubProducerAdapter::forceClosed` flag to make update visible to a callback thread
- Use unique data paths for venice-server-data in `AbstractClientEndToEndSetup` to avoid race condition

## How was this PR tested?

Internal CI and GHCI

### DIFF
--- a/internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestActiveActiveIngestion.java
+++ b/internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestActiveActiveIngestion.java
@@ -103,6 +103,7 @@ import org.testng.Assert;
 import org.testng.annotations.AfterClass;
 import org.testng.annotations.AfterMethod;
 import org.testng.annotations.BeforeClass;
+import org.testng.annotations.BeforeMethod;
 import org.testng.annotations.Test;
 
 
@@ -119,7 +120,7 @@ public class TestActiveActiveIngestion {
   private VeniceServerWrapper serverWrapper;
   private AvroSerializer serializer;
   private ControllerClient parentControllerClient;
-  private List<Closeable> additionalCloseablesPerMethod = new ArrayList<>();
+  private List<Closeable> additionalCloseablesPerMethod;
 
   @BeforeClass(alwaysRun = true)
   public void setUp() {
@@ -164,6 +165,11 @@ public class TestActiveActiveIngestion {
   public void cleanUp() {
     multiRegionMultiClusterWrapper.close();
     TestView.resetCounters();
+  }
+
+  @BeforeMethod(alwaysRun = true)
+  public void setUpBeforeMethod() {
+    additionalCloseablesPerMethod = new ArrayList<>();
   }
 
   @AfterMethod(alwaysRun = true)
--- a/internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestActiveActiveIngestion.java
+++ b/internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestActiveActiveIngestion.java
@@ -77,11 +77,9 @@ import com.linkedin.venice.writer.VeniceWriter;
 import com.linkedin.venice.writer.VeniceWriterFactory;
 import com.linkedin.venice.writer.VeniceWriterOptions;
 import io.tehuti.metrics.MetricsRepository;
-import java.io.Closeable;
 import java.io.File;
 import java.time.Instant;
 import java.time.temporal.ChronoUnit;
-import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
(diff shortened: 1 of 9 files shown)


FelixGV (dismissed):

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -2894,7 +2894,7 @@ public String getRealTimeTopic(String clusterName, String storeName) {
         if (store == null) {
           throwStoreDoesNotExist(clusterName, storeName);
         }
-        if (!store.isHybrid() && !store.isWriteComputationEnabled()) {
+        if (!store.isSystemStore() && !store.isHybrid() && !store.isWriteComputationEnabled()) {
           logAndThrow("Store " + storeName + " is not hybrid, refusing to return a realtime topic");
```
FelixGV, line 2898: What is the implication of adding this condition? System stores are allowed to be not fully initialized...?

sushantmane: Yeah. Let me rethink about this.

sushantmane: I think  I'll address meta store and push status store creation issues in a follow up PR as it might take some time and refactoring.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestActiveActiveIngestion.java
```diff
@@ -628,6 +644,7 @@ public void testAAIngestionWithStoreView() throws Exception {
             .setMockTime(testMockTime)
             .setRegionName("local-pubsub")
             .build());
+    additionalCloseablesPerMethod.add(localKafka);
```
nisargthakkar, line 647: Can we close it at the end of the test on our own? Or using try-with-resource would be better. Using a class variable will mean we need to run this single threaded - or we could run into other race conditions.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestChangeCaptureIngestion.java
```diff
@@ -201,6 +217,7 @@ public void testAAIngestionWithStoreView(CompressionStrategy compressionStrategy
             .setMockTime(testMockTime)
             .setRegionName(STANDALONE_REGION_NAME)
             .build());
+    additionalCloseablesPerMethod.add(localKafka);
```
nisargthakkar, line 220: Same as prev

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestActiveActiveIngestion.java
sushantmane: I think this is not a problem because we do not run tests from the single class concurrently  and hence each test run will have its own `additionalCloseablesPerMethod` (`@BeforeMethod).

nisargthakkar: But why do we want to leave this up to the test framework? Using try-with-resource seems like a much cleaner option to me. I also don't see the benefit of this just yet. Maybe if we were creating resources in different places than where we use them, this kind of a pattern would be helpful.

sushantmane: I've really silly argument to defend this. My hestistation to use try-wtih-resource in this case is because it will put rest of the test in the try block and somewhat affects readability.

nisargthakkar: Ugh, yeah. These are culture clashes 😂. Personally, I've gotten used to the try-with-resource and actually, like it quite a bit because you don't need to worry about closing resources when you no longer need them. So, it kind of also nudges us to keep these blocks smaller.

But if you have strong concerns, I can drop this.

sushantmane: Well I gave in already! 😄

nisargthakkar (approved):
LGTM

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestActiveActiveIngestion.java
nisargthakkar: 😂 Thanks

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestChangeCaptureIngestion.java
sushantmane: done

sushantmane:
Thanks for the review, @FelixGV & @nisargthakkar!

PR #718, opened 2023-10-30
[da-vinci] Prints more informative details in UpdateBuilder using unresolvedDatum object

### LINKED ISSUE #661: [Feature] Print more informative details in UpdateBuilder
### Willingness to contribute

### Feature Request Proposal

### Motivation

### Details

_No response_

elijahgrimaldi:
Resolved #661 

Catch UnresolvedUnionException in UpdateBuilder and print extracted unresolvedDatum object.

### DIFF
--- a/internal/venice-common/src/main/java/com/linkedin/venice/writer/update/UpdateBuilderImpl.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/writer/update/UpdateBuilderImpl.java
@@ -15,6 +15,7 @@ import java.util.Set;
 import org.apache.avro.Schema;
 import org.apache.avro.generic.GenericData;
 import org.apache.avro.generic.GenericRecord;
+import org.apache.avro.UnresolvedUnionException;
 import org.apache.commons.lang3.Validate;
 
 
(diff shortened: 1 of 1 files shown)


elijahgrimaldi:
I am not sure what the correct logging procedure is can you let me know if this is sufficient?

nisargthakkar:
Hi @elijahgrimaldi. Thanks for the contribution and welcome to Venice land.

For logging, we use Log4j in most places (except for CLI tools where we rely on `System.out.println`). You can search the codebase for examples in the format:
```
private static final Logger LOGGER = LogManager.getLogger(...);
```

Regarding this specific issue, it is tricky to get right because currently, Venice supports Avro versions 1.4+. The `UnresolvedUnionException` was only added in Avro 1.5, and before that, Avro threw `AvroRuntimeException`. So, to get this right, we need to do some if/else check on the actual exception to handle both cases

elijahgrimaldi:
> Hi @elijahgrimaldi. Thanks for the contribution and welcome to Venice land.
> 
> For logging, we use Log4j in most places (except for CLI tools where we rely on `System.out.println`). You can search the codebase for examples in the format:
> 
> ```
> private static final Logger LOGGER = LogManager.getLogger(...);
> ```
> 
> Regarding this specific issue, it is tricky to get right because currently, Venice supports Avro versions 1.4+. The `UnresolvedUnionException` was only added in Avro 1.5, and before that, Avro threw `AvroRuntimeException`. So, to get this right, we need to do some if/else check on the actual exception to handle both cases

`  private Exception validateUpdateRecordIsSerializable(GenericRecord updateRecord) {
    try {
      serializer.serialize(updateRecord);
    } catch (Exception serializationException) {
      if (serializationException instanceof AvroRuntimeException || serializationException instanceof UnresolvedUnionException) {
        Object unresolvedDatum = e.getUnresolvedDatum();
        LOGGER.error("Unresolved datum encountered: {}", unresolvedDatum, e);
      }
      return serializationException;
    }
    return null;
  }`

Does this correctly handle the case where the exception is an instance of either of these for the respective version? Also wouldn't I need to import these exception classes and how can I do that dependently on the version of Avro?

PR #719, opened 2023-10-30
[controller][vpj] Revert changes to MultiSchemaResponse/SchemaRoutes that impacts controllerClient#getAllReplicationMetadataSchemas()

sixpluszero:
## [controller][vpj] Revert changes to MultiSchemaResponse/SchemaRoutes that impacts controllerClient#getAllReplicationMetadataSchemas()

As a side-effect in PR: #704 I changed the Json property in controller class SchemaRoutes and MultiSchemaResponse to try to clean up API naming to make it makes more sense - ID-RMD_VERSION_ID. This leads to breaking changes in controller, it needs controller deployment to release VPJ and it might have break the API JSON serialization/deserialization. 
To make it easy to release the new TTL repush AAWC feature, this PR reverts the part of the change, so it no longer requires controller deployment. I will try to make another attempt apart from this PR to gracefully deprecate improper naming of JSON property.

## How was this PR tested?

Existing test should pass to make sure it is not a regression.


gaojieliu (approved):
Thanks

PR #720, opened 2023-10-30
[server] Stop reporting server replica statuses into meta stores

sushantmane:
## Stop reporting server replica statuses into meta stores
We have discontinued the use of replica statuses stored in meta system stores.  
This PR removes the logic for reporting replica statuses to meta stores to reduce  
resource usage on Venice servers.

## How was this PR tested?
CI


sixpluszero (approved):

sushantmane:
Thanks @xunyin8 & @sixpluszero for the review!

PR #721, opened 2023-10-30
[da-vinci]Log more informative details in UpdateBuilder using unresolvedDatum object

### LINKED ISSUE #661: [Feature] Print more informative details in UpdateBuilder

elijahgrimaldi:
Catch UnresolvedUnionException and AvroRuntimeException in UpdateBuilder and log extracted unresolvedDatum object for different versions of Avro

Resolves #661


elijahgrimaldi:
Should I remove the check for the UnresolvedUnionException?

nisargthakkar:
Yeah, let's leave this issue for now. It is tricky to get this right

PR #722, opened 2023-10-31
[da-vinci]Log more informative details in UpdateBuilder using unresolvedDatum object

### LINKED ISSUE #661: [Feature] Print more informative details in UpdateBuilder

elijahgrimaldi:
Resolves #661 

Catch UnresolvedUnionException and AvroRuntimeException in UpdateBuilder and log extracted unresolvedDatum object for different versions of Avro

### DIFF
--- a/internal/venice-common/src/main/java/com/linkedin/venice/writer/update/UpdateBuilderImpl.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/writer/update/UpdateBuilderImpl.java
@@ -134,8 +134,8 @@ public class UpdateBuilderImpl implements UpdateBuilder {
       serializer.serialize(updateRecord);
     } catch (Exception serializationException) {
       if (serializationException instanceof AvroRuntimeException || serializationException instanceof UnresolvedUnionException) {
-        Object unresolvedDatum = e.getUnresolvedDatum();
-        LOGGER.error("Unresolved datum encountered: {}", unresolvedDatum, e);
+        Object unresolvedDatum = serializationException.getUnresolvedDatum();
+        LOGGER.error("Unresolved datum encountered: {}", unresolvedDatum, serializationException);
       }
       return serializationException;
     }
--- a/internal/venice-common/src/main/java/com/linkedin/venice/writer/update/UpdateBuilderImpl.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/writer/update/UpdateBuilderImpl.java
@@ -20,6 +20,7 @@ import org.apache.commons.lang3.Validate;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
+
 @NotThreadsafe
 @Experimental
 public class UpdateBuilderImpl implements UpdateBuilder {
@@ -29,7 +30,6 @@ public class UpdateBuilderImpl implements UpdateBuilder {
   private final Set<String> collectionMergeFieldNameSet;
 
   private static final Logger LOGGER = LogManager.getLogger(UpdateBuilderImpl.class);
-
 
   /**
    * @param updateSchema Update schema that is derived from the value Record schema.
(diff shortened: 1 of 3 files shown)


FelixGV:
Thanks for the contribution @elijahgrimaldi! Looks like the CI is failing with a compilation issue. Can you try running `./gradlew compileJava compileTestJava compileIntegrationTestJava` and seeing where the problem is? Thanks!

internal/venice-common/src/main/java/com/linkedin/venice/writer/update/UpdateBuilderImpl.java
```diff
@@ -131,6 +133,10 @@ private Exception validateUpdateRecordIsSerializable(GenericRecord updateRecord)
     try {
       serializer.serialize(updateRecord);
     } catch (Exception serializationException) {
+      if (serializationException instanceof AvroRuntimeException || serializationException instanceof UnresolvedUnionException) {
+        Object unresolvedDatum = serializationException.getUnresolvedDatum();
```
FelixGV, line 137: That is probably the compilation failure... `UnresolvedUnionException` contains the `getUnresolvedDatum()` API, but `Exception` does not. Since `serializationException` is of type `Exception`, the compilation fails.

I think we should specifically `catch (UnresolvedUnionException e)` rather than catching `Exception`, and then we could use `getUnresolvedDatum` on it directly.

elijahgrimaldi: I made some changes to just catch UnresolvedUnionException specifically, let me know if this works

nickhansel:
lgtm

internal/venice-common/src/main/java/com/linkedin/venice/writer/update/UpdateBuilderImpl.java
```diff
@@ -130,6 +135,10 @@ public GenericRecord build() {
   private Exception validateUpdateRecordIsSerializable(GenericRecord updateRecord) {
     try {
       serializer.serialize(updateRecord);
+    } catch (UnresolvedUnionException serializationException) {
+      Object unresolvedDatum = serializationException.getUnresolvedDatum();
+      LOGGER.error("Unresolved datum encountered: {}", unresolvedDatum, serializationException);
+      return serializationException;
```
FelixGV, line 141: Ok so... I re-read the context of the issue that led us to create this ticket, and I remember it better now... the problem is that although the `UnresolvedUnionException` already prints the `unresolvedDatum` in its exception message, that is not always sufficient to debug the issue. We would also want to print the type/schema of that datum. So I would suggest something along those lines instead:

```
      Object unresolvedDatum = serializationException.getUnresolvedDatum();
      String unresolvedDatumType = unresolvedDatum instanceof GenericContainer
          ? ((GenericContainer) unresolvedDatum).getSchema().toString()
          : unresolvedDatum.getClass().getSimpleName();
      return new VeniceException("The following type does not conform to any branch of the union: " + unresolvedDatumType, serializationException);
```

Since the code calling `validateUpdateRecordIsSerializable` will throw the exception, and that should eventually get printed higher up in the stack, I don't think it's necessary to print it ourselves down here. Double-printing exceptions is sometimes confusing, so I think it's better to have all the info we need within the one exception.

FelixGV (commented):
Thanks Elijah, looks like you got your workspace set up right 🚀 

I took another look at the issue and the code change itself and left another round of comment. I think we are pretty close to being able to merge this...

Thanks!

internal/venice-common/src/main/java/com/linkedin/venice/writer/update/UpdateBuilderImpl.java
elijahgrimaldi: Ah I see. I added the changes let me know if it looks good.

FelixGV:
Hi @elijahgrimaldi, the code LGTM!

Unfortunately, we are now hitting a code coverage failure with the latest change. I think it should be fairly simple to get the new code tested. Would you mind taking a look at `UpdateBuilderImplTest` and adding a new function in there to pass invalid input, and assert that it throws an exception as expected?

FelixGV (approved):
Thanks for continuing to iterate on this, and for making your first contribution to Venice 🎉🍾🚀🚢

PR #724, opened 2023-10-31
[vpj] VPJ should not close and release resources on its own

nisargthakkar:
## VPJ should not close and release resources on its own

VPJ with D2 fails to send final push job details to controller because VPJ internally closes the D2 client and the subsequent `ControllerClient` request fails. The URL based ControllerClient doesn't have this issue because the `close` method is essentially a no-op.

## How was this PR tested?

GH CI


clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
```diff
@@ -3421,10 +3417,9 @@ private void killJobAndCleanup(
         LOGGER.info("Offline push job has been killed, topic: {}", topicInfo.topic);
       }
     }
-    close();
```
nisargthakkar, line 3424: This is the crux of the fix

nisargthakkar:
Leaving in draft phase till all release blockers are resolved. This is not high priority

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
sushantmane: We should still be calling close() after sending report, right?

nisargthakkar: Calling `close` is the responsibility of the code that creates the `VenicePushJob` object. It seems weird that we would close an object during it's execution

sushantmane (approved):
LGTM! Thanks!

PR #725, opened 2023-10-31
[server] Tweak logs for better debugging

adamxchen:

During my oncall shifts, I find the logging to detect non-existing topics is a bit noisy and doesn't provide much information when we filter by ERROR. It should be WARNING.

Also, sometimes I see stats failed to register but there was no reason. Add that as well.

## How was this PR tested?

ghci.


lluwm (approved):
Thanks, LGTM!

PR #726, opened 2023-10-31
[server][da-vinci] reduced DIV warn logging frequency to reduce cpu wait especially during rewind

gaojieliu:
reduced DIV warn logging frequency to reduce cpu wait especially during rewind

## How was this PR tested?
CI

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
@@ -2777,12 +2777,16 @@ public abstract class StoreIngestionTask implements Runnable, Closeable {
 
       // TODO: remove this condition check after fixing the bug that drainer in leaders is validating RT DIV info
       if (consumerRecord.getValue().producerMetadata.messageSequenceNumber != 1) {
-        LOGGER.warn(
-            "Data integrity validation problem with: {} offset: {}, "
-                + "but consumption will continue since EOP is already received. Msg: {}",
-            consumerRecord.getTopicPartition(),
-            consumerRecord.getOffset(),
-            warningException.getMessage());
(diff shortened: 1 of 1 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -2777,12 +2777,16 @@ protected void validateMessage(
 
       // TODO: remove this condition check after fixing the bug that drainer in leaders is validating RT DIV info
       if (consumerRecord.getValue().producerMetadata.messageSequenceNumber != 1) {
-        LOGGER.warn(
-            "Data integrity validation problem with: {} offset: {}, "
-                + "but consumption will continue since EOP is already received. Msg: {}",
-            consumerRecord.getTopicPartition(),
-            consumerRecord.getOffset(),
-            warningException.getMessage());
+        String errorMsgIdentifier = consumerRecord.getTopicPartition().getPubSubTopic().getName() + "-"
```
sixpluszero, line 2780: but warnMessage should contain details of producer right? Should we ignore that?

sixpluszero: NVM. Didn't see the getClass().getName() this should be OK.

FelixGV (changes requested):
I would prefer that we do something a bit different...

Specifically for the case of the non-SOS first message, completely suppress the log (and probably stop incrementing the missing metric as well, as it is misleading...), basically, treat it as a fully ok write (assuming EOP is already gotten of course).

For all other missing data, log it everytime. No redudant filter anywhere...

gaojieliu:
This PR needs some re-work and I will take another look soon.

majisourav99 (approved):

gaojieliu:
@lluwm has already merged a fix:
https://github.com/linkedin/venice/pull/905
I will discard this one.

PR #727, opened 2023-11-01
[da-vinci][server] Alert on failed ingestions for both incomplete and completed partitions

huangminchn:
## Summary
Problem:
The "ingestion_task_errored_gauge" metric checks ingestion errors for partitions which explicitly reports ERROR to ZK. However, for COMPLETED partition, ERROR will not be reported in order to protect read path, but ingestion will stop without any alerts until we see a high lag after rolling bounce or users notice data staleness/inconsistency.

Fix:
Added a new set to track the partition IDs whose ingestion has failed and stopped. Life cycles of the partitions IDs in the set:
1. When a partition encounters exceptions, update both the partition-to-exception list and failed-partitions set;
2. When an unsubscribe action is triggered by Helix for the failed partition, remove it from both the partition-to-exception list and failed-partitions set;
3. When an unsubscribe action is triggered by internal logic, remove the partition from the partition-to-exception list (stop logging) but keep it in the failed-partitions set (keep alerting);
4. When an subscribe action happens for the failed partition, regardless of who triggers it, remove it from both trackings, since ingestion will start over again.

Resolves VENG-10930

## How was this PR tested?
Added unit test to verify that when partition-to-exception list is updated, failed-ingestion set would keep the failed partition.

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ConsumerAction.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ConsumerAction.java
@@ -26,26 +26,33 @@ public class ConsumerAction implements Comparable<ConsumerAction> {
 
   private long createTimestampInMs = System.currentTimeMillis();
 
+  private boolean isHelixTriggeredAction = true;
   private CompletableFuture<Void> future = new CompletableFuture<>();
 
-  public ConsumerAction(ConsumerActionType type, PubSubTopicPartition topicPartition, int sequenceNumber) {
-    this(type, topicPartition, sequenceNumber, null, Optional.empty());
+  public ConsumerAction(
+      ConsumerActionType type,
+      PubSubTopicPartition topicPartition,
+      int sequenceNumber,
+      boolean isHelixTriggeredAction) {
+    this(type, topicPartition, sequenceNumber, null, Optional.empty(), isHelixTriggeredAction);
   }
 
   public ConsumerAction(
       ConsumerActionType type,
       PubSubTopicPartition topicPartition,
       int sequenceNumber,
-      LeaderFollowerPartitionStateModel.LeaderSessionIdChecker checker) {
-    this(type, topicPartition, sequenceNumber, checker, Optional.empty());
+      LeaderFollowerPartitionStateModel.LeaderSessionIdChecker checker,
+      boolean isHelixTriggeredAction) {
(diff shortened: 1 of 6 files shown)


gaojieliu:
It seems the new set in `SIT` is very close to `partitionIngestionExceptionList`, so can we leverage the existing variables instead?
I assume we only want to track the completed partitions with error, right? The `PartitionExceptionInfo` contains a flag to indicate whether the errored partition is `completed` or not.

huangminchn:
> It seems the new set in `SIT` is very close to `partitionIngestionExceptionList`, so can we leverage the existing variables instead? I assume we only want to track the completed partitions with error, right? The `PartitionExceptionInfo` contains a flag to indicate whether the errored partition is `completed` or not.

Hi @gaojieliu , the problem is that after unsubscription, the partition with error would be removed from the list.

gaojieliu:
> > It seems the new set in `SIT` is very close to `partitionIngestionExceptionList`, so can we leverage the existing variables instead? I assume we only want to track the completed partitions with error, right? The `PartitionExceptionInfo` contains a flag to indicate whether the errored partition is `completed` or not.
> 
> Hi @gaojieliu , the problem is that after unsubscription, the partition with error would be removed from the list.

I see.
Shall we just update `SIT` to keep the partitiion id set  similar to `partitionIngestionExceptionList` and only skipping the clear of partition id set during unsub?
I saw you also update the `ConsumerAction` to indicate whether it is from Helix ST or not, and I wonder whether that is necessary or not.

gaojieliu (approved):
Thanks for the fix!

huangminchn:
> > > It seems the new set in `SIT` is very close to `partitionIngestionExceptionList`, so can we leverage the existing variables instead? I assume we only want to track the completed partitions with error, right? The `PartitionExceptionInfo` contains a flag to indicate whether the errored partition is `completed` or not.
> > 
> > 
> > Hi @gaojieliu , the problem is that after unsubscription, the partition with error would be removed from the list.
> 
> I see. Shall we just update `SIT` to keep the partitiion id set similar to `partitionIngestionExceptionList` and only skipping the clear of partition id set during unsub? I saw you also update the `ConsumerAction` to indicate whether it is from Helix ST or not, and I wonder whether that is necessary or not.

Synced up offline. Reason for the new flag in ConsumerAction:
If Helix sends out the ST (STANDBY to OFFLINE) to unsubsubscribe from the partition, this partition is not expected to be ONLINE anyway, so we don't need to alert on it anymore.
However, if the unsubscribe is from our internal error handling codes, we should keep alerting on it.

PR #731, opened 2023-11-02
[da-vinci] Speed up shutdown

gaojieliu:
There are two issues with DaVinci shutdown:
1. `VersionBackend#close` is calling `KafkaStoreIngestionService#shutdownStoreIngestionTask`, which
   would wait for up to 10s to complete the shutdown (best effort).
2. Each SIT will shutdown the partition sequentially.

When shutting down a DaVinci instance, the Store Versions hosted will be shut down sequentially and
if there are x number of DaVinci stores, the last store will receive the shutdown signal after 10 * x seconds,
which is too late.
If there are y number of partitions, based on the logs, each partition graceful shutdown would take roughly 2 seconds
for a hybrid store partition,
so it will take 2 * y seconds to gracefully shut down one `SIT`.
For some heavy DaVinci users, they normally use multiple DaVinci stores (5+) stores and each store would use a partition count
at least 20+, and based on the above calculation, it will take several mins to let DaVinci instance fully shutdown.

This code change will shut down all the subscribed Stores concurrently when closing DaVinciBackend and in the meantime, for DaVinci,
`SIT` will try to close the subscribed partitions concurrently as well.

TODO: we can evaluate whether this concurrent shutdown is applicable to Venice Server or not.
To me, it is not necessary mostly since normally, each Venice Server will host a big number of SIT and adding more concurrency in
partition level won't help much since the total number of CPU cores is limited.

## How was this PR tested?
CI

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
@@ -52,6 +52,7 @@ import com.linkedin.venice.service.AbstractVeniceService;
 import com.linkedin.venice.service.ICProvider;
 import com.linkedin.venice.stats.TehutiUtils;
 import com.linkedin.venice.utils.ComplementSet;
+import com.linkedin.venice.utils.DaemonThreadFactory;
 import com.linkedin.venice.utils.Utils;
 import com.linkedin.venice.utils.VeniceProperties;
 import com.linkedin.venice.utils.concurrent.VeniceConcurrentHashMap;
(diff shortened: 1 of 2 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -1314,21 +1321,43 @@ public void run() {
...
+          PartitionConsumptionState partitionConsumptionState = entry.getValue();
+          consumerUnSubscribeAllTopics(partitionConsumptionState);
+
+          if (ingestionCheckpointDuringGracefulShutdownEnabled) {
+            try {
+              waitForAllMessageToBeProcessedFromTopicPartition(
+                  new PubSubTopicPartitionImpl(versionTopic, partitionConsumptionState.getPartition()),
+                  partitionConsumptionState);
+            } catch (InterruptedException e) {
+              throw new VeniceException(e);
```
sushantmane, line 1335: Would it make sense ti catch IE and set back interrupt status and let rest of the shutdown sequence to continue?

gaojieliu: Good question.
When this function call receives an `InterruptedException`, it means the upstreaming service would like to terminate program ASAP and continuing to execute the following logic will put program stuck again, which I think is unexpected.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
```diff
@@ -423,8 +424,21 @@ public synchronized void close() {
...
+      /**
+       * {@link StoreBackend#close()} is time-consuming since the internal {@link VersionBackend#close()} call triggers
+       * {@link KafkaStoreIngestionService#shutdownStoreIngestionTask}, which can take up to 10s to return.
+       * So here we use a thread pool to shut down all the subscribed stores concurrently.
+       */
+      storeBackendCloseExecutor.submit(() -> storeBackend.close());
+    }
+    storeBackendCloseExecutor.shutdown();
+    try {
+      storeBackendCloseExecutor.awaitTermination(60, TimeUnit.SECONDS);
```
sixpluszero, line 439: If it failed to close for 60secs should we call executor.shutdownNow() to force shutdown?

gaojieliu: These are daemon threads, so it won't block JVM shutdown.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
sushantmane: Yeah right.

sushantmane (approved):
lgtm! Thanks, @gaojieliu.

PR #732, opened 2023-11-02
[server] Remove lag measurement empty RT check and its helper functions

shuhui-liu:
## Summary

Currently we have empty RT check in hybrid store lag measurement. If RT is empty, even though the calculated lag is positive due to trailing CMs or skippable data records in RT, we will mark the lag as 0.

However, two bugs are found in the empty RT check logic. Since RT is no longer empty after commit a5be207, instead of fixing the bugs, this commit deletes empty RT check and its helper functions.

## How was this PR tested?

CI


huangminchn (approved):
Thanks Shuhui!

PR #733, opened 2023-11-03
[server] Fixed union handling in AAWC

FelixGV:
There was a bug where we assumed unions always used null as the first branch.

Introduced a new SchemaUtils::unwrapOptionalUnion function which replaces several other functions from that class.

Thanks to @sixpluszero for finding that this is the root cause of an issue.

## How was this PR tested?
New existing unit tests.

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/schema/SchemaUtils.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/schema/SchemaUtils.java
@@ -5,14 +5,17 @@ import static com.linkedin.venice.schema.rmd.v1.CollectionRmdTimestamp.DELETED_E
 import static org.apache.avro.Schema.Type.ARRAY;
 import static org.apache.avro.Schema.Type.LONG;
 import static org.apache.avro.Schema.Type.MAP;
+import static org.apache.avro.Schema.Type.NULL;
 import static org.apache.avro.Schema.Type.RECORD;
 import static org.apache.avro.Schema.Type.STRING;
+import static org.apache.avro.Schema.Type.UNION;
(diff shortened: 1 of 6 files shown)


ZacAttack:
Didn't we do work previously to make sure that null was always the first branch?  I seem to recall an issue awhile back where we had to backfill on this and put guard rails in the controller....

sixpluszero:
> Didn't we do work previously to make sure that null was always the first branch? I seem to recall an issue awhile back where we had to backfill on this and put guard rails in the controller....

This must be very old schema, even though I don't know how to check that. v1-v3 seems invalid to me for the original incident store.

ZacAttack (approved):
I guess it doesn't hurt to be defensive.  I still seem to recall that null not being first breaks some avro rule.

FelixGV:
> I guess it doesn't hurt to be defensive. I still seem to recall that null not being first breaks some avro rule.

No that is not correct AFAIK.

The rule in Avro is that the type of the default value must correspond to the type of the first branch of the enum. Therefore, both of these are valid:

```
{
  "name": "field1",
  "doc": "Epoch time of when a heartbeat is generated.",
  "type": ["null", "long"],
  "default": null
},
{
  "name": "field2",
  "doc": "Epoch time of when a heartbeat is generated.",
  "type": ["long", "null"],
  "default": 0
}
```

But both of these are invalid:

```
{
  "name": "field1",
  "doc": "Epoch time of when a heartbeat is generated.",
  "type": ["null", "long"],
  "default": 0
},
{
  "name": "field2",
  "doc": "Epoch time of when a heartbeat is generated.",
  "type": ["long", "null"],
  "default": null
}
```

PR #734, opened 2023-11-03
[server][da-vinci-client] Add compaction cancelled metric, add details to node_removable 

majisourav99:
## Add compaction cancelled metric, add details to node_removabl

Added compaction cancellation metric to get more insight on rocksdb SST file size usage. Also adds more details to node removable result to indicate why a node considered removable as passing wrong ID silently tells its removable.

## How was this PR tested?

GHCI

### DIFF
--- a/services/venice-controller/src/main/java/com/linkedin/venice/controller/InstanceStatusDecider.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/controller/InstanceStatusDecider.java
@@ -112,7 +112,7 @@ public class InstanceStatusDecider {
                   resourceName,
                   clusterName,
                   result.getSecond());
-              return NodeRemovableResult.nonremoveableResult(
+              return NodeRemovableResult.nonRemovableResult(
                   resourceName,
                   NodeRemovableResult.BlockingRemoveReason.WILL_LOSE_DATA,
                   result.getSecond());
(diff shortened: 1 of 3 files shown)


services/venice-controller/src/main/java/com/linkedin/venice/controller/InstanceStatusDecider.java
```diff
@@ -67,7 +67,8 @@ static NodeRemovableResult isRemovable(
     try {
       // If instance is not alive, it's removable.
       if (!HelixUtils.isLiveInstance(clusterName, instanceId, resources.getHelixManager())) {
-        return NodeRemovableResult.removableResult();
+        return NodeRemovableResult
+            .removableResult("Instance " + instanceId + " not found, either dead or wrong instance ID");
```
sushantmane, line 71: Or we can just say it's not part of the liveInstance for this cluster anymore?

sushantmane (dismissed):
lgtm. Left a minor comment. Thanks, @majisourav99 !

services/venice-controller/src/main/java/com/linkedin/venice/controller/InstanceStatusDecider.java
majisourav99: Sure

sushantmane (approved):
Thanks!

PR #735, opened 2023-11-03
[changelog][server] Fix incoming write schema id

ZacAttack:
This is to fix a WC/schema evolution bug.  Tests to follow.  But the changecapture should contain the schema of the merge result which should be the superset schema not the write compute one.

Resolves #XXX


clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
```diff
@@ -495,8 +495,7 @@ protected void processMessageAndMaybeProduceToKafka(
       // only extension of IngestionTask which does a read from disk before applying the record. This makes the
       // following function
       // call in this context much less obtrusive, however, it implies that all views can only work for AA stores
-      int valueSchemaId =
-          rmdWithValueSchemaID != null ? rmdWithValueSchemaID.getValueSchemaId() : incomingValueSchemaId;
+      int valueSchemaId = mergeConflictResult.getValueSchemaId();
```
FelixGV, line 498: Nitpick... can we just delete this variable and put `mergeConflictResult.getValueSchemaId()` inline, below. It's used in only one place from what I can tell.

FelixGV: No wait, actually, this one need not change.

FelixGV: It is the `incomingValueSchemaId` below that needs to be replaced...

ZacAttack: Ah you are right.  The incoming is what needs to be fixed.  But now I'm dubious about valueSchemaId given the lambda

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumerImpl.java
```diff
@@ -610,18 +610,6 @@ protected <T> T bufferAndAssembleRecordChangeEvent(
                 null),
             pubSubTopicPartition,
             readerSchemaId);
-
-        // assembledRecord = chunkingAdapter.get(
```
FelixGV, line 614: Thanks for cleaning up the dead code!

clients/da-vinci-client/src/main/java/com/linkedin/davinci/storage/chunking/AbstractAvroChunkingAdapter.java
```diff
@@ -127,6 +128,36 @@ public T get(
...
 
+  public ByteBufferValueRecord<T> getWithSchemaId(
+      AbstractStorageEngine store,
+      int partition,
+      ByteBuffer key,
+      boolean isChunked,
+      T reusedValue,
+      BinaryDecoder reusedDecoder,
+      ReadResponse response,
+      int readerSchemaId,
```
FelixGV, line 139: Is there any chance of getting rid of the superfluous params in this new API?

The `readerSchemaId` one, for example, causes some lookups at the call site, which I would prefer to avoid if the param is not even used...

The passed in `response` param is also null, so we could get rid of the conditional code related to it later on in the implementation...

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/record/ByteBufferValueRecord.java
```diff
@@ -0,0 +1,27 @@
...
+ * This class encapsulates a value from venice storage accompanied by the schema
+ * id that was used to serialize the value.
+ *
+ * TODO: This class should probably be superseded by {@link com.linkedin.davinci.store.record.ValueRecord}. Unfortunately,
+ * MANY interfaces in the ingestion path rely on the Bytebuffer interface, where ValueRecord relies on ByteBuf. Until
+ * we rectify that, this is our stand in.
+ */
+public class ByteBufferValueRecord<T> {
+  T value;
+  int writerSchemaId = -1;
```
FelixGV, line 13: Can we make these properties final?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
```diff
@@ -435,7 +436,7 @@ protected void processMessageAndMaybeProduceToKafka(
     switch (msgType) {
       case PUT:
         mergeConflictResult = mergeConflictResolver.put(
-            oldValueProvider,
+            Lazy.of(() -> oldValueProvider.get().getValue()),
```
FelixGV, line 439: Is this the only call site of `MergeConflictResolver.put()` (outside of tests)? If that is the case, then could we change the signature to take in a `Lazy` containing the new type, rather than using a doubly-wrapped lazy?

FelixGV (commented):
Thanks! This looks like a great improvement! I left some minor comments that are mostly cosmetic, though I hope we could address them.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
ZacAttack: I started to do that, but it gets kind of hairy fast.  The Lazy<ByteBuffer> trickles into a number of objects in MergeResolver.  ValueAndRmd is where I started to throw my hands up because it uses setters from contexts where it's non trivial to get a version.  So.... I kinda gave up :(

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/record/ByteBufferValueRecord.java
ZacAttack: Sure can.  Let me do that.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/storage/chunking/AbstractAvroChunkingAdapter.java
ZacAttack: Good idea.

ZacAttack: Updated, readerSchemaId I guess is used. intellij led me to believe another implementation was getting used where it was getting thrown away, but it turns out it is used by the AbsctractAvroChunkingAdapter implementation (where ChunkingAdapter drops it on the floor).  Stepping through it with the debugger confirmed it.  But I did drop some other parameters from the signature.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/record/ByteBufferValueRecord.java
ZacAttack: Done, and intellij prompted me to convert the class to a record, so I did.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/storage/chunking/AbstractAvroChunkingAdapter.java
FelixGV: Hmm ok... but in our case it won't go through the code path where it's used, and we could just hardcode -1 into it rather than perform the lookup, right?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
FelixGV: Sure, np. Thanks for looking into it!

clients/da-vinci-client/src/main/java/com/linkedin/davinci/storage/chunking/AbstractAvroChunkingAdapter.java
ZacAttack: Huh.  I did just hardcode it and you are right.  And now I think I get why.  The RawBytesDecoder shouldn't be doing any translating, it just returns raw bytes (hence the name).  It just needed it because of the interface contract. So removed it.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/record/ByteBufferValueRecord.java
ZacAttack: and promptly put it back to class. Boo

FelixGV (dismissed):
LGTM, LFG!!!!! 🚀🚀🚀🚀🚀🚀🚀

FelixGV:
So... I think we need to do some code coverage for this 🤷 ...

ZacAttack:
> So... I think we need to do some code coverage for this 🤷 ...

Fixed

FelixGV (approved):
Thanks Zac!

PR #736, opened 2023-11-03
[server] Best effort to extract optional collection type for type validation before Active/Active partial update collection operation

sixpluszero:
## [server] Best effort to extract optional collection type for type validation before Active/Active partial update collection operation
This PR removes strict optional collection type check before applying collection partial update in Active/Active store.
Previously it requires [null, collection] to be the union collection schema. This PR uses the SchemaUtils.unwrapOptionalUnion() API to retrieve the collection type.


FelixGV (approved):
Nice cleanup! Thanks!

PR #741, opened 2023-11-06
[dvc][common][test] leader complete status to follower part 1: leader produce HB and SOS to local VT with new header

m-nagarajan:
## Summary
Issue: For hybrid stores, followers are marked completed once EOP in VT is received and before leaders are marked complete and when leader handover happens to one of the completed follower instances, even though the ingestion is not completed, it will be marked completed and ready to serve.

Fix:
Part 1:
The proposal is to have a new `isLeaderCompleted` Kafka header which can be added to heartbeat `SOS` messages. The value of that header would be:0 or 1. As soon as the leader partition determines that it is complete (according to whatever rule it determines completion by, usually offset lag, but could be time lag also) the leader will then immediately send a heartbeat message containing the header: `isLeaderCompleted` = 1. All future heartbeats will also contain that same header.

Implementation details of part 1:
1. Added a new `PubSubMessageHeader` in `PubSubMessageHeaders` to hold this info in HB messages sent from leader to VT. This information won't be produced to RT.
2. `PubSubMessageHeaders` will be added as a part of `PubSubMessage` during deserialization of the consumed records to be later used by the followers in part 2.
3. using `PartitionConsumptionState#isCompletionReported()` to determine whether leader is marked completed or not and piggybacking on `reportCompleted()` to initiate heartbeat message immediately once leader is marked completed.

Part 2 (TBD):
This includes followers reading the header added in part 1 and using that as an additional factor to decide whether it should be completed or not. 

## How was this PR tested?
Tested whether the new header is being received by consuming the version topic in an integration test

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/ImmutableChangeCapturePubSubMessage.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/ImmutableChangeCapturePubSubMessage.java
@@ -2,6 +2,7 @@ package com.linkedin.davinci.consumer;
 
 import com.linkedin.venice.pubsub.adapter.kafka.ApacheKafkaOffsetPosition;
 import com.linkedin.venice.pubsub.api.PubSubMessage;
+import com.linkedin.venice.pubsub.api.PubSubMessageHeaders;
 import com.linkedin.venice.pubsub.api.PubSubTopicPartition;
 import java.util.Objects;
 
@@ -68,5 +69,10 @@ public class ImmutableChangeCapturePubSubMessage<K, V> implements PubSubMessage<
   @Override
   public boolean isEndOfBootstrap() {
     return isEndOfBootstrap;
+  }
+
+  @Override
+  public PubSubMessageHeaders getPubSubMessageHeaders() {
+    return null;
   }
 }
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/ImmutableChangeCapturePubSubMessage.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/ImmutableChangeCapturePubSubMessage.java
@@ -1,10 +1,7 @@
 package com.linkedin.davinci.consumer;
 
-import static com.linkedin.venice.writer.VeniceWriter.EMPTY_MSG_HEADERS;
-
 import com.linkedin.venice.pubsub.adapter.kafka.ApacheKafkaOffsetPosition;
 import com.linkedin.venice.pubsub.api.PubSubMessage;
-import com.linkedin.venice.pubsub.api.PubSubMessageHeaders;
 import com.linkedin.venice.pubsub.api.PubSubTopicPartition;
 import java.util.Objects;
 
@@ -71,10 +68,5 @@ public class ImmutableChangeCapturePubSubMessage<K, V> implements PubSubMessage<
   @Override
   public boolean isEndOfBootstrap() {
     return isEndOfBootstrap;
-  }
-
-  @Override
-  public PubSubMessageHeaders getPubSubMessageHeaders() {
-    return EMPTY_MSG_HEADERS;
   }
 }
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/ImmutableChangeCapturePubSubMessage.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/ImmutableChangeCapturePubSubMessage.java
@@ -1,5 +1,7 @@
 package com.linkedin.davinci.consumer;
 
+import static com.linkedin.venice.writer.VeniceWriter.EMPTY_MSG_HEADERS;
+
 import com.linkedin.venice.pubsub.adapter.kafka.ApacheKafkaOffsetPosition;
 import com.linkedin.venice.pubsub.api.PubSubMessage;
 import com.linkedin.venice.pubsub.api.PubSubMessageHeaders;
@@ -73,6 +75,6 @@ public class ImmutableChangeCapturePubSubMessage<K, V> implements PubSubMessage<
 
   @Override
   public PubSubMessageHeaders getPubSubMessageHeaders() {
-    return null;
+    return EMPTY_MSG_HEADERS;
   }
 }
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
@@ -1677,7 +1677,7 @@ public class LeaderFollowerStoreIngestionTask extends StoreIngestionTask {
         if (isCurrentVersion.getAsBoolean()) {
           amplificationFactorAdapter
               .executePartitionConsumptionState(pcs.getUserPartition(), PartitionConsumptionState::lagHasCaughtUp);
-          reportCompletedAndSendHeartBeat(pcs, partition, true);
+          reportCompleted(pcs, true);
         }
       }
     }
@@ -2578,23 +2578,6 @@ public class LeaderFollowerStoreIngestionTask extends StoreIngestionTask {
   }
 
   /**
-   * Once leader is marked completed, immediately send a heart beat message to the local VT such that
-   * followers don't have to wait till the periodic heartbeat to know that the leader is completed
-   *
-   * Also update veniceWriter of this status, such that the SOS messages sent from the leader to the
-   * followers can have this status.
-   */
-  @Override
-  protected void sendInstantHeartBeatAndUpdateVeniceWriter(
-      PartitionConsumptionState partitionConsumptionState,
-      PubSubTopicPartition pubSubTopicPartition) {
-    boolean isLeaderCompleted = partitionConsumptionState.isCompletionReported();
-    veniceWriter.get().setPartitionLeaderCompletionStatus(pubSubTopicPartition.getPartitionNumber(), isLeaderCompleted);
-    veniceWriter.get()
-        .sendHeartbeat(pubSubTopicPartition, null, DEFAULT_LEADER_METADATA_WRAPPER, ADD_LEADER_COMPLETED_HEADER);
-  }
-
-  /**
    * Unsubscribe from all the topics being consumed for the partition in partitionConsumptionState
    */
   @Override
@@ -3218,5 +3201,26 @@ public class LeaderFollowerStoreIngestionTask extends StoreIngestionTask {
       }
     }
     lastSendIngestionHeartbeatTimestamp = currentTimestamp;
+  }
+
+  /**
+   * Once leader is marked completed, immediately send a heart beat message to the local VT such that
+   * followers don't have to wait till the periodic heartbeat to know that the leader is completed
+   *
+   * Also update veniceWriter of this status, such that the SOS messages sent from the leader to the
+   * followers can have this status.
+   */
+  void reportCompleted(PartitionConsumptionState partitionConsumptionState, boolean forceCompletion) {
+    super.reportCompleted(partitionConsumptionState, forceCompletion);
+    if (partitionConsumptionState.getLeaderFollowerState().equals(LeaderFollowerStateType.LEADER)) {
+      int partitionId = partitionConsumptionState.getPartition();
+      VeniceWriter _veniceWriter = veniceWriter.get();
+      _veniceWriter.setPartitionLeaderCompletionStatus(partitionId, true);
+      _veniceWriter.sendHeartbeat(
+          new PubSubTopicPartitionImpl(versionTopic, partitionId),
+          null,
+          DEFAULT_LEADER_METADATA_WRAPPER,
+          ADD_LEADER_COMPLETED_HEADER);
+    }
   }
 }
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
@@ -1676,7 +1676,7 @@ public class LeaderFollowerStoreIngestionTask extends StoreIngestionTask {
         if (isCurrentVersion.getAsBoolean()) {
           amplificationFactorAdapter
               .executePartitionConsumptionState(pcs.getUserPartition(), PartitionConsumptionState::lagHasCaughtUp);
-          statusReportAdapter.reportCompleted(pcs, true);
+          reportCompletedAndSendHeartBeat(pcs, partition, true);
         }
       }
     }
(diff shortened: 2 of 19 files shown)


internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/restart/TestRestartServerAfterDeletingSstFilesWithActiveActiveIngestion.java
```diff
@@ -347,6 +361,40 @@ public void testActiveActiveStoreWithRMDAndRestartServer(boolean deleteSSTFiles,
           ExecutionStatus.COMPLETED);
     });
 
+    // create a kafka consumer to check for the messages with leaderCompletedStatus
+    Properties properties = new Properties();
+    properties.setProperty(ConfigKeys.KAFKA_BOOTSTRAP_SERVERS, pubSubBrokerWrapper.getAddress());
+    ApacheKafkaConsumerConfig apacheKafkaConsumerConfig =
+        new ApacheKafkaConsumerConfig(new VeniceProperties(properties), "testConsumer");
+    ApacheKafkaConsumerAdapter consumer =
```
sushantmane, line 369: Let's not tie up the test to any specific pubsub client. Could you please update it to use `PubSubConsumerAdapter`? See this example here: https://github.com/linkedin/venice/blob/d07f18606d10d9aaf9802ed7ed291767c2e64080/internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/pubsub/api/consumer/PubSubConsumerAdapterTest.java#L117

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/adapter/kafka/consumer/ApacheKafkaConsumerAdapter.java
```diff
@@ -56,7 +56,7 @@ public class ApacheKafkaConsumerAdapter implements PubSubConsumerAdapter {
   private final PubSubMessageDeserializer pubSubMessageDeserializer;
   private final ApacheKafkaConsumerConfig config;
 
-  ApacheKafkaConsumerAdapter(
+  public ApacheKafkaConsumerAdapter(
```
sushantmane, line 59: Shouldn't be public. See the comment about writing test against generic PubSubConsumerAdapter

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/ImmutableChangeCapturePubSubMessage.java
```diff
@@ -69,4 +70,9 @@ public int getPayloadSize() {
   public boolean isEndOfBootstrap() {
     return isEndOfBootstrap;
   }
+
+  @Override
+  public PubSubMessageHeaders getPubSubMessageHeaders() {
+    return null;
```
sushantmane, line 76: Prefer returning empty headers over null. Maybe define a constant EMPTY_HEADERS?

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/adapter/kafka/consumer/ApacheKafkaConsumerConfig.java
```diff
@@ -43,7 +43,7 @@ public class ApacheKafkaConsumerConfig {
   private final int topicQueryRetryIntervalMs;
   private final Duration defaultApiTimeout;
 
-  ApacheKafkaConsumerConfig(VeniceProperties veniceProperties, String consumerName) {
+  public ApacheKafkaConsumerConfig(VeniceProperties veniceProperties, String consumerName) {
```
sushantmane, line 46: Remove public from here as well

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/restart/TestRestartServerAfterDeletingSstFilesWithActiveActiveIngestion.java
m-nagarajan: Thanks for the pointer, changed it.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/ImmutableChangeCapturePubSubMessage.java
m-nagarajan: Reusing already defined `EMPTY_MSG_HEADERS`

m-nagarajan (commented):
Thanks @sushantmane for the review. Addressed review comments.

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/api/PubSubMessage.java
```diff
@@ -43,4 +43,6 @@ default int getPartition() {
    * @return whether this message marks the end of bootstrap.
    */
   boolean isEndOfBootstrap();
+
+  PubSubMessageHeaders getPubSubMessageHeaders();
```
majisourav99, line 47: you can have a default implementation of returning empty headers

m-nagarajan: Added this change. Thanks.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -2075,19 +2075,47 @@ protected DelegateConsumerRecordResult delegateConsumerRecord(
...
+                      .sendHeartbeat(
+                          topicPartition,
+                          callback,
+                          leaderMetadataWrapper,
+                          true,
+                          consumerRecord.getValue().producerMetadata.messageTimestamp,
+                          partitionConsumptionState.isCompletionReported());
+                } else {
+                  // Leaders forward SOS message to local VT: add isLeaderCompleted Header
+                  produceToLocalKafka(
```
sushantmane, line 2101: Do we really want to forward SOS messages? I thought we would only tag SOS messages from the leaders writer itself.

sushantmane (changes requested):
I've left two comments. I think it would be good to have this PR reviewed by @FelixGV as well. Thanks!

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/api/PubSubMessageDeserializer.java
```diff
@@ -21,6 +21,8 @@ public class PubSubMessageDeserializer {
   private static final Logger LOGGER = LogManager.getLogger(PubSubMessageDeserializer.class);
 
   public static final String VENICE_TRANSPORT_PROTOCOL_HEADER = "vtp";
+  /** Header to denote whether the leader is completed or not */
```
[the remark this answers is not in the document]
m-nagarajan: Fixed it.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
m-nagarajan: Makes sense to me and made the changes. We can continue to iterate on this in this PR or part 2 PR based on the teams opinion as well.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -2087,7 +2087,15 @@ protected DelegateConsumerRecordResult delegateConsumerRecord(
...
                     new LeaderMetadataWrapper(consumerRecord.getOffset(), kafkaClusterId);
                 PubSubTopicPartition topicPartition = new PubSubTopicPartitionImpl(getVersionTopic(), subPartition);
-                veniceWriter.get().sendHeartbeat(topicPartition, callback, leaderMetadataWrapper);
+                // Leaders forward HB SOS message to local VT with updated isLeaderCompleted header
+                veniceWriter.get()
+                    .sendHeartbeat(
+                        topicPartition,
+                        callback,
+                        leaderMetadataWrapper,
+                        true,
```
FelixGV, line 2096: Nitpick: It might help readability to create a constant for that, `LEADER_IS_COMPLETE_HEADER = true`, or similar.

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/ImmutablePubSubMessage.java
```diff
@@ -13,19 +14,33 @@ public class ImmutablePubSubMessage<K, V> implements PubSubMessage<K, V, Long> {
   private final long timestamp;
   private final int payloadSize;
 
+  private final PubSubMessageHeaders pubSubMessageHeaders;
```
FelixGV, line 17: I am slightly peeved at having an extra pointer per message, as that will increase the memory footprint of in-flight messages (even though it will almost always be null, null still takes 4B or 8B, depending on whether the heap size makes us use 64 bit pointers or not...).

We could use a primitive boolean instead (which would get padded up to 8 bits, no matter the pointer size). Or we could use extension to have a different class which carries the extra property only when needed.

These seem like premature optimizations so I guess it's fine to leave it as is for now.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -2569,6 +2577,23 @@ public long getRegionHybridOffsetLag(int regionId) {
...
+   * followers can have this status.
+   */
+  @Override
+  protected void sendInstantHeartBeatAndUpdateVeniceWriter(
+      PartitionConsumptionState partitionConsumptionState,
+      PubSubTopicPartition pubSubTopicPartition) {
+    boolean isLeaderCompleted = partitionConsumptionState.isCompletionReported();
+    veniceWriter.get()
+        .sendHeartbeat(pubSubTopicPartition, null, DEFAULT_LEADER_METADATA_WRAPPER, true, 0, isLeaderCompleted);
+    veniceWriter.get().setPartitionLeaderCompletionStatus(pubSubTopicPartition.getPartitionNumber(), isLeaderCompleted);
```
FelixGV, line 2594: It looks weird to me to pass `isLeaderCompleted` into `sendHeartbeat` and also into `setPartitionLeaderCompletionStatus`... I think we should either:

1. Say that the `VW` needs to remember the completion state, in which case we should call `setPartitionLeaderCompletionStatus(partition, isLeaderCompleted )` first, and then `sendHeartbeat` (but without passing `isLeaderCompleted`)... OR
2. Say that the `LFSIT` is responsible for remembering, in which case it should always pass the param and `VW` should not carry that state at all (and thus we delete the `setPartitionLeaderCompletionStatus` function and the data structure inside `VW`)...

FelixGV (commented):
Thanks, LGTM overall. Left a few minor comments.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -3495,6 +3493,31 @@ private ReadyToServeCheck getDefaultReadyToServeChecker() {
...
+    statusReportAdapter.reportCompleted(partitionConsumptionState, forceCompletion);
+    LOGGER.info("{} Partition {} is ready to serve", consumerTaskId, partition);
+    if (partitionConsumptionState.getLeaderFollowerState().equals(LeaderFollowerStateType.LEADER)) {
+      sendInstantHeartBeatAndUpdateVeniceWriter(
+          partitionConsumptionState,
+          new PubSubTopicPartitionImpl(versionTopic, partition));
+    }
+  }
+
+  protected abstract void sendInstantHeartBeatAndUpdateVeniceWriter(
```
gaojieliu, line 3517: I feel the structure is not quite right..

Even this function is called `reportCompletedAndSendHeartBeat`, it is not guaranteed that the HB will be sent out since there is a `leader` state check.

Shall we just call it `reportCompleted` and let `LFSIT` override this function and internally it will decide whether it should send out a HB or not.
With this way, we don't need this function any more: `sendInstantHeartBeatAndUpdateVeniceWriter`.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -3495,6 +3493,31 @@ private ReadyToServeCheck getDefaultReadyToServeChecker() {
     };
   }
 
+  void reportCompletedAndSendHeartBeat(PartitionConsumptionState partitionConsumptionState, int partition) {
```
gaojieliu, line 3496: So `partition` is part of `PartitionConsumptionState`, why do we need to pass two params here?

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
```diff
@@ -1635,7 +1684,9 @@ public void sendControlMessage(
               callback,
               updateCheckSum,
               leaderMetadataWrapper,
-              VENICE_DEFAULT_LOGICAL_TS).get();
+              VENICE_DEFAULT_LOGICAL_TS,
+              isStartOfSegment, // addLeaderCompletedStatus for SOS
```
gaojieliu, line 1688: So we want to add this header for every SOS message besides HB messages? What is the reason behind this behavior?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
m-nagarajan: Great suggestion. cleaned it up.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
m-nagarajan: Great point. Changed it. Chose to set the state in `VW` and then call `sendHeartbeat` without passing `isLeaderCompleted` as VW would need the state to send `SOS` as well.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
m-nagarajan: Done

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/ImmutablePubSubMessage.java
m-nagarajan: That is a great point & suggestions. Let me explore this further once I have the 2nd half of this PR ready.

m-nagarajan (commented):
Thanks @gaojieliu and @FelixGV for the review. Addressed review comments. Please take another look at it.

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
gaojieliu: Regarding `all control and data messages such that the follower is not waiting for heartbeats and can complete sooner than later.`, it is not completely true since the PR will send a HB message with `completed` header right away once it is ready to serve, and HB frequency is not bad (1 min interval), and I recall the interval to create a new segment is 5 mins, so this reason doesn't hold true from my understanding.

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
```diff
@@ -1384,25 +1339,29 @@ private CompletableFuture<PubSubProduceResult> sendMessage(
...
+    if (addLeaderCompleteState) {
       // copy protocolSchemaHeaders locally and add extra header for isLeaderCompleted
       returnPubSubMessageHeaders = new PubSubMessageHeaders();
       for (PubSubMessageHeader header: pubSubMessageHeaders.toList()) {
         returnPubSubMessageHeaders.add(header);
       }
       // 1 byte holding 0/1 basd on isLeaderCompleted
       byte[] val = new byte[1];
-      val[0] = (byte) (getPartitionLeaderCompletionStatus(partitionId) ? 1 : 0);
+      val[0] = (byte) ((leaderCompleteState == LeaderCompleteState.LEADER_COMPLETED) ? 1 : 0);
```
FelixGV, line 1364: I think it would look cleaner if the numerical value was a property of the enum, and this code here can call `leaderCompleteState.getValue()` or the like.

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
m-nagarajan: with the latest diff, only HB has this header. Thanks for the discussion.

FelixGV (commented):
Looks good overall. Just a minor nitpick...

gaojieliu (commented):
Please update the commit message to describe the updated behavior that the leader complete header will only be sent in HB SOS message.

Also, I didn't see any test cases for the new logic, so is it expected? or you want to keep all the test cases in the second PR?

FelixGV (dismissed):
LGTM, but please check Gaojie's latest comments as well in case there is something to address there.

internal/venice-common/src/main/java/com/linkedin/venice/writer/LeaderCompleteState.java
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * Completion state of the leader partition.
+ */
+public enum LeaderCompleteState {
+  /**
+   * Default state for the standby to know that the feature is not supported and
+   * don't consider leader's completion status for its completion.
+   */
+  NOT_SUPPORTED,
```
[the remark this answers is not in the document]
m-nagarajan: It will be used as default in pcs for the standby replica to be able to differentiate between whether this header is seen or not (rather than having a separate flag). It will be used in 2nd diff.

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
```diff
@@ -1340,9 +1340,33 @@ private CompletableFuture<PubSubProduceResult> sendMessage(
...
+    PubSubMessageHeaders returnPubSubMessageHeaders;
+    PubSubMessageHeaders pubSubMessageHeaders =
+        producerMetadata.getSegmentNumber() == 0 && producerMetadata.getMessageSequenceNumber() == 0
+            ? protocolSchemaHeaders
+            : EMPTY_MSG_HEADERS;
+
+    if (addLeaderCompleteState) {
+      // copy protocolSchemaHeaders locally and add extra header for isLeaderCompleted
+      returnPubSubMessageHeaders = new PubSubMessageHeaders();
+      for (PubSubMessageHeader header: pubSubMessageHeaders.toList()) {
```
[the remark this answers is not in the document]
m-nagarajan: Thats true. Added a comment for future changes to create headers in advance for hot paths.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/restart/TestRestartServerAfterDeletingSstFilesWithActiveActiveIngestion.java
```diff
@@ -439,50 +511,97 @@ public void testActiveActiveStoreWithRMDAndRestartServer(boolean deleteSSTFiles,
...
+            pubSubConsumer.poll(10 * Time.MS_PER_SECOND);
+        for (Map.Entry<PubSubTopicPartition, List<PubSubMessage<KafkaKey, KafkaMessageEnvelope, Long>>> entry: messages
+            .entrySet()) {
+          List<PubSubMessage<KafkaKey, KafkaMessageEnvelope, Long>> pubSubMessages = entry.getValue();
+          for (PubSubMessage<KafkaKey, KafkaMessageEnvelope, Long> message: pubSubMessages) {
+            PubSubMessageHeaders pubSubMessageHeaders = message.getPubSubMessageHeaders();
+            for (PubSubMessageHeader header: pubSubMessageHeaders.toList()) {
+              if (header.key().equals(VENICE_LEADER_COMPLETION_STATE_HEADER)) {
+                isLeaderCompletionHeaderFound.set(true);
+                if (LeaderCompleteState.valueOf(header.value()[0]) == LEADER_COMPLETED) {
```
m-nagarajan, line 589: @gaojieliu added this check in an existing test to check whether the new header is being sent and validating the values. Will be adding more tests in the 2nd diff.

PR #742, opened 2023-11-06
[controller] Add metric for disabled partition map.

majisourav99:
##  Add metric for disabled partition map.

Since disabled metric can cause a cluster to be in undeployable state, this PR is adding an alert to indicate disabled partition per cluster. 
The proposed action would be for oncall to investigate what error caused the alert and mitigate by restarting those hosts.

### DIFF
--- a/services/venice-controller/src/main/java/com/linkedin/venice/pushmonitor/AbstractPushMonitor.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/pushmonitor/AbstractPushMonitor.java
@@ -10,6 +10,7 @@ import static com.linkedin.venice.pushmonitor.OfflinePushStatus.HELIX_RESOURCE_N
 
 import com.linkedin.venice.controller.HelixAdminClient;
 import com.linkedin.venice.controller.VeniceControllerConfig;
+import com.linkedin.venice.controller.stats.ErrorPartitionStats;
 import com.linkedin.venice.exceptions.VeniceException;
 import com.linkedin.venice.exceptions.VeniceNoStoreException;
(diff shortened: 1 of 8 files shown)


services/venice-controller/src/main/java/com/linkedin/venice/pushmonitor/AbstractPushMonitor.java
```diff
@@ -755,6 +761,7 @@ public void disableReplica(String instance, int partitionId) {
             kafkaTopic,
             Collections.singletonList(HelixUtils.getPartitionName(kafkaTopic, partitionId)));
         disabledReplicaMap.computeIfAbsent(instance, k -> new HashSet<>()).add(partitionId);
+        disabledPartitionStats.recordDisabledPartition();
```
huangminchn, line 764: Currently do we reenable any partition? If so, we need to decrement the count?

huangminchn: Talked with Sourav offline; reenable will happen after the node restarts; however, the Count metric will also be reset to 0 after node restarts, so there is no need to do decrement with the current code base.

huangminchn (approved):
Thanks a lot Sourav!

services/venice-controller/src/main/java/com/linkedin/venice/pushmonitor/AbstractPushMonitor.java
majisourav99: since we renable only on restart, it will clear the metric then.

PR #743, opened 2023-11-09
 [controller][server] TopicManager refactoring and performance improvements

sushantmane:
## TopicManager Refactoring and Performance Improvements

Currently, we have three components that work together to fetch topic metadata,
such as `endOffset` and the timestamp of the last produced message. These
components are `TopicManager`, `PartitionOffsetFetcher`, and
`CachedPubSubMetadataGetter`. We create one `TopicManager` instance per broker
address/region, which is shared by all the components in the system.
`PartitionOffsetFetcher` is a part of the topic manager. There is one instance of
`CachedPubSubMetadataGetter` per `StoreIngestionTask`, which makes use of
`TopicManager` and `PartitionOffsetFetcher` to retrieve topic metadata.

`PartitionOffsetFetcher` uses pubsub admin and consumer clients to fetch offsets on
behalf of components like `CachedPubSubMetadataGetter` and other components. Each
`PartitionOffsetFetcher` has just one instance of a consumer, which is protected
with a consumer mutex since consumers are not thread-safe. This means that
`PartitionOffsetFetcher` can handle only one request at a time. The consumer mutex
is coarse-grained, which means it holds the lock even when making calls with the
admin client, which is unnecessary since admin clients are thread-safe.

Also, when the mutex is held for topics that do not exist or have metadata fetching
issues, we perform retries while holding the mutex. This results in the mutex being
held for a relatively longer time when there is an issue with one topic, directly
causing delays for other requests waiting for the lock. We have often observed high
contention on this mutex, leading to several issues.

The `CachedPubSubMetadataGetter` fetches offsets asynchronously using
`TopicManager`/`PartitionOffsetFetcher`, and caches them in a local cache. These
asynchronous requests are initiated when an entry is either not in the cache or
when it is stale. Currently, these asynchronous requests are served by the ForkJoin
common thread pool.

When there is an issue like a non-existent topic, we see high contention on the
consumer mutex, as discussed earlier. This results in ForkJoin threads being
blocked on the consumer lock for longer periods. When all the ForkJoin common
thread pools get blocked, other components in the system that rely on them are
starved, which affects their functionality.

To enhance the resiliency of `TopicManager` in the face of failures, this PR
introduces the following high-level changes:

1) Admin operations during metadata fetch are now performed without holding any lock.
This prevents bad topics from acquiring the consumer in the first place and reduces
the chances of contention when getting the consumer.

2) A configurable fetcher consumer pool is introduced for each topic manager. This
allows us to process requests with other consumers even if one is busy.

3) We have discontinued the use of the ForkJoin common thread pool for submitting
async fetch offset requests. Instead, dedicated fetcher thread pools are used for this purpose.

### Miscellaneous
- This commit introduced the following configs:
  1. `pubsub.topic.manager.metadata.fetcher.consumer.pool.size`  - A config to specify number of PubSub consumer clients to use per topic manager (default: 2) 
  2. `pubsub.topic.manager.metadata.fetcher.thread.pool.size` -  Controls the size of per TM thread-pool used for get offset async requests (default: 2).
- The `TopicManagerRepository.SSLPropertiesSupplier` has been extracted into `PubSubPropertiesSupplier`.
- The `CachedPubSubMetadataGetter` and `PartitionOffsetFetcher` have been merged to form `TopicManagerMetadataFetcher`.
- The method `getOffsetLagFor` in `AggKafkaConsumerService` has been renamed to `getConsumerLagBasedOnMetrics`.
- The method `getPartitionOffsetLag` in `ActiveActiveStoreIngestionTask` has been renamed to `getPartitionOffsetLagBasedOnMetrics`.
- `TopicManagerContext` has been created/extracted from `TopicManagerRepository.builder()` to pass the required dependencies for `TopicManagerRepository`.

## How was this PR tested?

CI, frankenwar testing


internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicManagerStats.java
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  // Record latency for a specific sensor type if the topic manager stats is not null.
+  static void recordLatency(
+      TopicManagerStats topicManagerStats,
+      OCCURRENCE_LATENCY_SENSOR_TYPE sensorType,
+      long requestLatencyMs) {
+    if (topicManagerStats == null || sensorType == null) {
+      return;
+    }
```
FelixGV, line 57: That looks a bit weird to me... why make the function static and take an instance of itself? Is this a way to deal with disabled metrics? Maybe an alternative pattern would be that the `TopicManagerStats` constructor accepts a null `MetricsRepository` but in that case it becomes no-op. So the class containing the `TopicManagerStats` need not do a null check every time it calls a function on it.

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicManagerRepository.java
```diff
@@ -0,0 +1,72 @@
...
+
+  /**
+   * By default, return TopicManager for local PubSub cluster.
+   */
+  public TopicManager getLocalTopicManager() {
+    return getTopicManager(localPubSubAddress);
+  }
+
+  public TopicManager getTopicManager(String pubSubAddress) {
+    return topicManagers.computeIfAbsent(pubSubAddress, k -> createTopicManager(pubSubAddress));
```
FelixGV, line 43: The second param could be `this:: createTopicManager`, if you'd like.

FelixGV (commented):
A few comments, all minor... looks great so far. I still have about half to review.

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicMetadataFetcher.java
```diff
@@ -0,0 +1,673 @@
...
+  }
+
+  // load the cache with the latest offset
+  void populateCacheWithLatestOffset(PubSubTopicPartition pubSubTopicPartition) {
+    long now = System.nanoTime();
+    // if there is no entry in the cache or the entry has expired, we will try to populate the cache
+    ValueAndExpiryTime<Long> cachedValue = latestOffsetCache.putIfAbsent(
+        pubSubTopicPartition,
+        new ValueAndExpiryTime<>((long) StatsErrorCode.LAG_MEASUREMENT_FAILURE.code, now + cachedEntryTtlInNs));
+    if (cachedValue == null || (cachedValue.getExpiryTimeNs() <= now && cachedValue.tryAcquireUpdateLock())) {
```
lluwm, line 350: Nit, do you think we could switch the order here in the if condition to something below. I am thinking of a small inefficiency case here, where two threads get the `cachedValue` at the same time, after thread 1 passed `cachedValue.getExpiryTimeNs() <= now` condition, now thread 2 got the lock, finished the `updateValue(latestOffset)`. Now thread 1 continue with getting the lock and updating value etc, but it does't have to (as value has just been updated by thread 2), if it `getExpiryTimeNs` after   `tryAcquireUpdateLock`



`if (cachedValue == null || (cachedValue.tryAcquireUpdateLock() && cachedValue.getExpiryTimeNs() <= now)`

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicMetadataFetcher.java
```diff
@@ -0,0 +1,673 @@
...
+              pubSubMessageDeserializer,
+              pubSubClusterAddress);
+
+      closeables.add(pubSubConsumerAdapter);
+      if (!pubSubConsumerPool.offer(pubSubConsumerAdapter)) {
+        throw new VeniceException("Failed to initialize topicMetadataFetcherPool");
+      }
+    }
+
+    threadPoolExecutor = new ThreadPoolExecutor(
```
lluwm, line 101: Can we add some comments here to give a note to the reader that all async method in this class are executed by this dedicated thread pool?

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicManager.java
```diff
@@ -426,16 +364,24 @@ public boolean updateTopicMinInSyncReplica(PubSubTopic topicName, int minISR)
...
+  /**
+   * Get retention time for all topics in the pubsub cluster.
+   * @return a map of topic name to retention time in MS.
+   */
   public Map<PubSubTopic, Long> getAllTopicRetentions() {
-    return pubSubReadOnlyAdminAdapter.get().getAllTopicRetentions();
+    long startTime = System.currentTimeMillis();
+    Map<PubSubTopic, Long> topicRetentions = pubSubAdminAdapter.getAllTopicRetentions();
+    TopicManagerStats
+        .recordLatency(topicManagerStats, GET_ALL_TOPIC_RETENTIONS, System.currentTimeMillis() - startTime);
```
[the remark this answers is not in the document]
lluwm: Take a look at the `utils.Timer.run` and see if it can remove a bit of the boilerplate in this pattern.

lluwm (commented):
Nice work and it makes a lot of sense to me to have a centralized place/class to coordinate the work of getting offset. Thank you @sushantmane!

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/pubsub/api/producer/PubSubProducerAdapterTest.java
```diff
@@ -0,0 +1,4 @@
+package com.linkedin.venice.pubsub.api.producer;
+
+public class PubSubProducerAdapterTest {
```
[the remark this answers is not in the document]
sushantmane: Deleted.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/config/VeniceServerConfig.java
```diff
@@ -512,6 +518,11 @@ public VeniceServerConfig(VeniceProperties serverProperties, Map<String, Map<Str
     maxRequestSize = (int) serverProperties.getSizeInBytes(SERVER_MAX_REQUEST_SIZE, 256 * 1024);
     topicOffsetCheckIntervalMs =
         serverProperties.getInt(SERVER_SOURCE_TOPIC_OFFSET_CHECK_INTERVAL_MS, (int) TimeUnit.SECONDS.toMillis(60));
+    this.topicManagerMetadataFetcherConsumerPoolSize = serverProperties.getInt(
+        PUBSUB_TOPIC_MANAGER_METADATA_FETCHER_CONSUMER_POOL_SIZE,
+        PUBSUB_TOPIC_MANAGER_METADATA_FETCHER_CONSUMER_POOL_SIZE_DEFAULT_VALUE);
+    this.topicManagerMetadataFetcherThreadPoolSize = serverProperties
+        .getInt(PUBSUB_TOPIC_MANAGER_METADATA_FETCHER_THREAD_POOL_SIZE, topicManagerMetadataFetcherConsumerPoolSize);
```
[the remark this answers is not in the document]
sushantmane: Done

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/pubsub/api/producer/PubSubProducerAdapterTest.java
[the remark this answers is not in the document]
sushantmane: Removed, thanks

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/AggKafkaConsumerService.java
```diff
@@ -370,7 +369,11 @@ public ConsumedDataReceiver<List<PubSubMessage<KafkaKey, KafkaMessageEnvelope, L
       throw new VeniceException(
           "Kafka consumer service must exist for version topic: " + versionTopic + " in Kafka cluster: " + kafkaURL);
     }
-
+    TopicManager topicManager = storeIngestionTask.getTopicManager(kafkaURL);
+    if (topicManager != null) {
+      // prefetch and cache the latest offset for the topic partition
+      topicManager.prefetchAndCacheLatestOffset(pubSubTopicPartition);
```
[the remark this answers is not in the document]
sushantmane: yeah that's right.

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicManagerRepository.java
sushantmane: Yeah that's more concise. Updated.

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicMetadataFetcher.java
```diff
@@ -0,0 +1,673 @@
...
+  }
+
+  // API: Get the latest offsets for all partitions of a topic
+  /**
+   * Get the latest offsets for all partitions of a topic. This is a blocking call.
+   * @param topic topic to get latest offsets for
+   * @return a map of partition id to latest offset. If the topic does not exist, an empty map is returned.
+   */
+  Int2LongMap getTopicLatestOffsets(PubSubTopic topic) {
+    PubSubConsumerAdapter pubSubConsumerAdapter = acquireConsumer();
```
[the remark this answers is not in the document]
sushantmane: `acquireConsumer()` works on a `BlockingQueue` and is expected to return consumer client or wait until one becomes available.  It is expected to throw only `InterruptedException`, in which it doesn't remove consumer from the blocking queue.  So IMO having it in `try` doesn't add any value.  

We do need to make sure that `releaseConsumer` is called every time control reaches a `try` block`.

We could consider wrapping `BlockingQueue<PubSubConsumerAdapter>` in `AutoClosable` `PubSubConsumerResource` so that we can use it in `try-with-resource`. The cost will be an additional wrapper object per function call.

FelixGV (commented):
Finished first pass... a few more comments. Nothing dramatic, LG overall.

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicMetadataFetcher.java
```diff
@@ -0,0 +1,673 @@
...
+        // out of range or the topic does not have any message
+        return latestOffset;
+      }
+      return result;
+    } finally {
+      releaseConsumer(pubSubConsumerAdapter);
+    }
+  }
+
+  long getOffsetForTimeWithRetries(PubSubTopicPartition pubSubTopicPartition, long timestamp, int retries) {
```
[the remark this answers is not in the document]
FelixGV: I would prefer to continue leveraging `RetryUtils` otherwise the retry code is very bespoke and scattered... I have noticed that the new retry code has no sleep in-between attempts (neither exponential nor fixed), which does not seem ideal to me. Even if indeed we want retries without backoff, we should be able to achieve that as well with `RetryUtils`.

internal/venice-common/src/test/java/com/linkedin/venice/pubsub/manager/TopicMetadataFetcherTest.java
```diff
@@ -0,0 +1,5 @@
+package com.linkedin.venice.pubsub.manager;
+
+public class TopicMetadataFetcherTest {
```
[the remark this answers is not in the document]
sushantmane: Updated

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicManagerStats.java
FelixGV: Sorry... I didn't notice the notification for this comment... I think you're doing what I suggested in the currently submitted version of the code... thanks!

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicManagerRepository.java
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  public List<TopicManager> getAllTopicManagers() {
+    return new ArrayList<>(topicManagers.values());
```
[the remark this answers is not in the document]
FelixGV: In general, we want to avoid sharing internal data structures between classes (though in the case of tests, that is sometimes useful so we tolerate it, albeit at lower visibility...). In this case, it may be harmless since the collection is immutable, but that is just an implementation, and if it changes later on for some reason, the code may be left in a more fragile state, where some class can inadvertently mutate the internal state of another class through some ill-defined entry point...

The idea of having a public function specifically to invalidate the cache is to offer a precise contract of what can be done with this class, rather than offering lower-level building blocks and letting the caller get creative (which is sometimes good, but also a double-edged sword)...

Anyway, this is a minor nitpick and I leave it up to you.

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicMetadataFetcher.java
lluwm: It's probably a benign race (just efficiency rather than correctness issue), but do we want to just leave the race there due to the overhead you mentioned? On the other hand, there is not much overhead, if we look at the way of acquiring the "lock", it's not really a lock and it's non blocking + no need to release, is that right?

```
    boolean tryAcquireUpdateLock() {
      return isUpdateInProgress.compareAndSet(false, true);
    }
```

sushantmane: I believe there are tradeoffs to consider with each of these options:

1) `(cachedValue.tryAcquireUpdateLock() && cachedValue.getExpiryTimeNs() <= now)`
   With this option, for each request, we attempt to acquire the lock (update the atomic variable). If multiple threads attempt to update the atomic variable simultaneously, we incur a performance penalty. Additionally, with this approach, we must ensure that the lock is released (setting `isUpdateInProgress` to `false`) if the value is not expired, so that subsequent callers can obtain the lock.

2) `(cachedValue.getExpiryTimeNs() <= now && cachedValue.tryAcquireUpdateLock())`
   In this case, we acquire the lock only when necessary and hence reduced the chances of multiple threads trying to update atomic var at the same time. The downside is that occasionally we may encounter the race condition you mentioned.

I believe option (2) is simpler and relatively better than option (1).

We can further improve option (2) by skipping the update (although this means we will be checking the expiry time twice: once before grabbing the lock and once after obtaining it) if the value is not expired.

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicManagerRepository.java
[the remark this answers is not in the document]
sushantmane: Introducing the invalidateCache API in TopicManagerRepository does reveal a bit about what's going on under the hood of the TopicManagers. But it does seem like an improvement over our current approach. I think I agree with you

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicMetadataFetcher.java
lluwm: Do you think we can make this improvement in this pr?

> We can further improve option (2) by skipping the update (although this means we will be checking the expiry time twice: once before grabbing the lock and once after obtaining it) if the value is not expired.

sushantmane: Yeah sure, I'll update the PR. Thanks!

sushantmane: Updated. Could you please take another look? Thanks!

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaStoreIngestionService.java
```diff
@@ -317,17 +315,18 @@ public KafkaStoreIngestionService(
...
-        .setMetricsRepository(metricsRepository)
-        .setLocalKafkaBootstrapServers(serverConfig.getKafkaBootstrapServers())
-        .setPubSubConsumerAdapterFactory(pubSubClientsFactory.getConsumerAdapterFactory())
-        .setTopicDeletionStatusPollIntervalMs(DEFAULT_TOPIC_DELETION_STATUS_POLL_INTERVAL_MS)
-        .setTopicMinLogCompactionLagMs(DEFAULT_KAFKA_MIN_LOG_COMPACTION_LAG_MS)
-        .setKafkaOperationTimeoutMs(DEFAULT_KAFKA_OPERATION_TIMEOUT_MS)
-        .setPubSubProperties(this::getPubSubSSLPropertiesFromServerConfig)
-        .setPubSubAdminAdapterFactory(pubSubClientsFactory.getAdminAdapterFactory())
-        .build();
+    TopicManagerContext topicManagerContext =
```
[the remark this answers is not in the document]
sushantmane: Update

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicManager.java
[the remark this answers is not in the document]
sushantmane: > startTime rather than the elapsedTime as input param. Then it can do the clock call and subtraction interna

Updated

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicManagerRepository.java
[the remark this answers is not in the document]
sushantmane: Replaced `TMR::getAllTopicManagers` with `TMR::invalidateTopicManagerCaches`

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/api/PubSubConsumerAdapter.java
```diff
@@ -139,10 +147,18 @@ default long getOffsetLag(PubSubTopicPartition pubSubTopicPartition) {
    * @return The latest offset, which is zero or a positive value if an offset was collected by the consumer,
    *         or -1 if the offset cannot be determined or is not applicable.
    */
-  default long getLatestOffset(PubSubTopicPartition pubSubTopicPartition) {
+  default long getEndOffsetBasedOnMetrics(PubSubTopicPartition pubSubTopicPartition) {
```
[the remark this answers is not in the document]
sushantmane: Reverted changes to `PubSubConsumerAdapter` API names

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicManager.java
```diff
@@ -534,43 +507,48 @@ public void ensureTopicIsDeletedAndBlock(PubSubTopic pubSubTopic) {
...
+        if (attempts == PubSubConstants.PUBSUB_TOPIC_DELETE_RETRY_TIMES) {
           logger.error("Topic deletion for topic {} {}! Giving up!!", pubSubTopic, errorMessage, e);
           throw e;
         }
       }
     }
   }
 
+  // TODO: Evaluate if we need synchronized here
   public synchronized Set<PubSubTopic> listTopics() {
```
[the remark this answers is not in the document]
sushantmane: IMO listTopics is an expensive API.. until we've a mechanism in place to prevent simultaneous list topic request I prefer to keep `synchronized`

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicManager.java
```diff
@@ -714,7 +714,7 @@ public void close() {
       logger.warn("{} is already closed", this);
     }
     try {
-      closeAsync().get(600, TimeUnit.SECONDS);
+      closeAsync().get(2, TimeUnit.MINUTES);
```
FelixGV, line 717: That is a long time... why wait so long? What if we crash and the close doesn't get executed at all, would it be a problem? I don't think there is anything in TopicManager that must absolutely complete at all costs before shutting down, is there?

FelixGV (commented):
Just one minor comment... probably good to go very soon! Thanks for continuing to iterate on this big change!

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicManager.java
sushantmane: >I don't think there is anything in TopicManager that must absolutely complete at all costs before shutting down

Yeah, I think you're correct. 


I considered this as a defense against unbounded blocking during close, and since it's not expected to occur frequently I was thinking maybe 2min is ok.   

Should we consider reducing it to perhaps 30 seconds?

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicManager.java
```diff
@@ -522,10 +478,15 @@ public void ensureTopicIsDeletedAndBlock(PubSubTopic pubSubTopic) {
 
     logger.info("Deleting topic: {}", pubSubTopic);
     try {
-      pubSubWriteOnlyAdminAdapter.get().deleteTopic(pubSubTopic, Duration.ofMillis(kafkaOperationTimeoutMs));
+      long startTime = System.currentTimeMillis();
```
gaojieliu, line 481: Having a generic question about the admin operation latency:
Will these operation take more than 1ms or less than 1ms?
If these operations would normally take less than 1ms, we need to use nano seconds here, otherwise, the latency metric can report 0.

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicManager.java
```diff
@@ -651,72 +618,110 @@ public synchronized boolean containsTopicAndAllPartitionsAreOnline(
...
+    return topicMetadataFetcher.getProducerTimestampOfLastDataMessageCached(pubSubTopicPartition);
   }
 
   /**
    * Get offsets for only one partition with a specific timestamp.
    */
-  public long getPartitionOffsetByTime(PubSubTopicPartition pubSubTopicPartition, long timestamp) {
-    return partitionOffsetFetcher.getPartitionOffsetByTime(pubSubTopicPartition, timestamp);
+  public long getOffsetByTime(PubSubTopicPartition pubSubTopicPartition, long timestamp) {
+    return topicMetadataFetcher.getOffsetForTimeWithRetries(pubSubTopicPartition, timestamp, 25);
```
gaojieliu, line 686: how do we come up with `25` here?

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicManager.java
```diff
@@ -651,72 +618,110 @@ public synchronized boolean containsTopicAndAllPartitionsAreOnline(
...
+    return this.pubSubClusterAddress;
   }
 
-  // For testing only
-  public void setTopicConfigCache(Cache<PubSubTopic, PubSubTopicConfiguration> topicConfigCache) {
-    this.topicConfigCache = topicConfigCache;
+  @Override
+  public void close() {
+    if (isClosed.get()) {
+      logger.warn("{} is already closed", this);
```
gaojieliu, line 712: `return` directly if the TopicManager is already closed?

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicManager.java
```diff
@@ -651,72 +618,110 @@ public synchronized boolean containsTopicAndAllPartitionsAreOnline(
...
+    }
+    CompletableFuture.runAsync(() -> {
```
gaojieliu, line 714: Why closing it asynchronously?

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicManager.java
sushantmane: Removed blocking wait on close

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/api/PubSubConsumerAdapter.java
```diff
@@ -127,10 +127,18 @@ public interface PubSubConsumerAdapter extends AutoCloseable, Closeable {
    * @return The offset lag, which is zero or a positive value if a valid lag was collected by the consumer,
    *         or -1 if the lag cannot be determined or is not applicable.
    */
-  default long getOffsetLag(PubSubTopicPartition pubSubTopicPartition) {
+  default long getConsumerLagBasedOnMetrics(PubSubTopicPartition pubSubTopicPartition) {
```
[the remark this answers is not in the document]
sushantmane: Since there is no objection to keep name change in other places except `PubSubConsumerAdapter` I'm marking this comment as resolved

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicMetadataFetcher.java
```diff
@@ -0,0 +1,748 @@
...
+      // find the beginning offset
+      long earliestOffset =
+          pubSubConsumerAdapter.beginningOffset(pubSubTopicPartition, PUBSUB_OFFSET_API_TIMEOUT_DURATION_DEFAULT_VALUE);
+      if (earliestOffset == latestOffset) {
+        return Collections.emptyList(); // no records in this topic partition
+      }
+
+      // consume latest records
+      long consumePastOffset = Math.max(Math.max(latestOffset - lastRecordsCount, earliestOffset) - 1, -1);
+      pubSubConsumerAdapter.subscribe(pubSubTopicPartition, consumePastOffset);
```
gaojieliu, line 567: Since this class is using shared consumer here, shall we check whether the returned consumer has subscribed any other topic partitions or not?
Maybe it is a good practice to reset the consumer after acquiring the consumer from the consumer pool.

gaojieliu (commented):
Looks great overall, and left some minor comments.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -2568,20 +2568,20 @@ public long getBatchReplicationLag() {
           String sourceKafkaURL = getSourceKafkaUrlForOffsetLagMeasurement(pcs);
           // Consumer might not exist after the consumption state is created, but before attaching the corresponding
           // consumer.
-          long offsetLagOptional = getPartitionOffsetLag(sourceKafkaURL, currentLeaderTopic, pcs.getUserPartition());
+          long offsetLagOptional =
```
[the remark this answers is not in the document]
sushantmane: Good catch. Let me rename it

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicManager.java
sushantmane: based on what I've observed, the latencies of pub-sub ops typically fall within the millisecond or sub-millisecond range. To obtain more accurate measurements, we can convert these latencies to microsecond.

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicManager.java
sushantmane: It is from the current/existing code

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicManager.java
sushantmane: Oh yeah I missed return here. Thanks

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicManager.java
sushantmane: We want TM close to be best effort operation and don't want to block for it to complete

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicMetadataFetcher.java
sushantmane: Currently, any `subscribe` action performed on these shared consumers will always be followed by an `unsubscribe`. When we obtain a consumer, it should not have any existing subscriptions. If it does, it indicates a problem with this code/system.

We can adopt super defensive approach as you suggested, however.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/AggKafkaConsumerService.java
```diff
@@ -393,24 +392,32 @@ public ConsumedDataReceiver<List<PubSubMessage<KafkaKey, KafkaMessageEnvelope, L
 
     versionTopicStoreIngestionTaskMapping.put(storeIngestionTask.getVersionTopic().getName(), storeIngestionTask);
     consumerService.startConsumptionIntoDataReceiver(pubSubTopicPartition, lastOffset, dataReceiver);
-
+    TopicManager topicManager = storeIngestionTask.getTopicManager(kafkaURL);
+    if (topicManager != null) {
+      // prefetch and cache the latest offset for the topic partition
+      topicManager.prefetchAndCacheLatestOffset(pubSubTopicPartition);
```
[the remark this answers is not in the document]
sushantmane: done

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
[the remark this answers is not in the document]
sushantmane: done

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicManager.java
sushantmane: done

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicManager.java
sushantmane: Update to record latency in microseconds instead of millis

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicManager.java
gaojieliu: So is this a new logic or existing behavior?

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicManager.java
gaojieliu: I see.. `25` looks weird, but it is fine to leave it this way since it exists before the PR.

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicMetadataFetcher.java
gaojieliu: yeah, `unsubscribe` can be skipped if some unexpected error happen.

sushantmane: Update with defensive check

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicManager.java
sushantmane: New logic

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicManagerStats.java
```diff
@@ -0,0 +1,59 @@
...
+  EnumMap<SENSOR_TYPE, Sensor> getSensorsByTypes() {
+    return sensorsByTypes;
+  }
+
+  void recordLatency(SENSOR_TYPE sensorType, long startTimeInNs) {
+    if (sensorsByTypes == null || sensorType == null) {
+      return;
+    }
+    // convert ns to us and record
+    sensorsByTypes.get(sensorType).record((System.nanoTime() - startTimeInNs) >> 10);
```
gaojieliu, line 57: There is a utility method: `LatencyUtils#getLatencyInMS`, which can be leveraged here.
And `>>10` is not accurate for the conversion from NS to MS.

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicManager.java
gaojieliu: Got it.
So you worry that TopicManager#close can take a long time and it might block the stop procedure, right?
Today, I think most of the `close` functions are doing sync logic and if we want to speed it up, we can execute the closing logic concurrently if there are multiple closables.
If we can make it sync, it would be easier to reason IMO.

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicManagerStats.java
sushantmane: I avoided using `LatencyUtils#getLatencyInMS` because it relies on division ( which is expensive than shift) and we can tolerate slightly less precision in this context IMO

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicManager.java
sushantmane: Yeah initially TM::close was sync and we were closing all TMs in TMR concurrently. But I removed it after discussion with @FelixGV: we decided waiting for it to complete doesn't add significant value.

Here is the commit that removed sync close: https://github.com/linkedin/venice/pull/743/commits/70be6aa13f323a5edbbb557b02f6999d178f4d8d

gaojieliu: Got it.
If the impact has been discussed extensively, I trusted your conclusion.

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicManagerStats.java
sushantmane: You're right, but we have to start somewhere! 😄

I can revert it if using `LatencyUtils#getLatencyInMS` seems better choice

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicManager.java
FelixGV: Note that all metrics are in floating point, so even if we measure in nanos, we can still report it as fractional millis (i.e. via `LatencyUtils::getLatencyInMS`). Microseconds are uncommon, however... I don't know if we have any other metrics in that unit. I suggest we stick to (fractional) millis for conformity with the rest of the metrics code.

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicManagerStats.java
FelixGV: I think we should use `LatencyUtils::getLatencyInMS`... and if eventually we decide to use bit shifting instead of division, let us do it in that function so it affects everything! (but like Gaojie said, I don't know that we're at that stage yet...)

sushantmane: Updated PR to use `LatencyUtils::getLatencyInMS`

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicManager.java
sushantmane: Updated the PR to use `LatencyUtils::getLatencyInMS`

FelixGV (approved):
Thank you Sushant! Let's take this monster PR across the finish line 🚀🚢 !!!!!

sushantmane:
Thanks a lot @FelixGV, @lluwm, and @gaojieliu for the review, discussions, and help in improving this PR. 🙏 🚀

PR #744, opened 2023-11-09
[changelog] Tweak after image consumer to not consume view topic

ZacAttack:
## [changelog] Tweak after image consumer to not consume view topic
This changes the behavior of the after image consumer so that it consumes the version topic instead of consuming the change capture topic.  In order to fully accomodate this, a few changes had to be applied, including:

- Seek API's for the After image consumer had to supply a topic suffix so as to always stay on version topics.
- seek to EOP for now has to basically consume the version topic and then find the coordinates of the EOP message.  Because this has to be done also when we detect that a new version is available, we use a second consumer which is doing the search/seek for the EOP message.  Once it finds it, it then resubscribes the client to the appropriate coordinate.
- Tweak message processing so that it returns deletes that are found on the version topic (tweaks to tests reflect this).  To my knowledge, deletes in the version topic should only happen in the case of TTL repushes, and this isn't really done in production yet.
- The after image consumer uses a background thread to detect if a version swap has happened.  This is a temporary measure until we publish version swaps to version topics.  Once we have that, we should delete this logic, as this implementation is a little fraught in terms of possible edge cases.

## How was this PR tested?
Added new integration and unit tests to verify the behavior

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceAfterImageConsumerImpl.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceAfterImageConsumerImpl.java
@@ -55,10 +55,14 @@ public class VeniceAfterImageConsumerImpl<K, V> extends VeniceChangelogConsumerI
     super(changelogClientConfig, consumer);
     versionSwapDetectionThread = new VersionSwapDetectionThread();
     internalSeekConsumer = seekConsumer;
+    versionSwapDetectionIntervalTimeInMs = changelogClientConfig.getVersionSwapDetectionIntervalTimeInMs();
   }
 
   @Override
   public Collection<PubSubMessage<K, ChangeEvent<V>, VeniceChangeCoordinate>> poll(long timeoutInMs) {
+    if (!versionSwapDetectionThread.isAlive()) {
+      versionSwapDetectionThread.start();
+    }
     return internalPoll(timeoutInMs, "");
   }
 
@@ -85,7 +89,7 @@ public class VeniceAfterImageConsumerImpl<K, V> extends VeniceChangelogConsumerI
 
           // We need to get the internal consumer as we have to intercept the control messages that we would normally
           // filter out from the user
-          PubSubConsumerAdapter consumerAdapter = internalSeekConsumer.get().pubSubConsumer;
+          PubSubConsumerAdapter consumerAdapter = internalSeekConsumer.get().getPubSubConsumer();
 
           Map<PubSubTopicPartition, List<PubSubMessage<KafkaKey, KafkaMessageEnvelope, Long>>> polledResults;
           Map<Integer, Boolean> endOfPushConsumedPerPartitionMap = new HashMap<>();
@@ -129,8 +133,8 @@ public class VeniceAfterImageConsumerImpl<K, V> extends VeniceChangelogConsumerI
               // We polled all EOP messages, stop polling!
               break;
             }
-            this.seekToCheckpoint(checkpoints).get();
           }
+          this.seekToCheckpoint(checkpoints).get();
         } catch (InterruptedException | ExecutionException e) {
           throw new VeniceException(
               "Seek to End of Push Failed for store: " + storeName + " partitions: " + partitions.toString(),
@@ -141,7 +145,8 @@ public class VeniceAfterImageConsumerImpl<K, V> extends VeniceChangelogConsumerI
     });
   }
 
-  public CompletableFuture<Void> internalSeek(
+  @Override
+  protected CompletableFuture<Void> internalSeek(
       Set<Integer> partitions,
       PubSubTopic targetTopic,
       SeekFunction seekAction) {
@@ -185,7 +190,7 @@ public class VeniceAfterImageConsumerImpl<K, V> extends VeniceChangelogConsumerI
 
         // Seek to end of push
         if (currentVersion != maxVersion) {
-          // get current subscriptions and seek to endOfPus
+          // get current subscriptions and seek to endOfPush
           Set<Integer> partitions = new HashSet<>();
           for (PubSubTopicPartition partitionSubscription: subscriptions) {
             partitions.add(partitionSubscription.getPartitionNumber());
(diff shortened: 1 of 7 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceAfterImageConsumerImpl.java
```diff
@@ -1,17 +1,210 @@
...
+                  }
+                }
+              }
+            }
+            if (endOfPushConsumedPerPartitionMap.values().stream().allMatch(e -> e)) {
+              // We polled all EOP messages, stop polling!
+              break;
+            }
+          }
+          this.seekToCheckpoint(checkpoints).get();
```
haoxu07, line 137: Is it possible we will get the thread seek to EOP endlessly here, as `versionSwapDetectionThread` will keep triggering `seekToEndOfPush` and the main consumer always seek to checkpoints and then poll and then seekToCheckpoint?

ZacAttack: Not sure I follow, but I don't think so.  versionSwapDetectionThread is single threaded and we do a blocking call on seekToCheckpoint, and will only trigger the switch if the consumer isn't on the current version topic.  So I don't understand how this would loop.

haoxu07: I think I get the idea. This switch will only be triggered with new current version detected by the thread. So it might be good to log here as we jumped to the newer version and the EOP postition we find for debugging purpose.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceAfterImageConsumerImpl.java
```diff
@@ -1,17 +1,210 @@
...
+  }
+
+  @Override
+  public CompletableFuture<Void> seekToEndOfPush(Set<Integer> partitions) {
+    return CompletableFuture.supplyAsync(() -> {
+      synchronized (internalSeekConsumer) {
+        try {
+          // TODO: This implementation basically just scans the version topic until it finds the EOP message. The
+          // approach
+          // we'd like to do is instead add the offset of the EOP message in the VT, and then just seek to that offset.
```
haoxu07, line 86: Not related to this change, wondering how would you add the offset of EOP in the VT so that we can easily to query?

clients/da-vinci-client/src/test/java/com/linkedin/davinci/consumer/VeniceChangelogConsumerImplTest.java
```diff
@@ -138,8 +142,8 @@ public void testConsumeBeforeAndAfterImage() throws ExecutionException, Interrup
     verify(mockPubSubConsumer).subscribe(oldVersionTopicPartition, OffsetRecord.LOWEST_OFFSET);
 
     veniceChangelogConsumer.subscribe(new HashSet<>(Arrays.asList(0))).get();
-    veniceChangelogConsumer.seekToEndOfPush();
-    verify(mockPubSubConsumer, times(2)).subscribe(oldVersionTopicPartition, OffsetRecord.LOWEST_OFFSET);
+    // veniceChangelogConsumer.seekToEndOfPush();
```
haoxu07, line 145: Nit: remove this.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceAfterImageConsumerImpl.java
```diff
@@ -1,17 +1,210 @@
...
+      SeekFunction seekAction) {
+    if (!versionSwapDetectionThread.isAlive()) {
+      versionSwapDetectionThread.start();
+    }
+    return super.internalSeek(partitions, targetTopic, seekAction);
+  }
+
+  private class VersionSwapDetectionThread extends Thread {
+    VersionSwapDetectionThread() {
+      super("Version-Swap-Detection-Thread");
```
haoxu07, line 161: So we might have 1 thread per AfterImageConsumer, is it possible we will have multiple AfterImageConsumer for one store? If so, we would better to add more informative name to this thread. This better carry the informatition carried from Factory creating consumer provided by customer.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumerImpl.java
```diff
@@ -377,12 +385,15 @@ public CompletableFuture<Void> subscribeAll() {
 
   @Override
   public CompletableFuture<Void> seekToTimestamps(Map<Integer, Long> timestamps) {
+    return internalSeekToTimestamps(timestamps, ChangeCaptureView.CHANGE_CAPTURE_TOPIC_SUFFIX);
```
haoxu07, line 388: As we fixed the suffix type here, So seekToTimestamps and seekToTail would only meaningful to ChangeEventConsumer not afterImageConsumer?

haoxu07 (commented):
Overall I feel the protocol of version jumping is OK, just left some comments about other changes.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceAfterImageConsumerImpl.java
ZacAttack: I'm going to add it into a system store and make it queryable for the client.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceAfterImageConsumerImpl.java
ZacAttack: I'm not sure I can think of a scheme that would be especially helpful other then adding a random number at the end of the name.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumerImpl.java
ZacAttack: That is not the case.  We override this method in the afterimage consumer so as to not provide the suffix.  It's meaningful to both.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceAfterImageConsumerImpl.java
ZacAttack: Added thread id to the loggers.  I think that'd be a better approach in all cases because then the id would match up in thread dumps.

haoxu07 (approved):
LGTM! Thanks for preparing this! My only concern is we might need to give unique name or id when we construct every consumer from factory. But we can add it in the future change.

PR #746, opened 2023-11-10
[changelog] Patch topic dumper to work on change capture topics

ZacAttack:
## [changelog] Patch topic dumper to work on change capture topics
Previously, the topic dumper only worked on RT and version topics, now this should work on SR
and CC topics.

## How was this PR tested?
In production!


haoxu07 (approved):
LGTM!

PR #747, opened 2023-11-13
Fix string cast issue in VeniceProperties

m-nagarajan:
## Summary
Metadata periodic refresh failed with below for `PartitionerParams`
```
java.lang.ClassCastException: class org.apache.avro.util.Utf8 cannot be cast to class java.lang.String

```
Fix is to use `toString()` in `VeniceProperties` constructor rather than `(String)`

## How was this PR tested?
GH CI

### DIFF
--- a/clients/venice-client/src/test/java/com/linkedin/venice/fastclient/meta/RequestBasedMetadataTestUtils.java
+++ b/clients/venice-client/src/test/java/com/linkedin/venice/fastclient/meta/RequestBasedMetadataTestUtils.java
@@ -69,12 +69,15 @@ public class RequestBasedMetadataTestUtils {
       Schema storeValueSchema) {
     D2TransportClient d2TransportClient = mock(D2TransportClient.class);
 
+    Map<String, String> partitionerParams = new HashMap<>();
+    partitionerParams.put("testKey", "testValue");
+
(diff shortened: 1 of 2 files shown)


gaojieliu (approved):

clients/venice-client/src/test/java/com/linkedin/venice/fastclient/meta/RequestBasedMetadataTestUtils.java
```diff
@@ -69,12 +69,15 @@ public static D2TransportClient getMockD2TransportClient(
       Schema storeValueSchema) {
     D2TransportClient d2TransportClient = mock(D2TransportClient.class);
 
+    Map<String, String> partitionerParams = new HashMap<>();
+    partitionerParams.put("testKey", "testValue");
```
nisargthakkar, line 73: I understand this is a very simple change, but the test added does not test the problem this is trying to fix

m-nagarajan: `RequestBasedMetadataTest#testMetadata` will fail without the fix. Discussed offline with Nisarg and resolving this.

PR #748, opened 2023-11-14
[server] Cleanup CV states at start/stop time for II

gaojieliu:
## Summary
In the past, when II is enabled, the `StorageService` in the main process won't load on-disk databases, which made the original CV cleanup logic doesn't work for II at startup time.
With this code change when II is enabled, `HelixParticipationService` will try to collect the loaded databases from main process and the isolated process for CV state cleanup.

## How was this PR tested?
CI


m-nagarajan (approved):
Thanks @gaojieliu LGTM, left a NIT

clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/IsolatedIngestionBackend.java
```diff
@@ -248,6 +252,28 @@ public void prepareForShutdown() {
     isShuttingDown.set(true);
   }
 
+  @Override
+  public Map<String, Set<Integer>> getLoadedStoreAndUserPartitionsMapping() {
+    /**
+     * Merge the loaded store partitions from main process and isolated process.
+     */
+    Map<String, Set<Integer>> localMapping = super.getLoadedStoreAndUserPartitionsMapping();
+    LoadedStoreUserPartitionMapping isolatedMapping = mainIngestionRequestClient.getLoadedStoreUserPartitionMapping();
```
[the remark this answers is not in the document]
gaojieliu: Thanks and fixed it.

m-nagarajan (approved):
Thanks @gaojieliu

clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/IsolatedIngestionBackend.java
```diff
@@ -248,6 +252,28 @@ public void prepareForShutdown() {
...
+  @Override
+  public Map<String, Set<Integer>> getLoadedStoreUserPartitionsMapping() {
+    /**
+     * Merge the loaded store partitions from main process and isolated process.
+     */
+    Map<String, Set<Integer>> localMapping = super.getLoadedStoreUserPartitionsMapping();
+    LoadedStoreUserPartitionMapping isolatedMapping = mainIngestionRequestClient.getLoadedStoreUserPartitionMapping();
+    LOGGER.info(
+        "Got loaded store partition mapping from isolated process: {}",
+        isolatedMapping.storeUserPartitionMapping);
```
sixpluszero, line 264: nit: For prod server, this will be a very big log line containing thousands of mapping... it is not going to look good though. But probably this is what we can get now as trimming some of them is also useless....

gaojieliu: yes, there will be one list per store version, and I feel it is useful to track this behavior since II communication is tricky and there is no read traffic at start/stop time, so the impact to the online traffic is none and it will pollute the log a little bit and we can evaluate it later.

sixpluszero (approved):
Thanks Gaojie for the fix!

PR #750, opened 2023-11-14
[da-vinci][server][controller] Fetch latest update schema available for push status writer in all Venice components

sixpluszero:
## [da-vinci][server][controller] Fetch latest update schema available for push status writer in all Venice components
Previously, PushStatusStoreWriter uses PUSH_STATUS_SYSTEM_SCHEMA_STORE's current protocol id as value schema, and relying on a config: PUSH_STATUS_STORE_DERIVED_SCHEMA_ID (default = 1) to as update schema id, and use a hardcoded specific write op record as partial update record. 
This is wrong as 1-1 is our legacy schema type which is a union with content of [update schema, del op]. The actual static update record is only the first branch of the union. And we found that Fast-Avro fails to deserialize this schema as writer schema and 1-3 as reader schema. Even though vanilla Avro does not throw exception, but it does not give correct deserialized result as well. 
This PR fixed the above issue by asking all components that use PushStatusStoreWriter to fetch the latest value schema and latest update schema from StoreSchemaFetcher and schema repository. PushStatusStoreWriter will no longer use protocol ID / value schema from AvroProtocolDefinition as it might not be registered. It will also not rely on config to find update protocol ID.

This PR also performs some cleanup:
1. Remove PushStatusStoreRecordDeleter and merge functionality into PushStatuStoreWriter.
2. Make PushStatusStoreWriter/PushStatusStoreReader no longer optional in Admin code as it has been rolled out for more than 2 years.
3. Some other code cleanup.

## How was this PR tested?
Existing tests;
Adding new unit tests to pass coverage.


internal/venice-common/src/main/java/com/linkedin/venice/pushstatushelper/PushStatusStoreWriter.java
```diff
@@ -119,49 +120,76 @@ public void addToSupposedlyOngoingIncrementalPushVersions(
...
         storeName,
         instanceName);
     veniceWriterCache.prepareVeniceWriter(storeName)
-        .update(
-            pushStatusKey,
-            writeComputeRecord,
-            AvroProtocolDefinition.PUSH_STATUS_SYSTEM_SCHEMA_STORE.getCurrentProtocolVersion(),
-            derivedSchemaId,
-            null);
+        .update(pushStatusKey, updateBuilder.build(), valueSchemaId, derivedSchemaId, null);
```
FelixGV, line 148: Wow, it's nice to see how much less verbose the `UpdateBuilder` API is compared to using the `SpecificRecord`!

clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
```diff
@@ -187,8 +189,15 @@ public DaVinciBackend(
           new VeniceWriterFactory(backendProps.toProperties(), pubSubClientsFactory.getProducerAdapterFactory(), null);
       String pid = Utils.getPid();
       String instanceName = Utils.getHostName() + "_" + (pid == null ? "NA" : pid);
-      int derivedSchemaID = backendProps.getInt(PUSH_STATUS_STORE_DERIVED_SCHEMA_ID, 1);
-      pushStatusStoreWriter = new PushStatusStoreWriter(writerFactory, instanceName, derivedSchemaID);
+
+      // Fetch latest update schema's protocol ID for Push Status Store from Router.
+      ClientConfig pushStatusStoreClientConfig = ClientConfig.cloneConfig(clientConfig)
+          .setStoreName(VeniceSystemStoreType.DAVINCI_PUSH_STATUS_STORE.getZkSharedStoreName());
+      StoreSchemaFetcher schemaFetcher = ClientFactory.createStoreSchemaFetcher(pushStatusStoreClientConfig);
```
FelixGV, line 196: Can we do this in a try-with-resources so that it gets closed when we don't need it anymore?

FelixGV (commented):
Nice work! Code looks much better! Left just one minor comment.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
sixpluszero: Yeah, just fixed.

internal/venice-common/src/main/java/com/linkedin/venice/pushstatushelper/PushStatusStoreWriter.java
sixpluszero: Yeah, we should use it everywhere whenever possible. There should be basically no special case for system store.

FelixGV (approved):
Thank you!

PR #751, opened 2023-11-15
[da-vinci] Improve NativeMetadataRepository schema retrieval efficiency

xunyin8:
## Summary
[da-vinci] Improve NativeMetadataRepository schema retrieval efficiency

Currently metadata refresh in DVC would generate many unnecessary schema requests to the meta system store. The severity increases with the number of DVC instances and the number of value schema versions for the subscribed stores. Since value schemas are immutable in Venice we will now:

1. Only retrieve value schemas that we don't have and reuse the SchemaData object that we cache in the NativeMetadataRepository.

## How was this PR tested?
New unit test and existing integration test.


gaojieliu (approved):
Thanks for the quick fix.

PR #752, opened 2023-11-15
[router][server] Return Http status 503 for VeniceNoStoreException

majisourav99:
## Return Http status 503 for VeniceNoStoreException

Venice routers retry error requests only for 500+ http status codes. But when server returns BAD_REQUEST (400) for VeniceNoStoreException for existing resources, router does not trigger any error retries which causes unhealthy request on the client side. This PR changes the status code to 503 for such cases so that router can initiate error retries.

## How was this PR tested?

CI

### DIFF
--- a/services/venice-server/src/main/java/com/linkedin/venice/listener/StorageReadRequestHandler.java
+++ b/services/venice-server/src/main/java/com/linkedin/venice/listener/StorageReadRequestHandler.java
@@ -359,20 +359,17 @@ public class StorageReadRequestHandler extends ChannelInboundHandlerAdapter {
   }
 
   private HttpResponseStatus getHttpResponseStatus(VeniceNoStoreException e) {
-    HttpResponseStatus status = HttpResponseStatus.BAD_REQUEST;
-
-    // return SERVICE_UNAVAILABLE to kick off error retry in router when store version resource exists
     String topic = e.getStoreName();
     String storeName = Version.parseStoreFromKafkaTopicName(topic);
     int version = Version.parseVersionFromKafkaTopicName(topic);
     Store store = metadataRepository.getStore(storeName);
-    if (store == null) {
-      return status;
+
+    if (store == null || store.getCurrentVersion() != version) {
+      return HttpResponseStatus.BAD_REQUEST;
     }
-    if (store.getCurrentVersion() == version) {
-      status = HttpResponseStatus.SERVICE_UNAVAILABLE;
-    }
-    return status;
+
+    // return SERVICE_UNAVAILABLE to kick off error retry in router when store version resource exists
+    return HttpResponseStatus.SERVICE_UNAVAILABLE;
   }
 
   /**
--- a/services/venice-server/src/main/java/com/linkedin/venice/listener/StorageReadRequestHandler.java
+++ b/services/venice-server/src/main/java/com/linkedin/venice/listener/StorageReadRequestHandler.java
@@ -265,10 +265,11 @@ public class StorageReadRequestHandler extends ChannelInboundHandlerAdapter {
                           earlyTerminationException.getMessage(),
                           earlyTerminationException.getHttpResponseStatus()));
                 } else if (e instanceof VeniceNoStoreException) {
+                  // return SERVICE_UNAVAILABLE to kick off error retry in router
                   context.writeAndFlush(
                       new HttpShortcutResponse(
                           "No storage exists for: " + ((VeniceNoStoreException) e).getStoreName(),
-                          HttpResponseStatus.BAD_REQUEST));
+                          HttpResponseStatus.SERVICE_UNAVAILABLE));
                 } else {
                   LOGGER.error("Exception thrown in parallel batch get for {}", request.getResourceName(), e);
                   HttpShortcutResponse shortcutResponse =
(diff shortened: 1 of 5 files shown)


services/venice-server/src/test/java/com/linkedin/venice/listener/StorageReadRequestHandlerTest.java
```diff
@@ -2,6 +2,7 @@
 
 import static com.linkedin.venice.read.RequestType.SINGLE_GET;
 import static com.linkedin.venice.router.api.VenicePathParser.TYPE_STORAGE;
+import static io.netty.handler.codec.http.HttpResponseStatus.*;
```
sixpluszero, line 5: Let's remove static import.

services/venice-server/src/main/java/com/linkedin/venice/listener/StorageReadRequestHandler.java
```diff
@@ -313,8 +314,8 @@ public void channelRead(ChannelHandlerContext context, Object message) throws Ex
           }
           context.writeAndFlush(response);
         } catch (VeniceNoStoreException e) {
-          context.writeAndFlush(
-              new HttpShortcutResponse("No storage exists for: " + e.getStoreName(), HttpResponseStatus.BAD_REQUEST));
+          HttpResponseStatus status = getHttpResponseStatus(e);
```
sixpluszero, line 317: I got a question here: When it throw NoStoreException, does it mean storeRepository cannot find the store? If that's the case, the new method you add should not help right?

services/venice-server/src/test/java/com/linkedin/venice/listener/StorageReadRequestHandlerTest.java
majisourav99: Sure, hate Intellij not saving import rules :(

services/venice-server/src/main/java/com/linkedin/venice/listener/StorageReadRequestHandler.java
```diff
@@ -357,6 +358,23 @@ public void channelRead(ChannelHandlerContext context, Object message) throws Ex
...
+    String storeName = Version.parseStoreFromKafkaTopicName(topic);
+    int version = Version.parseVersionFromKafkaTopicName(topic);
+    Store store = metadataRepository.getStore(storeName);
+    if (store == null) {
+      return status;
+    }
+    if (store.getCurrentVersion() == version) {
+      status = HttpResponseStatus.SERVICE_UNAVAILABLE;
+    }
+    return status;
```
gaojieliu, line 375: Can we tweak the logic for readability?
```
if (store == null || store.getCurrentVersion() != version) {
    return HttpResponseStatus.BAD_REQUEST;
}
return HttpResponseStatus.SERVICE_UNAVAILABLE;
```

services/venice-server/src/test/java/com/linkedin/venice/listener/StorageReadRequestHandlerTest.java
```diff
@@ -621,4 +622,28 @@ public void testMisRoutedStoreVersion() throws Exception {
...
+    doReturn(1).when(store).getCurrentVersion();
+    when(storageEngine.isClosed()).thenReturn(true);
+    doReturn(null).when(storageEngineRepository).getLocalStorageEngine(any());
+    doReturn(keyString.getBytes()).when(request).getKeyBytes();
+    StorageReadRequestHandler requestHandler = createStorageReadRequestHandler();
+    requestHandler.channelRead(context, request);
+    ArgumentCaptor<HttpShortcutResponse> shortcutResponseArgumentCaptor =
+        ArgumentCaptor.forClass(HttpShortcutResponse.class);
+    verify(context).writeAndFlush(shortcutResponseArgumentCaptor.capture());
+    Assert.assertEquals(shortcutResponseArgumentCaptor.getValue().getStatus(), SERVICE_UNAVAILABLE);
```
gaojieliu, line 647: Can we validate the `BAD_REQUEST` code path as well by this unit test?

gaojieliu (commented):
Thanks for the quick fix and left two minor comments.

services/venice-server/src/main/java/com/linkedin/venice/listener/StorageReadRequestHandler.java
majisourav99: No, it throws that even if cannot see the storage engine loaded, which is kind of misnomer. we should ideally throw a new NoStorageEngineFoundException in getStorageEngineOrThrow.

sixpluszero: Oh I see. So this is the II CV bug fix on router side then.

services/venice-server/src/main/java/com/linkedin/venice/listener/StorageReadRequestHandler.java
majisourav99: Sure

services/venice-server/src/test/java/com/linkedin/venice/listener/StorageReadRequestHandlerTest.java
majisourav99: Sure

services/venice-server/src/main/java/com/linkedin/venice/listener/StorageReadRequestHandler.java
majisourav99: right

gaojieliu (approved):

PR #753, opened 2023-11-15
[vpj] Refactor VPJ logic to allow different compute engines for data writer job

nisargthakkar:
## Refactor VPJ logic to allow different compute engines for data writer job

This PR refactors a bunch of VPJ logic for the main MR job (referred to as "data writer" job) to allow different compute execution engines. In a future PR, Spark will be added as an execution engine and will replace MapReduce eventually.

The following is a summary of the changes:
1. `ComputeJob` is the interface to abstract the complexities of the execution engines
2. For the data writer job, an `abstract class` called `DataWriterComputeJob` has been added to add some common functionality that will be shared and doesn't change when the compute engine changes.
    * For MapReduce, `DataWriterMRJob` is the implementation of this abstraction
3. Different execution engines have different ways to collect and report metrics. A generic interface called `TaskTracker` is added to abstract this aspec.
4. For the data writer tasks, an interface called `DataWriterTaskTracker` is added for the same reason
    * For MapReduce, there are two implementations of this interface:
        1. `ReporterBackedMapReduceDataWriterTaskTracker`: which is used when the MR job is running
        2. `CounterBackedMapReduceDataWriterTaskTracker`: which is used when the MR job is finished
5. `VenicePushJob.java` had a lot of MR setup logic. Now, we rely on `DataWriterComputeJob` and `DataWriterTaskTracker` as abstractions for the data writer job.
6. `PushJobSetting`, `StoreSetting`, `TopicInfo` have merged and moved into `PushJobSetting` and also moved to a dedicated class  to improve sharability, reduce the size of `VenicePushJob` class and make the compute abstraction interfaces cleaner
7. `EngineTaskConfigProvider` - this class is used to get information about a job/task from within a running task
    * For MapReduce, `MapReduceEngineTaskConfigProvider` is the implementation of this interface.
8. The `RecordReader` logic has been cleaned up to encapsulate the Avro serde inside `AbstractAvroRecordReader`. Previously, `VeniceAvroRecordReader` and `VeniceVsonRecordReader` implemented the serialization logic.
9. `AbstractInputRecordProcessor` - This is the abstraction for the `Map` task. `AbstractVeniceMapper` extends `AbstractInputRecordProcessor`.
10. `AbstractPartitionWriter` - This is the abstraction for the `Reduce` task. `VeniceReducer` and `VeniceKafkaInputReducer` extend `AbstractPartitionWriter`.
11. `AbstractDataWriterTask` - This is the equivalent of `AbstractMapReduceTask`. `AbstractInputRecordProcessor`, `AbstractPartitionWriter`, `KafkaInputFormatCombiner` and `ValidateSchemaAndBuildDictMapper` extend `AbstractDataWriterTask`.
12. Public constants moved from `VenicePushJob` to `VenicePushJobConstants`
13. Iterating of input moved out from `AbstractVeniceRecordReader` into `VeniceRecordIterator`. `AbstractVeniceRecordReader` is now responsible for extracting raw key and value bytes from individual records.

## Config Changes
1. `data.writer.compute.job.class` - this config is used to specify the FQCN of the compute job

## How was this PR tested?

GH CI. Tested this code manually in staging environment for Avro, Vson and other internally supported systems


adamxchen:
I'd like too review this. What's your suggested review order for this large amount of changes?

nisargthakkar:
> I'd like too review this. What's your suggested review order for this large amount of changes?

I feel the order mentioned in the Summary would be a good order to review

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/mapreduce/datawriter/jobs/DataWriterMRJob.java
```diff
@@ -0,0 +1,394 @@
...
+        KAFKA_CONFIG_PREFIX,
+        KafkaInputRecordReader.KIF_RECORD_READER_KAFKA_CONFIG_PREFIX);
+    int passThroughPrefixListSize = passThroughPrefixList.size();
+    if (passThroughPrefixListSize > 1) {
+      /**
+       * The following logic will make sure there are no two prefixes, which will be fully overlapped.
+       * The algo is slightly inefficient, but it should be good enough for VPJ.
+       */
+      for (int i = 0; i < passThroughPrefixListSize; ++i) {
+        for (int j = 0; j < passThroughPrefixListSize; ++j) {
```
sixpluszero, line 230: nit: j can start from i + 1, and below we can do two startsWith check. It is slightly better.

sixpluszero: But actually I don't understand the check and the comment, could you give an example?

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/mapreduce/datawriter/jobs/DataWriterMRJob.java
```diff
@@ -0,0 +1,394 @@
...
+        props.getString(ZSTD_COMPRESSION_LEVEL, String.valueOf(Zstd.maxCompressionLevel())));
+    conf.setBoolean(ZSTD_DICTIONARY_CREATION_SUCCESS, pushJobSetting.isZstdDictCreationSuccess);
+
+    /** Allow overriding properties if their names start with {@link HADOOP_PREFIX}.
+     *  Allow overriding properties if their names start with {@link VeniceWriter.VENICE_WRITER_CONFIG_PREFIX}
+     *  Allow overriding properties if their names start with {@link ApacheKafkaProducerConfig.KAFKA_CONFIG_PREFIX}
+     *  Allow overriding properties if their names start with {@link KafkaInputRecordReader.KIF_RECORD_READER_KAFKA_CONFIG_PREFIX}
+     **/
+    List<String> passThroughPrefixList = Arrays.asList(
+        VeniceWriter.VENICE_WRITER_CONFIG_PREFIX,
```
sixpluszero, line 220: The above comment has 4 allows, but here only has 3. Is that expected?

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/mapreduce/datawriter/jobs/DataWriterMRJob.java
```diff
@@ -0,0 +1,394 @@
...
+    jobConf.setMapOutputValueClass(BytesWritable.class);
+    if (pushJobSetting.isSourceKafka) {
+      jobConf.setReducerClass(VeniceKafkaInputReducer.class);
+    } else {
+      jobConf.setReducerClass(VeniceReducer.class);
+    }
+  }
+
+  // Visible for testing
+  public void setJobClientWrapper(JobClientWrapper jobClientWrapper) {
```
sixpluszero, line 329: Just curious, if it is for unit test only, it can be default modifier, no need to be public.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/task/TaskTracker.java
```diff
@@ -0,0 +1,13 @@
...
+
+/**
+ * An interface to report task health and track task progress for different execution engines
+ */
+public interface TaskTracker {
+  default void heartbeat() {
+  }
+
+  default float getProgress() {
+    return -1;
```
sixpluszero, line 11: Hmm why progress is a float? Could you add JavaDoc to the interface methods so it could be easier for reader?

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/task/datawriter/DataWriterTaskTracker.java
```diff
@@ -0,0 +1,106 @@
+package com.linkedin.venice.hadoop.task.datawriter;
+
+import com.linkedin.venice.hadoop.task.TaskTracker;
+
+
+/**
+ * An interface to report and retrieve metrics related to data writer tasks.
+ */
+public interface DataWriterTaskTracker extends TaskTracker {
```
sixpluszero, line 9: 1. Should this be moved to some tracker package along with the 2 implementations?
2. 

> For MapReduce, there are two implementations of this interface:
ReporterBackedMapReduceDataWriterTaskTracker: which is used when the MR job is running
CounterBackedMapReduceDataWriterTaskTracker: which is used when the MR job is finished

Regarding this explanation, I am a bit confused. I think the ComputeJob has only one TaskTracker right? Or I am missing something...

sixpluszero (commented):
Thanks for the change!

I did not finish a pass. Actually I only took a look at point 1-5. There are some questions I am not clear but I will continue tomorrow.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/task/datawriter/AbstractPartitionWriter.java
```diff
@@ -257,19 +230,21 @@ protected boolean isEnableWriteCompute() {
...
+      DataWriterTaskTracker dataWriterTaskTracker) {
     /**
      * Don't use {@link BytesWritable#getBytes()} since it could be padded or modified by some other records later on.
      */
-    byte[] keyBytes = key.copyBytes();
     if (!values.hasNext()) {
       throw new VeniceException("There is no value corresponding to key bytes: " + ByteUtils.toHexString(keyBytes));
     }
-    byte[] valueBytes = values.next().copyBytes();
+    byte[] valueBytes = values.next();
```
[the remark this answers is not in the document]
sixpluszero: Oh is it because you changed the input type from BytesWritable to byte[]?

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/task/datawriter/AbstractPartitionWriter.java
```diff
@@ -710,8 +595,33 @@ private String printDuplicateKey(byte[] keyBytes) {
 
     @Override
     public void close() {
-      Utils.closeQuietlyWithErrorLogged(recordReader);
-      Utils.closeQuietlyWithErrorLogged(keySerializer);
+      // Nothing to do
+    }
+  }
+
+  public class DelegatingProducerCallback implements PubSubProducerCallback {
```
sixpluszero, line 602: I suppose this is the same as the removed method ReducerProduceCallback?

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/mapreduce/datawriter/reduce/VeniceReducer.java
```diff
@@ -0,0 +1,235 @@
...
+    if (this.reporter == null || !this.reporter.equals(reporter)) {
+      this.reporter = reporter;
+      return true;
+    }
+    return false;
+  }
+
+  // Visible for testing
+  @Override
+  protected boolean getExceedQuotaFlag() {
```
sixpluszero, line 98: nit: If it is for unit test, all the "protected" can be removed.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/DefaultInputDataInfoProvider.java
```diff
@@ -365,35 +351,41 @@ private Pair<Schema, Schema> checkAvroSchemaConsistency(
...
            * read the records to get training samples
            */
-          InputDataInfoProvider.loadZstdTrainingSamples(recordReader, pushJobZstdConfig);
+          InputDataInfoProvider.loadZstdTrainingSamples(fileIterator, pushJobZstdConfig);
         }
       }
     }
-    return new Pair<>(recordReader.getFileSchema(), recordReader.getStoreSchema());
+
+    Utils.closeQuietlyWithErrorLogged(fileIterator);
```
sixpluszero, line 372: We could use try with resource?

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/input/kafka/ttl/VeniceRmdTTLFilter.java
```diff
@@ -48,13 +50,11 @@ public abstract class VeniceRmdTTLFilter<INPUT_VALUE> extends AbstractVeniceFilt
...
-    long ttlInMs = TimeUnit.SECONDS.toMillis(props.getLong(VenicePushJob.REPUSH_TTL_IN_SECONDS));
-    long ttlStartTimestamp = props.getLong(VenicePushJob.REPUSH_TTL_START_TIMESTAMP);
-    this.filterTimestamp = ttlStartTimestamp - ttlInMs - 1;
-    this.schemaSource = new HDFSSchemaSource(
-        props.getString(VenicePushJob.VALUE_SCHEMA_DIR),
-        props.getString(VenicePushJob.RMD_SCHEMA_DIR));
+    ttlPolicy = TTLResolutionPolicy.valueOf(props.getInt(REPUSH_TTL_POLICY));
+    long ttlStartTimestamp = props.getLong(REPUSH_TTL_START_TIMESTAMP);
+    // Filter anything that is before the start timestamp.
+    this.filterTimestamp = ttlStartTimestamp - 1;
```
sixpluszero, line 56: This is not correct right? filter timestamp = TTL job start TS - rewind time in MS - 1.

internal/venice-common/src/main/java/com/linkedin/venice/status/PushJobDetailsStatus.java
```diff
@@ -6,7 +6,7 @@
 
 
 public enum PushJobDetailsStatus {
-  STARTED(0), COMPLETED(1), ERROR(2), NOT_CREATED(3), UNKNOWN(4), TOPIC_CREATED(5), WRITE_COMPLETED(6), KILLED(7),
+  STARTED(0), COMPLETED(1), ERROR(2), NOT_CREATED(3), UNKNOWN(4), TOPIC_CREATED(5), DATA_WRITER_COMPLETED(6), KILLED(7),
```
sixpluszero, line 9: Do we need to change this name? I am afraid it might lead to any compatibility issue.

sixpluszero (commented):
I think I basically complete a pass on the source code change (but I am pretty sure I could missed some files) I did not look into test files changes yet. 
I left some comments and questions, please feel free to take a look when you recover and back. Thank you for the changes!

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/jobs/ComputeJob.java
```diff
@@ -0,0 +1,46 @@
...
+    Status(int value) {
+      this.value = value;
+    }
+
+    public static Status valueOf(int value) {
+      return values[value];
+    }
+
+    public boolean isEnded() {
+      return this == SUCCEEDED || this == FAILED || this == FAILED_VERIFICATION || this == KILLED;
```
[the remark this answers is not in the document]
nisargthakkar: I don't think I need the ordinal. Let me see if it can be removed.

I can do `isTerminal` method, and it works in this case. When writing it originally, I was thinking we'd need other cases like `isRunning`, `isEnded`, etc and thought that maybe the conditional approach would have been better. But now that the only method needed is `isEnded`, I'll do the method you suggested.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/jobs/DataWriterComputeJob.java
```diff
@@ -0,0 +1,169 @@
...
+    if (dataWriterTaskTracker == null) {
+      throw new VeniceException("DataWriterTaskTracker is not set. Unable to validate the job status.");
+    }
+
+    PushJobSetting pushJobSetting = getPushJobSetting();
+
+    if (pushJobSetting.inputHasRecords) {
+      final long taskTrackerClosedCount = dataWriterTaskTracker.getPartitionWriterCloseCount();
+      if (pushJobSetting.isSourceKafka && pushJobSetting.repushTTLEnabled) {
+        LOGGER.info("Repush with ttl filtered out {} records", dataWriterTaskTracker.getRepushTtlFilterCount());
```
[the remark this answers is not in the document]
nisargthakkar: This will be used when we add repush logic to spark. I can remove it and add a todo or error message or something. Currently, it is dead code

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/jobs/DataWriterComputeJob.java
```diff
@@ -0,0 +1,169 @@
...
+    if (writeAclAuthorizationFailureCount != 0) {
+      throw new VeniceException("Expect 0 ACL authorization failure. Got count: " + writeAclAuthorizationFailureCount);
+    }
+    final long duplicateKeyWithDistinctCount = dataWriterTaskTracker.getDuplicateKeyWithDistinctValueCount();
+    if (duplicateKeyWithDistinctCount != 0) {
+      throw new VeniceException(
+          "Expect 0 duplicated key with distinct value. Got count: " + duplicateKeyWithDistinctCount);
+    }
+    final long totalKeySize = dataWriterTaskTracker.getTotalKeySize();
+    if (totalKeySize != 0) {
```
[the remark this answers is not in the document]
nisargthakkar: This function checks error scenarios. We would not expect 0 key size, but if we see it, it is an error and an exception will be thrown

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/jobs/DataWriterComputeJob.java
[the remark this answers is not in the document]
nisargthakkar: I added it since I thought it was one line and I'd forget to add it later when doing KIF work

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/mapreduce/datawriter/jobs/DataWriterMRJob.java
nisargthakkar: This code has been copied over from `VenicePushJob` class so I didn't look through it in detail: https://github.com/linkedin/venice/blob/d788799205965e267cdf278a418f21c9375e7fad/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java#L3090-L3106

It looks like it checks if no `passThroughPrefix` is a prefix of any other `passThroughPrefix`. So, we want to check `prefix1.prefix2` is not allowed if `prefix1` is in the list. Similarly, if `prefix2` is also on the list, then we should not allow `prefix1.prefix2` or `prefix2.prefix1`, so, we can't do `j = i+1; ...` type of loop

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/mapreduce/datawriter/jobs/DataWriterMRJob.java
nisargthakkar: There is some problem with the packaging. This particular function is invoked by `VenicePushJob` class, but the code there is triggered by tests

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/mapreduce/datawriter/jobs/DataWriterMRJob.java
nisargthakkar: Yeah, looks like the code is correct, but comment is wrong. I'll update it since `HADOOP_CONF` is not a `passThroughPrefix`, but the prefix needs to be stripped out

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/task/TaskTracker.java
nisargthakkar: Oh yeah, I missed that. Will do it

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/task/datawriter/DataWriterTaskTracker.java
nisargthakkar: 1. Currently, I have this packaging:
    a. com.linkedin.venice.hadoop.task.TaskTracker
    b. com.linkedin.venice.hadoop.task.datawriter.DataWriterTaskTracker
    c. com.linkedin.venice.hadoop.mapreduce.datawriter.task.CounterBackedMapReduceDataWriterTaskTracker
    d. com.linkedin.venice.hadoop.mapreduce.datawriter.task.ReporterBackedMapReduceDataWriterTaskTracker
    e.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/PushJobSetting.java
```diff
@@ -0,0 +1,159 @@
...
+  public transient RepushInfoResponse repushInfoResponse;
+
+  // Schema-properties
+  public boolean isAvro = true;
+  public int valueSchemaId; // Value schema id retrieved from backend for valueSchemaString
+  public int derivedSchemaId = -1;
+  public String keyField;
+  public String valueField;
+
+  public Schema fileSchema;
```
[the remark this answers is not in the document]
nisargthakkar: We need both in some place. So, instead of repeated `toString`, I added both to this object since it is expected to be an object that holds state during VPJ job.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/mapreduce/engine/MapReduceEngineTaskConfigProvider.java
```diff
@@ -0,0 +1,57 @@
...
+import org.apache.hadoop.mapred.TaskID;
+
+
+/**
+ * An implementation of {@link EngineTaskConfigProvider} to get information about a running job on the MapReduce engine.
+ */
+public class MapReduceEngineTaskConfigProvider implements EngineTaskConfigProvider {
+  // Ideally, we should be using the new hadoop configs ("mapreduce." prefix) but the values of the new and old configs
+  // are different. Continuing to use the old configs for now.
+  public static final String MAPRED_TASK_ID_PROP_NAME = "mapred.task.id";
```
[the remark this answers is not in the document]
nisargthakkar: Yeah. I don't remember it clearly right now. It's been a while. The comment I added above is the only thing I remember. If you want me to look deeper, I can check and get back

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/mapreduce/engine/MapReduceEngineTaskConfigProvider.java
```diff
@@ -0,0 +1,57 @@
...
+  @Override
+  public int getTaskId() {
+    TaskAttemptID taskAttemptID = TaskAttemptID.forName(getJobConf().get(MAPRED_TASK_ID_PROP_NAME));
+    if (taskAttemptID == null) {
+      throw new UndefinedPropertyException(
+          MAPRED_TASK_ID_PROP_NAME + " not found in the " + JobConf.class.getSimpleName());
+    }
+    TaskID taskID = taskAttemptID.getTaskID();
+    if (taskID == null) {
+      throw new NullPointerException("taskAttemptID.getTaskID() is null");
```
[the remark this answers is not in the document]
nisargthakkar: This is copied code from here: https://github.com/linkedin/venice/blob/87b957d0f2bf5ac98891421cfd0465136083bf17/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/AbstractMapReduceTask.java#L78-L81. I think it's okay to leave it as is. What do you think?

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/task/datawriter/AbstractInputRecordProcessor.java
```diff
@@ -51,46 +52,46 @@ public abstract class AbstractVeniceMapper<INPUT_KEY, INPUT_VALUE> extends Abstr
...
   protected AbstractVeniceRecordReader<INPUT_KEY, INPUT_VALUE> veniceRecordReader;
-  protected FilterChain<INPUT_VALUE> veniceFilterChain;
+  private static final byte[] EMPTY_BYTES = new byte[0];
+  private final AtomicReference<byte[]> processedKey = new AtomicReference<>();
+  private final AtomicReference<byte[]> processedValue = new AtomicReference<>();
+  private boolean firstRecord = true;
 
-  @Override
-  public void map(
+  // TODO: Find a better name
```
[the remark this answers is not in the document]
nisargthakkar: Oh, I forgot to remove this 🤦. I think there are other instances of such TODOs. I'll audit all of them

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/input/recordreader/avro/AbstractAvroRecordReader.java
```diff
@@ -0,0 +1,134 @@
...
+          .createRecord(fileSchema.getName(), fileSchema.getDoc(), fileSchema.getNamespace(), fileSchema.isError());
+      outputSchema.setFields(outputSchemaFields);
+    } else {
+      outputSchema = fileSchema;
+    }
+
+    Schema.Field outputValueField = getField(outputSchema, valueFieldStr);
+    valueFieldPos = outputValueField.pos();
+
+    this.generatePartialUpdateRecordFromInput = updateSchema != null;
```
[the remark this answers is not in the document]
nisargthakkar: I now moved it into `VeniceAvroRecordReader`

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/task/datawriter/AbstractPartitionWriter.java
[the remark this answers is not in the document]
nisargthakkar: Yeah, exactly. The `copyBytes` is now done in `VeniceReducer#reduce`:
```
    processValuesForKey(key.copyBytes(), mapIterator(values, BytesWritable::copyBytes), dataWriterTaskTracker);
```

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/task/datawriter/AbstractPartitionWriter.java
```diff
@@ -435,6 +393,11 @@ private void telemetry() {
     }
   }
 
+  // Visible for testing
+  protected boolean getExceedQuotaFlag() {
```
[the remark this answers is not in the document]
nisargthakkar: This method gets called by `VeniceReducer#getExceedQuotaFlag` and since the packages are different, it is set as `protected`

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/task/datawriter/AbstractPartitionWriter.java
```diff
@@ -444,8 +407,9 @@ public void close() throws IOException {
         boolean shouldEndAllSegments = false;
         try {
           veniceWriter.flush();
-          shouldEndAllSegments = messageErrored.get() == 0 && messageSent == messageCompleted.get()
-              && previousReporter.getProgress() == 1.0;
+          shouldEndAllSegments =
+              messageErrored.get() == 0 && messageSent == messageCompleted.get() && (dataWriterTaskTracker == null
+                  || dataWriterTaskTracker.getProgress() == 1.0 || dataWriterTaskTracker.getProgress() == -1);
```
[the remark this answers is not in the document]
nisargthakkar: Yeah, I'll update the javadoc. In MR, the progress is a float/double between 0 and 1. So, I kept the same convention in our abstracted system too.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/mapreduce/datawriter/reduce/VeniceReducer.java
nisargthakkar: The problem here is with the structure of tests - I did not add tests specific to `AbstractPartitionWriter`, and so, it falls back to the existing tests for `VeniceReducer`. However, since most of the core functionality has moved to `AbstractPartitionWriter`, there is a chaining of function calls to set these configs.

`AbstractPartitionWriter` and `VeniceReducer` are in different packages and we will not be able to invoke the methods in `AbstractPartitionWriter` if the access specifier is not `protected`

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/DefaultInputDataInfoProvider.java
nisargthakkar: Yeah, that would be better. I didn't do that since I'd have to catch the exceptions and handle them, but it is a safer code style. Will update it

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/input/kafka/ttl/VeniceRmdTTLFilter.java
nisargthakkar: I modified the variables and constants from what was used before. `REPUSH_TTL_START_TIMESTAMP` is set as `TTL job start TS - rewind time in MS`

internal/venice-common/src/main/java/com/linkedin/venice/status/PushJobDetailsStatus.java
nisargthakkar: I don't think it will be a problem since we don't use these names anywhere - just their valus

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/task/datawriter/AbstractPartitionWriter.java
nisargthakkar: Yeah

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/mapreduce/datawriter/jobs/DataWriterMRJob.java
nisargthakkar: I re-read your comment and I'll make that change

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/jobs/ComputeJob.java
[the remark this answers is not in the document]
nisargthakkar: It looks like the ordinals and the `isEnded` are not used. I'll remove both of these

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/mapreduce/datawriter/jobs/DataWriterMRJob.java
nisargthakkar: I also feel  that this check is not necessary, as the prefixes are constants. But I didn't make any functional change to this part in the refactoring. I feel we could later move these checks to a test case instead of running it in a running VPJ job

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/PushJobSetting.java
```diff
@@ -0,0 +1,159 @@
...
+import java.io.Serializable;
+import java.util.Map;
+import org.apache.avro.Schema;
+import org.apache.hadoop.mapred.Partitioner;
+
+
+/**
+ * This class carries the state for the duration of the VenicePushJob. Consider breaking changes carefully.
+ */
+public class PushJobSetting implements Serializable {
```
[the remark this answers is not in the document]
nisargthakkar: The reason I added this is this is the class that holds the state during the lifetime of the push job. It is also present in many interfaces now. Internally at LI we add some implementations of these interfaces to support Dali. Similar to this, other users of Venice might also set such custom implementations. So, when we make changes to this state, we should be mindful about breaking such integrations

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/mapreduce/engine/MapReduceEngineTaskConfigProvider.java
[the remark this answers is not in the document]
nisargthakkar: The different values for old and new configs come from MR framework and not anything Venice controls

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/task/datawriter/AbstractInputRecordProcessor.java
```diff
@@ -51,46 +52,46 @@ public abstract class AbstractVeniceMapper<INPUT_KEY, INPUT_VALUE> extends Abstr
...
-      throws IOException {
+  private void maybeSprayAllPartitions(
+      BiConsumer<byte[], byte[]> recordEmitter,
+      DataWriterTaskTracker dataWriterTaskTracker) {
     /** First map invocation, since the {@link recordKey} will be set after this. */
-    if (TASK_ID_NOT_SET == getTaskId()) {
+    if (getTaskId() == TASK_ID_NOT_SET) {
       throw new IllegalStateException("attemptID not set!");
     }
     if (TASK_ID_WHICH_SHOULD_SPRAY_ALL_PARTITIONS != getTaskId()) {
```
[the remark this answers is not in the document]
nisargthakkar: I don't remember it clearly, but I think it was added to ensure that all reducers get launched even if there is no data for that partition. I believe this was done to be confident that a job doesn't get completed prematurely.

So, Mapper with task id 0 will emit a dummy record (key is empty `byte[]`) for each partition. The reducer will then detect this dummy record and drop it.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/task/datawriter/AbstractPartitionWriter.java
nisargthakkar: Actually, this was not true. The `ReducerProduceCallback` is still there in VeniceReducer. However, after I went through the implementation, I found I should move that callback and it's implementation into `AbstractPartitionWriter` and removed `DelegatingProducerCallback` completely.

internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
```diff
@@ -1347,12 +1347,12 @@ private ConfigKeys() {
   /**
    * The job tracker identifier as part of a map reduce job id.
    */
-  public static final String PUSH_JOB_MAP_REDUCE_JT_ID = "push.job.map.reduce.jt.id";
+  public static final String PUSH_JOB_COMPUTE_JOB_ID = "push.job.compute.job.id";
```
ZacAttack, line 1350: Just a question vs. an actual issue.  I suspect this is filled in exclusively programmatically, theres no one that we break by changing this string?

clients/venice-push-job/src/test/java/com/linkedin/venice/hadoop/mapreduce/datawriter/reduce/TestVeniceReducer.java
```diff
@@ -71,15 +78,15 @@ public void testReducerPutWithTooLargeValueAndChunkingDisabled() {
     AbstractVeniceWriter mockWriter = mock(AbstractVeniceWriter.class);
     when(mockWriter.put(any(), any(), anyInt(), any(), any()))
         .thenThrow(new RecordTooLargeException("expected exception"));
-    testReduceWithTooLargeValueAndChunkingDisabled(mockWriter, setupJobConf());
+    testReduceWithTooLargeValueAndChunkingDisabled(mockWriter, setupJobConf(100));
```
ZacAttack, line 81: Is it ever not 100?

internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
nisargthakkar: Yep. This is set and used only by the reducers

clients/venice-push-job/src/test/java/com/linkedin/venice/hadoop/mapreduce/datawriter/reduce/TestVeniceReducer.java
nisargthakkar: Yeah, there is one test where the value is 1 and a few tests where this value is obtained from a `dataProvider` and the space of values there is `1, 10, 1000`

PR #754, opened 2023-11-15
[WIP][da-vinci] Allow for custom materialized views in Da-Vinci

elijahgrimaldi:
## Summary
This is a work in progress to give users the ability to create different ‘views’ of a given Venice data set without having to either reimplement Venice versioning/bootstrapping or push a transformed copy of a dataset which might already exist in Venice.  

We would like to give users the ability to set up individual application specific data transformation, storage or filtering to specify what Davinci should do as it ingests every record.


elijahgrimaldi:
Should we also add an overloaded constructor for the generic/specific davinci cliens?

PR #755, opened 2023-11-16
[controller] Delete da-vinci push status asynchronously

shuhui-liu:
## Summary

When controller zk listener thread handles error push and deletes user store version, it will produce a delete message to davinci push status system store RT. If RT does not exist, zk listener thread will wait infinitely. Since the thread holds the store write lock forever, admin execution task threads will be blocked when processing the store's admin operations. What's worse, as admin execution tasks are shared, other stores' admin operations cannot be processed either.

This commit submits a task to delete davinci push status, wait for up to 30s to get the result, and cancel the task after timeout.

## How was this PR tested?

New integ test

### DIFF
--- a/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
@@ -231,7 +231,12 @@ import java.util.Optional;
 import java.util.Properties;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import java.util.function.Consumer;
 import java.util.stream.Collectors;
 import javax.annotation.Nonnull;
@@ -321,6 +326,7 @@ public class VeniceHelixAdmin implements Admin, StoreCleaner {
   private static final int INTERNAL_STORE_GET_RRT_TOPIC_ATTEMPTS = 3;
   private static final long INTERNAL_STORE_RTT_RETRY_BACKOFF_MS = TimeUnit.SECONDS.toMillis(5);
   private static final int PARTICIPANT_MESSAGE_STORE_SCHEMA_ID = 1;
+  private static final long PUSH_STATUS_STORE_WRITER_TIMEOUT_MS = TimeUnit.SECONDS.toMillis(30);
 
   static final int VERSION_ID_UNSET = -1;
 
@@ -3117,7 +3123,6 @@ public class VeniceHelixAdmin implements Admin, StoreCleaner {
   /**
    * Delete version from cluster, removing all related resources
    */
-  @Override
   public void deleteOneStoreVersion(String clusterName, String storeName, int versionNumber) {
     deleteOneStoreVersion(clusterName, storeName, versionNumber, false);
   }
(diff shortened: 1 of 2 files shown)


services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -3156,11 +3155,16 @@ private void deleteOneStoreVersion(String clusterName, String storeName, int ver
...
-          getPushStatusStoreWriter().deletePushStatus(
-              storeName,
-              deletedVersion.get().getNumber(),
-              Optional.empty(),
-              deletedVersion.get().getPartitionCount());
+          PubSubTopic pushStatusStoreRT = pubSubTopicRepository
+              .getTopic(Version.composeRealTimeTopic(VeniceSystemStoreUtils.getDaVinciPushStatusStoreName(storeName)));
+          if (getTopicManager().containsTopicAndAllPartitionsAreOnline(pushStatusStoreRT)) {
+            getPushStatusStoreWriter().deletePushStatus(storeName, deletedVersion.get().getNumber(), Optional.empty(),
+                deletedVersion.get().getPartitionCount());
```
FelixGV, line 3162: Hmm... even though we check here, there is still a race condition between the time of check and the time of use... Also, the check might be expensive (maybe not a concern for this particular code path, but it is a concern in general, given our Kafka MD load tends to be high already...).

So I wonder if we should make the delete go async in all cases... whether it blocks or not should not block a ZK thread.

Please LMK what you think.

shuhui-liu: Yeah current commit aims to reduce the possibility of producing to a nonexistent topic but does not eliminate it. The chance of topic deletion between topic existence check and writing message to the topic should be very low...I can change the topic existence check to topic truncation check to even lower the possibility.

Regarding "make the delete go async", do you mean that making the "writing delete message to system store RT" part async?

FelixGV: Yeah, like... don't even check the topic existence, but execute the delete in some other thread pool, with a max execution time. Do we need to block on this delete before moving forward to the next step of this function? If yes, we can block in-line (but with a timeout), if not, let it go fully async. Or a third option is to block only at the end of the function... IDK what is the right semantic, but in all cases, the point is: let's never allow the code to block forever.

Or anyway, that's just what I'm thinking off the top of my head. Please LMK if you have a concern with this approach.

shuhui-liu: Thanks for the elaboration. The next step of the function is to check and delete user store RT, which should not be blocked by producing system store delete message. IMHO, we still need a timeout in this case. Without timeout, when producing to nonexistent topics, threads will block forever and finally there will be no thread available in the thread pool to produce system store delete message for other stores/versions.

shuhui-liu: I tried a few ways to go async for dvc push status deletion and here are the results:
1. One executor service per class, fire and forget: if a topic does not exist, the thread will wait forever, other stores/versions cannot delete dvc push status.
2. Two executor services per class, one for execution and the other for monitor, fire and forget, the monitor thread waits for 30s and cancel task after timeout: if task1 and task2 are submitted at the same time, and task1 does not finish in 30s, the monitor thread will cancel both task1 and task2, even though task2 is not run at all.
3. One executor service per class, fire, wait, and cancel task after timeout: same problem with 2, task2 could be cancelled without running.
4. One executor service per method call, fire, wait, and cancel task after timeout: stores won't affect each other, thread will not wait forever, executor service will shutdown after timeout or getting the result.

I decided to go with option 4 and updated the PR. When you have time, please take a look and let me know if you have any concern.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -3156,11 +3161,26 @@ private void deleteOneStoreVersion(String clusterName, String storeName, int ver
           cleanUpViewResources(new Properties(), store, deletedVersion.get().getNumber());
         }
         if (store.isDaVinciPushStatusStoreEnabled()) {
-          getPushStatusStoreWriter().deletePushStatus(
-              storeName,
-              deletedVersion.get().getNumber(),
-              Optional.empty(),
-              deletedVersion.get().getPartitionCount());
+          ExecutorService executor = Executors.newSingleThreadExecutor();
```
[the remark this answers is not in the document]
shuhui-liu: Would the executor thread get blocked indefinitely? considering that we have future.get with timeout, and cancel the task after timeout.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -3156,11 +3161,26 @@ private void deleteOneStoreVersion(String clusterName, String storeName, int ver
...
+              () -> getPushStatusStoreWriter().deletePushStatus(
+                  storeName,
+                  deletedVersion.get().getNumber(),
+                  Optional.empty(),
+                  deletedVersion.get().getPartitionCount()));
+          try {
+            future.get(PUSH_STATUS_STORE_WRITER_TIMEOUT_MS, TimeUnit.MILLISECONDS);
+            LOGGER.info("Deleted da-vinci push status of {}_v{}", storeName, versionNumber);
+          } catch (InterruptedException e) {
+            LOGGER.error("Got interrupted when deleting da-vinci push status of {}_v{}", storeName, versionNumber);
```
FelixGV, line 3175: Swallowing interrupted exceptions is typically considered a bad practice... I think in Venice we mostly use the approach 4.3 mentioned in [this blog post](https://www.baeldung.com/java-interrupted-exception#3-custom-exception-handling), though feel free to pick any approach you prefer...

FelixGV (commented):
LGTM overall, just left a minor comment. I think we could merge this soon...

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
shuhui-liu: Thanks for catching this! I adopted approach 4.2 in the blog post, because I wanna RT deletion (next step of the function) to be executed. If throwing an exception, potential RT deletion won't happen after this specific version deletion.

FelixGV (approved):
Thanks Shuhui!

FelixGV (approved):
Thank you!

PR #757, opened 2023-11-17
[server] Introduced a repair service for stuck shared consumer

gaojieliu:
Recently, we noticed that some shared consumers were stuck since they
were blocked by producing to some non-existing topics, and many other
store versions got affected since the stuck consumer is being shared
by many other stores.
This PR introduces a way to detect and repair stuck consumer for above
situation.
In the high level, there will be a thread periodically check whether consumer#poll
is being invoked regularly or not, and if not, the thread will scan all
the ingestion tasks to see whether they are producing to any non-existing topics
or not. And for the store ingestion task, which is talking to some non-existing topic,
the repair thread will try to close the stuck KafkaProducer and kill the corresponding
ingestion task.
This PR also checked the non-existing topics for a configurable duration to tolerate
tranisent topic metadata propagaton delay.

5 new server configs:
```
server.stuck.consumer.repair.enabled : default true
server.stuck.consumer.repair.second : default 1 min
server.stuck.consumer.repair.threshold.second : default 5 mins
server.non.existing.topic.ingestion.task.kill.threshold.second: default 15 mins
server.non.existing.topic.check.retry.interval.second: default 1 min
```

3 new server metrics:
```
.StuckConsumerRepair--stuck_consumer_found.OccurrenceRate
.StuckConsumerRepair--ingestion_task_repair.OccurrenceRate
.StuckConsumerRepair--repair_failure.OccurrenceRate
```

## How was this PR tested?
CI

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/AggKafkaConsumerService.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/AggKafkaConsumerService.java
@@ -66,10 +66,7 @@ public class AggKafkaConsumerService extends AbstractVeniceService {
   private final Function<String, String> kafkaClusterUrlResolver;
 
   private final Map<String, StoreIngestionTask> versionTopicStoreIngestionTaskMapping = new VeniceConcurrentHashMap<>();
-  private final long stuckConsumerRepairThresholdMs;
-  private final long nonExistingTopicIngestionTaskKillThresholdMs;
   private ScheduledExecutorService stuckConsumerRepairExecutorService;
-  private final StuckConsumerRepairStats stuckConsumerRepairStats;
 
   public AggKafkaConsumerService(
       final PubSubConsumerAdapterFactory consumerFactory,
@@ -100,11 +97,6 @@ public class AggKafkaConsumerService extends AbstractVeniceService {
     this.sslPropertiesSupplier = sslPropertiesSupplier;
     this.kafkaClusterUrlResolver = serverConfig.getKafkaClusterUrlResolver();
 
-    this.stuckConsumerRepairThresholdMs =
-        TimeUnit.SECONDS.toMillis(serverConfig.getStuckConsumerDetectionRepairThresholdSecond());
-    this.nonExistingTopicIngestionTaskKillThresholdMs =
-        TimeUnit.SECONDS.toMillis(serverConfig.getNonExistingTopicIngestionTaskKillThresholdSecond());
-    this.stuckConsumerRepairStats = new StuckConsumerRepairStats(metricsRepository);
     if (serverConfig.isStuckConsumerRepairEnabled()) {
       this.stuckConsumerRepairExecutorService = Executors.newSingleThreadScheduledExecutor(
           new DaemonThreadFactory(this.getClass().getName() + "-StuckConsumerRepair"));
@@ -113,10 +105,10 @@ public class AggKafkaConsumerService extends AbstractVeniceService {
           getStuckConsumerDetectionAndRepairRunnable(
               kafkaServerToConsumerServiceMap,
               versionTopicStoreIngestionTaskMapping,
-              stuckConsumerRepairThresholdMs,
-              nonExistingTopicIngestionTaskKillThresholdMs,
+              TimeUnit.SECONDS.toMillis(serverConfig.getStuckConsumerDetectionRepairThresholdSecond()),
+              TimeUnit.SECONDS.toMillis(serverConfig.getNonExistingTopicIngestionTaskKillThresholdSecond()),
               TimeUnit.SECONDS.toMillis(serverConfig.getNonExistingTopicCheckRetryIntervalSecond()),
-              stuckConsumerRepairStats,
+              new StuckConsumerRepairStats(metricsRepository),
               killIngestionTaskRunnable),
           intervalInSeconds,
           intervalInSeconds,
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/AggKafkaConsumerService.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/AggKafkaConsumerService.java
@@ -4,6 +4,7 @@ import static com.linkedin.venice.ConfigKeys.KAFKA_BOOTSTRAP_SERVERS;
 
 import com.linkedin.davinci.config.VeniceServerConfig;
 import com.linkedin.davinci.ingestion.consumption.ConsumedDataReceiver;
+import com.linkedin.davinci.stats.StuckConsumerRepairStats;
 import com.linkedin.venice.exceptions.VeniceException;
 import com.linkedin.venice.kafka.TopicManagerRepository;
 import com.linkedin.venice.kafka.protocol.KafkaMessageEnvelope;
@@ -16,15 +17,23 @@ import com.linkedin.venice.pubsub.api.PubSubTopic;
 import com.linkedin.venice.pubsub.api.PubSubTopicPartition;
 import com.linkedin.venice.service.AbstractVeniceService;
 import com.linkedin.venice.throttle.EventThrottler;
+import com.linkedin.venice.utils.DaemonThreadFactory;
 import com.linkedin.venice.utils.SystemTime;
+import com.linkedin.venice.utils.Utils;
 import com.linkedin.venice.utils.concurrent.VeniceConcurrentHashMap;
 import io.tehuti.metrics.MetricsRepository;
 import it.unimi.dsi.fastutil.objects.Object2IntMap;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
 import java.util.Set;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.function.Consumer;
 import java.util.function.Function;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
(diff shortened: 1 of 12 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/AggKafkaConsumerService.java
```diff
@@ -100,6 +131,64 @@ public void stopInner() throws Exception {
...
+      }
+      if (!scanStoreIngestionTaskToFixStuckConsumer) {
+        return;
+      }
+      stuckConsumerRepairStats.recordStuckConsumerFound();
+
+      AtomicBoolean repairSomeIngestionTask = new AtomicBoolean(false);
+      versionTopicStoreIngestionTaskMapping.forEach((vt, sit) -> {
+        if (!sit.isProducingTopicHealthy()) {
+          LOGGER.info(
```
xunyin8, line 172: The killed ingestion could be a serving version. I think this should at least be a WARN level log if not ERROR. Will the kill also trigger alert for ingestion error or stalled alert? IIUC after the killed SIT requires the oncall to look at it.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/AggKafkaConsumerService.java
```diff
@@ -100,6 +131,64 @@ public void stopInner() throws Exception {
...
+              "The ingestion topics (version topic and/or real-time topic) are not healthy for "
+                  + "store version: {}, will kill the ingestion task to try to unblock shared consumer",
+              vt);
+          /**
+           * The following function call will interrupt all the stuck {@link org.apache.kafka.clients.producer.KafkaProducer#send} call
+           * to non-existing topics.
+           */
+          sit.closeVeniceWriters(false);
+          sit.kill();
```
lluwm, line 181: I checked `KafkaStoreIngestionService.killConsumptionTask` and there seems to have some clean up work after calling `StoreIngestionTask.kill`. Are those cleanup steps necessary for here as well?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/AggKafkaConsumerService.java
gaojieliu: Good idea, today, kill doesn't emit any metrics and as you mentioned, at least we should emit metrics to track stale replicas for the current version.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/AggKafkaConsumerService.java
lluwm: Does `versionTopicStoreIngestionTaskMapping` contain all possible SIT on this server? Just to think out loud, do we need to cross-check that the SIT that's going to be killed does indeed have a topic partition assigned to the slow `KafkaConsumer` (or `ConsumptionTask`), which causes `getMaxElapsedTimeMSSinceLastPollInConsumerPool` to alarm in the first place. My worry is that if isProducingTopicHealthy? or in general, it is okay to kill any SIT as long as it doesn't pass the  `isProducingTopicHealthy` test

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/AggKafkaConsumerService.java
xunyin8: We have some stalled ingestion metric already. If they don't fire after the SIT is killed then those metrics are also not working...

FelixGV (commented):
Is there any commonality between this and the ingestion repair task @ZacAttack built for the case where a remote Kafka cluster was down? Is there any opportunity to combine these two, or are they completely distinct?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/AggKafkaConsumerService.java
gaojieliu: This is a good question.
`KafkaStoreIngestionService.killConsumptionTask` does more cleanup than the `kill`, such as cleanup locks and handle to the ingestion task, and I should potentially call `KafkaStoreIngestionService#killConsumptionTask` instead.

`versionTopicStoreIngestionTaskMapping` contains all the live `SIT` on the same server.

Regarding the following:
```
Just to think out loud, do we need to cross-check that the SIT that's going to be killed does indeed have a topic partition assigned to the slow KafkaConsumer (or ConsumptionTask), which causes getMaxElapsedTimeMSSinceLastPollInConsumerPool to alarm in the first place. My worry is that if isProducingTopicHealthy? or in general, it is okay to kill any SIT as long as it doesn't pass the isProducingTopicHealthy test
```
I thought about it during the implementation, and checking the attached SITs for a particular stuck consumer is more complex because with partition-wise shared consumers, one SIT can be assigned to multiple consumers, and in consumer-level, there is no concept of `SIT`, but `ConsumedDataReceiver`.

In the high-level, I think the ingestion task should be killed if the producing topics are deleted, such as RT or VT, and I guess you might be worrying that this can potentially trigger some edge cases, such as `TopicManager#containsTopic` return `false`, but the topic does exist somehow.

I am thinking adding some retries into the function: `isProducingTopicHealthy` to make it more solid.

gaojieliu:
> Is there any commonality between this and the ingestion repair task @ZacAttack built for the case where a remote Kafka cluster was down? Is there any opportunity to combine these two, or are they completely distinct?

Based on what you said, these two are quite different.

This PR will kill the ingestion task if the producing topics are deleted and it only check the local fabric, and if the local producing topic don't exist, it will kill the whole task.

Regarding the resilience improvement to tolerate some remote Kafka cluster failure, I assume the detection can be different (maybe overlapping in some way), the local StoreIngestionTask needs to carry on.

BTW, there are two places remote Kafka clusters could kick in:
1. Remote consumption.
2. Lag measurement.


And this PR is mainly focusing on the producing to local Kafka fabric, which can stuck if the topic disappeared in the middle.

@ZacAttack 
Can you chime in here to share more details about the task you are working on?
Maybe I had some misunderstanding.

FelixGV:
> > Is there any commonality between this and the ingestion repair task @ZacAttack built for the case where a remote Kafka cluster was down? Is there any opportunity to combine these two, or are they completely distinct?
> Based on what you said, these two are quite different.

Yes they have different purposes but I mean in the sense that they are background tasks that periodically check some condition and potentially take some corrective action. It may be desirable to have a single "background task framework" for these kinds of activities, rather than spin up a bunch of bespoke thread pools and stuff for each individual need.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -3777,4 +3788,25 @@ protected boolean shouldUpdateUpstreamOffset(PubSubMessage<KafkaKey, KafkaMessag
...
+   * 2. Whether the real-time topic in local colo exists or not for hybrid store.
+   */
+  public boolean isProducingTopicHealthy() {
+    if (isDaVinciClient) {
+      /**
+       * DaVinci doesn't produce to any topics.
+       */
+      return true;
+    }
+    if (!topicManagerRepository.getTopicManager().containsTopic(this.versionTopic)) {
```
majisourav99, line 3804: Is it better to have a timestamped based retries here? like if we cannot find the topic for over some period of checks then only declare its unhealthy? nit: also can we rename to `isProducingTopicExists`

gaojieliu: Yeah, I replied with the similar idea to @lluwm 's comment

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/AggKafkaConsumerService.java
```diff
@@ -100,6 +141,104 @@ public void stopInner() throws Exception {
...
+        for (Map.Entry<String, StoreIngestionTask> entry: versionTopicIngestionTaskMappingForNonExistingTopic
+            .entrySet()) {
+          String versionTopic = entry.getKey();
+          StoreIngestionTask sit = entry.getValue();
+          try {
+            if (sit.isProducingVersionTopicHealthy()) {
+              /**
+               * If the version topic becomes available after retries, remove it from the tracking map.
+               */
+              versionTopicIngestionTaskMappingForNonExistingTopic.remove(versionTopic);
```
majisourav99, line 210: Is this even allowed ? wont it throw `ConcurrentModificationException` ?

gaojieliu: This is allowed and I do have a test to cover this case.

gaojieliu: Good catch and fixed it in the latest PR.

lluwm (approved):
Thanks for the fix, @gaojieliu! LGTM!

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/AggKafkaConsumerService.java
```diff
@@ -100,6 +134,106 @@ public void stopInner() throws Exception {
...
+         * The following function call will interrupt all the stuck {@link org.apache.kafka.clients.producer.KafkaProducer#send} call
+         * to non-existing topics.
+         */
+        sit.closeVeniceWriters(false);
+        killIngestionTaskRunnable.accept(vt);
+        repairSomeIngestionTask.set(true);
+        stuckConsumerRepairStats.recordIngestionTaskRepair();
+      });
+      if (!repairSomeIngestionTask.get()) {
+        LOGGER.error(
```
majisourav99, line 232: curious, what this check for ? It means we found some stuck consumer, but no unhealthy topic ? should we init `repairSomeIngestionTask` to false only if `versionTopicIngestionTaskMappingForNonExistingTopic` size is > 0 ?
also nit: `repairSomeIngestionTask` rename to isRepairedSomeIngestionTask to indicate its a boolean.

gaojieliu:
```
It means we found some stuck consumer, but no unhealthy topic
```
We want to track this case via metric and logging.

Regarding other minor comments, I can pick it up in another PR since perf diff is minor.

PR #758, opened 2023-11-17
[server][controller] Handle exceptions from KafkaConumer::ParitionsFor API

sushantmane:
## Handle exceptions from KafkaConumer::ParitionsFor API

Resolves #XXX

## How was this PR tested?

CI

### DIFF
--- a/internal/venice-common/src/main/java/com/linkedin/venice/pubsub/adapter/kafka/consumer/ApacheKafkaConsumerAdapter.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/pubsub/adapter/kafka/consumer/ApacheKafkaConsumerAdapter.java
@@ -129,7 +129,8 @@ public class ApacheKafkaConsumerAdapter implements PubSubConsumerAdapter {
         lastReadOffset);
   }
 
-  private boolean isValidTopicPartition(PubSubTopicPartition pubSubTopicPartition) {
+  // visible for testing
+  boolean isValidTopicPartition(PubSubTopicPartition pubSubTopicPartition) {
     if (pubSubTopicPartition == null) {
       throw new IllegalArgumentException("PubSubTopicPartition cannot be null");
     }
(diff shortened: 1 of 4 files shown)


internal/venice-common/src/main/java/com/linkedin/venice/pubsub/adapter/kafka/consumer/ApacheKafkaConsumerAdapter.java
```diff
@@ -103,7 +103,7 @@ public void subscribe(PubSubTopicPartition pubSubTopicPartition, long lastReadOf
     }
 
     // Check if the topic-partition exists
-    if (!isValidTopicPartition(pubSubTopicPartition)) {
+    if (config.shouldCheckTopicExistenceBeforeConsuming() && !isValidTopicPartition(pubSubTopicPartition)) {
```
huangminchn, line 106: I thought we cannot skip this check?

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/adapter/kafka/consumer/ApacheKafkaConsumerAdapter.java
```diff
@@ -149,6 +149,13 @@ private boolean isValidTopicPartition(PubSubTopicPartition pubSubTopicPartition)
         throw new PubSubClientException(
             "Interrupted while waiting for validation of topic-partition: " + pubSubTopicPartition,
             e);
+      } catch (Exception e) {
```
huangminchn, line 152: Can we only handle Kafka RetriableException here?

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/adapter/kafka/consumer/ApacheKafkaConsumerAdapter.java
```diff
@@ -468,7 +475,13 @@ public Long endOffset(PubSubTopicPartition pubSubTopicPartition) {
    */
   @Override
   public List<PubSubTopicPartitionInfo> partitionsFor(PubSubTopic topic) {
-    List<PartitionInfo> partitionInfos = this.kafkaConsumer.partitionsFor(topic.getName());
+    List<PartitionInfo> partitionInfos = null;
+    try {
+      partitionInfos = this.kafkaConsumer.partitionsFor(topic.getName());
+    } catch (Exception e) {
+      LOGGER.error("Exception thrown when attempting to get partitions for topic: {}", topic, e);
+    }
```
huangminchn, line 483: I don't want us to swallow InterruptException; can we capture Kafka exception only?

huangminchn (changes requested):

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/adapter/kafka/consumer/ApacheKafkaConsumerAdapter.java
sushantmane: Ideally we should not. Added this check so that if we need to revert back to Kafka clients and want to have old behavior due to mitigate unknown issues

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/adapter/kafka/consumer/ApacheKafkaConsumerAdapter.java
sushantmane: good point

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/adapter/kafka/consumer/ApacheKafkaConsumerAdapter.java
sushantmane: done

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/adapter/kafka/consumer/ApacheKafkaConsumerAdapter.java
huangminchn: I see.

PR #760, opened 2023-11-17
[controller][router] Safe backup version deletion

majisourav99:
##  Safe backup version deletion

Venice backup version deletion currently oblivious to the state of version being served by the router. If due to some issues router did not switch to the latest version, the controller might delete versions which are being read, which would cause catastrophic failures of client requests.  

This PR validates 
1. The current version state from every router and every server before proceeding to delete the version.  A new router end point is added to server the in memory current version by that router.
2.  A new server metadata fetch endpoint added to query the current version for every server.

TODO: make all version deletion in controller async, ie early delete backup version and repush deleting current version after push completion to go through this `StoreBackupVersionCleanupService` .

## How was this PR tested?

CI test

### DIFF
--- a/internal/venice-common/src/main/java/com/linkedin/venice/helix/ZkRoutersClusterManager.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/helix/ZkRoutersClusterManager.java
@@ -2,6 +2,7 @@ package com.linkedin.venice.helix;
 
 import com.linkedin.venice.VeniceResource;
 import com.linkedin.venice.exceptions.VeniceException;
+import com.linkedin.venice.meta.Instance;
 import com.linkedin.venice.meta.RoutersClusterConfig;
 import com.linkedin.venice.meta.RoutersClusterManager;
 import com.linkedin.venice.utils.HelixUtils;
@@ -9,6 +10,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.stream.Collectors;
 import org.apache.helix.AccessOption;
 import org.apache.helix.manager.zk.ZkBaseDataAccessor;
 import org.apache.helix.zookeeper.impl.client.ZkClient;
@@ -45,7 +47,7 @@ public class ZkRoutersClusterManager
 
   private final CachedResourceZkStateListener zkStateListener;
 
-  private final Set<String> liveRouterInstanceSet;
+  private final Set<Instance> liveRouterInstanceSet;
 
   public ZkRoutersClusterManager(
       ZkClient zkClient,
@@ -81,7 +83,12 @@ public class ZkRoutersClusterManager
     routersClusterConfig = newRoutersClusterConfig;
     zkClient.subscribeStateChanges(zkStateListener);
     // force a live router count update
-    liveRouterInstanceSet.addAll(zkClient.getChildren(getRouterRootPath()));
+    liveRouterInstanceSet.clear();
+    liveRouterInstanceSet.addAll(
+        zkClient.getChildren(getRouterRootPath())
+            .stream()
+            .map(HelixUtils::getInstanceFromHelixInstanceName)
+            .collect(Collectors.toSet()));
     changeLiveRouterCount(liveRouterInstanceSet.size());
     LOGGER.info("Refresh finished for cluster {}'s {}.", clusterName, getClass().getSimpleName());
   }
@@ -94,7 +101,7 @@ public class ZkRoutersClusterManager
     zkClient.unsubscribeStateChanges(zkStateListener);
   }
 
-  public Set<String> getLiveRouterInstances() {
+  public Set<Instance> getLiveRouterInstances() {
     return liveRouterInstanceSet;
   }
 
@@ -107,8 +114,8 @@ public class ZkRoutersClusterManager
     try {
       zkClient.createEphemeral(getRouterPath(instanceId));
       LOGGER.info("Add router: {} into live routers.", instanceId);
-      changeLiveRouterCount(zkClient.getChildren(getRouterRootPath()).size());
-      liveRouterInstanceSet.add(instanceId);
+      liveRouterInstanceSet.add(HelixUtils.getInstanceFromHelixInstanceName(instanceId));
+      changeLiveRouterCount(liveRouterInstanceSet.size());
     } catch (ZkNoNodeException e) {
       // For a new cluster, the path for routers might not be created, try to create it.
       try {
@@ -135,8 +142,8 @@ public class ZkRoutersClusterManager
           "Error when deleting router " + instanceId + " from zk path " + getRouterPath(instanceId),
           e);
     }
-    changeLiveRouterCount(zkClient.getChildren(getRouterRootPath()).size());
-    liveRouterInstanceSet.remove(instanceId);
+    liveRouterInstanceSet.remove(HelixUtils.getInstanceFromHelixInstanceName(instanceId));
+    changeLiveRouterCount(liveRouterInstanceSet.size());
     LOGGER.info("Removed router {} from live routers temporarily.", instanceId);
   }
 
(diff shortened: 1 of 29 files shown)


internal/venice-common/src/main/java/com/linkedin/venice/helix/ZkRoutersClusterManager.java
```diff
@@ -95,6 +95,11 @@ public void clear() {
     zkClient.unsubscribeStateChanges(zkStateListener);
   }
 
+  public List<String> getLiveRouterInstances() {
+    List<String> list = zkClient.getChildren(getRouterRootPath());
```
gaojieliu, line 99: In theory, the live Router instances can be cached since this class implements `IZkChildListener`, so that we don't need to talk to ZK directly for this.

services/venice-controller/src/main/java/com/linkedin/venice/controller/StoreBackupVersionCleanupService.java
```diff
@@ -39,8 +64,12 @@ public class StoreBackupVersionCleanupService extends AbstractVeniceService {
   private final Thread cleanupThread;
   private final long sleepInterval;
   private final long defaultBackupVersionRetentionMs;
+
+  private final Map<String, String> urlMap = new HashMap<>();
   private final AtomicBoolean stop = new AtomicBoolean(false);
 
+  private final CloseableHttpAsyncClient storeClient;
```
gaojieliu, line 71: `storeClient` name is confusing since it is not going to talk to any Venice store.
Maybe just call it `httpClient` instead?

services/venice-controller/src/main/java/com/linkedin/venice/controller/StoreBackupVersionCleanupService.java
```diff
@@ -91,6 +125,83 @@ protected static boolean whetherStoreReadyToBeCleanup(Store store, long defaultB
...
+        .stream()
+        .map(
+            instance -> urlMap
+                .computeIfAbsent(instance, k -> HelixUtils.getInstanceFromHelixInstanceName(k).getUrl(true)))
+        .collect(Collectors.toList());
+
+    for (String routerURL: urlList) {
+      try {
+        HttpGet routerRequest = new HttpGet(routerURL + "/" + TYPE_CURRENT_VERSION + "/" + store.getName());
+        HttpResponse response = storeClient.execute(routerRequest, null).get();
```
gaojieliu, line 141: Let us specify a timeout here to avoid infinite wait?

services/venice-controller/src/main/java/com/linkedin/venice/controller/StoreBackupVersionCleanupService.java
```diff
@@ -91,6 +125,83 @@ protected static boolean whetherStoreReadyToBeCleanup(Store store, long defaultB
...
+              instance,
+              store.getName());
+          return false;
+        }
+        byte[] responseBody;
+        try (InputStream bodyStream = response.getEntity().getContent()) {
+          responseBody = IOUtils.toByteArray(bodyStream);
+        }
+        RecordDeserializer<MetadataResponseRecord> metadataResponseRecordRecordDeserializer =
+            SerializerDeserializerFactory.getAvroGenericDeserializer(MetadataResponseRecord.SCHEMA$);
```
gaojieliu, line 188: `MetadataResponseRecord` is evolvable, right?
Does the response header contain the schema id?

services/venice-controller/src/main/java/com/linkedin/venice/controller/StoreBackupVersionCleanupService.java
```diff
@@ -91,6 +125,83 @@ protected static boolean whetherStoreReadyToBeCleanup(Store store, long defaultB
...
+        GenericRecord metadataResponse = metadataResponseRecordRecordDeserializer.deserialize(responseBody);
+
+        MetadataResponseRecord metadataResponseRecord =
```
gaojieliu, line 191: Interesting..
Why can't we use the deserialized record directly?

services/venice-controller/src/main/java/com/linkedin/venice/controller/StoreBackupVersionCleanupService.java
gaojieliu: To make it simple, maybe we can expose a new endpoint in server as well as the one in Router and it will be in json format instead of Avro.

gaojieliu (commented):
Since you are working on this feature, can you add a new metric in Router/Server to track the current version per store?
In case, there is an issue happened, we will be able to find it out via metrics quickly.

gaojieliu:
Shall we add a new feature flag to control this logic in case it doesn't work properly?
Also shall we emit some metrics to track backup version deletion behavior?

services/venice-controller/src/main/java/com/linkedin/venice/controller/StoreBackupVersionCleanupService.java
majisourav99: Sure

majisourav99:
> Since you are working on this feature, can you add a new metric in Router/Server to track the current version per store? In case, there is an issue happened, we will be able to find it out via metrics quickly.

Sure, I will add that in a separate PR.

internal/venice-common/src/main/java/com/linkedin/venice/helix/ZkRoutersClusterManager.java
majisourav99: added a cache for router instances.

services/venice-controller/src/main/java/com/linkedin/venice/controller/StoreBackupVersionCleanupService.java
```diff
@@ -60,26 +86,35 @@ protected StoreBackupVersionCleanupService(
     this.sleepInterval = TimeUnit.MINUTES.toMillis(5);
     this.defaultBackupVersionRetentionMs = multiClusterConfig.getBackupVersionDefaultRetentionMs();
     this.time = time;
+    this.httpAsyncClient = HttpAsyncClients.custom()
```
gaojieliu, line 89: Shall we only spin up the http client when this feature is enabled?

services/venice-controller/src/main/java/com/linkedin/venice/controller/StoreBackupVersionCleanupService.java
```diff
@@ -104,6 +211,14 @@ protected boolean cleanupBackupVersion(Store store, String clusterName) {
     List<Version> versions = store.getVersions();
     List<Version> readyToBeRemovedVersions = new ArrayList<>();
     int currentVersion = store.getCurrentVersion();
+
+    // Do not delete version unless all routers and all servers are on same current version
+    if (multiClusterConfig.getControllerConfig(clusterName).isBackupVersionMetadataFetchBasedCleanupEnabled()
+        && (!validateAllRouterOnCurrentVersion(store, clusterName, currentVersion)
+            || !validateAllServerOnCurrentVersion(store, clusterName, currentVersion))) {
+      return false;
```
gaojieliu, line 219: Shall we emit a metric to track this behavior since it is unexpected?

services/venice-controller/src/main/java/com/linkedin/venice/controller/StoreBackupVersionCleanupService.java
majisourav99: Currently the config is multi-cluster. So spinup could not be controlled for all unless we make it a cluster->client map or single config. Do you think spinning up a client is expensie?

services/venice-controller/src/main/java/com/linkedin/venice/controller/StoreBackupVersionCleanupService.java
majisourav99: Sure, added a metric to track skipped version deletion.

services/venice-controller/src/main/java/com/linkedin/venice/controller/StoreBackupVersionCleanupService.java
gaojieliu: If this service is shared by multiple clusters, it is fine to spin it up regardless.
The cost should be low when not being used.

internal/venice-common/src/main/java/com/linkedin/venice/helix/ZkRoutersClusterManager.java
```diff
@@ -281,10 +293,13 @@ protected void triggerRouterClusterConfigChangedEvent(RoutersClusterConfig newCo
    * will be executed to process the event.
    */
   @Override
-  public void handleChildChange(String parentPath, List<String> currentChilds) throws Exception {
+  public void handleChildChange(String parentPath, List<String> currentChildren) throws Exception {
     int oldLiveRouterCount = liveRouterCount;
-    changeLiveRouterCount(currentChilds.size());
-    LOGGER.info("Live router count has been changed from: {} to: {}.", oldLiveRouterCount, currentChilds.size());
+    liveRouterInstanceSet.clear();
+    liveRouterInstanceSet
```
gaojieliu, line 299: `clear` and `addAll` can be risky when some user is trying to read all the live Router instances. 
Why not just swap it instead?
Essentially, we can create a new instance set and assign it to `livceRouterInstanceSet` once it is completed.

gaojieliu (approved):

PR #761, opened 2023-11-17
[server][controller] Disable topic-existence check by default and handle exceptions from KafkaConumer::paritionsFor API

sushantmane:
## Disable topic-existence check by default and handle exceptions from KafkaConumer::paritionsFor API 
-  Catch exceptions from `KafkaConsumer::paritionsFor` API
- Add a config to enable/disable topic-partition check before subscription action


FelixGV (approved):
Thanks!

sushantmane:
Thanks @FelixGV for the review!

PR #762, opened 2023-11-17
[server] Fix regression in new push status store write schema retrieval logic

sixpluszero:
## [server] Fix regression in new push status store write schema retrieval logic

When server is deployed in a standalone fashion, the cluster does not have 
** venice_system_store_PUSH_STATUS_SYSTEM_SCHEMA_STORE**
So the new push status store writer initialization logic will fail.
This PR adds an exception catch logic and fall back to old fashion.

## How was this PR tested?

Will test in PCL testing.

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/HelixParticipationService.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/HelixParticipationService.java
@@ -16,7 +16,6 @@ import com.linkedin.davinci.storage.StorageMetadataService;
 import com.linkedin.davinci.storage.StorageService;
 import com.linkedin.venice.common.VeniceSystemStoreType;
 import com.linkedin.venice.exceptions.VeniceException;
-import com.linkedin.venice.exceptions.VeniceNoStoreException;
 import com.linkedin.venice.helix.HelixAdapterSerializer;
 import com.linkedin.venice.helix.HelixInstanceConverter;
 import com.linkedin.venice.helix.HelixPartitionStatusAccessor;
@@ -321,7 +320,7 @@ public class HelixParticipationService extends AbstractVeniceService
       valueSchemaEntry = helixReadOnlySchemaRepository.getSupersetOrLatestValueSchema(dummyPushStatusStoreName);
       updateSchemaEntry =
           helixReadOnlySchemaRepository.getLatestDerivedSchema(dummyPushStatusStoreName, valueSchemaEntry.getId());
-    } catch (VeniceNoStoreException e) {
+    } catch (VeniceException e) {
       LOGGER.warn(
           "ZK-shared system store is not available in current environment, will fall back to last known protocol version instead.");
       int valueSchemaId = AvroProtocolDefinition.PUSH_STATUS_SYSTEM_SCHEMA_STORE.getCurrentProtocolVersion();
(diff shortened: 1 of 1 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/HelixParticipationService.java
```diff
@@ -310,11 +314,24 @@ private void asyncStart() {
...
+          helixReadOnlySchemaRepository.getLatestDerivedSchema(dummyPushStatusStoreName, valueSchemaEntry.getId());
+    } catch (VeniceNoStoreException e) {
+      LOGGER.warn(
+          "ZK-shared system store is not available in current environment, will fall back to last known protocol version instead.");
+      int valueSchemaId = AvroProtocolDefinition.PUSH_STATUS_SYSTEM_SCHEMA_STORE.getCurrentProtocolVersion();
+      Schema valueSchema = AvroProtocolDefinition.PUSH_STATUS_SYSTEM_SCHEMA_STORE.getCurrentProtocolVersionSchema();
+      valueSchemaEntry = new SchemaEntry(valueSchemaId, valueSchema);
+      updateSchemaEntry = new DerivedSchemaEntry(
+          valueSchemaId,
+          1,
```
FelixGV, line 332: Is this the WC protocol version? Is it safe to hardcode this?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/HelixParticipationService.java
```diff
@@ -310,11 +314,24 @@ private void asyncStart() {
...
-    DerivedSchemaEntry updateSchemaEntry =
-        helixReadOnlySchemaRepository.getLatestDerivedSchema(dummyPushStatusStoreName, valueSchemaEntry.getId());
+    SchemaEntry valueSchemaEntry;
+    DerivedSchemaEntry updateSchemaEntry;
+    try {
+      String dummyPushStatusStoreName = VeniceSystemStoreType.DAVINCI_PUSH_STATUS_STORE.getSystemStoreName("dummy");
+      valueSchemaEntry = helixReadOnlySchemaRepository.getSupersetOrLatestValueSchema(dummyPushStatusStoreName);
+      updateSchemaEntry =
+          helixReadOnlySchemaRepository.getLatestDerivedSchema(dummyPushStatusStoreName, valueSchemaEntry.getId());
+    } catch (VeniceNoStoreException e) {
```
majisourav99, line 324: Should we try to catch other exception here? I see `VeniceException` is also thrown from above code.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/HelixParticipationService.java
sixpluszero: We were using this: veniceProperties.getInt(PUSH_STATUS_STORE_DERIVED_SCHEMA_ID, 1) actually and we never set the value so it was 1 all the time.

I believe we should never hit the fall back case when we really need to use push status store. Because the ZK shared store is not in the cluster, that means any r/w to push status store should fail, as user push status store should not be there.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/HelixParticipationService.java
sixpluszero: Ok. I can change to VeniceException.

majisourav99 (approved):
LGTM. Thanks for the quick fix.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/HelixParticipationService.java
FelixGV: Maybe we should disable the subsequent write instead...

PR #764, opened 2023-11-20
[client][server] Emit system store metric only for server

majisourav99:
## Emit system store metric only for server, but not thin/fast clients

System store metrics are useful only in server side. This PR prevents emission of meta system store metrics in DaVinci/thin clients.

## How was this PR tested?

CI


huangminchn (approved):
Thanks Sourav!

PR #766, opened 2023-11-20
[vpj] Use source grid fabric for target region fabric

majisourav99:
## [vpj] Use source grid fabric for target region fabric

Currently when target region push is enabled, it picks the default NR source fabric as the target region. When source fabric region and default native replication source are different, there would be 2 issues for target region push.

1. The source VT would be created in default NR region which then will be replayed back to the source grid fabric VT, thus data would be making double hop.
2. The SOP for repush will be missing in source grid fabric as it was created only in default NR fabric, thus leading to stuck push.

This PR resolves the above.

### DIFF
--- a/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
+++ b/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
@@ -2488,7 +2488,7 @@ public class VenicePushJob implements AutoCloseable {
     if (jobSetting.isTargetedRegionPushEnabled && jobSetting.targetedRegions == null) {
       // only override the targeted regions if it is not set and it is a single region push
       // use source grid fabric as target region to reduce data hop, else use default NR source
-      if (StringUtils.isEmpty(jobSetting.sourceGridFabric)) {
+      if (!StringUtils.isEmpty(jobSetting.sourceGridFabric)) {
         jobSetting.targetedRegions = jobSetting.sourceGridFabric;
       } else {
         jobSetting.targetedRegions = storeResponse.getStore().getNativeReplicationSourceFabric();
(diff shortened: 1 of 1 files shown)


nisargthakkar (approved):

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
```diff
@@ -2487,7 +2487,12 @@ private StoreSetting validateStoreSettingAndPopulate(ControllerClient controller
 
     if (jobSetting.isTargetedRegionPushEnabled && jobSetting.targetedRegions == null) {
       // only override the targeted regions if it is not set and it is a single region push
-      jobSetting.targetedRegions = storeResponse.getStore().getNativeReplicationSourceFabric();
+      // use source grid fabric as target region to reduce data hop, else use default NR source
+      if (!StringUtils.isEmpty(jobSetting.sourceGridFabric)) {
+        jobSetting.targetedRegions = jobSetting.sourceGridFabric;
+      } else {
+        jobSetting.targetedRegions = storeResponse.getStore().getNativeReplicationSourceFabric();
+      }
```
shuhui-liu, line 2495: Thanks for the fix. My review is a bit late, PR is already merged...This works when user does not specify target region. But if user specifies the target region which is different with source grid fabric, the repush step will still not complete because SOP is missing in source grid fabric. To resolve that, I think we can reset the VeniceWriter before running the repush job.

majisourav99: Thanks for your review. Yes if user specifies a region for target colo push, this would not work. The target colo region option when it does not match with source grid would be inefficient way to push data. We can either error in that case or create a writer using the target region. Let me open a PR for it

PR #769, opened 2023-11-21
[dvc][server] Added safeguard in LFSIT::updateOffsetsFromConsumerRecord

FelixGV:
There seems to be a race condition where the RT offset gets persisted into the VT offset field of the PCS. This change guards against this bookkeeping error, and logs information if it happens.

## How was this PR tested?
CI


sushantmane (approved):
Should we also log current role of replica?

FelixGV:
> Should we also log current role of replica?

It is included in the `PCS::toString` 🤝

PR #771, opened 2023-11-22
[server] ViewWriter use correct original old value fetcher

sixpluszero:
## [server] ViewWriter use correct original old value fetcher

Patch the missing line in original hot fix

## How was this PR tested?

N/A


nisargthakkar (approved):
LGTM

PR #772, opened 2023-11-23
[server] Eliminated some boxing in WC code

FelixGV:
The SortBasedCollectionFieldOpHandler was using List::add instead of
PrimitiveLongList::addPrimitive. This was shown in a profiling to be a
big source of memory allocation, so we are eliminating this boxing in
this commit.

## How was this PR tested?
CI

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/schema/merge/SortBasedCollectionFieldOpHandler.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/schema/merge/SortBasedCollectionFieldOpHandler.java
@@ -128,10 +128,10 @@ public class SortBasedCollectionFieldOpHandler extends CollectionFieldOperationH
     newElements.addAll(toPutList);
 
     // Add elements and timestamps for the collection-merge part.
-    activeElementToTsMap.forEach((activeElement, activeTimestamp) -> {
-      newElements.add(activeElement);
-      newActiveTimestamps.add(activeTimestamp);
-    });
+    for (Map.Entry<Object, Long> entry: activeElementToTsMap.entrySet()) {
+      newElements.add(entry.getKey());
+      newActiveTimestamps.addPrimitive(entry.getValue().longValue());
+    }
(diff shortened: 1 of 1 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/schema/merge/SortBasedCollectionFieldOpHandler.java
```diff
@@ -128,10 +128,10 @@ public UpdateResultStatus handlePutList(
     newElements.addAll(toPutList);
 
     // Add elements and timestamps for the collection-merge part.
-    activeElementToTsMap.forEach((activeElement, activeTimestamp) -> {
-      newElements.add(activeElement);
-      newActiveTimestamps.add(activeTimestamp);
-    });
+    for (Map.Entry<Object, Long> entry: activeElementToTsMap.entrySet()) {
+      newElements.add(entry.getKey());
+      newActiveTimestamps.addPrimitive(entry.getValue().longValue());
```
sixpluszero, line 133: Might be a dumb question but just wanted to confirm: So the entry.getValue() is already Long, that means it was already boxed to Long before. Hmm then why changing to this addPrimitive(long) API could save here?

>  public boolean add(Long o) {
    return addPrimitive(o);
  }

FelixGV: Yeah in that case it shouldn't change anything. There was one specific code path that seemed to account for a large fraction of allocations (25% IIRC!) due to Long boxing. I also changed the other code paths that use the `PrimitiveLongList` for the sake of consistency but the other paths may not have as much impact.

sixpluszero: I see. Thanks for the change!

sixpluszero (approved):

PR #775, opened 2023-11-28
[dvc][common][server][test]leader complete status part 2: DVC and standby completes based on the consumed leader complete status

m-nagarajan:
## Summary
Part 2 for [PR 741](https://github.com/linkedin/venice/pull/741)

1. [`Deadlock` issue from part 1]: In part 1, we made a change that caused the leader replicas to send a heartbeat with leader completed header to VT when they report `COMPLETED`. This was done by the drainer thread using the same producer that originally produced the message that made the leader to be marked `COMPLETED`, which created a circular loop: [`producer -> producer buffer -> VT -> callback -> drainer queue -> producer`]. This led to a deadlock situation when both the producer buffer and the drainer queue were full, and the first message in the drainer queue made the leader report `COMPLETED`. This resulted in the drainer being blocked on the producer buffer being full, and the producer being blocked on the drainer queue being full. To avoid this, we removed the part that sent the heartbeat to VT directly, and instead we set the `LFSIT#lastSendIngestionHeartbeatTimestamp` to 0 such that `maybeSendIngestionHeartbeat()` force sends the HB to RT whenever a leader partition is reported completed apart from the interval based one.
2. Also added another change wrt `amplification factor`: Leader partition reads HB from RT and propagates the heartbeat to all sub partitions' VT
3. In `maybeSendIngestionHeartbeat()`: Don't send heartbeat to non-existing RT topics if amplification factor is enabled resulting in stuck consumption task.
4. Followers reads the header added in heartbeat and updates its state in `PartitionConsumptionState` which will be used when checking whether lag is acceptable or not for both offset based lag and time-based lag for both `DaVinciClient` and `STANDYBY` replicas.
5. Found some issue with stores without AA: In non-AA stores, consumer task is unable to read HB messages from RT topic (though a test consumer can still read it), leading to standby replicas waiting for leader completion state header indefinitely, so disabling this feature for non-AA stores until that issue is resolved.
6. New configs:
`server.leader.complete.state.check.in.follower.enabled` => to enable this feature. Disabled by default and should be enabled only after `Venice tag` `0.4.154` is fully rolled out which supports sending HBs.
`server.leader.complete.state.check.in.follower.valid.interval.ms` => configure the time interval within which the HB should be considered valid. Default 5mins.

## How was this PR tested?
GH CI

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/config/VeniceServerConfig.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/config/VeniceServerConfig.java
@@ -76,7 +76,7 @@ import static com.linkedin.venice.ConfigKeys.SERVER_KAFKA_CONSUMER_OFFSET_COLLEC
 import static com.linkedin.venice.ConfigKeys.SERVER_KAFKA_MAX_POLL_RECORDS;
 import static com.linkedin.venice.ConfigKeys.SERVER_KAFKA_PRODUCER_POOL_SIZE_PER_KAFKA_CLUSTER;
 import static com.linkedin.venice.ConfigKeys.SERVER_LEADER_COMPLETE_STATE_CHECK_IN_FOLLOWER_ENABLED;
-import static com.linkedin.venice.ConfigKeys.SERVER_LEADER_COMPLETE_STATE_CHECK_VALID_INTERVAL_MS;
+import static com.linkedin.venice.ConfigKeys.SERVER_LEADER_COMPLETE_STATE_CHECK_IN_FOLLOWER_VALID_INTERVAL_MS;
 import static com.linkedin.venice.ConfigKeys.SERVER_LEAKED_RESOURCE_CLEANUP_ENABLED;
 import static com.linkedin.venice.ConfigKeys.SERVER_LEAKED_RESOURCE_CLEAN_UP_INTERVAL_IN_MINUTES;
 import static com.linkedin.venice.ConfigKeys.SERVER_LOCAL_CONSUMER_CONFIG_PREFIX;
@@ -451,8 +451,8 @@ public class VeniceServerConfig extends VeniceClusterConfig {
   private final int metaStoreWriterCloseConcurrency;
 
   private final long ingestionHeartbeatIntervalMs;
-  private final boolean leaderCompleteStateCheckEnabled;
-  private final long leaderCompleteStateCheckValidIntervalMs;
+  private final boolean leaderCompleteStateCheckInFollowerEnabled;
+  private final long leaderCompleteStateCheckInFollowerValidIntervalMs;
   private final boolean stuckConsumerRepairEnabled;
   private final int stuckConsumerRepairIntervalSecond;
   private final int stuckConsumerDetectionRepairThresholdSecond;
@@ -760,10 +760,10 @@ public class VeniceServerConfig extends VeniceClusterConfig {
         serverProperties.getInt(SERVER_NON_EXISTING_TOPIC_INGESTION_TASK_KILL_THRESHOLD_SECOND, 15 * 60); // 15 mins
     nonExistingTopicCheckRetryIntervalSecond =
         serverProperties.getInt(SERVER_NON_EXISTING_TOPIC_CHECK_RETRY_INTERNAL_SECOND, 60); // 1min
-    leaderCompleteStateCheckEnabled =
+    leaderCompleteStateCheckInFollowerEnabled =
         serverProperties.getBoolean(SERVER_LEADER_COMPLETE_STATE_CHECK_IN_FOLLOWER_ENABLED, false);
-    leaderCompleteStateCheckValidIntervalMs =
-        serverProperties.getLong(SERVER_LEADER_COMPLETE_STATE_CHECK_VALID_INTERVAL_MS, TimeUnit.MINUTES.toMillis(5));
+    leaderCompleteStateCheckInFollowerValidIntervalMs = serverProperties
+        .getLong(SERVER_LEADER_COMPLETE_STATE_CHECK_IN_FOLLOWER_VALID_INTERVAL_MS, TimeUnit.MINUTES.toMillis(5));
   }
 
   long extractIngestionMemoryLimit(
@@ -1318,12 +1318,12 @@ public class VeniceServerConfig extends VeniceClusterConfig {
     return ingestionHeartbeatIntervalMs;
   }
 
-  public boolean isLeaderCompleteStateCheckEnabled() {
-    return leaderCompleteStateCheckEnabled;
+  public boolean isLeaderCompleteStateCheckInFollowerEnabled() {
+    return leaderCompleteStateCheckInFollowerEnabled;
   }
 
-  public long getLeaderCompleteStateCheckValidIntervalMs() {
-    return leaderCompleteStateCheckValidIntervalMs;
+  public long getLeaderCompleteStateCheckInFollowerValidIntervalMs() {
+    return leaderCompleteStateCheckInFollowerValidIntervalMs;
   }
 
   public boolean isStuckConsumerRepairEnabled() {
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
@@ -1289,7 +1289,7 @@ public class ActiveActiveStoreIngestionTask extends LeaderFollowerStoreIngestion
     boolean isLagAcceptable = offsetLag <= offsetThreshold;
 
     if (isLagAcceptable && isHybridFollower(pcs)) {
-      if (!getServerConfig().isLeaderCompleteStateCheckEnabled()) {
+      if (!getServerConfig().isLeaderCompleteStateCheckInFollowerEnabled()) {
         isLagAcceptable = true;
       } else if (!pcs.isFirstHeartBeatSOSReceived()) {
         // wait for the first HB to know if the leader supports sending LeaderCompleteState or not
@@ -1301,7 +1301,7 @@ public class ActiveActiveStoreIngestionTask extends LeaderFollowerStoreIngestion
         // check if the leader is completed and the last update time was within the configured time
         isLagAcceptable = pcs.isLeaderCompleted()
             && ((System.currentTimeMillis() - pcs.getLastLeaderCompleteStateUpdateInMs()) <= getServerConfig()
-                .getLeaderCompleteStateCheckValidIntervalMs());
+                .getLeaderCompleteStateCheckInFollowerValidIntervalMs());
       }
     }
 
--- a/clients/da-vinci-client/src/test/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTaskTest.java
+++ b/clients/da-vinci-client/src/test/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTaskTest.java
@@ -13,7 +13,7 @@ import static com.linkedin.venice.ConfigKeys.SERVER_DATABASE_CHECKSUM_VERIFICATI
 import static com.linkedin.venice.ConfigKeys.SERVER_ENABLE_LIVE_CONFIG_BASED_KAFKA_THROTTLING;
 import static com.linkedin.venice.ConfigKeys.SERVER_INGESTION_HEARTBEAT_INTERVAL_MS;
 import static com.linkedin.venice.ConfigKeys.SERVER_LEADER_COMPLETE_STATE_CHECK_IN_FOLLOWER_ENABLED;
-import static com.linkedin.venice.ConfigKeys.SERVER_LEADER_COMPLETE_STATE_CHECK_VALID_INTERVAL_MS;
+import static com.linkedin.venice.ConfigKeys.SERVER_LEADER_COMPLETE_STATE_CHECK_IN_FOLLOWER_VALID_INTERVAL_MS;
 import static com.linkedin.venice.ConfigKeys.SERVER_LOCAL_CONSUMER_CONFIG_PREFIX;
 import static com.linkedin.venice.ConfigKeys.SERVER_NUM_SCHEMA_FAST_CLASS_WARMUP;
 import static com.linkedin.venice.ConfigKeys.SERVER_PROMOTION_TO_LEADER_REPLICA_DELAY_SECONDS;
@@ -2496,7 +2496,7 @@ public abstract class StoreIngestionTaskTest {
     propertyBuilder.put(SERVER_LOCAL_CONSUMER_CONFIG_PREFIX, VeniceProperties.empty());
     propertyBuilder.put(SERVER_REMOTE_CONSUMER_CONFIG_PREFIX, VeniceProperties.empty());
     propertyBuilder.put(SERVER_INGESTION_HEARTBEAT_INTERVAL_MS, 1000);
-    propertyBuilder.put(SERVER_LEADER_COMPLETE_STATE_CHECK_VALID_INTERVAL_MS, 1000);
+    propertyBuilder.put(SERVER_LEADER_COMPLETE_STATE_CHECK_IN_FOLLOWER_VALID_INTERVAL_MS, 1000);
     extraProperties.forEach(propertyBuilder::put);
 
     Map<String, Map<String, String>> kafkaClusterMap = new HashMap<>();
@@ -2735,7 +2735,7 @@ public abstract class StoreIngestionTaskTest {
 
     Map<String, Object> extraServerProperties = new HashMap<>();
     extraServerProperties.put(SERVER_INGESTION_HEARTBEAT_INTERVAL_MS, 5000L);
-    extraServerProperties.put(SERVER_LEADER_COMPLETE_STATE_CHECK_VALID_INTERVAL_MS, 5000L);
+    extraServerProperties.put(SERVER_LEADER_COMPLETE_STATE_CHECK_IN_FOLLOWER_VALID_INTERVAL_MS, 5000L);
     extraServerProperties.put(SERVER_LEADER_COMPLETE_STATE_CHECK_IN_FOLLOWER_ENABLED, true);
 
     StoreIngestionTaskFactory ingestionTaskFactory = getIngestionTaskFactoryBuilder(
@@ -3084,7 +3084,7 @@ public abstract class StoreIngestionTaskTest {
 
     Map<String, Object> serverProperties = new HashMap<>();
     serverProperties.put(SERVER_INGESTION_HEARTBEAT_INTERVAL_MS, 5000L);
-    serverProperties.put(SERVER_LEADER_COMPLETE_STATE_CHECK_VALID_INTERVAL_MS, 5000L);
+    serverProperties.put(SERVER_LEADER_COMPLETE_STATE_CHECK_IN_FOLLOWER_VALID_INTERVAL_MS, 5000L);
     serverProperties.put(SERVER_LEADER_COMPLETE_STATE_CHECK_IN_FOLLOWER_ENABLED, leaderCompleteStateCheckEnabled);
 
     StoreIngestionTaskFactory ingestionTaskFactory = getIngestionTaskFactoryBuilder(
--- a/internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
@@ -2032,8 +2032,8 @@ public class ConfigKeys {
    * the follower replica is marked completed based on the old heartbeat messages from
    * a previous leader replica.
    */
-  public static final String SERVER_LEADER_COMPLETE_STATE_CHECK_VALID_INTERVAL_MS =
-      "server.leader.complete.state.check.valid.interval.ms";
+  public static final String SERVER_LEADER_COMPLETE_STATE_CHECK_IN_FOLLOWER_VALID_INTERVAL_MS =
+      "server.leader.complete.state.check.in.follower.valid.interval.ms";
 
   /**
    * Whether to enable stuck consumer repair in Server.
--- a/internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/integration/utils/VeniceServerWrapper.java
+++ b/internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/integration/utils/VeniceServerWrapper.java
@@ -26,7 +26,7 @@ import static com.linkedin.venice.ConfigKeys.SERVER_HTTP2_INBOUND_ENABLED;
 import static com.linkedin.venice.ConfigKeys.SERVER_INGESTION_HEARTBEAT_INTERVAL_MS;
 import static com.linkedin.venice.ConfigKeys.SERVER_INGESTION_ISOLATION_APPLICATION_PORT;
 import static com.linkedin.venice.ConfigKeys.SERVER_INGESTION_ISOLATION_SERVICE_PORT;
-import static com.linkedin.venice.ConfigKeys.SERVER_LEADER_COMPLETE_STATE_CHECK_VALID_INTERVAL_MS;
+import static com.linkedin.venice.ConfigKeys.SERVER_LEADER_COMPLETE_STATE_CHECK_IN_FOLLOWER_VALID_INTERVAL_MS;
 import static com.linkedin.venice.ConfigKeys.SERVER_NETTY_GRACEFUL_SHUTDOWN_PERIOD_SECONDS;
 import static com.linkedin.venice.ConfigKeys.SERVER_PARTITION_GRACEFUL_DROP_DELAY_IN_SECONDS;
 import static com.linkedin.venice.ConfigKeys.SERVER_PROMOTION_TO_LEADER_REPLICA_DELAY_SECONDS;
@@ -252,7 +252,7 @@ public class VeniceServerWrapper extends ProcessWrapper implements MetricsAware 
               PUB_SUB_ADMIN_ADAPTER_FACTORY_CLASS,
               pubSubBrokerWrapper.getPubSubClientsFactory().getAdminAdapterFactory().getClass().getName())
           .put(SERVER_INGESTION_HEARTBEAT_INTERVAL_MS, 5000)
-          .put(SERVER_LEADER_COMPLETE_STATE_CHECK_VALID_INTERVAL_MS, 5000);
+          .put(SERVER_LEADER_COMPLETE_STATE_CHECK_IN_FOLLOWER_VALID_INTERVAL_MS, 5000);
       if (sslToKafka) {
         serverPropsBuilder.put(KAFKA_SECURITY_PROTOCOL, SecurityProtocol.SSL.name);
         serverPropsBuilder.put(KafkaTestUtils.getLocalCommonKafkaSSLConfig(SslUtils.getTlsConfiguration()));


clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -752,6 +758,106 @@ private StoragePartitionConfig getStoragePartitionConfig(
...
+                .setLastLeaderCompleteStateUpdateInMs(consumerRecord.getValue().producerMetadata.messageTimestamp);
+            isLeaderCompleteHeaderFound = true;
+            break; // only interested in this header here
+          }
+        }
+        if (!isLeaderCompleteHeaderFound) {
+          // reset LeaderCompleteState: If a leader originally sent this header but later is rolled back to a version
+          // that doesn't support this header or this TP gets a new leader which doesn't support this header yet.
+          newState = LEADER_COMPLETE_STATE_UNKNOWN;
+        }
```
huangminchn, line 794: So for AA store, when we do a colo by colo rollout, servers with old codes would not produce this header, and thus followers with new codes would be back and forth on leader complete state. In theory, it's okay, right?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -752,6 +758,106 @@ private StoragePartitionConfig getStoragePartitionConfig(
...
+          LOGGER.debug(
+              "LeaderCompleteState for store {} version {} partition {} received from leader: {} and is unchanged from the previous state",
+              storeName,
+              versionNumber,
+              partitionConsumptionState.getPartition(),
+              newState);
+        }
+        if (!partitionConsumptionState.isFirstHeartBeatSOSReceived()) {
+          partitionConsumptionState.setFirstHeartBeatSOSReceived(true);
+        }
```
huangminchn, line 815: So this bool means follower see the HB, but not necessarily the leader complete state header. Just to confirm this is what we want?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -2097,16 +2100,24 @@ protected DelegateConsumerRecordResult delegateConsumerRecord(
...
+                // Factor is configured to be more than 1) with updated LeaderCompleteState header:
+                // Adding the headers during this phase instead of adding it to RT directly simplifies the logic
+                // of how to identify the HB SOS from the correct version or whether the HB SOS is from the local
+                // colo or remote colo, as the header inherited from an incorrect version or remote colos might
+                // provide incorrect information about the support of the header and the leader state.
+                List<Integer> subPartitions =
+                    PartitionUtils.getSubPartitions(partitionConsumptionState.getUserPartition(), amplificationFactor);
+                for (int _subPartition: subPartitions) {
+                  PubSubTopicPartition topicPartition = new PubSubTopicPartitionImpl(getVersionTopic(), _subPartition);
+                  sendIngestionHeartbeat(
```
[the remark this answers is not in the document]
huangminchn: Let's replace "replicate" with "propagate" or "forward"

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -752,6 +758,106 @@ private StoragePartitionConfig getStoragePartitionConfig(
...
+  protected boolean isLagAcceptableForHybridStore(PartitionConsumptionState pcs, long offsetLag, long offsetThreshold) {
+    boolean isLagAcceptable = (offsetLag <= offsetThreshold);
+
+    if (isHybridMode() && (isDaVinciClient || pcs.getLeaderFollowerState().equals(STANDBY))) {
+      if (!isLagAcceptable) {
+        return false;
+      }
+
+      // non AA stores have issues reading HB SOS from the RT leading to the standby replicas waiting for
+      // leader completion state header indefinitely, so disabling it until that issue is resolved
```
huangminchn, line 837: That's because leaders in some colos might be running on old codes? If so, we will have the same problem when doing canary deployment in one colo though.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -752,6 +758,106 @@ private StoragePartitionConfig getStoragePartitionConfig(
...
+      if (!(getServerConfig().isLeaderCompleteStateCheckEnabled() && this.isActiveActiveReplicationEnabled())) {
+        return true;
+      }
+
+      // wait for the first HB to know if the leader supports sending LeaderCompleteState or not
+      if (!pcs.isFirstHeartBeatSOSReceived()) {
+        return false;
+      }
```
huangminchn, line 845: I believe this is the problem for one colo (non-AA store). If the replica on the canary node is follower replica, and the leader is on very old codes which don't even produce HB. This follower replica will never be online?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -752,6 +758,106 @@ private StoragePartitionConfig getStoragePartitionConfig(
...
+
+      // if the leader don't support LeaderCompleteState
+      if (pcs.getLeaderCompleteState().equals(LEADER_COMPLETE_STATE_UNKNOWN)) {
+        return true;
+      }
+
+      // check if the leader is completed and the last update time was within the configured time
+      return (pcs.isLeaderCompleted()
+          && ((System.currentTimeMillis() - pcs.getLastLeaderCompleteStateUpdateInMs()) <= getServerConfig()
+              .getLeaderCompleteStateCheckValidIntervalMs()));
```
huangminchn, line 855: 1 minute is too strict IMHO. I would start with something around 10 minutes.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -1680,13 +1808,15 @@ private void checkConsumptionStateWhenStart(
...
-      // ready-to-server checker.
+      // For leaders: if offset lag is acceptable, report completed directly, otherwise rely on the
+      // normal ready-to-serve checker. Followers call read-to-serve checker as it looks for leader
+      // completion status as well along with lag.
       boolean isCompletedReport = false;
       long offsetLagDeltaRelaxFactor = serverConfig.getOffsetLagDeltaRelaxFactorForFastOnlineTransitionInRestart();
       long previousOffsetLag = newPartitionConsumptionState.getOffsetRecord().getOffsetLag();
-      if (hybridStoreConfig.isPresent() && newPartitionConsumptionState.isEndOfPushReceived()) {
+      if (hybridStoreConfig.isPresent()
+          && newPartitionConsumptionState.getLeaderFollowerState() == LeaderFollowerStateType.LEADER
```
huangminchn, line 1818: This code is run for OFFLINE to STANDBY transition, leader replicas will not have a chance to run this code IMO. Why adding this check?

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/PubSubTopicImpl.java
```diff
@@ -17,7 +17,7 @@ public class PubSubTopicImpl implements PubSubTopic {
    * instances can be gotten from, so that the internal state is computed just once, and then reused for the lifetime
    * of the topic.
    */
-  PubSubTopicImpl(String name) {
+  public PubSubTopicImpl(String name) {
```
huangminchn, line 20: The javaDoc says the constructor should remain as private.

huangminchn (changes requested):
Thanks Manoj!

clients/da-vinci-client/src/main/java/com/linkedin/davinci/config/VeniceServerConfig.java
```diff
@@ -757,6 +760,9 @@ public VeniceServerConfig(VeniceProperties serverProperties, Map<String, Map<Str
         serverProperties.getInt(SERVER_NON_EXISTING_TOPIC_INGESTION_TASK_KILL_THRESHOLD_SECOND, 15 * 60); // 15 mins
     nonExistingTopicCheckRetryIntervalSecond =
         serverProperties.getInt(SERVER_NON_EXISTING_TOPIC_CHECK_RETRY_INTERNAL_SECOND, 60); // 1min
+    leaderCompleteStateCheckEnabled = serverProperties.getBoolean(SERVER_LEADER_COMPLETE_STATE_CHECK_ENABLED, true);
+    leaderCompleteStateCheckValidIntervalMs =
+        serverProperties.getLong(SERVER_LEADER_COMPLETE_STATE_CHECK_VALID_INTERVAL_MS, TimeUnit.MINUTES.toMillis(1));
```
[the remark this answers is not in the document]
m-nagarajan: Yes, to make sure that stale leader completion status is not used by followers to mark them complete.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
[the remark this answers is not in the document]
m-nagarajan: Makes sense, moved it to a function.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
m-nagarajan: This will be needed only until all the servers are running the latest version. During transition, my understanding is that it's okay if old or the new way is used to decide the follower completion status.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
m-nagarajan: In case of source colo of batch push: Leader writing to local VT comes into picture only after leader starts consuming from RT. so, before the leader could send its first message with this header to local VT, follower will think that leader don't have the updated code and it will mark itself as completed once VT is completed. So the solution is to make the hybrid store wait until the first HB is received before making that decision.

m-nagarajan (commented):
Thanks for the review @huangminchn. Replied to some of the questions.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -1896,6 +1900,169 @@ private void recordRegionHybridConsumptionStats(
...
+        consumerRecord,
+        partitionConsumptionState,
+        leaderProducedRecordContext,
+        partition,
+        kafkaUrl,
+        beforeProcessingRecordTimestampNs);
+    LeaderMetadataWrapper leaderMetadataWrapper = new LeaderMetadataWrapper(consumerRecord.getOffset(), kafkaClusterId);
+    List<Integer> subPartitions =
+        PartitionUtils.getSubPartitions(partitionConsumptionState.getUserPartition(), amplificationFactor);
+    for (int _subPartition: subPartitions) {
```
FelixGV, line 2054: Nitpick... I think we prefer not using underscores. In some projects, they use the underscore prefix for class variables (not local variables) but in Venice we don't use underscores for anything.

m-nagarajan: Corrected this. Thanks.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
m-nagarajan: Removed this change.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -3,7 +3,10 @@
 import static com.linkedin.davinci.kafka.consumer.ConsumerActionType.RESET_OFFSET;
 import static com.linkedin.davinci.kafka.consumer.ConsumerActionType.SUBSCRIBE;
 import static com.linkedin.davinci.kafka.consumer.ConsumerActionType.UNSUBSCRIBE;
+import static com.linkedin.davinci.kafka.consumer.LeaderFollowerStateType.STANDBY;
```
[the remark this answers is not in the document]
m-nagarajan: Fixed this.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -3199,35 +3270,12 @@ protected void maybeSendIngestionHeartbeat() {
...
-        try {
-          veniceWriter.get().sendHeartbeat(topicPartition, null, DEFAULT_LEADER_METADATA_WRAPPER);
-        } catch (Exception e) {
-          String errorMessage = String.format(
-              "Failed to send ingestion heartbeat for topic: %s, partition: %s",
-              topicPartition.getPubSubTopic().getName(),
-              topicPartition.getPartitionNumber());
-          LOGGER.error(errorMessage, e);
-        }
+        sendIngestionHeartbeat(
```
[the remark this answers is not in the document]
m-nagarajan: Thats correct, added comments.

m-nagarajan (commented):
Thanks @FelixGV and @huangminchn for the comments. Addressed them.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -1895,6 +1900,169 @@ private void recordRegionHybridConsumptionStats(
...
+
+    return isLagAcceptable;
+  }
+
+  /**
+   * HeartBeat SOS messages carry the leader completion state in the header. This function extracts the leader completion
+   * state from that header and updates the {@param partitionConsumptionState} accordingly. <p>
+   * If there is no leader completion state header, reset the leader completion state to
+   * {@link LeaderCompleteState#LEADER_COMPLETE_STATE_UNKNOWN} as the leader don't know about this header
+   * (using old version of the code) or the leader may have rolled back to a version that doesn't support this header or
```
gaojieliu, line 1976: What if Venice Server rolled back to a version, which doesn't emit HB messages at all, so will the function handle that?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -1895,6 +1900,169 @@ private void recordRegionHybridConsumptionStats(
...
+   * this topic partition gets a new leader which doesn't support this header yet.
+   */
+  protected void getAndUpdateLeaderCompletedState(
+      PubSubMessage<KafkaKey, KafkaMessageEnvelope, Long> consumerRecord,
+      PartitionConsumptionState partitionConsumptionState) {
+    if (isHybridFollower(partitionConsumptionState)) {
+      ControlMessage controlMessage = (ControlMessage) consumerRecord.getValue().payloadUnion;
```
gaojieliu, line 1983: So the passed message is guaranteed to be `ControlMessage`? Shall we have a type check here?

gaojieliu: Another way would be passing `ControlMessage` param directly along with other required info, such as headers.

m-nagarajan: It was called from only one place and there was a check before that for control Message. But modified it to pass the CM param along with other params directly rather than the `consumerRecord`.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
m-nagarajan: This won't handle that and will keep on waiting for the 1st HB. But as all the servers are on versions that sends HB now, should we also handle this scenario?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -1895,6 +1900,169 @@ private void recordRegionHybridConsumptionStats(
...
+      PartitionConsumptionState pcs,
+      long offsetLag,
+      long offsetThreshold,
+      boolean shouldLogLag,
+      boolean isOffsetBasedLag,
+      long latestConsumedProducerTimestamp) {
+    boolean isLagAcceptable = offsetLag <= offsetThreshold;
+
+    if (isLagAcceptable && isHybridFollower(pcs)) {
+      // non AA stores have issues reading HB SOS from the RT leading to the standby replicas waiting for
```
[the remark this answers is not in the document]
m-nagarajan: Added a bit more detailed comment.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -1895,6 +1900,169 @@ private void recordRegionHybridConsumptionStats(
...
+    return isHybridMode() && (isDaVinciClient || partitionConsumptionState.getLeaderFollowerState().equals(STANDBY));
+  }
+
+  /**
+   * Checks whether the lag is acceptable for hybrid stores
+   *
+   * If the instance is a standby or DaVinciClient: Also check if
+   * 1. leaderCompleteStatus header has been received and
+   * 2. leader was completed and
+   * 3. the last update time was within 5 seconds
```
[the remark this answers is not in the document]
m-nagarajan: Yes, I added this comment during the initial changes and later didn't correct it. Modified it now, thanks.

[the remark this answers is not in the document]
huangminchn: Should be 5 minutes by default; yes, it's configurable.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/config/VeniceServerConfig.java
```diff
@@ -757,6 +760,10 @@ public VeniceServerConfig(VeniceProperties serverProperties, Map<String, Map<Str
         serverProperties.getInt(SERVER_NON_EXISTING_TOPIC_INGESTION_TASK_KILL_THRESHOLD_SECOND, 15 * 60); // 15 mins
     nonExistingTopicCheckRetryIntervalSecond =
         serverProperties.getInt(SERVER_NON_EXISTING_TOPIC_CHECK_RETRY_INTERNAL_SECOND, 60); // 1min
+    leaderCompleteStateCheckEnabled = serverProperties.getBoolean(SERVER_LEADER_COMPLETE_STATE_CHECK_ENABLED, true);
```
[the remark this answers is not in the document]
m-nagarajan: This is more comprehensive, so used this now. thanks.

m-nagarajan (commented):
Thanks @gaojieliu for the review. Addressed review comments.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
huangminchn: I thought about it for some time, the idea of writing correct codes still prevail. Maybe this is the case for our prod environment, but maybe not for external users of Venice (if any). If they use to run on older version (without HB code), and suddenly upgrade to this code version on one canary node; my understanding is that followers on that canary node will be stuck forever.

I could think of 3 options:
1. change the default value of config "isLeaderCompleteStateCheckEnabled" to false; update the javaDoc of that config, saying that users should not enable this until release xxx is fully rolled out. Then on our side, we can override this config in venice-backend immediately (so the default for us is true)
2. make a major version bump, explicitly stating that this version is not compatible with previous version.
3. Compute the elapsed time from the time when TOPIC_SWITCH CM is created to now; if the time is still within "getLeaderCompleteStateCheckValidIntervalMs" (reuse this config), return lag not acceptable; otherwise, return "boolean isLagAcceptable"

What do you think?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
huangminchn: SG! Could you please add a log here with redundant log filtering: if HB with header is found but not valid anymore, log the value of "System.currentTimeMillis() - pcs.getLastLeaderCompleteStateUpdateInMs()".

huangminchn (changes requested):
I haven't gone through the entire diff yet, but I want to share what I have so far, so that we can make progress together.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/config/VeniceServerConfig.java
```diff
@@ -760,9 +760,11 @@ public VeniceServerConfig(VeniceProperties serverProperties, Map<String, Map<Str
         serverProperties.getInt(SERVER_NON_EXISTING_TOPIC_INGESTION_TASK_KILL_THRESHOLD_SECOND, 15 * 60); // 15 mins
     nonExistingTopicCheckRetryIntervalSecond =
         serverProperties.getInt(SERVER_NON_EXISTING_TOPIC_CHECK_RETRY_INTERNAL_SECOND, 60); // 1min
-    leaderCompleteStateCheckEnabled = serverProperties.getBoolean(SERVER_LEADER_COMPLETE_STATE_CHECK_ENABLED, true);
+    leaderCompleteStateCheckEnabled =
+        serverProperties.getBoolean(SERVER_LEADER_COMPLETE_STATE_CHECK_IN_FOLLOWER_ENABLED, true);
+    /** As this is used to ignore the stale heartbeats, defaulting to 5 minutes, inline with {@link SERVER_PROMOTION_TO_LEADER_REPLICA_DELAY_SECONDS} */
     leaderCompleteStateCheckValidIntervalMs =
```
huangminchn, line 766: In theory, these two things are not related; my understanding is that when (1) leader check is enabled and (2) follower hasn't received the first HB message from leader, follower doesn't even use this valid interval config; therefore, even though leader takes 30 minutes to start acting as leader, follower will not go into COMPLETE state until it sees the first HB.

However, the leader promotion delay will matter if we decide to pick [option 3 from the other issue being discussed](https://github.com/linkedin/venice/pull/775#discussion_r1419741991), and reuse this invalid interval config for that purpose.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -1896,6 +1902,185 @@ private void recordRegionHybridConsumptionStats(
...
+  protected boolean isHybridFollower(PartitionConsumptionState partitionConsumptionState) {
+    return isHybridMode() && (isDaVinciClient || partitionConsumptionState.getLeaderFollowerState().equals(STANDBY));
+  }
+
+  /**
+   * Checks whether the lag is acceptable for hybrid stores
+   * <p>
+   * If the instance is a standby or DaVinciClient: Also check if <br>
+   * 1. the feature is enabled and is AA enabled: In non AA stores, consumer task is unable to read HB messages from RT
+   *           topic (though a test consumer can still read it), leading to standby replicas waiting for leader completion
```
huangminchn, line 1915: I thought the problem was the leader complete state would be jumping among UNKNOWN, INCOMPLETE and COMPLETE randomly, but seems like that's not the case from this comment. Why is consumer not able to read HB from RT?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -1896,6 +1902,185 @@ private void recordRegionHybridConsumptionStats(
...
+      long offsetLag,
+      long offsetThreshold,
+      boolean shouldLogLag,
+      boolean isOffsetBasedLag,
+      long latestConsumedProducerTimestamp) {
+    boolean isLagAcceptable = offsetLag <= offsetThreshold;
+
+    if (isLagAcceptable && isHybridFollower(pcs)) {
+      if (!(getServerConfig().isLeaderCompleteStateCheckEnabled() && this.isActiveActiveReplicationEnabled())) {
+        isLagAcceptable = true;
```
huangminchn, line 1934: This looks like a bug to me. If AA enable is false (non-AA store), we return lag acceptable immediately?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -1896,6 +1902,185 @@ private void recordRegionHybridConsumptionStats(
...
+      PartitionConsumptionState pcs,
+      long offsetLag,
+      long offsetThreshold,
+      boolean shouldLogLag,
+      boolean isOffsetBasedLag,
+      long latestConsumedProducerTimestamp) {
+    boolean isLagAcceptable = offsetLag <= offsetThreshold;
+
+    if (isLagAcceptable && isHybridFollower(pcs)) {
+      if (!(getServerConfig().isLeaderCompleteStateCheckEnabled() && this.isActiveActiveReplicationEnabled())) {
```
huangminchn, line 1933: We have ActiveActiveStoreIngestionTask; it might be easier to just override this function in AASIT.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
huangminchn: but the codes do match the comment:

> 1. the feature is enabled and is AA enabled: In non AA stores, consumer task is unable to read HB messages from RT

So the HB based approach for follower completion report only work for AA stores?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
huangminchn: Have similar concern: https://github.com/linkedin/venice/pull/775#discussion_r1419741991

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -1896,6 +1902,185 @@ private void recordRegionHybridConsumptionStats(
...
+        kafkaUrl,
+        beforeProcessingRecordTimestampNs);
+    LeaderMetadataWrapper leaderMetadataWrapper = new LeaderMetadataWrapper(consumerRecord.getOffset(), kafkaClusterId);
+    List<Integer> subPartitions =
+        PartitionUtils.getSubPartitions(partitionConsumptionState.getUserPartition(), amplificationFactor);
+    LeaderCompleteState leaderCompleteState =
+        LeaderCompleteState.getLeaderCompleteState(partitionConsumptionState.isCompletionReported());
+    /**
+     * The maximum value between the original producer timestamp and the timestamp when the message is added to the RT topic is used:
+     * This approach addresses scenarios wrt clock drift where the producer's timestamp is consistently delayed by several minutes,
```
huangminchn, line 2063: Cool! But I am curious; is it defensive coding, or did you really encounter this problem in tests somehow?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -3276,6 +3461,20 @@ protected void maybeSendIngestionHeartbeat() {
...
+   * such that {@link #maybeSendIngestionHeartbeat()} will send HB SOS to the respective RT topics
+   * rather than waiting for the timer to send HB SOS.
+   */
+  @Override
+  void reportCompleted(PartitionConsumptionState partitionConsumptionState, boolean forceCompletion) {
+    super.reportCompleted(partitionConsumptionState, forceCompletion);
+    if (isHybridMode() && partitionConsumptionState.getLeaderFollowerState().equals(LEADER)) {
+      // reset lastSendIngestionHeartbeatTimestamp to force sending HB SOS to the respective RT topics.
+      lastSendIngestionHeartbeatTimestamp.set(0);
+    }
```
huangminchn, line 3478: Good idea!

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/PartitionConsumptionState.java
```diff
@@ -193,6 +193,11 @@ public class PartitionConsumptionState {
 
   private LeaderCompleteState leaderCompleteState;
   private long lastLeaderCompleteStateUpdateInMs;
+  /**
+   * In case of hybrid store, wait until the first HB is received, which might have the leader completed header and base
+   * the decision on that. This will be used only during the transition phase to avoid 2 phase deployment and once all
+   * servers are running the latest code, this can be removed.
```
huangminchn, line 199: We can only remove it after doing a major version bump and claiming that it's no longer compatible with old releases.

huangminchn (changes requested):
Thanks Manoj!

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
m-nagarajan: it is defensive for now, discussed in the design doc.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
m-nagarajan: Made this change.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
m-nagarajan: yes that's right. The issue with non AA stores needs to be further debugged and can be added as a new fix on top of this.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
m-nagarajan: Lets go with 1st option as per Min's comment.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
m-nagarajan: its already getting logged inside `checkAndLogIfLagIsAcceptableForHybridStore`

m-nagarajan (commented):
Addressed review comments from @huangminchn

huangminchn (approved):
Thanks a lot for the hard work Manoj!

huangminchn (dismissed):

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
m-nagarajan: Thanks for the suggestions, Min. As per our discussions: dropping 3rd option as it falls back to the old version (not checking HB) if the follower is running slow, which is what we are trying to fix in the first place.  Going with the first option.

huangminchn (approved):

PR #776, opened 2023-11-29
[server] Fixed rare race condition for TokenBucket tryConsume()

xunyin8:
## [server] Fixed rare race condition for TokenBucket tryConsume()

TokenBucket relies on consumption itself to update/refill its tokens. There is a rare race condition that could happen and cause very few consumption requests to be rejected even when the token bucket is refilled and has sufficient tokens. The race is caused by the old implementation of tryConsume()

return noRetryTryConsume(tokensToConsume) || (update() && noRetryTryConsume(tokensToConsume));

This is problematic because a small window where token is empty so it enters update() but the bucket just got updated by another thread so the first nextUpdateTime check will be false and hence returning false from update() and causing the tryConsume() to return false when in fact there are plenty of token available and was just refilled.

An alternative fix would be to check the bucket within update() but we might as well always try to consume again.

## How was this PR tested?
Existing unit and integration test, will war drop if needed

### DIFF
--- a/internal/venice-common/src/main/java/com/linkedin/venice/throttle/TokenBucket.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/throttle/TokenBucket.java
@@ -112,6 +112,7 @@ public class TokenBucket {
   }
 
   public boolean tryConsume(long tokensToConsume) {
+    tokensRequestedSinceLastRefill.getAndAdd(tokensToConsume);
     if (noRetryTryConsume(tokensToConsume)) {
       return true;
     } else {
@@ -121,7 +122,6 @@ public class TokenBucket {
   }
 
   private boolean noRetryTryConsume(long tokensToConsume) {
-    tokensRequestedSinceLastRefill.getAndAdd(tokensToConsume);
     long tokensThatWereAvailable = tokens.getAndAccumulate(tokensToConsume, (existing, toConsume) -> {
       if (toConsume <= existing) { // there are sufficient tokens
         return existing - toConsume;
--- a/internal/venice-common/src/main/java/com/linkedin/venice/throttle/TokenBucket.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/throttle/TokenBucket.java
@@ -71,10 +71,10 @@ public class TokenBucket {
   }
 
   /**
-   *
-   * @return true if tokens may have been added, false if short circuited and no tokens were added
+   * Check and add tokens if conditions are met. Note that token may have been updated by another thread even when the
+   * function is short-circuited. Consumers of the token bucket should always retry.
    */
-  private boolean update() {
+  private void update() {
     if (clock.millis() > nextUpdateTime) {
       synchronized (this) {
         long timeNow = clock.millis();
@@ -99,9 +99,6 @@ public class TokenBucket {
           nextUpdateTime = timeNow + refillIntervalMs;
         }
       }
-      return true;
-    } else {
-      return false;
     }
   }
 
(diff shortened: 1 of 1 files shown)


internal/venice-common/src/main/java/com/linkedin/venice/throttle/TokenBucket.java
```diff
@@ -71,10 +71,10 @@ public TokenBucket(long capacity, long refillAmount, long refillInterval, TimeUn
...
 
   /**
-   *
-   * @return true if tokens may have been added, false if short circuited and no tokens were added
+   * Check and add tokens if conditions are met. Note that token may have been updated by another thread even when the
+   * function is short-circuited. Consumers of the token bucket should always retry.
    */
-  private boolean update() {
+  private void update() {
     if (clock.millis() > nextUpdateTime) {
```
majisourav99, line 78: Looks like `nextUpdateTime` is checked but not safely updated inside.  Put this under the sync-block ?

xunyin8: Synced offline, this fix should be sufficient but there are other alternative like a read-write lock that could be easier to reason about but maybe less performant. wdyt @gaojieliu

internal/venice-common/src/main/java/com/linkedin/venice/throttle/TokenBucket.java
```diff
@@ -115,7 +112,12 @@ public long getStaleTokenCount() {
   }
 
   public boolean tryConsume(long tokensToConsume) {
-    return noRetryTryConsume(tokensToConsume) || (update() && noRetryTryConsume(tokensToConsume));
+    if (noRetryTryConsume(tokensToConsume)) {
```
lluwm, line 115: I don't think it is caused by your change, but it sounds possible that`noRetryTryConsume` could be called 2 times: 

noRetryTryConsume -> false
update
noRetryTryConsume -> true/false.

in this case, `tokensToConsume` can be double counted in `tokensRequestedSinceLastRefill`. I don't know if that's okay or, but just think it loud here.

xunyin8: We don't double count because we only decrement IF we can consume:
```
  private boolean noRetryTryConsume(long tokensToConsume) {
    tokensRequestedSinceLastRefill.getAndAdd(tokensToConsume);
    long tokensThatWereAvailable = tokens.getAndAccumulate(tokensToConsume, (existing, toConsume) -> {
      if (toConsume <= existing) { // there are sufficient tokens
        return existing - toConsume;
      } else {
        return existing; // insufficient tokens, do not consume any
      }
    });
    return tokensToConsume <= tokensThatWereAvailable;
  }
```

lluwm: I mean `tokensRequestedSinceLastRefill` is unconditionally updated here, isn't it? Would it matter?

xunyin8: I see, `tokensRequestedSinceLastRefill` is used to calculate usage ratio. Could use read-write lock to have `update()` return true/false properly so we can avoid extra `noRetryTryConsume` calls when we know no token is updated and we should just give up.

lluwm: Yeah, that should work, or if `noRetryTryConsume` is private and `tryConsume` is its only caller, we can move
`tokensRequestedSinceLastRefill.getAndAdd(tokensToConsume);` into `tryConsume`, so that it will be 1 per tryConsume call? Up to your preference

gaojieliu: @lluwm 's suggestion SGTM.

lluwm:
Thank you @xunyin8 and it looks good to me!

majisourav99 (approved):
LGTM. Probably if we want to refactor the refilling and/or locking we can do it separate.

xunyin8:
> LGTM. Probably if we want to refactor the refilling and/or locking we can do it separate.

Gaojie had similar idea. Refactoring could potentially introduce new edge cases and would be less performant because there will be more synchronization.

PR #777, opened 2023-11-29
[dvc] Enable custom views of Venice data set in Da Vinci client

elijahgrimaldi:
## Summary
This is a work in progress to give users the ability to create different ‘views’ of a given Venice data set without having to either reimplement Venice versioning/bootstrapping or push a transformed copy of a dataset which might already exist in Venice.

## How was this PR tested?

CI

We would like to give users the ability to set up individual application specific data transformation, storage or filtering to specify what Davinci should do as it ingests every record.


clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/DaVinciRecordTransformer.java
```diff
@@ -0,0 +1,47 @@
...
+ * into Da Vinci's local storage (e.g. RocksDB).
+ *
+ * N.B.: The inputs are wrapped inside {@link Lazy} so that if the implementation need not look at
+ * them, the deserialization cost is not paid.
+ *
+ * @param <K> type of the input key
+ * @param <V> type of the input value
+ * @param <O> type of the output value
+ */
+public interface DaVinciRecordTransformer<K, V, O extends GenericRecord> {
```
ZacAttack, line 21: Regarding O extending GenericRecord. This might be too loose.  It seems from the code we require that GenericRecord have "key" and "value", so the structure of the output isn't really up to the user to define.

We probably need some kind of Record class to contain the outputted K/V in order to be sure the code will work (otherwise if the user just renders some random GenericRecord without key or value fields, it'll NPE).

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -3025,6 +3048,11 @@ private int processKafkaDataMessage(
         }
         keyLen = keyBytes.length;
 
+        if (recordTransformer != null) {
+          Lazy<byte[]> lazyKey = Lazy.of(() -> keyBytes);
+          GenericRecord keptRecord = recordTransformer.delete(lazyKey, producedPartition);
```
ZacAttack, line 3053: It seems keptRecord isn't used?  We'll need to carry the output of the key/partition over to removeFromStorageEngine in order for it to properly get cleaned up.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -2999,14 +3004,32 @@ private int processKafkaDataMessage(
...
+          Lazy<byte[]> lazyKey = Lazy.of(() -> keyBytes);
+          Lazy<ByteBuffer> lazyValue = Lazy.of(() -> put.putValue);
+          GenericRecord transformedRecord = recordTransformer.put(lazyKey, lazyValue, producedPartition);
+          ByteBuffer keyByteBuffer = (ByteBuffer) transformedRecord.get("key");
+          byte[] newKeyBytes = new byte[keyByteBuffer.remaining()];
+          keyByteBuffer.get(newKeyBytes);
+          ByteBuffer valueByteBuffer = (ByteBuffer) transformedRecord.get("value");
+          byte[] valueBytes = new byte[valueByteBuffer.remaining()];
+          valueByteBuffer.get(valueBytes);
+          int newProducedPartition = Objects.hash(newKeyBytes, put.putValue);
```
ZacAttack, line 3019: This isn't quite right.  Use VenicePartitioner.getPartitionId(keybytes, numPartitions).  Theres already a VenicePartitioner member of StoreIngestionTask, so should be an easy tweak.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
elijahgrimaldi: I think I may be misunderstanding what the delete method in the recordTransformer should do. My thought process was that the delete method would be called in the recordTransformer which handles the deletion of the key that is seperate (maybe internally or in the outPutCollector we discussed?) from the removeFromStorageEngine function, and is defined to maybe return something if the user wants to hold onto to the record for some reason. So right now I wasn't sure what to do with keptRecord. But now I am thinking that the key/partition needs to be transformed in order to delete it from the right place in the storage engine, and one can choose to not remove it from the storage engine by letting it return the null default value.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -2999,14 +3004,27 @@ private int processKafkaDataMessage(
...
-            // so we need to re-calculate partition.
-            // Followers are not affected since they are always consuming from VTs.
-            producedPartition,
-            keyBytes,
-            put,
-            currentTimeMs);
+
+        // Do transorfmation recompute key, value and partition
+        if (recordTransformer != null) {
+          Lazy<byte[]> lazyKey = Lazy.of(() -> keyBytes);
```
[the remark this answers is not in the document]
ZacAttack: I'll need to poke around.  I'm not sure we'd have the POJO deserializer handy from the ingestion task. We'll probably have to have the user pass one in?  That might make the most sense (as they might be partial to a certain version of the upstream venice schema).

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -3227,6 +3256,46 @@ private void deserializeValue(
     }
   }
 
+  private GenericRecord deserializeKeyAndReturn(
```
ZacAttack, line 3259: I don't think we need two separate functions for key and value.  I think we can just add an argument for SchemaEntry and just invoke it with:

deserializeAvroObjectAndReturn(schemaId, object, record, schemaRepository.getValueSchema(storeName, schemaId)) (and change it to getKeySchema for the other call).

ZacAttack: One other thing, let's a TODO here.  Something I 'think' is the case, this code will not work if the data which is ingested is compressed or chunked.

Now, compression or chunking isn't ALWAYS enabled for a store.  In fact, it's probably in the minority.  But we'll need to do some work here I think to make this work with those features.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/DaVinciRecordTransformer.java
```diff
@@ -0,0 +1,48 @@
...
+
+  Schema getValueOutputSchema();
+
+  /**
+   * @param key to be put
+   * @param value to be put
+   * @param partition the put key/value belong to
+   * @return the object to keep in storage, or null if the put should be skipped
+   */
+  O put(Lazy<K> key, Lazy<V> value, int partition);
```
ZacAttack, line 36: Why do we need the partition here?  I'm not certain that needs to be exposed to the user.

ZacAttack:
Other gotcha, StoreIngestionTaskTest seems to be failing to compile due to the change in signature for LeaderFollowerStoreIngestionTask, can you tweak that as well?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/AvroGenericDaVinciClient.java
```diff
@@ -677,13 +678,14 @@ private void initBackend(
...
-      Optional<ObjectCacheConfig> cacheConfig) {
+      Optional<ObjectCacheConfig> cacheConfig,
+      DaVinciRecordTransformer recordTransformer) {
     synchronized (AvroGenericDaVinciClient.class) {
       if (daVinciBackend == null) {
         logger
             .info("Da Vinci Backend does not exist, creating a new backend for client: " + clientConfig.getStoreName());
         daVinciBackend = new ReferenceCounted<>(
-            new DaVinciBackend(clientConfig, configLoader, managedClients, icProvider, cacheConfig),
+            new DaVinciBackend(clientConfig, configLoader, managedClients, icProvider, cacheConfig, recordTransformer),
```
ZacAttack, line 688: This will always pass null, you need to bring the line at 735 up before this, or just pass in the value of daVinciConfig.getRecordTransformer (and maybe forgo the member variable entirely from this class?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/AvroGenericDaVinciClient.java
```diff
@@ -719,7 +721,7 @@ public synchronized void start() {
     logger.info("Starting client, storeName=" + getStoreName());
     VeniceConfigLoader configLoader = buildVeniceConfig();
     Optional<ObjectCacheConfig> cacheConfig = Optional.ofNullable(daVinciConfig.getCacheConfig());
-    initBackend(clientConfig, configLoader, managedClients, icProvider, cacheConfig);
+    initBackend(clientConfig, configLoader, managedClients, icProvider, cacheConfig, recordTransformer);
```
ZacAttack, line 724: I deleted the last comment, but yeah I think this is the case.  recordTransformer is always null here because we don't set the member until L735.  We need to bump it up, or, just forgo the member variable and always read the config.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -3107,8 +3126,17 @@ private int processKafkaDataMessage(
           delete = ((Delete) leaderProducedRecordContext.getValueUnion());
         }
         keyLen = keyBytes.length;
-
+        // Commenting out, since only supporting value transformation for now deleting a key should be the same
+        // if (recordTransformer != null) {
```
ZacAttack, line 3130: Let's just delete this for now and bring it back when come back to key transformation later.  We'll still have the history here in this PR for later if we ever need to copy it back.

ZacAttack (approved):
Looks good!

PR #782, opened 2023-12-01
Hotfix 0.4.161 stuck consumer metric and repair

gaojieliu:
[server][da-vinci] Construct the MaxElapsedTimeSinceLastPoll Gauge metric with function instead of value (#781)
    
    Constructing Venice Gauge metric with value is pointless; Venice Gauge metric
    can only work with function.

    [server] Introduced a repair service for stuck shared consumer (#757)
    
    * [server] Introduced a repair service for stuck shared consumer
    
    Recently, we noticed that some shared consumers were stuck since they
    were blocked by producing to some non-existing topics, and many other
    store versions got affected since the stuck consumer is being shared
    by many other stores.
    This PR introduces a way to detect and repair stuck consumer for above
    situation.
    In the high level, there will be a thread periodically check whether consumer#poll
    is being invoked regularly or not, and if not, the thread will scan all
    the ingestion tasks to see whether they are producing to any non-existing topics
    or not. And for the store ingestion task, which is talking to some non-existing topic,
    the repair thread will try to close the stuck KafkaProducer and kill the corresponding
    ingestion task.
    This PR also checked the non-existing topics for a configurable duration to tolerate
    tranisent topic metadata propagaton delay.
    
    5 new server configs:
    
    server.stuck.consumer.repair.enabled : default true
    server.stuck.consumer.repair.second : default 1 min
    server.stuck.consumer.repair.threshold.second : default 5 mins
    server.non.existing.topic.ingestion.task.kill.threshold.second: default 15 mins
    server.non.existing.topic.check.retry.interval.second: default 1 min
    
    
    3 new server metrics:
    
    .StuckConsumerRepair--stuck_consumer_found.OccurrenceRate
    .StuckConsumerRepair--ingestion_task_repair.OccurrenceRate
    .StuckConsumerRepair--repair_failure.OccurrenceRate

## How was this PR tested?
CI

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/AggKafkaConsumerService.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/AggKafkaConsumerService.java
@@ -4,6 +4,7 @@ import static com.linkedin.venice.ConfigKeys.KAFKA_BOOTSTRAP_SERVERS;
 
 import com.linkedin.davinci.config.VeniceServerConfig;
 import com.linkedin.davinci.ingestion.consumption.ConsumedDataReceiver;
+import com.linkedin.davinci.stats.StuckConsumerRepairStats;
 import com.linkedin.venice.exceptions.VeniceException;
 import com.linkedin.venice.kafka.TopicManagerRepository;
 import com.linkedin.venice.kafka.protocol.KafkaMessageEnvelope;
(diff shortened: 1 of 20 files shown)


huangminchn (approved):
Thanks Gaojie!

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/AggKafkaConsumerService.java
```diff
@@ -100,6 +134,106 @@ public void stopInner() throws Exception {
...
+          } catch (Exception e) {
+            LOGGER.error("Got exception while checking topic existence for version topic: {}", versionTopic, e);
+          } finally {
+            Utils.sleep(nonExistingTopicRetryIntervalMs);
+          }
+        }
+      }
+
+      AtomicBoolean repairSomeIngestionTask = new AtomicBoolean(false);
+      versionTopicIngestionTaskMappingForNonExistingTopic.forEach((vt, sit) -> {
```
huangminchn, line 217: Nitpick: why not do it inside the same loop? So that we can apply the "nonExistingTopicRetryIntervalMs" wait time for the actual works of killing SIT.

gaojieliu: `nonExistingTopicRetryIntervalMs` is not for killing SIT, but a delay for topic existence check retry.
Killing SITs, which are producing to unknown version topics doesn't need to have any delay.

huangminchn (approved):

PR #783, opened 2023-12-01
[doc] Add PlantUML diagrams for Venice

adamxchen:

Added plantUML codes to the repo.

## How was this PR tested?

https://adamxchen.github.io/venice/docs/dev_guide/venice_write_path


manujose0:
Is there any way to render it for the venicedb.org ?

diagrams/README.md
```diff
@@ -0,0 +1,12 @@
+### Introduction
```
nisargthakkar, line 1: Can we put it in `docs/uml/` and add it as excluded? https://github.com/linkedin/venice/blob/main/docs/_config.yml#L8

adamxchen:
> Is there any way to render it for the venicedb.org ?

Yes, we can do it!

diagrams/README.md
adamxchen: I moved them to `/assets/uml/`

docs/architecture/venice_architecture.md
```diff
@@ -0,0 +1,9 @@
+---
+layout: default
+title: Venice Architecture
+has_children: true
+permalink: /docs/architecture
```
FelixGV, line 5: Hmm... I wonder if this is the right location for this doc. On one hand, I think it may be good to have a high-level architecture section. On the other hand, this particular page you're contributing goes pretty deep into the implementation details and I think it may be geared especially towards Venice developers/contributors, so I wonder if it should be housed under Developer Guides.

WDYT?

docs/architecture/venice_write_path.md
```diff
@@ -0,0 +1,42 @@
...
+| VT    | Version Topic. The version topic is a kafka topic where holds all the data for a given store.               |
+| RT    | Real time Topic. A real time topic is also a kafka topic but it only contains data from nearline producers. |
+| ZK    | Apache ZooKeeper. It's used in multiple Venice components to persist Venice metadata.                       |
+| Helix | Apache Helix. It's used in Venice to manage the cluster and assign partitions to different servers.         |
+| EV    | Helix External View. Venice uses it to represent the state of a server in a leader/follower mode.           |
+| CV    | Helix Customized View. Venice uses it to represent the ingestion state of a server for a given store.       |
+
+### Venice Store Ingestion on server side
+
+In Venice, a table or the dataset is called a store and each store has multiple partitions. Helix assign leaders and
```
FelixGV, line 30: I think it may be good to put "Store" at the top of the terminology table. Then we could mention in the VT row that there are many VTs per Store, and in the RT row that there is one RT per Store. WDYT?

docs/architecture/venice_write_path.md
```diff
@@ -0,0 +1,42 @@
...
+followers status on each partition to different servers. When assignments are done, each server will create multiple
+`StoreIngestionTask`s to handle ingestion for the partitions it's assigned to. The diagram belows describes the data flow of how a 
+store ingestion task's created and ingest data on the server side.
+
+![Server Ingestion Diagram](../assets/images/server_ingestion.svg)
```
FelixGV, line 35: This diagram is fully accurate AFAICT. Thanks a lot for putting it together!

Seeing it makes me want to refactor things a bit because the SIT is a little bit of a God class at this point, and the code would probably be easier to follow if the data flow did not go right then back to the left again. That being said, it is good to document the current state and that diagram succeeds in doing that.

docs/architecture/venice_write_path.md
```diff
@@ -0,0 +1,42 @@
...
+
+### Venice Store Ingestion report 
+
+When ingestion is in different stages, the server reports the ingestion status. The ingestion signal is propagated to 
+various downstream for different business logics. The diagram below describes the signal flow of the ingestion report.
+
+![Server Ingestion Report Diagram](../assets/images/server_ingestion_report.svg)
```
FelixGV, line 42: This second diagram is also accurate, I believe. Although I seem to recall that @xunyin8 recently deleted the code which updates the meta store, since FC is using a different mechanism now... If that is true, maybe that part could be deleted in the diagram as well, to simplify things slightly.

Also, I wonder if there is a way in our docs theme to make images pop out in full size when clicking on them... I managed to do it by right-clicking and loading just the image in another browser tab, but it may be a good usability improvement to just be able to do it with a normal click, since these diagrams are pretty detailed and benefit from being seen in full resolution. It's totally fine to leave that for later though.

docs/assets/uml/server_ingestion.puml
```diff
@@ -0,0 +1,46 @@
...
+collections  StoreBufferDrainers as SBDS
+end box
+database "StorageEngine" as db
+
+helix -> KIS:Issue state transitions \naccording to partition assignments
+KIS -> SITS:Manage lifecycle and \nsend commands for SITs
+
+
+' # leader behavior
+SITS -[#red]> remote:Leaders subscribe to \nsource kafka topics \n(a remote Version Topic or Real-time Topic) \nwhen consumes remotely
```
FelixGV, line 33: s/consumes/consuming/

docs/assets/uml/server_ingestion.puml
```diff
@@ -0,0 +1,46 @@
...
+remote -[#red]> KCTS:Pull msgs from kafka
+KCTS -[#red]> VT:Produce msgs to local VT \nwhen leader consumes remotely
+VT -[#red]> SBDS:Msgs Acked by Kafka \nthen records queued up into drainers
+
+' # follower behavior
+SITS -[#green]> VT:Followers subscribe to local VT
+VT -[#green]> KCTS:Pull msgs from kafka
+KCTS -[#green]> SBDS:Records queued into drainers
+
+SBDS -> SITS:Dequeue the records and invoke methods from SITS to persist
```
FelixGV, line 43: "SITS" looks weird IMHO... Can we say "SIT" or "SITs"?

docs/assets/uml/server_ingestion_report.puml
```diff
@@ -0,0 +1,46 @@
...
+participant "ZooKeeper" as ZK
+participant "Venice Controller" as controller
+participant "Helix" as helix
+participant "Venice component\n(server/controller/router)" as venice
+sra -> ind:Report various ingestion status
+
+ind --> log
+log -> log:Log the message
+ind --> LF
+LF -> LF:Update status on \na countdown latch which controls \na Helix thread \nblocking on OFFLINE -> STANDBY\nuntil a resource in a terminal state
```
FelixGV, line 35: until a resource _is_ in a terminal state?

docs/architecture/venice_write_path.md
```diff
@@ -0,0 +1,42 @@
...
+| ZK    | Apache ZooKeeper. It's used in multiple Venice components to persist Venice metadata.                       |
+| Helix | Apache Helix. It's used in Venice to manage the cluster and assign partitions to different servers.         |
+| EV    | Helix External View. Venice uses it to represent the state of a server in a leader/follower mode.           |
+| CV    | Helix Customized View. Venice uses it to represent the ingestion state of a server for a given store.       |
+
+### Venice Store Ingestion on server side
+
+In Venice, a table or the dataset is called a store and each store has multiple partitions. Helix assign leaders and
+followers status on each partition to different servers. When assignments are done, each server will create multiple
+`StoreIngestionTask`s to handle ingestion for the partitions it's assigned to. The diagram belows describes the data flow of how a
```
FelixGV, line 32: While it is true that the server creates multiple SITs, it may be good to point out the mapping, which is that there is one SIT per store-version. Multiple partitions of the same store-version are handled by the same SIT instance.

FelixGV (commented):
Great initiative! Thanks a lot Adam. Left several minor comments about the content. Should be easy to address, so I think we could merge this soon.

docs/architecture/venice_architecture.md
adamxchen: I agree with that. I even thought about adding this to the user_guide and ended up creating a new section for it. However, the Dev guide sounds like a good place. It helps developers get up to speed in understanding the code structures and data flow. Let me create a subfolder `venice_internal` for this.

docs/architecture/venice_write_path.md
adamxchen: I thought about this a few times but decided to keep this way because I think if contributors need to write codes for SITs - unless we are going to overhaul it soon - they need to understand this state. I personally got confused by this God class for a while and when I drew the diagram, it's way more clear to me.

docs/architecture/venice_write_path.md
adamxchen: I think you are right. It got removed by #720. Nice to know and clean it up!

The image pop out seems to be a UI-related stuff with the Github page them [just the docs](https://github.com/just-the-docs/just-the-docs). We can revisit it later.

adamxchen:
@FelixGV please review when you get the chance! Appreciate the feedback and have updated the docs per the suggestions.

FelixGV (approved):
Thanks a lot Adam. This is great work!

PR #784, opened 2023-12-01
[controller] Add disabled replica cleanup service

majisourav99:
## Add disabled replica cleanup service

Venice disables a leader replica during future version ingestion error for Helix to elect a new leader replica. But that could create deployment issue when many partition with disabled replica becomes online as the disabled partition stay in OFFLINE state. This PR adds a periodic refresh service to reenable disabled replicas on every cluster's servers in child controller. This a best effort as re-enabling does not guarantee its successful transition back to LEADER. 
The config to enable this feature `"controller.renable.disabled.replica.enabled"`

### DIFF
--- a/internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
@@ -379,8 +379,7 @@ public class ConfigKeys {
   /**
    * The following config is to control whether to enable backup version cleanup based on retention policy or not at cluster level.
    */
-  public static final String CONTROLLER_RENABLE_DISABLED_REPLICA_ENABLED =
-      "controller.renable.disabled.replica.enabled";
+  public static final String CONTROLLER_ENABLE_DISABLED_REPLICA_ENABLED = "controller.enable.disabled.replica.enabled";
 
   // Server specific configs
   public static final String LISTENER_PORT = "listener.port";
--- a/internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
@@ -376,6 +376,12 @@ public class ConfigKeys {
   public static final String KME_REGISTRATION_FROM_MESSAGE_HEADER_ENABLED =
       "kme.registration.from.message.header.enabled";
 
+  /**
+   * The following config is to control whether to enable backup version cleanup based on retention policy or not at cluster level.
+   */
+  public static final String CONTROLLER_RENABLE_DISABLED_REPLICA_ENABLED =
+      "controller.renable.disabled.replica.enabled";
+
   // Server specific configs
   public static final String LISTENER_PORT = "listener.port";
   public static final String GRPC_READ_SERVER_PORT = "grpc.read.server.port";
--- a/services/venice-controller/src/main/java/com/linkedin/venice/controller/DisabledPartitionEnablerService.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/controller/DisabledPartitionEnablerService.java
@@ -1,6 +1,8 @@
 package com.linkedin.venice.controller;
 
 import com.linkedin.venice.service.AbstractVeniceService;
+import com.linkedin.venice.utils.SystemTime;
+import com.linkedin.venice.utils.Time;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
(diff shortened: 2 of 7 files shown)


services/venice-controller/src/main/java/com/linkedin/venice/controller/DisabledPartitionEnablerService.java
```diff
@@ -0,0 +1,83 @@
...
+              || !admin.isLeaderControllerFor(clusterName)) {
+            continue;
+          }
+          boolean didEnable = false;
+          try {
+            didEnable = admin.enableDisabledPartition(clusterName, "", true);
+          } catch (Exception e) {
+            LOGGER.error("Encountered exception while enabling disabled partition in cluster: {}", clusterName, e);
+          }
+          if (didEnable) {
```
m-nagarajan, line 66: curious on why introducing an extra layer of sleep of 10h after enabling the partitions in each cluster apart from the 10h sleep in the top of `run()`

internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
```diff
@@ -369,6 +369,12 @@ private ConfigKeys() {
   public static final String KME_REGISTRATION_FROM_MESSAGE_HEADER_ENABLED =
       "kme.registration.from.message.header.enabled";
 
+  /**
+   * The following config is to control whether to enable backup version cleanup based on retention policy or not at cluster level.
+   */
+  public static final String CONTROLLER_RENABLE_DISABLED_REPLICA_ENABLED =
```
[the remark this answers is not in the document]
majisourav99: Sure. will change the name.

services/venice-controller/src/main/java/com/linkedin/venice/controller/DisabledPartitionEnablerService.java
majisourav99: good catch. since renabling might cause new bootstraps, we should limit it, but another 10hr is too long. Let me change that.

services/venice-controller/src/main/java/com/linkedin/venice/controller/DisabledPartitionEnablerService.java
```diff
@@ -0,0 +1,83 @@
...
+  private final long sleepInterval;
+  private final AtomicBoolean stop = new AtomicBoolean(false);
+
+  public DisabledPartitionEnablerService(
+      VeniceHelixAdmin admin,
+      VeniceControllerMultiClusterConfig multiClusterConfig) {
+    this.admin = admin;
+    this.multiClusterConfig = multiClusterConfig;
+    this.allClusters = multiClusterConfig.getClusters();
+    this.sleepInterval = TimeUnit.HOURS.toMillis(10);
```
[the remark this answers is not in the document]
majisourav99: I rather not add another config which probably wont be tuned.

services/venice-controller/src/test/java/com/linkedin/venice/controller/TestDisabledPartitionEnablerService.java
```diff
@@ -0,0 +1,40 @@
...
+    Set<String> clusters = new HashSet<>();
+    doReturn(true).when(admin).isLeaderControllerFor(any());
+    clusters.add(clusterName);
+    doReturn(true).when(controllerConfig).isEnableDisabledReplicaEnabled();
+    doReturn(clusters).when(config).getClusters();
+    TestMockTime time = new TestMockTime();
+    DisabledPartitionEnablerService service = new DisabledPartitionEnablerService(admin, config, time);
+
+    service.startInner();
+    verify(admin, atLeast(1)).enableDisabledPartition(clusterName, "", true);
```
m-nagarajan, line 36: This checks whether `enableDisabledPartition()` is being called in the timer, which is great, but should we also add a test to disable multiple partitions and check if it's re-enabled by this service?

m-nagarajan (approved):
LGTM thanks @majisourav99

internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
```diff
@@ -376,6 +376,11 @@ private ConfigKeys() {
   public static final String KME_REGISTRATION_FROM_MESSAGE_HEADER_ENABLED =
       "kme.registration.from.message.header.enabled";
 
+  /**
+   * The following config is to control whether to enable backup version cleanup based on retention policy or not at cluster level.
```
adamxchen, line 380: this comment doesn't seem to be accurate. Isn't this config for enabling the OFFLINE replica?

PR #785, opened 2023-12-02
[server] Forced kill producers when ingestion task is killed

huangminchn:
## Summary
If StoreIngestionTask receives a kill-job command from Controller or operators, that means this store version is abandoned and it wouldn't have a chance to come online again. In that case, do a forced kill on resources like producer instead of trying to gracefully stop the ingestion task.

## How was this PR tested?
CI


gaojieliu (approved):
Thanks for the fix!

PR #786, opened 2023-12-03
[admin-tool] Verify the migration status of system stores during migration

### LINKED ISSUE #651: [BUG] Store migration should verify the cluster-discovery of system stores too
### Willingness to contribute

### Venice version

0.4.139

### System information

- **OS Platform and Distribution (e.g., Linux Ubuntu 20.0)**: Mariner 5.15.111.1-1.cm2
- **JDK version**: 17

### Describe the problem

During a store's migration, a subsequent empty push to the store was stuck in EOPR and did not complete. The CV and OfflinePushes znodes showed that all replicas of all partitions were completed.

In the controller these logs were seen:

```
2023/09/21 15:55:51.171 INFO [PushMonitorUtils] [pool-41-thread-4] [venice-controller-war] [] Getting Da Vinci push status for topic: CAIExperimentalStore_v4
2023/09/21 15:55:51.174 WARN [LoggingFilter] [R2 Callback Executor-4-8] [venice-controller-war] [AAYF4Ia2TrpWPETn/wGUAw==] onStreamError: ex: com.linkedin.r2.message.stream.StreamException (Received error 400 from server for URI https://lva1-app
96617.prod.linkedin.com:8438/storage/venice_system_store_davinci_push_status_store_CAIExperimentalStore/BAAIAAAAAg==) wire: {} rpc-trace: (lva1-app19053,venice-controller-war,No Context Set,2023/09/21 15:55:51.171)
com.linkedin.r2.message.stream.StreamException: Received error 400 from server for URI https://lva1-app96617.prod.linkedin.com:8438/storage/venice_system_store_davinci_push_status_store_CAIExperimentalStore/BAAIAAAAAg== 
The router port was corresponding to the src cluster.
```

The issue here is that the discovery cluster switch happens during migration when there are requests made to the src cluster's routers between --complete-migration and --end-migration. In the case of system stores, it is possible that there is no traffic during this period. Hence, post migration, the clients will not be able to switch to the new clusters.

A suggested way to fix this would be add code into the --complete-migration or --end-migration commands in the venice-admin-tool to check the appropriate system stores.

### Tracking information

_No response_

### Code to reproduce bug

_No response_

- [X] `Admin Tool`: This is the stand-alone client used for ad-hoc operations on Venice.

kvargha:

Added migration status checks for all system stores during checkMigrationStatus in the AdminTool. This was necessary because migration status checks were added for cluster stores but not system stores. So post migration, clients may try to submit requests to system stores that are no longer available.

Resolves #651

## How was this PR tested?

Added unit test.


kvargha:
@ZacAttack can you please trigger the CI? Thanks.

ZacAttack (changes requested):
This is looking great!  Left a comment.

Theres a test called testStoreMigrationWithDaVinciPushStatusSystemStore().  I think we could probably add a check in that test case that just looks at the checkMigration status output and verifies the system store is in both clusters.  Basically inserting an extra check between lines 327 and 328 in TestStoreMigration.java.  Once we have that I think this is ready to ship!

clients/venice-admin-tool/src/main/java/com/linkedin/venice/AdminTool.java
```diff
@@ -1648,6 +1649,27 @@ private static void checkMigrationStatus(CommandLine cmd) {
     ControllerClient destControllerClient = new ControllerClient(destClusterName, veniceUrl, sslFactory);
 
     ChildAwareResponse response = srcControllerClient.listChildControllers(srcClusterName);
+
+    for (VeniceSystemStoreType systemStoreType: VeniceSystemStoreType.values()) {
+      MultiStoreInfoResponse srcClusterStoreInfo = srcControllerClient.getClusterStores(srcClusterName);
```
ZacAttack, line 1654: I may have given bad advice, apologies :P 

It looks like the behavior of getClusterStores has an unfortunate behavior today.  In order for it to work, it has to be wired to the controller which happens to be the leader for administering that cluster.  This... is an annoying thing for an operator to figure out.  So need to restructure this a bit.

What we can is leverage another method in this class called queryStoreList.  This will consult the operator passed in controller endpoint for the complete store list.  We can then filter it down to the system store's we care about and print the migration status.

Sorry about that -_-;

kvargha: No worries!

I've made the modifications. 

Is there any way we can assert that `checkMigrationStatus()` is working properly in the test?

Also, how do I view the log output when running tests?

Thanks!

ZacAttack:
> > This is looking great! Left a comment.
> > Theres a test called testStoreMigrationWithDaVinciPushStatusSystemStore(). I think we could probably add a check in that test case that just looks at the checkMigration status output and verifies the system store is in both clusters. Basically inserting an extra check between lines 327 and 328 in TestStoreMigration.java. Once we have that I think this is ready to ship!
> 
> Hey @ZacAttack, I modified checkMigrationStatus to check the migration status for all system store types, and added a checkMigrationStatus between startMigration and completeMigration in the test.
> 
> Regarding checking the checkMigrationStatus output, how would I go about doing that given the function has a void return type?

Yeah, admittedly the way this code is architected right now it doesn't test easy.

So let's do this to make it a bit better.  Right now checkMigrationStatus function takes only one argument.  It's the cmd.  Let's pass to it a closure for where is should print the migration status.  Right now printMigrationStatus just parses out a string and then passes it to System.err.println().  If we could have it instead populate a string we can validate the output of, then that'd be ideal.

So something like:

`  @FunctionalInterface
  interface PrintFunction {
    void apply(String message);
  }
  private static void printMigrationStatus(ControllerClient controller, String storeName) {
    printMigrationStatus(controller, storeName, System.err::println);
  }
  private static void printMigrationStatus(ControllerClient controller, String storeName, PrintFunction printFunction) {
   .....blah blah blah
       printFunction.apply("\t" + storeName + " belongs to cluster " + controller.discoverCluster(storeName).getCluster()
        + " according to cluster discovery");
`
And then we would augment the interface of checkMigrationStatus(CommandLine cmd) to be now checkMigrationStatus(CommandLine cmd, PrintFunction printFunction) and pass that printFunction along.

Once we've done that, then we can invoke checkMigrationStatus with a PrintFunction that we've instrumented to store into a string that we can examine with Asserts.

clients/venice-admin-tool/src/main/java/com/linkedin/venice/AdminTool.java
```diff
@@ -1648,6 +1648,26 @@ private static void checkMigrationStatus(CommandLine cmd) {
     ControllerClient destControllerClient = new ControllerClient(destClusterName, veniceUrl, sslFactory);
 
     ChildAwareResponse response = srcControllerClient.listChildControllers(srcClusterName);
+
+    // Check the migration status for system stores in the source and destination clusters
+    MultiStoreResponse srcClusterStores = srcControllerClient.queryStoreList();
+    for (String currStoreName: srcClusterStores.getStores()) {
+      VeniceSystemStoreType systemStoreType = VeniceSystemStoreType.getSystemStoreType(currStoreName);
+
+      if (systemStoreType != null && currStoreName.matches(systemStoreType.getSystemStoreName(currStoreName))) {
```
ZacAttack, line 1657: Other tweak.  This if statement never evaluates to true.  I think it's because currStoreName is a larger string, and systemStoreType.getSystemStoreName(currStoreName) does... I'm not sure, but probably something which is a trim of currStoreName.

I think we can tweak this to be: systemStoreType.extractRegularStoreName(currStoreName).equals(storeName)

This I think would also avoid gotchas where theres a storeName which is a substring of a different store.

clients/venice-admin-tool/src/main/java/com/linkedin/venice/AdminTool.java
```diff
@@ -1648,6 +1648,26 @@ private static void checkMigrationStatus(CommandLine cmd) {
     ControllerClient destControllerClient = new ControllerClient(destClusterName, veniceUrl, sslFactory);
 
     ChildAwareResponse response = srcControllerClient.listChildControllers(srcClusterName);
+
+    // Check the migration status for system stores in the source and destination clusters
+    MultiStoreResponse srcClusterStores = srcControllerClient.queryStoreList();
```
ZacAttack, line 1653: And then one last tweak.  Let's move these lines into the below if statements that check parent controller and child controllers (I guess by copying it twice or wrapping into a function, either works).

The logic right now for this tools is to

if (controllerUrl is NOT a parent controller) {
  only list migration status for this controllerUrl (one datacenter)
} else {
   list migration status for parent controller
   for() { list status for each child controller }
}

In the above flow we basically would like to execute your code for each datacenter (parent and each child) that we check.

kvargha:
Thanks for pointing out these tweaks!

> > > This is looking great! Left a comment.
> > > Theres a test called testStoreMigrationWithDaVinciPushStatusSystemStore(). I think we could probably add a check in that test case that just looks at the checkMigration status output and verifies the system store is in both clusters. Basically inserting an extra check between lines 327 and 328 in TestStoreMigration.java. Once we have that I think this is ready to ship!
> > 
> > 
> > Hey @ZacAttack, I modified checkMigrationStatus to check the migration status for all system store types, and added a checkMigrationStatus between startMigration and completeMigration in the test.
> > Regarding checking the checkMigrationStatus output, how would I go about doing that given the function has a void return type?
> 
> Yeah, admittedly the way this code is architected right now it doesn't test easy.
> 
> So let's do this to make it a bit better. Right now checkMigrationStatus function takes only one argument. It's the cmd. Let's pass to it a closure for where is should print the migration status. Right now printMigrationStatus just parses out a string and then passes it to System.err.println(). If we could have it instead populate a string we can validate the output of, then that'd be ideal.
> 
> So something like:
> 
> ```
>   @FunctionalInterface
>   interface PrintFunction {
>     void apply(String message);
>   }
>   private static void printMigrationStatus(ControllerClient controller, String storeName) {
>     printMigrationStatus(controller, storeName, System.err::println);
>   }
>   private static void printMigrationStatus(ControllerClient controller, String storeName, PrintFunction printFunction) {
>    .....blah blah blah
>        printFunction.apply("\t" + storeName + " belongs to cluster " + controller.discoverCluster(storeName).getCluster()
>         + " according to cluster discovery");
> ```
> 
> And then we would augment the interface of checkMigrationStatus(CommandLine cmd) to be now checkMigrationStatus(CommandLine cmd, PrintFunction printFunction) and pass that printFunction along.
> 
> Once we've done that, then we can invoke checkMigrationStatus with a PrintFunction that we've instrumented to store into a string that we can examine with Asserts.

Since we can't get the output of checkMigrationStatus from the test file due to it being executed in main, could we have a global variable where we can store the log output of printMigrationStatus like so?

`private static ArrayList<String> systemStoreStatuses = new ArrayList<>();
    
    public static ArrayList<String> getSystemStoreStatuses() {
        return systemStoreStatuses;
    }
    
    private static void printMigrationStatus(ControllerClient controller, String storeName, PrintFunction printFunction) {
        systemStoreStatuses.add(printFunction.apply("Current status"));
    }`

One issue with this is systemStoreStatuses could grow infinitely and may cause an out-of-memory error.

Another idea I had was to make a public version of checkMigrationStatus to directly read the return value from within the test.

What are your thoughts?

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestStoreMigration.java
```diff
@@ -325,6 +325,7 @@ public void testStoreMigrationWithDaVinciPushStatusSystemStore() throws Exceptio
               .assertEquals(pushStatusStoreReader.getPartitionStatus(storeName, 1, 0, Optional.empty()).size(), 1));
 
       startMigration(parentControllerUrl, storeName);
+      checkMigrationStatus(parentControllerUrl, storeName);
```
ZacAttack, line 328: Referring to comment about main():

We don't have to sweat main or global variables now that we have a print function closure.  We can change the signature of checkMigrationStatus which takes a PrintFunction we define here.  Our PrintFunction in our test would look like:

      List<String> statusOutput = new ArrayList<>();
      AdminTool.PrintFunction printFunction = (string) -> {statusOutput.add(string)};
      checkMigrationStatus(parentControllerUrl, storeName,printFunction);
      Assert.assertTrue(statusOutput.contains(storeName));
      // assert ssystem storeName etc.

No global variable needed, no execution of main needed.  Admittedly a full integration test would go end to end with main, but I think this is ok for now.

clients/venice-admin-tool/src/main/java/com/linkedin/venice/AdminTool.java
```diff
@@ -1614,7 +1614,16 @@ private static void migrateStore(CommandLine cmd) {
...
+  }
+
   private static void printMigrationStatus(ControllerClient controller, String storeName) {
+    printMigrationStatus(controller, storeName, System.err::println);
+  }
+
+  private static void printMigrationStatus(ControllerClient controller, String storeName, PrintFunction printFunction) {
     StoreInfo store = controller.getStore(storeName).getStore();
 
     System.err.println("\n" + controller.getClusterName() + "\t" + controller.getLeaderControllerUrl());
```
ZacAttack, line 1629: Let's use the print Function on the rest of these println's

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestStoreMigration.java
kvargha: Thank you for the clarification regarding the closure usage.

For this are we bypassing main? Since `printFunction` isn't a string I can't pass it into the function below as an argument inside the test file.

```
private void checkMigrationStatus(String controllerUrl, String storeName) throws Exception {
    String[] checkMigrationStatusArgs = { "--migration-status", "--url", controllerUrl, "--store", storeName,
        "--cluster-src", srcClusterName, "--cluster-dest", destClusterName };
    AdminTool.main(checkMigrationStatusArgs);
}
```

If so, would it be ok to make `checkMigrationStatus` in AdminTool.java a public function so I can directly invoke it in the test file?

kvargha:
Hey @ZacAttack, `DiffCoverage` keeps failing.

I was reading over the documentation and it said that JaCoCo and DiffCoverage don't work with integration tests. Is there a specific unit test file that I can test `checkMigrationStatus` against?

ZacAttack:
> Hey @ZacAttack, `DiffCoverage` keeps failing.
> 
> I was reading over the documentation and it said that JaCoCo and DiffCoverage don't work with integration tests. Is there a specific unit test file that I can test `checkMigrationStatus` against?

I sent you som details on the Venice slack.  Should get you most the way on this.

kvargha:
@ZacAttack This should be good to go now. Can you please trigger the CI?

ZacAttack (approved):
Kicked off CI.  Once that shows green I think we are ready to merge this now.  Thanks!

PR #787, opened 2023-12-04
[server] Fix edge case in AAWC list element update

sixpluszero:
## [server] Fix edge case in AAWC list element update
This PR tries to fix a bug in AAWC logic on list element update.
The bug is: when trying to update a pre-existing element in the list field, it only check if the TS is different from the existing element, not greater than the existing element TS. It might lead to update timestamp rewind for the element, and if there is another DELETE that happens between the two updates TS, certain update order can lead to data being deleted - thus data loss.

## How was this PR tested?

Add new unit test.


gaojieliu (approved):
Thanks for the fix!

PR #788, opened 2023-12-05
[server] Fix edge case in AAWC list element update

sixpluszero:
## [server] Fix edge case in AAWC list element update
This PR tries to fix a bug in AAWC logic on list element update.
The bug is: when trying to update a pre-existing element in the list field, it only check if the TS is different from the existing element, not greater than the existing element TS. It might lead to update timestamp rewind for the element, and if there is another DELETE that happens between the two updates TS, certain update order can lead to data being deleted - thus data loss.

## How was this PR tested?
New unit test

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/schema/merge/SortBasedCollectionFieldOpHandler.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/schema/merge/SortBasedCollectionFieldOpHandler.java
@@ -561,13 +561,9 @@ public class SortBasedCollectionFieldOpHandler extends CollectionFieldOperationH
         activeElementToTsMap.remove(toAddElement);
         newPutOnlyPartLength--;
       }
-      if (activeTimestamp == null) {
+      if (activeTimestamp == null || activeTimestamp < modifyTimestamp) {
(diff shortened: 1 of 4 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/schema/merge/SortBasedCollectionFieldOpHandler.java
```diff
@@ -564,8 +564,7 @@ private UpdateResultStatus handleModifyCollectionMergeList(
       if (activeTimestamp == null) {
         activeElementToTsMap.put(toAddElement, modifyTimestamp);
         updated = true;
-      } else if (activeTimestamp != modifyTimestamp) {
-        // activeElementToTsMap.remove(toAddElement);
+      } else if (activeTimestamp < modifyTimestamp) {
         activeElementToTsMap.put(toAddElement, modifyTimestamp);
         updated = true;
```
nisargthakkar, line 570: Should we merge these two branches? They're the same

sixpluszero: Good catch. Missed this in main branch though

sushantmane (approved):

PR #791, opened 2023-12-05
[build] upgraded rocksdb dep to 8.8.1

gaojieliu:
## Summary
The latest rocksdb release will enable this feature: `level_compaction_dynamic_level_bytes` by default to reduce the space amplification, and we did some study and we thought this can help us reduce the disk usage for high write throughput stores. Related post for this feature:
https://github.com/facebook/rocksdb/wiki/Leveled-Compaction

## How was this PR tested?
CI

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBServerConfig.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBServerConfig.java
@@ -219,7 +219,6 @@ public class RocksDBServerConfig {
 
   private final long rocksDBBlockCacheSizeInBytes;
   private final long rocksDBRMDBlockCacheSizeInBytes;
-  private final long rocksDBBlockCacheCompressedSizeInBytes;
   private final boolean rocksDBBlockCacheStrictCapacityLimit;
   private final boolean rocksDBSetCacheIndexAndFilterBlocks;
   private final int rocksDBBlockCacheShardBits;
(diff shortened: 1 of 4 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBServerConfig.java
```diff
@@ -219,7 +219,6 @@ public class RocksDBServerConfig {
 
   private final long rocksDBBlockCacheSizeInBytes;
   private final long rocksDBRMDBlockCacheSizeInBytes;
-  private final long rocksDBBlockCacheCompressedSizeInBytes;
```
FelixGV, line 222: Is this removed in the latest version of RocksDB?

FelixGV (approved):
Thank you!

majisourav99 (commented):
Can we expose `setLevelCompactionDynamicLevelBytes` and set it to false explicitly and then through becnhmark/experiment we enable it slowly ? I believe 8.4+ makes it default true.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBServerConfig.java
gaojieliu: yes.

gaojieliu:
> Can we expose `setLevelCompactionDynamicLevelBytes` and set it to false explicitly and then through becnhmark/experiment we enable it slowly ? I believe 8.4+ makes it default true. Worst case we will have option to disable it if something goes bad.

Personally, I don't want to introduce this config if it is enabled by default, and there are a few reasons:
1. RocksDB team is confident to have this feature enabled by default.
2. There must be some new features/configs enabled in the latest release, which we are not aware, and I don't think we want to introduce configs for these as well.

So my plan is to evaluate this new release as a whole and roll it back completely since the change is more than this option.

PR #792, opened 2023-12-07
[fast-client] Wrong value schema id used when handling multi-get response in FC

xunyin8:
## [fast-client] Wrong value schema id used when handling multi-get response in FC

batchGetTransportRequestCompletionHandler should obtain the correct dataRecordDeserializer by using the MultiGetResponseRecordV1's per record value schema id field. The handler was using the transport response header's schema id which is inteded for the protocol version and not the write schema version for the data records. As a result, FC is running into data issues when the user record's write schema id > 1 with new fields/data.

## How was this PR tested?
Existing tests and modified an unit test to ensure the correct value schema id is obtained. Also war dropped to confirm the root cause and fix.


m-nagarajan (commented):
Thanks @xunyin8 for figuring out this issue and the quick fix. I think the change looks good. Should we modify one integration test to have 2 schema versions and verify the data based on v2?

xunyin8:
> Thanks @xunyin8 for figuring out this issue and the quick fix. I think the change looks good. Should we modify one integration test to have 2 schema versions and verify the data based on v2?

I think the unit test is sufficient for this specific issue, but will add one in case there are other issues with multiple value schemas.

m-nagarajan (dismissed):
LGTM, thanks @xunyin8

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/DispatchingAvroGenericStoreClient.java
```diff
@@ -572,6 +571,9 @@ private void batchGetTransportRequestCompletionHandler(
     VeniceCompressor compressor =
         metadata.getCompressor(transportClientResponse.getCompressionStrategy(), requestContext.currentVersion);
     for (MultiGetResponseRecordV1 r: records) {
+      // Since the underlying FastSerializerDeserializerFactory is caching deserializers by reader, writer and object
+      // class. Retrieving the corresponding deserializer per record should be fast here after initial warmup.
+      RecordDeserializer<V> dataRecordDeserializer = getDataRecordDeserializer(r.getSchemaId());
```
FelixGV, line 576: Nitpick: I think we could move this line to just after `nanoTsBeforeDeserialization`, since looking up the deserializer is technically part of the deserialization time.

Also, another nitpick in that area... we do:

```
      totalDecompressionTimeForResponse += System.nanoTime() - nanoTsBeforeDecompression;
      long nanoTsBeforeDeserialization = System.nanoTime();
```

I don't think we need to do two `System.nanoTime()` calls in a row, we should be able to do only one, e.g. something like:

```
      long nanoTsBeforeDeserialization = System.nanoTime();
      V deserializedValue = dataRecordDeserializer.deserialize(decompressRecord);
      requestContext.recordRecordDeserializationTime(
          transportClientResponse.getRouteId(),
          getLatencyInNS(nanoTsBeforeDeserialization));
      MultiKeyRequestContext.KeyInfo<K> k = keyInfos.get(r.keyIndex);
      keysSeen.add(r.keyIndex);
      callback.onRecordReceived(k.getKey(), deserializedValue);
      totalDecompressionTimeForResponse += nanoTsBeforeDeserialization - nanoTsBeforeDecompression;
```

And one more nitpick here (last one, I promise 😅 ...), why do we measure the per-record deserialization time and the total decompression time? Why not measure the total time in both cases? If we're trying to analyze the client's performance and figure out where the bottleneck is, it seems confusing and counter-productive to have these two measurements at different granularity, because then we would need to normalize by the number of keys (per route!) in order to compare them to one another. Finding out what is the bigger bottleneck would be non-straightforward...

FelixGV (commented):
Thanks for the fast turnaround Xun!

I left 3 nitpicky comments... they could be left for another PR if you think they're too troublesome to address, but I think they're probably fairly simple so it would be great if we could knock 'em off before forgetting...

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/DispatchingAvroGenericStoreClient.java
majisourav99, line 576: Looks like there are multiple hops inside `getDataRecordDeserializer`. Should we cache the lookup locally ?

xunyin8: Great suggestions, will fix.

xunyin8: Addressed the first two points. I think Nisarg already fixed the third point in his refactor since recordRecordDeserilizationTime is aggregating by routeID
```
  void recordRecordDeserializationTime(String routeId, long latencyInNS) {
    Validate.notNull(routeId);
    routeRequests.get(routeId).recordDeserializationTime.addAndGet(latencyInNS);
  }
```

FelixGV: In other code paths we have optimized this access pattern with `AvroStoreDeserializerCache`. It is in `venice-common`, so FC should have access to it already.

FelixGV: Sure. It seems a bit unfortunate to do the `routeRequests.get(routeId)` lookup per record but that's ok... No need to bloat the current PR. Thanks for addressing the first two!

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/DispatchingAvroGenericStoreClient.java
```diff
@@ -571,16 +570,22 @@ private void batchGetTransportRequestCompletionHandler(
     long totalDecompressionTimeForResponse = 0;
     VeniceCompressor compressor =
         metadata.getCompressor(transportClientResponse.getCompressionStrategy(), requestContext.currentVersion);
+    // Despite FastSerializerDeserializerFactory is caching deserializers by reader, writer and object class.
+    // Retrieving the corresponding deserializer still requires constructing some container object which could add up.
+    Map<Integer, RecordDeserializer<V>> recordDeserializerMap = new HashMap<>();
```
majisourav99, line 575: can we make it class level variable?

FelixGV: We cannot make it a class property if we lookup only by 1 ID (the reader schema), because the writer schema (which should be the latest schema) can change over time. If we want to make it a class property, then we need to look up by two IDs, and we already have a class tailor-made for this: `AvroStoreDeserializerCache`. This is likely the optimal approach, IMHO.

If we keep it within the scope of the function only, then we can lookup only by reader schema ID and assume the writer schema doesn't change much in the span of one function call. In that case it's ok to have it as a locally scoped variable. Though a `SparseConcurrentList` would be more optimal than a Integer-keyed map...

xunyin8: Yeah exactly what @FelixGV said `getDataRecordDeserializer` gets the latest reader schema. If we cache it at the class level then even if the reader schema is updated it won't reflect until writer schema changes (triggers new compute if absent call) or app is bounced. That's different behavior than our thin-client today and could cause issues. 

@FelixGV you mean reader schema will not change in the span of one request call right? There could be different writer schemas in one call because keys are written at different times for a hybrid store.

Will checkout `SparseConcurrentList`.

majisourav99: Yeah, just looked at `AvroStoreDeserializerCache` impl. It would be the better approach.

gaojieliu: `AvroStoreDeserializerCache` will work for generic store client and for specific store client, it may not work since the reader schema carried in SpecificRecord may not be same as the schemas registered in schema repo.

xunyin8: @gaojieliu what is your recommendation?

gaojieliu: For Generic store client, it will just work and for specific store client, we can overload it by specifying the reader schema at construction time: `AvroSpecificStoreDeserializerCache`.
And that will potentially be easier to wrap up all the deserializer  cache related logics into separate class instead of repeating the same code in many different places.

gaojieliu (commented):
Looks good overall, left one minor comment.

internal/venice-common/src/main/java/com/linkedin/venice/serialization/avro/AvroSpecificStoreDeserializerCache.java
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  private AvroSpecificStoreDeserializerCache(IntFunction<Schema> schemaGetter, Class<V> valueClass) {
+    this.schemaGetter = schemaGetter;
+    this.cache = new SparseConcurrentList<>();
+    this.valueClass = valueClass;
+  }
+
+  @Override
+  public RecordDeserializer<V> getDeserializer(int writerSchemaId, int readerSchemaId) {
```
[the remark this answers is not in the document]
xunyin8: I tried it but it will cause errors and require refactoring of ChunkingAdapter since it was originally implemented when both specific and generic deser cache was sharing the same class and invoked with the same `getDeserializer(int writerSchemaId, int readerSchemaId)`. We can do it in another PR for the refactor since this PR is getting bigger and bigger.

gaojieliu (dismissed):

gaojieliu (approved):

PR #794, opened 2023-12-08
[server] Add a new option to allocate dedicated consumer pool for leader replicas of AA/WC stores

gaojieliu:
## Summary
Based on the offline analysis, the logic to handle writes from Active/Active or Write-compute stores in consumer threads is much heavier than other stores, which would slow down the ingestion speed of regular stores significantly when there are some high write throughput to AA or WC stores.
This code change introduces an option to assign the leader replicas of AA or WC stores to a dedicated consumer pool, so that the regular store ingestion won't be affected by the above combination.

The write compute flag is store-level config, so it can change in the lifetime of a particular version, and this code change won't react of write compute flag change until a restart.

New configs:
server.dedicated.consumer.pool.for.aa.wc.leader.enabled: default false server.dedicated.consumer.pool.size.for.aa.wc.leader: default 5

This code change will introduce a new set of consumer service metrics with suffix: `_for_aa_wc_leader` when this feature is enabled.

## How was this PR tested?
CI


FelixGV (approved):
Thanks, looks pretty clean to me!

FelixGV (approved):

lluwm (approved):

PR #795, opened 2023-12-09
[alpini][router] Repaired the Alpini build

FelixGV:
The Alpini build got broken by the Netty bump in 9163931. This does not seem to affect the main code, which has been running in production successfully, but some of the tests fail due to this. In addition, some tests are failing due to subtle differences between JDK versions.

This commit is a series of changes to fix all of that.

- Tweaked TestRequestLogHandler so that it works in all JDK versions.

- Skipped TestHeaderUtils::testRandomWeakUUID on JDK 17 since it does not work.

- Created a TestNettyUtil with a function to detect when Epoll is not available (as is the case on mac) and throw a SkipException so that tests don't fail.

- Deleted several unused classes that were broken by the Netty bump, including: ContentCompressor, HttpClientResponseHandler, Http2ClientResponseHandler and Http2AwareChannelPoolFactory, as well as their accompanying tests.

- Also added a dependency to org.bouncycastle:bcpkix-jdk15on which is necessary in JDK17.

Misc:

- Added JDK 8 and 11 to AlpiniUnitTests, like we do for Venice unit tests. They will still run only when the Alpini code changes.

## How was this PR tested?
CI.


internal/alpini/netty4/alpini-netty4-base/src/test/java/com/linkedin/alpini/util/TestNettyUtil.java
```diff
@@ -0,0 +1,23 @@
...
+
+public class TestNettyUtil {
+  public static <T> T skipEpollIfNotFound(Supplier<T> eventLoopGroupSupplier) {
+    try {
+      return eventLoopGroupSupplier.get();
+    } catch (NoClassDefFoundError e) {
+      if (e.getMessage().equals("Could not initialize class io.netty.channel.epoll.EpollEventLoop")) {
+        throw new SkipException("EPoll not available.");
+      }
+      throw e;
```
majisourav99, line 15: Here and in the next catch, do we expect to specified exception with different messages? or its just for safety ?
nit: since we are doing similar action can we catch using `catch (NoClassDefFoundError | UnsatisfiedLinkError e)` ?

FelixGV: I wanted to verify the error message so that the catch is as specific as possible (even more specific than just the exception type) in order to let any other kind of issue bubble up.

We could use the pipe notation, but I would still want to check the exception messages, so I don't think it would reduce the code size much...

majisourav99 (approved):
Thanks a lot Felix for cleaning up the alpini build.

FelixGV:
> Thanks a lot Felix for cleaning up the alpini build.

Thanks for the review!

PR #797, opened 2023-12-12
[server][da-vinci] Delete Unused non NR Metrics

ZacAttack:
## Delete Unused non NR Metrics

We seem to have a pile of metrics that harken back to pre-NR. This change removes them.

Resolves #XXX

## How was this PR tested?
Ran CI and verified in production hosts these metrics are always blank.


majisourav99 (approved):
LGTM. Thanks Zac

PR #798, opened 2023-12-12
[server][test] Allow and skip acl checks for health check uri  from clients to server

m-nagarajan:
## Summary
Fast client warms up H2 connections by hitting the `/health` endpoint in servers but as uri validation for requests from clients to servers happens in `StoreAclHandler`, added changes to allow the `/heatlh` uri and by-pass acl validation since it is just a health check endpoint without any pii.

## How was this PR tested?
unit tests and GHCI


adamxchen (dismissed):
LGTM. Have one nit.

PR #799, opened 2023-12-12
[server] Added info to the exception in AASIT::getUpstreamKafkaUrlFromKafkaValue

FelixGV:
The exception now looks like this:

com.linkedin.venice.exceptions.VeniceException: No Kafka cluster ID found in the cluster ID to Kafka URL map. Got cluster ID -1 and ID to cluster URL map {0=>url0, 1=>url1}. Source Kafka: sourceKafkaURL; TP(topic: 'topic', partition: 0); Offset: 100; Message type: CONTROL_MESSAGE/TOPIC_SWITCH; ProducerMetadata: {"producerGUID": null, "segmentNumber": 0, "messageSequenceNumber": 0, "messageTimestamp": 0, "logicalTimestamp": 0}; LeaderMetadataFooter: {"hostName": null, "upstreamOffset": 0, "upstreamKafkaClusterId": -1}

## How was this PR tested?
New unit test.

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
@@ -1252,9 +1252,13 @@ public class ActiveActiveStoreIngestionTask extends LeaderFollowerStoreIngestion
       throw new VeniceException(
           String.format(
               "No Kafka cluster ID found in the cluster ID to Kafka URL map. "
-                  + "Got cluster ID %d and ID to cluster URL map %s",
+                  + "Got cluster ID %d and ID to cluster URL map %s. "
+                  + "Message type: %d; ProducerMetadata: %s; LeaderMetadataFooter: %s",
               kafkaValue.leaderMetadataFooter.upstreamKafkaClusterId,
-              kafkaClusterIdToUrlMap));
+              kafkaClusterIdToUrlMap,
+              kafkaValue.messageType,
+              kafkaValue.producerMetadata,
+              kafkaValue.leaderMetadataFooter));
     }
     return upstreamKafkaURL;
   }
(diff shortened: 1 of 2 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
```diff
@@ -1252,9 +1252,13 @@ private String getUpstreamKafkaUrlFromKafkaValue(KafkaMessageEnvelope kafkaValue
       throw new VeniceException(
           String.format(
               "No Kafka cluster ID found in the cluster ID to Kafka URL map. "
-                  + "Got cluster ID %d and ID to cluster URL map %s",
+                  + "Got cluster ID %d and ID to cluster URL map %s. "
+                  + "Message type: %d; ProducerMetadata: %s; LeaderMetadataFooter: %s",
```
huangminchn, line 1256: I think we need a few more info:
1. which topic is this message from
2. the offset of this message
3. which Kafka cluster is it from (this one might be difficult; if so, it's good to have the above two info first)

huangminchn (changes requested):
Thanks a lot Felix!

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
FelixGV: Sure, I'll change the function signature so I can add that.

FelixGV: Done in the latest commit.

lluwm (dismissed):

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
FelixGV: I was missing the third one, but actually, it is easy to add as well, so I did in the latest commit. All 3 infos you asked for are included now.

lluwm (approved):
LGTM! Thanks for adding these extra info

huangminchn (approved):
Thanks a lot Felix!

PR #800, opened 2023-12-13
Fix BootstrappingVeniceChangelogConsumer no metadata issue

zx0401:

Fix BootstrappingVeniceChangelogConsumer no metadata issue. Previously BootstrappingVeniceChangelogConsumer can store the data in local RocksDB. However, it couldn't successfully bootstrap from RocksDB as we didn't save the metadata correctly. This PR is to fix it.

## How was this PR tested?

Tested by running integration test `testBootstrappingVeniceChangelogConsumer_WithPollAndCatchUp` twice with same store name. On the second run, verified the log to show we could load data directly from rocksDB instead of running pollAndCatchUp.

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/InternalLocalBootstrappingVeniceChangelogConsumer.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/InternalLocalBootstrappingVeniceChangelogConsumer.java
@@ -9,7 +9,6 @@ import static com.linkedin.davinci.store.rocksdb.RocksDBServerConfig.ROCKSDB_LEV
 import static com.linkedin.davinci.store.rocksdb.RocksDBServerConfig.ROCKSDB_PLAIN_TABLE_FORMAT_ENABLED;
 import static com.linkedin.venice.ConfigKeys.DATA_BASE_PATH;
 import static com.linkedin.venice.kafka.partitionoffset.PartitionOffsetFetcherImpl.DEFAULT_KAFKA_OFFSET_API_TIMEOUT;
-import static com.linkedin.venice.offsets.OffsetRecord.LOWEST_OFFSET;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.linkedin.davinci.callback.BytesStreamingCallback;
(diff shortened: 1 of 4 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangeCoordinate.java
```diff
@@ -96,11 +85,16 @@ protected VeniceChangeCoordinate(String topic, PubSubPosition pubSubPosition, In
 
   public static String convertVeniceChangeCoordinateToStringAndEncode(VeniceChangeCoordinate veniceChangeCoordinate)
       throws IOException {
-    veniceChangeCoordinate.writeExternal(_objectOutputStream.get());
-    _objectOutputStream.get().flush();
-    _objectOutputStream.get().close();
```
ZacAttack, line 101: Was this doing something bad?  I had figured we'd use member variables here to help with GC

zx0401: Yes,` _byteArrayOutputStream.get().toByteArray();` will keep the previously saved value and just appended new value. For example: https://stackoverflow.com/questions/35450584/is-it-possible-to-clear-bytearrayoutputstream. Since we only use this in VeniceChangelogCheckpointThread which runs every 20 seconds, it shouldn't cause obvious performance impact.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/InternalLocalBootstrappingVeniceChangelogConsumer.java
```diff
@@ -308,6 +312,10 @@ private Collection<PubSubMessage<K, ChangeEvent<V>, VeniceChangeCoordinate>> pol
       currentPartitionState.currentPubSubPosition = record.getOffset();
```
ZacAttack, line 312: So I 'thought' this is where we figured the bug was?  Doing the checkpointing just based off of currentPubSubPosition from the last record consumed isn't quite the right thing to do since the poll method is going to screen out control messages.

Looking at the code in kafkaConsumer I could have swore we had a method for asking the consumer the high offset, but it looks like I am mistaken....  I'll put up a change for this.

zx0401: Yes, if there is any control message got filtered out we'll get stuck here. I didn't change it in this PR as currently this will fail the EI only as it doesn't have frequent updates.

ZacAttack (approved):
Looks good!

PR #801, opened 2023-12-14
[server][da-vinci-client] Add server current version metric and other changes

majisourav99:
## Add server current version metric

Add server current version metric + added logging when server gets error request for easier debugging.

## How was this PR tested?

GHCI

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/VeniceVersionedStatsReporter.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/VeniceVersionedStatsReporter.java
@@ -27,10 +27,8 @@ public class VeniceVersionedStatsReporter<STATS, STATS_REPORTER extends Abstract
 
     this.isSystemStore = VeniceSystemStoreUtils.isSystemStore(storeName);
 
-    if (!isSystemStore) {
-      registerSensor("current_version", new Gauge(() -> currentVersion));
-      registerSensor("future_version", new Gauge(() -> futureVersion));
-    }
+    registerSensor("current_version", new Gauge(() -> currentVersion));
+    registerSensor("future_version", new Gauge(() -> futureVersion));
 
     this.currentStatsReporter = statsSupplier.get(metricsRepository, storeName + "_current");
     if (!isSystemStore) {
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/VeniceVersionedStatsReporter.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/VeniceVersionedStatsReporter.java
@@ -27,8 +27,10 @@ public class VeniceVersionedStatsReporter<STATS, STATS_REPORTER extends Abstract
 
     this.isSystemStore = VeniceSystemStoreUtils.isSystemStore(storeName);
 
-    registerSensor("current_version", new Gauge(() -> currentVersion));
-    registerSensor("future_version", new Gauge(() -> futureVersion));
+    if (!isSystemStore) {
+      registerSensor("current_version", new Gauge(() -> currentVersion));
+      registerSensor("future_version", new Gauge(() -> futureVersion));
+    }
 
     this.currentStatsReporter = statsSupplier.get(metricsRepository, storeName + "_current");
     if (!isSystemStore) {
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/VeniceVersionedStatsReporter.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/VeniceVersionedStatsReporter.java
@@ -7,14 +7,10 @@ import com.linkedin.venice.stats.AbstractVeniceStats;
 import com.linkedin.venice.stats.Gauge;
 import com.linkedin.venice.stats.StatsSupplier;
 import io.tehuti.metrics.MetricsRepository;
-import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.function.DoubleSupplier;
 
 
 public class VeniceVersionedStatsReporter<STATS, STATS_REPORTER extends AbstractVeniceStatsReporter<STATS>>
     extends AbstractVeniceStats {
-  private static final AtomicBoolean IS_VERSION_STATS_SETUP = new AtomicBoolean(false);
-
   private int currentVersion = NON_EXISTING_VERSION;
   private int futureVersion = NON_EXISTING_VERSION;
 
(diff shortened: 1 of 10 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/VeniceVersionedStatsReporter.java
```diff
@@ -7,14 +7,10 @@
...
 import com.linkedin.venice.stats.StatsSupplier;
 import io.tehuti.metrics.MetricsRepository;
-import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.function.DoubleSupplier;
 
 
 public class VeniceVersionedStatsReporter<STATS, STATS_REPORTER extends AbstractVeniceStatsReporter<STATS>>
     extends AbstractVeniceStats {
-  private static final AtomicBoolean IS_VERSION_STATS_SETUP = new AtomicBoolean(false);
-
```
huangminchn, line 17: Why do we remove this? Could you please share more details?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/VeniceVersionedStatsReporter.java
```diff
@@ -31,20 +27,9 @@ public VeniceVersionedStatsReporter(
...
-     * we could have a centralized stats version dispatcher that listens to Helix data changes,
-     * keeps track of store version and notifies each stats.
-     * TODO: get ride of {@link IS_VERSION_STATS_SETUP}
-     */
-    if (IS_VERSION_STATS_SETUP.compareAndSet(false, true)) {
-      registerSensor("current_version", new VersionStat(() -> (double) currentVersion));
-      if (!isSystemStore) {
-        registerSensor("future_version", new VersionStat(() -> (double) futureVersion));
-      }
+    if (!isSystemStore) {
```
gaojieliu, line 30: Shall we emit this for system store as well since it is very cheap?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/AbstractStorageEngine.java
```diff
@@ -356,7 +356,7 @@ public synchronized void close() throws VeniceException {
     tmpList.parallelStream().forEach(Partition::close);
     LOGGER.info(
         "Closing {} rockDB partitions of store {} took {} ms",
-        partitionList.size(),
```
gaojieliu, line 359: Nit: no difference with/without this change, right?

majisourav99: there is, it prints 0 for spart list.

gaojieliu: I see.
So sparse list doesn't implement this method properly, so either we should reimplement that method or throw exception to claim it is not supported, otherwise, it is error prone.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/VeniceVersionedStatsReporter.java
majisourav99: Sure

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/AbstractStorageEngine.java
majisourav99: Actually sparseList.size would return size of array even if its sparsely populated. So if a host has partition 1 and 100. size will return 101.

gaojieliu: Got it.
It is still not properly implemented from my understanding.
@FelixGV 
Shall we re-implement it to return the actual element count? since array.size is more like an implementation detail, which shouldn't change the semantics of `size` function.

majisourav99: Actually Felix implemented a function `nonNullSize` for this purpose.

gaojieliu: Shouldn't switch the semantics instead since `#size` is supposed to return `non-null` element num.

majisourav99: For sparse list we can 2 types of size. 
1.  the underlying array's size (capacity) which includes both empty and non-empty slots.
2.  nonNullSize which Felix added. 
Are you asking if size should be same as nonNullSize?

FelixGV: I don't know if there is an easy answer to this one... because as a list, I think it's reasonable for a user to expext having seen every element if they do this:

```
for (int i = 0; i < list.size(); i++) {
  list.get(i);
}
```

The above does indeed traverse all elements, as a user might reasonably expect. If we make `size` return the non null size, then the above traversal code is likely to short-circuit too soon (i.e. unless the list happens to be densely populated).

So... directionally, I think the clean thing to do is to evolve this SCL utility into a full-fledged Map, rather than a List. Then it could become less confusing. As of now, it is kind of in-between a Map and a List, and that causes confusion.

gaojieliu: The underlying array is an implementation detail and the `size` function is supposed to respect API contract defined by the interface from my understanding:
```
    /**
     * Returns the number of elements in this list.  If this list contains
     * more than <tt>Integer.MAX_VALUE</tt> elements, returns
     * <tt>Integer.MAX_VALUE</tt>.
     *
     * @return the number of elements in this list
     */
    int size();
```
I supposed the above `number of elements` are the ones we explicitly put in, right?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/VeniceVersionedStatsReporter.java
```diff
@@ -31,21 +27,8 @@ public VeniceVersionedStatsReporter(
 
     this.isSystemStore = VeniceSystemStoreUtils.isSystemStore(storeName);
 
-    /**
-     * All stats which are aware of version should line up with Helix listener. It's duplicate
```
adamxchen, line 35: what's the reason to remove this?

majisourav99: Its replaced by the new metric. Previously it was using some shady `static final AtomicBoolean` for no reason.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/AbstractStorageEngine.java
majisourav99: I see your point. Currently `size` kinda returns the capacity of the array.  But developers could easily make mistakes if they want to iterate the array using for loop with size() and miss to visit all the elements. BTW can we separate the discussion from this PR ?

gaojieliu: Sure, that will be fine.
Please cut a ticket to track this since this API is error-prone from the user POV.

gaojieliu (approved):

PR #802, opened 2023-12-14
[server][dvc] Lag measurement safeguard and logging improvements

FelixGV:
[server][dvc] Lag measurement safeguard and logging improvements

It seems that sometimes the offset lag used in lag measurement can
be negative, which results in the lag being considered acceptable
and the replica becoming ready to serve (prematurely).

It seems that negative lag is a somewhat normal phenomenon since
the formula is basically max_offset - consumed_offset, and the
former is cached, so if stale it could be smaller than the latter
(which of course doesn't make sense, but is an artifact of the
caching). This is a benign case.

However, there is also a case where the max_offset retrieval can
fail, in which case a negative error code would be returned. This
also leads to negative lag, but it is a mistake.

This commit adds safeguards to prevent prematurely declaring
ready-to-serve. There are a bunch of code paths interacting with
offset retrieval, which makes this commit a bit big. In order to
improve maintainability, the tricky bits are centralized as much
as possible into a new SIT::measureLagWithCallToPubSub function.

Miscellaneous:

- Added a get_partition_latest_offset_with_retry_error metric.

- Deleted the SLOPPY_OFFSET_CATCHUP_THRESHOLD in SIT, which is
  not necessary and is, well, sloppy.

- Introduced a new EmptyPubSubMessageHeaders which is immutable
  (the previous EMPTY_MSG_HEADERS in VeniceWriter was mutable).

- Fixed a NPE in AbstractVeniceStatsReporter.

- Fixed a concurrency issue about closing the non-threadsafe
  Kafka consumer in PartitionOffsetFetcher.

- Fixed the sensor name of KafkaConsumerServiceStats's offset 
  present/absent metrics. Added a test for this as well.

- Tweaked the API in AbstractVeniceStats so that it is less
  error-prone. The registerSensor function with sensorFullName
  parameter is now private, to avoid misuse.

There are also various log tweaks to make logs less voluminous:

- Deleted two log lines from SIT::getDefaultReadyToServeChecker
  - "Reopen partition {}_{} for reading after ready-to-serve."
  - "Partition {} synced offset"

- Passed shouldLogLag false in SIT::updateOffsetLagInMetadata.

- Deleted two functions from LFSIT which were used only to 
  specify default values to some parameters in a deeper 
  function. Kept only measureRTOffsetLagForSingleRegion, which 
  is the first function of the call chain.

- Rewrote the TopicPartitionsOffsetsTracker so that it is a bit 
  more efficient, cannot have invalid lag, doesn't needlessly 
  log a bunch of lines every minute or so, and uses less code.

Test changes:

- Overhauled DataProviders in StoreIngestionTaskTest. Rather
  than using a bunch of booleans which are hard reason about
  in test logs, we now use enums, which makes it easy to read
  which test permutations passed or failed. In one case, this 
  allows squashing 2 booleans (of which 1 of 4 permutations is
  invalid) into a single 3 items enum, thus reducing the test 
  permutations count.

- Fixed testOffsetSyncBeforeGracefulShutDown flakiness in 
  StoreIngestionTaskTest.

- SITTest::testGetAndUpdateLeaderCompletedState which was 
  mutating the static instance of VW.EMPTY_MSG_HEADERS, thus 
  contaminating subsequent tests (when running in the same JVM,
  as is the case in IntelliJ, but not in Gradle).

- Added a unit test for parent stats.

## How was this PR tested?
Existing tests. Broke them a few times before finding a way that the safeguard doesn't cause failures.

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
@@ -1,5 +1,6 @@
 package com.linkedin.davinci.kafka.consumer;
 
+import static com.linkedin.davinci.ingestion.LagType.OFFSET_LAG;
 import static com.linkedin.davinci.kafka.consumer.ConsumerActionType.LEADER_TO_STANDBY;
 import static com.linkedin.davinci.kafka.consumer.ConsumerActionType.STANDBY_TO_LEADER;
 import static com.linkedin.davinci.kafka.consumer.LeaderFollowerStateType.IN_TRANSITION_FROM_STANDBY_TO_LEADER;
@@ -19,6 +20,7 @@ import static java.util.concurrent.TimeUnit.MINUTES;
 import com.linkedin.davinci.client.DaVinciRecordTransformer;
 import com.linkedin.davinci.config.VeniceStoreVersionConfig;
 import com.linkedin.davinci.helix.LeaderFollowerPartitionStateModel;
+import com.linkedin.davinci.ingestion.LagType;
 import com.linkedin.davinci.schema.merge.CollectionTimestampMergeRecordHelper;
 import com.linkedin.davinci.schema.merge.MergeRecordHelper;
 import com.linkedin.davinci.storage.chunking.ChunkedValueManifestContainer;
@@ -1915,13 +1917,13 @@ public class LeaderFollowerStoreIngestionTask extends StoreIngestionTask {
       long offsetLag,
       long offsetThreshold,
       boolean shouldLogLag,
-      boolean isOffsetBasedLag,
+      LagType lagType,
       long latestConsumedProducerTimestamp) {
     boolean isLagAcceptable = offsetLag <= offsetThreshold;
 
     if (shouldLogLag) {
       logLag(
-          isOffsetBasedLag,
+          lagType,
           pcs.getPartition(),
           isLagAcceptable,
           latestConsumedProducerTimestamp,
@@ -1934,7 +1936,7 @@ public class LeaderFollowerStoreIngestionTask extends StoreIngestionTask {
   }
 
   protected void logLag(
-      boolean isOffsetBasedLag,
+      LagType lagType,
       int partition,
       boolean isLagAcceptable,
       long latestConsumedProducerTimestamp,
@@ -1944,10 +1946,10 @@ public class LeaderFollowerStoreIngestionTask extends StoreIngestionTask {
     LOGGER.info(
         "{} [{} lag] partition {} is {}. {}Lag: [{}] {} Threshold [{}]. {}",
         this.consumerTaskId,
-        isOffsetBasedLag ? "Offset" : "Time",
+        lagType.prettyString(),
         partition,
         (isLagAcceptable ? "not lagging" : "lagging"),
-        (isOffsetBasedLag ? "" : "The latest producer timestamp is " + latestConsumedProducerTimestamp + ". "),
+        (lagType == OFFSET_LAG ? "" : "The latest producer timestamp is " + latestConsumedProducerTimestamp + ". "),
         lag,
         (isLagAcceptable ? "<" : ">"),
         threshold,
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
@@ -1924,19 +1924,38 @@ public class LeaderFollowerStoreIngestionTask extends StoreIngestionTask {
     boolean isLagAcceptable = offsetLag <= offsetThreshold;
 
     if (shouldLogLag) {
-      LOGGER.info(
-          "{} [{} lag] partition {} is {}lagging. {}Lag: [{}] {} Threshold [{}]",
-          isOffsetBasedLag ? "Offset" : "Time",
-          consumerTaskId,
+      logLag(
+          isOffsetBasedLag,
           pcs.getPartition(),
-          (isLagAcceptable ? "not " : ""),
-          (isOffsetBasedLag ? "" : "The latest producer timestamp is " + latestConsumedProducerTimestamp + ". "),
+          isLagAcceptable,
+          latestConsumedProducerTimestamp,
           offsetLag,
-          (isLagAcceptable ? "<" : ">"),
-          offsetThreshold);
+          offsetThreshold,
+          "");
     }
 
     return isLagAcceptable;
+  }
+
+  protected void logLag(
+      boolean isOffsetBasedLag,
+      int partition,
+      boolean isLagAcceptable,
+      long latestConsumedProducerTimestamp,
+      long offsetLag,
+      long offsetThreshold,
+      String lagLogFooter) {
+    LOGGER.info(
+        "{} [{} lag] partition {} is {}lagging. {}Lag: [{}] {} Threshold [{}]{}",
+        this.consumerTaskId,
+        isOffsetBasedLag ? "Offset" : "Time",
+        partition,
+        (isLagAcceptable ? "not " : ""),
+        (isOffsetBasedLag ? "" : "The latest producer timestamp is " + latestConsumedProducerTimestamp + ". "),
+        offsetLag,
+        (isLagAcceptable ? "<" : ">"),
+        offsetThreshold,
+        lagLogFooter);
   }
 
   /**
@@ -3272,6 +3291,14 @@ public class LeaderFollowerStoreIngestionTask extends StoreIngestionTask {
         ? offsetFromConsumer
         : cachedPubSubMetadataGetter
             .getOffset(getTopicManager(sourceRealTimeTopicKafkaURL), leaderTopic, partitionToGetLatestOffsetFor);
+
+    if (lastOffsetInRealTimeTopic < 0) {
+      LOGGER.warn(
+          "Unexpected! Got a negative lastOffsetInRealTimeTopic ({})! Will return Long.MAX_VALUE as the lag.",
+          lastOffsetInRealTimeTopic,
+          new VeniceException("Exception not thrown, just for logging purposes."));
+      return Long.MAX_VALUE;
+    }
 
     if (latestLeaderOffset == -1) {
       // If leader hasn't consumed anything yet we should use the value of 0 to calculate the exact offset lag.
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
@@ -99,6 +99,7 @@ import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.BiConsumer;
 import java.util.function.BooleanSupplier;
+import java.util.function.LongPredicate;
 import java.util.function.Predicate;
 import java.util.function.Supplier;
 import org.apache.avro.generic.GenericRecord;
@@ -715,7 +716,7 @@ public class LeaderFollowerStoreIngestionTask extends StoreIngestionTask {
    */
   private boolean isLocalVersionTopicPartitionFullyConsumed(PartitionConsumptionState pcs) {
     long localVTOff = pcs.getLatestProcessedLocalVersionTopicOffset();
-    long localVTEndOffset = getKafkaTopicPartitionEndOffSet(localKafkaServer, versionTopic, pcs.getPartition());
+    long localVTEndOffset = getTopicPartitionEndOffSet(localKafkaServer, versionTopic, pcs.getPartition());
     if (localVTEndOffset == StatsErrorCode.LAG_MEASUREMENT_FAILURE.code) {
       return false;
     }
@@ -1066,16 +1067,16 @@ public class LeaderFollowerStoreIngestionTask extends StoreIngestionTask {
       }
     }
     if (!isDataRecoveryCompleted) {
-      long dataRecoverySourceVTEndOffset = cachedPubSubMetadataGetter.getOffset(
-          topicManagerRepository.getTopicManager(nativeReplicationSourceVersionTopicKafkaURL),
-          versionTopic,
-          partitionConsumptionState.getPartition());
-
-      // If the last few records in source VT is old then we can also complete data recovery if the leader idles and we
+      // If the last few records in source VT is old then we can also complete data recovery if the leader idles, and we
       // have reached/passed the end offset of the VT (around the time when ingestion is started for that partition).
-      long localVTOffsetProgress = partitionConsumptionState.getLatestProcessedLocalVersionTopicOffset();
-      // -2 because getOffset returns the next available offset (-1) and we are skipping the remote TS message (-1).
-      boolean isAtEndOfSourceVT = dataRecoverySourceVTEndOffset - 2 <= localVTOffsetProgress;
+      // We subtract 1 because we are skipping the remote TS message.
+      long localVTOffsetProgress = partitionConsumptionState.getLatestProcessedLocalVersionTopicOffset() - 1;
+      long dataRecoveryLag = measureLagWithCallToPubSub(
+          nativeReplicationSourceVersionTopicKafkaURL,
+          versionTopic,
+          partitionConsumptionState.getPartition(),
+          localVTOffsetProgress);
+      boolean isAtEndOfSourceVT = dataRecoveryLag <= 0;
       long lastTimestamp = getLastConsumedMessageTimestamp(partitionConsumptionState.getPartition());
       if (isAtEndOfSourceVT && LatencyUtils.getElapsedTimeInMs(lastTimestamp) > newLeaderInactiveTime) {
         LOGGER.info(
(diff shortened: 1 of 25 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -3268,6 +3287,14 @@ private long getLatestLeaderOffsetAndHybridTopicOffset(
         : cachedPubSubMetadataGetter
             .getOffset(getTopicManager(sourceRealTimeTopicKafkaURL), leaderTopic, partitionToGetLatestOffsetFor);
 
+    if (lastOffsetInRealTimeTopic < 0) {
+      LOGGER.warn(
```
majisourav99, line 3291: can we rate limit this logging?

FelixGV: We can... but IDK if we should 🤔 ... maybe it should be a metric instead 🤷 ?

FelixGV: Actually there is already a metric for that called `getLatestOffsetIsAbsent` which is emitted from the `KafkaConsumerService`, called just above. So I guess it's fine to rate limit this log.

FelixGV: No that's not correct, actually... the `getLatestOffsetIsAbsent` will not be called in cases where get into this if block. I may want another metric for that...

m-nagarajan (commented):
Thanks @FelixGV. LGTM, just left a NIT comment.

FelixGV:
> Thanks @FelixGV. LGTM, just left a NIT comment.

Thanks Manoj but I'm still iterating on this... I believe there are more edge cases related to negative lag which I'm still trying to resolve. The code is unfortunately quite tangled which makes my progress slow ): ...

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -1921,21 +1909,40 @@ protected boolean checkAndLogIfLagIsAcceptableForHybridStore(
...
+  protected void logLag(
+      boolean isOffsetBasedLag,
+      int partition,
+      boolean isLagAcceptable,
+      long latestConsumedProducerTimestamp,
+      long offsetLag,
+      long offsetThreshold,
+      String lagLogFooter) {
+    LOGGER.info(
+        "{} [{} lag] partition {} is {}lagging. {}Lag: [{}] {} Threshold [{}]{}",
```
majisourav99, line 1934: nit: can we put spaces between `{}` and messages? also `logLag` sounds bit funny, maybe rename to something like `logOffsetLag` ?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -3271,6 +3268,16 @@ private long getLatestLeaderOffsetAndHybridTopicOffset(
         : cachedPubSubMetadataGetter
             .getOffset(getTopicManager(sourceRealTimeTopicKafkaURL), leaderTopic, partitionToGetLatestOffsetFor);
 
+    if (lastOffsetInRealTimeTopic < 0) {
+      if (!REDUNDANT_LOGGING_FILTER.isRedundantException("Got a negative lastOffsetInRealTimeTopic")) {
+        LOGGER.warn(
+            "Unexpected! Got a negative lastOffsetInRealTimeTopic ({})! Will return Long.MAX_VALUE as the lag.",
```
majisourav99, line 3274: can we also print the value of Long.MAX_VALUE for easier comparison from the log.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
FelixGV: In the final string there is nothing stuck to anything else. That being said, I refactored the formatting a bit to minimize cases where `{}` is stuck to the next token...

The function is also used for time lag, so putting offset in the name seems wrong to me... but I'm open to other name suggestions if any.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
FelixGV: Sure.

clients/da-vinci-client/src/test/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTaskTest.java
```diff
@@ -2263,16 +2264,34 @@ public void testDelayedTransitionToOnlineInHybridMode(boolean isActiveActiveRepl
...
+      }
       for (int i = 0; i < MESSAGES_BEFORE_EOP; i++) {
         try {
-          localVeniceWriter.put(getNumberedKey(i), getNumberedValue(i), SCHEMA_ID).get();
+          CompletableFuture<PubSubProduceResult> future =
+              localVeniceWriter.put(getNumberedKey(i), getNumberedValue(i), SCHEMA_ID);
+          PubSubProduceResult result = future.get();
+          int partition = result.getPartition();
+          long offset = messageCountPerPartition[partition];
+          messageCountPerPartition[partition] = ++offset;
```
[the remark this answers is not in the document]
FelixGV: Done.

FelixGV:
So... I think I got all the code paths I wanted to get in the last commits... but now I'm getting failures in the unit tests and code cov CI task 😅 ... I'll have to check that out, so it's still not quite ready for review.

sushantmane:
>  It seems that negative lag is a somewhat normal phenomenon since the formula is basically max_offset - consumed_offset, and the former is cached, so if stale it could be smaller than the latter (which of course doesn't make sense, but is an artifact of the caching). This is a benign case.
 
 IIUC, we use cached offset only if metric based offset is not present. Metric based offset is usually not present when at least one record for the TP is not consumed yet. So the cases where we may not have metric based offset is when there is no new record to consume. In such cases, if we fall back on cached offset we should get the actual end offset of the TP and since there are no new records end offset should remain unchanged. So I am wondering under what circumstances we get a negative lag.

FelixGV:
> > It seems that negative lag is a somewhat normal phenomenon since the formula is basically max_offset - consumed_offset, and the former is cached, so if stale it could be smaller than the latter (which of course doesn't make sense, but is an artifact of the caching). This is a benign case.
> 
> IIUC, we use cached offset only if metric based offset is not present. Metric based offset is usually not present when at least one record for the TP is not consumed yet. So the cases where we may not have metric based offset is when there is no new record to consume. In such cases, if we fall back on cached offset we should get the actual end offset of the TP and since there are no new records end offset should remain unchanged. So I am wondering under what circumstances we get a negative lag.

That's a good question. The consumer metric-based offset collection can be disabled via a server config but the default is true and AFAICT we run with true also. So I'm not sure what is the scenario where it falls back to the older mechanism...

FelixGV:
> > > It seems that negative lag is a somewhat normal phenomenon since the formula is basically max_offset - consumed_offset, and the former is cached, so if stale it could be smaller than the latter (which of course doesn't make sense, but is an artifact of the caching). This is a benign case.
> > 
> > 
> > IIUC, we use cached offset only if metric based offset is not present. Metric based offset is usually not present when at least one record for the TP is not consumed yet. So the cases where we may not have metric based offset is when there is no new record to consume. In such cases, if we fall back on cached offset we should get the actual end offset of the TP and since there are no new records end offset should remain unchanged. So I am wondering under what circumstances we get a negative lag.
> 
> That's a good question. The consumer metric-based offset collection can be disabled via a server config but the default is true (enabled) and AFAICT we run with true also. So I'm not sure what is the scenario where it falls back to the older mechanism...

Actually, the consumer-based offset collection also has a caching/stale behavior. See here: https://github.com/linkedin/venice/blob/44ff335878ebefd51e9190c09c8e0d1130318797/internal/venice-common/src/main/java/com/linkedin/venice/pubsub/adapter/kafka/TopicPartitionsOffsetsTracker.java#L73

sushantmane:
> > > > It seems that negative lag is a somewhat normal phenomenon since the formula is basically max_offset - consumed_offset, and the former is cached, so if stale it could be smaller than the latter (which of course doesn't make sense, but is an artifact of the caching). This is a benign case.
> > > 
> > > 
> > > IIUC, we use cached offset only if metric based offset is not present. Metric based offset is usually not present when at least one record for the TP is not consumed yet. So the cases where we may not have metric based offset is when there is no new record to consume. In such cases, if we fall back on cached offset we should get the actual end offset of the TP and since there are no new records end offset should remain unchanged. So I am wondering under what circumstances we get a negative lag.
> > 
> > 
> > That's a good question. The consumer metric-based offset collection can be disabled via a server config but the default is true (enabled) and AFAICT we run with true also. So I'm not sure what is the scenario where it falls back to the older mechanism...
> 
> Actually, the consumer-based offset collection also has a caching/stale behavior. See here:
> 
> https://github.com/linkedin/venice/blob/44ff335878ebefd51e9190c09c8e0d1130318797/internal/venice-common/src/main/java/com/linkedin/venice/pubsub/adapter/kafka/TopicPartitionsOffsetsTracker.java#L73

Oh I see. Thanks

sushantmane (approved):
LGTM. Thanks!

internal/venice-common/src/main/java/com/linkedin/venice/kafka/partitionoffset/PartitionOffsetFetcherStats.java
```diff
@@ -21,11 +21,12 @@ public enum OCCURRENCE_LATENCY_SENSOR_TYPE {
   }
 
   private final Map<OCCURRENCE_LATENCY_SENSOR_TYPE, Sensor> sensorsByTypes;
+  private Sensor getPartitionLatestOffsetError;
```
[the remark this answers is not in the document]
FelixGV: That's right, since this does not record latency, I thought the other stats didn't make sense.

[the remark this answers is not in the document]
sushantmane: Ack

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/adapter/kafka/TopicPartitionsOffsetsTracker.java
```diff
@@ -64,51 +55,81 @@ public TopicPartitionsOffsetsTracker(@Nonnull Duration offsetsUpdateInterval, Du
...
-      return Objects.equals(metricName.name(), "records-lag") && (metric.metricValue() instanceof Double);
+      if (Objects.equals(metricName.name(), "records-lag")) {
+        Object metricValue = metric.metricValue();
+        if (metricValue instanceof Double) {
+          // Double is the way all metrics are internally represented in Kafka, but since we are dealing with lag, we
+          // want an integral type, so we cast it.
+          return ((Double) metricValue).longValue();
+        }
+      }
+      return INVALID;
```
sushantmane, line 127: Since INVALID returned for all cases except when metricValue is Double we can put it in the end of function and have single `return INVALID`?

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/adapter/kafka/TopicPartitionsOffsetsTracker.java
sushantmane, line 1: Since we use `TopicPartitionsOffsetsTracker` for non-kafka clients as well (via duplication), I was wondering if we should consider abstracting out common logic so that we can avoid duplication?

sushantmane (dismissed):
LGTM. Left two comments. Thanks!

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/adapter/kafka/TopicPartitionsOffsetsTracker.java
```diff
@@ -64,51 +55,81 @@ public TopicPartitionsOffsetsTracker(@Nonnull Duration offsetsUpdateInterval, Du
...
-        .getElapsedTimeInMs(lastMetricsCollectedTime.toEpochMilli()) < offsetsUpdateInterval.toMillis()) {
+      Consumer<byte[], byte[]> kafkaConsumer) {
+    long lastMetricsCollectedTimeSnapshot = lastMetricsCollectedTime;
+    long currentTime = System.currentTimeMillis();
+    long elapsedTime = currentTime - lastMetricsCollectedTimeSnapshot;
+    if (elapsedTime < offsetsUpdateIntervalMs) {
       return; // Not yet
     }
-    lastMetricsCollectedTime = Instant.now();
+    synchronized (this) {
```
sushantmane, line 69: Since this is called from `Consumer::poll` which is already called with a mutex/synchronized, do we need another `synchronized`?

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/adapter/kafka/TopicPartitionsOffsetsTracker.java
```diff
@@ -64,51 +55,81 @@ public TopicPartitionsOffsetsTracker(@Nonnull Duration offsetsUpdateInterval, Du
...
   public void updateEndAndCurrentOffsets(
       ConsumerRecords<byte[], byte[]> records,
-      Map<MetricName, ? extends Metric> metrics) {
-    if (lastMetricsCollectedTime != null && LatencyUtils
-        .getElapsedTimeInMs(lastMetricsCollectedTime.toEpochMilli()) < offsetsUpdateInterval.toMillis()) {
+      Consumer<byte[], byte[]> kafkaConsumer) {
+    long lastMetricsCollectedTimeSnapshot = lastMetricsCollectedTime;
+    long currentTime = System.currentTimeMillis();
+    long elapsedTime = currentTime - lastMetricsCollectedTimeSnapshot;
+    if (elapsedTime < offsetsUpdateIntervalMs) {
```
sushantmane, line 66: Since offsets are fetched from metrics IMO the cost is relatively trivial to update offsets upon each poll request. Can we consider removing this check altogether?

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/adapter/kafka/TopicPartitionsOffsetsTracker.java
FelixGV: Yeah for sure... I was wondering about that, actually. How would this be handled for non-Kafka pub subs...

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/adapter/kafka/TopicPartitionsOffsetsTracker.java
FelixGV: That is true. I guess it is defensive coding, in case this assumption changes, though that seems unlikely. I could remove the thread safety check then.

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/adapter/kafka/TopicPartitionsOffsetsTracker.java
sushantmane: For non-Kafka we'll only change the implementation of `getMetricEntryRecordsLag` (+ metric name and  it's type), everything will go in base.

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/adapter/kafka/TopicPartitionsOffsetsTracker.java
FelixGV: Hmm... possibly, yeah. I guess that would depend on a few factors, including:

1. What is the batch size? If the batch size is small (in the most extreme case, containing just a single record), then we would pay this overhead in a less amortized way.
2. How many metrics are there in the client? My understanding is that Kafka clients carry a fairly large number of metrics, so iterating over each one and comparing the name for each might actually be costly... but there may be another algorithm that is more efficient, e.g. hanging on to the `KafkaMetric` object itself rather than copying its content...

So I guess we could expose the param to tune the time interval, and try lowering it to see if the cost is noticeable. I could remove the safeguard against making the interval zero, since you pointed out that it is single threaded anyway.

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/adapter/kafka/TopicPartitionsOffsetsTracker.java
sushantmane: We can make `getMetricEntryRecordsLag` abstract method
```java
@Override
getMetricEntryRecordsLag(PubSubTopicPartition topicPartition) {
 //pus-sub specific logic that includes metric name formats and related stuff
}

```

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/adapter/kafka/TopicPartitionsOffsetsTracker.java
sushantmane: Good point. I agree it would be better to make interval tunable instead of removing it completely. We can do it in follow-up PR as well. Thanks

sushantmane (approved):
LGTM. Thanks a lot, @FelixGV!

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/adapter/kafka/TopicPartitionsOffsetsTracker.java
sushantmane: As per offline discussion we will take up this in a follow up PR

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/adapter/kafka/TopicPartitionsOffsetsTracker.java
sushantmane: As per offline discussion we will take up this in a follow up PR

PR #803, opened 2023-12-16
[dvc] Check to clear local data on record transformer value schema change between reinitializations

elijahgrimaldi:
In order to make sure that data is always ingested and transformed with the most up to date value output schema, we need to add an additional check in  `hasConflictPersistedStoreEngineConfig() ` to compare the new schema and the current schema. If there is a conflict in the schema, this function will return true.
Describe


ZacAttack:
Some conversation here: https://venicedb.slack.com/archives/C06ACQXR0JV/p1704840769706859

Basically, instead of adding a new config for the schema, let's instead consult the RecordTransformer that we already have.

ZacAttack (approved):
Ok!  Sorry for my confusion.  This looks good!

ZacAttack (approved):

PR #804, opened 2023-12-19
[WIP][da-vinci] Make chunk assembly a utility

kvargha:

This moves chunk deserialization and assembly to a utility, so it can be used in different areas in the da-vinci client. This is because other areas face the issue of message sizes being too large to be sent under one message, so they need to be serialized and chunked.

## How was this PR tested?

CI


ZacAttack:
Just a quick thing,

I 'think' you should make a new local branch or merge the branch your working on with main, this PR seems to be recording the changes from your last change.

kvargha:
New PR opened [here](https://github.com/linkedin/venice/pull/805)

PR #805, opened 2023-12-19
[da-vinci] Make chunk assembly a utility

kvargha:
## Summary

This moves chunk deserialization and assembly to a utility, so it can be used in different areas in da-vinci. This is because other areas face the issue of message sizes being too large to be sent under one message, so they need to be serialized and chunked.

## How was this PR tested?

CI

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/utils/ChunkAssembler.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/utils/ChunkAssembler.java
@@ -71,7 +71,7 @@ public class ChunkAssembler {
           keyBytes,
           ValueRecord.create(schemaId, valueBytes.array()).serialize());
       try {
-        assembledRecord = deserializeAndDecompressBytes(
+        assembledRecord = decompressAndDeserialize(
             recordDeserializer.get(),
             compressor,
             RawBytesChunkingAdapter.INSTANCE.get(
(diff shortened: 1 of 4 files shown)


kvargha:
@ZacAttack I removed the commits from my previous PR, but it looks like it forced closed the other PR I had so here's a new one.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/utils/ChunkAssembler.java
```diff
@@ -0,0 +1,150 @@
...
+import org.apache.logging.log4j.Logger;
+
+
+/*
+ * This class serves as a utility to deserialize and assemble chunks consumed from a Kafka topic 
+ */
+public class ChunkAssembler {
+  private static final Logger LOGGER = LogManager.getLogger(ChunkAssembler.class);
+
+  protected static final VeniceCompressor NO_OP_COMPRESSOR = new NoopCompressor();
```
ZacAttack, line 28: Do we need this here?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/utils/ChunkAssembler.java
```diff
@@ -0,0 +1,150 @@
...
+  protected final HashMap<Integer, VeniceCompressor> compressorMap = new HashMap<>();
```
ZacAttack, line 29: Same with this.

kvargha: Thank you for catching these!

kvargha: Should the LOGGER be kept since it's being used in `bufferAndAssembleRecord`?

ZacAttack: logger is ok to keep I think

ZacAttack (approved):

PR #807, opened 2023-12-20
[dvc] Reduce the number of successful heartbeat info logs

m-nagarajan:
## Summary
Heartbeat successfully sent is logged per topic partitions leading to a lot of logs per minute if the server hosts a lot of topic partitions.
`maybeSendIngestionHeartbeat()` which sends HB to all RT partitions will generate either 1 error log if there are failures or 1 debug log if everything is successful.
`propagateHeartbeatFromUpstreamTopicToLocalVersionTopic()` which forwards HB from RT to VT sends 1 error/debug log per topic partition if it succeeds/fails respectively. 
Both are protected inside a redundant logging filter (1 min).

## How was this PR tested?
GH CI

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
@@ -15,7 +15,6 @@ import static com.linkedin.venice.writer.VeniceWriter.DEFAULT_LEADER_METADATA_WR
 import static java.lang.Long.max;
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
 import static java.util.concurrent.TimeUnit.MINUTES;
-import static java.util.concurrent.TimeUnit.SECONDS;
 
 import com.linkedin.davinci.client.DaVinciRecordTransformer;
 import com.linkedin.davinci.config.VeniceStoreVersionConfig;
@@ -91,12 +90,13 @@ import java.util.Optional;
 import java.util.Properties;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.CompletionException;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.BiConsumer;
 import java.util.function.BiFunction;
 import java.util.function.BooleanSupplier;
@@ -140,7 +140,7 @@ import org.apache.logging.log4j.Logger;
  *           follower can subscribe back to VT using the recently updated VT offset.
  */
 public class LeaderFollowerStoreIngestionTask extends StoreIngestionTask {
-  private static final Logger LOGGER = LogManager.getLogger(LeaderFollowerStoreIngestionTask.class);
+  static Logger LOGGER = LogManager.getLogger(LeaderFollowerStoreIngestionTask.class);
 
   /**
    * The new leader will stay inactive (not switch to any new topic or produce anything) for
@@ -1910,9 +1910,8 @@ public class LeaderFollowerStoreIngestionTask extends StoreIngestionTask {
   /**
    * Checks whether the lag is acceptable for hybrid stores
    * <p>
-   * leaderCompleteStateCheckEnabled is not used for non AA stores: as consumer task is unable to read HB messages from
-   * RT topic (though a test consumer can still read it), leading to standby replicas waiting for leader completion
-   * state header indefinitely, so disabling it until that issue is resolved
+   * leaderCompleteStateCheckEnabled is not used for non AA stores: as SIT reads from parent RT for non AA
+   * stores but HB is written to local RT. This should be enabled once all hybrid stores are made AA.
    */
   @Override
   protected boolean checkAndLogIfLagIsAcceptableForHybridStore(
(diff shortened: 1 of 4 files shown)


adamxchen:
Thank you! I've been noticing this for a while but didn't have time to make the change. @m-nagarajan  I am wondering whether we can make it DEBUG level. The reason is probably we only need to check the heartbeat msg when something goes off and since the heartbeat msg is sent per minute, devs only need to wait for a minute or two to see the logging.

FelixGV (commented):
Is there a way to put more info in this log? Otherwise I'm afraid it may not be useful enough... What I have in mind is:

If a HB has been sent for all partitions that needed to send one within the past minute, we log that. If only a subset of those sent HBs, then we log which partitions didn't get it or something like that (still in just one log line though).

Maybe it's possible to put the log after the end of the partition loop in `maybeSendIngestionHeartbeat`?

m-nagarajan:
> Is there a way to put more info in this log? Otherwise I'm afraid it may not be useful enough... What I have in mind is:
> 
> If a HB has been sent for all partitions that needed to send one within the past minute, we log that. If only a subset of those sent HBs, then we log which partitions didn't get it or something like that (still in just one log line though).
> 
> Maybe it's possible to put the log after the end of the partition loop in `maybeSendIngestionHeartbeat`?

Thanks for the review.

@FelixGV Added more info at the end of `maybeSendIngestionHeartbeat`
@adamxchen changed the successful logs to debug

clients/da-vinci-client/src/test/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTaskTest.java
```diff
@@ -4028,6 +4033,111 @@ public void testMaybeSendIngestionHeartbeat(
...
+            mockKafkaConsumerProperties,
+            () -> true,
+            mockVeniceStoreVersionConfig,
+            0,
+            false,
+            Optional.empty(),
+            null);
+    ingestionTask.setPartitionConsumptionState(0, pcs0);
+    ingestionTask.setPartitionConsumptionState(1, pcs1);
+    Logger mockLogger = mock(Logger.class);
```
FelixGV, line 4100: I'm not a big fan of having tests for logging... I understand in this case the only change is to logging and so the principle of "there should always be a unit test that fails without the change" points towards adding a test for logging. But I feel like this could be treated as an exception... if the modified code is covered by tests, it could be sufficient, no need to actually have a test that fails when the logging is different.

These are just my two cents, and I don't think we need to delete this new test, but this is more of a forward looking comment (I don't think we need to write logging tests in the future...). Feel free to disagree, of course, though.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -3443,4 +3514,9 @@ void reportCompleted(PartitionConsumptionState partitionConsumptionState, boolea
       lastSendIngestionHeartbeatTimestamp.set(0);
     }
   }
+
+  // visible for testing
+  public static void setLogger(Logger logger) {
```
FelixGV, line 3519: We should make this package-private rather than public, if possible.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -136,7 +140,7 @@
  *           follower can subscribe back to VT using the recently updated VT offset.
  */
 public class LeaderFollowerStoreIngestionTask extends StoreIngestionTask {
-  private static final Logger LOGGER = LogManager.getLogger(LeaderFollowerStoreIngestionTask.class);
+  static Logger LOGGER = LogManager.getLogger(LeaderFollowerStoreIngestionTask.class);
```
FelixGV, line 143: Can it remain private, even though it is not final anymore? Technically, we should make it in lower case since it's not a constant anymore, but I'm fine with omitting that in the interest of minimizing the diff size...

FelixGV (commented):
LGTM overall. Just left a couple of minor nitpicks. Please take a look.

clients/da-vinci-client/src/test/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTaskTest.java
m-nagarajan: Thanks @FelixGV, read more about this and it makes sense to not have tests on logging and gave it an attempt to follow it from this PR itself. Modified the `maybeSendIngestionHeartbeat` API to be able to test these cases. Let me know what you think.

FelixGV (commented):
Thanks, the new approach is a bit more maintainable I think, in the sense that we can now tweak the logging format without needing to update tests. I left a request to add to the JavaDoc, but I think we're pretty much ready to merge here.

FelixGV (approved):
Thanks a lot! LGTM!

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -3430,7 +3430,7 @@ private CompletableFuture<PubSubProduceResult> sendIngestionHeartbeat(
    * in the RT, avoiding the need to differentiate between heartbeats from leaders of different versions (backup/current/future) and colos.
    */
   @Override
-  protected void maybeSendIngestionHeartbeat() {
+  protected void maybeSendIngestionHeartbeat(Optional<Set<String>> failedPartitions) {
```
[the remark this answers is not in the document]
m-nagarajan: Returning the set looks cleane, so opted for it and updated the JavaDoc for it.

m-nagarajan (commented):
Thanks @FelixGV and @adamxchen for the review 🚀

PR #808, opened 2023-12-21
[test] Use targeted region push in conjunction with repush

adamxchen:
## Summary

Maybe it doesn't worth a PR but we'd like to know whether targeted reigon push can work with repush and turned out our integ tests don't have that.

So, this PR adds this part so we can ensure this compatibility doesn't break.

Also tried to fix a flaky unit test..

## How was this PR tested?

GHCI.

### DIFF
--- a/services/venice-controller/src/test/java/com/linkedin/venice/controller/TestDisabledPartitionEnablerService.java
+++ b/services/venice-controller/src/test/java/com/linkedin/venice/controller/TestDisabledPartitionEnablerService.java
@@ -2,7 +2,10 @@ package com.linkedin.venice.controller;
 
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.anyString;
-import static org.mockito.Mockito.*;
+import static org.mockito.Mockito.doReturn;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.timeout;
+import static org.mockito.Mockito.verify;
 
 import com.linkedin.venice.utils.TestMockTime;
 import java.util.HashSet;
@@ -30,8 +33,7 @@ public class TestDisabledPartitionEnablerService {
     DisabledPartitionEnablerService service = new DisabledPartitionEnablerService(admin, config, time);
 
     service.startInner();
-    verify(admin, timeout(1000).atLeast(1)).enableDisabledPartition(clusterName, "", true);
-
+    verify(admin, timeout(1000).atLeastOnce()).enableDisabledPartition(clusterName, "", true);
   }
 
 }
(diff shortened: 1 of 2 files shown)


adamxchen:
The `TestDisabledPartitionEnablerService` is flaky... We spin up a thread and check the mocked result immediately. Sometimes it'd pass but sometimes it'd fail as well.. I am working on a small fix for this.

services/venice-controller/src/test/java/com/linkedin/venice/controller/TestDisabledPartitionEnablerService.java
```diff
@@ -30,11 +34,32 @@ public void testCleanupBackupVersion() throws Exception {
...
+
+  }
+
+  class DisabledPartitionEnablerServiceForTesting extends DisabledPartitionEnablerService {
+    public DisabledPartitionEnablerServiceForTesting(
+        VeniceHelixAdmin admin,
+        VeniceControllerMultiClusterConfig multiClusterConfig,
+        Time time) {
+      super(admin, multiClusterConfig, time);
+      this.cleanupThread = new Thread(new DisabledPartitionEnablerTaskForTesting(), "TestTask");
```
adamxchen, line 52: This probably isn't very elegant but other ways I've tried would be disruptive. Plz lmk if we have better way to test such cases.

services/venice-controller/src/test/java/com/linkedin/venice/controller/TestDisabledPartitionEnablerService.java
```diff
@@ -30,11 +35,32 @@ public void testCleanupBackupVersion() throws Exception {
     doReturn(true).when(controllerConfig).isEnableDisabledReplicaEnabled();
     doReturn(clusters).when(config).getClusters();
     TestMockTime time = new TestMockTime();
-    DisabledPartitionEnablerService service = new DisabledPartitionEnablerService(admin, config, time);
+    DisabledPartitionEnablerService service = new DisabledPartitionEnablerServiceForTesting(admin, config, time);
 
     service.startInner();
+    Assert.assertTrue(latch.await(1, TimeUnit.SECONDS));
     verify(admin, atLeast(1)).enableDisabledPartition(clusterName, "", true);
+    service.stopInner();
```
lluwm, line 43: Not an expert on Mockito, do you think [VerificationWithTimeout](https://site.mockito.org/javadoc/current/index.html?org/mockito/verification/VerificationWithTimeout.html) help in this case?

adamxchen: Yes! I think so! That probably require less code changes.. Let me incorporate that.

adamxchen: Done! @lluwm

lluwm (approved):
LGTM and thank you for fixing this flaky one!

PR #810, opened 2024-01-02
[server] Restore data partition in isolated process in Venice Server

sixpluszero:
## [server] Restore data partition in isolated process in Venice Server

This PR tries to address the issue of cleaning up data correctly for ingestion isolation enabled server.
Originally, we observed that certain data partition is not deleted when a OFFLINE -> DROPPED message is issued for stale version upon server restart. The message is not executed as we did not restore all the data partitions upon server start in any of the processes. 
The first fix we implemented is in  #617 , where we added a clean up service in isolated process. However, it is not correct, as the clean up service tries to remove partitions opened in the forked process, which is metadata partition only. This leaves the data partitions still unopened and lingering forever.

This PR tries to fix it thoroughly by adjusting the data restore logic for II. It will by default restore everything in II process. For those partitions to receive OFFLINE->DROPPED message, it will be sent to isolated process and get executed. For old storage engines, it shall also be opened and should be cleaned up eventually, as restore metadata partition will recreate the metadata partition, even if it is deleted before.

## How was this PR tested?
Existing integration tests.
Need to consider adding a new unit test.

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/IsolatedIngestionBackend.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/IsolatedIngestionBackend.java
@@ -390,15 +390,9 @@ public class IsolatedIngestionBackend extends DefaultIngestionBackend
       Runnable mainProcessCommandRunnable) {
     boolean isTopicPartitionHosted = isTopicPartitionHosted(topicName, partition);
 
-    // drop storage partition in main process if it does not exist
-    if (command == REMOVE_PARTITION && !isTopicPartitionHosted) {
-      mainProcessCommandRunnable.run();
-      return;
-    }
-
     do {
       if (isTopicPartitionHostedInMainProcess(topicName, partition)
-          || !isTopicPartitionHosted && command != START_CONSUMPTION) {
+          || (!isTopicPartitionHosted && command != START_CONSUMPTION && command != REMOVE_PARTITION)) {
         LOGGER.info("Executing command {} of topic: {}, partition: {} in main process.", command, topicName, partition);
         mainProcessCommandRunnable.run();
         return;
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/IsolatedIngestionBackend.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/IsolatedIngestionBackend.java
@@ -395,7 +395,9 @@ public class IsolatedIngestionBackend extends DefaultIngestionBackend
        * There are two cases where we execute the command to the main process:
        * (1) The main process metadata tells that the topic partition is hosted in main process.
        * (2) The topic partition has never been associated with the server, and the incoming command is not START_CONSUMPTION
-       * or REMOVE_PARTITION. For these two commands, it should by default be sent to isolated process.
+       * or REMOVE_PARTITION. For these two commands, it should by default be sent to isolated process. Here it needs to
+       * make sure topic partition subscription request has never been issued to the server, so the check "isTopicPartitionHosted"
+       * should be false in this case.
        */
       if (isTopicPartitionHostedInMainProcess(topicName, partition)
           || (!isTopicPartitionHosted && command != START_CONSUMPTION && command != REMOVE_PARTITION)) {
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/IsolatedIngestionBackend.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/IsolatedIngestionBackend.java
@@ -391,6 +391,12 @@ public class IsolatedIngestionBackend extends DefaultIngestionBackend
     boolean isTopicPartitionHosted = isTopicPartitionHosted(topicName, partition);
 
     do {
+      /**
+       * There are two cases where we execute the command to the main process:
+       * (1) The main process metadata tells that the topic partition is hosted in main process.
+       * (2) The topic partition has never been associated with the server, and the incoming command is not START_CONSUMPTION
+       * or REMOVE_PARTITION. For these two commands, it should by default be sent to isolated process.
+       */
       if (isTopicPartitionHostedInMainProcess(topicName, partition)
           || (!isTopicPartitionHosted && command != START_CONSUMPTION && command != REMOVE_PARTITION)) {
         LOGGER.info("Executing command {} of topic: {}, partition: {} in main process.", command, topicName, partition);
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/isolated/IsolatedIngestionServer.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/isolated/IsolatedIngestionServer.java
@@ -735,7 +735,7 @@ public class IsolatedIngestionServer extends AbstractVeniceService {
         storeVersionStateSerializer,
         partitionStateSerializer,
         storeRepository,
-        false,
+        !isDaVinciClient,
         true);
     storageService.start();
 
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/isolated/IsolatedIngestionServer.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/isolated/IsolatedIngestionServer.java
@@ -715,19 +715,22 @@ public class IsolatedIngestionServer extends AbstractVeniceService {
         metricsRepository,
         storeRepository,
         serverConfig.isUnregisterMetricForDeletedStoreEnabled());
-    /**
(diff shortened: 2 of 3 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/IsolatedIngestionBackend.java
```diff
@@ -390,15 +390,9 @@ void executeCommandWithRetry(
...
-    // drop storage partition in main process if it does not exist
-    if (command == REMOVE_PARTITION && !isTopicPartitionHosted) {
-      mainProcessCommandRunnable.run();
-      return;
-    }
-
     do {
       if (isTopicPartitionHostedInMainProcess(topicName, partition)
-          || !isTopicPartitionHosted && command != START_CONSUMPTION) {
+          || (!isTopicPartitionHosted && command != START_CONSUMPTION && command != REMOVE_PARTITION)) {
```
gaojieliu, line 395: Can you add some javadoc here? I am not sure I fully understand this logic.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/isolated/IsolatedIngestionServer.java
```diff
@@ -735,7 +735,7 @@ private void initializeIsolatedIngestionServer() {
         storeVersionStateSerializer,
         partitionStateSerializer,
         storeRepository,
-        false,
+        !isDaVinciClient,
```
gaojieliu, line 738: So there is no data partition leaking issue in DaVinci?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/isolated/IsolatedIngestionServerHandler.java
```diff
@@ -201,6 +201,7 @@ private IngestionTaskReport handleIngestionTaskCommand(IngestionTaskCommand inge
           isolatedIngestionServer.cleanupTopicState(topicName);
           break;
         case REMOVE_PARTITION:
+          isolatedIngestionServer.maybeSubscribeNewResource(topicName, partitionId);
```
gaojieliu, line 204: Can you explain why need to call `subscribe` when removing a partition?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/isolated/IsolatedIngestionServer.java
sixpluszero: The issue is only triggered by Helix issuing OFFLINE->DROPPED message upon server restart. It's likely because the version becomes stale when server is down/being restarted. The case is not frequent, and this case should not happen to DVC.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/isolated/IsolatedIngestionServerHandler.java
sixpluszero: It is caused by the check mechanism implemented here. When the server started, nothing has been initialized. When II process received a command, it should check whether the TP has been "initialized" (subscribed), if it is either not subscribed or has been ingested and hand-over to main (unsubscribed), II process should reject the command, as it is unexpected. To adapt to the change in this PR, we need to let the II process accept the remove partition command if the TopicPartition has never been processed by II. This will be the same behavior as the OFFLINE -> STANDBY.
The command will only change the TP's subscription state to true if it is never found in the state map. When a TP is handed over to main process, it will be changed to false.

sixpluszero: I can also try to add some explanation in the PR

clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/IsolatedIngestionBackend.java
sixpluszero: Updated the Java Doc. Please feel free to take a 2nd look when you get some time.

gaojieliu: I understand for `START_PARTITION`, it is possible since Venice Server would like to bring up a new partition replica, but how can it be possible for `REMOVE_PARTITION`?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/isolated/IsolatedIngestionServer.java
gaojieliu: Got it, and let us add these statements as javadoc.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/IsolatedIngestionBackend.java
sixpluszero: The Helix might want to drop a stale topic partition's replica, so it could send an OFFLINE->DROPPED message. The message could arrive when server is restarted. Since we changed the behavior to restore data partition in II, the message should arrive at II process.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/isolated/IsolatedIngestionServerHandler.java
gaojieliu: The current logic sound weird to me..
So What is the logic when II is disabled to handle such case, does Venice Server need to try to subscribe? I don't recall such logic.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/IsolatedIngestionBackend.java
sixpluszero: isTopicPartitionHosted means if the TP has ever gone through the subscription flow (OFFLINE->STANDBY)
Since for the case, it is only the data RocksDB opened by II process, not going through the flow, and no metadata has been updated, so isTopicPartitionHosted will be false.

gaojieliu: I see.
Please add the above statement as javadoc and it is not clear from the naming of this variable.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/isolated/IsolatedIngestionServer.java
sixpluszero: Sure, updated the statement.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/isolated/IsolatedIngestionServerHandler.java
sixpluszero: All this subscribe thingy is for II only.
We build the subscription metadata for each topic partition in main process and isolated process, this is needed as it can direct any helix command to go to the place where the resource is hosted. This is for correctness guarantee. 
For a new resource that metadata management does not recognize, by default the OFFLINE->STANDBY goes to II. For this PR's purpose, it change to make OFFLINE->DROPPED by default goes to II as well.

For non II and one server process only, all the command goes to the same process so we don't need the metadata management at all.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/IsolatedIngestionBackend.java
sixpluszero: Yeah, updated the Java doc as well.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/isolated/IsolatedIngestionServerHandler.java
gaojieliu: It just looks weird to put this logic here.
Can the following `dropStoragePartitionGracefully` function ignore whether the to-be-dropped partition has subscribed before or not?
If we put it here, I think it will be confusing to other people even through it can run correctly.
People might get illusion that before dropping a particular partition, Venice Server needs to subscribe it first and subscription in Server normally means: create the partition, subscribe to Kafka and start ingestion and more.
I recommend either ignore the check in the following function call or rename this function to something else since `subscribe` is abused here from my understanding.

sixpluszero: I see your concern, let me think about it...

gaojieliu (approved):

PR #813, opened 2024-01-04
[test] Fix flaky test IngestionHeartBeatTest#testIngestionHeartBeat

sixpluszero:
## [test] Fix flaky test IngestionHeartBeatTest#testIngestionHeartBeat
This PR tries to fix the flaky test. 
The cause is: in some cases: when evaluating the lag, the standby replica might already know the leader topic and get the lag == 1, instead of negative max long. This is calculated by local VT end offset - last processed VT. 
In the current test setup, inc push is always enabled in the test, in the non-aggregate mode, it will be subscribing to parent RT, and will not consume HB. Relaxing the requirement is needed for non-AA case.

In the future, let's consider blocking the non-agg mode + inc push
Also, we don't need to test with NR = T / F as NR is always true in prod env for long time.


m-nagarajan (approved):
Thanks @sixpluszero for fixing this test 🚀

m-nagarajan:
[celebrate]     Manoj Nagarajan reacted to your message:
________________________________

PR #814, opened 2024-01-05
[fc][test] clean up useStreamingBatchGetAsDefault and associated tests

m-nagarajan:
## Summary
cleanup `useStreamingBatchGetAsDefault` config which enables using streaming backend for `batchGet` API, its related code and tests to make it the only supported behavior.

## How was this PR tested?
GH CI


xunyin8 (approved):
Thanks for the cleanup, let's make sure all the tests are still working.

PR #815, opened 2024-01-08
[da-vinci][server] Tweaked the offset lag threshold logic

gaojieliu:
Previously, the store-level 'offsetLagThresholdToGoOnline' config is being used as partition-level, which has caused a lot of confusion since the users are expecting this would apply in store-level. This code change would divide this configured value by partition count for current/future versions to align the expectation, and for the backup version, we will continue to use the original logic since backup version doesn't serve any read traffic, which means it won't cause any user impact and the original logic can help speed up the restart procedure.

This code also removed some deprecated logic for incremental push to VT.

## How was this PR tested?
CI


clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -1625,6 +1619,33 @@ protected TopicSwitch resolveSourceKafkaServersWithinTopicSwitch(TopicSwitch ori
...
+    }
+    boolean isBackupVersion = versionNumber < storeRepository.getStore(storeName).getCurrentVersion();
+
+    /**
+     * For the backup version, we would like to speed up the ready-to-serve check during restart, and
+     * it will be using configured `OffsetLagThresholdToGoOnline` in partition-level.
+     */
+    return isBackupVersion
+        ? hybridStoreConfig.get().getOffsetLagThresholdToGoOnline()
+        : lagThreshold / subPartitionCount;
```
xunyin8, line 1646: I remember we used to have off by 1 issues around lag calculation so we advised users to not set lag threshold to 0. Are we all clear with that? Since this division might return `0` for user stores with large amount of partitions. e.g. 2000 partitions and lag threshold of 1000. Might be worth to have an integration test to validate the 0 lag threshold scenario.

Alternatively we can also set our own lower limit of like 1 for partition level offset lag threshold.

gaojieliu: Good question, and I am not aware of this `0` lag issue.
Let us sync up internally.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -803,149 +805,141 @@ protected boolean isReadyToServe(PartitionConsumptionState partitionConsumptionS
...
+      if (producerTimeLagThresholdInSeconds > 0) {
+        long producerTimeLagThresholdInMS = TimeUnit.SECONDS.toMillis(producerTimeLagThresholdInSeconds);
+        long latestConsumedProducerTimestamp =
+            partitionConsumptionState.getOffsetRecord().getLatestProducerProcessingTimeInMs();
+        if (amplificationFactor != 1) {
+          latestConsumedProducerTimestamp = getLatestConsumedProducerTimestampWithSubPartition(
+              latestConsumedProducerTimestamp,
+              partitionConsumptionState);
+        }
+        boolean timestampLagIsAcceptable = checkAndLogIfLagIsAcceptableForHybridStore(
```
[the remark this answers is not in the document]
gaojieliu: It should be checking the time-lag since it gets a time-lag threshold param.
There are some conflicts I will need to rebase, and I think this is to execute time-lag measurement logic instead of offset lag.

xunyin8 (dismissed):
LGTM, thanks Gaojie

xunyin8 (approved):

PR #816, opened 2024-01-08
[common][server][fc] Fix trust store handling for Grpc with SSL

nisargthakkar:
## Fix trust store handling for Grpc with SSL

Currently, `GrpcSslUtils#getTrustManagers` loads the trust store but expects the trust store to be of the same type as the key store. While this is true in our tests, it might not always be true in practice. This PR fixes this by handling trust store types explicitly. 

## How was this PR tested?

GH CI. Tested in internal jobs.


ZacAttack (approved):
Looks good, but out of curiosity, which environment did you run into this??

nisargthakkar:
> out of curiosity, which environment did you run into this??
I was testing Spark job launching with one of our internal systems and it uses GRPC. So, I thought I'd reuse this util instead of rewriting the same thing in our internal codes

PR #819, opened 2024-01-09
[da-vinci-client] Catch all errors and exceptions in ConsumptionTask …

lluwm:
…to avoid silent exit

In corner cases, we have seen shared consumer thread could exit silently without reporting any reason. This change adds a catch-all block to ensure that the thread doesn't die silently, so if it does exit somehow, we want to know the reason.

## How was this PR tested?

CI


gaojieliu (approved):
Thanks!

We need to setup alert against stuck consumer repair failure metric, so that we can collect the new error log quickly to figure out what is happening.

PR #820, opened 2024-01-09
[dvc] Collect transformer latency in ingestion stats

elijahgrimaldi:
## Summary

We want to be able to track the latency when using the DaVinci record transformer feature. 

## How was this PR tested?

CI

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/IngestionStats.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/IngestionStats.java
@@ -58,6 +58,7 @@ public class IngestionStats {
   public static final String NEARLINE_PRODUCER_TO_LOCAL_BROKER_LATENCY = "nearline_producer_to_local_broker_latency";
   public static final String NEARLINE_LOCAL_BROKER_TO_READY_TO_SERVE_LATENCY =
       "nearline_local_broker_to_ready_to_serve_latency";
+  public static final String TRANSFORMER_LATENCY = "transformer_latency";
 
   private static final MetricConfig METRIC_CONFIG = new MetricConfig();
   private StoreIngestionTask ingestionTask;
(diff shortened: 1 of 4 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/IngestionStats.java
```diff
@@ -168,6 +170,7 @@ public IngestionStats(VeniceServerConfig serverConfig) {
         localMetricRepository,
         METRIC_CONFIG,
         NEARLINE_LOCAL_BROKER_TO_READY_TO_SERVE_LATENCY);
+    transformerLatencySensor = new WritePathLatencySensor(localMetricRepository, METRIC_CONFIG, TRANSFORMER_LATENCY);
```
ZacAttack, line 173: We need to register this sensor.  But ideally, we should only register this sensor if the transformer is passed and is being updated.

Let's wrap the register in a function and then invoke that function in the ingestion task if a transformer is passed.

elijahgrimaldi: @ZacAttack Do you mind re-running the tests, I think the integration test A is being flaky

elijahgrimaldi: And also let me know if this looks good to you

ZacAttack:
The failing test suite is stuck on an NPE, let's take a look at that, but otherwise this is looking good.

elijahgrimaldi:
> The failing test suite is stuck on an NPE, let's take a look at that, but otherwise this is looking good.

@ZacAttack Ya you're right, in the testBatchStore it looks like when testing with a record transformer it isn't properly initializing the latency sensor. I will look into it later today.

ZacAttack (approved):
Looks good!

PR #821, opened 2024-01-11
[test] Slight tweaks to ExceptionUtils

FelixGV:
Made recursiveMessageContains work even in an edge case where the message is null but a cause has the message we're looking for. Added a unit test for this. Note that this is the only functional change of this commit, but the recursiveMessageContains function is only used in tests, hence the tag of the commit is just [test]... The motivation for this change is that this function is being used in another project that depends on Venice.

Reduced code size (without functional changes) of recursiveClassEquals.

## How was this PR tested?
New unit test.


sushantmane (approved):
LGTM. Thanks, @FelixGV!

PR #822, opened 2024-01-13
[admin-tool] Added command to recover stores from graveyard and admin channel

gaojieliu:
    When we accidentally remove some being-used stores from the Venice platform, we
    can use the newly added command to recover the metadata and schemas, but not data.
    The idea is as follows:
    1. If the store already exists, the command will exit.
    2. If the store is not found in store graveyard, not recoverable.
    3. Then retrieve the store config from graveyard.
    4. Scan the corresponding admin channel to collect the latest key/value schemas.
    5. Create the store, register the schemas and then update the store configs.
    
    Sample command:
    java -jar venice-admin-tool-all.jar --recover-store-metadata --url ${parent.controller.url} --venice-zookeeper-url {parent zookeeper url} --zk-ssl-config-file ./zk.ssl --kafka-bootstrap-servers {Parent Kafka url} --kafka-consumer-config-file ./ssl.config --store {store_name} --repair true --graveyard-clusters {cluster1,cluster2}
    
    Sample zk.ssl:
    zookeeper.client.secure=true
    zookeeper.clientCnxnSocket=org.apache.zookeeper.ClientCnxnSocketNetty
    zookeeper.ssl.keyStore.location=./identity.p12
    zookeeper.ssl.keyStore.password=xxxx
    zookeeper.ssl.keyStore.type=PKCS12
    zookeeper.ssl.trustStore.location=./cacerts
    zookeeper.ssl.trustStore.password=xxxxx
    zookeeper.ssl.trustStore.type=JKS
    jute.maxbuffer=4194304
    
    Sample ssl.config:
    security.protocol=SSL
    ssl.protocol=TLS
    ssl.trustmanager.algorithm=SunX509
    ssl.keymanager.algorithm=SunX509
    ssl.keystore.type=pkcs12
    ssl.keystore.location=./identity.p12
    ssl.keystore.password=xxxx
    ssl.key.password=xxxx
    ssl.truststore.type=JKS
    ssl.truststore.location=./cacerts
    ssl.truststore.password=xxxx
    ssl.secure.random.implementation=SHA1PRNG
    
    We can use the existing command in AdminTool to run an empty push against the recovered store if necessary.

This PR also added another command to backup the store graveyard to some local folder and sample usage:
java -jar venice-admin-tool-all.jar --backup-store-metadata-from-graveyard --venice-zookeeper-url {parent.zookeeper.url} --zk-ssl-config-file ./zk.ssl --backup-folder ./backup.2/ --graveyard_clusters {cluster1,cluster2}

## How was this PR tested?
Local test

### DIFF
--- a/clients/venice-admin-tool/src/main/java/com/linkedin/venice/AdminTool.java
+++ b/clients/venice-admin-tool/src/main/java/com/linkedin/venice/AdminTool.java
@@ -159,6 +159,7 @@ import org.apache.commons.cli.OptionGroup;
 import org.apache.commons.cli.Options;
 import org.apache.commons.cli.ParseException;
 import org.apache.commons.lang.StringUtils;
+import org.apache.helix.zookeeper.impl.client.ZkClient;
 
 
 public class AdminTool {
@@ -544,6 +545,12 @@ public class AdminTool {
           break;
         case CONFIGURE_STORE_VIEW:
           configureStoreView(cmd);
+          break;
+        case BACKUP_STORE_METADATA_FROM_GRAVEYARD:
+          backupStoreMetadataFromGraveyard(cmd);
+          break;
+        case RECOVER_STORE_METADATA:
+          recoverStoreMetadata(cmd);
           break;
         default:
           StringJoiner availableCommands = new StringJoiner(", ");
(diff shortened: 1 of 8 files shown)


clients/venice-admin-tool/src/main/java/com/linkedin/venice/AdminTool.java
```diff
@@ -545,6 +546,12 @@ public static void main(String args[]) throws Exception {
         case CONFIGURE_STORE_VIEW:
           configureStoreView(cmd);
           break;
+        case BACKUP_STORE_METADATA_FROM_GRAVEYARD:
```
FelixGV, line 549: Might be good to add an example of how to trigger this new command as well in the commit message.

clients/venice-admin-tool/src/main/java/com/linkedin/venice/RecoverStoreMetadata.java
```diff
@@ -0,0 +1,333 @@
...
+      boolean skipLastStoreCreation,
+      boolean doRepair,
+      List<String> graveyardClusterList,
+      String recoverCluster) throws IOException {
+
+    /**
+     * Try to see whether the store exists or not.
+     */
+    ControllerClient controllerClient =
+        ControllerClientFactory.getControllerClient(graveyardClusterList.get(0), url, sslFactory);
```
FelixGV, line 61: I don't understand why this is a list? Is it in case the operator doesn't know which cluster the store was originally part of (in which case they could supply a list of all the clusters?)... here there seems to be an assumption that the store must be found in the graveyard of the first cluster, which seems to defeat the purpose of passing a list? Can you comment on why a list is useful and what inputs are valid?

clients/venice-admin-tool/src/main/java/com/linkedin/venice/RecoverStoreMetadata.java
```diff
@@ -0,0 +1,333 @@
...
+        });
+        // Update store config
+        UpdateStoreQueryParams updateParams = new UpdateStoreQueryParams();
+        updateParams.setPartitionCount(deletedStore.getPartitionCount());
+        if (deletedStore.getPartitionerConfig() != null) {
+          updateParams.setPartitionerClass(deletedStore.getPartitionerConfig().getPartitionerClass())
+              .setPartitionerParams(deletedStore.getPartitionerConfig().getPartitionerParams())
+              .setAmplificationFactor(deletedStore.getPartitionerConfig().getAmplificationFactor());
+        }
+        ;
```
FelixGV, line 227: Nitpick: let's delete that...

FelixGV (commented):
LGTM! I just have a minor question on the cluster list param...

FelixGV:
There are also some static analysis and unit test code coverage issues we should address...

clients/venice-admin-tool/src/main/java/com/linkedin/venice/AdminTool.java
gaojieliu: Sure

clients/venice-admin-tool/src/main/java/com/linkedin/venice/RecoverStoreMetadata.java
gaojieliu: This is used to create a dummy ControllerClient to call `discoverCluster`, essentially, any cluster will work here.
I can try to make it easier to understand.

The reason to introduce a list of clusters since we don't know which clusters the deleted store belongs to.

gaojieliu: I am going to use this static function instead to mitigate the confusion:
```
    D2ServiceDiscoveryResponse discoveryResponse = ControllerClient.discoverCluster(url, storeName, sslFactory, 3);
    if (discoveryResponse.getCluster() != null) {
      throw new VeniceException(
          "Store: " + storeName + " exists in cluster: " + discoveryResponse.getCluster() + ", no need to recover");
    }
```

gaojieliu:
yes, I need to fix the unit test coverage issue.

FelixGV (approved):
Thanks Gaojie!

PR #823, opened 2024-01-17
[test] Fixed NPEs in RocksDBStorageEngineTest

FelixGV:
The testPartitioning and testRemovingPartitionTwice tests suffered from contamination from previous tests in the same function.

## How was this PR tested?

UT.

### DIFF
--- a/clients/da-vinci-client/src/test/java/com/linkedin/davinci/store/rocksdb/RocksDBStorageEngineTest.java
+++ b/clients/da-vinci-client/src/test/java/com/linkedin/davinci/store/rocksdb/RocksDBStorageEngineTest.java
@@ -22,6 +22,7 @@ import java.util.Optional;
 import java.util.Set;
 import org.testng.Assert;
 import org.testng.annotations.AfterClass;
+import org.testng.annotations.AfterMethod;
 import org.testng.annotations.BeforeClass;
 import org.testng.annotations.Test;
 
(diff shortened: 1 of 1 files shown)


clients/da-vinci-client/src/test/java/com/linkedin/davinci/store/rocksdb/RocksDBStorageEngineTest.java
```diff
@@ -67,6 +69,24 @@ public void cleanUp() throws Exception {
...
+  @AfterMethod
+  public void testCounter() {
+    this.testCount++;
+  }
+
+  /**
+   * Some tests require a reset if other tests have run before them, as they are sensitive to contamination.
+   *
+   * Alternatively, we could make {@link #setUp()} have {@link org.testng.annotations.BeforeMethod} and
+   * {@link #cleanUp()} have {@link AfterMethod}, though that makes the class take longer than the current approach.
```
adamxchen, line 81: nit: can we do that in this PR? I prefer this way as it's cleaner and don't get people confused when they touch this file in the future. However, if it takes a lot more effort than expected, I suppose we are okay with this too.

adamxchen (approved):
Thanks Felix!

clients/da-vinci-client/src/test/java/com/linkedin/davinci/store/rocksdb/RocksDBStorageEngineTest.java
FelixGV: This is how I did at first, but it made the test take longer to run, so I iterated until the current solution... effort-wise it's basically a no-op.

PR #824, opened 2024-01-18
[controller] Turn off A/A and inc push when turning off hybrid store config

sixpluszero:
## [controller] Turn off A/A and inc push when turning off hybrid store config
This PR adjusts the existing behavior that throws exception when trying to turn off hybrid on an inc push enabled store. This was added 1.5 year ago as part of tactical fix to prevent store to lose RT right away especially for inc push store. Based on my code reading, we will not delete RT right away, if there is active hybrid version. If that is true, the change should be considered safe.

The new behavior is: When disabling hybrid store config: 
1. If user also try to enable A/A or inc push together, we should fail loudly. 
2. Otherwise, we will disable A/A and inc push store config.

## How was this PR tested?
Added new unit test to confirm the logic is behaving as expected.

### DIFF
--- a/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
@@ -2315,6 +2315,14 @@ public class VeniceParentHelixAdmin implements Admin {
           && veniceHelixAdmin.isHybrid(updatedHybridStoreConfig);
       // Check if the store is being converted to a batch store
       boolean storeBeingConvertedToBatch = currStore.isHybrid() && !veniceHelixAdmin.isHybrid(updatedHybridStoreConfig);
+      LOGGER.info(
+          "DEBUGGING: {} {} {} {} {} ",
+          storeBeingConvertedToBatch,
+          activeActiveReplicationEnabled,
+          incrementalPushEnabled,
+          currStore.isHybrid(),
+          veniceHelixAdmin.isHybrid(updatedHybridStoreConfig));
+
       if (storeBeingConvertedToBatch && activeActiveReplicationEnabled.orElse(false)) {
         throw new VeniceHttpException(
             HttpStatus.SC_BAD_REQUEST,
--- a/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
@@ -2315,14 +2315,6 @@ public class VeniceParentHelixAdmin implements Admin {
           && veniceHelixAdmin.isHybrid(updatedHybridStoreConfig);
       // Check if the store is being converted to a batch store
       boolean storeBeingConvertedToBatch = currStore.isHybrid() && !veniceHelixAdmin.isHybrid(updatedHybridStoreConfig);
-      LOGGER.info(
-          "DEBUGGING: {} {} {} {} {} ",
-          storeBeingConvertedToBatch,
-          activeActiveReplicationEnabled,
-          incrementalPushEnabled,
-          currStore.isHybrid(),
-          veniceHelixAdmin.isHybrid(updatedHybridStoreConfig));
-
       if (storeBeingConvertedToBatch && activeActiveReplicationEnabled.orElse(false)) {
         throw new VeniceHttpException(
             HttpStatus.SC_BAD_REQUEST,
--- a/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
@@ -2313,7 +2313,20 @@ public class VeniceParentHelixAdmin implements Admin {
       // Check if the store is being converted to a hybrid store
       boolean storeBeingConvertedToHybrid = !currStore.isHybrid() && updatedHybridStoreConfig != null
           && veniceHelixAdmin.isHybrid(updatedHybridStoreConfig);
-
+      // Check if the store is being converted to a batch store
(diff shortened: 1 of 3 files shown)


services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
```diff
@@ -2325,6 +2338,12 @@ public void updateStore(String clusterName, String storeName, UpdateStoreQueryPa
         setStore.activeActiveReplicationEnabled = true;
         updatedConfigsList.add(ACTIVE_ACTIVE_REPLICATION_ENABLED);
       }
+      // When turning of hybrid store, we will also turn off A/A store config.
```
adamxchen, line 2341: nit: off

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
```diff
@@ -2337,18 +2356,12 @@ public void updateStore(String clusterName, String storeName, UpdateStoreQueryPa
...
-      // incremental push without enabling hybrid already (we will automatically convert to hybrid store with default
-      // configs).
-      if (veniceHelixAdmin.isHybrid(currStore.getHybridStoreConfig())
-          && !veniceHelixAdmin.isHybrid(updatedHybridStoreConfig) && setStore.incrementalPushEnabled) {
-        throw new VeniceHttpException(
-            HttpStatus.SC_BAD_REQUEST,
-            "Cannot convert store to batch-only, incremental push enabled stores require valid hybrid configs. "
-                + "Please disable incremental push if you'd like to convert the store to batch-only",
-            ErrorType.BAD_REQUEST);
+      // When turning of hybrid store, we will also turn off incremental store config.
```
adamxchen, line 2359: dittp

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
adamxchen, line 1: I wonder if a unit test would be sufficient to cover it. Hopefully it's not too hard to write one😂

sixpluszero: Just added one. Original unit test class missed one mocking which took me extra hour to debug :)

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
sixpluszero: Thanks for pointing out, just fixed.

services/venice-controller/src/test/java/com/linkedin/venice/controller/TestVeniceParentHelixAdmin.java
```diff
@@ -1846,6 +1846,50 @@ public void testUpdateStoreNativeReplicationSourceFabric() {
...
+
+    doReturn(CompletableFuture.completedFuture(new SimplePubSubProduceResultImpl(topicName, partitionId, 1, -1)))
+        .when(veniceWriter)
+        .put(any(), any(), anyInt());
+
+    when(zkClient.readData(zkMetadataNodePath, null)).thenReturn(null)
+        .thenReturn(AdminTopicMetadataAccessor.generateMetadataMap(1, -1, 1));
+
+    parentAdmin.initStorageCluster(clusterName);
+    parentAdmin.updateStore(
```
adamxchen, line 1874: Thanks a lot! Can you also add a few lines to confirm exceptions would be thrown as expected?

adamxchen: I want to make sure it would fail loudly and would not be removed by someone accidentally

sixpluszero: What do you mean by this? before the logic change, the Assertion will be thrown at the end because it won't disable hybrid config as expected. With updated logic, new store is no longer hybrid at the end of the test.

adamxchen: Sure. I was asking if you can add a few more tests for unhappy path where the exceptions would be thrown and we verify that.

adamxchen: Do we have tests to exercise line 2319 2325?

sixpluszero: Oh I see what you mean. This part I can add it

sixpluszero: Updated, could you please check again?

services/venice-controller/src/test/java/com/linkedin/venice/controller/TestVeniceParentHelixAdmin.java
```diff
@@ -1846,6 +1846,68 @@ public void testUpdateStoreNativeReplicationSourceFabric() {
...
+        .put(any(), any(), anyInt());
+
+    when(zkClient.readData(zkMetadataNodePath, null)).thenReturn(null)
+        .thenReturn(AdminTopicMetadataAccessor.generateMetadataMap(1, -1, 1));
+
+    parentAdmin.initStorageCluster(clusterName);
+    // When user disable hybrid but also try to manually turn on A/A or Incremental Push, update operation should fail
+    // loudly.
+    Assert.assertThrows(
+        () -> parentAdmin.updateStore(
```
adamxchen, line 1878: nit: we can call out that's `VeniceHttpException` will be thrown but it's okay too

adamxchen (approved):
Thank you for the fix!

PR #825, opened 2024-01-18
[test] fix restartServerByDeletingSSTFiles tests with checksum enabled

m-nagarajan:
## Summary
This test originally had the following flow:
* Delete the SST files.
* Restart the servers to simulate a crash and restart in the midst of SST file move.

However, this flow introduced an issue when checksum was enabled:
Graceful shutdown of the server after SST file deletion triggers a final offset sync, which throws an exception when files are missing during checksum calculations.

Modified the test to:
Stop the server -> Delete the SST files -> Start the server.

## How was this PR tested?
GH CI

### DIFF
--- a/clients/da-vinci-client/src/test/java/com/linkedin/davinci/store/rocksdb/RocksDBSstFileWriterTest.java
+++ b/clients/da-vinci-client/src/test/java/com/linkedin/davinci/store/rocksdb/RocksDBSstFileWriterTest.java
@@ -1,16 +1,20 @@
 package com.linkedin.davinci.store.rocksdb;
 
 import com.linkedin.venice.exceptions.VeniceException;
+import com.linkedin.venice.kafka.validation.checksum.CheckSum;
+import com.linkedin.venice.kafka.validation.checksum.CheckSumType;
 import com.linkedin.venice.utils.Utils;
 import com.linkedin.venice.utils.VeniceProperties;
 import java.io.File;
 import java.io.IOException;
+import java.nio.ByteBuffer;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Optional;
 import org.apache.commons.io.FileUtils;
 import org.rocksdb.EnvOptions;
 import org.rocksdb.Options;
+import org.rocksdb.RocksDBException;
 import org.testng.Assert;
 import org.testng.annotations.Test;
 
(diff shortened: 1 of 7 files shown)


adamxchen (commented):
> Graceful shutdown of the server after SST file deletion triggers a final offset sync, which throws an exception when files are missing during checksum calculations.

Can you add one more test (either a unit test or integ test) to test this path?

sixpluszero:
> > Graceful shutdown of the server after SST file deletion triggers a final offset sync, which throws an exception when files are missing during checksum calculations.
> 
> Can you add one more test (either a unit test or integ test) to test this path?

I think the change itself is good. Regarding the suggestion by Adam, I think it is OK to have it to validate the final sync() action as you mentioned.

Nit: Is it possible to merge the non-aa test with the A/A test with some test parameters? Not necessarily need to happen in this PR if you think it is complicated but I suppose it can just be turned off by setting store to A/A = false?

m-nagarajan:
@adamxchen @sixpluszero added more unit tests, please take another look.

clients/da-vinci-client/src/test/java/com/linkedin/davinci/store/rocksdb/RocksDBSstFileWriterTest.java
```diff
@@ -255,6 +259,124 @@ public void testOpenWithAllValidFiles() throws IOException {
...
+      // create 10 sst file
+      createSstFiles(10);
+
+      rocksDBSstFileWriter.open(checkpointedInfo, Optional.of(() -> "1".getBytes()));
+      // check that only 7 files exists (0-5 after cleanup by checkDatabaseIntegrity)
+      // and 6 after rocksDBSstFileWriter.open opens a new file
+      Assert.assertEquals(getNumberOfFilesInTempDirectory(), 7);
+      rocksDBSstFileWriter.put("key".getBytes(), ByteBuffer.wrap("value".getBytes()));
+      deleteAllSstFiles(getNumberOfFilesInTempDirectory());
+      // call sync after deleting all the files to mimic exception thrown during graceful shutdown of servers with
```
sixpluszero, line 377: Other change looks good, but I am confused by the intention of this particular test. Could you explain a bit more about this one?

m-nagarajan: This was the RC behind this test failing after enabling checksum calculation. so added a test to verify that the checksum fails if there are no sst files during the sync() which is expected.

sixpluszero (approved):
Thanks for the fix!

PR #826, opened 2024-01-19
[Test] Add an integration test for hybrid store RT topic deletion

lluwm:
This change adds a new integration test to verify that store ingestion and hybrid store RT deletion can be handled correctly in the shared consumer thread.


internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestHybridStoreDeletion.java
```diff
@@ -0,0 +1,213 @@
+package com.linkedin.venice.endToEnd;
+
+import static com.linkedin.davinci.store.rocksdb.RocksDBServerConfig.ROCKSDB_PLAIN_TABLE_FORMAT_ENABLED;
+import static com.linkedin.venice.ConfigKeys.*;
```
FelixGV, line 4: Can we remove the wildcard please?

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestHybridStoreDeletion.java
```diff
@@ -0,0 +1,213 @@
...
+            ClientConfig.defaultGenericClientConfig(storeNameSecond)
+                .setVeniceURL(veniceCluster.getRandomRouterURL()))) {
+
+      createStoresAndVersions(controllerClient, storeNames, streamingRewindSeconds, streamingMessageLag);
+
+      // Produce to the rt topic of the first store and allow the thread to produce some amount of data.
+      produceToStoreRTTopic(storeNameFirst, 200);
+
+      // Delete the rt topic of the first store.
+      controllerClient.deleteKafkaTopic(Version.composeRealTimeTopic(storeNameFirst));
```
FelixGV, line 140: Did it work? Can we use `assertCommand` to make sure?

Also, I believe this ends up calling `admin.truncateKafkaTopic(topicName)`, which asynchronously kicks off the deletion process by the `TopicCleanupService`, rather than deleting immediately. Moreover, there may be safeguards in the `TCS` preventing the deletion of a RT topic from a non-deleted hybrid store (not sure).

We should add a non-deterministic assertion which waits until the topic is actually deleted. Or maybe just use the Kafka tooling to directly delete the topic, rather than go through the controller client, controller, `TCS`...

FelixGV (commented):
Thanks for this test. It seems that the new test passes, which is unexpected, right? I point out an issue in the comments, though IDK if there are other issues. Let's try to get the test to fail, so that we can repro the issue we've seen...

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestHybridStoreDeletion.java
lluwm: Sure, will fix it.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestHybridStoreDeletion.java
lluwm: > Did it work? Can we use assertCommand to make sure?

I think so, for example, I can find the progress of deleting the rt topic from test logs. Of course we can use `assrtCommand` as suggested.
```
2024-01-23 10:50:06 - [] INFO [TopicCleanupServiceForParentController] [TopicCleanupTask] Retention policy for topic: hybrid-store-test-first_264c8084685a2_a9986bb4_rt is: 5000 ms, and it is deprecated, will delete it after 200 ms.
...
2024-01-23 10:50:06 - [] INFO [TopicManager [localhost:57659]] [TopicCleanupTask] Deleting topic: hybrid-store-test-first_264c8084685a2_a9986bb4_rt
...
2024-01-23 10:50:06 - [] INFO [TopicManager [localhost:57659]] [TopicCleanupTask] Topic: hybrid-store-test-first_264c8084685a2_a9986bb4_rt has been deleted
```
After the topic had been deleted, I was able to see e.g. similar error messages in the shared consumer thread.  
```
2024-01-23 10:50:08 - [testRegion-VeniceServer-57714] WARN [NetworkClient] [venice-shared-consumer-for-localhost:57659-t1] [Consumer clientId=localhost_localhost:57659_0, groupId=null] Error while fetching metadata with from source 1001 with correlation id 114 : {hybrid-store-test-first_264c8084685a2_a9986bb4_rt=UNKNOWN_TOPIC_OR_PARTITION}
2024-01-23 10:50:08 - [testRegion-VeniceServer-57714] WARN [NetworkClient] [venice-shared-consumer-for-localhost:57659-t1] [Consumer clientId=localhost_localhost:57659_0, groupId=null] Error while fetching metadata with from source 1001 with correlation id 117 : {hybrid-store-test-first_264c8084685a2_a9986bb4_rt=UNKNOWN_TOPIC_OR_PARTITION}
```

FelixGV: So... it sounds like there is no issue with a topic getting deleted while a consumer is subscribed to it. That is fine I guess (though unexpected). I think we should still have a stronger assertion that the topic is actually deleted, e.g. via `PubSubAdminAdapter::containsTopic`, before moving on to the next step.

Once we have that, then we could merge the PR. WDYT?

lluwm: That's also my understanding from this test, our system seems to be able to handle it correctly. However, there might be some other factors that in combination with a deleted RT topic to cause the dead consumer thread issue (thus not able to ingest data), and we are not quite clear on what the other factor is.

I agree and I will add the assertion in the test to make sure the topic is deleted before the next step.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestHybridStoreDeletion.java
```diff
@@ -1,7 +1,17 @@
...
+import static com.linkedin.venice.ConfigKeys.MIN_CONSUMER_IN_CONSUMER_POOL_PER_KAFKA_CLUSTER;
+import static com.linkedin.venice.ConfigKeys.PERSISTENCE_TYPE;
+import static com.linkedin.venice.ConfigKeys.SERVER_CONSUMER_POOL_SIZE_PER_KAFKA_CLUSTER;
+import static com.linkedin.venice.ConfigKeys.SERVER_DATABASE_CHECKSUM_VERIFICATION_ENABLED;
+import static com.linkedin.venice.ConfigKeys.SERVER_DATABASE_SYNC_BYTES_INTERNAL_FOR_DEFERRED_WRITE_MODE;
+import static com.linkedin.venice.ConfigKeys.SERVER_DEDICATED_DRAINER_FOR_SORTED_INPUT_ENABLED;
+import static com.linkedin.venice.ConfigKeys.SERVER_PROMOTION_TO_LEADER_REPLICA_DELAY_SECONDS;
+import static com.linkedin.venice.ConfigKeys.SERVER_SHARED_CONSUMER_ASSIGNMENT_STRATEGY;
+import static com.linkedin.venice.ConfigKeys.SSL_TO_KAFKA_LEGACY;
+import static com.linkedin.venice.kafka.TopicManager.*;
```
FelixGV, line 14: Sorry... one more 🥲 ...

FelixGV (commented):
One more minor comment...

Also, I seem to recall that the issue with deleted topics might have been if the producer got stuck writing to a disappeared topic... if that is the case, then it may be benign when the RT topic disappears (as your test seems to indicate...). It may be more of a problem if the VT disappears, and furthermore if more writes keep coming in through the RT, until the VT producer's buffer is filled up and starts imposing back pressure, at which point the `ConsumptionTask` could become stuck.

Maybe @gaojieliu has other ideas regarding this. As it stands now, it seems the new test cannot repro any issue, which seems concerning...

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestHybridStoreDeletion.java
lluwm: Ahh... I am sorry about my inattention here.

FelixGV (approved):
Thanks Lei. I think we can merge this in as is. It is good to have a test for this scenario.

That being sad, we should also continue working in this area to keep digging into whether deleted topics can cause hangs. For example, deleting a version-topic, such that the producer's buffer gets full. This can be a separate PR.

lluwm:
Sound good @FelixGV. Proceed to merge this change.

PR #828, opened 2024-01-23
[server][controller] Try ungraceful close when the graceful one faile…

lluwm:
…d in VeniceWriter

In VeniceWriter.close(), when gracefulClose is set to true, Kafka producer call the underlining method, which tries to flush the buffered data before closing it (see ApacheKafkaProducerAdapter.close). However, when exception happens (e.g. timeout), today, VeniceWriter only logs the exception, moves on, and consider the VeniceWriter being closed while it is actually not. (A leaking Venicewriter can cause several issues, e.g. a stuck consumer thread etc.)

This change adds a retry of ungraceful close if gracefulClose flag is set to true and it failed. In this case, ungraceful close will skip the flushing step and it should always succeed (e.g. StoreIngestionTask.kill()). For cases where gracefulClose is set to false, this change doesn't change anything.

## How was this PR tested?

Unit test.
CI.

### DIFF
--- a/internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
@@ -396,6 +396,10 @@ public class VeniceWriter<K, V, U> extends AbstractVeniceWriter<K, V, U> {
    */
   @Override
   public void close(boolean gracefulClose) {
+    close(gracefulClose, true);
+  }
+
+  public void close(boolean gracefulClose, boolean retryOnGracefulCloseFailure) {
     synchronized (closeLock) {
       if (isClosed) {
         return;
@@ -416,7 +420,7 @@ public class VeniceWriter<K, V, U> extends AbstractVeniceWriter<K, V, U> {
           producerAdapter.close(topicName, closeTimeOutInMs, gracefulClose);
           OPEN_VENICE_WRITER_COUNT.decrementAndGet();
         } catch (Exception e) {
-          handleExceptionInClose(e, gracefulClose);
+          handleExceptionInClose(e, gracefulClose, retryOnGracefulCloseFailure);
         } finally {
           threadPoolExecutor.shutdown();
           isClosed = true;
@@ -426,6 +430,12 @@ public class VeniceWriter<K, V, U> extends AbstractVeniceWriter<K, V, U> {
   }
 
   public CompletableFuture<VeniceResourceCloseResult> closeAsync(boolean gracefulClose) {
+    return closeAsync(gracefulClose, true);
+  }
+
+  public CompletableFuture<VeniceResourceCloseResult> closeAsync(
+      boolean gracefulClose,
+      boolean retryOnGracefulCloseFailure) {
     return CompletableFuture.supplyAsync(() -> {
       synchronized (closeLock) {
         if (isClosed) {
@@ -452,7 +462,7 @@ public class VeniceWriter<K, V, U> extends AbstractVeniceWriter<K, V, U> {
             producerAdapter.close(topicName, closeTimeOutInMs, gracefulClose);
             OPEN_VENICE_WRITER_COUNT.decrementAndGet();
           } catch (Exception e) {
-            handleExceptionInClose(e, gracefulClose);
+            handleExceptionInClose(e, gracefulClose, retryOnGracefulCloseFailure);
           } finally {
             threadPoolExecutor.shutdown();
             isClosed = true;
@@ -464,13 +474,13 @@ public class VeniceWriter<K, V, U> extends AbstractVeniceWriter<K, V, U> {
     }, threadPoolExecutor);
   }
 
-  void handleExceptionInClose(Exception e, boolean gracefulClose) {
+  void handleExceptionInClose(Exception e, boolean gracefulClose, boolean retryOnGracefulCloseFailure) {
     logger.warn("Swallowed an exception while trying to close the VeniceWriter for topic: {}", topicName, e);
     VENICE_WRITER_CLOSE_FAILED_COUNT.incrementAndGet();
 
     // For graceful close, swallow the exception and give another try to close it ungracefully.
     try {
-      if (gracefulClose) {
+      if (gracefulClose && retryOnGracefulCloseFailure) {
         producerAdapter.close(topicName, closeTimeOutInMs, false);
         OPEN_VENICE_WRITER_COUNT.decrementAndGet();
       }
--- a/internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
@@ -396,9 +396,25 @@ public class VeniceWriter<K, V, U> extends AbstractVeniceWriter<K, V, U> {
    */
   @Override
   public void close(boolean gracefulClose) {
-    close(gracefulClose, true);
+    try {
+      closeAsync(gracefulClose).get();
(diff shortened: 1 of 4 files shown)


internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
```diff
@@ -399,26 +400,28 @@ public void close(boolean gracefulClose) {
...
+          if (gracefulClose) {
+            endAllSegments(true);
+          }
+          // DO NOT call the {@link #PubSubProducerAdapter.close(int) version from here.}
+          // For non-shared producer mode gracefulClose will flush the producer
+
+          producerAdapter.close(topicName, closeTimeOutInMs, gracefulClose);
+          OPEN_VENICE_WRITER_COUNT.decrementAndGet();
+        } catch (Exception e) {
+          handleExceptionInClose(e, gracefulClose);
```
gaojieliu, line 419: Originally, I was thinking whether we should implement this logic in the app layer of `VeniceWriter` since the app can decide whether it would choose to close it forcibly when some exception happens.
Maybe we should expose a param in case some usage of `VeniceWriter#close` would like to graceful close all the time and fail fast if it doesn't work.

lluwm: I thought about this idea initially and then realized that:
1. `close` has multiple callers (5-6), we could have some duplicated codes, if we handle it at its caller side.
2. More important, today's `close` function swallows all exceptions and doesn't pass them to its caller, so that it leaves the impression to its caller that `close` should always succeed (but actually it is not true). 

> Originally, I was thinking whether we should implement this logic in the app layer of VeniceWriter since the app can decide whether it would choose to close it forcibly when some exception happens.

Yeah, we can add another boolean in the signature and based on its value to determine if a forceful close is required. 
> Maybe we should expose a param in case some usage of VeniceWriter#close would like to graceful close all the time and fail fast if it doesn't work.

gaojieliu: What I meant is still keeping this logic in `VeniceWriter` to avoid duplication, but exposing a boolean param to let the user have an option to fallback or not.

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
```diff
@@ -395,70 +396,101 @@ public static ControlMessage generateHeartbeatMessage(CheckSumType checkSumType)
...
-        // For non-shared producer mode gracefulClose will flush the producer
 
-        producerAdapter.close(topicName, closeTimeOutInMs, gracefulClose);
-        OPEN_VENICE_WRITER_COUNT.decrementAndGet();
-      } catch (Exception e) {
-        logger.warn("Swallowed an exception while trying to close the VeniceWriter for topic: {}", topicName, e);
-        VENICE_WRITER_CLOSE_FAILED_COUNT.incrementAndGet();
+      logger.info("Closing VeniceWriter for topic: {}, gracefulness: {}", topicName, gracefulClose);
+      try (Timer ignore = Timer.run(
+          elapsedTimeInMs -> logger.info("Closed VeniceWriter for topic: {} in {} ms", topicName, elapsedTimeInMs))) {
```
gaojieliu, line 410: Shall we mention this is `gracefully closed VeniceWriter` to make it clear since there is a logic for ungraceful shutdown?

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
lluwm: I think I did that in the new update. Would you take another look and see if it addressed your comments?

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
```diff
@@ -395,70 +396,101 @@ public static ControlMessage generateHeartbeatMessage(CheckSumType checkSumType)
...
+        try {
+          // If {@link #broadcastEndOfPush(Map)} was already called, the {@link #endAllSegments(boolean)}
+          // will not do anything (it's idempotent). Segments should not be ended if there are still data missing.
+          if (gracefulClose) {
+            endAllSegments(true);
```
gaojieliu, line 415: https://github.com/linkedin/venice/blob/main/internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java#L1693
Sending control messages to a non-existing topic can stuck forever since this function is calling `get()` without timeout.

If we don't address this logic, the fallback can only execute in the application logic to have a timer and ungraceful close the writer.

gaojieliu: Another strategy to this problem:
can we always close the VeniceWriter ungracefully?

gaojieliu: There is one more way to think about to address this stuck consumer issue:
1. Check the thread trace of the stuck consumer to see whether it is blocked by Producer#send.
2. If yes, check the producing topic exist or not.
3. If not, kill the corresponding VeniceWriter.

This will be a little hacky, but it would catch all the cases of stuck producers.

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
```diff
@@ -395,70 +396,101 @@ public static ControlMessage generateHeartbeatMessage(CheckSumType checkSumType)
...
+          }
+          // DO NOT call the {@link #PubSubProducerAdapter.close(int) version from here.}
+          // For non-shared producer mode gracefulClose will flush the producer
+
+          producerAdapter.close(topicName, closeTimeOutInMs, gracefulClose);
```
gaojieliu, line 420: Maybe we can pass `closeTimeOutInMS` to `endAllSegments` to avoid waiting forever when the producing topic doesn't exist.

internal/venice-common/src/test/java/com/linkedin/venice/writer/VeniceWriterUnitTest.java
```diff
@@ -533,4 +538,26 @@ public void testSendHeartbeat(boolean addLeaderCompleteHeader, LeaderCompleteSta
       }
     }
   }
+
+  // Write a unit test for the retry mechanism in VeniceWriter.close(true) method.
+  @Test(dataProvider = "True-and-False", dataProviderClass = DataProviderUtils.class, timeOut = 10 * Time.MS_PER_SECOND)
+  public void testVeniceWriterCloseRetry(boolean gracefulClose) throws ExecutionException, InterruptedException {
+    PubSubProducerAdapter mockedProducer = mock(PubSubProducerAdapter.class);
```
gaojieliu, line 545: We potentially should have some E2E integration test for this to test whether this will work for different scenarios or not.

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
lluwm: As discussed offline, we can leverage the closeAsync with wait to avoid the indefinite blocking case as you mentioned.

internal/venice-common/src/test/java/com/linkedin/venice/writer/VeniceWriterUnitTest.java
lluwm: Sure, a new integration test is added.

gaojieliu (approved):
Thanks!

PR #829, opened 2024-01-23
[FC][test] Update HTTP configs and libraries, add FC warmup config, and fix metric unit

m-nagarajan:
## Summary
* updated `validateAfterInactivity` and `connectionKeepAlive` configs to be indefinite 
* bump http libs to the latest versions: `httpAsyncClient`, `httpClient5`,  `httpCore5,` `httpCore5H2`, `httpClient`, `httpCore`
* added a config for FC warmup defaulted to true
* `requestSerializationTime` was reporting in NS, changed it to MS to keep it uniform with other metrics

## How was this PR tested?
GH CI


internal/venice-client-common/src/main/java/com/linkedin/venice/httpclient5/HttpClient5Utils.java
```diff
@@ -40,9 +41,10 @@ public static class HttpClient5Builder {
      * We need to use a high connect timeout to avoid reconnect issue, which might result in confusing logging and unhealthy requests.
      * For now, we remove the functions updating to connect timeout to avoid mistakes.
      */
-    private final Timeout CONNECT_TIMEOUT_IN_MILLISECONDS = Timeout.ofMilliseconds(TimeUnit.MINUTES.toMillis(1)); // 1m
-                                                                                                                  // by
-                                                                                                                  // default
+    private final Timeout CONNECT_TIMEOUT_IN_MINUTES = Timeout.ofMinutes(1);
```
m-nagarajan, line 44: @gaojieliu starting a comment here about java Doc for this class. How relevant is this currently and whether can this be removed?

internal/venice-client-common/src/main/java/com/linkedin/venice/httpclient5/HttpClient5Utils.java
```diff
@@ -85,6 +87,25 @@ public HttpClient5Builder setHttp1MaxConnectionsPerRoute(int http1MaxConnections
       return this;
     }
 
+    private ConnectionConfig getDefaultConnectionConfig() {
+      return ConnectionConfig.custom()
+          .setConnectTimeout(CONNECT_TIMEOUT_IN_MINUTES)
+          .setSocketTimeout(CONNECT_TIMEOUT_IN_MINUTES)
+          // Override default value of 2seconds to CONNECTION_INDEFINITE_TIMEOUT
+          .setValidateAfterInactivity(CONNECTION_INDEFINITE_TIMEOUT)
+          .setTimeToLive(CONNECTION_INDEFINITE_TIMEOUT)
```
xunyin8, line 96: Will there be any side effects to set TTL of connections to -1? In the javadoc it doesn't say if this means it will not enforce TTL or not, only that the default is null.

xunyin8: Checked offline and this is good.

xunyin8 (approved):
LGTM thanks!

PR #830, opened 2024-01-23
[common][controller] Delete all version topics before deleting RT topic

xunyin8:
## [common][controller] Delete all version topics before deleting RT topic

1. TopicCleanupService in child datacenters will now only delete RT topic if all version topics of the corresponding store is deleted in all child fabrics. This is to ensure there are no ongoing ingestions still polling for these RT topics before deleting them.

2. Cleaned up some excessive logging in store graveyard and FC related meta system store updates in cleanup service that's no longer applicable.

## How was this PR tested?
Existing tests and new unit tests.

### DIFF
--- a/services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/TopicCleanupService.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/TopicCleanupService.java
@@ -189,7 +189,7 @@ public class TopicCleanupService extends AbstractVeniceService {
             if (storeVersionTopicCount.containsKey(topic.getStoreName())) {
               canDelete = false;
               LOGGER.info(
-                  "Topic deletion for topic: {} is delayed due to {} version topics found in a remote fabric",
+                  "Topic deletion for topic: {} is delayed due to {} version topics found in a local/remote fabric",
                   topic.getName(),
                   storeVersionTopicCount.get(topic.getStoreName()));
               break;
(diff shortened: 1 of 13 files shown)


xunyin8:
Added an multi datacenter integration test and found some issues, addressing them now. Please hold on the review.

gaojieliu (commented):
TopicCleanupService is more and more complex, and I think we should start introduce some metrics to track the deletion progress and error scenarios.

xunyin8:
> TopicCleanupService is more and more complex, and I think we should start introduce some metrics to track the deletion progress and error scenarios.

Do you think two metrics, one for deletable topics and topics deleted is sufficient or we should split them into VT and RT topics for each category?

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/TopicCleanupService.java
```diff
@@ -200,28 +221,70 @@ void cleanupVeniceTopics() {
...
+          String pubSubBootstrapServer = multiClusterConfigs.getChildDataCenterKafkaUrlMap().get(childFabric);
+          if (localPubSubBootstrapServer.equals(pubSubBootstrapServer)) {
+            continue;
+          }
+          Set<PubSubTopic> remoteTopics = getTopicManager(pubSubBootstrapServer).getAllTopicRetentions().keySet();
+          clearAndPopulateStoreToVersionTopicCountMap(remoteTopics, multiDataCenterStoreToVersionTopicCount.get(i));
+          i++;
+        } catch (Exception e) {
+          LOGGER.error("Failed to refresh store to version topic count map for fabric {}", childFabric, e);
+        }
```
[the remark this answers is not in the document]
xunyin8: Will block RT topic deletion if we:

1. Unable to make sense of the child fabric list or kafka URLs (mis-configured)
2. Unable to fetch version topic info because kafka is down/unreachable like you mentioned in the above scenario.

gaojieliu:
> TopicCleanupService is more and more complex, and I think we should start introduce some metrics to track the deletion progress and error scenarios.
> 
> Do you think two metrics, one for deletable topics and topics deleted is sufficient or we should split them into VT and RT topics for each category?

Yes, I think it is good enough, and also capture the error scenarios, such as RT topic deletion is blocked.

gaojieliu:
All other code changes look good.

xunyin8:
> All other code changes look good.

Thanks for the review. I just updated with the new metrics.

services/venice-controller/src/main/java/com/linkedin/venice/controller/stats/TopicCleanupServiceStats.java
```diff
@@ -0,0 +1,33 @@
...
+public class TopicCleanupServiceStats extends AbstractVeniceStats {
+  private final Sensor deletableTopicsCountGaugeSensor;
+  private final Sensor deletedTopicsCountSensor;
+  private final Sensor topicDeletionErrorCountSensor;
+
+  public TopicCleanupServiceStats(MetricsRepository metricsRepository) {
+    super(metricsRepository, "TopicCleanupService");
+    deletableTopicsCountGaugeSensor = registerSensorIfAbsent("deletable_topics_count", new Gauge());
+    deletedTopicsCountSensor = registerSensorIfAbsent("deleted_topics_count", new Count());
+    topicDeletionErrorCountSensor = registerSensorIfAbsent("topic_deletion_error_count", new Count());
```
gaojieliu, line 19: Shall we use `Rate` instead of `Count`?

xunyin8: Good point, we don’t really care the exact count as much as the overall rate. Let me update it.

gaojieliu (approved):
Thanks!

PR #831, opened 2024-01-25
[admin-tool] Add support to dump UPDATE records and RMD records in Kafka Topic Dumper

sixpluszero:
## [admin-tool] Add support to dump UPDATE records and RMD records in Kafka Topic Dumper
This PR adds new mode to print data record (PUT/DELETE/UPDATE) as well as its corresponding RMD record to console, on top of its existing functionality of logging Kafka message metadata to console. It is very useful for debugging feature correctness. 

In the future, we can consider adding a key filter to log specific key only, to better focus on certain key.

## How was this PR tested?
Added new unit test to cover the message build up

### DIFF
--- a/clients/venice-admin-tool/src/main/java/com/linkedin/venice/AdminTool.java
+++ b/clients/venice-admin-tool/src/main/java/com/linkedin/venice/AdminTool.java
@@ -1616,7 +1616,9 @@ public class AdminTool {
       maxConsumeAttempts = Integer.parseInt(getOptionalArgument(cmd, Arg.MAX_POLL_ATTEMPTS));
     }
 
-    boolean logMetadataOnly = cmd.hasOption(Arg.LOG_METADATA.toString());
+    boolean logMetadata = cmd.hasOption(Arg.LOG_METADATA.toString());
+    boolean logDataRecord = cmd.hasOption(Arg.LOG_DATA_RECORD.toString());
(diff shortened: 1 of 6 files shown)


ZacAttack (dismissed):
Looks great!

ZacAttack:
One thing occurs to me though.  This will dump the content of fields.  That is useful I suppose, but I wonder if we should by default add some screening to the output to avoid data privacy violations.

Like, it's probably ok to display nulls or NoOps, but anything else we should probably screen with ######

sixpluszero:
> One thing occurs to me though.  This will dump the content of fields.  That is useful I suppose, but I wonder if we should by default add some screening to the output to avoid data privacy violations.
> 
> Like, it's probably ok to display nulls or NoOps, but anything else we should probably screen with ###### with an optional flag to display the data if it's truly necessary.

Hmm, my understanding is first the existing mode already supports dumping the whole data into files, what I added here is only to extend the functionality to be easier to use. So I mean I don't make it more or less safe compared to master branch.
Also, when we are using the dumper, we are already guarded by ACL. The optional flag you suggest, I think it does not add extra safety unless the option is also guarded with permission. It really depends on the purpose of the user. We could think of ways to make it more secure, but I think..it could happen in a future PR I suppose? To me for now it looks more like query.sh script...

ZacAttack (approved):

PR #833, opened 2024-01-26
[server] Skip success request metric tracking for /health request

gaojieliu:
/health request from Router doesn't have a store name, so the store-level tracking would break with a gaint log stacktrace, which is bloating log file, and this is a regression happened a few months ago.

## How was this PR tested?
CI


FelixGV (approved):
Thanks Gaojie!

gaojieliu:
I debugged further and this PR is not necessary.
The root cause is as follow with the version running in prod:
1. Fast-Client is using `/health` endpoint for conn warming.
2. Prod version doesn't have this code change: https://github.com/linkedin/venice/issues/798, which will bypass acl check.
3. Prod version will fail all the `/health` requests from fast-client.
4. When Venice Server fails the health request in acl handler, it won't setup `serverStatsContext.isHealthCheck()`.
5. Without this flag, `StatsHandler` will go ahead to call `serverStatsContext.successRequest`, but it will fail since there is no store name.

Once we deploy #798 , we shouldn't see such issue in prod any more.

PR #834, opened 2024-01-27
[server][davinci] Implement Heartbeat Monitoring Service

ZacAttack:
## [server][davinci] Implement Heartbeat Monitoring Service
This PR implements the heartbeat monitoring service.

The service functions on two ends.  Getting notified for partitions
that should be monitored (as leader or follower), and notified when
heartbeats are consumed (in the context of leader or follower consumption).

Stats are then reported and updated from a monitoring thread that is seperate
from ingestion tasks.  The purpose of this is to enable monitoring to be a true
catch all.  For a hybrid store that is assigned Leader/Follower in helix, theres
no acceptable situation where the server should not be consuming records for
a prolonged duration.  Ingestion tasks may die, get paused, or recycled.  That
is fine, but if Helix says we should be up, then we'll report based on that.

Resolves #XXX

## How was this PR tested?
New tests


clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/ingestion/heartbeat/HeartbeatMonitoringService.java
```diff
@@ -0,0 +1,198 @@
...
+  }
+
+  private void recordHeartbeat(
+      String store,
+      int version,
+      int partition,
+      String region,
+      Long timestamp,
+      Map<String, Map<Integer, Map<Integer, Map<String, Long>>>> heartbeatTimestamps) {
+    if (heartbeatTimestamps.get(store) != null) {
```
xunyin8, line 132: Should we emit any metric/logs if entries are missing? i.e. `removeLagMonitor` is called on the corresponding store partition but somehow the ingestion is still triggering `recordHeartbeat` (maybe when draining the drainer queue?).

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -2267,6 +2301,7 @@ protected DelegateConsumerRecordResult delegateConsumerRecord(
             } else {
               if (controlMessageType == START_OF_SEGMENT
                   && Arrays.equals(consumerRecord.getKey().getKey(), KafkaKey.HEART_BEAT.getKey())) {
+                recordHeartbeatReceived(partitionConsumptionState, consumerRecord, kafkaUrl);
```
xunyin8, line 2304: If we plan to use the monitored lag for data freshness then we should report this in the drainer (when data is getting persisted and PCS offset is getting updated) instead of when we receive it from the consumer right?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -2066,6 +2070,36 @@ private void propagateHeartbeatFromUpstreamTopicToLocalVersionTopic(
...
+  protected void recordHeartbeatReceived(
+      PartitionConsumptionState partitionConsumptionState,
+      PubSubMessage<KafkaKey, KafkaMessageEnvelope, Long> consumerRecord,
+      String kafkaUrl) {
+    if (heartbeatMonitoringService == null) {
+      // Not enabled!
+      return;
+    }
+    long producerTimeStamp =
+        max(consumerRecord.getPubSubMessageTime(), consumerRecord.getValue().producerMetadata.messageTimestamp);
```
xunyin8, line 2082: Related to the comment to potentially move `recordHeartbeatReceived` to when we receive the HB in drainer. In that case maybe the `consumerRecord.getValue().producerMetadata.messageTimestamp` would get messed up since it will be the timestamp when the record was copied into VT from RT?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/ingestion/heartbeat/HeartbeatMonitoringService.java
```diff
@@ -0,0 +1,198 @@
...
+      heartbeatTimestamps.get(version.getStoreName()).put(version.getNumber(), new VeniceConcurrentHashMap<>());
+    }
+    if (heartbeatTimestamps.get(version.getStoreName()).get(version.getNumber()).get(partition) == null) {
+      Map<String, Long> regionTimestamps = new VeniceConcurrentHashMap<>();
+      if (version.isActiveActiveReplicationEnabled()) {
+        for (String region: regionNames) {
+          regionTimestamps.put(region, DEFAULT_SENTINEL_HEARTBEAT_TIMESTAMP);
+        }
+      } else {
+        regionTimestamps.put(localRegionName, DEFAULT_SENTINEL_HEARTBEAT_TIMESTAMP);
```
xunyin8, line 71: Essentially this means if no heartbeat is received between `initializeEntry` and `record` (up to 60s with current configs) the lag metric will get a spike of 6h. This could generate a lot of spikes/noise when ingestion is just started or rebalanced. We could set up monitoring accordingly to fire only when observing a lag that exceeds threshold for more than let's say 5min. However, I'm wondering if we could just initialize it with `-1` and have the monitoring emit -1 to indicate no heartbeat have been received yet. Since we don't really about the lag without any heartbeat. `DEFAULT_SENTINEL_HEARTBEAT_TIMESTAMP` is just an arbitrary value and could lead to confusion/miscommunication during an incident where oncall will see a lag of 6h+delta but in reality the exact lag could be anything and needs to be manually verified.

xunyin8 (commented):
I like the simplicity of it which means it should also be pretty robust. Left some comments and concerns.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
ZacAttack: I suppose we could, are heartbeats queued to drainer today?

I'm not sure if that would make too much difference.  If drainer is blocked, then consumption will be blocked, so we'll still get the lag/alert.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
ZacAttack: I guess it would depend on what we queue to drainer.  We could record it into what we queue there maybe.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/ingestion/heartbeat/HeartbeatMonitoringService.java
ZacAttack: So the problem with -1 is that our alerting is threshhold and duration based.  I'm not sure if we have a good way to set up "this number should be between 0 and 20k" in a single alert.

Duration is what we would leverage.  We would spike up to 6 hours and increase until we get the next heartbeat.  Something like "lag should not be above 1 hour for 3 hours" or however long we think it should take for ingestion to consume even one event or catch up.

ZacAttack: Oh!  Also it's a max.  If we use -1, then it'll get screened out by healthy ingesting partitions.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/ingestion/heartbeat/HeartbeatMonitoringService.java
ZacAttack: That's kind of interesting 🤔 

So like... as a means to detect leaked ingestion?  I think that's not a bad idea, but maybe let's do that in a separate PR and focus on this problem we're trying to detect.

xunyin8: Sounds good.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
xunyin8: Yeah, it depends if you want to use this lag metric as a lower bound or not.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/ingestion/heartbeat/HeartbeatMonitoringService.java
xunyin8: I see... was just trying to see if there's a way to avoid these spikes if we have sufficient signals to confirm they are false positive.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
adamxchen: I'm not following here.  Isn't `producerMetadata.messageTimestamp` always older/smaller than `getPubSubMessageTime()`? What's the circumstances that a `producerMetadata.messageTimestamp` being unreliable?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
ZacAttack: Put into the drainer

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
ZacAttack: I actually followed this pattern from another bit of code which was doing this (I assumed the original author thought of some edge case).  The comment above it said something about clock skew being a potential concern, though I don't think using Max would completely alleviate that notion besides erring on the side of caution.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/ingestion/heartbeat/HeartbeatMonitoringService.java
ZacAttack: After musing on it, I think theres probably three approaches we could try.  They are:

- emit -1, but augment Max to give special accordance to the value of -1.  This would mean that it doesn't matter what other big value is out there, if -1 is in there we report it.  A conditional MAX I guess
- Emit a seperate metric for those partitions where we haven't gotten a heartbeat for a partition that we expect it for.
- Commit heartbeats to disk and emit based on the disk value, thus keeping the lag between restarts.  Defaulting to a large number when theres nothing on disk.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/ingestion/heartbeat/HeartbeatMonitoringService.java
```diff
@@ -0,0 +1,198 @@
...
+import io.tehuti.metrics.MetricConfig;
+import io.tehuti.metrics.MetricsRepository;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.TimeUnit;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+
+public class HeartbeatMonitoringService extends AbstractVeniceService {
```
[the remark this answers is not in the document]
ZacAttack: done.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
m-nagarajan: Added `max()` to alleviate the time skew occurring in the leader topic partition, which would causing followers to consider the heartbeat as stale (since they check for one minute freshness) and never become ready to serve. This is not a complete fix, but the chance of both the leader topic partition and Kafka broker's time going awry at the same time is rare or if it happens, there will likely be other major issues around it.
For monitoring purposes, having either the producer timestamp, broker timestamp, or the maximum of both doesn't really make a difference in my opinion, unless we need to monitor this delay by calculating it from two separate metrics. Therefore, this change looks good IMO. 

Nit: For clarity and to keep the reasoning together, we can convert this line into a small function and move the comments about the skew there.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/ingestion/heartbeat/HeartbeatMonitoringService.java
xunyin8: I guess for V1 we can leave it as it is. If we find it inconvenient or noisy can always improve it later.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/ingestion/heartbeat/HeartbeatMonitoringService.java
```diff
@@ -0,0 +1,255 @@
...
+      ReportLagFunction lagFunction) {
+    for (Map.Entry<String, Map<Integer, Map<Integer, Map<String, Long>>>> storeName: heartbeatTimestamps.entrySet()) {
+      for (Map.Entry<Integer, Map<Integer, Map<String, Long>>> version: storeName.getValue().entrySet()) {
+        for (Map.Entry<Integer, Map<String, Long>> partition: version.getValue().entrySet()) {
+          for (Map.Entry<String, Long> region: partition.getValue().entrySet()) {
+            lagFunction.apply(storeName.getKey(), version.getKey(), region.getKey(), region.getValue());
+          }
+        }
+      }
+    }
```
FelixGV, line 219: This one I'm not sure about... is the iteration safe in the presence of concurrent updates to these maps? It may be fine I suppose.

FelixGV (commented):
Left a few belated comments on this. Mostly stylistic, though it may possibly make the code safer.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/ingestion/heartbeat/HeartbeatMonitoringService.java
ZacAttack: This should be thread safe, I think the only situation where this might go sidewise is that an entry we retrieved from the iterator is no longer there, and then when we call recordLag() we just hit one of the null checks and then carry on.  Other scenario is something gets added that we don't loop through on this round of iteration, which is ok, because we'll just get it next time.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/ingestion/heartbeat/HeartbeatMonitoringService.java
```diff
@@ -0,0 +1,255 @@
...
+  // This default value SHOULD be a number that will trip an alert. Alerting on this should be duration based,
+  // so starting off with a large value is fine that is cleared once we get the next heartbeat
+  public long DEFAULT_SENTINEL_HEARTBEAT_TIMESTAMP = System.currentTimeMillis() - 21600000; // 6 hours ago
+
+  private final Set<String> regionNames;
+  private final String localRegionName;
+
+  // store -> version -> partition -> region -> timestamp
+  private final Map<String, Map<Integer, Map<Integer, Map<String, Long>>>> followerHeartbeatTimeStamps;
+  private final Map<String, Map<Integer, Map<Integer, Map<String, Long>>>> leaderHeartbeatTimeStamps;
```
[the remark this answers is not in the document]
ZacAttack: These are good suggestions.  I've added them to this PR: https://github.com/linkedin/venice/pull/880

PR #835, opened 2024-01-29
[docs] Reorganized docs and added some

FelixGV:
Moved all contribution guides to dev_guide/how_to/ and left the dev_guide section focused on internal implementation details of Venice.

Moved most posts and talks to a new Learn More section, under user_guides.

Added a high-level architecture diagram on the main page. Added links between pages.

## How was this PR tested?
Here: https://felixgv.github.io/venice/


docs/README.md
```diff
@@ -77,74 +83,69 @@ Write Compute includes two kinds of operations, which can be performed on the va
 N.B.: Currently, write compute is only supported in conjunction with active-passive replication. Support for 
 active-active replication is under development.
```
sixpluszero, line 84: Do you want to update this part? Although we found a bug last Friday, I think we, still onboarded a few PROD cases already. (Or if you feel like it can be updated after GA that's also ok)

sixpluszero (approved):
Looks good overall, except just one comment.

docs/README.md
FelixGV: The bugs we find are further and further apart, and that is a good sign of progress, but I would prefer removing this disclaimer after all known issues are fixed.

PR #836, opened 2024-01-29
[controller] Delete unused schema for batch stores

majisourav99:
## Delete unused schema

Many store owners evolve values schemas either manually or through pushjobs with auto-register value schema config. All value schemas are persisted in znode.

Overtime the set of value schemas could increase and create huge bloat.
Historical schemas could contain incompatible schema which prevent uses from evolving new schema.
This also might affect creation of superset schema with lots of unnecessary fields which are not even used.
This PR will try to trim set of unused value schemas.

The set of write schemas id used during ingestion are stored in system meta store for each store versions.

1. The parent controller would periodically scan the child colo for above set the of active schemas being used.

2. It will then consolidate the schema set and remove the past historical schemas from both parent and child colos to keep them in sync.
3. Currently will work with batch only stores and skip system meta stores.
4. Will only delete schemas with schema id less than active set of schema id.

The configs added

"controller.min.schema.count.keep" only delete schema for stores with this many schema count. default 20.
"controller.unused.value.schema.cleanup.enabled" to enable this feature
"controller.unused.schema.cleanup.interval.mins" interval of this cleanup task.


clients/venice-admin-tool/src/main/java/com/linkedin/venice/Command.java
```diff
@@ -499,6 +499,10 @@ public enum Command {
   BACKUP_STORE_METADATA_FROM_GRAVEYARD(
       "backup-store-metadata-from-graveyard", "Backup store metadata from graveyard in EI",
       new Arg[] { VENICE_ZOOKEEPER_URL, ZK_SSL_CONFIG_FILE, BACKUP_FOLDER }
+  ),
+  DELETE_VALUE_SCHEMA(
```
xunyin8, line 503: General question, will this break superset schema? Do we need to regenerate superset schema upon removing old schemas?

services/venice-controller/src/main/java/com/linkedin/venice/controller/UnusedValueSchemaCleanupService.java
```diff
@@ -0,0 +1,127 @@
...
+      VeniceHelixAdmin admin,
+      VeniceParentHelixAdmin parentHelixAdmin) {
+    this.multiClusterConfig = multiClusterConfig;
+    this.admin = admin;
+    this.scheduleIntervalMinutes = multiClusterConfig.getUnusedSchemaCleanupIntervalMinutes();
+    this.minSchemaCountToKeep = multiClusterConfig.getMinSchemaCountToKeep();
+    this.veniceParentHelixAdmin = parentHelixAdmin;
+  }
+
+  private Runnable getRunnableForSchemaCleanup() {
```
xunyin8, line 40: We should catch and report/log exceptions that may be thrown from schema deletion so we know if the feature is working for all stores or not.

services/venice-controller/src/main/java/com/linkedin/venice/controller/UnusedValueSchemaCleanupService.java
```diff
@@ -0,0 +1,127 @@
...
+            // skip latest value schema or super-set schema id
+            if (schemaId == store.getLatestSuperSetValueSchemaId() || admin.getHelixVeniceClusterResources(clusterName)
+                .getSchemaRepository()
+                .getSupersetOrLatestValueSchema(storeName)
+                .getId() == schemaId) {
+              continue;
+            }
+
+            // delete only if its not used and less than minimum of used schema id
+            if (!usedSchemaSet.contains(schemaId) && schemaId < minSchemaIdInUse) {
```
xunyin8, line 85: IIUC this is to prevent deletion of new schemas that the user might have added but haven't got a chance to push data with it right? Can we have an unit test for this specific scenario too?

services/venice-controller/src/main/java/com/linkedin/venice/controller/UnusedValueSchemaCleanupService.java
```diff
@@ -0,0 +1,127 @@
...
+              schemasToDelete.add(schemaId);
+              // maintain minimum of SCHEMA_COUNT_THRESHOLD schemas in repo
+              if (schemasToDelete.size() > allSchemas.size() - minSchemaCountToKeep) {
+                break;
+              }
+            }
+          }
+          // delete from child colos
+          admin.deleteValueSchemas(clusterName, store.getName(), schemasToDelete);
```
xunyin8, line 94: This ordering is important, in this case since the service is running in parent controller we should actually call `veniceParentHelixAdmin.deleteValueSchemas` before ` admin.deleteValueSchemas`. This way if one of the colo is not reachable or throwing exceptions we get to retry the deletion on the next run because we haven't deleted the unused schemas in parent yet. We should add an unit test for this scenario (where one colo is temporarily unreachable).

services/venice-controller/src/test/java/com/linkedin/venice/controller/TestUnusedValueSchemaCleanupService.java
```diff
@@ -0,0 +1,87 @@
...
+    UnusedValueSchemaCleanupService service = new UnusedValueSchemaCleanupService(config, admin, parentHelixAdmin);
+
+    service.startInner();
+    Set<Integer> unusedSchemas = new HashSet<>();
+    unusedSchemas.add(1);
+    unusedSchemas.add(2);
+    TestUtils.waitForNonDeterministicAssertion(
+        1,
+        TimeUnit.SECONDS,
+        () -> verify(admin, atLeast(1)).deleteValueSchemas(clusterName, store.getName(), unusedSchemas));
```
xunyin8, line 85: I think we should know exactly how many schemas will be deleted so can we use times(x) instead of atLeast(1)?

services/venice-controller/src/main/java/com/linkedin/venice/controller/UnusedValueSchemaCleanupService.java
```diff
@@ -0,0 +1,127 @@
...
+          }
+          Collection<SchemaEntry> allSchemas = veniceParentHelixAdmin.getValueSchemas(clusterName, storeName);
+
+          if (allSchemas.size() < minSchemaCountToKeep) {
+            continue;
+          }
+
+          Set<Integer> usedSchemaSet = veniceParentHelixAdmin.getInUseValueSchemaIds(clusterName, storeName);
+
+          if (usedSchemaSet.isEmpty()) {
```
xunyin8, line 67: Can we also have an unit test to make sure in the case where we are unable to `getInUseValueSchemaIds` due to errors or unreachable child colo, the schema deletion is skipped just like what you implemented here? (This is to make sure this behavior stays when some of these APIs get modified in the future to be used for other things as well)

xunyin8 (commented):
Mostly looks good, just a few concerns and request to have more unit tests.

sixpluszero:
Hmm, I have a question actually: If we delete certain schema IDs, for example, we have 1-5, and we delete 1, 3 due to inactivity. Will next schema upgrade from user add schema v6 or it will find the next available slot (say v1 in our case)?

majisourav99:
> Hmm, I have a question actually: If we delete certain schema IDs, for example, we have 1-5, and we delete 1, 3 due to inactivity. Will next schema upgrade from user add schema v6 or it will find the next available slot (say v1 in our case)?

Its will pick v6 (max schema id + 1) as the next schema.

clients/venice-admin-tool/src/main/java/com/linkedin/venice/Command.java
majisourav99: It should not, super-set schema might still contain fields from a deleted schema, but it wont be read. You might argue about optimization of the size of the super-set schema if older schema fields are dropped later, but most schema evolution are addition rather than deletion of fields.

services/venice-controller/src/main/java/com/linkedin/venice/controller/UnusedValueSchemaCleanupService.java
majisourav99: Sure

services/venice-controller/src/main/java/com/linkedin/venice/controller/UnusedValueSchemaCleanupService.java
majisourav99: Good point!  I will add a test case for that.

services/venice-controller/src/main/java/com/linkedin/venice/controller/UnusedValueSchemaCleanupService.java
xunyin8: I still don't see the unit test include a scenario for this, did I miss it?

services/venice-controller/src/main/java/com/linkedin/venice/controller/UnusedValueSchemaCleanupService.java
xunyin8: Same here, a simple unit test will do.

xunyin8 (dismissed):
LGTM thanks!

services/venice-controller/src/main/java/com/linkedin/venice/controller/UnusedValueSchemaCleanupService.java
```diff
@@ -0,0 +1,129 @@
...
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+
+
+public class UnusedValueSchemaCleanupService extends AbstractVeniceService {
```
gaojieliu, line 19: Can we add some javadoc to this new class?

services/venice-controller/src/main/java/com/linkedin/venice/controller/UnusedValueSchemaCleanupService.java
```diff
@@ -0,0 +1,129 @@
...
+          Collection<SchemaEntry> allSchemas = veniceParentHelixAdmin.getValueSchemas(clusterName, storeName);
+
+          if (allSchemas.size() < minSchemaCountToKeep) {
+            continue;
+          }
+
+          Set<Integer> usedSchemaSet = veniceParentHelixAdmin.getInUseValueSchemaIds(clusterName, storeName);
+
+          // if any of the child colo is unreachable, skip deletion.
+          if (usedSchemaSet.isEmpty()) {
```
gaojieliu, line 68: Logging?

services/venice-controller/src/main/java/com/linkedin/venice/controller/UnusedValueSchemaCleanupService.java
```diff
@@ -0,0 +1,129 @@
...
+            // skip latest value schema or super-set schema id
+            if (schemaId == store.getLatestSuperSetValueSchemaId() || admin.getHelixVeniceClusterResources(clusterName)
+                .getSchemaRepository()
+                .getSupersetOrLatestValueSchema(storeName)
+                .getId() == schemaId) {
+              continue;
+            }
+
+            // delete only if its not used and less than minimum of used schema id
+            if (!usedSchemaSet.contains(schemaId) && schemaId < minSchemaIdInUse) {
```
gaojieliu, line 86: I hope we can wrap this filtering logic into a function and write enough unit tests against it to make sure there is no surprise.

services/venice-controller/src/main/java/com/linkedin/venice/controller/UnusedValueSchemaCleanupService.java
```diff
@@ -0,0 +1,129 @@
...
+              // maintain minimum of SCHEMA_COUNT_THRESHOLD schemas in repo
+              if (schemasToDelete.size() > allSchemas.size() - minSchemaCountToKeep) {
+                break;
+              }
+            }
+          }
+          // delete from child colos before parent
+          if (veniceParentHelixAdmin.deleteValueSchemas(clusterName, store.getName(), schemasToDelete)) {
+            // delete from parent only if child colo deletion succeeds
+            admin.deleteValueSchemas(clusterName, store.getName(), schemasToDelete);
```
gaojieliu, line 97: Logging?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerConfig.java
```diff
@@ -543,6 +551,11 @@ public VeniceControllerConfig(VeniceProperties props) {
         props.getBoolean(KME_REGISTRATION_FROM_MESSAGE_HEADER_ENABLED, false);
     this.enableDisabledReplicaEnabled = props.getBoolean(CONTROLLER_ENABLE_DISABLED_REPLICA_ENABLED, false);
 
+    this.unusedValueSchemaCleanupServiceEnabled =
```
gaojieliu, line 554: Should we only consider enabling this service in parent controller only?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
```diff
@@ -624,6 +631,46 @@ private void sendAdminMessageAndWaitForConsumed(String clusterName, String store
     waitingMessageToBeConsumed(clusterName, storeName, message.executionId);
   }
 
+  @Override
+  public boolean deleteValueSchemas(String clusterName, String storeName, Set<Integer> inuseValueSchemaIds) {
```
gaojieliu, line 635: Shall we enforce the unused schema id check here? Allowing random value schema deletion is very risky.

internal/venice-common/src/main/java/com/linkedin/venice/system/store/MetaStoreWriter.java
```diff
@@ -251,6 +252,40 @@ public void writeReadyToServerStoreReplicas(
     });
   }
 
+  /**
+   * This function should be invoked during ingestion when execution status changes.
+   */
+  public void writeStoreReplicaStatus(
```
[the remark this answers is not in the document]
majisourav99: Good catch! probably leaked from rebasing. will remove it.

services/venice-controller/src/main/java/com/linkedin/venice/controller/UnusedValueSchemaCleanupService.java
majisourav99: Sure

services/venice-controller/src/main/java/com/linkedin/venice/controller/UnusedValueSchemaCleanupService.java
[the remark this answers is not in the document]
majisourav99: Yeah, I assumed there wont be gap in schemas being used in general. I could fine grain the set as you suggested.

services/venice-controller/src/main/java/com/linkedin/venice/controller/UnusedValueSchemaCleanupService.java
```diff
@@ -0,0 +1,129 @@
...
+            if (!usedSchemaSet.contains(schemaId) && schemaId < minSchemaIdInUse) {
+              schemasToDelete.add(schemaId);
+              // maintain minimum of SCHEMA_COUNT_THRESHOLD schemas in repo
+              if (schemasToDelete.size() > allSchemas.size() - minSchemaCountToKeep) {
+                break;
+              }
+            }
+          }
+          // delete from child colos before parent
+          if (veniceParentHelixAdmin.deleteValueSchemas(clusterName, store.getName(), schemasToDelete)) {
```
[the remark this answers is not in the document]
majisourav99: I could use admin channel, but Xun raised a point where if a child colo is unreachable, then we will delete the schema from parent even if the child colo could not delete them. I guess its fine since the admin channel will retry and delete it eventually?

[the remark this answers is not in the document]
gaojieliu: Yeah, the eventual consistency should be fine. 
For almost all the write operations, they should go through admin channel to achieve the similar guarantee about ordering and eventual consistency, otherwise, it will become hard to reason in the future.

clients/venice-admin-tool/src/main/java/com/linkedin/venice/AdminTool.java
```diff
@@ -734,6 +736,14 @@ private static void estimateDataRecoveryTime(CommandLine cmd) {
     }
   }
 
+  private static void deleteValueSchema(CommandLine cmd) {
+    String storeName = getRequiredArgument(cmd, Arg.STORE);
```
gaojieliu, line 740: So if we implement the auto schema cleanup logic, why do we still need the manual cleanup?
Manual cleanup can be unsafe without proper check.

internal/venice-common/src/main/java/com/linkedin/venice/system/store/MetaStoreWriter.java
```diff
@@ -360,6 +365,10 @@ private void update(
     });
   }
 
+  HelixReadOnlySchemaRepository getSchemaRepository() {
```
gaojieliu, line 368: So we don't clean up the unused schemas from meta store?

services/venice-controller/src/main/java/com/linkedin/venice/controller/UnusedValueSchemaCleanupService.java
```diff
@@ -0,0 +1,151 @@
...
+
+          Set<Integer> schemasToDelete = findSchemaIdsToDelete(
+              allSchemas,
+              store,
+              admin.getHelixVeniceClusterResources(clusterName).getSchemaRepository(),
+              inUseValueSchemaIds);
+
+          if (!schemasToDelete.isEmpty()) {
+            // delete from child colos
+            veniceParentHelixAdmin.deleteValueSchemas(clusterName, store.getName(), schemasToDelete);
```
gaojieliu, line 89: This is interesting, and this command is supposed to delete the unused schemas from both parent and child, no?

services/venice-controller/src/main/java/com/linkedin/venice/controller/UnusedValueSchemaCleanupService.java
```diff
@@ -0,0 +1,151 @@
...
+  }
+
+  Set<Integer> findSchemaIdsToDelete(
+      Collection<SchemaEntry> allSchemas,
+      Store store,
+      ReadWriteSchemaRepository schemaRepository,
+      Set<Integer> inUseValueSchemaIds) {
+    Set<Integer> schemasToDelete = new HashSet<>();
+
+    // assumes `getValueSchemas` returns ascending schema ids so that the older schemas are deleted first
```
gaojieliu, line 118: Let us not assume this order, and can we sort it before the cleanup?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -3965,6 +3969,61 @@ void setAccessControl(String clusterName, String storeName, boolean accessContro
...
+    Set<Integer> schemaIds = new HashSet<>();
+
+    // Fetch value schema id used by all existing store version
+    for (Version version: store.getVersions()) {
+      Map<String, String> map = new HashMap<>(2);
+      map.put(MetaStoreWriter.KEY_STRING_VERSION_NUMBER, Integer.toString(version.getNumber()));
+      StoreMetaKey key = MetaStoreDataType.VALUE_SCHEMAS_WRITTEN_PER_STORE_VERSION.getStoreMetaKey(map);
+      StoreMetaValue metaValue = getMetaStoreValue(key, storeName);
+
+      if (metaValue == null) {
```
gaojieliu, line 4000: `null` means that the meta store is not working properly, right?
Shall we throw an exception instead?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
```diff
@@ -624,6 +632,52 @@ private void sendAdminMessageAndWaitForConsumed(String clusterName, String store
     waitingMessageToBeConsumed(clusterName, storeName, message.executionId);
   }
 
+  @Override
+  public void deleteValueSchemas(String clusterName, String storeName, Set<Integer> unusedValueSchemaIds) {
+    Set<Integer> inuseValueSchemaIds = getInUseValueSchemaIds(clusterName, storeName);
```
gaojieliu, line 637: I assume this code path is for the admin operation, right?
Shall we get rid of the admin tool command to make it simpler?

internal/venice-common/src/main/java/com/linkedin/venice/system/store/MetaStoreWriter.java
majisourav99: I dont recall why I skipped system-stores, but given system schema are rarely updated, it might be ok to skip them ?

services/venice-controller/src/main/java/com/linkedin/venice/controller/UnusedValueSchemaCleanupService.java
majisourav99: Good catch! yeah right, this is a leak from previous code without admin channel call.

services/venice-controller/src/main/java/com/linkedin/venice/controller/UnusedValueSchemaCleanupService.java
majisourav99: sure

services/venice-controller/src/main/java/com/linkedin/venice/controller/UnusedValueSchemaCleanupService.java
```diff
@@ -0,0 +1,151 @@
...
+    for (SchemaEntry schemaEntry: allSchemas) {
+      int schemaId = schemaEntry.getId();
+      // skip latest value schema or super-set schema id
+      if (schemaId == store.getLatestSuperSetValueSchemaId()
+          || schemaRepository.getSupersetOrLatestValueSchema(store.getName()).getId() == schemaId) {
```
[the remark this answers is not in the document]
majisourav99: `store.getLatestSuperSetValueSchemaId()` returns the actual superset schema and store repo one returns either the superschema schema or latest schema. I could use the later one, but its just a double sanity check.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
majisourav99: Sure, but first lets wait for @FelixGV comment on this.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
majisourav99: also it could be just a newly populated version which has no writes yet ?

internal/venice-common/src/main/java/com/linkedin/venice/system/store/MetaStoreWriter.java
gaojieliu: Actually, I was referring to the following question:
When we cleanup unused schema from backend/ZK, do we clean these schema from meta store for the corresponding Venice store as well?

services/venice-controller/src/main/java/com/linkedin/venice/controller/UnusedValueSchemaCleanupService.java
```diff
@@ -0,0 +1,156 @@
...
+    return () -> {
+      if (stop) {
+        return;
+      }
+      for (String clusterName: multiClusterConfig.getClusters()) {
+        boolean cleanupEnabled =
+            multiClusterConfig.getControllerConfig(clusterName).isUnusedValueSchemaCleanupServiceEnabled();
+
+        // Get all stores for current cluster
+        List<Store> stores = admin.getAllStores(clusterName);
```
gaojieliu, line 59: I think `VeniceHelixParentAdmin` supports `getAllStores` call as well, so what is the main reason to use both internal admin and parent admin here?

services/venice-controller/src/main/java/com/linkedin/venice/controller/UnusedValueSchemaCleanupService.java
```diff
@@ -0,0 +1,156 @@
...
+              admin.getHelixVeniceClusterResources(clusterName).getSchemaRepository(),
+              inUseValueSchemaIds);
+
+          if (!schemasToDelete.isEmpty()) {
+            LOGGER.info(
+                "In cluster {}, store {} has the following unused schemas {}.",
+                clusterName,
+                storeName,
+                schemasToDelete);
+            if (cleanupEnabled) {
```
gaojieliu, line 92: I saw if you have already checked `cleanupEnabled` before the loop, so this is redundant, right?

internal/venice-common/src/main/java/com/linkedin/venice/system/store/MetaStoreWriter.java
majisourav99: It stores schema per store version, so when its deleted from zk, it wont be calling the delete method again.

services/venice-controller/src/main/java/com/linkedin/venice/controller/UnusedValueSchemaCleanupService.java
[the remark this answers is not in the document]
gaojieliu: So if the store doesn't have `superset schema` feature enabled, `store.getLatestSuperSetValueSchemaId()` will return `-1`, right?
I think we should make the code clean, otherwise, it is confusing.

services/venice-controller/src/main/java/com/linkedin/venice/controller/UnusedValueSchemaCleanupService.java
```diff
@@ -0,0 +1,156 @@
...
+    this.minSchemaCountToKeep = multiClusterConfig.getMinSchemaCountToKeep();
+    this.veniceParentHelixAdmin = parentHelixAdmin;
+  }
+
+  private Runnable getRunnableForSchemaCleanup() {
+    return () -> {
+      if (stop) {
+        return;
+      }
+      for (String clusterName: multiClusterConfig.getClusters()) {
```
gaojieliu, line 54: Shall we only run this service in the leader controller of the super cluster?
Right now, I didn't see such condition check, and that indicates this service will run in every parent controller, right?

services/venice-controller/src/main/java/com/linkedin/venice/controller/UnusedValueSchemaCleanupService.java
majisourav99: No, I removed the earlier check, so in a way it will collect data for unused schema without deletion.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerConfig.java
```diff
@@ -553,6 +561,11 @@ public VeniceControllerConfig(VeniceProperties props) {
...
+        props.getBoolean(CONTROLLER_UNUSED_VALUE_SCHEMA_CLEANUP_ENABLED, false);
+    this.unusedSchemaCleanupIntervalMinutes = props.getInt(CONTROLLER_UNUSED_SCHEMA_CLEANUP_INTERVAL_MINS, 600);
```
gaojieliu, line 566: So the default interval is 600mins = 10 hours?

services/venice-controller/src/main/java/com/linkedin/venice/controller/UnusedValueSchemaCleanupService.java
```diff
@@ -0,0 +1,156 @@
...
+        boolean cleanupEnabled =
+            multiClusterConfig.getControllerConfig(clusterName).isUnusedValueSchemaCleanupServiceEnabled();
+
+        // Get all stores for current cluster
+        List<Store> stores = admin.getAllStores(clusterName);
+        for (Store store: stores) {
+          String storeName = store.getName();
```
gaojieliu, line 61: Shall we wrap `try/catch` around one store cleanup logic, so that one problematic store won't block other stores' cleanup?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
gaojieliu: Sure, either way, I think we should throw exception to make sure we don't delete the schema, which can be potentially used soon.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -3965,6 +3969,61 @@ void setAccessControl(String clusterName, String storeName, boolean accessContro
...
+    }
+    return schemaIds;
+  }
+
+  public void deleteValueSchemas(String clusterName, String storeName, Set<Integer> unusedValueSchemaIds) {
+    Set<Integer> inuseValueSchemaIds = getInUseValueSchemaIds(clusterName, storeName);
+    boolean isCommon = unusedValueSchemaIds.stream().anyMatch(inuseValueSchemaIds::contains);
+    if (isCommon) {
+      LOGGER
+          .error("For store {} cannot delete value schema ids {} as they being used.", storeName, unusedValueSchemaIds);
```
gaojieliu, line 4017: Shall we throw exception here? Otherwise, this might results in inconsistency between child regions.

services/venice-controller/src/main/java/com/linkedin/venice/controller/UnusedValueSchemaCleanupService.java
gaojieliu: I see.

internal/venice-common/src/main/java/com/linkedin/venice/system/store/MetaStoreWriter.java
gaojieliu: I understood about the used schema tracking in meta store.
In meta store, we also store all the schemas and if we decide to delete unused schemas from ZK, we should delete these schemas from meta store as well, right?

services/venice-controller/src/main/java/com/linkedin/venice/controller/UnusedValueSchemaCleanupService.java
majisourav99: Good point, will add the leader check.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerConfig.java
majisourav99: yes, is that too large ?

services/venice-controller/src/main/java/com/linkedin/venice/controller/UnusedValueSchemaCleanupService.java
majisourav99: Sure, good to skip on bad stores.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerConfig.java
gaojieliu: That should be fine considering schema cleanup is not an urgent task.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
majisourav99: Sure

internal/venice-common/src/main/java/com/linkedin/venice/system/store/MetaStoreWriter.java
majisourav99: Good catch. updated to delete meta store schemas

gaojieliu (approved):
Thanks!

PR #838, opened 2024-01-30
[changelog] Use serializer cache to improve changecapture user performance

ZacAttack:
## [changelog] Use serializer cache to improve changecapture user performance

The deserializer lookup is pretty expensive.  This change should speed things up.

Resolves #XXX

## How was this PR tested?
normal test suite

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumerImpl.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumerImpl.java
@@ -110,6 +110,8 @@ public class VeniceChangelogConsumerImpl<K, V> implements VeniceChangelogConsume
 
   protected final ChunkAssembler chunkAssembler;
 
+  protected final AvroStoreDeserializerCache avrostoreDeserializerCache;
+
   public VeniceChangelogConsumerImpl(
       ChangelogClientConfig changelogClientConfig,
       PubSubConsumerAdapter pubSubConsumer) {
@@ -146,6 +148,8 @@ public class VeniceChangelogConsumerImpl<K, V> implements VeniceChangelogConsume
       this.userEventChunkingAdapter = GenericChunkingAdapter.INSTANCE;
       this.storeDeserializerCache = new AvroStoreDeserializerCache<>(storeRepository, storeName, true);
     }
+
+    avrostoreDeserializerCache = new AvroStoreDeserializerCache(storeRepository, storeName, true);
     LOGGER.info(
         "Start a change log consumer client for store: {}, with partition count: {} and view class: {} ",
         storeName,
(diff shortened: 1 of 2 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumerImpl.java
```diff
@@ -146,6 +148,8 @@ public VeniceChangelogConsumerImpl(
       this.userEventChunkingAdapter = GenericChunkingAdapter.INSTANCE;
       this.storeDeserializerCache = new AvroStoreDeserializerCache<>(storeRepository, storeName, true);
     }
+
+    avrostoreDeserializerCache = new AvroStoreDeserializerCache(storeRepository, storeName, true);
```
sixpluszero, line 152: looks good to me, just a nit: you could move this line up, and line 149 can reuse the same cache object for generic client?

sixpluszero (dismissed):

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumerImpl.java
ZacAttack: ugh!  Great point.

ZacAttack: fixed.

sixpluszero: Uh I don't think your PR will pass the unit test coverage. As I think last run should already failed on this https://github.com/linkedin/venice/actions/runs/7704830781

ZacAttack: on it.  This should pass

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumerImpl.java
```diff
@@ -761,9 +762,7 @@ private PubSubMessage<K, ChangeEvent<V>, VeniceChangeCoordinate> convertChangeEv
   }
 
   private V deserializeValueFromBytes(ByteBuffer byteBuffer, int valueSchemaId) {
-    Schema currentValueSchema = schemaReader.getValueSchema(valueSchemaId);
-    RecordDeserializer<V> valueDeserializer =
-        FastSerializerDeserializerFactory.getFastAvroGenericDeserializer(currentValueSchema, currentValueSchema);
+    RecordDeserializer<V> valueDeserializer = storeDeserializerCache.getDeserializer(valueSchemaId, valueSchemaId);
```
FelixGV, line 765: Not related to this optimization, but...

Using the same schema for reader and writer is typically frowned upon... is this the right behavior? I am wondering if we should align it with the behavior of the GenericRecord thin client, which always upconverts to the latest schema (i.e. latest as the reader schema). Writer schema must be whatever the payload came with, of course.

FelixGV (commented):
Change itself looks good but I left a more general question...

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumerImpl.java
ZacAttack: That makes sense, I can take that change in another PR (Ads team is a bit hot to try this out to determine if we need to do something more dramatic to get the performance better).  I'll file a BB

FelixGV: LFG! 🚀

FelixGV (approved):

PR #839, opened 2024-01-30
[server] Fix NPE in MergeConflictResolver#ignoreNewUpdate()

sixpluszero:
## [server] Fix NPE in MergeConflictResolver#ignoreNewUpdate()
This PR fixes NPE thrown in MergeConflictResolver#ignoreNewUpdate(): When checking a new UPDATE message, we will first check message timestamp and compared with old value RMD. The existing logic gets all the fields from update record and check operation type for each field. Here it makes a wrong assumption that field value cannot be null, and throw exception when a new field is being set to null.
The PR fixes the issue, removes code duplications in util methods.

## How was this PR tested?

New unit tests in different levels.

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/replication/merge/MergeConflictResolver.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/replication/merge/MergeConflictResolver.java
@@ -7,7 +7,7 @@ import static com.linkedin.venice.schema.rmd.v1.CollectionRmdTimestamp.PUT_ONLY_
 import static com.linkedin.venice.schema.rmd.v1.CollectionRmdTimestamp.TOP_LEVEL_COLO_ID_FIELD_POS;
 import static com.linkedin.venice.schema.rmd.v1.CollectionRmdTimestamp.TOP_LEVEL_TS_FIELD_POS;
 import static com.linkedin.venice.schema.writecompute.WriteComputeOperation.NO_OP_ON_FIELD;
-import static com.linkedin.venice.schema.writecompute.WriteComputeSchemaConverter.getFieldOperationType;
+import static com.linkedin.venice.schema.writecompute.WriteComputeOperation.getFieldOperationType;
 
 import com.linkedin.davinci.replication.RmdWithValueSchemaId;
 import com.linkedin.davinci.schema.merge.ValueAndRmd;
--- a/internal/venice-client-common/src/main/java/com/linkedin/venice/schema/writecompute/WriteComputeSchemaConverter.java
+++ b/internal/venice-client-common/src/main/java/com/linkedin/venice/schema/writecompute/WriteComputeSchemaConverter.java
@@ -4,7 +4,6 @@ import static com.linkedin.venice.schema.writecompute.WriteComputeConstants.WRIT
 import static com.linkedin.venice.schema.writecompute.WriteComputeOperation.LIST_OPS;
 import static com.linkedin.venice.schema.writecompute.WriteComputeOperation.MAP_OPS;
 import static com.linkedin.venice.schema.writecompute.WriteComputeOperation.NO_OP_ON_FIELD;
-import static com.linkedin.venice.schema.writecompute.WriteComputeOperation.PUT_NEW_FIELD;
 import static org.apache.avro.Schema.Field;
 import static org.apache.avro.Schema.Type;
 import static org.apache.avro.Schema.Type.RECORD;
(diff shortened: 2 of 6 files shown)


sixpluszero:
I don't know how to comment on the code that I did not make change. What I want to point out that the ignoreNewUpdate logic looks a bit weird still, as it does not actually check if certain field in WC is NO_OP. Feel free to comment if you think if we need to add check on this. 
UpdateBuilderImpl already makes sure that we cannot build an empty UPDATE record, so at least one field should be touched.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/replication/merge/MergeConflictResolver.java
```diff
@@ -7,7 +7,7 @@
 import static com.linkedin.venice.schema.rmd.v1.CollectionRmdTimestamp.TOP_LEVEL_COLO_ID_FIELD_POS;
 import static com.linkedin.venice.schema.rmd.v1.CollectionRmdTimestamp.TOP_LEVEL_TS_FIELD_POS;
 import static com.linkedin.venice.schema.writecompute.WriteComputeOperation.NO_OP_ON_FIELD;
-import static com.linkedin.venice.schema.writecompute.WriteComputeSchemaConverter.getFieldOperationType;
+import static com.linkedin.venice.schema.writecompute.WriteComputeOperation.getFieldOperationType;
```
FelixGV, line 10: Is this simple change the main code fix?

internal/venice-client-common/src/main/java/com/linkedin/venice/schema/writecompute/WriteComputeSchemaConverter.java
```diff
@@ -446,26 +443,4 @@ public Schema getNoOpOperation(String namespace) {
     noOpSchema.setFields(Collections.emptyList());
     return noOpSchema;
   }
-
-  public static WriteComputeOperation getFieldOperationType(Object writeComputeFieldValue) {
-    Utils.notNull(writeComputeFieldValue);
```
FelixGV, line 451: So I guess the fix is that the copy of this function in `WriteComputeOperation` skips this `notNull` check. This is fine. `null instanceof blah` returns false, so the rest of the code is good even without this null check.

FelixGV (approved):
I looked into it further. I think the change is good, though I have a concern about some other pre-existing part of the code... I don't think we should try to fix it all at once as the other issue is likely a can of worms... but I hope we can discuss it and brainstorm ideas to solve it.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/replication/merge/MergeConflictResolver.java
FelixGV: Yep, looking at the code further, I understand now...

internal/venice-client-common/src/main/java/com/linkedin/venice/schema/writecompute/WriteComputeSchemaConverter.java
sixpluszero: Yeah this is true.

internal/venice-client-common/src/main/java/com/linkedin/venice/schema/writecompute/WriteComputeSchemaConverter.java
```diff
@@ -446,26 +443,4 @@ public Schema getNoOpOperation(String namespace) {
...
-        return NO_OP_ON_FIELD;
-      }
-
-      if (writeComputeFieldSchemaName.endsWith(LIST_OPS.name)) {
-        return LIST_OPS;
-      }
-
-      if (writeComputeFieldSchemaName.endsWith(MAP_OPS.name)) {
-        return MAP_OPS;
-      }
```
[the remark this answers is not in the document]
sixpluszero: I see your concern, but I think it is more than just namespace right? 
For example, here is one field schema I print out that get passed into the util method:

> {"type":"record","name":"nullableMapFieldMapOps","namespace":"com.linkedin.avro","fields":[{"name":"mapUnion","type":{"type":"map","values":"int"},"default":{}},{"name":"mapDiff","type":{"type":"array","items":"string"},"default":[]}]}

For this one, if you when to instanceOf MapOps, then I think even the internal operation types should match right? But for different map type, it can be different?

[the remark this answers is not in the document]
FelixGV: Yeah, the type of the item in the collection can change, which makes this tricky... I wonder if there's a better way 🤔 ...

PR #841, opened 2024-01-31
[test] Stabilize the flaky test 'testStuckConsumerRepair'

lluwm:
This change aims to stabilize the flaky test 'testStuckConsumerRepair' with the following modifications:

1. Reduce the number of servers in cluster from 2 to 1.
2. Avoid pushing the 2nd data version as it will achieve the same testing purpose.

## How was this PR tested?
Internal CI.

### DIFF
--- a/internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestStuckConsumerRepair.java
+++ b/internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestStuckConsumerRepair.java
@@ -79,7 +79,7 @@ public class TestStuckConsumerRepair {
   private static VeniceClusterWrapper setUpCluster() {
     Properties extraProperties = new Properties();
     extraProperties.setProperty(DEFAULT_MAX_NUMBER_OF_PARTITIONS, "5");
-    VeniceClusterWrapper cluster = ServiceFactory.getVeniceCluster(1, 0, 1, 2, 1000000, false, false, extraProperties);
+    VeniceClusterWrapper cluster = ServiceFactory.getVeniceCluster(1, 0, 1, 1, 1000000, false, false, extraProperties);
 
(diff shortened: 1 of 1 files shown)


internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestStuckConsumerRepair.java
```diff
@@ -79,7 +79,7 @@ public void setUp() {
   private static VeniceClusterWrapper setUpCluster() {
     Properties extraProperties = new Properties();
     extraProperties.setProperty(DEFAULT_MAX_NUMBER_OF_PARTITIONS, "5");
-    VeniceClusterWrapper cluster = ServiceFactory.getVeniceCluster(1, 0, 1, 2, 1000000, false, false, extraProperties);
+    VeniceClusterWrapper cluster = ServiceFactory.getVeniceCluster(1, 0, 1, 1, 1000000, false, false, extraProperties);
```
sixpluszero, line 82: Why reducing server count here? Could you explain a bit? I mean, extra server should not be impacting the test right...?

lluwm: For simplification and to avoid the possibility that one host may not getting any partition assigned, thus not triggering the repair service to kick in at all. If that happens, the following assert would fail.

sixpluszero: I see, so that is the reason of the flakiness if I understand your comment correctly. Thanks!

sixpluszero (approved):

lluwm:
Thanks for the review @sixpluszero! Proceed to check in.

PR #842, opened 2024-02-01
[test] Close PubSubBrokerWrapper in TestActiveActiveIngestion

lluwm:
I noticed some KafkaBrokerWrapper leaks were reported when running the integration test in our internal CI. This is a trivial PR to fix these leaks by simply moving them into a try-with-resource block. This is no meaningful changes otherwise.


FelixGV (approved):
Nice, thank you!

PR #843, opened 2024-02-01
[common][router] Fixed redirect request for migrated store

xunyin8:
## [common][router] Fixed redirect request for migrated store
StoreAclHandler was trying to get the Store object for a is system store check. This step was before the VeniceVersionFinder and VenicePathParser which is where we perform the redirect for migrated store.

Modified the StoreAclHandler to check system store by name without accessing the Store object. Alternatively we could add a new SimpleChannelInboundHandler to perform the check as early as possible in the pipeline. However that would involve a lot of change since the 301 response is currently built as part of the ScatterGatherHelper. Other requests like metadata does not support redirect so it's a bit of an overkill to do that.
## How was this PR tested?


gaojieliu (approved):

We constantly discover issue with acl handler since our integration test doesn't cover ACL handlers.

To avoid future problems, I feel we might need to simulate acl handlers in Venice integration tests, so that acl related issues can be caught in the feature.
We don't have to address this here, and we can cut a ticket to track this.

PR #844, opened 2024-02-01
[producer][samza] Reinitalize transport client if a store has moved

ZacAttack:
## [producer][samza] Reinitalize transport client if a store has moved

This is to allow store moves to not break system producers.

Resolves #XXX


FelixGV (approved):
Thanks Zac!

PR #849, opened 2024-02-07
[controller][common] Check for killed pushjob before disabling a partition

majisourav99:
## Check if pushjob is killed before disabling a partition

Venice disables a leader replica if it hits ingestion error during future version push. But if the push job is killed then disabled partition metric creates confusion as to if those are disabled due to real ingestion error or not. This PR skips disabling such error partitions.

## How was this PR tested?

GHCI

### DIFF
--- a/internal/venice-common/src/main/java/com/linkedin/venice/pushmonitor/PushStatusDecider.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/pushmonitor/PushStatusDecider.java
@@ -192,14 +192,16 @@ public abstract class PushStatusDecider {
     boolean isLeaderCompleted = true;
     int previouslyDisabledErrorReplica = 0;
     for (Map.Entry<Instance, HelixState> entry: instanceToStateMap.entrySet()) {
-      ExecutionStatus currentStatus =
-          getReplicaCurrentStatus(partitionStatus.getReplicaHistoricStatusList(entry.getKey().getNodeId()));
+      List<StatusSnapshot> snapshotList = partitionStatus.getReplicaHistoricStatusList(entry.getKey().getNodeId());
+      ExecutionStatus currentStatus = getReplicaCurrentStatus(snapshotList);
       if (entry.getValue() == HelixState.LEADER) {
         if (!currentStatus.equals(COMPLETED)) {
           isLeaderCompleted = false;
         }
+        // Disable replica only if error is not from killjob and its not already disabled
         if (currentStatus.equals(ERROR) && callback != null
-            && !callback.isReplicaDisabled(entry.getKey().getNodeId(), partitionStatus.getPartitionId())) {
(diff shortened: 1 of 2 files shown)


internal/venice-common/src/main/java/com/linkedin/venice/pushmonitor/PushStatusDecider.java
```diff
@@ -235,6 +237,16 @@ protected ExecutionStatus getPartitionStatus(
     return STARTED;
   }
 
+  boolean isPushjobKilled(List<StatusSnapshot> snapshotList) {
+    for (StatusSnapshot snapshot: snapshotList) {
+      if (snapshot.getStatus() == ERROR
+          && snapshot.getIncrementalPushVersion().contains("signal to kill this consumer")) {
```
FelixGV, line 243: I guess string comparison is the best we can do ATM, given the way these components communicate, but hardcoding the string is too brittle IMHO. I suggest we create a constant containing `Received the signal to kill this consumer.` and have both places in the code refer to that. We should also add JavaDoc to that constant saying that this string is sent by the server and received by the controller, and should therefore not be changed otherwise the controller will fail to pick up the signal properly if server and controller are running different versions of the code...

FelixGV (commented):
LGTM overall, and thanks for adding the test. Just left a minor comment about adding a constant for the string of interest...

internal/venice-common/src/main/java/com/linkedin/venice/pushmonitor/PushStatusDecider.java
majisourav99: I was just thinking about it, many codebases have centralized Log messages class with some log-ID for each log and rest of the code uses that. Which makes things like canary tests/log stats deliver a better picture of the current state. Also log rate limiting also gets cleaner as it will try to limit by log-id, probably can be universally rate limited for all logs. Anyways will update the code.

internal/venice-common/src/main/java/com/linkedin/venice/LogMessages.java
```diff
@@ -0,0 +1,7 @@
+package com.linkedin.venice;
+
+public class LogMessages {
+  // Replica with ERROR status with the following message implies the job was killed instead of actual ingestion error
```
FelixGV, line 4: Sorry for the nitpick... but can we say that this string is used in server-to-controller communication and that it should not be changed, otherwise a cluster running with mixed versions might not behave properly? Also, let's please use JavaDoc `/** */` instead of `//` so that it populates the JavaDoc and tool tips properly...

majisourav99: Sure, updated the javadoc

FelixGV (approved):
Thanks!

PR #850, opened 2024-02-09
[WIP][dvc] Add user documentation on how to use the DaVinciRecordTransformer, version transformer, and add a metric to keep track of transformation errors

kvargha:
## Summary

1. Added a Read APIs section and a Da Vinci Client subsection, which contains implementation details on how to use the DaVinciRecordTransformer since it didn't previously exist. This can be found [here](https://kvargha.com/venice/docs/user_guide/read_api/da_vinci_client).
2.  Got rid of `TransformedRecord` since it was meant to be used for 1:N transformation, but it will have its own implementation for that.
3. Converted  `DaVinciRecordTransformer` from an interface to an abstract class so we can store `storeVersion` in the class itself so the can perform operations based on the lifecycle.
4. Added a sensor to keep track of errors when performing transformation.
5. Added lifecycle events to `DaVinciRecordTransformer` to monitor the start and end of ingestion.
6. Utilize a functional interface to create a new `DaVinciRecordTransformer` for each ingestion task.
7. Remove deserializeAvroObjectAndReturn in StoreIngestionTask as it's deprecated.

## How was this PR tested?
CI


clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/DaVinciRecordTransformer.java
```diff
@@ -34,6 +34,19 @@ public interface DaVinciRecordTransformer<K, V, O extends TransformedRecord> {
...
+  /**
+   * By default, calls the put function to be implemented.
+   * @param key to be put
+   * @param value to be put
+   * @param newVersionNumber version number being created
+   * @param oldVersionNumber version number being deleted
+   * @param readyToServe boolean of whether or not PartitionConsumptionState is ready to serve
+   * @return the object to keep in storage, or null if the put should be skipped
+   */
+  default O put(Lazy<K> key, Lazy<V> value, int newVersionNumber, int oldVersionNumber, boolean readyToServe) {
```
FelixGV, line 46: I realize that we were ambiguous when we discussed "versions" in the last contributor sync. I think what you're going for in this interface is a notion of "version" akin to MVCC or some other row-level versioning mechanism. We do have an equivalent notion to that in Venice which we call "RMD" (Replication Metadata), and it might be interesting to expose that in the interface, though ATM I don't believe we foresee any use case for that...

We also have another notion of versioning, which is the one we're interested in for this interface expansion, and it is much more coarse-grained. In Venice, a dataset is called "store" and each store contains multiple "store-versions". In the current context, a given write operation (put, delete...) will be applied to a single store-version. Or rather, it may be applied to all existing store-versions, but that will occur via separate invokations of this interface. So in this interface, we would want just a single version number param.

Passing the `readyToServe` param is interesting, and might be a viable design. Alternatively, the state change of becoming ready to serve might be best modeled by having a completely separate function. I guess it'd be good to get @ZacAttack's thoughts on that part...

docs/user_guide/da_vinci_client/da_vinci_client.md
```diff
@@ -0,0 +1,10 @@
+---
+layout: default
+title: Da Vinci Client
+parent: User Guides
+has_children: true
+permalink: /docs/user_guide/da_vinci_client
```
FelixGV, line 6: I'd like to suggest a different structure... We already have a `Write APIs` section under `User Guides`. I think I'd like to have a `Read APIs` page alongside it, and the `Da Vinci Client` page can be nested under that. The Record Transformer stuff can be a section within the DVC page itself, rather than nested as a separate page.

You already figured out the `grand_parent` stuff correctly, so that's good. The above restructuring would also need a `grand_parent` annotation on the deepest level (the DVC page).

docs/user_guide/da_vinci_client/da_vinci_record_transformer.md
```diff
@@ -0,0 +1,41 @@
...
+grand_parent: User Guides
+permalink: /docs/user_guide/da_vinci_client/da_vinci_record_transformer
+---
+
+# Da Vinci Record Transformer
+
+This feature enables applications to transform records as they're being consumed and stored in the Da Vinci Client. It's capable of handling records that are compressed and/or chunked.
+
+## Usage
+To use the record transformer, you will need to implement the `DaVinciRecordTransformer` interface and pass it into the `StoreIngestionTask` constructor. When a message is being consumed, the `DaVinciRecordTransformer` will modify the value before it is written to storage.
```
FelixGV, line 14: Nitpick: we want to respect a 120 characters per line limit in the docs as well. This makes it easier to comment on a specific part of the text, since GH only supports targeting at the granularity of entire lines when commenting. The markdown format ends up being rendered correctly with overflow and everything out of the box, no matter where we cut the lines.

For the `DaVinciRecordTransformer`, I would prefer that we put a link to the online JavaDoc, e.g. http://venicedb.org/javadoc/com/linkedin/davinci/client/DaVinciRecordTransformer.html in this case.

We should mention that the `DVRT` needs to be passed into the `DaVinciConfig`, not into the `SIT` (JavaDoc link: http://venicedb.org/javadoc/com/linkedin/davinci/client/DaVinciConfig.html#setRecordTransformer(com.linkedin.davinci.client.DaVinciRecordTransformer) ...). The transformer eventually makes its way down into the `SIT`, but that is an implementation detail, and we hope to keep this part of the documentation user-facing.

FelixGV (commented):
I see that you also successfully published your version of the docs here: https://kvargha.com/venice/ which is great (even setup a custom domain for it, good job :D !)... it is good to mention that URL in the PR description so that reviewers can verify that the doc renders well, links work, etc.

Left a few comments on both the doc and the new interface... Thanks a lot!

docs/user_guide/read_api/da_vinci_client.md
```diff
@@ -0,0 +1,52 @@
...
+    implements DaVinciRecordTransformer<Integer, String, TransformedRecord<Integer, String>> {
+  public Schema getKeyOutputSchema() {
+    return Schema.create(Schema.Type.INT);
+  }
+
+  public Schema getValueOutputSchema() {
+    return Schema.create(Schema.Type.STRING);
+  }
+
+  public TransformedRecord<Integer, String> put(Lazy<Integer> key, Lazy<String> value) {
```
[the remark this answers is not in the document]
kvargha: As of right now, the key in `TransformedRecord<K, V>` isn't in use. I wasn't part of the design discussion for this, so I'm not aware if a future use case has been identified.

Maybe @elijahgrimaldi can shine some light here?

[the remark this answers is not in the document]
elijahgrimaldi: I believe I implemented this while we were having discussions about how we might want to do 1:N transformations but it was left incomplete and now it just wraps the serialization logic for the transformed value. It is excessive for what we have now so I think we could get rid of it.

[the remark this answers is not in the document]
ZacAttack: Yeah the original intention was to allow for this extension because.... well... at the time it seemed so simple :)

clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/DaVinciRecordTransformer.java
ZacAttack: readyToServe is interesting.  But....  I'm struggling to think through how it'll be implemented.  I 'think' ready to serve right now is a construct that's per partition, and I don't think it would be a good thing to open up to the user the notion that partition 'X' is ready but parittion 'Y' is not.  I think I'd prefer them to view the whole dataset as ready or not.  So maybe let's remove this from the interface and instead rely on a discrete callback for the lifecycle of a version.

docs/user_guide/read_api/da_vinci_client.md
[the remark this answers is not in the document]
kvargha: I can go ahead and remove the key from `TransformedRecord`.

In [StoreIngestionTask](https://github.com/kvargha/venice/blob/88d834d26982e55c63da329bd708b8eacd812b1e/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java#L3144-L3145), after `put` has been called, it calls `getValueBytes` on the `TransformedRecord` object. I'm thinking of moving the object creation logic out of the `put` function and right before `getValueBytes` is called. Does that sound good @FelixGV?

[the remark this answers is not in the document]
elijahgrimaldi: I think we probably don't need the `TransformedRecord` class at all right now actually, if I remember correctly the original intent had to do with implementing some sort of output collector and all we really need is the serializer if we are transforming values 1:1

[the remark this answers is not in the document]
kvargha: I agree. We can do away with `TransformedRecord ` and move `getValueBytes` to `DaVinciRecordTransformer` as a default function.

Since we're no longer collecting output, I'm thinking we should rename `put` to `transformValue`, and maybe do away with the `delete` method. Thoughts?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/DaVinciRecordTransformer.java
```diff
@@ -34,6 +34,18 @@ public interface DaVinciRecordTransformer<K, V, O extends TransformedRecord> {
    */
   O put(Lazy<K> key, Lazy<V> value);
 
+  /**
```
gaojieliu, line 37: I have some questions about these two functions:
```
  Schema getKeyOutputSchema();
  Schema getValueOutputSchema();
```
Will these schemas be the ones used to persist the transformed data in the local storage?
If yes, that might introduce some risks.
Today, DVC will leverage the on-disk snapshot during restart/deployment and if the previous snapshot is using a different schemas because of logic change after a new deployment, the newly deployed DVC app shouldn't leverage the previous snapshot, and even with the same schemas, the transforming logic can be changed, and keeping ingesting on top of the previous snapshot can be problematic as well.

If we want to go with this direction, I think we should version the transformer, which should cover both schemas and transformation logic.
At the end of the day, it will be still a best effort.

A naive way to address this concern is to bootstrap every time, which might not be ideal for some use case, which want to leverage the previous snapshot to speed up bootstrapping.

elijahgrimaldi: Right now we have an additional check on restart whether the schema has changed compared to the on-disk snapshot or if a transformer is present in the new deployment when it previously wasn't and will reingest. The reingestion part is what is meant by bootstrapping correct? Although this doesn't account for change in transformer logic.

elijahgrimaldi: I can add a version to the transformer, that we can state the user should pass a monotonically increasing version number when they update with a new transformer. On restart we can just check this instead.

elijahgrimaldi: Maybe we can leverage the snapshot if there was no transformation before, and apply the new transformation logic locally instead of ingestion over a network?

ZacAttack: Let's take this up in another PR.  I think the idea of a versioning scheme outside of a schema change makes sense, but let's think on how best to approach that.

clients/da-vinci-client/src/test/java/com/linkedin/davinci/config/DaVinciConfigTest.java
```diff
@@ -43,7 +42,7 @@ public void testRecordTransformerEnabled() {
 
   @Test
   public void testGetAndSetRecordTransformer() {
-    DaVinciRecordTransformer transformer = new TestRecordTransformer();
+    DaVinciRecordTransformer transformer = new TestRecordTransformer(0);
```
ZacAttack, line 45: We'll need to call new on this in the IngestionTask in order to supply the version, So we probably need a factory of some kind passed to the davinci client.

ZacAttack (approved):
I think this looks good, great work!

PR #852, opened 2024-02-12
[controller][h2v][common][router] Improve resiliency in ZSTD dictionary retrieval

xunyin8:
## [controller][h2v][common][router] Improve resiliency in ZSTD dictionary retrieval
1. Catch throwable in router dictionary fetching thread so the thread doesn't die when encountering issues fetching dictionary for one store.

2. Replaced static final field for empty push dictionary in ZstdWithDictCompressor to Lazy wrapped final fields in corresponding classes. This way it doesn't get initialized if it's not used. This is the case for routers since it was initialized but never used. Failure to initialize the field will render the entire ZstdWithDictCompressor class unable to initialize.

## How was this PR tested?

### DIFF
--- a/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
@@ -415,6 +415,8 @@ public class VeniceHelixAdmin implements Admin, StoreCleaner {
   private final Object LIVENESS_HEARTBEAT_CLIENT_LOCK = new Object();
   private AvroSpecificStoreClient<BatchJobHeartbeatKey, BatchJobHeartbeatValue> livenessHeartbeatStoreClient = null;
 
+  private final Lazy<ByteBuffer> emptyPushZSTDDictionary;
+
   public VeniceHelixAdmin(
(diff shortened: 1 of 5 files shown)


services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -683,6 +685,8 @@ public void handleDeletedInstances(Set<Instance> deletedInstances) {
         }
       });
     }
+    emptyPushZSTDDictionary =
+        Lazy.of(() -> ByteBuffer.wrap(ZstdWithDictCompressor.buildDictionaryOnSyntheticAvroData()));
```
FelixGV, line 689: Can we also update the `buildDictionaryOnSyntheticAvroData` function so that everything within it is within a try catch, and we could log whatever happens before rethrowing?

FelixGV (commented):
Thanks Xun. Just one minor comment...

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
xunyin8: Good call, will do.

FelixGV (approved):
Thanks Xun!

PR #853, opened 2024-02-13
[controller] Decrement replica count in disabled partition stat metric

majisourav99:
##  Decrement replica count in disabled partition stat metric

There was a mismatch in increment and decrement count in disabled replica metric. On hitting ingestion error replicas are disabled for a partition, but during resetting them its done per host resources which can contain multiple partitions which can lead to mismatch in metric. This PR decrements the metric by the number of partition being reset. Also collect instances from live instance monitor instead of of 'INSTANCE' path which can contain stale resource.
## How was this PR tested?

GHCI


FelixGV (approved):
Thanks Sourav!

PR #855, opened 2024-02-13
[controller] Fixed NPE in StoreInfo::fromStore

FelixGV:
When this function receives a null Store parameter, it will return null itself.

There are four calls to this function, and they have all been inspected to ensure that they deal properly with nulls.

In the case of VHA::migrateStore, there are some usability improvements so that the VeniceNoStoreException includes the cluster the store is in, if it is in a different one than the supplied source cluster. There is a new short-circuiting logic as well in case the store is already in the desired destination cluster.

## How was this PR tested?
Unit tests.

### DIFF
--- a/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
@@ -7123,7 +7123,7 @@ public class VeniceHelixAdmin implements Admin, StoreCleaner {
          * This filtering should not be needed because {@link #getAllStores(String)} should not return any null items,
          * but just in case, we have it here as defensive code...
          */
-        .filter(storeInfo -> storeInfo != null)
+        .filter(Objects::nonNull)
         .map(StoreInfo::fromStore)
         .collect(Collectors.toList());
   }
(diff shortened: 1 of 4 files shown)


services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -7094,6 +7119,11 @@ public void configureActiveActiveReplication(
   public ArrayList<StoreInfo> getClusterStores(String clusterName) {
     // Return all stores at this step in the process
     return (ArrayList<StoreInfo>) getAllStores(clusterName).stream()
+        /**
+         * This filtering should not be needed because {@link #getAllStores(String)} should not return any null items,
+         * but just in case, we have it here as defensive code...
+         */
+        .filter(storeInfo -> storeInfo != null)
```
sushantmane, line 7126: We can also use `filter(Objects::nonNull)`

sushantmane (approved):
LGTM! Thanks!

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
FelixGV: Oh yeah nice, I thought there was something like that but I was blanking out...

sushantmane (approved):

PR #856, opened 2024-02-13
[thin-client][producer][router] SchemaReader will only fetch single schema for all the APIs

sixpluszero:
## [thin-client][producer][router] SchemaReader will only fetch single schema for all the APIs

This PR refactors the RouterBackedSchemaReader implementation to make sure no batch schema fetching under new router endpoint. 

1. It will lazily fetch individual value schema by value schema ID and update schema by value schema ID. 
2. For get value schema id by schema and get latest value schema APIs, we introduced a new router endpoint to /value_schema_ids/schema_name. It will fetch all value schema IDs and then fetch all value schema IDs one by one and derive latest value schema id.
3. For schema refresher thread, it will use the same end point to perform schema refreshing.
4. If new router endpoint is not available, we should fall back to fetch schema by batch.
5. This PR introduces a behavior change: If get value schema / update schema failed router call (throws exception), we will catch it and return null SchemaEntry.

## How was this PR tested?
N/A

### DIFF
--- a/clients/venice-thin-client/src/main/java/com/linkedin/venice/client/schema/RouterBackedSchemaReader.java
+++ b/clients/venice-thin-client/src/main/java/com/linkedin/venice/client/schema/RouterBackedSchemaReader.java
@@ -33,6 +33,7 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Function;
@@ -67,7 +68,7 @@ public class RouterBackedSchemaReader implements SchemaReader {
   private final Map<Integer, DerivedSchemaEntry> valueSchemaIdToUpdateSchemaEntryMap = new VeniceConcurrentHashMap<>();
   private final AtomicReference<SchemaEntry> latestValueSchemaEntry = new AtomicReference<>();
   private final AtomicInteger supersetSchemaIdAtomic = new AtomicInteger(SchemaData.INVALID_VALUE_SCHEMA_ID);
-
+  private final AtomicBoolean shouldRefreshLatestValueSchemaEntry = new AtomicBoolean(false);
   private final String storeName;
   private final InternalAvroStoreClient storeClient;
   private final boolean externalClient;
@@ -413,13 +414,15 @@ public class RouterBackedSchemaReader implements SchemaReader {
 
   private SchemaEntry maybeUpdateLatestValueSchemaEntry() {
     SchemaEntry latest = latestValueSchemaEntry.get();
-    if (latest == null) {
+    if (latest == null || shouldRefreshLatestValueSchemaEntry.get()) {
       /**
-       * Every time it sees latestValueSchemaEntry is null, it will try to update it once.
+       * Every time it sees latestValueSchemaEntry is null or the flag to update latest schema entry is set to true,
+       * it will try to update it once.
        * The update is expensive, but we expect it to be filled after the first fetch, as each store should have at least
        * one active value schema.
        */
       updateAllValueSchemaEntriesAndLatestValueSchemaEntry(false);
+      shouldRefreshLatestValueSchemaEntry.compareAndSet(true, false);
       latestValueSchemaEntry.set(latestValueSchemaEntry.get());
       latest = latestValueSchemaEntry.get();
     }
@@ -453,6 +456,9 @@ public class RouterBackedSchemaReader implements SchemaReader {
         if (entry == null) {
           return NOT_EXIST_VALUE_SCHEMA_ENTRY;
         }
+        // Every time when we fetch a new value schema to cache during non-force-refresh logic, we should try to mark
+        // the flag as true.
+        shouldRefreshLatestValueSchemaEntry.compareAndSet(false, true);
         valueSchemaMapR.put(entry.getSchema(), valueSchemaId);
         return entry;
       });
--- a/clients/venice-thin-client/src/main/java/com/linkedin/venice/client/schema/RouterBackedSchemaReader.java
+++ b/clients/venice-thin-client/src/main/java/com/linkedin/venice/client/schema/RouterBackedSchemaReader.java
@@ -183,7 +183,7 @@ public class RouterBackedSchemaReader implements SchemaReader {
 
   @Override
   public Schema getValueSchema(int id) {
-    SchemaEntry valueSchemaEntry = maybeUpdateValueSchemaEntryMapById(id, false);
+    SchemaEntry valueSchemaEntry = maybeUpdateAndFetchValueSchemaEntryById(id, false);
     if (!isValidSchemaEntry(valueSchemaEntry)) {
       LOGGER.warn("Got null value schema from Venice for store: {} and id: {}", storeName, id);
       return null;
@@ -223,7 +223,7 @@ public class RouterBackedSchemaReader implements SchemaReader {
 
   @Override
   public Schema getUpdateSchema(int valueSchemaId) {
-    DerivedSchemaEntry updateSchemaEntry = maybeUpdateUpdateSchemaEntryMapById(valueSchemaId, false);
+    DerivedSchemaEntry updateSchemaEntry = maybeUpdateAndFetchUpdateSchemaEntryById(valueSchemaId, false);
     if (isValidSchemaEntry(updateSchemaEntry)) {
       return updateSchemaEntry.getSchema();
     }
@@ -238,7 +238,7 @@ public class RouterBackedSchemaReader implements SchemaReader {
       LOGGER.warn("Got null latest value schema from Venice for store: {}.", storeName);
       return null;
     }
-    DerivedSchemaEntry updateSchemaEntry = maybeUpdateUpdateSchemaEntryMapById(latestValueSchema.getId(), false);
+    DerivedSchemaEntry updateSchemaEntry = maybeUpdateAndFetchUpdateSchemaEntryById(latestValueSchema.getId(), false);
     if (isValidSchemaEntry(updateSchemaEntry)) {
       return updateSchemaEntry;
     }
@@ -298,10 +298,9 @@ public class RouterBackedSchemaReader implements SchemaReader {
    */
   private void updateAllValueSchemas(boolean forceRefresh) {
     try {
+      Set<Integer> valueSchemaIdSchema;
       try {
-        for (int id: fetchAllValueSchemaIdsFromRouter()) {
-          maybeUpdateValueSchemaEntryMapById(id, forceRefresh);
-        }
+        valueSchemaIdSchema = fetchAllValueSchemaIdsFromRouter();
       } catch (Exception e) {
         LOGGER.warn(
             "Caught exception when trying to fetch all value schema IDs from router, will fetch all value schema entries instead.");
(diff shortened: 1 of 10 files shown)


clients/venice-thin-client/src/main/java/com/linkedin/venice/client/schema/RouterBackedSchemaReader.java
```diff
@@ -300,6 +290,294 @@ private SchemaEntry fetchSingleSchema(String requestPath, boolean isValueSchema)
...
+   * If the new router endpoint /value_schema_ids/ is not available, it should fall back to query all value schemas at once.
+   */
+  private void updateAllValueSchemas(boolean forceRefresh) {
+    try {
+      Set<Integer> valueSchemaIdSet;
+      try {
+        valueSchemaIdSet = fetchAllValueSchemaIdsFromRouter();
+      } catch (Exception e) {
+        LOGGER.warn(
+            "Caught exception when trying to fetch all value schema IDs from router, will fetch all value schema entries instead.");
```
gaojieliu, line 307: Do we know the difference between non-existing endpoint and some true error even the endpoint exist?
For non-existing endpoint, potentially we can skip the logging, since it is not quite useful IMO.

clients/venice-thin-client/src/main/java/com/linkedin/venice/client/schema/RouterBackedSchemaReader.java
```diff
@@ -300,6 +290,294 @@ private SchemaEntry fetchSingleSchema(String requestPath, boolean isValueSchema)
...
+   * (1) Exists in the cached schema map, we will directly return it.
+   * (2) Does not exist in the cached schema map, we will perform one time single schema fetch to request it from router,
+   * and update the result in the cached schema map.
+   * (3) If the schema was queried previously and not found, we will:
+   * -- (a) Return not found, if forceRefresh flag == false.
+   * -- (b) Perform one time router query, update and return the result if forceRefresh flag == true.
+   */
+  private SchemaEntry maybeUpdateAndFetchValueSchemaEntryById(int valueSchemaId, boolean forceRefresh) {
+    if (forceRefresh) {
+      SchemaEntry entry = fetchValueSchemaEntryFromRouter(valueSchemaId);
```
gaojieliu, line 444: When should we force refresh when the schema id is available locally?

clients/venice-thin-client/src/main/java/com/linkedin/venice/client/schema/RouterBackedSchemaReader.java
```diff
@@ -300,6 +290,294 @@ private SchemaEntry fetchSingleSchema(String requestPath, boolean isValueSchema)
...
+      }
+    } else {
+      return valueSchemaEntryMap.computeIfAbsent(valueSchemaId, (schemaId) -> {
+        SchemaEntry entry = fetchValueSchemaEntryFromRouter(valueSchemaId);
+        if (entry == null) {
+          return NOT_EXIST_VALUE_SCHEMA_ENTRY;
+        }
+        // Every time when we fetch a new value schema to cache during non-force-refresh logic, we should try to mark
+        // the flag as true.
+        shouldRefreshLatestValueSchemaEntry.compareAndSet(false, true);
```
gaojieliu, line 461: I don't fully get this variable, so can you explain more?

clients/venice-thin-client/src/main/java/com/linkedin/venice/client/schema/RouterBackedSchemaReader.java
```diff
@@ -300,6 +290,294 @@ private SchemaEntry fetchSingleSchema(String requestPath, boolean isValueSchema)
...
+
+  private boolean isValidSchemaId(int valueSchemaId) {
+    return valueSchemaId != SchemaData.INVALID_VALUE_SCHEMA_ID;
+  }
+
+  private MultiSchemaIdResponse fetchMultiSchemaIdResponse(String requestPath) throws VeniceClientException {
+    MultiSchemaIdResponse schemaIdResponse;
+    byte[] response = executeRouterRequest(requestPath);
+
+    if (response == null) {
```
gaojieliu, line 563: Can `response` be `null`? Or maybe it can be `null` if the response is `errored`, so shall we log the error message here?

clients/venice-thin-client/src/main/java/com/linkedin/venice/client/schema/RouterBackedSchemaReader.java
```diff
@@ -300,6 +290,294 @@ private SchemaEntry fetchSingleSchema(String requestPath, boolean isValueSchema)
...
+    if (latest == null || shouldRefreshLatestValueSchemaEntry.get()) {
+      /**
+       * Every time it sees latestValueSchemaEntry is null or the flag to update latest schema entry is set to true,
+       * it will try to update it once.
+       * The update is expensive, but we expect it to be filled after the first fetch, as each store should have at least
+       * one active value schema.
+       */
+      updateAllValueSchemaEntriesAndLatestValueSchemaEntry(false);
+      shouldRefreshLatestValueSchemaEntry.compareAndSet(true, false);
+      latestValueSchemaEntry.set(latestValueSchemaEntry.get());
```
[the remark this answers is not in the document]
sixpluszero: Use check -> lock -> check now to fix the issue.

clients/venice-thin-client/src/main/java/com/linkedin/venice/client/schema/RouterBackedSchemaReader.java
sixpluszero: Do you mean if the schema is available in local and it is not INVALID_SCHEMA, then we don't have to call it again?

sixpluszero: Yeah, I think given the fact that schema is static, and will not change its content, I think it makes sense.

clients/venice-thin-client/src/main/java/com/linkedin/venice/client/schema/RouterBackedSchemaReader.java
sixpluszero: When we fetch a new schema from backend, this means the latest value schema id we computed might be already stale, so next time when we are trying to get latest value schema ID, we should perform one time request to fetch it.

clients/venice-thin-client/src/main/java/com/linkedin/venice/client/schema/RouterBackedSchemaReader.java
```diff
@@ -44,19 +48,27 @@
 public class RouterBackedSchemaReader implements SchemaReader {
   public static final String TYPE_KEY_SCHEMA = "key_schema";
   public static final String TYPE_VALUE_SCHEMA = "value_schema";
-  public static final String TYPE_GET_UPDATE_SCHEMA = "update_schema";
+  public static final String TYPE_VALUE_SCHEMA_ID = "value_schema_ids";
+  public static final String TYPE_UPDATE_SCHEMA = "update_schema";
```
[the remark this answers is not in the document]
sixpluszero: I think it will have to change router API.
Also, for these APIs, I think the existing pattern does not contain such `get_xxx` thing, so I'd suggest we just keep it, only change the constant string variable name.

clients/venice-thin-client/src/main/java/com/linkedin/venice/client/schema/RouterBackedSchemaReader.java
sixpluszero: I think it still throws VeniceClientHttpException with a different error message inside executeRouterRequest. I feel like it is hard to distinguish it from other query errors. Given the fact that it should not be called frequently, I suppose it is fine. What do you think?

gaojieliu: The only issue I can see is that the error log for the non-existing endpoint can be a little confusing.
If it is hard to differentiate, we can leave it this way.

clients/venice-thin-client/src/main/java/com/linkedin/venice/client/schema/RouterBackedSchemaReader.java
gaojieliu: Got it.

clients/venice-thin-client/src/main/java/com/linkedin/venice/client/schema/RouterBackedSchemaReader.java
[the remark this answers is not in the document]
gaojieliu: `update_schema` name is a little confusing since it can be used to update an existing value schema.
`get_update_schema` is more clear since it indicates it is a read-only op.
If it is hard to change it, I guess we can live with it since it is internal.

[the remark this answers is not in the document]
sixpluszero: Yeah, I think it is not easy to update the endpoint name easily...maybe let's just leave it as it is for now...

clients/venice-thin-client/src/main/java/com/linkedin/venice/client/schema/RouterBackedSchemaReader.java
```diff
@@ -300,6 +298,309 @@ private SchemaEntry fetchSingleSchema(String requestPath, boolean isValueSchema)
...
+        try {
+          maybeFetchValueSchemaEntryById(id, forceRefresh);
+        } catch (VeniceClientException e) {
+          LOGGER.error("Got exception while trying fetch schema id: " + id + " from router for store: " + storeName);
+        }
+      }
+
+    } catch (Exception ex) {
+      throw new VeniceClientException(
+          "Got exception while trying to fetch all value schemas for store: " + storeName,
```
gaojieliu, line 334: So the code would catch the exception for each individual schema fetch and the outer layer would try to re-throw the exception.
What is the expected behavior here? Is the schema refresh is a best effort or all or nothing?
I saw `maybeFetchValueSchemaEntryById` would update the current schema map directly, so it is not all or nothing , so maybe we can maintain the last exception and re-throw it after the loop?

clients/venice-thin-client/src/main/java/com/linkedin/venice/client/schema/RouterBackedSchemaReader.java
```diff
@@ -300,6 +298,309 @@ private SchemaEntry fetchSingleSchema(String requestPath, boolean isValueSchema)
...
+        valueSchemaIdToUpdateSchemaEntryMap.put(valueSchemaId, derivedSchemaEntry);
+        return derivedSchemaEntry;
+      }
+    } else {
+      return valueSchemaIdToUpdateSchemaEntryMap.computeIfAbsent(valueSchemaId, (id) -> {
+        DerivedSchemaEntry derivedSchemaEntry = fetchUpdateSchemaEntryFromRouter(id);
+        if (derivedSchemaEntry == null) {
+          return NOT_EXIST_UPDATE_SCHEMA_ENTRY;
+        }
+        return derivedSchemaEntry;
```
gaojieliu, line 517: After the change, it seems `forceRefresh` param is useless since it will use the schema if exists and call Router endpoint if absent.
This comment applies to function: `maybeFetchValueSchemaEntryById` as well.

clients/venice-thin-client/src/main/java/com/linkedin/venice/client/schema/RouterBackedSchemaReader.java
sixpluszero: Umm, I think I get what you mean. The outer loop is older code in this PR, I can update this part.

sixpluszero: Oh because fetchAllValueSchemaEntriesFromRouter also need to be handled. But anyway let me try to adjust the handling in this method.

gaojieliu (approved):
Thanks!

PR #857, opened 2024-02-13
[server] apply lossy offset rewind check before data persistence

gaojieliu:
Previously, we discovered that the lossy offset rewind check was applied after data persistence, which made the lossy check useless since the persisted data will be always same as the one just processed. This PR introduced a dry-mode to persisted offset persisting function:
1. Avoid big refactoring.
2. Run the dry-run mode before data persistence to check whether offset rewind is benign or not and run the regular mode after data persistence to persist the processed offset.

This PR also adds an integration test to validate the split brain issue handling logic.

## How was this PR tested?
CI

### DIFF
--- a/internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestHybrid.java
+++ b/internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestHybrid.java
@@ -2,6 +2,7 @@ package com.linkedin.venice.endToEnd;
 
 import static com.linkedin.davinci.store.rocksdb.RocksDBServerConfig.ROCKSDB_PLAIN_TABLE_FORMAT_ENABLED;
 import static com.linkedin.venice.ConfigKeys.DEFAULT_MAX_NUMBER_OF_PARTITIONS;
+import static com.linkedin.venice.ConfigKeys.INSTANCE_ID;
 import static com.linkedin.venice.ConfigKeys.KAFKA_BOOTSTRAP_SERVERS;
 import static com.linkedin.venice.ConfigKeys.PERSISTENCE_TYPE;
 import static com.linkedin.venice.ConfigKeys.SERVER_CONSUMER_POOL_SIZE_PER_KAFKA_CLUSTER;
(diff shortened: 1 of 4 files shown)


internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestHybrid.java
```diff
@@ -271,6 +274,165 @@ public void testHybridInitializationOnMultiColo() throws IOException {
...
+        veniceWriter2.flush();
+        veniceWriter1.broadcastEndOfPush(Collections.emptyMap());
+        veniceWriter1.flush();
+      }
+
+      TestUtils.waitForNonDeterministicAssertion(100, TimeUnit.SECONDS, true, () -> {
+        // Now the store should have version 1
+        JobStatusQueryResponse jobStatus = controllerClient.queryJobStatus(Version.composeKafkaTopic(storeName, 1));
+        Assert.assertFalse(jobStatus.isError(), "Error in getting JobStatusResponse: " + jobStatus.getError());
+        assertEquals(jobStatus.getStatus(), "ERROR");
```
shuhui-liu, line 431: I'm a bit confused with the test assertion. Does the push job succeed with the split brain problem? If so, why job status is ERROR? If not, why `jobStatus.isError()` is false?

gaojieliu: The push job should fail because of the split-brain issue before reporting completed.
`jobStatus.isError` is to check whether Controller is serving the metadata request or not.
Even the job status is `ERROR`, `JobStatusQueryResponse` itself shouldn't be error.

shuhui-liu: Gotcha, thanks!

shuhui-liu (approved):
LGTM

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestHybrid.java
shuhui-liu: Gotcha, thanks!

PR #858, opened 2024-02-14
[router] Fixed NPE in ReadRequestThrottler

FelixGV:
There could be a race condition where a listener of the ReadRequestThrottler got called while the constructor was not fully done executing, leading to a NPE for accessing a field that was not yet initialized. The fix is to register listeners at the very end.

Also eliminated the RoutingDataRepository constructor param and class property since it was unused.

## How was this PR tested?
New unit test.

### DIFF
--- a/services/venice-router/src/main/java/com/linkedin/venice/router/throttle/ReadRequestThrottler.java
+++ b/services/venice-router/src/main/java/com/linkedin/venice/router/throttle/ReadRequestThrottler.java
@@ -8,7 +8,6 @@ import com.linkedin.venice.helix.ZkRoutersClusterManager;
 import com.linkedin.venice.meta.ReadOnlyStoreRepository;
 import com.linkedin.venice.meta.RoutersClusterConfig;
 import com.linkedin.venice.meta.RoutersClusterManager;
-import com.linkedin.venice.meta.RoutingDataRepository;
 import com.linkedin.venice.meta.Store;
 import com.linkedin.venice.meta.StoreDataChangedListener;
 import com.linkedin.venice.router.VeniceRouterConfig;
@@ -39,7 +38,6 @@ public class ReadRequestThrottler implements RouterThrottler, StoreDataChangedLi
   private static final Logger LOGGER = LogManager.getLogger(ReadRequestThrottler.class);
   private final ZkRoutersClusterManager zkRoutersManager;
   private final ReadOnlyStoreRepository storeRepository;
-  private final RoutingDataRepository routingDataRepository;
   private final long maxRouterReadCapacity;
   private int lastRouterCount;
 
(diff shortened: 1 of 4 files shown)


services/venice-router/src/test/java/com/linkedin/venice/router/throttle/ReadRequestThrottlerTest.java
```diff
@@ -368,4 +378,87 @@ public void testDisableMaxCapacityProtection() {
...
+      }
+    };
+
+    ReadRequestThrottler readRequestThrottler = new ReadRequestThrottler(
+        mock(ZkRoutersClusterManager.class),
+        storeRepository,
+        mock(AggRouterHttpRequestStats.class),
+        mock(VeniceRouterConfig.class));
+
+    assertFalse(listenerThrewException.get(), "The listener should not receive an exception!");
```
sushantmane, line 460: Value of `listenerThrewException` is never updated. So this assertion will always succeed.

services/venice-router/src/main/java/com/linkedin/venice/router/throttle/ReadRequestThrottler.java
```diff
@@ -84,23 +80,25 @@ public ReadRequestThrottler(
...
-    this.storeRepository.registerStoreDataChangedListener(this);
     this.stats = stats;
     this.maxRouterReadCapacity = maxRouterReadCapacity;
     this.lastRouterCount = zkRoutersManager.getExpectedRoutersCount();
     this.perStoreRouterQuotaBuffer = perStoreRouterQuotaBuffer;
     this.idealTotalQuotaPerRouter = calculateIdealTotalQuotaPerRouter();
-    this.storesThrottlers = new AtomicReference<>(buildAllStoreReadThrottlers());
     this.isNoopThrottlerEnabled = false;
+
+    /** Calling {@link #buildAllStoreReadThrottlers()} should be done after all internal state is initialized */
```
majisourav99, line 97: Looks like there was another bug in the initialization in the constructor of this class. https://github.com/linkedin/venice/pull/399/files

Thanks for catching another!

majisourav99 (approved):
Thanks Felix!

services/venice-router/src/test/java/com/linkedin/venice/router/throttle/ReadRequestThrottlerTest.java
FelixGV: Ah, good catch 😅 ... I meant to update it in the catch of the registration function...

services/venice-router/src/main/java/com/linkedin/venice/router/throttle/ReadRequestThrottler.java
FelixGV: Yep, this class is the gift that keeps on giving 🤣

services/venice-router/src/test/java/com/linkedin/venice/router/throttle/ReadRequestThrottlerTest.java
FelixGV: Fixed in the latest commit.

sushantmane (approved):
LGTM! Thanks @FelixGV!

PR #860, opened 2024-02-15
[server] Remove a global disk quota metric

adamxchen:
## Summary

I noticed we have a `global disk quota` metric which should always be flat when there's no quota change. We have had `disk_usage_in_bytes` to tell us the real-time disk usage and `storage_quota_used` which computes the percentage. The `global disk quota` is useless IMO and we should get rid of it as it's also a store-level metric so can be expensive. 

## How was this PR tested?

GHCI


ZacAttack (approved):
Looks good.  I double checked customer dashboards, we already emit the quota usage as a function of percent.  So I too don't think is particularly useful.

adamxchen:
@ZacAttack  can you please approve it again?

ZacAttack (approved):

PR #861, opened 2024-02-16
[controller] Increase future timeout while fetching metadata

majisourav99:
## Increase future timeout while fetching metadata in controller

Currently the mtadata fetch request from server times out sporadically due to low future.get timeout of 500ms. This PR increase that resolve those timeouts. 

## How was this PR tested?

Canaried in controller host.

### DIFF
--- a/services/venice-controller/src/main/java/com/linkedin/venice/controller/StoreBackupVersionCleanupService.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/controller/StoreBackupVersionCleanupService.java
@@ -105,7 +105,7 @@ public class StoreBackupVersionCleanupService extends AbstractVeniceService {
     });
     Optional<SSLFactory> sslFactory = admin.getSslFactory();
     this.httpAsyncClient = HttpAsyncClients.custom()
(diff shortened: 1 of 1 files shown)


services/venice-controller/src/main/java/com/linkedin/venice/controller/StoreBackupVersionCleanupService.java
```diff
@@ -105,7 +105,7 @@ protected StoreBackupVersionCleanupService(
     });
     Optional<SSLFactory> sslFactory = admin.getSslFactory();
     this.httpAsyncClient = HttpAsyncClients.custom()
-        .setDefaultRequestConfig(RequestConfig.custom().setSocketTimeout(2000).build())
+        .setDefaultRequestConfig(RequestConfig.custom().setSocketTimeout(10000).build())
```
sixpluszero, line 108: Hmm, are you trying to set it to 10s? It seems very long timeout to me, could you give any context about this change? Like why it would take up to 10s..

sixpluszero (approved):
Sync offline, I think we can give it a try, if that's not going to work, I feel like server side should have some issue responding to requests.

PR #862, opened 2024-02-16
[controller] Parent controller should not update the current version

nisargthakkar:
## Parent controller should not update the current version

Currently, there are some codepaths that lead to parent controller updating the current version. Another unexpected code path that gets triggered in parent controller is `RealTimeTopicSwitcher#transmitVersionSwapMessage`. Since this piece wasn't written with the intention of getting executed in parent controller, it causes failures in executing tasks from the admin channel. This causes the Admin channel to get stuck for this store.

This commit adds checks that prevent the current version in parent controller to be updated.

## How was this PR tested?

CI

### DIFF
--- a/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
@@ -3719,22 +3719,24 @@ public class VeniceHelixAdmin implements Admin, StoreCleaner {
   @Override
   public void setStoreCurrentVersion(String clusterName, String storeName, int versionNumber) {
     storeMetadataUpdate(clusterName, storeName, store -> {
-      // Parent colo should not update the current version of a store
-      if (!isParent()) {
-        if (store.getCurrentVersion() != Store.NON_EXISTING_VERSION) {
-          if (versionNumber != Store.NON_EXISTING_VERSION && !store.containsVersion(versionNumber)) {
-            throw new VeniceException("Version:" + versionNumber + " does not exist for store:" + storeName);
-          }
-
-          if (!store.isEnableWrites()) {
-            throw new VeniceException(
-                "Unable to update store:" + storeName + " current version since store writeability is false");
-          }
-        }
-        int previousVersion = store.getCurrentVersion();
-        store.setCurrentVersion(versionNumber);
-        realTimeTopicSwitcher.transmitVersionSwapMessage(store, previousVersion, versionNumber);
+      if (isParent()) {
+        // Parent colo should not update the current version of a store
+        return store;
       }
+
+      if (store.getCurrentVersion() != Store.NON_EXISTING_VERSION) {
+        if (versionNumber != Store.NON_EXISTING_VERSION && !store.containsVersion(versionNumber)) {
+          throw new VeniceException("Version: " + versionNumber + " does not exist for store:" + storeName);
+        }
+
+        if (!store.isEnableWrites()) {
+          throw new VeniceException(
+              "Unable to update store: " + storeName + " current version since store writeability is false");
+        }
+      }
+      int previousVersion = store.getCurrentVersion();
+      store.setCurrentVersion(versionNumber);
+      realTimeTopicSwitcher.transmitVersionSwapMessage(store, previousVersion, versionNumber);
       return store;
     });
   }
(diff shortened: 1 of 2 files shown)


services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -3720,19 +3720,22 @@ Pair<Store, Version> waitVersion(String clusterName, String storeName, int versi
...
+          if (versionNumber != Store.NON_EXISTING_VERSION && !store.containsVersion(versionNumber)) {
+            throw new VeniceException("Version:" + versionNumber + " does not exist for store:" + storeName);
+          }
 
-        if (!store.isEnableWrites()) {
-          throw new VeniceException(
-              "Unable to update store:" + storeName + " current version since store writeability is false");
+          if (!store.isEnableWrites()) {
+            throw new VeniceException(
+                "Unable to update store:" + storeName + " current version since store writeability is false");
```
ZacAttack, line 3732: nit: (that I think 'I' originally did, but I'll make you fix it XD) add a space after store: and Version: in the two exception messages.

majisourav99 (commented):
While you are at it can you fix the code in `transmitVersionSwapMessage` as well ?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -3720,19 +3720,22 @@ Pair<Store, Version> waitVersion(String clusterName, String storeName, int versi
   @Override
   public void setStoreCurrentVersion(String clusterName, String storeName, int versionNumber) {
     storeMetadataUpdate(clusterName, storeName, store -> {
-      if (store.getCurrentVersion() != Store.NON_EXISTING_VERSION) {
-        if (versionNumber != Store.NON_EXISTING_VERSION && !store.containsVersion(versionNumber)) {
-          throw new VeniceException("Version:" + versionNumber + " does not exist for store:" + storeName);
-        }
+      // Parent colo should not update the current version of a store
+      if (!isParent()) {
```
sushantmane, line 3724: Can we do this check outside and avoid extra indentation...
```java
if (isParent()) {
  return;
}
```

nisargthakkar: Sure

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -3720,14 +3720,19 @@ Pair<Store, Version> waitVersion(String clusterName, String storeName, int versi
   @Override
```
nisargthakkar, line 3720: @smaji comment in a random place to start a thread:

> While you are at it can you fix the code in `transmitVersionSwapMessage` as well ?
This function will not get triggered in parent controller anymore. Do we still need a fix there?

majisourav99: It would be nice to block it in parent from accidental usage.

gaojieliu:
Can you explain why non-zero parent current-version can affect other admin operations? Maybe we should enhance that piece of logic as well?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
nisargthakkar: I don't know about that. I'd rather not put band-aids to functions that weren't written to be executed in the parent controller. As I see it, entering `transmitVersionSwap` function in parent controller is an issue in itself. So, throwing an exception is the right thing to do

I think we need to refactor our `VeniceHelixAdmin` and `VeniceParentHelixAdmin` logics to be better structured with better protection. There are a lot of chances of unintentionally doing the wrong thing since `VeniceHelixAdmin` runs in both child and parent controllers. We generally forget to handle "Parent controller" logic in this `VeniceHelixAdmin`

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -3720,14 +3720,19 @@ Pair<Store, Version> waitVersion(String clusterName, String storeName, int versi
...
   public void setStoreCurrentVersion(String clusterName, String storeName, int versionNumber) {
```
nisargthakkar, line 3721: >
 Can you explain why non-zero parent current-version can affect other admin operations? Maybe we should enhance that piece of logic as well?

This is the code in `VeniceHelixAdmin#setStoreCurrentVersion`:
```
      int previousVersion = store.getCurrentVersion();
      store.setCurrentVersion(versionNumber);
      realTimeTopicSwitcher.transmitVersionSwapMessage(store, previousVersion, versionNumber);
```

Here, if current version in parent gets updated, and then the version cycles out of the parent Zk (through enough VPJ afterwards), then it causes exceptions in `RealTimeTopicSwitcher#transmitVersionSwap`:
```
    Version previousStoreVersion = store.getVersion(previousVersion)
        .orElseThrow(
            () -> new VeniceException(
                "Corresponding version " + previousVersion + " does not exist for store: " + store.getName()));
    Version nextStoreVersion = store.getVersion(nextVersion)
        .orElseThrow(
            () -> new VeniceException(
                "Corresponding version " + nextVersion + " does not exist for store: " + store.getName()));
```

This `RealTimeTopicSwitcher#transmitVersionSwap` is not supposed to execute in the parent controller. The short-circuit added in `VeniceHelixAdmin#setStoreCurrentVersion` will prevent this function from executing.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
majisourav99: Yeah, I see `rollForwardToFutureVersion` and backup version are also unprotected in `VeniceHelixAdmin` from parent controller running it. throwing exception also block the admin channel, no ?

nisargthakkar: Yeah. Throwing exception will block the channel. I think it is expected to have these two functions in `VeniceHelixAdmin` as currently, we support a single colo mode of deployment (most tests use this)

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
gaojieliu: I see, thanks for the explanation!

majisourav99 (approved):
LGTM. Thanks!

PR #866, opened 2024-02-20
[server] Skip lag monitor setup if the version info is not found in the store metadata

sushantmane:
## Skip lag monitor setup if the version info is not found in the store metadata
Previously, if version info for a store wasn't available in the metadata repo,
the HB lag monitor would throw an exception, causing state transitions to fail.
This change adds a configurable wait for version info to become available. If it
doesn't, we skip the lag monitor setup and log the failure.

New config:
`server.max.wait.for.version.info.ms (default: 5000 milliseconds)`
A config to control the maximum duration to wait for the version  
information to become available in the store metadata repository  
before skipping Heartbeat (HB) lag monitor setup activity during  
state transition.

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/LeaderFollowerPartitionStateModel.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/LeaderFollowerPartitionStateModel.java
@@ -14,9 +14,12 @@ import com.linkedin.venice.meta.ReadOnlyStoreRepository;
 import com.linkedin.venice.meta.Store;
 import com.linkedin.venice.meta.Version;
 import com.linkedin.venice.utils.LatencyUtils;
+import com.linkedin.venice.utils.Pair;
+import java.time.Duration;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
+import java.util.function.BiConsumer;
 import org.apache.helix.NotificationContext;
 import org.apache.helix.model.Message;
 import org.apache.helix.participant.statemachine.StateModelInfo;
@@ -104,7 +107,7 @@ public class LeaderFollowerPartitionStateModel extends AbstractPartitionStateMod
       try {
         long startTimeForSettingUpNewStorePartitionInNs = System.nanoTime();
         setupNewStorePartition();
-        store.getVersion(version).ifPresent(v -> heartbeatMonitoringService.addFollowerLagMonitor(v, getPartition()));
+        updateLagMonitor(message.getResourceName(), heartbeatMonitoringService::addFollowerLagMonitor);
         logger.info(
             "Completed setting up new store partition for {} partition {}. Total elapsed time: {} ms",
             resourceName,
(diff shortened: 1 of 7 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/LeaderFollowerPartitionStateModel.java
```diff
@@ -104,7 +104,7 @@ public void onBecomeStandbyFromOffline(Message message, NotificationContext cont
       try {
         long startTimeForSettingUpNewStorePartitionInNs = System.nanoTime();
         setupNewStorePartition();
-        heartbeatMonitoringService.addFollowerLagMonitor(store.getVersion(version).get(), getPartition());
+        store.getVersion(version).ifPresent(v -> heartbeatMonitoringService.addFollowerLagMonitor(v, getPartition()));
```
ZacAttack, line 107: We need this to fail more loudly, or, in the absence of this still set up a lag monitor.

Our goal is to set all alerting on this metric.  If we randomly aren't setting up the alert, then it won't be very useful.

sushantmane: I agree. For now I can add a log statement but we should "understand and fix why version metadata is missing."

clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/LeaderFollowerPartitionStateModel.java
```diff
@@ -201,6 +192,27 @@ public void onBecomeOfflineFromError(Message message, NotificationContext contex
     logger.warn("unexpected state transition from ERROR to OFFLINE");
   }
 
+  private void updateLagMonitor(String resourceName, BiConsumer<Version, Integer> lagMonFunction) {
+    try {
+      String storeName = Version.parseStoreFromKafkaTopicName(resourceName);
+      int storeVersion = Version.parseVersionFromKafkaTopicName(resourceName);
+      Pair<Store, Version> res = getStoreRepo().waitVersion(storeName, storeVersion, Duration.ofMillis(500));
```
nisargthakkar, line 199: 500 millis is not going to be enough. The waitVersion retries this call every 1 second. Any concern in making this timeout to 30 seconds - and probably configurable through some config?

sushantmane: Good catch. Let me bump it to higher value.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/LeaderFollowerPartitionStateModel.java
```diff
@@ -201,6 +192,27 @@ public void onBecomeOfflineFromError(Message message, NotificationContext contex
...
+      Store store = res.getFirst();
+      Version version = res.getSecond();
+      if (store == null || version == null) {
+        logger.warn(
```
ZacAttack, line 203: Let's bump this to error, this is pretty bad if we get here since we won't get metrics.

sushantmane: done

clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/LeaderFollowerPartitionStateModel.java
sushantmane: done

clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/LeaderFollowerPartitionStateModel.java
sushantmane: Added configurable wait and log statement

ZacAttack (dismissed):
Looks good!

nisargthakkar:
Can you update the commit message before merging?

sushantmane:
> Can you update the commit message before merging?

I've already updated it. Should I make any specific modifications?

sushantmane:
Thanks @nisargthakkar & @ZacAttack for the review!

ZacAttack (approved):

PR #867, opened 2024-02-20
[producer] Skip metrics for the SchemaReaders inside online producer

nisargthakkar:
## Skip metrics for the `SchemaReader` inside online producer

Currently, the clients internal to online producer also registers metrics. While this is not an issue by itself, it could cause issues if online producer and Venice client are used in the same application. This PR disables the metric registration since these metrics are not very useful currently.

## How was this PR tested?

GH CI

    - Store clients created internal to Online producer will not register and emit any metrics


sushantmane (approved):

sushantmane (approved):
LGTM! Thanks @nisargthakkar!

PR #868, opened 2024-02-21
[Router] Add router count change listner in read quota throttler

majisourav99:
## Add router count change listner in read quota throttler

Commit https://github.com/linkedin/venice/pull/548/files#diff-a543d0bd2b70372ac582bc358f240ee3507b8c3bf14f9e7c8829fb2e9c0de362 removed the live router count change listener in 'ReadRequestThrottler'  which made router quota susceptible to mismatched quota issues during router instances modifications.
This PR fixes that and add tests to capture this issue.

## How was this PR tested?

integ test.

### DIFF
--- a/services/venice-router/src/main/java/com/linkedin/venice/router/throttle/ReadRequestThrottler.java
+++ b/services/venice-router/src/main/java/com/linkedin/venice/router/throttle/ReadRequestThrottler.java
@@ -89,6 +89,7 @@ public class ReadRequestThrottler implements RouterThrottler, StoreDataChangedLi
     this.storeQuotaCheckTimeWindow = storeQuotaCheckTimeWindow;
     this.stats = stats;
     this.maxRouterReadCapacity = maxRouterReadCapacity;
+    this.zkRoutersManager.subscribeRouterCountChangedEvent(this);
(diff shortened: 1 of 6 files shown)


services/venice-router/src/main/java/com/linkedin/venice/router/throttle/ReadRequestThrottler.java
```diff
@@ -89,6 +89,7 @@ public ReadRequestThrottler(
     this.storeQuotaCheckTimeWindow = storeQuotaCheckTimeWindow;
     this.stats = stats;
     this.maxRouterReadCapacity = maxRouterReadCapacity;
+    this.zkRoutersManager.subscribeRouterCountChangedEvent(this);
```
gaojieliu, line 92: Can we get rid of `RoutersClusterConfig` and the related code as well?
This config and related listeners are not being used in prod.

gaojieliu (approved):
Thanks for the fix and cleanup.

services/venice-router/src/main/java/com/linkedin/venice/router/throttle/ReadRequestThrottler.java
majisourav99: Remove the reference to `RoutersClusterConfig`

PR #869, opened 2024-02-21
[server] Metric Name fix and math fix for HeartbeatMonitor

ZacAttack:
## [server] Metric Name fix and math fix for HeartbeatMonitor

The resulting metric name contained an extra set of '--' when being hooked up to the reporter.  Additionally, the metric was reporting the time of the last heartbeat, but not the time SINCE the last heartbeat.

Resolves #XXX

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/ingestion/heartbeat/HeartbeatMonitoringService.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/ingestion/heartbeat/HeartbeatMonitoringService.java
@@ -220,14 +220,15 @@ public class HeartbeatMonitoringService extends AbstractVeniceService {
   }
 
   protected void record() {
+    long currentTime = System.currentTimeMillis();
     recordLags(
         leaderHeartbeatTimeStamps,
         ((storeName, version, region, lag) -> versionStatsReporter
-            .recordLeaderLag(storeName, version, region, System.currentTimeMillis() - lag)));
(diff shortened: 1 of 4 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/ingestion/heartbeat/HeartbeatMonitoringService.java
```diff
@@ -222,10 +222,12 @@ protected void recordLags(
...
     recordLags(
         leaderHeartbeatTimeStamps,
-        ((storeName, version, region, lag) -> versionStatsReporter.recordLeaderLag(storeName, version, region, lag)));
+        ((storeName, version, region, lag) -> versionStatsReporter
+            .recordLeaderLag(storeName, version, region, System.currentTimeMillis() - lag)));
     recordLags(
         followerHeartbeatTimeStamps,
-        ((storeName, version, region, lag) -> versionStatsReporter.recordFollowerLag(storeName, version, region, lag)));
+        ((storeName, version, region, lag) -> versionStatsReporter
+            .recordFollowerLag(storeName, version, region, System.currentTimeMillis() - lag)));
```
sushantmane, line 230: Can we just pass start time here and then do subtraction (to get elapsed time) in `HeartbeatStat::recordLeaderLag` & HeartbeatStat::recordFollowerLag`.  This way we don't need to invoke `System.currentTimeMillis()` twice for each call..?

sushantmane (approved):
LGTM! Thanks @ZacAttack!

PR #870, opened 2024-02-22
[vpj] Bump Apache Spark version to 3.3.3

nisargthakkar:
## Bump Apache Spark version to 3.3.3

Apache Spark has a critical vulnerability in versions prior to 3.3.3. ([CVE-2023-22946](https://github.com/advisories/GHSA-329j-jfvr-rhr6)). This PR bumps Spark version to 3.3.3 which has a fix for this vulnerability.

## How was this PR tested?

GH CI. Tested manually on test flows

### DIFF
--- a/clients/venice-push-job/build.gradle
+++ b/clients/venice-push-job/build.gradle
@@ -55,6 +55,11 @@ dependencies {
     exclude group: 'org.apache.hadoop', module: 'hadoop-client-api'
   }
 
+  // Spark versions 3.2.X - 3.3.X are compiled with antlr4 4.8. In our classpath, antlr4 version 4.5 is used. This
+  // discrepancy causes errors at runtime.
+  implementation libraries.antlr4
+  implementation libraries.antlr4Runtime
+
   implementation project(':clients:venice-thin-client') // Needed by the KME SchemaReader
 
   implementation libraries.commonsIo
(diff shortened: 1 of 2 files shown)


clients/venice-push-job/build.gradle
```diff
@@ -55,6 +55,11 @@ dependencies {
     exclude group: 'org.apache.hadoop', module: 'hadoop-client-api'
   }
 
+  // Spark versions 3.2.X - 3.3.X are compiled with antlr4 4.8. In our classpath, antlr4 version 4.5 is used. This
+  // discrepancy causes errors at runtime.
```
FelixGV, line 59: Hmm, that is weird... if we don't have an explicit dependency on antlr, then why isn't Gradle resolving the transitive deps to the highest one requested automatically? Anyway, it's not a blocker to merge, but I'm just curious...

FelixGV (approved):

clients/venice-push-job/build.gradle
nisargthakkar: Yeah, it was surprising to me too. I saw that the version for `antlr4-runtime` was updated ot 4.8 transitively. However, `pegasus` (r2/d2) also pulls in `antlr4` module (version 4.5). We can also explore excluding this specific package if we find that it has no business at runtime (like `errorprone`/`checkerframework`/etc)

FelixGV: I'm all for dep exclusion... we carry way too much baggage. Pegasus in particular is pretty heavyweight. Oh well.

PR #873, opened 2024-02-23
[venice-server][metrics] Remove period from metric name

ZacAttack:
## [venice-server][metrics] Remove period from metric name

This seems to be wreaking some havoc with how Tehuti namespaces metric names.  So replacing it with a '-'.

Resolves #XXX


sushantmane (changes requested):
Can you pass metrics stats name through `TehutiUtils.fixMalformedMetricName` just to be safe?

sushantmane (approved):

PR #874, opened 2024-02-23
[all] Improved Avro serialization reliability

FelixGV:
When Avro serialization fails for any reason, there is a ThreadLocal BinaryEncoder which can become corrupted. This showed up in unit tests where one test corrupted it and another then produced invalid payloads, but it could also (at least theoretically) happen in production code.

The solution is to catch all exceptions, rather than just IOException, and to make the reusable encoder null, so that another one gets built the next time we serialize. It seems there is no way to properly reset an encoder that has been corrupted in this way, so we abandon reuse in those cases.

In addition to the above fix, the improved union error handling logic has been moved from the online producer to the client-common lib. This requires doing a bit of trickery because the UnresolvedUnionException only exists in modern Avro versions, and so we have to isolate its use into a separate class and make sure we don't load that class if it is not available on the classpath.

## How was this PR tested?
Lots of new unit tests...

Better error messages for union-related issues in all serialization code paths, not just the online producer.


adamxchen (approved):
LGTM! Thanks for the clean up.

PR #875, opened 2024-02-24
[compat][server][fast-client] Use backend store batch-get limit config

xunyin8:
## [compat][server][fast-client] Use backend store batch-get limit config
1. Removed fast-client side config maxAllowedKeyCntInBatchGetReq. Use the store config value set for batchGetLimit instead to avoid confusion and reduce toil. Default value set to 150.

2. Fast-client will still fail fast for requests that exceeded the limit. However, the exception will now be thrown when the caller calls get() on the future instead of during request construction for better error handling/logging for use cases where venice client is used through another intermediate plateform or library.

3. Evolved MetadataResponseRecord protocol to v2 to include batchGetLimit field. All existing running fast-client uses request based metadata and are on a version that supports forward compatibility so the change should be safe.

4. Split MetadataRetriever into IngestionMetadataRetriever and ReadMetadataRetriever to move read metadata handling out of the already crowded KafkaStoreIngestionService. 

## How was this PR tested?
New and existing unit + integration tests.

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
@@ -241,8 +241,6 @@ public class DaVinciBackend implements Closeable {
           clusterInfoProvider,
           storeRepository,
           schemaRepository,
-          Optional.empty(),
-          Optional.empty(),
           null,
           metricsRepository,
           Optional.of(kafkaMessageEnvelopeSchemaReader),
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaStoreIngestionService.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaStoreIngestionService.java
@@ -20,8 +20,6 @@ import com.linkedin.davinci.config.VeniceServerConfig;
 import com.linkedin.davinci.config.VeniceStoreVersionConfig;
 import com.linkedin.davinci.helix.LeaderFollowerPartitionStateModel;
 import com.linkedin.davinci.listener.response.AdminResponse;
-import com.linkedin.davinci.listener.response.MetadataResponse;
-import com.linkedin.davinci.listener.response.ServerCurrentVersionResponse;
 import com.linkedin.davinci.listener.response.TopicPartitionIngestionContextResponse;
 import com.linkedin.davinci.notifier.LogNotifier;
 import com.linkedin.davinci.notifier.PartitionPushStatusNotifier;
@@ -41,14 +39,10 @@ import com.linkedin.venice.client.store.ClientConfig;
 import com.linkedin.venice.common.VeniceSystemStoreType;
 import com.linkedin.venice.exceptions.VeniceException;
 import com.linkedin.venice.exceptions.VeniceNoStoreException;
-import com.linkedin.venice.helix.HelixCustomizedViewOfflinePushRepository;
-import com.linkedin.venice.helix.HelixInstanceConfigRepository;
 import com.linkedin.venice.helix.HelixReadOnlyZKSharedSchemaRepository;
 import com.linkedin.venice.kafka.protocol.KafkaMessageEnvelope;
 import com.linkedin.venice.kafka.protocol.state.PartitionState;
 import com.linkedin.venice.meta.ClusterInfoProvider;
-import com.linkedin.venice.meta.Instance;
-import com.linkedin.venice.meta.Partition;
 import com.linkedin.venice.meta.ReadOnlyLiveClusterConfigRepository;
 import com.linkedin.venice.meta.ReadOnlySchemaRepository;
 import com.linkedin.venice.meta.ReadOnlyStoreRepository;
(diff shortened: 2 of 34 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaStoreIngestionService.java
```diff
@@ -1325,6 +1325,11 @@ public MetadataResponse getMetadata(String storeName) {
       response.setLatestSuperSetValueSchemaId(latestSuperSetValueSchemaId);
       response.setRoutingInfo(routingInfo);
       response.setHelixGroupInfo(helixGroupInfo);
+      if (store.getBatchGetLimit() > 0) {
```
FelixGV, line 1328: IDK the background here so maybe there is a good reason... but why is this function part of the `KafkaStoreIngestionService`? It does not seem to make sense to have it here 🤔 ...

FelixGV: I guess it's part of the `MetadataRetriever` interface, so that's why... Some of the functions in that interface (`getConsumptionSnapshots` and `getTopicPartitionIngestionContext`) are related to ingestion, so we probably need to keep that interface in any case... while the other 3 functions (`getStoreVersionCompressionDictionary`, `getMetadata` and `getCurrentVersionResponse`) are accessing objects which are passed into the `KSIS` constructor by the `VeniceServer` (or other code paths...) so it might make sense to pass those objects into a different container and have the remaining three functions be decoupled from `KSIS`, but IDK if it's worth it... Looks like the `schemaRepo`, `customizedViewRepository` and `helixInstanceConfigRepository` are only used by `getMetadata` so we could remove those completely from `KSIS` if we decoupled... It seems a tiny bit cleaner, but IDK if it's worth the effort. WDYT?

xunyin8: Yeah I took a look and it's not super clean since we still need `KSIS` to implement `MetadataRetriever` to provide the two consumption related metadata as you mentioned. I don't mind doing some cleanup if we are ok with another service class that solely acts as a container to share some reference for the repositories to provide metadata lookup.

FelixGV: Well... I guess it'd be nice to move the latter 3 functions to a different class, since `KSIS` is already fairly big, at 1400 LoC, so it's gravitating towards God class territory. So splitting the `MetadataRetriever` interface into two, e.g. perhaps `IngestionMetadataRetriever` and `ReadMetadataRetriever` or something like that, and being able to remove some of the objects passed into the `KSIS` constructor (which is also quite large at this point) might help tidy things up a bit. Personally, I think it's ok to spend a bit more time to leave the codebase in a cleaner state than we found it, but I leave it up to you to make the call.

xunyin8: I agree, will make the refactor and changes.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
```diff
@@ -241,8 +241,6 @@ public DaVinciBackend(
           clusterInfoProvider,
           storeRepository,
           schemaRepository,
-          Optional.empty(),
-          Optional.empty(),
```
FelixGV, line 245: Nice! I didn't expect the cleanup to reach all the way here, but that's good! It means the code is better decoupled...

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaStoreIngestionService.java
```diff
@@ -217,8 +200,6 @@ public KafkaStoreIngestionService(
       ClusterInfoProvider clusterInfoProvider,
       ReadOnlyStoreRepository metadataRepo,
       ReadOnlySchemaRepository schemaRepo,
-      Optional<CompletableFuture<HelixCustomizedViewOfflinePushRepository>> customizedViewFuture,
-      Optional<CompletableFuture<HelixInstanceConfigRepository>> helixInstanceFuture,
```
FelixGV, line 221: Nitpick: I think the `sslFactory` is also unused. Can you check and delete that, if applicable?

FelixGV (commented):
Thanks! That looks great! Just one minor nitpick...

Also, please mention this refactoring in the commit message when you squash it!

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaStoreIngestionService.java
xunyin8: Done, unfortunately it's still used for KME protocol so can only be converted from class field to local variable to be passed along.

FelixGV (approved):
Great work! Thanks!!

PR #876, opened 2024-02-26
Update bootstrappingVeniceChangelogConsumer with sync offset and allow multi-consumers

zx0401:
[changelog][BootstrappingVeniceChangelogConsumer] Update bootstrappingVeniceChangelogConsumer to sync offset during processing and allow multi-consumers with different consumer id.

- Update bootstrappingVeniceChangelogConsumer to sync offset during processing when processedRecordSizeSinceLastSync > databaseSyncBytesInterval.
- Add consumerId in the bootstrappingVeniceChangelogConsumer constructor so we can allow multi-consumers creation. Each consumer will have its own state, rocksDB path, etc.

## How was this PR tested?

- unit test
- integration test

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/ChangelogClientConfig.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/ChangelogClientConfig.java
@@ -19,8 +19,10 @@ public class ChangelogClientConfig<T extends SpecificRecord> {
   private int controllerRequestRetryCount;
 
   private String bootstrapFileSystemPath;
-
   private long versionSwapDetectionIntervalTimeInMs = 600000L;
+
+  // Default is 32 MB
+  private long databaseSyncBytesInterval = 33554432;
 
   public ChangelogClientConfig(String storeName) {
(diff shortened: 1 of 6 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/ChangelogClientConfig.java
```diff
@@ -142,6 +144,15 @@ public ChangelogClientConfig setVersionSwapDetectionIntervalTimeInMs(long interv
     return this;
   }
 
+  public long getDatabaseSyncBytesInterval() {
+    return databaseSyncBytesInterval;
+  }
+
+  public ChangelogClientConfig setDatabaseSyncBytesInterval(long databaseSyncBytesInterval) {
```
ZacAttack, line 151: let's add a bit of javadoc here to specify what this is for.

zx0401: Sure, added.

ZacAttack (approved):
Looks good

PR #878, opened 2024-02-27
[vpj] send out pushjob details while hitting exception.

majisourav99:
## Send out pushjob details while hitting exception

During target colo push if the later push job fails, it does not send out the pushjob details to the controller, which later breaks the h2v availability report. This PR sends out pushjob details before throwing exceptions.

## How was this PR tested?

GHCI


clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
```diff
@@ -2450,6 +2455,7 @@ void pollStatusUntilComplete(
           VenicePushJob.this.pushJobSetting.storeResponse.getStore().getBootstrapToOnlineTimeoutInHours();
       long durationMs = System.currentTimeMillis() - pollStartTimeMs;
       if (durationMs > TimeUnit.HOURS.toMillis(bootstrapToOnlineTimeoutInHours)) {
+        sendPushJobDetailsToController();
```
nisargthakkar, line 2458: Can you explain a bit about how this will help? IIUC, if an exception is thrown, the catch clause will send push job details after marking it as an ERROR

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
```diff
@@ -859,6 +859,8 @@ public void run() {
         sendPushJobDetailsToController();
         closeVeniceWriter();
       } catch (Exception ex) {
+        sendPushJobDetailsToController();
+        closeVeniceWriter();
```
adamxchen, line 863: How about moving 859-863 to the final block if we have to invoke them anyways?

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
```diff
@@ -2429,12 +2436,11 @@ void pollStatusUntilComplete(
           // but if the majority of datacenters have completed, we give up on the unreachable datacenter
           // and start truncating the data topic.
           // 2) For targeted region push, not all targeted regions have completed.
-
           StringBuilder errorMsg = new StringBuilder().append("Push job error reported by controller: ")
               .append(pushJobSetting.veniceControllerUrl)
               .append("\ncontroller response: ")
               .append(response);
-
+          sendPushJobDetailsToController();
```
adamxchen, line 2443: Why is this needed if we already have line 862-863?

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
nisargthakkar: Can we veify that the colo status is being constructed correctly? I don't see how adding more `sendPushJobDetailsToController()` will fix the issue of incomplete colo statuses

PR #879, opened 2024-02-28
[all] Fix an infinite recursion issue in VeniceWriter

lluwm:
Today, VeniceWriter (VW) has a bug to run into an infinite recursion when it sends a message in the VeniceWriter and eventually it throws StackOverflowError. For a particular partition, the conditions to trigger the issue are:

1. VW has a cached Segment
2. The segment is neither started nor ended
3. The elapsed time for the segment is greater than MAX_ELAPSED_TIME_FOR_SEGMENT_IN_MS

To fix it, we can break the infinite loop by always recording the new start time.

## How was this PR tested?
GH action

### DIFF
--- a/internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
@@ -238,10 +238,10 @@ public class VeniceWriter<K, V, U> extends AbstractVeniceWriter<K, V, U> {
    */
   private final Segment[] segments;
   /**
-   * Map of partition to its segment creation time in milliseconds.
+   * Map of partition to its segment start time in milliseconds.
    * -1: the current segment is ended
    */
-  private final long[] segmentsCreationTimeArray;
+  private final long[] segmentsStartTimeArray;
   private final KeyWithChunkingSuffixSerializer keyWithChunkingSuffixSerializer = new KeyWithChunkingSuffixSerializer();
   private final ChunkedValueManifestSerializer chunkedValueManifestSerializer =
       new ChunkedValueManifestSerializer(true);
@@ -361,13 +361,13 @@ public class VeniceWriter<K, V, U> extends AbstractVeniceWriter<K, V, U> {
       if (this.numberOfPartitions <= 0) {
         throw new VeniceException("Invalid number of partitions: " + this.numberOfPartitions);
       }
-      this.segmentsCreationTimeArray = new long[this.numberOfPartitions];
+      this.segmentsStartTimeArray = new long[this.numberOfPartitions];
       // Prepare locks for all partitions instead of using map to avoid the searching and creation cost during
       // ingestion.
       this.partitionLocks = new Object[this.numberOfPartitions];
       for (int i = 0; i < numberOfPartitions; i++) {
         partitionLocks[i] = new Object();
-        segmentsCreationTimeArray[i] = -1L;
+        segmentsStartTimeArray[i] = -1L;
       }
       this.segments = new Segment[this.numberOfPartitions];
       OPEN_VENICE_WRITER_COUNT.incrementAndGet();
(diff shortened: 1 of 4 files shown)


internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
sushantmane, line 1: I'm wondering if we should consider re-writing this code so that there is no chance of recursion at all...

gaojieliu: That is fair callout, and we need to consider how to refactor the code to avoid this recursive pattern.
Today, the problem happens since the segmentation happens inside `getSegment`, which is being used to construct a KME message.
The loop is as follows:
```
sendMessage
getKafkaMessageEnvelope
getSegment
startSegment
sendStartOfSegment
sendControlMessage
sendMessage
```
This is a long loop, which contains several steps.
One way is to move the segmentation logic to the upper level into `put`/`update`/`delete` calls and we can evaluate how big change can be.
Feel free to share your thoughts here as well.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/writer/VeniceWriterTest.java
```diff
@@ -227,4 +230,53 @@ public void testVeniceWriterClose(boolean doFlush) {
       }
     }
   }
+
+  /**
+   * This is a regression test for the VeniceWriter issue where the VeniceWriter could run into
```
FelixGV, line 235: Thanks for writing the integration test. That is good, but is it possible to also add a unit test which repros this issue?

FelixGV (commented):
Thanks! Good catch!

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
lluwm: I agree and I am not a big fan of having recursive calls in basic functionality such as `sendMessage`, if we could avoid them. What's your opinion if we tackle the code refactor problem in a follow up change? The reasoning behind is that this issue is currently a blocker, thus we probably want to fix it first to unblock ourselves. Meanwhile, this would also buy us more time to refactor code with better understanding/reasoning and cautions. WDYT?

sushantmane: Here is one way I think we could do achieve this with minimal code changes just by separate/restructuring some logic

```console
Future sendMessage()
    --> getOrUpdateSegment() 
        if segment does not exist
          - createSegment(); getKme(); stampKmeWithSoS(); sendToProducer() // NO sendControlMessage()
        if segement exists
          if we need to end the segment
            - getKme(); stampKmeWithEoS(); sendToProducer() // NO sendControlMessage() 
            - createSegment(); getKme(); stampKmeWithSoS(); sendToProducer() // NO sendControlMessage()
        return the segment 
     --> getKme(); stampKmeWithMessage(); stampKmeWithDivInfo(); sendToProducer() // NO sendControlMessage()



KME getKme() {
    // does not deal with the segement manupulation; just returns the empty envelope
}


Future sendToProducer() {
    // does not manupulate the segment or div info; just hand over the envelope to the producer
}
```

sushantmane: > this issue is currently a blocker

I agree this issue is critical, but it isn't necessarily the release blocker IMHO.

gaojieliu: Isn't it true?
If this issue happens, GCN can happen, since it can cause consumer thread or drainer thread auto-exit, which will result in push job failure and data staleness, which are severe prod outages.
Release blocker covers both the issues introduced by new release and newly discovered critical issues.
If we don't want to have a regular release to include this, we may go with a hot-fix release considering the severity if happens.

sushantmane: IIUC the chances of this happening are fairly low... 

That being said I don't mind if we want to do refactoring in a separate PR

gaojieliu: The chance is low, but the consequence is severe..
So we should be ready for it if it happens, and it does happen in the past a few times already as consumer auto-termination. 
Since we identify the problem, we should rollout the fix ASAP to avoid another prod incident.

lluwm: If we are okay with the option to do code refactoring in a follow up PR, I will resolve this comment and file a ticket to track this work

gaojieliu (approved):
Thanks Lei!

PR #880, opened 2024-02-28
[server][metrics] Add filtering to heartbeat versioned stats for hybrid stores

ZacAttack:
## [server][metrics] Add filtering to heartbeat versioned stats for hybrid stores

VersionedStats assumes all metrics should be applied for all stores.  This change overrides methods which add stats to check to see if the node should report metrics for the store getting added or not.

Resolves #XXX

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/AbstractVeniceAggVersionedStats.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/AbstractVeniceAggVersionedStats.java
@@ -50,7 +50,7 @@ public abstract class AbstractVeniceAggVersionedStats<STATS, STATS_REPORTER exte
     loadAllStats();
   }
 
-  public final synchronized void loadAllStats() {
+  public synchronized void loadAllStats() {
     metadataRepository.getAllStores().forEach(store -> {
       addStore(store.getName());
       updateStatsVersionInfo(store.getName(), store.getVersions(), store.getCurrentVersion());
@@ -85,7 +85,7 @@ public abstract class AbstractVeniceAggVersionedStats<STATS, STATS_REPORTER exte
     return stats;
   }
 
-  private VeniceVersionedStats<STATS, STATS_REPORTER> addStore(String storeName) {
+  protected VeniceVersionedStats<STATS, STATS_REPORTER> addStore(String storeName) {
     return aggStats.computeIfAbsent(
         storeName,
         s -> new VeniceVersionedStats<>(metricsRepository, storeName, statsInitiator, reporterSupplier));
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/ingestion/heartbeat/HeartbeatMonitoringService.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/ingestion/heartbeat/HeartbeatMonitoringService.java
@@ -65,7 +65,7 @@ public class HeartbeatMonitoringService extends AbstractVeniceService {
         () -> new HeartbeatStat(new MetricConfig(), regionNames),
         (aMetricsRepository, s) -> new HeartbeatStatReporter(aMetricsRepository, s, regionNames),
         leaderHeartbeatTimeStamps,
-        leaderHeartbeatTimeStamps);
+        followerHeartbeatTimeStamps);
   }
 
   private synchronized void initializeEntry(
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/ingestion/heartbeat/HeartbeatMonitoringService.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/ingestion/heartbeat/HeartbeatMonitoringService.java
@@ -63,7 +63,8 @@ public class HeartbeatMonitoringService extends AbstractVeniceService {
         metricsRepository,
         metadataRepository,
         () -> new HeartbeatStat(new MetricConfig(), regionNames),
-        (aMetricsRepository, s) -> new HeartbeatStatReporter(aMetricsRepository, s, regionNames));
+        (aMetricsRepository, s) -> new HeartbeatStatReporter(aMetricsRepository, s, regionNames),
+        this);
   }
 
   private synchronized void initializeEntry(
@@ -231,6 +232,12 @@ public class HeartbeatMonitoringService extends AbstractVeniceService {
   @FunctionalInterface
   interface ReportLagFunction {
     void apply(String storeName, int version, String region, long lag);
+  }
+
+  Set<String> getReportedStores() {
+    Set<String> monitoredStores = leaderHeartbeatTimeStamps.keySet();
+    monitoredStores.addAll(followerHeartbeatTimeStamps.keySet());
+    return monitoredStores;
   }
 
   private class HeartbeatReporterThread extends Thread {
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/ingestion/heartbeat/HeartbeatMonitoringService.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/ingestion/heartbeat/HeartbeatMonitoringService.java
@@ -95,12 +95,12 @@ public class HeartbeatMonitoringService extends AbstractVeniceService {
       Map<String, Map<Integer, Map<Integer, Map<String, Long>>>> heartbeatTimestamps,
       Version version,
       int partition) {
-    heartbeatTimestamps.computeIfPresent(version.getStoreName(), (storeKey, map1) -> {
-      map1.computeIfPresent(version.getNumber(), (versionKey, map2) -> {
-        map2.remove(partition);
-        return map2;
+    heartbeatTimestamps.computeIfPresent(version.getStoreName(), (storeKey, versionMap) -> {
+      versionMap.computeIfPresent(version.getNumber(), (versionKey, partitionMap) -> {
+        partitionMap.remove(partition);
+        return partitionMap;
       });
-      return map1;
+      return versionMap;
     });
   }
 
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/ingestion/heartbeat/HeartbeatMonitoringService.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/ingestion/heartbeat/HeartbeatMonitoringService.java
@@ -71,7 +71,8 @@ public class HeartbeatMonitoringService extends AbstractVeniceService {
   private synchronized void initializeEntry(
       Map<String, Map<Integer, Map<Integer, Map<String, Long>>>> heartbeatTimestamps,
       Version version,
-      int partition) {
+      int partition,
+      boolean isFollower) {
     // We don't monitor heartbeat lag for non hybrid versions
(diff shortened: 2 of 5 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/ingestion/heartbeat/HeartbeatMonitoringService.java
```diff
@@ -74,36 +76,32 @@ private synchronized void initializeEntry(
...
       Version version,
       int partition) {
-    if (heartbeatTimestamps.get(version.getStoreName()) != null) {
-      if (heartbeatTimestamps.get(version.getStoreName()).get(version.getNumber()) != null) {
-        if (heartbeatTimestamps.get(version.getStoreName()).get(version.getNumber()).get(partition) != null) {
-          heartbeatTimestamps.get(version.getStoreName()).get(version.getNumber()).remove(partition);
-        }
-      }
-    }
+    heartbeatTimestamps.computeIfPresent(version.getStoreName(), (storeKey, map1) -> {
```
adamxchen, line 98: nit: can we call them `perVersionMap`, `perPartitionMap` etc. like you did for line 190? Prefer to not have them as `map1`, `map2`

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/ingestion/heartbeat/HeartbeatVersionedStats.java
```diff
@@ -23,4 +34,42 @@ public void recordLeaderLag(String storeName, int version, String region, long l
...
+
+  @Override
+  public void handleStoreChanged(Store store) {
+    Set<String> stores = getReportedStores();
+    if (stores.contains(store.getName())) {
+      updateStatsVersionInfo(store.getName(), store.getVersions(), store.getCurrentVersion());
+    }
+  }
+
+  Set<String> getReportedStores() {
```
adamxchen, line 65: I am trying to understand why we need to do this. Is it because `metadataRepository` could have some stale stores that are no longer exist on the host?

ZacAttack: No.  The reason why we have to do this is far dumber then that.  The parent class of this class calls it's own methods from it's constructor.  Specifically, it calls the loadAllStats() function which in turn invokes our own overridden method and then calls this.  Since in java the first line of constructor for a subclass has to be super(...), this means that this function will get called before these maps are assigned/initialized.  So we have to add this dumb defensive check for that.

The TODO is here to tell us that when we get around to fixing that code, we can remove this.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/ingestion/heartbeat/HeartbeatVersionedStats.java
```diff
@@ -23,4 +34,42 @@ public void recordLeaderLag(String storeName, int version, String region, long l
   public void recordFollowerLag(String storeName, int version, String region, long lag) {
     getStats(storeName, version).recordFollowerLag(region, lag);
   }
+
+  @Override
+  public synchronized void loadAllStats() {
+    Set<String> stores = getReportedStores();
+    metadataRepository.getAllStores().forEach(store -> {
```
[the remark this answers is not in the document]
ZacAttack: Yeah that's right, but that's ok.  We don't really want to start reporting until Helix assigns a state transition.  When the reporting thread runs it'll invoke the right methods for getting the stat onboarded, loadAllStats is just being proactive at start up time.  That's fine, but it's not desirable for this case.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/ingestion/heartbeat/HeartbeatMonitoringService.java
ZacAttack: fixed

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/ingestion/heartbeat/HeartbeatVersionedStats.java
```diff
@@ -2,18 +2,29 @@
...
 import com.linkedin.venice.stats.StatsSupplier;
 import io.tehuti.metrics.MetricsRepository;
+import java.util.Collections;
+import java.util.Map;
+import java.util.Set;
 import java.util.function.Supplier;
 
 
 public class HeartbeatVersionedStats extends AbstractVeniceAggVersionedStats<HeartbeatStat, HeartbeatStatReporter> {
+  private Map<String, Map<Integer, Map<Integer, Map<String, Long>>>> leaderMonitors;
```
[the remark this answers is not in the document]
ZacAttack: fixed

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/ingestion/heartbeat/HeartbeatVersionedStats.java
ZacAttack: Actually, I think I misunderstood your question.  That is right.  As partitions are moved off of hosts, we'd like to stop reporting on them.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/ingestion/heartbeat/HeartbeatVersionedStats.java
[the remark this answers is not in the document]
FelixGV: Hmm... but it looks like `loadAllStats` is ONLY called by the constructor, so shall we make it a no-op, rather than have any code at all in it? Or maybe I'm not understanding correctly what the goal of this override is 🤔

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/ingestion/heartbeat/HeartbeatVersionedStats.java
```diff
@@ -23,4 +34,42 @@ public void recordLeaderLag(String storeName, int version, String region, long l
...
+  }
+
+  Set<String> getReportedStores() {
+    if (leaderMonitors == null || followerMonitors == null) {
+      // TODO: We have to do this because theres a self call in the constructor
+      // of the superclass of this class. We shouldn't have to do this
+      return Collections.EMPTY_SET;
+    }
+    Set<String> monitoredStores = leaderMonitors.keySet();
+    monitoredStores.addAll(followerMonitors.keySet());
```
FelixGV, line 72: `CHM::keySet` does not support additions, it is immutable, according to the JavaDoc and code... can you write a unit test that exercises `handleStoreChanged` (or any of the methods calling this function, besides the constructor) and see if this code actually works?

If the above bug suspicion is correct, then it can be fixed easily can copying the key set into a new set... but I feel like creating a combined set is for the sake of doing one `contains` on it is too expensive... can you refactor this so that we call `containsKey` on the two maps of interest instead?

ZacAttack: Sure

ZacAttack: Fixed

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/ingestion/heartbeat/HeartbeatVersionedStats.java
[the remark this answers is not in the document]
ZacAttack: That would probably be legit to do.  I'll do that.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/ingestion/heartbeat/HeartbeatVersionedStats.java
```diff
@@ -23,4 +32,32 @@ public void recordLeaderLag(String storeName, int version, String region, long l
...
   }
+
+  @Override
+  public synchronized void loadAllStats() {
+    // No-op
+  }
+
+  @Override
+  public void handleStoreCreated(Store store) {
+    if (isStoreAssignedToThisNode(store.getName())) {
```
nisargthakkar, line 43: When would this get called? The store will be created before a version is created. After a version is created, a SN will get assigned a partition. So, do we need this? Can we modify `handleStoreChanged` to call `addStore` only when `isStoreAssignedToThisNode` is `true`?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/AbstractVeniceAggVersionedStats.java
```diff
@@ -50,7 +50,7 @@ public AbstractVeniceAggVersionedStats(
     loadAllStats();
   }
 
-  public final synchronized void loadAllStats() {
+  public synchronized void loadAllStats() {
     metadataRepository.getAllStores().forEach(store -> {
       addStore(store.getName());
       updateStatsVersionInfo(store.getName(), store.getVersions(), store.getCurrentVersion());
```
nisargthakkar, line 56: So, a higher level question. Should every SN register metrics for every store? Can we instead move the improvement of registering only the necessary metrics one-level up so everything benefits?

ZacAttack: I think we should do something like that, but I think in order to do that we have to look at every implementer of this class and figure out what the policy should be.

Probably what we could do is add an overridable method which supports filtering of a store, and have them implement that.  But I think doing that dive is out of scope for this PR.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/ingestion/heartbeat/HeartbeatVersionedStats.java
ZacAttack: Overriding addStore might actually be the better way to go as I think about it.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/AbstractVeniceAggVersionedStats.java
nisargthakkar: Yeah, I agree about it being out of scope

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/ingestion/heartbeat/HeartbeatVersionedStats.java
ZacAttack: oof.  Actually I remember why I didn't do this for addStore.  addStore isn't a void method and I was worried about triggering NPE's somewhere else.  I guess coming back to storeCreated.  storeCreated should honestly probably NoOp as well as you are correct.  There's no way a version will be created when a store is first created, so theres not really any reason to react here.

nisargthakkar (approved):
LGTM

PR #881, opened 2024-02-28
[doc] VIP-4: Store Lifecycle Hooks

FelixGV:
See the rendered version of this VIP here: https://felixgv.github.io/venice/docs/proposals/vip-4

This VIP proposes several new configs, metrics and APIs. See PR for details.


docs/proposals/proposals.md
```diff
@@ -16,3 +16,4 @@ This folder includes all the Venice Improvement Proposals. Click to read further
 | [VIP-1](vip-1.md)                                    | Authentication Service API                                                     | Accepted         |
 | [VIP-2](https://github.com/linkedin/venice/pull/513) | Removing Per Record Offset Metadata From Venice-Server Storage With Heartbeats | Under Discussion |
 | [VIP-3](vip-3.md)                                    | Rust Server Read Path                                                          | Under Discussion |
+| [VIP-4](vip-4.md)                                    | Store Lifecycle Hooks                                                          | Under Discussion |
```
gaojieliu, line 19: Isn't it more about `Store Version Lifecycle` rather than `Store Lifecycle`?

docs/proposals/vip-4.md
```diff
@@ -0,0 +1,217 @@
...
+|                    `preStoreVersionCreation` |        ✅         |         ✅         |    ✅     |
+|                   `postStoreVersionCreation` |        ✅         |         ✅         |    ✅     |
+|  `preStartOfStoreVersionIngestionForDaVinci` |       2️⃣        |        2️⃣        |   2️⃣    |
+| `postStartOfStoreVersionIngestionForDaVinci` |       2️⃣        |        2️⃣        |   2️⃣    |
+|          `postStoreVersionLeaderReplication` |        ✅         |         ❌         |    ❌     |
+|                        `preStoreVersionSwap` |        ✅         |         ✅         |    ✅     |
+|                       `postStoreVersionSwap` |        ✅         |         ✅         |    ✅     |
+|                            `preEndOfPushJob` |        ✅         |         ✅         |    ✅     |
+|                           `postEndOfPushJob` |        ✅         |         ✅         |    ✅     |
+
```
gaojieliu, line 119: This is covering a bigger scope than region-by-region version swap and are we supposed to support all of them?

docs/proposals/vip-4.md
```diff
@@ -0,0 +1,217 @@
...
+It is also possible to consider invoking some hooks in one of these location while other hooks would be executed 
+elsewhere. The table below summarizes the feasibility and tradeoffs for each of the proposed hooks:
+
+|                                              | Child Controller | Parent Controller | Push Job |
+|---------------------------------------------:|:----------------:|:-----------------:|:--------:|
+|                          `preStartOfPushJob` |        ❌         |         ✅         |    ✅     |
+|                         `postStartOfPushJob` |        ❌         |         ✅         |    ✅     |
+|                      `preSchemaRegistration` |        ❌         |         ✅         |   1️⃣    |
+|                     `postSchemaRegistration` |        ✅         |         ✅         |   1️⃣    |
```
gaojieliu, line 109: Schema registration is a global change today, so are you considering to make it region specific to avoid global outage? Also schema rollback is not supported today, and it needs client to support read-on-write schema support.

docs/proposals/vip-4.md
```diff
@@ -0,0 +1,217 @@
...
+
+* `venice.store.lifecycle.hooks.threads`: Number of threads used by the hooks framework to execute all hooks.
+
+* `venice.store.lifecycle.hooks.wait.interval.seconds`: The time to wait before re-invoking a hook which returned `WAIT` 
+  (default: 60 seconds).
+
+* `venice.store.lifecycle.hooks.configs.<arbitrary>`: Any number of configs to be passed (after clipping everything
+  before the `<arbitrary>` part) into the hooks constructor.
+
+In addition, the store config will get a new `Map<String, String>` of store-level config overrides.
```
majisourav99, line 161: Why do we need this ? these can be overriden from pushjob itself?

docs/proposals/vip-4.md
```diff
@@ -0,0 +1,217 @@
...
+
+1. Within the push job.
+2. Within the parent controller.
+3. Within the child controllers.
+
+It is also possible to consider invoking some hooks in one of these location while other hooks would be executed 
+elsewhere. The table below summarizes the feasibility and tradeoffs for each of the proposed hooks:
+
+|                                              | Child Controller | Parent Controller | Push Job |
+|---------------------------------------------:|:----------------:|:-----------------:|:--------:|
```
majisourav99, line 105: can we consider safety of running the validation hook in various components ? vpj would be the most safe ?

docs/proposals/vip-4.md
```diff
@@ -0,0 +1,217 @@
...
+controller could not trigger the hook.
+
+2️⃣ : The hook for the start of ingestion for Da Vinci Clients is tricky for a few reasons. The start of ingestion is 
+controlled by updating the Meta Store, which is a non-replicated system store updated by child controllers, so those
+must be involved (either by running the hook there in the first place, or by having some mechanism that enables the
+parent or VPJ to inform the child controllers of when the system store is eligible for getting updated, such as by
+adding a new field to the `AddVersion` admin channel command). However, see Rollback Support below for why running hooks
+in the child controller may be insufficient.
+
+### Rollback Support
```
majisourav99, line 133: So the rollback option only makes sense in `postStoreVersionSwap` stage? for previous steps what does rollback mean?

docs/assets/images/push_job_steps.drawio
```diff
@@ -0,0 +1,250 @@
+<mxfile host="Electron" modified="2024-02-26T21:48:03.768Z" agent="Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) draw.io/23.1.5 Chrome/120.0.6099.109 Electron/28.1.0 Safari/537.36" etag="BhlpWibByn0qgMP8agoQ" version="23.1.5" type="device">
```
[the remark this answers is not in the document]
FelixGV: Sorry that this is unclear. The "Server Read Traffic SERVING" step is the swap. From my understanding, if there are DVC instances reporting their status into the local DVPS system store, then the child controller will wait for all heartbeating DVC instances to report their completion before doing the swap, and therefore the server swap depends on the DVC swap. Please LMK if this is not the right understanding (or if you think we should change this behavior for some reason...).

And so regarding the swap being in the per-region flow, it is there, but named as "serving" rather than "swap".

docs/proposals/proposals.md
FelixGV: In the current proposal, it is both, and I consider store-version lifecycle to be a subset of the store lifecycle.

Schema registration is a global event which I think we may want to emit monitoring events for, so that is why I included it.

docs/proposals/vip-4.md
FelixGV: We don't necessarily need to support everything. As I mentioned later in the doc, I recommend skipping the replication complete hook, since that one is most easily accomplished as a child controller hook, and there are other reasons why I think that is not the most straightforward approach. If we could have everything, I think replication complete would be an interesting event to emit, but it is not strictly required in the short-term.

Conversely, there could be other hooks we want but which are currently omitted...

In other words, the exact scope of the first iteration can be tweaked, although many of these hooks are must haves in the short-term.

docs/proposals/vip-4.md
FelixGV: Please take a look at the code included in this PR, which I think makes things more clear than the text.

Some hooks return the `StoreLifecycleEventOutcome` enum, which allows only `PROCEED` or `ABORT`. This is the case of the schema registration hook.

Other hooks return the `StoreVercionLifecycleEventOutcome` enum, which allows not only `PROCEED` and `ABORT`, but also `WAIT` and `ROLLBACK`.

Finally, some hooks return `void` and thus have no ability to affect anything.

IOW, the proposed API uses strong typing to make it very clear and strict what each hook can or cannot accomplish.

docs/proposals/vip-4.md
FelixGV: Good question!

This is not indicated in this doc... but we have some internal requirement to control in a centralized manner some of the hooks' behaviors. This will be easier to accomplish via store configs than push job configs.

For behaviors that we wish to allow the user to override arbitrarily, it is conceivable to have those as VPJ configs...

docs/proposals/vip-4.md
FelixGV: Yes, that is true.

For hooks needing complex logic, we may want to isolate their business logic into a separate service, and have the Venice-side hook be just a thin client for interacting with the other service...

But if we were to pack a lot of logic into the hook itself, then the more widely distributed and per-user nature of VPJ would indeed be safer...

But running inside VPJ for fine-grained control of push behaviors is also potentially tedious... it seems like it might be easier to build the hooks framework in the location which already contains a lot of the relevant logic and state. But that's not an obvious conclusion, and certainly, all options are feasible.

docs/proposals/vip-4.md
FelixGV: If a hook returns `ROLLBACK` prior to the first region having swapped, then it's simply equivalent to `ABORT`... but if any other region already swapped, then it can be done.

The main purpose of the `preStoreVersionSwap` is for the option to return `WAIT`.

If we want, we could have three enums instead of two... and the additional enum would contain `PROCEED`, `ABORT` and `WAIT` but not `ROLLBACK`... but it seems a bit tedious to have this many enums... the API is simpler with fewer enums, IMHO.

But feel free to disagree, of course. LMK what you think.

docs/proposals/vip-4.md
```diff
@@ -0,0 +1,217 @@
...
+* [StoreLifecycleEventOutcome](/venice/javadoc/com/linkedin/venice/hooks/StoreLifecycleEventOutcome.html), which is the signal
+  returned by some hooks to indicate that a given step should proceed or abort.
+* [StoreVersionLifecycleEventOutcome](/venice/javadoc/com/linkedin/venice/hooks/StoreVersionLifecycleEventOutcome.html), which
+  is the signal returned by some other hooks which need more fine-grained control over the workflow. In addition to 
+  proceeding and aborting, this also provides the option to wait, which tells the hooks framework to try invoking the 
+  hook again later, and rollback, which tells the framework to rollback to the previous store-version in all regions.
+
+## Proposed Design
+
+The main design consideration is where to execute the hooks. At a high level, there are three options:
```
adamxchen, line 95: Are we going to only choose one option? Just brainstorm here. Customers may choose to execute one individual hook per region for a life cycle event and after it run another hook that only executes once globally. Should this be supported? What's the max flexibility we want to provide?

docs/assets/images/push_job_steps.drawio.svg
adamxchen, line 1: Love the diagram! Thanks a lot for making it!

docs/proposals/vip-4.md
```diff
@@ -0,0 +1,217 @@
...
+  which the parent controller does periodically in order for the child to inform the parent of the need to rollback, or 
+  else build a new mechanism for the child to directly interact with the parent (or even with other child controllers 
+  directly...).
+
+* If the hooks are executed in VPJ, then it would need to interact with the parent via the controller client to trigger 
+  the rollback.
+
+### Configs
+
+There needs to be new configs:
```
sixpluszero, line 149: Regarding these hooks implementation, I am wondering how will they look like. 
Will it be some classes to be instantiated at run time using reflection and executed? If so, I can see it is going to be registered at LI Miltiproduct level (Like VPJ or Venice-backend), just as how partitioner implementation works. If so, it will not be very flexible for partners to register and evolve hooks implementation for their runs, as it requires new version roll out and deployment... 

For the beginning it might be fine, but I am wondering just like our custom partitioner vision, should we create a service that allows user to dynamically register hooks implementation.

docs/proposals/vip-4.md
FelixGV: Yeah, good comments. A few thoughts on these points:

1. For the first iteration of this work, I think it would simplify things if we ran the hooks framework in only one component. My preference is to pick the parent controller, but I think VPJ might be fine too (different tradeoffs).
2. In the future, I think it is conceivable to expand the scope of hooks and that some of them could run inside other components, such as in the child controllers, or maybe even in the servers (I don't have a use case in mind for that, but theoretically, there could be...).
3. Regarding the flexibility of letting a hook implementer decide if a hook will run once per region or once globally, I think using either the parent controller or VPJ probably gets us closest to that... by letting one component handle all of the regions, then it may facilitate making certain types of decisions based on whether another regional hook already did something or not (although it involves maintaining state which could still be tricky, but not insurmountable...).

docs/assets/images/push_job_steps.drawio.svg
FelixGV: Yeah, thanks! I think it turned out well. I've been enjoying draw.io quite a bit for making open source-able diagrams... I'm thinking of also showing this diagram in the documentation, but not sure where yet.

docs/proposals/vip-4.md
FelixGV: Yeah so... I think your point is totally right regarding custom partitioners. That stuff is too difficult for users to register and evolve, and dynamic registration would be super.

Although I have not made that clear in the doc (but I should...), in the case of these lifecycle hooks, the targeted stakeholder is the Venice operator, not the Venice user, so I don't think dynamic registration is critical. Dynamic registration is not completely useless in the context of hooks, but I would say it's a fairly minor nice to have. As such, I would prefer leaving it out of scope of this work. Though if in the future we end up building dynamic registration for the sake of custom partitioner, then we could certainly use it for hooks as well.

docs/assets/images/push_job_steps.drawio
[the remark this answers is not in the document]
FelixGV: I added notes in the intro to capture the above details.

docs/proposals/vip-4.md
```diff
@@ -0,0 +1,248 @@
...
+
+* The compression job is optional, and whether it runs or not depends on configurations.
+
+* The data push job writes to a store-version pub sub topic in one of the regions (typically the one which is local to
+  where the push job is running), but all regions start ingesting right away, as soon as the data push job begins 
+  writing. For regions that are remote from the topic the data push job is writing to, the leader servers are performing 
+  replication, while in the region which contains that topic, the replication is a no-op.
+
+* The `SERVING` step is also called "version swap". If there are Da Vinci Clients, they will ingest and swap on their
+  own, and the child controller of that region will wait until all DVC instances have swapped (started serving) before
```
gaojieliu, line 41: Is this true?
I recalled DaVinci will report completed, the switch will wait for the version switch decided by the Venice Controller, and it won't switch right after completion.
cc @sixpluszero

docs/proposals/vip-4.md
```diff
@@ -0,0 +1,248 @@
...
+submitting this proposal, it is undetermined whether this work will apply to stream reprocessing jobs. In terms of 
+priority, we care mostly about supporting full pushes from offline grids, and it may be fine to leave stream 
+reprocessing out of scope, although depending on the design details we choose, we may be able to support stream 
+reprocessing "for free" as well.
+
+The goal is for lifecycle hooks to achieve the following use cases:
+
+* Orchestrate how data is served in each region, including:
+  * Ensuring a minimum delay (e.g. 1 hour) between each region beginning to serve the new store-version.
+  * Delaying the swapping of a new store-version to be within some time of day (e.g. during "business hours").
```
gaojieliu, line 83: Do we treat business hour range as a global setting or per-store setting? Different customers might have different assumption about the range of the business hours.

docs/proposals/vip-4.md
gaojieliu: There are different types of complexity if we choose different options.
The lifecycle management is stateful, which means if we choose to execute it in Controller, we may need to transfer the hook state during rebalance. Essentially, checkpointing and recovery will be required when changing Leader Controller for a given cluster, and if the hooks are executed in PushJob, I assume we don't need to handle such checkpointing/recovery logic since push job executor won't change during the lifetime of a certain run.

docs/proposals/vip-4.md
gaojieliu: I guess the solution can be a combination of collaboration among several components.
For example, VPJ can be the orchestrator to maintain the state of the lifecycle execution, and Controller can be a stateless component for lifecycle management, but it just responds to the VPJ ask to execute the corresponding hooks/runtime blindly if required (or we think some hooks are better or more convenient to be executed in Controller instead of VPJ) without maintaining the stage of the lifecycle management.

docs/proposals/proposals.md
gaojieliu: Sounds good.

docs/proposals/vip-4.md
FelixGV: No... as I said, please take a look at the abstract class in this PR, which should make the contract of each hook clear.

The schema registration has two hooks: pre- and post-registration. Only the pre-registration hook has the ability to abort. The post-hook is `void` and thus for informational purpose only.

docs/proposals/vip-4.md
```diff
@@ -0,0 +1,248 @@
...
+In order to support the ability for the hooks which return the `StoreVersionLifecycleEventOutcome` to rollback, the most
+natural way to achieve this is likely to involve the parent controller, either by having those hooks run there in the 
+first place, or by having a propagation mechanism to it:
+
+* If the hooks are executed in the child controller, the propagation mechanism might be to extend the job status check 
+  which the parent controller does periodically in order for the child to inform the parent of the need to rollback, or 
+  else build a new mechanism for the child to directly interact with the parent (or even with other child controllers 
+  directly...).
+
+* If the hooks are executed in VPJ, then it would need to interact with the parent via the controller client to trigger
```
gaojieliu, line 170: Orchestrator can be different from hook executor.
BTW, for the delayed promotion to other regions, I assume we are going to halt the push job and still report the final status in the push job by either abort or rolling forward, right?
There are some costs with this approach since VPJ executor will take more resources than before, and if we don't halt push job, the final status won't be easily observed by the end user.
Let us call out the additional cost/impact to the existing components if applicable.

docs/proposals/vip-4.md
gaojieliu: There is a benefit to let Parent Controller to be the orchestrator, stream processing job can be easily supported since it doesn't have an executor of push job as VPJ.

docs/proposals/vip-4.md
FelixGV: I think the absolute must haves are:

```
preStartOfStoreVersionIngestionForDaVinci
preStoreVersionSwap
postStoreVersionSwap
preEndOfPushJob
```

And it would be cleaner (but not necessarily critical) to have those as well:

```
postStartOfPushJob
postSchemaRegistration
preStoreVersionCreation
postStoreVersionCreation
postEndOfPushJob
```

The rest are less important in the short-term.

docs/proposals/vip-4.md
FelixGV: > I guess the solution can be a combination of collaboration among several components.

Yep, it can be.

> There is a benefit to let Parent Controller to be the orchestrator, stream processing job can be easily supported since it doesn't have an executor of push job as VPJ.

That's right... and there are other similar issues related to using VPJ instead of the parent controller.

For example, schema registration (which we would like to emit a monitoring event for) would be caught in all cases by the controller but not by VPJ (there are other solutions of course, such as emitting the monitoring even in the other components besides VPJ that can register schemas, but it's potentially whack-a-mole, whereas the controller will for sure see all schema registration calls).

Another example is version swap... it is possible to invoke this via the controller client / admin tool outside of a push job, and we may want to invoke hooks in those cases too (at least for monitoring event purposes, but possibly for flow control purposes, although for flow control we might want to support an override to "skip hooks" or override specific hook configs, when calling from the admin tool...).

docs/proposals/vip-4.md
FelixGV: Yeah that's right... state management is easier when using VPJ...

That being said, I am not sure that we have a use case that absolutely must have state.

Another aspect related to this is that if VPJ dies abruptly today (without being able to call its graceful kill/cancel function, which does happen sometimes) then as of today the push is capable of completing on its own, even without VPJ running (VPJ is only an observer). Whereas if we make VPJ the orchestrator, then VPJ dying can result in a hanging push (unless we adopt the VPJ heartbeat mechanism, which would be a good thing to have in and of itself, but causes scope creep in this project...).

docs/proposals/vip-4.md
```diff
@@ -0,0 +1,248 @@
...
+  SERVING" step.
+
+## Problem Statement 
+
+Pushing to all regions in parallel makes the push faster, but it also means that if the push causes an issue, the impact
+is going to be global (affecting all regions). It would be desirable to have certain checks and balances that reduce the
+blast radius in cases where the content of the push causes issues. Examples of such issues include:
+
+* Data quality issues:
+  * Incomplete data (due to issues in data generation logic, or in upstream datasets).
```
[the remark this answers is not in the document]
FelixGV: We can do that, yeah. We could leverage the push status system store which contains heuristics about all pushes to compare the size of the previous push to the size of the incoming one, and then have some threshold for the maximum % change allowed...

docs/proposals/vip-4.md
```diff
@@ -0,0 +1,248 @@
...
+  * Change in semantics (e.g. some embeddings trained by a new ML model / weights / params / etc. are incompatible with 
+    that used at inference-time).
+* Schema-related issues:
+  * Some optional fields which used to always be populated get deleted (or get populated with null) and the reading app 
+    fails due to lack of null checking. This kind of app-side bug can happen even though a schema evolution is fully 
+    compatible.
+* Infra issues:
+  * Larger payloads take more resources, resulting in lack of capacity and thus latency degradation.
+  * Certain types of yet-unknown infra bugs are somehow triggered by a data push.
```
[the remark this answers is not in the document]
FelixGV: Good question... I did not mean that, actually... I meant issues that happen at swap time, such as the issue we had recently where the total size of all schemas exceeded the max response size.

For issues that happen due to ingestion bugs, my understanding is that past issues we've had in that category happened when we turned on some new functionality in a store for the first time (such as zstd compression on a hybrid store) and then did a global push to that store. For that kind of scenario, where test an ingestion-related functionality for the first time, I think we should use the currently-supported flavor of targeted region push (which is not going away as part of this work), and this allows us to ingest in just one region. If the targeted region push succeeds, then we can use that new functionality as part of a global push. If we ramp new ingestion-related functionality using this methodology, we should be able to avoid global outages (at least the kinds we have had in the past)...

docs/proposals/vip-4.md
FelixGV: That is not specified as part of this proposal, but the proposal supports full flexibility, including:

1. Mandatory hooks.
2. Opt-in (disabled by default, can be enabled via the per-store hook config bag).
3. Opt-out (enabled by default, can be disabled via the per-store hook config bag).

docs/proposals/vip-4.md
FelixGV: > Orchestrator can be different from hook executor.

Can you elaborate on what you're thinking here?

> BTW, for the delayed promotion to other regions, I assume we are going to halt the push job and still report the final status in the push job by either abort or rolling forward, right?
> There are some costs with this approach since VPJ executor will take more resources than before, and if we don't halt push job, the final status won't be easily observed by the end user.
> Let us call out the additional cost/impact to the existing components if applicable.

Hmm... I think we would definitely want the VPJ driver/executor (which we run in Azkaban) to stall as long as the hooks are instructing to wait, so that the final status of the job reflects the reality. But the driver job is pretty lightweight from my understanding, so I don't expect the cost of that running longer to be significant... the costly part, from my understanding, is the (MR) data push job but that would not take longer due to hooks.

docs/proposals/vip-4.md
majisourav99: given we would have schema deletion API, if there a risk in deleting bad as a rollback step? I assume some clients might have already refreshed to the latest schema, but given its bad to beging with, might be ok to delete?

FelixGV: Yeah I think it's conceivable to support schema deletion as a hook at some point, though we would need more than just rollback to make it safe... we would need to rollback to the backup version, and wait for the future version we rolled back from to be fully deleted...

Given that we're going in the direction of automatically deleting unused schemas, I don't think we need to involve hooks... we can just let the automation run its course...

FelixGV: FYI, I've added a column in the table to indicate the actions each hook is allowed to return, so that the doc is self-contained and reading the code is less important... I hope that helps make things more clear!

docs/proposals/vip-4.md
gaojieliu: @sixpluszero Can you confirm ^?

docs/proposals/vip-4.md
```diff
@@ -0,0 +1,273 @@
...
+  * ✋ Wait: Let the hooks framework re-run this step later (i.e. 1 minute later, by default).
+  * ↩️ Rollback: Let the store go back to the store-version it had prior to beginning the push job (in all regions). 
+
+* The last three columns are the feasibility of implementing this hook in a given component:
+  * **CC** Child Controller.
+  * **PC** Parent Controller.
+  * **VPJ** Venice Push Job.
+  * ✅ It is feasible to implement this hook within this component (without unreasonable complexity).
+  * ❌ It is NOT feasible to implement this hook within this component (without unreasonable complexity).
+  * 1️⃣ Schema registration hooks in push jobs could only be invoked in cases where auto-registration is enabled and
```
gaojieliu, line 170: Just curious, wouldn't schema registration have the same consequence regardless of the trigger?

docs/proposals/vip-4.md
```diff
@@ -0,0 +1,273 @@
...
+  * ✅ It is feasible to implement this hook within this component (without unreasonable complexity).
+  * ❌ It is NOT feasible to implement this hook within this component (without unreasonable complexity).
+  * 1️⃣ Schema registration hooks in push jobs could only be invoked in cases where auto-registration is enabled and 
+  the new schema originates from the push job itself, whereas schema registrations which are performed directly on the 
+  controller could not trigger the hook.
+  * 2️⃣ The hook for the start of ingestion for Da Vinci Clients is tricky for a few reasons. The start of ingestion is 
+  controlled by updating the Meta Store, which is a non-replicated system store updated by child controllers, so those
+  must be involved (either by running the hook there in the first place, or by having some mechanism that enables the
+  parent or VPJ to inform the child controllers of when the system store is eligible for getting updated, such as by
+  adding a new field to the `AddVersion` admin channel command). However, see Rollback Support below for why running
```
gaojieliu, line 177: Today, DaVinci doesn't support quick rollback as the backup version will be deleted immediately right after the version swap, so are we considering to support fast rollback in DaVinci by keeping two versions in the disk of DaVinci nodes?

docs/proposals/vip-4.md
gaojieliu: The delay in the driver will introduce some cost as each driver is an independent process in Azkaban executor, which pre-allocates some fixed amount of memory for JVM heap, and too many hanging processes might kill the Azkaban executors.
I don't have the data points and these are purely theoretical analysis.

docs/proposals/vip-4.md
```diff
@@ -0,0 +1,273 @@
...
+   DVC ingestion (depending on the result of `preStartOfStoreVersionIngestionForDaVinci`) which the child controller 
+   would honor by holding off on writing to the Meta Store.
+
+2. The parent controller would create the new store-version with `versionSwapDeferred = true`.
+
+3. When the parent controller sees that a region has completed ingestion, it would invoke the `preStoreVersionSwap`
+   hook for that region, and if the hook responds with `PROCEED`, then it would send an admin command targeted for that
+   region to swap the current version.
+
+4. If at any stage the hooks respond with `ROLLBACK` then the parent controller would send more admin channel commands
```
gaojieliu, line 251: When we rollback a store version, we don't rollback the registered schema, right?

internal/venice-common/src/main/java/com/linkedin/venice/controllerapi/JobStatusQueryResponse.java
```diff
@@ -57,10 +59,24 @@ public void setStatusDetails(String statusDetails) {
...
+   * N.B.: Older versions of the controller did not support this timestamp, so this can be null.
+   *
+   * @return the UNIX Epoch timestamp when the value of {@link #getStatus()} was last updated (for child controllers),
+   *         null for parent controllers and for older versions of child controllers.
+   */
+  public Long getStatusUpdateTimestamp() {
+    return this.statusUpdateTimestamp;
+  }
+
+  public void setStatusUpdateTimestamp(Long statusUpdateTimestamp) {
```
gaojieliu, line 72: This timestamp is to track version status update?

internal/venice-common/src/main/java/com/linkedin/venice/controllerapi/JobStatusQueryResponse.java
```diff
@@ -92,6 +108,20 @@ public void setExtraDetails(Map<String, String> extraDetails) {
...
+   * N.B.: Older versions of the controller did not support these timestamps, so this map can be null.
+   *
+   * @return a map of region name -> UNIX Epoch timestamp indicating when the value corresponding to the same key in
+   *         {@link #getExtraInfo()} was last updated, or null if not available.
+   */
+  public Map<String, Long> getExtraInfoUpdateTimestamp() {
+    return this.extraInfoUpdateTimestamp;
+  }
+
+  public void setExtraInfoUpdateTimestamp(Map<String, Long> extraInfoUpdateTimestamp) {
```
gaojieliu, line 121: Can you elaborate what kind of timestamp will be populated into this map?

gaojieliu: I assume some of the hooks would leverage this timestamp to decide the wait time after version swap, right?

internal/venice-common/src/main/java/com/linkedin/venice/controllerapi/JobStatusQueryResponse.java
gaojieliu: After reading the following code,  I understand this timestamp is used to track the timestamp of the latest status update.

internal/venice-common/src/main/java/com/linkedin/venice/controllerapi/JobStatusQueryResponse.java
gaojieliu: And this is for the aggregated results from child regions.

internal/venice-common/src/main/java/com/linkedin/venice/hooks/StoreLifecycleHooks.java
```diff
@@ -0,0 +1,311 @@
...
+   * <br>
+   * N.B.: this hook returns a {@link StoreLifecycleEventOutcome}, and not a {@link StoreVersionLifecycleEventOutcome},
+   * because the new store-version is not created yet at the start of the push job.<br>
+   * <br>
+   * Cardinality: once per push job.
+   */
+  public StoreLifecycleEventOutcome preStartOfPushJob(
+      String clusterName,
+      String storeName,
+      VeniceProperties jobProperties,
```
gaojieliu, line 117: If we decide to execute the hooks in parent controller, do we have a way to grab the push job properties?

docs/proposals/vip-4.md
majisourav99: Thats a great point, given users planning to use business hour version swap thru controller client (probably thats not part of the scope here), having hooks to validate the action would be great. Though we need to have a way to communicate/rollback the action to the user.

docs/proposals/vip-4.md
FelixGV: Yes, the consequences should be the same regardless of the trigger. The point of this line is to say that if we execute the schema registration hooks inside VPJ, then we will have a blindspot by design (which is not ideal...).

docs/proposals/vip-4.md
FelixGV: Fast rollback in DVC is orthogonal to this project (out of scope).

docs/proposals/vip-4.md
FelixGV: Correct. And the schema hook does not provide the ability to rollback.

internal/venice-common/src/main/java/com/linkedin/venice/controllerapi/JobStatusQueryResponse.java
FelixGV: Yeah, the hook could leverage these timestamps to decide if it waited long enough after a region got swapped.

internal/venice-common/src/main/java/com/linkedin/venice/hooks/StoreLifecycleHooks.java
FelixGV: I thought it might be useful for the hook to know the properties, though I'm not sure that we have a concrete use case that must have that as of today.

If we want to keep this parameter here, and execute the hooks in the controller, then we may need to pass the job props to the `/request_topic` endpoint. Alternatively, the job props are in the `PushJobDetails` system store and we may be able to query them out of there (in which case, maybe we should wrap it inside a `Lazy` so that it's only queried if the hook needs it).

gaojieliu: Since there is no known use case requires this, shall we remove it?

docs/proposals/vip-4.md
gaojieliu: Got it.

docs/proposals/vip-4.md
gaojieliu: I understand it is orthogonal, and I was wondering whether we should speed it up because of this project since DVC rollback could take 10+ mins depending on the store size and during that period, there will be prod impact even inside a single colo.

docs/proposals/vip-4.md
gaojieliu: Got it.

And `Schema registration` hook will be invoked regardless of the trigger, right?

FelixGV: Only if the hook is executed in the controller...

internal/venice-common/src/main/java/com/linkedin/venice/hooks/StoreLifecycleHooks.java
FelixGV: Done!

internal/venice-common/src/main/java/com/linkedin/venice/pushmonitor/ExecutionStatusWithDetails.java
```diff
@@ -30,4 +40,14 @@ public String getDetails() {
   public boolean isNoDaVinciStatusReport() {
     return noDaVinciStatusReport;
   }
+
+  /**
+   * N.B.: This is not available in older controller versions, and not available for all statuses (e.g. if the status is
+   * {@link ExecutionStatus#NOT_CREATED}, it does not have this timestamp).
+   *
+   * @return the UNIX Epoch timestamp of the last update to the status, or null if not available.
```
majisourav99, line 48: nit: return some sentinel value for non-existing status instead of null. might save us from NPE in future.

docs/proposals/vip-4.md
majisourav99: So the reasoning here that increasing wait time on pushes might cause too many Azkaban jobs to run simultaneously and kill the executors ? If thats the case, then random slowness in vpj or fabric connection issues would cause same thing, which we do not see that often.

internal/venice-common/src/main/java/com/linkedin/venice/hooks/StoreLifecycleHooks.java
```diff
@@ -0,0 +1,304 @@
...
+
+  /**
+   * Invoked prior to starting a new job. The hook has the option of aborting the job.<br>
+   * <br>
+   * N.B.: this hook returns a {@link StoreLifecycleEventOutcome}, and not a {@link StoreVersionLifecycleEventOutcome},
+   * because the new store-version is not created yet at the start of the push job.<br>
+   * <br>
+   * Cardinality: once per push job.
+   */
+  public StoreLifecycleEventOutcome preStartOfPushJob(
```
majisourav99, line 114: Just thinking out aloud, what cases pre/post start PushJobHook would we want abort the job?

internal/venice-common/src/main/java/com/linkedin/venice/pushmonitor/ExecutionStatusWithDetails.java
FelixGV: We could, although mis-handling a sentinel value is still a bug, even though it does not cause a NPE. IMHO the main benefit of sentinel value instead of boxed/nullable types is to limit memory usage, but in this case I think that doesn't matter... It could also be an Optional, though that also does not prevent bugs in my opinion. BTW, do you know how does Jackson handle new fields getting added? Is there a way to set the default value when dealing with payloads encoded by older code (i.e. which are missing the new field)? If it is a boxed type, then I'm pretty sure a missing field is going to be null, but if it's a primitive type, IDK...

internal/venice-common/src/main/java/com/linkedin/venice/hooks/StoreLifecycleHooks.java
FelixGV: Sorry, forgot to answer that. I think we don't have a concrete use case for this hook at the moment. It is included mostly for symmetry, since all the other hooks have pre and post. But we could exclude it from scope of the MVP if need be.

majisourav99 (approved):
Thank you Felix!

FelixGV:
> Thank you Felix!

Thanks @majisourav99 !

PR #882, opened 2024-02-28
[vpj] Handle extra null default in ETL+VPJ value schema

nisargthakkar:
## Handle extra `null` default in ETL+VPJ value schema

Currently, ETL+VPJ flow makes strict assumptions about schema. It doesn't allow the value schema to have `null` as a default value if the original value schema didn't have a `null` default. This commit allows schemas where the `value` field has a `null` default.

## How was this PR tested?

Added unit tests, integration tests

### DIFF
--- a/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/input/recordreader/avro/AbstractAvroRecordReader.java
+++ b/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/input/recordreader/avro/AbstractAvroRecordReader.java
@@ -1,6 +1,7 @@
 package com.linkedin.venice.hadoop.input.recordreader.avro;
 
 import com.linkedin.avroutil1.compatibility.AvroCompatibilityHelper;
+import com.linkedin.avroutil1.compatibility.AvroSchemaUtil;
 import com.linkedin.venice.etl.ETLUtils;
 import com.linkedin.venice.etl.ETLValueSchemaTransformation;
 import com.linkedin.venice.exceptions.VeniceException;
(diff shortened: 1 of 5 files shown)


clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/input/recordreader/avro/AbstractAvroRecordReader.java
```diff
@@ -60,7 +61,21 @@ public AbstractAvroRecordReader(
         if (fileField.name().equals(valueFieldStr)) {
           fieldSchema = ETLUtils.getValueSchemaFromETLValueSchema(fieldSchema, etlValueSchemaTransformation);
         }
-        outputSchemaFields.add(AvroCompatibilityHelper.newField(fileField).setSchema(fieldSchema).build());
+        // Since we are stripping the schema of the value field of the "null" type, we need to handle the case where the
+        // default value of the schema is null and the schema is not nullable. In this case, we create a schema without
```
ZacAttack, line 65: I'm not sure I understand this one.  So we stripped out the null type, but this right now chokes if the default value is null?  Is that right?

ZacAttack (approved):
Looks ok, though I'm a little confused on the premise

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/input/recordreader/avro/AbstractAvroRecordReader.java
nisargthakkar: Yes, exactly. The problem arises because when we do `AvroCompatibilityHelper.newField(fileField)`, then it copies all properties of `fileField`

PR #883, opened 2024-02-29
[all] Improve debugging experience

sushantmane:
## Improve debugging experience

- Adjust the indexes of threads created with DaemonThreadFactory to use
  zero-based indexing. This change simplifies mapping of threads with
  drainer queues and consumers by eliminating the need for additional mental
  calculations.
- Override the `ImmutablePubSubMessage::toString` method to include additional
  information such as topic-partition and offset in the message output.

### DIFF
--- a/internal/venice-common/src/main/java/com/linkedin/venice/pubsub/ImmutablePubSubMessage.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/pubsub/ImmutablePubSubMessage.java
@@ -82,4 +82,9 @@ public class ImmutablePubSubMessage<K, V> implements PubSubMessage<K, V, Long> {
   public PubSubMessageHeaders getPubSubMessageHeaders() {
     return pubSubMessageHeaders;
   }
+
+  @Override
+  public String toString() {
+    return "PubSubMessage{" + topicPartition + ", offset=" + offset + ", timestamp=" + timestamp + '}';
+  }
 }
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/ImmutableChangeCapturePubSubMessage.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/ImmutableChangeCapturePubSubMessage.java
@@ -69,4 +69,10 @@ public class ImmutableChangeCapturePubSubMessage<K, V> implements PubSubMessage<
   public boolean isEndOfBootstrap() {
     return isEndOfBootstrap;
   }
+
+  @Override
+  public String toString() {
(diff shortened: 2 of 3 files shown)


sixpluszero (dismissed):

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/ImmutablePubSubMessage.java
```diff
@@ -82,4 +82,9 @@ public boolean isEndOfBootstrap() {
   public PubSubMessageHeaders getPubSubMessageHeaders() {
     return pubSubMessageHeaders;
   }
+
+  @Override
+  public String toString() {
+    return "PubSubMessage{topicPartition=" + topicPartition + ", offset=" + offset + ", timestamp=" + timestamp + '}';
```
haoxu07, line 88: Nit: as PubSubTopicPartition already has format like this: TP(topic: '" + topic.getName() + "', partition: " + partitionNumber + "), maybe we can get rid of 'topicPartition='. Besides, do you think adding payloadSize and boolean isEndOfBootstrap will help debugging?

sushantmane: Good catch. Thanks. Updated. 

>Besides, do you think adding payloadSize and boolean isEndOfBootstrap will help debugging?
IMO this is not very helpful.

sixpluszero: Yeah, and what does isEndOfbootstrap means? I see it always returns false.

haoxu07 (approved):
Thank you very much for this improvement. The thread id starting with 0 do helps a lot on checking log and heapdum.

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/ImmutablePubSubMessage.java
sushantmane: I think it's used in CC context but I don't know it's significance

sushantmane:
> The thread id starting with 0 do helps a lot on checking log and heapdump.

Yeah fully agree.

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/ImmutablePubSubMessage.java
sushantmane: Added `ImmutableChangeCapturePubSubMessage::toString` which logs `isEoB` as well

sixpluszero (approved):

sushantmane:
Thanks @sixpluszero and @haoxu07 for the review!

PR #884, opened 2024-02-29
[all] Fix an infinite recursion issue in VeniceWriter (#879)

lluwm:
* [all] Fix an infinite recursion issue in VeniceWriter

Today, VeniceWriter (VW) has a bug to run into an infinite recursion when it sends a message in the VeniceWriter and eventually it throws StackOverflowError. For a particular partition, the conditions to trigger the issue are:

1. VW has a cached Segment
2. The segment is neither started nor ended
3. The elapsed time for the segment is greater than MAX_ELAPSED_TIME_FOR_SEGMENT_IN_MS

To fix it, we break the infinite loop by always recording the new start time. A new integration test, as well as a unit test, is added to capture any regressions.

Resolves #XXX

### DIFF
--- a/internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
@@ -238,10 +238,10 @@ public class VeniceWriter<K, V, U> extends AbstractVeniceWriter<K, V, U> {
    */
   private final Segment[] segments;
   /**
-   * Map of partition to its segment creation time in milliseconds.
+   * Map of partition to its segment start time in milliseconds.
    * -1: the current segment is ended
    */
-  private final long[] segmentsCreationTimeArray;
+  private final long[] segmentsStartTimeArray;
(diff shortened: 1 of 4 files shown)


gaojieliu (approved):

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
```diff
@@ -1972,20 +1972,18 @@ private Segment getSegment(int partition, boolean sendEndOfSegment) {
   private Segment startSegment(int partition) {
     synchronized (this.partitionLocks[partition]) {
       Segment currentSegment = segments[partition];
-      long segmentStartTime = segmentsCreationTimeArray[partition];
 
       if (currentSegment == null) {
         currentSegment = new Segment(partition, 0, checkSumType);
-        segmentStartTime = time.getMilliseconds();
```
ZacAttack, line 1979: What the heck did these even do in the first place?

ZacAttack (approved):
Looks good!

lluwm:
Thanks for the review! Notice that this is a clean merge without a need to resolve any conflict. Proceed to check in.

PR #885, opened 2024-03-01
[davinci][server] Add more write-path metrics for observability

adamxchen:

This PR adds some metrics on the write path to provide fine-granular view, specifically:
1. Add `leader_div_completion_latency` to measure how long it takes for a leader node begins to process each record pulled from pubsub system, e.g. kafka, to the time it finishes DIV(date ingestion validation).
2. Add `internal_process_latency`, which measures how long for a drainer to process a record internally and `process_record_error`, which count the number of errors happen for internal processing in `StoreBufferService`(SBS). The term "internal process" refers to the process nodes persist records on the local disk.
3. Add `storage_engine_delete_latency`, which measures the long to delete a record from store engine.
4. Remove `memory_usage_for_writer_num_` metrics for SBS since they are not particularly helpful for debugging. We used to use them to identify drainer skew issue tho.
5. Remove `producer_to_follower_consumer_latency` and `producer_to_leader_consumer_latency` since they can be deduced using existing metrics
6. Removed `Total` for `consumer_poll_result_num`
7. Added `producer_callback_latency` to measure the total time spent in the producer callback.

## How was this PR tested?

GHCI

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreBufferService.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreBufferService.java
@@ -83,6 +83,25 @@ public class StoreBufferService extends AbstractStoreBufferService {
         this::getMinMemoryUsagePerDrainer);
   }
 
+  /**
+   * Constructor for testing
+   */
+  public StoreBufferService(
+      int drainerNum,
+      long bufferCapacityPerDrainer,
+      long bufferNotifyDelta,
+      boolean queueLeaderWrites,
+      StoreBufferServiceStats stats) {
+    this.drainerNum = drainerNum;
+    this.blockingQueueArr = new ArrayList<>();
+    this.bufferCapacityPerDrainer = bufferCapacityPerDrainer;
+    for (int cur = 0; cur < drainerNum; ++cur) {
+      this.blockingQueueArr.add(new MemoryBoundBlockingQueue<>(bufferCapacityPerDrainer, bufferNotifyDelta));
+    }
+    this.leaderRecordHandler = queueLeaderWrites ? this::queueLeaderRecord : StoreBufferService::processRecord;
+    this.storeBufferServiceStats = stats;
+  }
+
   protected MemoryBoundBlockingQueue<QueueNode> getDrainerForConsumerRecord(
       PubSubMessage<KafkaKey, KafkaMessageEnvelope, Long> consumerRecord,
       int subPartition) {
@@ -309,7 +328,6 @@ public class StoreBufferService extends AbstractStoreBufferService {
             .sorted(comparing(Map.Entry::getValue, reverseOrder()))
             .limit(count)
             .collect(toList());
-        // TODO: break this done so we can let it emit latency metrics too.
         int finalIndex = index;
         slowestEntries.forEach(
             entry -> LOGGER
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreBufferService.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreBufferService.java
@@ -4,6 +4,7 @@ import static java.util.Collections.reverseOrder;
 import static java.util.Comparator.comparing;
 import static java.util.stream.Collectors.toList;
 
+import com.linkedin.davinci.stats.StoreBufferServiceStats;
 import com.linkedin.venice.common.Measurable;
 import com.linkedin.venice.exceptions.VeniceChecksumException;
 import com.linkedin.venice.exceptions.VeniceException;
@@ -16,6 +17,7 @@ import com.linkedin.venice.pubsub.api.PubSubMessage;
 import com.linkedin.venice.pubsub.api.PubSubTopicPartition;
 import com.linkedin.venice.utils.DaemonThreadFactory;
 import com.linkedin.venice.utils.PartitionUtils;
+import io.tehuti.metrics.MetricsRepository;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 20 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreBufferService.java
adamxchen, line 1: I moved the nested class from top to bottom for readability.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreBufferService.java
```diff
@@ -472,138 +201,421 @@ public void drainBufferedRecordsFromTopicPartition(PubSubTopicPartition topicPar
...
+      // print drainer info when there is a slow drainer.
+      if (slowDrainerExists) {
+        MemoryBoundBlockingQueue<QueueNode> queue = blockingQueueArr.get(index);
+        int count = queue.getMemoryUsage() > 0.8 * bufferCapacityPerDrainer ? 5 : 1;
+        List<Map.Entry<PubSubTopicPartition, Long>> slowestEntries = drainer.topicToTimeSpent.entrySet()
+            .stream()
+            .sorted(comparing(Map.Entry::getValue, reverseOrder()))
+            .limit(count)
+            .collect(toList());
+        // TODO: break this done so we can let it emit latency metrics too.
```
xunyin8, line 312: NIT: typo in the comment but also wondering if the existence of a slow drainer itself is worth a metric.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreBufferService.java
```diff
@@ -472,138 +201,421 @@ public void drainBufferedRecordsFromTopicPartition(PubSubTopicPartition topicPar
...
+              node.getBeforeProcessingRecordTimestampNs());
+
+          /**
+           * Complete {@link QueueNode#queuedRecordPersistedFuture} since the processing for the current record is done.
+           */
+          if (recordPersistedFuture != null) {
+            recordPersistedFuture.complete(null);
+          }
+          long latencyInMS = System.currentTimeMillis() - startTime;
+          this.stats.recordInternalProcessingLatency(latencyInMS);
```
xunyin8, line 564: This might be a dumb question but what do we use `internalProcessingLatency` for? The metric seems to cover a variety of behaviors since `processRecord` can be called on control messages and data records. The expected latency can be very different for different messages. Especially if message is rejected/skipped by DIV the latency will be really low. Basically not sure how this metric can be useful when any value could be either good or bad.

xunyin8 (commented):
Mostly looks good, have some concerns on internal process latency metric.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreBufferService.java
adamxchen: This is a old comment since I was moving some codes around... Let me remove this.
To your questions, no I don't think we need to add a specific metric here as we are already adding some processing latency metrics at the store-level and having them at the drainer-level is a duplicate effort.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreBufferService.java
adamxchen: Good question! It's true that this metric doesn't differentiate records and we probably cannot tell whether slowness happens in which behavior (in that case, I think we need to add a new metric for that behavior or check different stats, e.g. `min` for control message 
 or `max` for large data record). However, it gives us the insight on how fast a drainer can process the record. Right now, we don't have the data and existing end-to-end metrics include the drainer queueing time which makes them unreliable to understand the internal processing latency. With this metric, we can also deduce how long a record is queued in the drainer roughly..

xunyin8: I see, but data record's process latency should mostly be storage engine put latency which we already have a metric for. Given that this is not a store level metric I guess it's fine to add it and see the behavior.

clients/da-vinci-client/src/test/java/com/linkedin/davinci/kafka/consumer/StoreBufferServiceTest.java
```diff
@@ -213,7 +224,7 @@ public void testGetDrainerIndexForConsumerRecord(boolean queueLeaderWrites) {
 
   @Test(dataProviderClass = DataProviderUtils.class, dataProvider = "True-and-False")
   public void testRunWhenThrowVeniceCheckSumFailException(boolean queueLeaderWrites) throws Exception {
```
xunyin8, line 226: Do we not want to verify any of the new metrics in this test class? e.g. `recordInternalProcessingError`

services/venice-server/src/test/java/com/linkedin/venice/stats/AggVersionedDIVStatsTest.java
```diff
@@ -108,43 +108,24 @@ public void testStatsCanUpdateVersionStatus() {
...
 
     double v1ProducerToSourceBrokerLatencyMs = 811d;
     double v1SourceBrokerToLeaderConsumerLatencyMs = 211d;
-    double v1ProducerToLeaderConsumerLatencyMs = 1011d;
     stats.recordLeaderLatencies(
         storeName,
         1,
         consumerTimestampMs,
         v1ProducerToSourceBrokerLatencyMs,
-        v1SourceBrokerToLeaderConsumerLatencyMs,
```
xunyin8, line 121: Similar here, any verification for the newly added `leader_div_completion_latency`?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreBufferService.java
adamxchen: yes, I agree that storage engine put would be the major contributor. As we potentially add more codes to the internal processing path, this new metric gives some insights on how other code paths perform.

clients/da-vinci-client/src/test/java/com/linkedin/davinci/kafka/consumer/StoreBufferServiceTest.java
adamxchen: IMO, we don't have to add many tests for `stats` and usually they get tested as a part of business logic. I will make sure they are emitted correctly after deploying to per-production. That being said, I've added some tests for SBS stats. Please take a look @xunyin8

services/venice-server/src/test/java/com/linkedin/venice/stats/AggVersionedDIVStatsTest.java
adamxchen: please see my reply above

clients/da-vinci-client/src/test/java/com/linkedin/davinci/kafka/consumer/StoreBufferServiceTest.java
xunyin8: Right, we don't have to add **many** tests for stats but we should at least have one smoke test and most of the time it's not to verify the new stats are added correctly but prevent regressions that break stats. We've had numerous incidents where a change accidentally broke some stats.

xunyin8 (approved):

clients/da-vinci-client/src/test/java/com/linkedin/davinci/kafka/consumer/StoreBufferServiceTest.java
adamxchen: Acked. I will see if I can add more on this in my next PR. The SBS is easy to add but for SIT, something was not working for me so I decided to defer it.

adamxchen: I still couldn't get it working.. Let me merge this one for now and we can add it later.

PR #886, opened 2024-03-02
[vpj] Allow enabling compression metric collection without using mapper to build dictionary

nisargthakkar:
## Allow enabling compression metric collection without using mapper to build dictionary

Currently, `use.mapper.to.build.dictionary` is auto enabled when `compression.metric.collection.enabled` is enabled. This commit separates the two configs for migration purposes and `use.mapper.to.build.dictionary` will not be auto-enabled when `compression.metric.collection.enabled` is enabled and it now needs to be explicitly enabled. Eventually, `use.mapper.to.build.dictionary` will be removed along with the `ValidateSchemaAndBuildDict` job.

Operationally, we have noticed that often, the VPJ jobs wait in the queue to acquire resources from Yarn despite the job only requesting a container for one map task and no reduce tasks. When abstracting the compute engines used by VPJ, we decided to not abstract the `ValidateSchemaAndBuildDict` job. We decided this because the initial reasons to add a dedicated job are no longer valid:
1. Zstd dictionary creation would sometimes [crash the JVM](https://github.com/luben/zstd-jni/issues/253)
    * This issue has now been fixed in the library; and
    *  We also have guardrails for this in our code
2. Ability to build larger dictionaries: 
    * We believe the VPJ executors can have enough memory to allow them to build good enough dictionaries
    * At LinkedIn, we have found that dictionaries of size <150 KB and using 200 MB of samples offer great compression ratios. (Upto 40X in some cases). And for this, our executors only need 256 MB of memory.

## How was this PR tested?

Updated unit tests. GH CI runs all tests. Will test on test jobs soon.

    - `use.mapper.to.build.dictionary` will not be auto-enabled when `compression.metric.collection.enabled` is enabled and it now needs to be explicitly enabled
    - This is okay because this piece only offers analytics around optimization and there are no users of this functionality other than ourselves

### DIFF
--- a/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/DefaultInputDataInfoProvider.java
+++ b/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/DefaultInputDataInfoProvider.java
@@ -266,7 +266,8 @@ public class DefaultInputDataInfoProvider implements InputDataInfoProvider {
         if (!pushJobSetting.isIncrementalPush) {
           if (!pushJobSetting.useMapperToBuildDict) {
             /** If dictionary compression is enabled for version, read the records to get training samples */
-            if (pushJobSetting.storeCompressionStrategy == CompressionStrategy.ZSTD_WITH_DICT) {
+            if (pushJobSetting.storeCompressionStrategy == CompressionStrategy.ZSTD_WITH_DICT
+                || pushJobSetting.compressionMetricCollectionEnabled) {
               InputDataInfoProvider.loadZstdTrainingSamples(fileIterator, pushJobZstdConfig);
             }
           } else {
--- a/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/DefaultInputDataInfoProvider.java
+++ b/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/DefaultInputDataInfoProvider.java
@@ -87,8 +87,8 @@ public class DefaultInputDataInfoProvider implements InputDataInfoProvider {
     }
 
     if (!pushJobSetting.isIncrementalPush && !pushJobSetting.useMapperToBuildDict) {
-      if (this.pushJobSetting.storeCompressionStrategy == CompressionStrategy.ZSTD_WITH_DICT) {
-        LOGGER.info("Zstd compression enabled for {}", pushJobSetting.storeName);
+      if (pushJobSetting.storeCompressionStrategy == CompressionStrategy.ZSTD_WITH_DICT
+          || pushJobSetting.compressionMetricCollectionEnabled) {
         initZstdConfig(fileStatuses.length);
       }
     }
@@ -294,7 +294,7 @@ public class DefaultInputDataInfoProvider implements InputDataInfoProvider {
   }
 
   @Override
-  public byte[] getZstdDictTrainSamples() {
+  public byte[] trainZstdDictionary() {
     return pushJobZstdConfig.getZstdDictTrainer().trainSamples();
   }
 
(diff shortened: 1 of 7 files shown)


m-nagarajan:
Before we disable/remove this, can we increase sample and dict sizes to get some numbers?

nisargthakkar:
> Before we disable/remove this, can we increase sample and dict sizes to get some numbers? 

This change doesn't reduce the scope of the feature. The dict building will eventually only run on the driver instead of a dedicated compute job. The spark driver has more compute and memory than the driver that we have for Map-Reduce. So, we can still collect these analytics at a later time when needed

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/DefaultInputDataInfoProvider.java
```diff
@@ -357,7 +357,8 @@ protected Pair<Schema, Schema> getAvroFileHeader(FileSystem fs, Path path, boole
         if (!pushJobSetting.isIncrementalPush) {
           if (!pushJobSetting.useMapperToBuildDict) {
             /** If dictionary compression is enabled for version, read the records to get training samples */
-            if (pushJobSetting.storeCompressionStrategy == CompressionStrategy.ZSTD_WITH_DICT) {
+            if (pushJobSetting.storeCompressionStrategy == CompressionStrategy.ZSTD_WITH_DICT
+                || pushJobSetting.compressionMetricCollectionEnabled) {
```
m-nagarajan, line 361: similar check needed in `getMetadataFromSequenceFile`

m-nagarajan (changes requested):
Thanks @nisargthakkar. Left a comment.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/DefaultInputDataInfoProvider.java
nisargthakkar: Hmm, interesting. Good catch. Wonder why the test didn't catch this

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
```diff
@@ -1120,13 +1127,17 @@ protected static boolean shouldBuildZstdCompressionDictionary(
       return false;
     }
 
+    if (pushJobSetting.isIncrementalPush) {
+      LOGGER.info("No compression dictionary will be generated as the push type is incremental push");
+      return false;
+    }
+
+    if (!inputFileHasRecords) {
```
m-nagarajan, line 1135: This could be a duplicate log as we check and log it later in the function as well.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/DefaultInputDataInfoProvider.java
```diff
@@ -259,27 +260,13 @@ private Map<String, String> getMetadataFromSequenceFile(
       FileSystem fs,
       Path path,
       boolean isZstdDictCreationRequired) {
-    LOGGER.debug("path:{}", path.toUri().getPath());
     VeniceVsonRecordReader recordReader = getVeniceVsonRecordReader(fs, path);
-    try (VeniceVsonFileIterator fileIterator = new VeniceVsonFileIterator(fs, path, recordReader)) {
```
m-nagarajan, line 264: Thanks for cleaning these duplicate checks.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/DefaultInputDataInfoProvider.java
```diff
@@ -296,6 +283,17 @@ private VeniceVsonFileIterator getVeniceVsonFileIterator(FileSystem fs, Path pat
 
   @Override
   public byte[] trainZstdDictionary() {
+    int collectedNumberOfSamples = pushJobZstdConfig.getCollectedNumberOfSamples();
```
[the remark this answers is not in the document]
nisargthakkar: I didn't move that because it also emits a metric, and I think we would like to have that metric till we keep that job around. With this change, there is no way to signal the caller if the dict is based on data, or dummy data

m-nagarajan (approved):
LGTM, Thanks @nisargthakkar 🚀

nisargthakkar:
Thanks for the review, @m-nagarajan

PR #888, opened 2024-03-04
[server] Do not treat quota rejection towards error request metric

majisourav99:
## Do not treat quota rejection towards error request metric

Quota rejection should not be counted as error request, which triggers unnecessary alerts on Venice. it’s “success” as in Venice successfully reported the “quota rejection” to client.

## How was this PR tested?

integ test


services/venice-server/src/main/java/com/linkedin/venice/listener/StatsHandler.java
```diff
@@ -204,15 +205,15 @@ public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise)
             : serverStatsContext.getCurrentStats().getStoreStats(serverStatsContext.getStoreName());
         serverStatsContext.recordBasicMetrics(serverHttpRequestStats);
         double elapsedTime = LatencyUtils.getLatencyInMS(serverStatsContext.getRequestStartTimeInNS());
-        // if ResponseStatus is either OK or NOT_FOUND and the channel write is succeed,
+        // if ResponseStatus is either OK or NOT_FOUND or TOO_MANY_REQUESTS and the channel write is succeed,
         // records a successRequest in stats. Otherwise, records a errorRequest in stats;
         if (result.isSuccess() && (serverStatsContext.getResponseStatus().equals(OK)
-            || serverStatsContext.getResponseStatus().equals(NOT_FOUND))) {
+            || serverStatsContext.getResponseStatus().equals(NOT_FOUND)
+            || serverStatsContext.getResponseStatus().equals(TOO_MANY_REQUESTS))) {
```
xunyin8, line 212: I'm wondering if we should call `successRequest` for `TOO_MANY_REQUESTS`. In today's router we still emit too many requests as unhealthy: https://github.com/linkedin/venice/blob/fbcace8e57302661a81ca85f698e370a216f134b/services/venice-router/src/main/java/com/linkedin/venice/router/api/RouterExceptionAndTrackingUtils.java#L142

I understand that we shouldn't call `errorRequest` since `TOO_MANY_REQUESTS` is an user error and shouldn't trigger the same alert that makes Venice devs believe something is wrong with the server. However, I'm concerned that calling `successRequest` on `TOO_MANY_REQUESTS` may unintentionally obfuscate our success metric. Perhaps we should simply do nothing.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/fastclient/FastClientIndividualFeatureConfigurationTest.java
```diff
@@ -125,11 +127,17 @@ public void testServerReadQuota() throws Exception {
       assertTrue(clientException.getMessage().contains("VeniceClientRateExceededException"));
     }
     boolean readQuotaRejected = false;
+    boolean errorRequest = false;
+
     for (MetricsRepository serverMetric: serverMetrics) {
       if (serverMetric.getMetric(readQuotaRejectedString).value() > 0) {
         readQuotaRejected = true;
       }
+      if (serverMetric.getMetric(errorRequestString).value() > 0) {
```
xunyin8, line 136: Can we also add an unit test for `StatsHandler.write` if possible for the different `getResponseStatus`?

services/venice-server/src/main/java/com/linkedin/venice/listener/StatsHandler.java
majisourav99: As discussed offline, 429 is not counted against either success or error requests.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/fastclient/FastClientIndividualFeatureConfigurationTest.java
majisourav99: Looks like to unit test we need to create dummy ctx which needs too many boilterplace class method overrides. Probably I will need to refactor the method to cleanly write unit test, which I am keeping outside the scope of this critical PR.

xunyin8 (approved):
Thanks LGTM!

PR #889, opened 2024-03-06
[controller] Parent controller allows updating the current version in a special UpdateStore call

nisargthakkar:
## Parent controller allows updating the current version in a special UpdateStore call

In PR #862, we prevented parent controllers from updating the current version in Zk. However, we added a provision to allow the current version to still be updated by operators in a restricted mode. However, this custom upgrade doesn't work as it too gets skipped due to certain conditional checks. This commit fixes that issue.

## How was this PR tested?

GH CI

### DIFF
--- a/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
@@ -4258,10 +4258,12 @@ public class VeniceHelixAdmin implements Admin, StoreCleaner {
   }
 
   private void internalUpdateStore(String clusterName, String storeName, UpdateStoreQueryParams params) {
-    /**
-     * Check whether the command affects this region.
-     */
+    // There are certain configs that are only allowed to be updated in child regions. We might still want the ability
+    // to update such configs in the parent region via the Admin tool for operational reasons. So, we allow such updates
+    // if the regions filter only specifies one region, which is the parent region.
     boolean onlyParentRegionFilter = false;
+
+    // Check whether the command affects this region.
     if (params.getRegionsFilter().isPresent()) {
       Set<String> regionsFilter = parseRegionsFilterList(params.getRegionsFilter().get());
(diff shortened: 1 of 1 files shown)


services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -4394,7 +4410,7 @@ private void internalUpdateStore(String clusterName, String storeName, UpdateSto
 
       if (currentVersion.isPresent()) {
         if (childRegionOnlyConfigUpdateAllowed) {
```
majisourav99, line 4412: can we reuse the variable `onlyParentRegionFilter` ? Also simply the logic something  like

`if (currentVersion.isPresent()) {
        setStoreCurrentVersion(clusterName, storeName, currentVersion.get(), onlyParentRegionFilter);
 }`

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -3718,9 +3718,22 @@ Pair<Store, Version> waitVersion(String clusterName, String storeName, int versi
...
+  private void setStoreCurrentVersion(
+      String clusterName,
+      String storeName,
+      int versionNumber,
+      boolean allowedInParent) {
     storeMetadataUpdate(clusterName, storeName, store -> {
-      if (isParent()) {
+      if (isParent() && !allowedInParent) {
         // Parent colo should not update the current version of a store
+        LOGGER.info("Updating current version in parent region is not allowed. Skipping.");
```
majisourav99, line 3736: And add the logging here. Also never add logs without any identifying features, like store/cluster names.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
nisargthakkar: Makes sense. Updated

majisourav99 (approved):
Thanks Nisarg

PR #890, opened 2024-03-07
[server] Do not collect stats for metadata requests

majisourav99:
## Do not collect stats for metadata requests

Metadata requests for otherwise unused stores triggers metric collection, which for clusters with high number of stores creates memory pressure. This PR skips metric recording for metadata requests which includes `current_version/health and metatdata` calls.

## How was this PR tested?

GHCI


xunyin8 (approved):
Thanks Sourav, it'd be great if we had tests to verify these filters but looks like it involves the `StatsHandler`class again which could be difficult without some refactoring.

majisourav99:
> Thanks Sourav, it'd be great if we had tests to verify these filters but looks like it involves the `StatsHandler`class again which could be difficult without some refactoring.

Yeah! that class came to bite us again.

PR #891, opened 2024-03-07
[vpj] Fix deterministic GUID generation for VPJ

nisargthakkar:
## [vpj] Fix deterministic GUID generation for VPJ

Recently, when abstracting compute engines, the logic to compute a deterministic GUID was changed to be less reliant on the underlying compute engine. However, this change had a bug where the uniqueness of the GUID was tied to the uniqueness of the `jobId` used to trigger the `VenicePushJob`. This commit generates a unique GUID per VPJ job.

## How was this PR tested?

GH CI. Updated an integration test

### DIFF
--- a/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/mapreduce/datawriter/jobs/DataWriterMRJob.java
+++ b/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/mapreduce/datawriter/jobs/DataWriterMRJob.java
@@ -206,7 +206,8 @@ public class DataWriterMRJob extends DataWriterComputeJob {
         props.getString(ZSTD_COMPRESSION_LEVEL, String.valueOf(Zstd.maxCompressionLevel())));
     conf.setBoolean(ZSTD_DICTIONARY_CREATION_SUCCESS, pushJobSetting.isZstdDictCreationSuccess);
 
-    // Deterministic producerGUID generator
+    // We generate a random UUID once, and the tasks of the compute job can use this to build the same producerGUID
+    // deterministically.
     UUID producerGuid = UUID.randomUUID();
     conf.setLong(PUSH_JOB_GUID_MOST_SIGNIFICANT_BITS, producerGuid.getMostSignificantBits());
     conf.setLong(PUSH_JOB_GUID_LEAST_SIGNIFICANT_BITS, producerGuid.getLeastSignificantBits());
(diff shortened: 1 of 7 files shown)


clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/mapreduce/datawriter/jobs/DataWriterMRJob.java
```diff
@@ -203,6 +206,11 @@ private void setupDefaultJobConf(JobConf conf, PushJobSetting pushJobSetting, Ve
         props.getString(ZSTD_COMPRESSION_LEVEL, String.valueOf(Zstd.maxCompressionLevel())));
     conf.setBoolean(ZSTD_DICTIONARY_CREATION_SUCCESS, pushJobSetting.isZstdDictCreationSuccess);
 
+    // Deterministic producerGUID generator
+    UUID producerGuid = UUID.randomUUID();
+    conf.setLong(PUSH_JOB_GUID_MOST_SIGNIFICANT_BITS, producerGuid.getMostSignificantBits());
+    conf.setLong(PUSH_JOB_GUID_LEAST_SIGNIFICANT_BITS, producerGuid.getLeastSignificantBits());
```
FelixGV, line 212: I don't understand this... random is not deterministic, or is it?

FelixGV (commented):
The goal of having a deterministic GUID generation was to support MR speculative execution. Can we ensure that we maintain this support across engines?

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/mapreduce/datawriter/jobs/DataWriterMRJob.java
nisargthakkar: It is random - so, it helps to guarantee it to be unique across different VPJ jobs. But because we generate this random UUID once, and use that during the execution, it will be "deterministic" for all tasks of that compute job.

nisargthakkar: I updated the comments around this to make it more explanatory

internal/venice-common/src/main/java/com/linkedin/venice/guid/GuidUtils.java
```diff
@@ -31,25 +34,17 @@ private static GuidGenerator getGuidGenerator(VeniceProperties properties) {
...
-        guidGeneratorPrefix = 0;
-      } else {
-        guidGeneratorPrefix = jobId.hashCode();
-      }
-
       guidGenerator = new DeterministicGuidGenerator(
-          guidGeneratorPrefix,
-          properties.getLong(ConfigKeys.PUSH_JOB_COMPUTE_TASK_ID, 0L));
+          properties.getLong(PUSH_JOB_GUID_MOST_SIGNIFICANT_BITS, 0L),
+          properties.getLong(PUSH_JOB_GUID_LEAST_SIGNIFICANT_BITS, 0L));
```
FelixGV, line 39: Using `0L` as a default seems wrong... I think we should fail instead if the required config is not found.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/mapreduce/datawriter/jobs/DataWriterMRJob.java
FelixGV: Thanks, the improved comment is good.

internal/venice-common/src/main/java/com/linkedin/venice/guid/GuidUtils.java
nisargthakkar: That makes sense. I'll update it. Probably also need to update some tests

FelixGV (approved):
Thanks! LGTM!

nisargthakkar:
Thanks @FelixGV !!

PR #892, opened 2024-03-08
[fast-client] Complete error retry for fast client multi-get

xunyin8:
## [fast-client] Complete error retry for fast client multi-get
1. Error retry was already covered by long tail retry. This means we have error retry at the request level and not at the per route level. This means keys that are not done yet by the long tail retry threshold time will be retried regardless if it's due to latency or error. We have unit tests for various scenarios to complete the table: original request | retry request
fail             | fail
fail             | success
late success     | fail
late success     | success

2. Added the ability to differentiate 429 from other errors so retry is skipped altogether if we encountered 429 (assuming subsequent retry will also be rejected).

## How was this PR tested?
Unit and integration tests

### DIFF
--- a/clients/venice-client/src/main/java/com/linkedin/venice/fastclient/RetriableAvroGenericStoreClient.java
+++ b/clients/venice-client/src/main/java/com/linkedin/venice/fastclient/RetriableAvroGenericStoreClient.java
@@ -110,8 +110,14 @@ public class RetriableAvroGenericStoreClient<K, V> extends DelegatingAvroStoreCl
     final CompletableFuture<V> retryFuture = new CompletableFuture<>();
     final CompletableFuture<V> finalFuture = new CompletableFuture<>();
 
+    AtomicReference<Throwable> savedException = new AtomicReference<>();
     // create a retry task
     Runnable retryTask = () -> {
+      if (savedException.get() != null && isExceptionCausedByTooManyRequests(savedException.get())) {
+        // Defensive code, abort retry if original request failed due to 429
+        retryFuture.completeExceptionally(savedException.get());
+        return;
+      }
       super.get(requestContext, key).whenComplete((value, throwable) -> {
         if (throwable != null) {
           retryFuture.completeExceptionally(throwable);
@@ -146,6 +152,7 @@ public class RetriableAvroGenericStoreClient<K, V> extends DelegatingAvroStoreCl
       } else {
         // Trigger the retry right away when receiving any error that's not a 429 otherwise try to cancel any scheduled
         // retry
+        savedException.set(throwable);
         if (!timeoutFuture.isDone()) {
           timeoutFuture.cancel();
           if (!isExceptionCausedByTooManyRequests(throwable)) {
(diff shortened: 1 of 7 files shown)


clients/venice-client/src/main/java/com/linkedin/venice/fastclient/RetriableAvroGenericStoreClient.java
```diff
@@ -266,6 +268,17 @@ private <R extends MultiKeyRequestContext<K, V>, RESPONSE> void retryStreamingMu
 
     Runnable retryTask = () -> { // Look at the remaining keys and setup completion
       if (!pendingKeysFuture.isEmpty()) {
+        Throwable throwable = savedException.get();
```
m-nagarajan, line 271: Will it be better to do this when the original request completes rather than when the retry task starts, to free up the scheduler more proactively in case of lot of retries happening?

in retryStreamingMultiKeyRequest:
1. schedule the retry task first: scheduledRetryTask
2. send the original request along and pass in scheduledRetryTask to it
3. when original request completes with 429,  scheduledRetryTask can be cancelled there rather than waiting for retry task to executed and then cancel it.
4. Have the current change as a defensive code just in case the retry task starting while cancelling it.

what do you think?

m-nagarajan (commented):
Thanks @xunyin8 for the change. Left a comment.

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/RetriableAvroGenericStoreClient.java
xunyin8: Good call, let me make that change.

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/RetriableAvroGenericStoreClient.java
```diff
@@ -341,12 +353,24 @@ public void onCompletion(Optional<Exception> exception) {
           if (!exceptionToSave.isPresent()) {
             finalRequestCompletionFuture.complete(null);
           } else {
+            boolean shouldCompleteRequestFuture = false;
+            if (scheduledRetryTask != null && isExceptionCausedByTooManyRequests(exceptionToSave.get())) {
+              // Check if the exception is 429, if so cancel the retry if the retry task exists
```
m-nagarajan, line 358: as the retry task checks `savedException`, for the defensive code to be useful, I think we need to do this after 
`if (!savedException.compareAndSet(null, exceptionToSave.get())) {` block.

m-nagarajan (commented):
Mostly LGTM @xunyin8, just left some minor comments.

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/RetriableAvroGenericStoreClient.java
xunyin8: Good catch, will swap it for better coverage but just want to be clear that it's still not bullet proof and we try to cancel/avoid the retry as much as we can. Since the retry could be triggered already before we reach this block of code in the original request.

m-nagarajan: yes, agree with this.

m-nagarajan (approved):
LGTM, thanks @xunyin8

PR #894, opened 2024-03-09
[server] Fix NPE when collecting heartbeat_delay_ms and QuotaRcuPerSecondAllowed stats

nisargthakkar:
## Fix NPE when collecting `heartbeat_delay_ms` and `QuotaRcuPerSecondAllowed` stats

We noticed NPE exceptions in deployed instances in the metric collection code paths. This commit handles `null` gracefully and prevents these NPEs.

`AbstractVeniceAggVersionedStats#updateStatsVersionInfo` has the following code:
```
    if (newCurrentVersion != versionedStats.getCurrentVersion()) {
      versionedStats.setCurrentVersion(newCurrentVersion);
    }
...
    if (futureVersion != versionedStats.getFutureVersion()) {
      versionedStats.setFutureVersion(futureVersion);
    }
```

If there is no future version, the `if` block will never get executed. The `setFutureVersion` is what sets the `stat` object. However, the `heartbeat_delay_ms` metric already got registered like:
```
(ignored, ignored2) -> getStats().getFollowerLag(region).getMax()
```
So, `getStats()` would return `null` in such a case.

For `QuotaRcuPerSecondAllowed`, this code just avoids an NPE. I have not done much analysis into why `storeRepository.getStore(storeName)` could be `null`. We can investigate this more if we see a degradation in the metrics.

## How was this PR tested?

GH CI

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -717,6 +717,7 @@ ext.createDiffFile = { ->
       // da-vinci-client
       ':!clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java',
       ':!clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/isolated/IsolatedIngestionServer.java',
+      ':!clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/ingestion/heartbeat/HeartbeatStatReporter.java',
 
       // venice-client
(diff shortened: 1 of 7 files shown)


ZacAttack (dismissed):
Looks good, just gotta puzzle the checks

ZacAttack (approved):

PR #895, opened 2024-03-11
[server][da-vinci] Remove noise when clearing offsets without metadata partition

huangminchn:
## Summary
When a version is retired, both Helix resource and ingestion task will be deleted/killed; when Helix resource is deleted, OFFLINE->DROPPED transition message will be sent to server; and when ingestion task is killed, a kill-job message will be sent to servers; these two deletions are happening independently; if transitions are done before kill-job, and if ingestion task is trying to flush RESET_OFFSET actions, servers will fail to find metadata partition.

If metadata partition doesn't exist when trying to clear offsets, there is nothing left to clear and thus no point throwing exception for it.

## How was this PR tested?
CI

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/AbstractStorageEngine.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/AbstractStorageEngine.java
@@ -58,7 +58,7 @@ public abstract class AbstractStorageEngine<Partition extends AbstractStoragePar
   // Using a large positive number for metadata partition id instead of -1 can avoid database naming issues.
   public static final int METADATA_PARTITION_ID = 1000_000_000;
 
-  private final String storeName;
+  private final String storeVersionName;
   private final SparseConcurrentList<Partition> partitionList = new SparseConcurrentList<>();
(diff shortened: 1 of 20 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/AbstractStorageEngine.java
```diff
@@ -579,7 +579,8 @@ public synchronized Optional<OffsetRecord> getPartitionOffset(int partitionId) {
    */
   public synchronized void clearPartitionOffset(int partitionId) {
     if (!metadataPartitionCreated()) {
-      throw new StorageInitializationException("Metadata partition not created!");
+      LOGGER.info("Metadata partition not created; there is nothing to clear!");
```
majisourav99, line 582: can you add more info in the log like resource name/partition etc.

huangminchn: Good call! Updated.

majisourav99 (approved):
Thanks for clarifying the store to storeVersion field name!

PR #897, opened 2024-03-12
[server] Extend retry for new superset schema fetching.

haoxu07:

Previous when user updates a new value schema to the AAWC store, the ongoing ingestion will get acknowledged by ZK store to refresh the schema. Previous retry for fetching the new schema from zk is 3 times with a fixed 100ms, which is very short. Without fetching new schema, the store ingestion could fail.  For a real-prod system, usually server need to wait for 3~5 seconds to fetch this schema after new schema registered.  Here we use exponential back off retry.

## How was this PR tested?

Test in prod system by update test store's value schema.

- [ *] No. You can skip the rest of this section.

### DIFF
--- a/internal/venice-common/src/main/java/com/linkedin/venice/helix/HelixReadOnlySchemaRepository.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/helix/HelixReadOnlySchemaRepository.java
@@ -163,7 +163,10 @@ public class HelixReadOnlySchemaRepository implements ReadOnlySchemaRepository, 
     Store store = getStoreRepository().getStore(storeName);
     int supersetSchemaId = store.getLatestSuperSetValueSchemaId();
     AtomicReference<SchemaEntry> supersetSchemaEntry = new AtomicReference<>();
-    RetryUtils.executeWithMaxAttempt(() -> {
+    long currentTimestamp = System.currentTimeMillis();
+    List<Class<? extends Throwable>> retriableExceptions =
+        Collections.singletonList(InvalidVeniceSchemaException.class);
+    RetryUtils.executeWithMaxAttemptAndExponentialBackoff(() -> {
       try {
         getSchemaLock().writeLock().lock();
         SchemaData schemaData = getSchemaMap().get(storeName);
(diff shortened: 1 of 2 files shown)


internal/venice-common/src/main/java/com/linkedin/venice/helix/HelixReadOnlySchemaRepository.java
```diff
@@ -163,7 +163,10 @@ SchemaEntry forceRefreshSupersetSchemaWithRetry(String storeName) {
     Store store = getStoreRepository().getStore(storeName);
     int supersetSchemaId = store.getLatestSuperSetValueSchemaId();
     AtomicReference<SchemaEntry> supersetSchemaEntry = new AtomicReference<>();
-    RetryUtils.executeWithMaxAttempt(() -> {
+    Long currentTimestamp = System.currentTimeMillis();
```
sixpluszero, line 166: nit: Use long instead of Long to avoid auto boxing

internal/venice-common/src/main/java/com/linkedin/venice/helix/HelixReadOnlySchemaRepository.java
```diff
@@ -176,7 +179,13 @@ SchemaEntry forceRefreshSupersetSchemaWithRetry(String storeName) {
       } finally {
         getSchemaLock().writeLock().unlock();
       }
-    }, 3, Duration.ofMillis(100), Collections.singletonList(InvalidVeniceSchemaException.class));
+    }, 10, Duration.ofSeconds(1), Duration.ofMinutes(1), Duration.ofMinutes(5), retriableExceptions);
+    Long timePassed = System.currentTimeMillis() - currentTimestamp;
```
sixpluszero, line 183: same here.

sixpluszero (approved):
Thanks Hao! Looks good to me. Left a minor comment, should be easy to resolve.

internal/venice-common/src/main/java/com/linkedin/venice/helix/HelixReadOnlySchemaRepository.java
haoxu07: Fixed

sixpluszero (approved):

PR #898, opened 2024-03-13
[vpj] Clean up PushJobSetting to remove DataWriterMRJob specific fields

nisargthakkar:
## Clean up `PushJobSetting` and remove `DataWriterMRJob` specific fields

`PushJobSetting` has two configs that are specific to `DataWriterMRJob`. This is an anti-pattern and this commit moves them into `DataWriterMRJob`

## How was this PR tested?

GH CI


m-nagarajan (approved):
LGTM.  Thanks.

PR #899, opened 2024-03-14
[server] Fixed a few server metrics that weren't working properly

xunyin8:
## [server] Fixed a few server metrics that weren't working properly
1. client_connection_count gauge was not thread safe and resulting in negative numbers.

2. Ensure all KPS from different requests (single, multi and compute) are recorded by server stats so it can be used capacity planning for both thin-client and fast-client use cases.

3. Renamed a SN read quota metrics for clarity and fixed their behavior (QPS vs KPS).

## How was this PR tested?
Integration tests

### DIFF
--- a/internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/fastclient/FastClientIndividualFeatureConfigurationTest.java
+++ b/internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/fastclient/FastClientIndividualFeatureConfigurationTest.java
@@ -73,10 +73,10 @@ public class FastClientIndividualFeatureConfigurationTest extends AbstractClient
     }
     String readQuotaStorageNodeTokenBucketRemaining =
         ".venice-storage-node-token-bucket--QuotaRcuTokensRemaining.Gauge";
-    String readQuotaRequestedQPSString = "." + storeName + "--quota_requested_qps.Count";
-    String readQuotaRejectedQPSString = "." + storeName + "--quota_rejected_qps.Count";
-    String readQuotaRequestedKPSString = "." + storeName + "--quota_requested_kps.Count";
-    String readQuotaRejectedKPSString = "." + storeName + "--quota_rejected_kps.Count";
+    String readQuotaRequestedQPSString = "." + storeName + "--quota_request.Rate";
+    String readQuotaRejectedQPSString = "." + storeName + "--quota_rejected_request.Rate";
+    String readQuotaRequestedKPSString = "." + storeName + "--quota_request_key_count.Rate";
+    String readQuotaRejectedKPSString = "." + storeName + "--quota_rejected_key_count.Rate";
     String readQuotaAllowedUnintentionally = "." + storeName + "--quota_unintentionally_allowed_key_count.Count";
     String readQuotaUsageRatio = "." + storeName + "--quota_requested_usage_ratio.Gauge";
     String errorRequestString = ".total--error_request.OccurrenceRate";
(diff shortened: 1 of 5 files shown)


services/venice-server/src/main/java/com/linkedin/venice/listener/ServerStatsContext.java
```diff
@@ -208,9 +208,11 @@ public HttpResponseStatus getResponseStatus() {
   public void setRequestType(RequestType requestType) {
     switch (requestType) {
       case MULTI_GET:
+      case MULTI_GET_STREAMING:
         currentStats = multiGetStats;
         break;
       case COMPUTE:
+      case COMPUTE_STREAMING:
```
gaojieliu, line 215: So we have streaming request hitting Venice Server from fast client now?

services/venice-server/src/main/java/com/linkedin/venice/stats/ServerQuotaUsageStats.java
```diff
@@ -21,19 +20,19 @@ public class ServerQuotaUsageStats extends AbstractVeniceStats {
...
-    requestedQPS = registerSensor("quota_rcu_requested", new Count());
-    requestedKPS = registerSensor("quota_rcu_requested_key", new Total());
-    rejectedQPS = registerSensor("quota_rcu_rejected", new Count());
-    rejectedKPS = registerSensor("quota_rcu_rejected_key", new Total());
-    allowedUnintentionallyKPS = registerSensor("quota_rcu_allowed_unintentionally", new Count());
+    requestedQPS = registerSensor("quota_requested_qps", new Count());
+    requestedKPS = registerSensor("quota_requested_kps", new Count());
+    rejectedQPS = registerSensor("quota_rejected_qps", new Count());
+    rejectedKPS = registerSensor("quota_rejected_kps", new Count());
+    allowedUnintentionallyKPS = registerSensor("quota_unintentionally_allowed_key_count", new Count());
```
gaojieliu, line 27: These metric names are a little confusing.
Can we mutate them in the following way?
```
quota_request
quota_request_key_count
quota_rejected_request
quota_rejected_key_count
````
I think we can use `Rate` metric type to make the semantics more explicit.
What do you think?


I assume these metrics cover all request types, right?

File: services/venice-server/src/main/java/com/linkedin/venice/listener/ServerStatsContext.java (hunk 1 above)
Comment (xunyin8) [reply services/venice-server/src/main/java/com/linkedin/venice/listener/ServerStatsContext.java:215]: You mean compute streaming specifically? Not yet but PYMK could be ramping to it in Q4.

File: services/venice-server/src/main/java/com/linkedin/venice/stats/ServerQuotaUsageStats.java (hunk 2 above)
Comment (xunyin8) [reply services/venice-server/src/main/java/com/linkedin/venice/stats/ServerQuotaUsageStats.java:27]: Sure, will change them to the suggested names.

Yes, these metrics cover all request types (single, multi, compute, etc.).

File: services/venice-server/src/main/java/com/linkedin/venice/listener/ServerStatsContext.java (hunk 1 above)
Comment (gaojieliu) [reply services/venice-server/src/main/java/com/linkedin/venice/listener/ServerStatsContext.java:215]: Got it.

**Review by gaojieliu — APPROVED** [review]

File: internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/fastclient/FastClientIndividualFeatureConfigurationTest.java [hunk 3]
```diff
@@ -72,9 +73,11 @@ public void testServerReadQuota() throws Exception {
...
     String readQuotaStorageNodeTokenBucketRemaining =
         ".venice-storage-node-token-bucket--QuotaRcuTokensRemaining.Gauge";
-    String readQuotaRequestedString = "." + storeName + "--quota_rcu_requested.Count";
-    String readQuotaRejectedString = "." + storeName + "--quota_rcu_rejected.Count";
-    String readQuotaAllowedUnintentionally = "." + storeName + "--quota_rcu_allowed_unintentionally.Count";
+    String readQuotaRequestedQPSString = "." + storeName + "--quota_request.Rate";
+    String readQuotaRejectedQPSString = "." + storeName + "--quota_rejected_request.Rate";
+    String readQuotaRequestedKPSString = "." + storeName + "--quota_request_key_count.Rate";
+    String readQuotaRejectedKPSString = "." + storeName + "--quota_rejected_key_count.Rate";
+    String readQuotaAllowedUnintentionally = "." + storeName + "--quota_unintentionally_allowed_key_count.Count";
```
adamxchen, line 80: for my curiosity, why do we have quota assigned to clients unintentionally?

xunyin8: SN quota right now is fail pass. Meaning if for some reason SN didn't initialize properly or we can't fetch/calculate the quota then the enforcer will pretend the user has infinite quota and allow the requests unintentionally. We can change this behavior later once we believe the quota system is stable enough to do so (this metric never went above zero since we fixed some delayed CV update after restart which was months ago).

PR #900, opened 2024-03-14
[all] Bumped up the tehuti dep to pick up the latest release

gaojieliu:
Bump up the tehuti dep to pick up the latest AsyncGauge execution optimization.

## How was this PR tested?
CI

### DIFF
--- a/internal/venice-client-common/src/main/java/com/linkedin/venice/utils/metrics/MetricsRepositoryUtils.java
+++ b/internal/venice-client-common/src/main/java/com/linkedin/venice/utils/metrics/MetricsRepositoryUtils.java
@@ -1,10 +1,8 @@
 package com.linkedin.venice.utils.metrics;
 
-import com.linkedin.venice.utils.DaemonThreadFactory;
-import io.tehuti.metrics.AsyncGaugeConfig;
 import io.tehuti.metrics.MetricConfig;
 import io.tehuti.metrics.MetricsRepository;
-import java.util.concurrent.Executors;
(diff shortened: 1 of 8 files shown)


internal/venice-client-common/src/main/java/com/linkedin/venice/utils/metrics/MetricsRepositoryUtils.java
```diff
@@ -1,39 +1,34 @@
...
+import io.tehuti.metrics.stats.AsyncGauge;
 import java.util.concurrent.TimeUnit;
 
 
 /**
  * Utility functions to help create common metrics repository.
  */
 public class MetricsRepositoryUtils {
-  public static MetricsRepository createMultiThreadedMetricsRepository(String threadName) {
+  public static MetricsRepository createMultiThreadedMetricsRepository() {
```
sixpluszero, line 13: I don't have the exact context, but this change looks strange. 
Beyond bumping dependency, does it remove the unique name of the thread that operates on the metric? Could you explain a bit about the reason?

gaojieliu: Yes, it changed the API and doesn't allow to specify the thread name any more.
It only allows to specify the thread count, and internally it will create a daemon thread pool with the same prefix.
It is not common to use multiple MetricsRepository but just in test, so I think it should be fine.

sixpluszero (approved):

PR #902, opened 2024-03-15
[controller] DaVinci offline instance wait time is now configurable

huangminchn:
## Summary
Previously the offline instance wait time is hardcoded to 5 minutes; this PR makes it configureable with config name:
da.vinci.offline.instance.wait.time.in.minutes

Default value is set to 15 minutes.

## How was this PR tested?
CI

### DIFF
--- a/internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
@@ -1757,6 +1757,13 @@ public class ConfigKeys {
       "push.status.store.heartbeat.expiration.seconds";
 
   /**
+   * Controller will wait for unresponsive DaVinci instances for this amount of time before announcing the push fails
+   * on DaVinci side.
+   */
+  public static final String DA_VINCI_OFFLINE_INSTANCE_WAIT_TIME_IN_MINUTES =
+      "da.vinci.offline.instance.wait.time.in.minutes";
+
+  /**
    * Derived schemaId for push status store write compute.
    */
   public static final String PUSH_STATUS_STORE_DERIVED_SCHEMA_ID = "push.status.store.derived.schema.id";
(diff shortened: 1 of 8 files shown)


internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
```diff
@@ -1756,6 +1756,13 @@ private ConfigKeys() {
   public static final String PUSH_STATUS_STORE_HEARTBEAT_EXPIRATION_TIME_IN_SECONDS =
```
gaojieliu, line 1756: @sixpluszero 
I thought this config is used to control when to mark a DaVinci instance as offline or not, no?

sixpluszero: Yes, this is the original config for PushStatusReader to see if the retrieved HB for an instance is stale or not.

gaojieliu: Sure, can we just bump this config to make Controller more tolerant to handle the spiky push status write from massive DVC instances during push? Do we still need this new change?

sixpluszero: Actually this sounds like a good idea... If tweak this to 20-30 mins for example, instances might be good after the offset spike comes down...

sixpluszero: Reading the HB retrieval code, if HB is not available at all, ti will be marked as 0 and instance will still be considered offline. But this is based on the fact that certain instance has been reporting status. I think it will report one HB before it report STARTED, because HB reporting thread does not have initial delay

gaojieliu: If there is no HB from a certain DVC instance, that instance won't show up in the push status tracking map for that specific store version, right? I assume the HB event would be the first event in the tracking per DVC instance.
If that works, let us try to tweak the existing config to see whether it works before adding a new one.

huangminchn: We have increased the timeout for determining offline DVC instance using an existing config. I am going to discard this PR.

PR #903, opened 2024-03-18
[vpj] Cleanup dictionary fetching logic from various sources

nisargthakkar:
## Cleanup dictionary fetching logic from various sources

Currently VPJ either builds a dictionary or fetches it from various sources to optimize the compression ratios. However, this logic seemed a bit confusing and there was scope to streamline the implementation. This commit splits `VenicePushJob#getCompressionDictionary()` into two functions - `VenicePushJob#getCompressionDictionary()` and `VenicePushJob#fetchOrBuildCompressionDictionary()`. `VenicePushJob#fetchOrBuildCompressionDictionary()` handles the logic of how to get the compression dictionary based on various store and job properties, and `VenicePushJob#getCompressionDictionary()` unifies the error handling and fallback to synthetic dictionaries if the compression strategy is `ZSTD_WITH_DICT`.

The case where `use.mapper.to.build.dictionary` is `false`, and dictionary build fails, would previously throw exceptions (even for `ZSTD_WITH_DICT`). This change handles it gracefully and uses a dictionary built on synthetic data instead.

## How was this PR tested?

GH CI. Also tested manually

### DIFF
--- a/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
+++ b/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
@@ -1127,7 +1127,7 @@ public class VenicePushJob implements AutoCloseable {
       if (!inputFileHasRecords) {
         if (pushJobSetting.storeCompressionStrategy == CompressionStrategy.ZSTD_WITH_DICT) {
           LOGGER.info(
-              "compression strategy is {} with no input records: dictionary will be generated from synthetic data or current version data for hybrid stores",
+              "Compression strategy is {} with no input records. A dictionary will be generated from synthetic data or current version data for hybrid stores",
               pushJobSetting.storeCompressionStrategy);
         } else {
           LOGGER.info("No compression dictionary will be generated as there are no records");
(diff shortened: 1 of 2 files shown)


clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
```diff
@@ -1425,8 +1425,37 @@ private ControllerClient getControllerClient(
...
+
+    if (pushJobSetting.storeCompressionStrategy != CompressionStrategy.ZSTD_WITH_DICT) {
+      LOGGER.info(
+          "No dictionary fetched. But since the compression strategy is not {}, it is not required",
+          CompressionStrategy.ZSTD_WITH_DICT);
+      pushJobSetting.isZstdDictCreationSuccess = false;
+      return Optional.empty();
+    }
+
+    LOGGER.info("No dictionary fetched. Creating a default dictionary since compression strategy is ZSTD_WITH_DICT");
```
xunyin8, line 1453: In this branch would we get any alert/notification without diving deep into the logs? I'm worried that if `CompressionStrategy.ZSTD_WITH_DICT` is enabled but we failed to fetch due to some random reason wouldn't it impact compression performance without us knowing about it?

nisargthakkar: Yes, it affects the compression ratios. But the alternative for `ZSTD_WITH_DICT` (without a lot of refactoring) is to fail the push job. In case the compression dictionary is bad for the store, the push will fail due to quota exceeding

xunyin8: I see, give the push a chance with default dictionary does sound better for now. We might want to document this as a comment or something so we can add more signals if dictionary fetch fails frequently.

xunyin8 (approved):

PR #904, opened 2024-03-18
[server] Server read quota usage ratio metric improvement

xunyin8:
## [server] Server read quota usage ratio metric improvement
1. Previously the quota ratio is only updated whenever the token bucket is refilled. This is problematic for stores that have a low KPS relative to their allocated quota since it won't be refilled very often. As a result we could see the old usage ratio for extended periods of time if the token bucket is not refilled frequently.

2. To improve this we will perform the calculation whenever the metric is read by using the requested token since last refill and elasped time since last refill to compute the average KPS and divide by refill KPS.

## How was this PR tested?
existing unit and integration tests

### DIFF
--- a/internal/venice-common/src/main/java/com/linkedin/venice/throttle/TokenBucket.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/throttle/TokenBucket.java
@@ -18,9 +18,8 @@ public class TokenBucket {
   private final Clock clock;
   private final AtomicLong tokens;
   private final AtomicLong tokensRequestedSinceLastRefill;
-  private volatile long previousRefillTime = 0;
+  private volatile long previousRefillTime;
   private volatile long nextUpdateTime;
-  private volatile double staleUsageRatio = 0;
 
   /**
    * This constructor should only be used by tests.  Application should not specify it's own instance of Clock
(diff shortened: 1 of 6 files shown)


nisargthakkar (approved):
LGTM

nisargthakkar (dismissed):
LGTM

services/venice-server/src/test/java/com/linkedin/venice/stats/AggServerQuotaUsageStatsTest.java
```diff
@@ -0,0 +1,66 @@
...
+    long batchSize2 = 200;
+    aggServerQuotaUsageStats.recordAllowed(storeName, batchSize);
+    aggServerQuotaUsageStats.recordAllowed(storeName, batchSize);
+    aggServerQuotaUsageStats.recordAllowed(storeName2, batchSize2);
+
+    double totalQPS = metricsRepository.getMetric(readQuotaRequestedQPSString).value()
+        + metricsRepository.getMetric(readQuotaRequestedQPSString2).value();
+    double totalKPS = metricsRepository.getMetric(readQuotaRequestedKPSString).value()
+        + metricsRepository.getMetric(readQuotaRequestedKPSString2).value();
+    Assert.assertEquals(metricsRepository.getMetric(totalReadQuotaRequestedQPSString).value(), totalQPS, 0.01);
```
nisargthakkar, line 35: Should we also assert the expected raw values of these metrics?

xunyin8: The rate metric calculation time window is very unpredictable, I tried it and it returns a different value depending on execution/call time.

nisargthakkar (approved):
Sweet. Thank you

PR #905, opened 2024-03-19
[dvc] Don't throw DIV MISSING data fault for new segment after EOP

lluwm:
Today, for hybrid stores after EOP is received, DIV creates a new segment and throws data missing fault when:

1. EOP is received
2. no previous segment is found for the input guid and partition in the record
3. sequence number in the record is > 0

Actually, this does not seem to be right simply because that by looking at the data that DIV has at the moment, it's too early to make the conclusion that missing data scenario happened (e.g. this is the very first message that DIV receives from the guid for that partition).

In this change, DIV accepts that this is a new message from the guid for the partition and not throws any fault. (also a bit of code refactoring in related function)

## How was this PR tested?

- New unit test
- GHA


clients/da-vinci-client/src/main/java/com/linkedin/davinci/validation/PartitionTracker.java
```diff
@@ -385,9 +385,11 @@ private void trackSequenceNumber(
       segment.start();
       segment.setLastRecordProducerTimestamp(consumerRecord.getValue().producerMetadata.messageTimestamp);
       return;
-    } else if (segment.isNewSegment() && incomingSequenceNumber == previousSequenceNumber) {
-      if (segment.getSequenceNumber() > 0 && !tolerateMissingMsgs.get()) {
-        throw DataFaultType.MISSING.getNewException(segment, consumerRecord);
+    }
+
+    if (incomingSequenceNumber == previousSequenceNumber) {
+      if (!segment.isNewSegment() && incomingSequenceNumber > 0) {
```
nisargthakkar, line 391: We should also check `!tolerateMissingMsgs.get()`, right?

lluwm: The next line is changed to throw DUPLICATE exception, and for duplicate exception, do we still need the `tolerateMissingMsgs` check?

lluwm: Just to clarify on this , today the behavior is (assuming incomingSequenceNumber == previousSequenceNumber):

isNewSegment == true  --> throw MISSING
isNewSegment == false --> throw DUPLICATE

after this change:
isNewSegment == true  --> return (no error)
isNewSegment == false --> throw DUPLICATE

gaojieliu:
Did you verify whether all the missing message logs are caused by this reason or not?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/validation/PartitionTracker.java
nisargthakkar: > isNewSegment == true --> return (no fault)
This should be changed to `isNewSegment == true && prevSegment = 0` (or however we track previously unknown segments). Otherwise, it is still a `MISSING` message error since there is an entire segment missing

lluwm: I see, do you mean the e.g. `hasPreviousSegment` flag? I think the scenario that you pointed to is probably covered in `trackSegment` which is called earlier before this function `trackSequenceNumber`. There,  if `we see a new segment` && `prev seg still exists` && `not tolerate missing message`, we throw the MISSING fault.

```
      } else if (incomingSegmentNumber > previousSegmentNumber) {
        if (tolerateMissingMsgs.get()) {
          return initializeNewSegment(consumerRecord, endOfPushReceived, true);
        } else {
          throw DataFaultType.MISSING.getNewException(previousSegment, consumerRecord);
        }
```

huangminchn: Let's stop using "incomingSequenceNumber > 0" in any condition check; this idea is wrong from the beginning.
We thought that when there is a new segment but the new segment doesn't start from sequence number 0, then it indicates data loss - this idea was wrong.
So far, I can think of 3 cases where a new segment appears, and its sequence number is positive, but no data loss:
1. When buffer replay happens for the first time, leader switches to some offset of RT, there is no guarantee that leader would happen to land on START_OF_SEGMENT message.
2. Node X used to be leader and consume RT, it tracks GUID "Samza" up to segment 10, sequence 10, and then leadership is handed over to Node Y. Node Y replicates some data from RT, and hands over back to Node X which resumes consumption from where Y leaves behind, and the first RT message X encounters could be segment 11, sequence 1, which is a new segment from GUID "Samza", and its sequence number is positive, but we can't tell if there is data loss.
3. The incident that you discovered - the DIV clean up process kicked in mistakenly and clears DIV info for GUID "Samza"; when Node X consumes a RT message from the same GUID again, it will be a new Segment, and sequence number could be positive, but again, we can't tell if there is data loss.

Therefore, I would recommend removing this check "incomingSequenceNumber > 0".

Then let's consider if we have enough information to differentiate the real data loss and then confidently throw a MISSING message exception. My understanding is it's possible to do it in follower replica which always consumes VT and it can use "tolerateMissingMsgs" to check if VT is compacted at this offset; but it's impossible to do it in leader replica unless we start tracking the below two information:
1. If this GUID has previous segment, and we want to throw missing exception because new segment has positive sequence number, is there a leadership handover? If so, what are the last segment and sequence number from GUID "Samza" that the previous leader produces?
2. If this GUID doesn't have previous segment and we want to throw missing exception due to the same reason, does a DIV clean up happen? If so, what are the last segment and sequence number from this GUID before the clean-up?
To get these two info, we need to put Samza producer metadata into VT records; and revert clean-up feature. KME upgrade is somewhat feasible, but reverting clean-up doesn't make sense.

However, we should definitely keep this part:
1. If "tolerateMissingMsgs" is false (no compaction)
2. newSegment is false
3. but incomingSequenceNumber > previousSequenceNumber + 1
There is definitely data-loss.

@FelixGV @gaojieliu Please take a look when you have time. When newSegment is true, I don't think the current implementation is able to tell if there is data loss.

huangminchn (changes requested):

clients/da-vinci-client/src/main/java/com/linkedin/davinci/validation/PartitionTracker.java
gaojieliu: Sure.
The follower check makes sense and today it can only validate VT, not RT.
In Leader replica, there are 3 use cases we can NOT differentiate, which means this check is problematic, and it is noisy, but can't provide enough value.
Personally, I would recommend to disable DIV check completely for RT processing, only validate VT.
Once we address all the known issues, we resume the RT DIV check in leader replicas.

lluwm: I like Ming's and Gaojie's idea in general, basically, for RT DIV checks, either DIV doesn't warn at all or it only warns when DIV is 100% certain that this is truly a data loss and keeps silence for suspicious ones.

gaojieliu: Let us get @FelixGV 's opinions on this.

lluwm: @huangminchn, regarding `incomingSequenceNumber > 0`, in the new change, we don't throw DIV MISSING fault anymore for the condition that `incomingSequenceNumber == previousSequenceNumber`. `incomingSequenceNumber > 0` is now used for throw DUPLICATE.

isNewSegment == true --> return (no fault)
isNewSegment == false --> throw DUPLICATE

huangminchn: Right, but it's wrong to use "incomingSequenceNumber > 0" to make any decision; let's remove this condition.

Gaojie is correct that RT DIV can't be accurate. At the end of the day, we need to choose between false positive (warn but actually no data loss) and false negative (less warning logs but maybe data loss happens)

lluwm: sounds good, condition removed.

FelixGV: IMHO, it makes sense to eliminate false positives. But I don't think we need to disable RT DIV completely in order to achieve that... do we?

gaojieliu: With the current design, can we eliminate all the false positives?
At least, the false positives triggered by rebalance can be addressed easily at this moment.

huangminchn (approved):
Thanks a lot Lei! The current implementation is better than what we have before; we removed some false positives; no change to false negative with the current design.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/validation/PartitionTracker.java
lluwm: Not all false positives, but some of them, especially the ones when hybrid rewind time and DIV cleanup maxAge exists a gap. Due to this gap, DIV cleanup component unnecessarily deletes segment, but later DIV itself creates a new segment again and throws data missing fault, and then the same segment got deleted again and recreated again ... repeatedly

PR #907, opened 2024-03-20
[changelog] Make changelog consumer creation parallel in VeniceChangelogConsumerClientFactory

sixpluszero:
## [changelog] Make changelog consumer creation parallel in VeniceChangelogConsumerClientFactory
In user onboarding we noticed that multiple consumer will spent extra time on creation. During log investigation, we found that this factory method is the root cause. 
The synchronized keyword makes consumer creation sequential, including Apache consumer creation and StorageService restoration. Each of them could take 1-2s so in total it took ~11 in practice to complete. 
This change should make it truly parallel in creation and save ~10s during bootstrap.

## How was this PR tested?

N/A

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumerClientFactory.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumerClientFactory.java
@@ -58,7 +58,7 @@ public class VeniceChangelogConsumerClientFactory {
   /**
    * Default method to create a {@link VeniceChangelogConsumer} given a storeName.
    */
-  public synchronized <K, V> VeniceChangelogConsumer<K, V> getChangelogConsumer(String storeName) {
+  public <K, V> VeniceChangelogConsumer<K, V> getChangelogConsumer(String storeName) {
     return getChangelogConsumer(storeName, null);
   }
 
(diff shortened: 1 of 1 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumerClientFactory.java
```diff
@@ -98,29 +100,25 @@ public synchronized <K, V> BootstrappingVeniceChangelogConsumer<K, V> getBootstr
...
-
-      return new LocalBootstrappingVeniceChangelogConsumer(
-          newStoreChangelogClientConfig,
-          consumer != null
-              ? consumer
-              : getConsumer(newStoreChangelogClientConfig.getConsumerProperties(), consumerName),
-          consumerId);
-    });
+    return (BootstrappingVeniceChangelogConsumer<K, V>) storeClientMap
+        .computeIfAbsent(suffixConsumerIdToStore(storeName, consumerId), name -> {
```
sushantmane, line 107: can you change `storeClientMap` to `ConcurrentHashMap` so that to guarantee one consumer per client

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumerClientFactory.java
```diff
@@ -23,6 +23,8 @@
 
 
 public class VeniceChangelogConsumerClientFactory {
+  private static final ApacheKafkaConsumerAdapterFactory kafkaConsumerAdapterFactory =
```
sushantmane, line 26: Can we use interface here `PubSubConsumerAdapterFactory`?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumerClientFactory.java
sixpluszero: Good catch. I prefer VeniceConcurrentHashMap directly everywhere if possible. Fixed.

sushantmane (dismissed):
LGTM! Thanks, @sixpluszero!

sushantmane (approved):

PR #908, opened 2024-03-21
[server] Remove upstream offset calculation and offset record update in processTopicSwitch() for both leader and follower

sixpluszero:
## [server] Remove upstream offset calculation and offset record update in processTopicSwitch() for both leader and follower

The original issue is, when re-bootstrapping a current/backup version partition, when hitting proessTopicSwitch, it will calculate rewind and get the offset by rewind. It will persist the offset to the OffsetRecord. However, due to RT message retention, old messages got removed, and the newly calculated offset is pointing to newer message's offset. When it is processing the message in VT (which was consumed by leader from RT), the old message's offset < persisted offset, and SIT will throw exception, thinking multiple leader is consuming.

After checking the code, I think we do such logic in current code:
**Leader**

1.  processTopicSwitch(drainer): Calculate offset, persist.
2.  leaderExecuteTopicSwitch(consumer long running task thread): Calculate offset, persist, subscribe

**Follower**

1.  processTopicSwitch(drainer): Calculate offset, persist, set leader topic.

I think for both leader and follower, processTopicSwitch does not need to perform extra calculation, as follower should trust what leader produce into VT. For leader, that's also redundant calculation.

This PR removes two rewind calculation from both LFSIT and AASIT in API processTopicSwitch().

## How was this PR tested?

Tweak unit test to make sure no such persist API call in both L & F.

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
@@ -1086,58 +1086,6 @@ public class ActiveActiveStoreIngestionTask extends LeaderFollowerStoreIngestion
     PubSubTopic newSourceTopic = pubSubTopicRepository.getTopic(newSourceTopicName);
     Map<String, Long> upstreamStartOffsetByKafkaURL = new HashMap<>(topicSwitch.sourceKafkaServers.size());
 
-    /*
-    // Calculate the start offset based on start timestamp
-    if (!isDaVinciClient) {
-      final int newSourceTopicPartition = partitionConsumptionState.getSourceTopicPartitionNumber(newSourceTopic);
-      AtomicInteger numberOfContactedBrokers = new AtomicInteger(0);
-      topicSwitch.sourceKafkaServers.forEach(sourceKafkaURL -> {
-        long rewindStartTimestamp;
-        // calculate the rewind start time here if controller asked to do so by using this sentinel value.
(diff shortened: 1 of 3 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
```diff
@@ -1080,60 +1082,10 @@ protected boolean processTopicSwitch(
...
-      });
 
-      if (numberOfContactedBrokers.get() == 0) {
-        throw new VeniceException("Failed to query any broker for rewind!  Aborting topic switch processing!");
-      }
-
-      upstreamStartOffsetByKafkaURL.forEach((sourceKafkaURL, upstreamStartOffset) -> {
-        partitionConsumptionState.getOffsetRecord().setLeaderUpstreamOffset(sourceKafkaURL, upstreamStartOffset);
-      });
-    }
```
gaojieliu, line 1136: Nit:
Shall we just pass an empty map into the following function call?
```
syncTopicSwitchToIngestionMetadataService(topicSwitch, partitionConsumptionState, Collections.emptyMap());
```

gaojieliu (approved):
Looks good, thanks!

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
sixpluszero: Updated the PR. Yes, it is not even necessary to print an empty map, so I just removed it. In leader's execute TS logic, it prints out the actual upstream and offset it will subscribe to and start with, so I think it is fine.

gaojieliu (approved):

PR #910, opened 2024-03-21
[server] Add threadsafe mode to venice-server which adjusts message processing order

ZacAttack:
## [server][WIP] Add threadsafe mode to venice-server which adjusts message processing order
This is an initial phase PR.  It is seen as the minimal set of changes needed in order to add a mode where writes on leader are committed to rocksdb prior to producing.  This change in order has the following impacts:

**Drainer is skipped on leaders:**
 In a later refactor it might be prudent to remove the drainer entirely.  However, in order to best accomodate that, it would likely make sense to execute a kind of batching logic when flushing to rocksdb.  We do not attempt to make this change in this PR.

**DCR logic must change:**
 Since writes are persisted to rocksdb prior to producing to Kafka, we now must accomodate for the possibility of left over state on a leader.  To address this, we add a new mode to the merge conflict resolution logic where upon a perfect tie (on value and timestamp), we resolve to produce the repeated record.  The intention here is to be able to be certain that a write which was persisted to rocksdb on leader but not produced doesn't end up getting lost due to failing DCR.

**Transient Record is disabled**
 transient record cache is disabled for those ingestion tasks which enable this mode.  This is itself was one of the goals, but we should go here with some validation.  Most clusters in production end up seeing pretty low cache hit rate on transient record cache in production, however, there is at least one use case that gets as high as a 20% hit rate.  Theoretically, we may be able to avoid taking too much hit here as we are able to give the memory savings to rocksdb cache, but this needs vetting.  If this doesn't work, then we will need to replace the transient record cache with a simple size/time based cache.

There are also some cleanups here and there.  Getting rid of some code paths that we no longer need and cleaning up others.

NOTE: Integration tests haven't been completely added to this PR yet.  Part of that is because while switching some of the existing integration tests to this mode, some tests are failing.  This needs some more diagnosis.  Hence the WIP tag.

Resolves #XXX


clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -2231,9 +2263,28 @@ public void processConsumerRecord(
...
+      LeaderProducedRecordContext leaderProducedRecordContext,
+      PartitionConsumptionState partitionConsumptionState,
+      int subPartition) {
+
+    long syncBytesInterval = partitionConsumptionState.isDeferredWrite()
+        ? databaseSyncBytesIntervalForDeferredWriteMode
+        : databaseSyncBytesIntervalForTransactionalMode;
+    boolean recordsProcessedAboveSyncIntervalThreshold = (syncBytesInterval > 0
+        && (partitionConsumptionState.getProcessedRecordSizeSinceLastSync() >= syncBytesInterval));
+    defaultReadyToServeChecker.apply(partitionConsumptionState, recordsProcessedAboveSyncIntervalThreshold);
```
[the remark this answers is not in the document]
ZacAttack: oh.  So the recordsProcessedAboveSyncIntervalThreshold includes the synced bytes?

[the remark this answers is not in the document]
ZacAttack: Hrm.  I guess so, that would make sense.  I'll bump it down.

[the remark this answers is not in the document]
ZacAttack: acutally.  I don't think this will get included in that calculation in any case.  So I don't think it matters?  I'll leave it though where it is because this was the ordering that was in the original code.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/PartitionConsumptionState.java
```diff
@@ -565,6 +573,10 @@ public void setTransientRecord(
       int valueLen,
       int valueSchemaId,
       GenericRecord replicationMetadataRecord) {
+    if (this.threadSafeMode) {
+      // NoOp
```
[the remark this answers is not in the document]
ZacAttack: I think the problem then becomes how do we shrink the TR then.  Right now theres an implicit TR max size that's the size of the number of requests in flight between consumption and commit to rocksdb (we evict it once it's gone through the drainer).  In this new scheme I'm not sure when we'd evict.  One way to do that would be to swap the TR cache for a caffeing cache thats size/time bound.  But I'm trying to keep this first patch as small as possible.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestActiveActiveIngestion.java
```diff
@@ -127,6 +127,7 @@ public void setUp() {
     Properties serverProperties = new Properties();
     serverProperties.setProperty(ConfigKeys.SERVER_PROMOTION_TO_LEADER_REPLICA_DELAY_SECONDS, Long.toString(1));
     serverProperties.put(ROCKSDB_PLAIN_TABLE_FORMAT_ENABLED, false);
+    serverProperties.put(SERVER_INGESTION_TASK_THREAD_SAFE_MODE, false);
```
sixpluszero, line 130: Since this mode is default false, can we make it to run both true and false for some sophisticated tests for AAWC?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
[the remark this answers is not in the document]
sixpluszero: Hmm, actually, it looks like if maybeSyncOffset == true here, then you are doing the sync twice?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -1602,13 +1602,27 @@ protected void produceToLocalKafka(
       String kafkaUrl,
       int kafkaClusterId,
       long beforeProcessingRecordTimestampNs) {
+
+    if (this.runInThreadSafeMode) {
+      // Write to rocksdb. At time of writing, this is the last step after a huge amount of processing and compression
+      // and whatnot. At this stage we do not sync the offset, instead doing that after successfully produce.
+      this.processConsumerRecord(
```
sixpluszero, line 1609: I just suddenly think about this issue and left this comment so it might be totally wrong...

The issue I believe today we have in CC processing logic is:
In order to have good processing throughput, we use another thread to do View processing and producing, and does not block the same consumer thread to fetch next record from the same partition which might be the same key.
If it is the same key, then I feel like the problem is still there. Even if we disable TR, and do the drainer work directly here but since the upstream async logic is still unchanged, the race condition is still there. If let's say next key is the same key, then it will fetch from storage directly, assuming VW producing is a bit slow, and the commit does not happen. Then you loss the first update you just processed and directly apply the 2nd update in the original record...

sixpluszero (commented):
Thanks for the update, now I am not sure if I totally follow the logic as I post my concern in one of the reply......we probably should chat a bit offline to chew the code change.
Also, we probably need some tests coverage to prove our old mode can fail in the race condition case and new mode is correct, even if my concern is eventually proved invalid...

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
sixpluszero: I think I am wrong, given we don't play around with TR, it should be good. Let me re-think about this.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -2834,6 +2885,14 @@ private int internalProcessConsumerRecord(
           LOGGER.error("Failed to record Record heartbeat with message: ", e);
         }
       } else {
+        // TODO: This is a hack. Today the code kind of does a backdoor change to a key in the leaderProducer callback.
```
sixpluszero, line 2888: Actually I don't really follow this hack, can you explain a bit? I remember the issue being the leader does not chunk the record before it is produced to VT, and it seems like here it does not solve the issue? So the Leader and the Follower will still have different view?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderProducerCallback.java
```diff
@@ -337,12 +344,36 @@ void produceDeprecatedChunkDeletionToStoreBufferService(ChunkedValueManifest man
...
+      LeaderProducedRecordContext leaderProducedRecordContext,
+      int subPartition,
+      String kafkaUrl,
+      long beforeProcessingRecordTimestampNs,
+      long currentTimeForMetricsMs,
+      boolean chunkDeprecation) throws InterruptedException {
+    if (this.syncOffsetsOnlyAfterProducing && !chunkDeprecation) {
+      // sync offsets
+      ingestionTask
+          .maybeSyncOffsets(consumedRecord, leaderProducedRecordContext, partitionConsumptionState, subPartition);
```
FelixGV, line 369: Does the PCS instance passed here has a chance of having been modified by another thread (the processing thread)? I imagine we don't clone the PCS in order to make them immutable, so I wonder if what we would be checkpointing here is guaranteed to represent the correct state, up until what was just produced, rather than up until what's been consumed by another thread?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/PartitionConsumptionState.java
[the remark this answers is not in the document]
FelixGV: Actually, one of the goals of this refactoring is to get rid of the `TRC`, since that is going to yield an efficiency improvement, and reduce overall complexity. IOW... if there is a reason why we would need to keep the TRC, then I think we need to scrutinize this thoroughly, and I wonder in that case if the refactoring is worth it at all...

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -3834,27 +3887,38 @@ protected abstract DelegateConsumerRecordResult delegateConsumerRecord(
...
   protected enum DelegateConsumerRecordResult {
     /**
-     * The consumer record has been produced to local version topic by leader.
+     * The consumer record should be produced to Kafka
      */
-    PRODUCED_TO_KAFKA,
+    PRODUCE_TO_KAFKA,
     /**
-     * The consumer record has been put into drainer queue; the following cases will result in putting to drainer directly:
+     * The consumer record needs to be put into drainer queue; the following cases will result in putting to drainer directly:
```
FelixGV, line 3911: We should delete the O/B/O reference below, right?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/replication/merge/MergeConflictResolver.java
```diff
@@ -3,9 +3,6 @@
 import static com.linkedin.venice.schema.rmd.RmdConstants.REPLICATION_CHECKPOINT_VECTOR_FIELD_POS;
 import static com.linkedin.venice.schema.rmd.RmdConstants.TIMESTAMP_FIELD_POS;
 import static com.linkedin.venice.schema.rmd.RmdTimestampType.PER_FIELD_TIMESTAMP;
-import static com.linkedin.venice.schema.rmd.v1.CollectionRmdTimestamp.PUT_ONLY_PART_LENGTH_FIELD_POS;
-import static com.linkedin.venice.schema.rmd.v1.CollectionRmdTimestamp.TOP_LEVEL_COLO_ID_FIELD_POS;
-import static com.linkedin.venice.schema.rmd.v1.CollectionRmdTimestamp.TOP_LEVEL_TS_FIELD_POS;
```
FelixGV, line 8: Nitpick: why are we removing those? It makes the code a bit more verbose, and it adds trivially modified lines into the git blame, which inflates the size of an already complex PR... is there any benefit? Perhaps there is ambiguity between the constants coming from different classes? (In which case I'd be fine with keeping the change...)

clients/da-vinci-client/src/main/java/com/linkedin/davinci/schema/merge/PerFieldTimestampMergeRecordHelper.java
```diff
@@ -27,14 +27,11 @@ public UpdateResultStatus putOnField(
...
-      if (newFieldCompletelyReplaceOldField) {
+
+      if (newFieldValue != oldFieldValue) {
         oldRecord.put(oldRecordField.pos(), newFieldValue);
       }
-      return newFieldCompletelyReplaceOldField
-          ? UpdateResultStatus.COMPLETELY_UPDATED
-          : UpdateResultStatus.NOT_UPDATED_AT_ALL;
-
+      return UpdateResultStatus.COMPLETELY_UPDATED;
```
FelixGV, line 34: So I guess this is the change referenced in the commit message about:

> **DCR logic must change:**
> Since writes are persisted to rocksdb prior to producing to Kafka, we now must accomodate for the possibility of left over state on a leader. To address this, we add a new mode to the merge conflict resolution logic where upon a perfect tie (on value and timestamp), we resolve to produce the repeated record. The intention here is to be able to be certain that a write which was persisted to rocksdb on leader but not produced doesn't end up getting lost due to failing DCR.

I don't understand this, however... why would there be changes persisted in the leader but not in the VT, in cases of a perfect tie? If it is a perfect tie, then I assume the leader need not persist any changes either, right...?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/schema/merge/SortBasedCollectionFieldOpHandler.java
```diff
@@ -773,7 +773,7 @@ private UpdateResultStatus handleModifyPutOnlyMap(
     deletedKeys.sort(String::compareTo);
 
     // Step 3: Set new active elements and their active timestamps.
-    setNewMapActiveElementAndTs(
+    setNewMapActiveElementAndTimestamp(
```
FelixGV, line 776: Nitpick... I would prefer trivial renames to be in a separate PR, to minimize the size of complex PRs... but up to you.

FelixGV (commented):
Thanks for working on this! I wish I would have reviewed it sooner, but better late than never I guess... my comments are pretty minor, I think... but I have a few questions as well.

I don't see which integration tests are disabled. Is this a stale part of the commit message / PR body? Let's delete that part if it is stale...

Thanks again!

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderProducerCallback.java
ZacAttack: That's a good catch, yeah it could happen.

PR #911, opened 2024-03-21
[server][dvc] Lag arithmetic fix

FelixGV:
There is an edge case where an empty topic with an end offset of zero gets mistaken for an error code, and results in returning Long.MAX_VALUE as the lag instead.

## How was this PR tested?
It was not... tests should be added.

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
@@ -2431,13 +2431,19 @@ public abstract class StoreIngestionTask implements Runnable, Closeable {
       // A negative value means there was a problem in measuring the end offset, and therefore we return "infinite lag"
       return Long.MAX_VALUE;
     } else if (endOffset == 0) {
-      // The topic is empty and therefore, by definition, there cannot be any lag
+      /**
+       * Topics which were never produced to have an end offset of zero. Such topics are empty and therefore, by
+       * definition, there cannot be any lag.
+       *
+       * Note that the reverse is not true: a topic can be currently empty and have an end offset above zero, if it had
+       * messages produced to it before, which have since then disappeared (e.g. due to time-based retention).
+       */
       return 0;
     }
 
     /**
-     * An empty topic should have an end offset of zero. A topic with a single message in it will have an end offset of
-     * 1, while that single message will have offset 0. In such single message topic, a consumer which fully scans the
+     * A topic with an end offset of zero is empty. A topic with a single message in it will have an end offset of 1,
+     * while that single message will have offset 0. In such single message topic, a consumer which fully scans the
      * topic would have a current offset of 0, while the topic has an end offset of 1, and therefore we need to subtract
      * 1 from the end offset in order to arrive at the correct lag of 0.
      */
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
@@ -2412,13 +2412,27 @@ public abstract class StoreIngestionTask implements Runnable, Closeable {
       PubSubTopic topic,
       int partition,
       long currentOffset) {
-    if (currentOffset < 0) {
+    return measureLagWithCallToPubSub(pubSubServerName, topic, partition, currentOffset, this::getTopicManager);
+  }
+
+  protected static long measureLagWithCallToPubSub(
+      String pubSubServerName,
+      PubSubTopic topic,
+      int partition,
+      long currentOffset,
+      Function<String, TopicManager> topicManagerProvider) {
+    if (currentOffset < OffsetRecord.LOWEST_OFFSET) {
+      // -1 is a valid offset, which means that nothing was consumed yet, but anything below that is invalid.
       return Long.MAX_VALUE;
     }
-    TopicManager tm = getTopicManager(pubSubServerName);
+    TopicManager tm = topicManagerProvider.apply(pubSubServerName);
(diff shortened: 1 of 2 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -2416,11 +2416,18 @@ protected long measureLagWithCallToPubSub(
       return Long.MAX_VALUE;
     }
     TopicManager tm = getTopicManager(pubSubServerName);
-    long endOffset = tm.getLatestOffsetCached(topic, partition) - 1;
+    long endOffset = tm.getLatestOffsetCached(topic, partition);
```
sushantmane, line 2419: what if we returned lag `0` if endOffset is `0`?

FelixGV: So... I guess the rationale is that if the end offset is zero, then it's an empty partition, and therefore there must be zero lag. We can do that if you think it makes the code clearer. Although the rest of the function would still be the same right? (i.e. we still the -1 later on, etc.)

FelixGV: I adopted your suggestion in the latest commit, and also added a unit test.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -2412,15 +2412,36 @@ protected long measureLagWithCallToPubSub(
...
+      // -1 is a valid offset, which means that nothing was consumed yet, but anything below that is invalid.
       return Long.MAX_VALUE;
     }
-    TopicManager tm = getTopicManager(pubSubServerName);
-    long endOffset = tm.getLatestOffsetCached(topic, partition) - 1;
+    TopicManager tm = topicManagerProvider.apply(pubSubServerName);
+    long endOffset = tm.getLatestOffsetCached(topic, partition);
     if (endOffset < 0) {
+      // A negative value means there was a problem in measuring the end offset, and therefore we return "infinite lag"
       return Long.MAX_VALUE;
```
gaojieliu, line 2432: Shall we add a warn log here as it is unexpected?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -2412,15 +2412,36 @@ protected long measureLagWithCallToPubSub(
...
+    } else if (endOffset == 0) {
```
sixpluszero, line 2433: Why do we have this branch? I think it can be merged into below branch. EndOffset = 0 then currentOffset must be -1? then below formula also applies

FelixGV: That is true. It was a suggestion from @sushantmane. It is not strictly necessary, but might be easier to read for some (maybe not all...?) people.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -2412,15 +2412,36 @@ protected long measureLagWithCallToPubSub(
...
+      // The topic is empty and therefore, by definition, there cannot be any lag
+      return 0;
     }
-    return endOffset - currentOffset;
+
+    /**
+     * An empty topic should have an end offset of zero. A topic with a single message in it will have an end offset of
```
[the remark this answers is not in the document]
FelixGV: So I guess it would be more accurate to say that a topic with end offset of zero has always been empty. Whereas a topic which had messages in it before may now be empty (due either to time retention or to log compaction where all keys have gotten a null value [which we do not do, but anyway...]). In such case where a topic has had messages before but is empty now, the end offset would be > 0.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
FelixGV: There is already an `ERROR` log in `TopicMetadataFetcher::getLatestOffsetCached` when returning the negative error code, so I would rather not add another one.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
sushantmane: After thinking about this more, I think maybe we should remove this branch.  Sorry for the trouble

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
gaojieliu: SG

sushantmane (approved):
LGTM. Thanks, @FelixGV!

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
FelixGV: IMHO, I think it's fine to keep it, as it helps structure the comments to explain what each offset state means. Though the code could be one instruction shorter and achieve the same outcome, I don't think it matters in this case from a performance standpoint.

sixpluszero (approved):

PR #912, opened 2024-03-22
[dvc][server]Make recordNearlineLocalBrokerToReadyToServeLatency ignore catchup messages after push

m-nagarajan:
## Summary
While measuring nearline latency from local broker to ready to serve, the messages catching up after a push records longer latencies to these metrics. This changes records `lagCaughtUpTimeInMs` and considers the nearline messages that were ingested only after that time to ignore the already ingested messages. 

## How was this PR tested?
GH CI

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/PartitionConsumptionState.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/PartitionConsumptionState.java
@@ -48,9 +48,11 @@ public class PartitionConsumptionState {
   private boolean deferredWrite;
   private boolean errorReported;
   private boolean lagCaughtUp;
-  // record the time when the lag is caught up to record nearline metrics
-  // for messages produced after this time to ignore the messages that are
-  // getting caught up after a push
+  /**
+   * Save the time when the lag is caught up to record nearline metrics
+   * for messages produced after this time to ignore the old messages that are
+   * getting caught up after a push.
+   */
   private long lagCaughtUpTimeInMs;
   private boolean completionReported;
   private boolean isSubscribed;
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/PartitionConsumptionState.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/PartitionConsumptionState.java
@@ -48,6 +48,10 @@ public class PartitionConsumptionState {
   private boolean deferredWrite;
   private boolean errorReported;
   private boolean lagCaughtUp;
+  // record the time when the lag is caught up to record nearline metrics
+  // for messages produced after this time to ignore the messages that are
+  // getting caught up after a push
+  private long lagCaughtUpTimeInMs;
   private boolean completionReported;
   private boolean isSubscribed;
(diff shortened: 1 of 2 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/PartitionConsumptionState.java
```diff
@@ -48,6 +48,10 @@ public class PartitionConsumptionState {
   private boolean deferredWrite;
   private boolean errorReported;
   private boolean lagCaughtUp;
+  // record the time when the lag is caught up to record nearline metrics
+  // for messages produced after this time to ignore the messages that are
+  // getting caught up after a push
+  private long lagCaughtUpTimeInMs;
```
FelixGV, line 54: Nitpick: let's use JavaDoc for field documentation.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -3815,21 +3815,23 @@ private void recordNearlineLocalBrokerToReadyToServerLatency(
...
+     * Record nearline latency only when it's a hybrid store, the lag has been caught up and ignore
+     * messages that are getting caughtup. Sometimes the producerTimestamp can be -1 if the
+     * leaderProducedRecordContext had an error after callback. Don't record latency for invalid timestamps.
      */
     if (!isUserSystemStore() && isHybridMode() && partitionConsumptionState.hasLagCaughtUp()) {
-      long afterProcessingRecordTimestampMs = System.currentTimeMillis();
       long producerTimestamp = (leaderProducedRecordContext == null)
           ? kafkaMessageEnvelope.producerMetadata.messageTimestamp
           : leaderProducedRecordContext.getProducedTimestampMs();
       if (producerTimestamp > 0) {
```
FelixGV, line 3826: Nitpick: I think you could replace this check entirely with `partitionConsumptionState.isNearlineMetricsRecordingValid` since that function includes `producerTimeStampInMs > lagCaughtUpTimeInMs` which is equivalent... so no need for two nested ifs.

FelixGV (commented):
Thanks for the fix. Just left two nitpicks but it LGTM overall.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
m-nagarajan: Else part of this logs the invalid timestamps. Didn't want to log the else case for `isNearlineMetricsRecordingValid`, so added a nesting here. What do you think?

m-nagarajan (commented):
Thanks @FelixGV for the review.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
FelixGV: Oh yeah, sure, I missed that.

FelixGV (approved):
Thanks!

PR #913, opened 2024-03-25
[vpj] Allow configuring explicit TTL age in VPJ

nisargthakkar:
## Allow configuring explicit TTL age in VPJ

Currently, [Repush with TTL](https://venicedb.org/docs/user_guide/ttl#repush-with-ttl) uses the store's `rewindTimeInSeconds` hybrid config  as the TTL time for eviction. Some use-cases might require configurable TTL times through configs on their Venice Push Jobs. This commit adds a new config `repush.ttl.seconds` that can be set to specify such TTL configs.

This commit also improves some unrelated logging in VPJ.

## How was this PR tested?

GHCI

### DIFF
--- a/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
+++ b/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
@@ -2087,6 +2087,8 @@ public class VenicePushJob implements AutoCloseable {
           jobSetting.repushTTLStartTimeMs =
               pushJobSetting.jobStartTimeMs - (storeRewindTimeInSeconds * Time.MS_PER_SECOND);
         }
+
+        LOGGER.info("Will evict records older than epoch time: {} ms", jobSetting.repushTTLStartTimeMs);
       }
     }
 
--- a/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
+++ b/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
@@ -58,6 +58,7 @@ import static com.linkedin.venice.hadoop.VenicePushJobConstants.POLL_STATUS_RETR
 import static com.linkedin.venice.hadoop.VenicePushJobConstants.POST_VALIDATION_CONSUMPTION_ENABLED;
 import static com.linkedin.venice.hadoop.VenicePushJobConstants.PUSH_JOB_STATUS_UPLOAD_ENABLE;
 import static com.linkedin.venice.hadoop.VenicePushJobConstants.REPUSH_TTL_ENABLE;
+import static com.linkedin.venice.hadoop.VenicePushJobConstants.REPUSH_TTL_SECONDS;
 import static com.linkedin.venice.hadoop.VenicePushJobConstants.REPUSH_TTL_START_TIMESTAMP;
 import static com.linkedin.venice.hadoop.VenicePushJobConstants.REWIND_EPOCH_TIME_BUFFER_IN_SECONDS_OVERRIDE;
 import static com.linkedin.venice.hadoop.VenicePushJobConstants.REWIND_EPOCH_TIME_IN_SECONDS_OVERRIDE;
@@ -144,6 +145,7 @@ import com.linkedin.venice.utils.AvroSupersetSchemaUtils;
 import com.linkedin.venice.utils.ByteUtils;
 import com.linkedin.venice.utils.DictionaryUtils;
 import com.linkedin.venice.utils.EncodingUtils;
+import com.linkedin.venice.utils.LatencyUtils;
 import com.linkedin.venice.utils.PartitionUtils;
 import com.linkedin.venice.utils.ReflectUtils;
 import com.linkedin.venice.utils.RegionUtils;
@@ -221,7 +223,6 @@ public class VenicePushJob implements AutoCloseable {
   // Total input data size, which is used to talk to controller to decide whether we have enough quota or not
   private InputDataInfoProvider.InputDataInfo inputDataInfo;
 
-  private long jobStartTimeMs;
   private Properties veniceWriterProperties;
   private JobClientWrapper jobClientWrapper;
   private SentPushJobDetailsTracker sentPushJobDetailsTracker;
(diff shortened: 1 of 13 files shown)


clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
```diff
@@ -360,6 +362,30 @@ private PushJobSetting getPushJobSetting(VeniceProperties props) {
...
+            "Both " + REPUSH_TTL_SECONDS + " and " + REPUSH_TTL_START_TIMESTAMP + " are set. Please set only one.";
+        throw new VeniceException(message);
+      }
+
+      if (repushTtlSeconds >= 0) {
+        pushJobSettingToReturn.repushTTLStartTimeMs =
+            pushJobSettingToReturn.jobStartTimeMs - (repushTtlSeconds * Time.MS_PER_SECOND);
+      } else if (repushTtlStartTimestamp >= 0) {
+        pushJobSettingToReturn.repushTTLStartTimeMs = repushTtlStartTimestamp;
+      }
```
xunyin8, line 386: should be log the final `repushTTLStartTimeMs` for information and debugging purpose?

xunyin8 (dismissed):
LGTM, one minor NIT. Feel free to drop if you think it's unnecessary.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
nisargthakkar: Ah, yes. I was thinking we should, but then I forgot to log it 🤦

sixpluszero:
Could you explain a bit for "some use-cases might require configurable TTL times"? I remember we have such thing `repush.ttl.start.timestamp` as a config to expose the start time?

nisargthakkar:
>
Could you explain a bit for "some use-cases might require configurable TTL times"? I remember we have such thing `repush.ttl.start.timestamp` as a config to expose the start time?

Sure. In `repush.ttl.start.timestamp`, users need to specify the exact start timestamp they would like to use. The problem with this is this is hard to set in orchestration systems (like Azkaban) where the job configs are statically configured at build time, rather than being computed programmatically.

With the new config, users can specify a TTL time on their VPJ jobs. The main benefit I can think of currently is that it can help reduce the reliance on the store config. The problem with having a very high rewind time store config is that we end up using that rewind time value as the retention for the RT topic. Also, due to compliance reasons, it could end up restricting the amount of rewind that TTL jobs can support without setting an explicit start time.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
```diff
@@ -360,6 +362,30 @@ private PushJobSetting getPushJobSetting(VeniceProperties props) {
...
+      if (repushTtlSeconds >= 0 && repushTtlStartTimestamp >= 0) {
+        String message =
+            "Both " + REPUSH_TTL_SECONDS + " and " + REPUSH_TTL_START_TIMESTAMP + " are set. Please set only one.";
+        throw new VeniceException(message);
+      }
+
+      if (repushTtlSeconds >= 0) {
+        pushJobSettingToReturn.repushTTLStartTimeMs =
+            pushJobSettingToReturn.jobStartTimeMs - (repushTtlSeconds * Time.MS_PER_SECOND);
+      } else if (repushTtlStartTimestamp >= 0) {
```
majisourav99, line 384: No `else if` needed, right?  may be a ternary op:

`pushJobSettingToReturn.repushTTLStartTimeMs = repushTtlStartTimestamp >= 0 ? repushTtlStartTimestamp : pushJobSettingToReturn.jobStartTimeMs - (repushTtlSeconds * Time.MS_PER_SECOND)

sixpluszero:
> > Could you explain a bit for "some use-cases might require configurable TTL times"? I remember we have such thing `repush.ttl.start.timestamp` as a config to expose the start time?
> 
> Sure. In `repush.ttl.start.timestamp`, users need to specify the exact start timestamp they would like to use. The problem with this is this is hard to set in orchestration systems (like Azkaban) where the job configs are statically configured at build time, rather than being computed programmatically.
> 
> With the new config, users can specify a TTL time on their VPJ jobs. The main benefit I can think of currently is that it can help reduce the reliance on the store config. The problem with having a very high rewind time store config is that we end up using that rewind time value as the retention for the RT topic. Since RT topic doesn't have compaction, it means that it would hold data for longer durations, and that could have compliance issues. Hence, it ends up restricting the amount of rewind that TTL jobs can support when specifying a TTL age.

Sure. I acknowledge having an override is good, as it provides flexibility for operator.  
But just for my knowledge, if as you mentioned having high rewind time will lead to compliance issue, in practice should we ask them to update the rewind time, instead of setting it in the job conf?

docs/user_guide/ttl.md
```diff
@@ -17,11 +17,42 @@ minimum, not maximum, TTL. It is also advisable to schedule the periodic purging
...
+The major differences between both approaches is the original data source and how real-time buffer is replayed.
 
-There are two ways to achieve TTL compliance, via [empty push](#empty-push) or via [Repush with TTL](#experimental-repush-with-ttl).
+|                 | Source Data for TTL    | Real-time buffer replay                                               |
+|-----------------|------------------------|-----------------------------------------------------------------------|
+| Empty push      | None                   | Replay real-time buffer with `hybrid-rewind-seconds` config           |
+| Repush with TTL | Existing version topic | Replay real-time buffer with `rewind.time.in.seconds.override` config |
 
-## Empty push
+## Repush with TTL
```
majisourav99, line 28: Thank you for updating the doc as well!  Curious why are we allowing the store rewind time and ttl period to be separately configured ? If for compliance then the rewind time would also cause problem ?

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
nisargthakkar: There are 3 ways to configure `repushTTLStartTimeMs` - 
1. `repush.ttl.seconds` (Optional)
2. `repush.ttl.start.timestamp` (Optional)
3. Store rewind time config

This piece of code in this thread runs during config time - and not run method (for improving testability) and checks for Points 1 and 2. In the `run` method, before setting the final job config on the compute engine, we check if `jobStartTimeMs` is still `<0` (unset), then we use the one from hybrid configs.

docs/user_guide/ttl.md
nisargthakkar: Why does TTL period need to be configured through "rewind time"? What if use cases want 1 year of TTL? GDPR compliance would only allow a max of 30 days. So, that restricts repush from supporting 1 year TTLs because we only use the same store config for TTL and topic retention. This config allows us to disambiguate the two things.

majisourav99 (approved):

nisargthakkar:
Thanks @xunyin8, @majisourav99 and @sixpluszero

PR #914, opened 2024-03-25
[dvc] Create snapshots after batch update

ishwarya-citro:
## Create snapshots after batch update

After a batch update, we want to create a database snapshot. The snapshots will be under the database directory in the "./snapshot" folder for each batch-push version. The "./snapshot" folder includes the SST files, which will be used for the blob transfer.

## How was this PR tested?

Integration / Unit Tests

Created ".snapshot_files" folder: 
Screenshot 2024-03-25 at 4 04 59 PM

Snapshot Created:
Screenshot 2024-03-21 at 11 54 50 AM


clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBSstFileWriter.java
```diff
@@ -77,14 +80,20 @@ public ReusableObjects() {
...
   private Optional<Supplier<byte[]>> expectedChecksumSupplier;
   private final String storeName;
   private final int partitionId;
   private final EnvOptions envOptions;
   private final Options options;
   private final boolean isRMD;
+
   private final RocksDBServerConfig rocksDBServerConfig;
 
+  protected Checkpoint makeCheckpoint(RocksDB rocksDB) {
```
adamxchen, line 93: nit: how about renaming it to `createCheckPoint`? and might also give a comment saying this is testing-only since I see you mocked its output in the unit test

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBSstFileWriter.java
```diff
@@ -490,6 +500,24 @@ public void ingestSSTFiles(RocksDB rocksDB, List<ColumnFamilyHandle> columnFamil
     }
   }
 
+  public void createSnapshot(RocksDB rocksDB) {
+    try {
+      Checkpoint checkpoint = makeCheckpoint(rocksDB);
+      String checkpointPath = this.fullPathForPartitionDBSnapshot;
```
adamxchen, line 506: The value doesn't change so do we have to re-assign the variables? Is it for making the codes shorter?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBSstFileWriter.java
```diff
@@ -490,6 +500,24 @@ public void ingestSSTFiles(RocksDB rocksDB, List<ColumnFamilyHandle> columnFamil
...
+
+      if (Files.exists(Paths.get(checkpointPath))) {
```
adamxchen, line 508: What's the circumstances the snapshot directory already exist? Is it the second push? If so, what's the expected behavior?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBSstFileWriter.java
```diff
@@ -490,6 +500,24 @@ public void ingestSSTFiles(RocksDB rocksDB, List<ColumnFamilyHandle> columnFamil
...
+        return;
+      }
+
+      LOGGER.info("Start creating snapshots in directory: {}", checkpointPath);
+      checkpoint.createCheckpoint(checkpointPath);
+
+      LOGGER.info("Finished creating snapshots in directory: {}", checkpointPath);
+    } catch (RocksDBException e) { // | IOException e) {
```
adamxchen, line 516:
```
 // | IOException e) {
``` 
Do we have to add it back?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBSstFileWriter.java
```diff
@@ -490,6 +500,24 @@ public void ingestSSTFiles(RocksDB rocksDB, List<ColumnFamilyHandle> columnFamil
...
+      throw new VeniceException("Received exception during RocksDB's snapshot creation", e);
```
adamxchen, line 517: Can you please add a little more details in the exception? Like what's the path so it's easier to debug using the exception msg. (IDK if `e` contains it or not. If yes, feel free to discard this comment)

clients/da-vinci-client/src/test/java/com/linkedin/davinci/store/rocksdb/RocksDBSstFileWriterTest.java
```diff
@@ -1,8 +1,11 @@
 package com.linkedin.davinci.store.rocksdb;
 
+import static org.mockito.Mockito.*;
```
adamxchen, line 3: please do not import all. That's a config we can tweak in the Intellij.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/ReplicationMetadataRocksDBStoragePartition.java
```diff
@@ -206,6 +206,7 @@ public synchronized void endBatchWrite() {
 
     if (deferredWrite) {
       rocksDBSstFileWriter.ingestSSTFiles(rocksDB, getColumnFamilyHandleList());
+      rocksDBSstFileWriter.createSnapshot(rocksDB);
```
adamxchen, line 209: We need to control this new behavior, including some codes below, via a feature flag so we won't create snapshot for every stores.. I think we can use `StoragePartitionConfig` or `VeniceStoreVersionConfig` for this. Can you take a look and add that?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBSstFileWriter.java
```diff
@@ -106,6 +115,7 @@ public RocksDBSstFileWriter(
     this.envOptions = envOptions;
     this.options = options;
     this.fullPathForTempSSTFileDir = fullPathForTempSSTFileDir;
+    this.fullPathForPartitionDBSnapshot = RocksDBUtils.composeSnapshotDir(dbDir, storeName, partitionId);
```
adamxchen, line 118: Plz see my first comment. If this feature is not enabled, we don't need to create the path for this store.

clients/da-vinci-client/src/test/java/com/linkedin/davinci/store/rocksdb/RocksDBSstFileWriterTest.java
```diff
@@ -268,6 +273,40 @@ public void testOpenAndSyncWithAllValidFiles() throws IOException {
     }
   }
 
+  @Test
+  public void testCreateSnapshot() throws RocksDBException {
```
adamxchen, line 277: Can we also have a test to ensure it catches exceptions when create snapshot?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/ReplicationMetadataRocksDBStoragePartition.java
adamxchen: Also, you probably don't need it here since `ReplicationMetadataRocksDBStoragePartition` extends `RocksDBStoragePartition` and you have added the behavior in `RocksDBStoragePartition#endBatchWrite`

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestBatch.java
```diff
@@ -23,6 +23,7 @@
 import static com.linkedin.venice.utils.IntegrationTestPushUtils.createStoreForJob;
 import static com.linkedin.venice.utils.IntegrationTestPushUtils.defaultVPJProps;
 import static com.linkedin.venice.utils.IntegrationTestPushUtils.updateStore;
+import static com.linkedin.venice.utils.TestUtils.*;
```
adamxchen, line 26: Ditto

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestBatch.java
```diff
@@ -650,6 +653,41 @@ public void testKafkaInputAAStore() throws Exception {
...
+      }
+    };
+
+    String storeName = testBatchStore(
+        inputDir -> new KeyAndValueSchemas(writeSimpleAvroFileWithStringToStringSchema(inputDir)),
+        properties -> {},
+        validator);
+
+    // Re-push with Kafka Input
+    testRepush(storeName, validator);
```
adamxchen, line 671: It's good to test Kafka Input but can we have a test that's for regular batch too? You can use a boolean like other test methods to control which type of push jobs you want to kick off.

internal/venice-test-common/src/main/java/com/linkedin/venice/utils/TestUtils.java
```diff
@@ -897,4 +899,50 @@ public static void checkMissingFieldInAvroRecord(GenericRecord record, String fi
...
+        result.add(file.getParent()); // Add the parent directory to the result list
+        break; // Stop searching this directory once we find a file with the desired extension
+      }
+    }
+    return result;
+  }
+
+  public static boolean directoryContainsFolder(String directoryPath, String folderName) {
+    File directory = new File(directoryPath);
+    if (directory.exists() && directory.isDirectory()) {
```
adamxchen, line 936: AFIAK, `isDirectory` checks whether it exists or not, the `directory.exists()` is redundant. We can remove it

internal/venice-test-common/src/main/java/com/linkedin/venice/utils/TestUtils.java
```diff
@@ -897,4 +899,50 @@ public static void checkMissingFieldInAvroRecord(GenericRecord record, String fi
...
+    if (!directory.exists() || !directory.isDirectory()) {
+      return result;
+    }
+    return searchForFileExtension(directory, fileExtension.toLowerCase());
+  }
+
+  public static List<String> searchForFileExtension(File directory, String fileExtension) {
+    List<String> result = new ArrayList<>();
+    if (!directory.canRead()) {
+      System.err.println("Cannot read directory: " + directory.getAbsolutePath());
```
adamxchen, line 915: Let's use `LOGGER` instead of `System.err`

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestBatch.java
```diff
@@ -650,6 +653,41 @@ public void testKafkaInputAAStore() throws Exception {
...
+
+    Path folderPath1 = Paths.get(BASE_DATA_PATH_1);
```
adamxchen, line 673: Can we move this line up to 669 and verify there's snapshots created for version 1 but there's nothing for `BASE_DATA_PATH_2`? Then after line 674, we verify BASE_DATA_PATH_2 also has snapshots created so it validates the full lifecycle of regular push and repush.

internal/venice-test-common/src/main/java/com/linkedin/venice/utils/TestUtils.java
```diff
@@ -897,4 +899,50 @@ public static void checkMissingFieldInAvroRecord(GenericRecord record, String fi
...
+      return result;
+    }
+    File[] files = directory.listFiles();
+    if (files == null) {
+      System.err.println("Error reading directory: " + directory.getAbsolutePath());
```
adamxchen, line 920: Ditto.

adamxchen (changes requested):
Great progress! Please take a look at the comments and iterate this PR!

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestBatch.java
ishwarya-citro: hmm. It looks like `BASE_DATA_PATH_1` and `BASE_DATA_PATH_2` belong to different venice server ([1](https://github.com/linkedin/venice/blob/main/internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestBatchForIngestionIsolation.java#L32) and [2](https://github.com/linkedin/venice/blob/main/internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestBatchForIngestionIsolation.java#L34)).  

I think you mean to test the following ? - 
1. testBatchStore (first push) : make sure we have v1 w/ snapshots but not any other versions 
2. testRepush (consecutive pushes) : make sure we have v2 w/ snapshots

i can cover this test case but pl lmk if thats not what you mean

adamxchen: Yes, that's what I was asking for. I thought these 2 paths are for different versions but I could be wrong. If you can do that, it'd be great.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBSstFileWriter.java
adamxchen: I misread some codes when I made the comment and forgot to delete this.
I meant to ask when we have new pushes, would the content under the same path get overwritten? If yes, there's no concern

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBServerConfig.java
```diff
@@ -1,5 +1,7 @@
 package com.linkedin.davinci.store.rocksdb;
 
+import static com.linkedin.venice.ConfigKeys.*;
```
adamxchen, line 3: there's a Intellij config we can change to avoid it being too smart to import all. That's `Named count to use static import with '*'`

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBServerConfig.java
```diff
@@ -375,6 +378,8 @@ public RocksDBServerConfig(VeniceProperties props) {
     this.atomicFlushEnabled = props.getBoolean(ROCKSDB_ATOMIC_FLUSH_ENABLED, true);
     this.separateRMDCacheEnabled = props.getBoolean(ROCKSDB_SEPARATE_RMD_CACHE_ENABLED, false);
 
+    this.blobTransferBatchOnlyEnabled = props.getBoolean(ENABLE_BLOB_TRANSFER_FOR_BATCH_ONLY_STORE, false);
```
adamxchen, line 381: This config doesn't seem to be right. This `RocksDBServerConfig` is for host-level configs not store-level. Would `StoragePartitionConfig` or `VeniceStoreVersionConfig` work?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/ReplicationMetadataRocksDBStoragePartition.java
adamxchen: I am not sure I follow. When `deferredWrite` is false, the method returns for both `ReplicationMetadataRocksDBStoragePartition` and `RocksDBStoragePartition`. When it is true, we end up calling `createSnapshot` twice.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestBatch.java
adamxchen: It comes back again.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBSstFileWriter.java
ishwarya-citro: removed this

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBServerConfig.java
ishwarya-citro: sry will change the intellij setting now

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBServerConfig.java
ishwarya-citro: I didn't find the docs clear on what the configs were used for. I used `RocksDBServerConfig` because the [RocksDBSstFileWriter](https://github.com/linkedin/venice/blob/ca88d722de394beea99371486e294025574546a1/clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBSstFileWriter.java#L121) already had access to i, and we wouldn't have to make so many changes. I can refactor to use `StoragePartitionConfig` or `VeniceStoreVersionConfig`

adamxchen: Please consider using latter two since they are more close to `store-level` config. Though `RocksDBServerConfig` is easy to access w/o touching many codes, this is just not the right config and can impact code readability/maintainability

clients/da-vinci-client/src/main/java/com/linkedin/davinci/config/VeniceStoreVersionConfig.java
```diff
@@ -30,13 +31,15 @@ public class VeniceStoreVersionConfig extends VeniceServerConfig {
...
   private boolean restoreMetadataPartition = true;
+  private boolean blobTransferBatchOnlyEnabled;
 
   public VeniceStoreVersionConfig(
       String storeVersionName,
       VeniceProperties storeProperties,
       Map<String, Map<String, String>> kafkaClusterMap) {
     super(storeProperties, kafkaClusterMap);
     this.storeVersionName = storeVersionName;
+    this.blobTransferBatchOnlyEnabled = storeProperties.getBoolean(ENABLE_BLOB_TRANSFER_FOR_BATCH_ONLY_STORE, false);
```
adamxchen, line 42: Is it possible that we add the validation somewhere so when the config is enabled for a store that is not compatible, we might want to throw exception.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBSstFileWriter.java
```diff
@@ -100,15 +109,18 @@ public RocksDBSstFileWriter(
...
       boolean isRMD,
-      RocksDBServerConfig rocksDBServerConfig) {
+      RocksDBServerConfig rocksDBServerConfig,
+      boolean blobTransferBatchOnlyEnabled) {
     this.storeName = storeName;
     this.partitionId = partitionId;
     this.envOptions = envOptions;
     this.options = options;
     this.fullPathForTempSSTFileDir = fullPathForTempSSTFileDir;
+    this.fullPathForPartitionDBSnapshot = RocksDBUtils.composeSnapshotDir(dbDir, storeName, partitionId);
```
adamxchen, line 119: if the feature is not enabled, we don't need to create the directory.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBStoragePartition.java
```diff
@@ -453,6 +468,12 @@ public synchronized void endBatchWrite() {
      * the last SST file written is finished.
      */
     rocksDBSstFileWriter.ingestSSTFiles(rocksDB, columnFamilyHandleList);
+    createSnapshot();
+  }
+
+  @Override
+  public synchronized void createSnapshot() {
+    rocksDBSstFileWriter.createSnapshot(rocksDB);
```
adamxchen, line 476: We should addd a check here too. If the feature is not enabled, calling this method would result in one exception.

internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
```diff
@@ -183,6 +183,11 @@ private ConfigKeys() {
   public static final String ENABLE_ACTIVE_ACTIVE_REPLICATION_AS_DEFAULT_FOR_HYBRID_STORE =
       "enable.active.active.replication.as.default.for.hybrid.store";
 
+  /**
+   * Store-level config to enable blob transfers (feature flag)
+   */
+  public static final String ENABLE_BLOB_TRANSFER_FOR_BATCH_ONLY_STORE = "enable.blob.transfer.for.batch.only.store";
```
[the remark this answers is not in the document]
adamxchen: Actually one more thought is since this is a store-level property, we need to add it to [StoreMetaValue](https://github.com/linkedin/venice/blob/f07d0c13edcd64e730a88925390b4c739f4ac4f5/internal/venice-common/src/main/resources/avro/StoreMetaValue/v18/StoreMetaValue.avsc) so it also lives on ZK. What do you think?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/config/VeniceStoreVersionConfig.java
ishwarya-citro: Is this what you're asking - 
If the `ENABLE_BLOB_TRANSFER_FOR_BATCH_ONLY_STORE` is enabled for a different config type, we should throw an error?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/storage/StorageService.java
```diff
@@ -203,7 +203,8 @@ private void initInternalStorageEngineFactories() {
             serverConfig,
             rocksDBMemoryStats,
             storeVersionStateSerializer,
-            partitionStateSerializer));
+            partitionStateSerializer,
+            configLoader));
```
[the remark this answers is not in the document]
ishwarya-citro: ah, how did i miss that! thanks for the call out.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBSstFileWriter.java
```diff
@@ -49,7 +51,7 @@ private static class ReusableObjects {
     public ReusableObjects() {
       directKeyBuffer = ByteBuffer.allocateDirect(1024 * 1024);
       directValueBuffer = ByteBuffer.allocateDirect(1024 * 1024 + 128); // Adding another 128 bytes considering
-                                                                        // potential overhead of metadata
+      // potential overhead of metadata
```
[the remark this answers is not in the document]
ishwarya-citro: maybe this is a intellij setting doing this

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBStoragePartition.java
ishwarya-citro: thanks for the catch

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestBatch.java
```diff
@@ -1336,4 +1340,60 @@ public void testKafkaInputBatchJobSucceedsWhenSourceTopicIsEmpty() throws Except
...
+        Assert.assertEquals(avroClient.get(Integer.toString(i)).get().toString(), "test_name_" + i);
+      }
+    };
+
+    String storeName = testBatchStore(
+        inputDir -> new KeyAndValueSchemas(writeSimpleAvroFileWithStringToStringSchema(inputDir)),
+        properties -> {},
+        validator);
+
+    verifySnapshotFolders();
```
adamxchen, line 1358: Do you think we can clean up directories before we conduct 2nd push so we can ensure the 2nd push also creates the snapshots?

adamxchen (approved):

adamxchen:
:ship:  Thanks a lot for the hard work! I think this looks good to go
However, I’d like to check two things:
1. We would have another PR to add the config from ZK to VeniceStoreVersionConfig, right?
2. Can you please add some validation in the aforementioned PR such that when updating store/version config for blob transfer, only batch store can enable this? They can be added in `internalUpdateStore` I believe.

ishwarya-citro:
> 🚢 Thanks a lot for the hard work! I think this looks good to go However, I’d like to check two things:
> 
> 1. We would have another PR to add the config from ZK to VeniceStoreVersionConfig, right?
> 2. Can you please add some validation in the aforementioned PR such that when updating store/version config for blob transfer, only batch store can enable this? They can be added in `internalUpdateStore` I believe.

Yes! Will address them in a separate PR

PR #915, opened 2024-03-26
[producer] Skip metrics for the store client inside online producer

nisargthakkar:
## Skip metrics for the store client inside online producer

In #867, we skipped metric emission for `SchemaReader`, but we missed skipping metric emission for the store client that we create to query topic information from routers. This commit addresses this gap.

## How was this PR tested?

GH CI


gaojieliu (approved):
Thanks!

PR #916, opened 2024-03-27
[vpj] Added retries when fetching end offset in a repush job

gaojieliu:
Added retries when fetching end offset in a repush job

## How was this PR tested?

CI


majisourav99 (approved):
Thanks Gaojie

PR #917, opened 2024-03-27
[controller] Check for bad default during new schema registration

majisourav99:
## Check for bad default during new schema registration

During new store creation and new schema registration Venice does validate the union default for schemas using `AvroSchemaUtils.validateAvroSchemaStr`, but it does not check wrong numeric default types and allows wrong numeric default to be registered. This later causes ingestion error in server as it does strict validation of schemas. 
This PR will make sure controller schema addition code path does a strict validation.

## How was this PR tested?

CI test


xunyin8 (dismissed):
LGTM

xunyin8 (approved):

PR #918, opened 2024-03-27
[controller] Fix multiple AdminExecutionTasks working on the same store at the same time.

haoxu07:

Current state: when we create `AdminExecutionTask` , we did not check if there is `AdminExecutionTask` for one store on the fly, so if one `AdminExecutionTask`  is waiting for a lock (e.g. updating  store operation waiting for the store level write lock and that lock is currently unavailable), it is possible that `AdminConsumptionTask` will create multiple `AdminExecutionTask` s for a single store until they occupy all threads from `ExecutorService` 's thread pool, they are all waiting for the same store-level lock.

Besides, `executorService.invokeAll(tasks, processingCycleTimeoutInMs, TimeUnit.MILLISECONDS)`
will cancel every invoked tasks, even the task is emitted by not getting thread from pool to execute.  `ExecutorService` will try to cancel the `AdminExecutionTask`  waiting for a lock after timeout, but that will not terminate that thread (as the acquiring locking operation by `AutoCloseableLock` is not interruptible).   Then that `AdminExecutionTask`  will keep occupying one thread from `ExecutorService` 's thread pool until the lock is released. 

This PR has the following changes to address the issue:
1) Adding check to see if there is `AdminExecutionTask` is running for a store.
2) Integration test to simulate the lock blocking one thread from pool and recover after acquiring the lock.

## How was this PR tested?

CI

### DIFF
--- a/internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/controller/kafka/consumer/AdminConsumptionTaskIntegrationTest.java
+++ b/internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/controller/kafka/consumer/AdminConsumptionTaskIntegrationTest.java
@@ -36,6 +36,7 @@ import com.linkedin.venice.writer.VeniceWriter;
 import com.linkedin.venice.writer.VeniceWriterOptions;
 import java.io.IOException;
 import java.util.Properties;
+import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import org.testng.Assert;
@@ -111,7 +112,7 @@ public class AdminConsumptionTaskIntegrationTest {
   }
 
   @Test(timeOut = TIMEOUT)
-  public void testParallelAdminExecutionTasks() throws IOException {
+  public void testParallelAdminExecutionTasks() throws IOException, InterruptedException {
     try (ZkServerWrapper zkServer = ServiceFactory.getZkServer();
         PubSubBrokerWrapper pubSubBrokerWrapper = ServiceFactory.getPubSubBroker(
             new PubSubBrokerConfigs.Builder().setZkWrapper(zkServer).setRegionName(STANDALONE_REGION_NAME).build());
@@ -146,14 +147,16 @@ public class AdminConsumptionTaskIntegrationTest {
             getStoreCreationMessage(clusterName, storeName, owner, keySchema, valueSchema, executionId);
         writer.put(new byte[0], goodMessage, AdminOperationSerializer.LATEST_SCHEMA_ID_FOR_ADMIN_OPERATION);
 
-        TestUtils.waitForNonDeterministicAssertion(TIMEOUT * 3, TimeUnit.MILLISECONDS, () -> {
+        TestUtils.waitForNonDeterministicAssertion(TIMEOUT, TimeUnit.MILLISECONDS, () -> {
           Assert.assertTrue(controller.getVeniceAdmin().hasStore(clusterName, storeName));
         });
 
         // Spin up a thread to occupy the store write lock to simulate the blocking admin execution task thread.
-        Runnable infiniteLockOccupy = getRunnable(controller, storeName);
+        CountDownLatch lockOccupyThreadStartedSignal = new CountDownLatch(1);
+        Runnable infiniteLockOccupy = getRunnable(controller, storeName, lockOccupyThreadStartedSignal);
         Thread infiniteLockThread = new Thread(infiniteLockOccupy, "infiniteLockOccupy: " + storeName);
         infiniteLockThread.start();
+        Assert.assertTrue(lockOccupyThreadStartedSignal.await(5, TimeUnit.SECONDS));
 
         // Here we wait here to send every operation to let each consumer pool has at most one admin operation from
         // this store, as the waiting time of 5 seconds > ADMIN_CONSUMPTION_CYCLE_TIMEOUT_MS setting.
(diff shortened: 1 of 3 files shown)


internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/controller/kafka/consumer/AdminConsumptionTaskIntegrationTest.java
```diff
@@ -101,6 +109,103 @@ public void testSkipMessageEndToEnd() throws ExecutionException, InterruptedExce
...
+
+        // Spin up a thread to occupy the store write lock to simulate the blocking admin execution task thread.
+        Runnable infiniteLockOccupy = getRunnable(controller, storeName);
+        Thread infiniteLockThread = new Thread(infiniteLockOccupy, "infiniteLockOccupy: " + storeName);
+        infiniteLockThread.start();
+
+        for (int i = 0; i < adminConsumptionMaxWorkerPoolSize; i++) {
+          // Here we wait here to send every operation to let each consumer pool has at most one admin operation from
+          // this store.
+          Utils.sleep(5000);
```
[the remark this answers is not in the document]
haoxu07: Here we create a store update request every 5 seconds > ADMIN_CONSUMPTION_CYCLE_TIMEOUT_MS(3 seconds), which will simulate a scenario:  first store's update will not finish due to not able to acquiring the store level lock. With previous implementation with checking if there is existing `AdminExecutionTask`, previous logic will try to create one new  `AdminExecutionTask` to exhaust the `ADMIN_CONSUMPTION_MAX_WORKER_THREAD_POOL_SIZE`, since each creation will be triggered by new polling of message of each store update. In short, 5 seconds waiting with 3 seconds timeout is try to achieve: all store update message will be in separate consumer poll and invoke a check if a store has an ongoing `AdminExecutionTask`.

haoxu07:
> I have a few questions:
> 
> 1. is it expected to run some admin logic, which is not interruptible? I think we should make all the logic interruptible, otherwise, it can block the graceful shutdown.
> 2. Is there any particular type of admin message, which would stuck or take longer time than others?
> 
> We face a similar situation when implementing async gauge execution, and we adopted the following strategy, which seems to work well: [tehuti-io/tehuti#31](https://github.com/tehuti-io/tehuti/pull/31)
> 
> cc @xunyin8

Regarding your questions:
 One case for logic non-interruptible is:  our current store level lock acquiring inside `VeniceHelixAdmin` through `AutoCloseableLock` is not interruptible by using lock.lock(): https://github.com/linkedin/venice/blob/f07d0c13edcd64e730a88925390b4c739f4ac4f5/internal/venice-common/src/main/java/com/linkedin/venice/utils/locks/AutoCloseableMultiLock.java#L18. When a thread is waiting for the lock, it is not interruptible. We could improve this by using lock.lockInterruptibly(), but I found it will some test failure, which I am trying to figure out. 

But thanks for suggested solutions, let me figure out if it works.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/controller/kafka/consumer/AdminConsumptionTaskIntegrationTest.java
```diff
@@ -101,6 +110,125 @@ public void testSkipMessageEndToEnd() throws ExecutionException, InterruptedExce
...
+              pubSubBrokerWrapper.getPubSubClientsFactory().getProducerAdapterFactory();
+          VeniceWriter<byte[], byte[], byte[]> writer =
+              IntegrationTestPushUtils.getVeniceWriterFactory(pubSubBrokerWrapper, pubSubProducerAdapterFactory)
+                  .createVeniceWriter(new VeniceWriterOptions.Builder(adminTopic.getName()).build())) {
+        int executionId = 1;
+        byte[] goodMessage =
+            getStoreCreationMessage(clusterName, storeName, owner, keySchema, valueSchema, executionId);
+        writer.put(new byte[0], goodMessage, AdminOperationSerializer.LATEST_SCHEMA_ID_FOR_ADMIN_OPERATION);
+
+        TestUtils.waitForNonDeterministicAssertion(TIMEOUT * 3, TimeUnit.MILLISECONDS, () -> {
```
lluwm, line 149: Nit, the test is set to timeout after `TIMEOUT` ms, but the wait here is `3 * TIMEOUT`, I guess we can reduce it to TIMEOUT? (also for the later places)

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/controller/kafka/consumer/AdminConsumptionTaskIntegrationTest.java
```diff
@@ -101,6 +110,125 @@ public void testSkipMessageEndToEnd() throws ExecutionException, InterruptedExce
...
+          Assert.assertTrue(controller.getVeniceAdmin().hasStore(clusterName, storeName));
+        });
+
+        // Spin up a thread to occupy the store write lock to simulate the blocking admin execution task thread.
+        Runnable infiniteLockOccupy = getRunnable(controller, storeName);
+        Thread infiniteLockThread = new Thread(infiniteLockOccupy, "infiniteLockOccupy: " + storeName);
+        infiniteLockThread.start();
```
lluwm, line 156: To avoid potential race and improve the test stability, can we use a CountDownLatch as we did in e.g. in `TopicManagerTest.testContainsTopicWithExpectationAndRetry` to wait for the `infiniteLockThread` to actually in execution state before sending the disable write admin operation?

lluwm (commented):
Looks good overall, just some comments on the integration part.

lluwm:
@haoxu07, LGTM and thanks for the fix.

lluwm (approved):

PR #919, opened 2024-03-29
[admin-tool][controller][dvc][samza][test][vpj] Job status reporting to include detailed DVC push error

m-nagarajan:
## Summary
- Add new DVC error specific terminal enums in `ExecutionStatus` and similar ones in `PushJobCheckpoints` and use it instead of the generic `ExecutionStatus#ERROR` (which was/will be used for server/other task status).
     - `DVC_INGESTION_ERROR_DISK_FULL`
     - `DVC_INGESTION_ERROR_MEMORY_LIMIT_REACHED` 
     - `DVC_INGESTION_ERROR_TOO_MANY_DEAD_INSTANCES`
     - `DVC_INGESTION_ERROR_OTHER`
- Add a new property `rootStatus` in `ExecutionStatus`. `rootStatus` for these 4 new enums will be `ERROR` and for other existing enums `rootStatus` will be itself. Idea is to use status.getRootStatus() for most of the cases.
- These new enums will be returned back from DVC to backend/push job polling (via DVC push status store) and update the respective new checkpoints in VPJ. Still continue to set `PushJobDetailsStatus#ERROR` for `pushJobDetails.overallStatus`

One other implementation considered was: Add a new key in push status system store for each store version to add the specific error details and Push job monitor looks up for these extra details after it sees `ExecutionStatus#ERROR` => prints it in the VPJ logs and updates the `PushJobCheckpoints` accordingly.

`Deployment order:`
These new statuses are emitted by DaVinci hosts. Other components should be deployed before the Davinci release to be able to act on these new statuses accordingly.

## How was this PR tested?
GH CI


internal/venice-client-common/src/main/java/com/linkedin/venice/pushmonitor/ExecutionStatus.java
```diff
@@ -88,26 +88,55 @@ public enum ExecutionStatus {
...
+  /** There are too many dead DaVinci instances leading to failed push job */
+  DVC_INGESTION_ERROR_TOO_MANY_DEAD_INSTANCES(true, false, true, true, ERROR, 20),
+
+  /** Other uncategorized DaVinci Client errors */
+  DVC_INGESTION_ERROR_OTHER(true, false, true, true, ERROR, 21);
 
   final boolean isJobStatus;
   final boolean isTaskStatus;
-  final boolean isUsedByHybridStoresOnly;
+  final boolean isUsedByDaVinciClientOnly;
```
gaojieliu, line 108: Why this change?
Do we use this field: `isUsedByHybridStoresOnly` today?

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
```diff
@@ -2428,7 +2430,10 @@ void pollStatusUntilComplete(
               .append(pushJobSetting.veniceControllerUrl)
               .append("\ncontroller response: ")
               .append(response);
-
+          if (overallStatus.isDVCIngestionError()) {
+            this.pushJobDetails.pushJobLatestCheckpoint =
+                PushJobCheckpoints.valueOf(overallStatus.toString()).getValue();
```
gaojieliu, line 2435: Shall we have a try/catch here in case the DVC error is not recognized when we add more DaVinci error status, but H2V hasn't been updated yet?

gaojieliu (commented):

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
m-nagarajan: Thanks for catching this.
The code at line 2410 will fail first for the same reason. 
`      ExecutionStatus overallStatus = ExecutionStatus.valueOf(response.getStatus());`
Curious whether we should rely on deployment order here (VPJ driver before DVC client) or add some work around code. Work around only makes sense if we assume that the unknown status are DVC specific errors, if not it can be any random newly added status which can mean anything.

m-nagarajan (commented):
Thanks @gaojieliu for the review. Addressed some comments and added follow up for some.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
m-nagarajan: Thinking further, added a function `getExecutionStatusFromControllerResponse` for getting `overallStatus` which throws a detailed exception in case of mismatches.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
```diff
@@ -733,4 +738,18 @@ public void endOfIncrementalPushReceived(
       });
     }
   };
+
+  static ExecutionStatus getDaVinciErrorStatus(String message, Exception e) {
+    ExecutionStatus status = DVC_INGESTION_ERROR_OTHER;
+    if (e instanceof MemoryLimitExhaustedException) {
+      status = DVC_INGESTION_ERROR_MEMORY_LIMIT_REACHED;
+    } else if (e instanceof VeniceException) {
+      if (message.startsWith("Disk is full")) {
```
gaojieliu, line 747: Is it possible to have a specific exception like `DiskFullException` instead of relying on the message?

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
```diff
@@ -594,7 +596,7 @@ protected void setVeniceWriter(VeniceWriter<KafkaKey, byte[], byte[]> veniceWrit
   }
 
   // Visible for testing
-  protected void setSentPushJobDetailsTracker(SentPushJobDetailsTracker sentPushJobDetailsTracker) {
+  public void setSentPushJobDetailsTracker(SentPushJobDetailsTracker sentPushJobDetailsTracker) {
```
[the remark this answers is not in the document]
gaojieliu: SG

internal/venice-client-common/src/main/java/com/linkedin/venice/pushmonitor/ExecutionStatus.java
gaojieliu: We do have a few `true` for the existing enums and they are not DaVinci only:
```
  /** Tasks belonging to a Hybrid Store emits this instead of {@link ExecutionStatus#COMPLETED} when it consumes a EOP message */
  END_OF_PUSH_RECEIVED(true, true, true, false, 4),
  /** Tasks belonging to a Hybrid Store emits this instead when it consumes a SOBR message */
  @Deprecated
  START_OF_BUFFER_REPLAY_RECEIVED(false, true, true, false, 5),
  /** Tasks belonging to a Hybrid Store emits this instead when it consumes a TS message */
  TOPIC_SWITCH_RECEIVED(false, true, true, false, 6),

CATCH_UP_BASE_TOPIC_OFFSET_LAG(false, true, true, false, 13),
```

With this renaming, I assume we should update the 3rd param to be `false`, right?

gaojieliu (commented):
Left a few more minor comments and the change looks good to me overall.

internal/venice-client-common/src/main/java/com/linkedin/venice/pushmonitor/ExecutionStatus.java
m-nagarajan: Thanks for catching this. Fixed it and added small tests for these fields to reduce copy paste errors in the future.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
m-nagarajan: Good call. Added `DiskLimitExhaustedException`.

gaojieliu (approved):
Thanks!

sixpluszero (approved):
Thanks for the changes!

internal/venice-client-common/src/main/java/com/linkedin/venice/pushmonitor/ExecutionStatus.java
sushantmane, line 1: What is the motivation for changing the terminal status of `END_OF_PUSH_RECEIVED`? Wouldn’t this impact buffer replay behavior?

PR #920, opened 2024-03-29
[changelog] Add RocksDB block cache metric to changeLogClientConfig; Expose RocksDBMemoryStats to changelogConsumer

sixpluszero:
## [changelog] Add RocksDB block cache metric to changeLogClientConfig; Expose RocksDBMemoryStats to changelogConsumer

After customer adopts multiple changelog consumer, we encounter OOM and application killed by consumer. Based on current config setup, we suspect it is caused by multiple RocksDB instance opening 16GB block cache and eventually it use more memory than it should. Since we don't expose RocksDBMemoryStats today, we cannot make sure this is 100% the root cause, but from the behavior change from 1 consumer to 8 consumer with the same dataset, it is likely the cause.

Personally I think it will be great if multiple consumer can share same RocksDB instance without performance impact during bootstrap and consumption. Based on current implementation it would be hard to do so, so we will put this PR as an incremental improvement.

This PR opens up the config for user to set per consumer cache size limit. We will verify from the new metrics in the new client release.

## How was this PR tested?

Existing test.

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/ChangelogClientConfig.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/ChangelogClientConfig.java
@@ -28,9 +28,9 @@ public class ChangelogClientConfig<T extends SpecificRecord> {
   private long databaseSyncBytesInterval = 32 * 1024 * 1024L;
 
   /**
-   * RocksDB block cache size per BootstrappingVeniceChangelogConsumer. Default is 1 GB.
+   * RocksDB block cache size per BootstrappingVeniceChangelogConsumer. Default is 1 MB.
    */
-  private long rocksDBBlockCacheSizeInBytes = 1024 * 1024 * 1024L;
+  private long rocksDBBlockCacheSizeInBytes = 1024 * 1024L;
 
   public ChangelogClientConfig(String storeName) {
     this.innerClientConfig = new ClientConfig<>(storeName);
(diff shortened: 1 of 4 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/ChangelogClientConfig.java
```diff
@@ -25,7 +25,12 @@ public class ChangelogClientConfig<T extends SpecificRecord> {
    * This will be used in BootstrappingVeniceChangelogConsumer to determine when to sync updates with the underlying
    * storage engine, e.g. flushes entity and offset data to disk. Default is 32 MB.
    */
-  private long databaseSyncBytesInterval = 33554432;
+  private long databaseSyncBytesInterval = 32 * 1024 * 1024L;
+
+  /**
+   * RocksDB block cache size per BootstrappingVeniceChangelogConsumer. Default is 16 GB.
+   */
+  private long rocksDBBlockCacheSizeInBytes = 16 * 1024 * 1024 * 1024L;
```
ZacAttack, line 33: Let's.... set this to something kind of small?  Maybe even nonexistent.  The block cache for bootstrapping consumer is practically never used.  During consumption we only write to rocksdb, and when we read the rocksdb, it'll be after a restart, so the block cache will be cold.

In fact.  I wonder if we shouldn't just turn it off

ZacAttack (approved):
Awesome, thanks!

PR #921, opened 2024-03-29
[controller] get future version for finished pushes

majisourav99:
## Get future version for finished pushes.

Currently list_future_versions works only for ongoing pushes, if the push finishes then it will return invalid version. But in colo-by-colo push workflow, finished pushes not imply that is no future version as current version may not have swapped to it (deferredVersionsSwap = true). This PR will return the largest online version larger than current version as future version.

## How was this PR tested?

CI test.


huangminchn (approved):
Thanks Sourav! LGTM!

PR #922, opened 2024-03-29
[Server] Keep producerStateMaxAgeMs unchanged if time-based clearing of expired DIV state is disabled

sushantmane:
## How was this PR tested?
CI


majisourav99 (commented):
Can we make a test for this in OSS ?

sushantmane:
> Can we make a test for this in OSS ?

SIT is very convoluted to write UTs.. so would prefer skipping it

majisourav99 (approved):
Thanks for the quick fix

sushantmane:
Thanks for the review, @majisourav99!

PR #924, opened 2024-04-01
[dvc] Removed DVC remote query feature

gaojieliu:
As it is hard to predict the DVC KPS and we don't have a valid use case, which would query remote Venice on purpose, this PR removes this feature completely after marking it as deprecated for some time.

Resolves #XXX

## How was this PR tested?
CI

Customers shouldn't rely on the remote query feature while using DaVinci client.


clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/NonLocalAccessPolicy.java
```diff
@@ -1,21 +0,0 @@
-package com.linkedin.davinci.client;
-
-public enum NonLocalAccessPolicy {
```
nisargthakkar, line 3: Let's leave this enum as is and mark it deprecated instead. Once we've gotten everyone to migrate off, we can remove it. Otherwise, it becomes a breaking change.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/DaVinciConfig.java
```diff
@@ -99,15 +84,6 @@ public DaVinciConfig setStorageClass(StorageClass storageClass) {
...
   }
 
-  public NonLocalAccessPolicy getNonLocalAccessPolicy() {
-    return nonLocalAccessPolicy;
-  }
-
-  public DaVinciConfig setNonLocalAccessPolicy(NonLocalAccessPolicy nonLocalAccessPolicy) {
-    this.nonLocalAccessPolicy = nonLocalAccessPolicy;
-    return this;
-  }
```
nisargthakkar, line 109: Let's leave these two functions as is and mark them deprecated instead. We can leave the setter empty and the getter return `FAIL_FAST`.

Once we've gotten everyone to migrate off, we can remove it. Otherwise, it becomes a breaking change.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/NonLocalAccessPolicy.java
gaojieliu: We will bump the major version to force customer to get rid of the usage.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/DaVinciConfig.java
gaojieliu: We will bump the major version to force customer to get rid of the usage.

nisargthakkar (approved):

PR #925, opened 2024-04-01
[server][da-vinci-client] Skip updating pushstatus if CV update fails

majisourav99:
## Skip updating pushstatus if CV update fails

If CV update fails after a push completion, updating the pushstatus would cause error requests as router uses CV to send request to servers. This PR changes the order of push notifier to update CV before offline push status and if CV update fails, it skips updating the offline push status.


clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/HelixParticipationService.java
```diff
@@ -392,6 +391,7 @@ private void asyncStart() {
        * the notifier is added.
        */
       partitionPushStatusAccessorFuture.complete(partitionPushStatusAccessor);
+      ingestionBackend.addPushStatusNotifier(pushMonitorNotifier);
```
gaojieliu, line 394: Why this change?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/IngestionNotificationDispatcher.java
```diff
@@ -72,6 +73,10 @@ private void report(
       try {
         function.apply(notifier);
       } catch (Exception ex) {
+        if (notifier instanceof PartitionPushStatusNotifier) {
```
gaojieliu, line 76: The ideal solution should be letting Controller look at CV instead of `OfflinePushStatus` as `CV` is being used as the SoT for routing.
I wonder how hard it is to switch to CV in Controller for batch push.
Also this fix is a little error-prone as the user can attach the notifiers in a different order and if we decide to go with this approach, I think we need to consolidate offline push status notifier and CV notifier into one wrapper class and always make sure CV error won't mark offline push status as completed.

gaojieliu (changes requested):

gaojieliu (commented):
Code change looks good overall.
Shall we remove `PushMonitorNotifier` as well? I know `LeaderErrorNotifier` inherits `PushMonitorNotifier` and I guess we can update it to inherit the new `PushStatusNotifier`.

Also let us cut a ticket to track the task to switch to CV in Controller as two SoT might cause other unexpected issues in the future.

gaojieliu (approved):
Thanks!

PR #926, opened 2024-04-02
[WIP][server] Calculate upstream offset from TopicSwitch if needed when transitioning into LEADER

sixpluszero:
## [server] Calculate upstream offset from TopicSwitch if needed when transitioning into LEADER

PR #908 fixed the bug that follower may calculate and store wrong rewind offset if some RT messages are gone because of message retention.
However, the fix is not complete. It missed the case during leader transition: The to-be-leader will store TS (if it sees one), it will update the leader topic, but it will not calculate the rewind and persist it after the PR change. The result is that: when it becomes leader after 5 min inactive time, it will start from -1, as no upstream offset has been computed. 

This PR proposes to fix the case by adding the same offset calculation logic to API startConsumingAsLeader.

It happens to improve the resiliency of AA subscription as well by reusing the same code. As in the past, the to-be-leader will not re-try calculating offset for unreachable broker via repair task.

## How was this PR tested?

Not yet. I am planning to add an integration test to attack this scenario.

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
@@ -826,6 +826,90 @@ public class ActiveActiveStoreIngestionTask extends LeaderFollowerStoreIngestion
   }
 
   @Override
+  protected Map<String, Long> calculateLeaderUpstreamOffsetWithTopicSwitch(
+      PartitionConsumptionState partitionConsumptionState,
+      TopicSwitch topicSwitch,
+      PubSubTopic newSourceTopic,
+      List<CharSequence> unreachableBrokerList) {
+    Map<String, Long> upstreamOffsetsByKafkaURLs = new HashMap<>(topicSwitch.sourceKafkaServers.size());
+    final PubSubTopicPartition sourceTopicPartition = partitionConsumptionState.getSourceTopicPartition(newSourceTopic);
+    topicSwitch.sourceKafkaServers.forEach(sourceKafkaURL -> {
+      Long upstreamStartOffset =
+          partitionConsumptionState.getLatestProcessedUpstreamRTOffsetWithNoDefault(sourceKafkaURL.toString());
+      if (upstreamStartOffset == null || upstreamStartOffset < 0) {
+        long rewindStartTimestamp;
+        // calculate the rewind start time here if controller asked to do so by using this sentinel value.
+        if (topicSwitch.rewindStartTimestamp == REWIND_TIME_DECIDED_BY_SERVER) {
+          rewindStartTimestamp = calculateRewindStartTime(partitionConsumptionState);
+          LOGGER.info(
+              "{} leader calculated rewindStartTimestamp {} for {}",
+              ingestionTaskName,
+              rewindStartTimestamp,
+              sourceTopicPartition);
+        } else {
+          rewindStartTimestamp = topicSwitch.rewindStartTimestamp;
+        }
+        if (rewindStartTimestamp > 0) {
+          PubSubTopicPartition newSourceTopicPartition =
+              new PubSubTopicPartitionImpl(newSourceTopic, sourceTopicPartition.getPartitionNumber());
+          try {
+            upstreamStartOffset =
+                getTopicPartitionOffsetByKafkaURL(sourceKafkaURL, newSourceTopicPartition, rewindStartTimestamp);
+          } catch (Exception e) {
+            /**
+             * This is actually tricky. Potentially we could return a -1 value here, but this has the gotcha that if we
+             * have a non-symmetrical failure (like, lor1 can't talk to the ltx1 broker) this will result in a remote
+             * colo rewinding to a potentially non-deterministic offset when the remote becomes available again. So
+             * instead, we record the context of this call and commit it to a repair queue to rewind to a
+             * consistent place on the RT
+             *
+             * NOTE: It is possible that the outage is so long and the rewind so large that we fall off retention. If
+             * this happens we can detect and repair the inconsistency from the offline DCR validator.
+             */
+            unreachableBrokerList.add(sourceKafkaURL);
+            upstreamStartOffset = OffsetRecord.LOWEST_OFFSET;
+            LOGGER.error(
+                "Failed contacting broker {} when processing topic switch! for {}. Setting upstream start offset to {}",
+                sourceKafkaURL,
+                sourceTopicPartition,
+                upstreamStartOffset);
+            hostLevelIngestionStats.recordIngestionFailure();
+
+            // Add to repair queue
+            if (remoteIngestionRepairService != null) {
+              this.remoteIngestionRepairService.registerRepairTask(
+                  this,
+                  buildRepairTask(
+                      sourceKafkaURL.toString(),
+                      sourceTopicPartition,
+                      rewindStartTimestamp,
+                      partitionConsumptionState));
+            } else {
+              // If there isn't an available repair service, then we need to abort in order to make sure the error is
+              // propagated up
+              throw new VeniceException(
+                  String.format(
+                      "Failed contacting broker (%s) and no repair service available!  Aborting topic switch processing for %s. Setting upstream start offset to %d",
+                      sourceKafkaURL,
+                      sourceTopicPartition,
+                      upstreamStartOffset));
+            }
+          }
+        } else {
+          upstreamStartOffset = OffsetRecord.LOWEST_OFFSET;
+        }
+      }
+      upstreamOffsetsByKafkaURLs.put(sourceKafkaURL.toString(), upstreamStartOffset);
+    });
+
+    if (unreachableBrokerList.size() >= ((topicSwitch.sourceKafkaServers.size() + 1) / 2)) {
+      // We couldn't reach a quorum of brokers and that's a red flag, so throw exception and abort!
+      throw new VeniceException("Couldn't reach any broker!!  Aborting topic switch triggered consumer subscription!");
+    }
+    return upstreamOffsetsByKafkaURLs;
+  }
+
+  @Override
   protected void startConsumingAsLeader(PartitionConsumptionState partitionConsumptionState) {
     final int partition = partitionConsumptionState.getPartition();
     final OffsetRecord offsetRecord = partitionConsumptionState.getOffsetRecord();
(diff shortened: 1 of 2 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
```diff
@@ -825,6 +825,97 @@ protected void produceToLocalKafka(
...
+            LOGGER.info(
+                "Get upstreamStartOffset: {} for source URL: {}, topic: {}, rewind timestamp: {}",
+                upstreamStartOffset,
+                sourceKafkaURL,
+                newSourceTopicPartition,
+                rewindStartTimestamp);
+          } catch (Exception e) {
+            /**
+             * This is actually tricky. Potentially we could return a -1 value here, but this has the gotcha that if we
+             * have a non-symmetrical failure (like, lor1 can't talk to the ltx1 broker) this will result in a remote
```
FelixGV, line 868: Let's replace these names with `region1`, `region2` as it doesn't make sense outside of LI...

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
```diff
@@ -907,83 +1023,12 @@ protected void leaderExecuteTopicSwitch(
     final PubSubTopic currentLeaderTopic =
         partitionConsumptionState.getOffsetRecord().getLeaderTopic(pubSubTopicRepository);
     final PubSubTopicPartition sourceTopicPartition = partitionConsumptionState.getSourceTopicPartition(newSourceTopic);
-    Map<String, Long> upstreamOffsetsByKafkaURLs = new HashMap<>(topicSwitch.sourceKafkaServers.size());
-
     List<CharSequence> unreachableBrokerList = new ArrayList<>();
-    topicSwitch.sourceKafkaServers.forEach(sourceKafkaURL -> {
```
FelixGV, line 913: This `leaderExecuteTopicSwitch` function is very long so even in the absence of needing to call this code from somewhere else, it's nice to break it up into a separate function. Thanks!

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestHybrid.java
```diff
@@ -2081,6 +2082,68 @@ public void testHybridWithPartitionWiseConsumer() throws Exception {
     }
   }
 
+  @Test(timeOut = 180 * Time.MS_PER_SECOND)
+  public void testLeaderShouldCalculateRewindDuringPromotion() {
```
FelixGV, line 2086: Thanks for adding an integ test. Just out of curiosity, how reliable have you seen this test be at catching the issue in the absence of the main code fix?

FelixGV (commented):
LGTM overall. Just left a minor comment and a question. Thanks!

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -796,11 +821,26 @@ protected void startConsumingAsLeader(PartitionConsumptionState partitionConsump
...
+    String leaderSourceKafkaURL = leaderSourceKafkaURLs.iterator().next();
+    if (leaderStartOffset < 0 && leaderTopic.isRealTime()) {
+      // TODO: Add description.
+      TopicSwitch topicSwitch = partitionConsumptionState.getTopicSwitch().getTopicSwitch();
+      if (topicSwitch == null) {
+        throw new VeniceException(
+            "New leader does not have topic switch, unable to switch to realtime leader topic: "
+                + leaderTopicPartition);
+      }
+      leaderStartOffset = calculateLeaderUpstreamOffsetWithTopicSwitch(
```
huangminchn, line 835: Shall we update PCS with this new number?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -856,16 +896,13 @@ protected void leaderExecuteTopicSwitch(
...
-        upstreamStartOffset = getTopicPartitionOffsetByKafkaURL(
-            newSourceKafkaServer,
-            newSourceTopicPartition,
-            topicSwitch.rewindStartTimestamp);
-      } else {
-        upstreamStartOffset = OffsetRecord.LOWEST_OFFSET;
-      }
+    if (upstreamStartOffset < 0 && newSourceTopic.isRealTime()) {
+      upstreamStartOffset = calculateLeaderUpstreamOffsetWithTopicSwitch(
+          partitionConsumptionState,
```
huangminchn, line 901: Could you please elaborate? Why do we replace rewind time check with source topic check?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
```diff
@@ -825,6 +825,97 @@ protected void produceToLocalKafka(
...
+          partitionConsumptionState.getLatestProcessedUpstreamRTOffsetWithNoDefault(sourceKafkaURL.toString());
+      if (upstreamStartOffset == null || upstreamStartOffset < 0) {
+        if (rewindStartTimestamp > 0) {
+          PubSubTopicPartition newSourceTopicPartition =
+              new PubSubTopicPartitionImpl(newSourceTopic, sourceTopicPartition.getPartitionNumber());
+          try {
+            upstreamStartOffset =
+                getTopicPartitionOffsetByKafkaURL(sourceKafkaURL, newSourceTopicPartition, rewindStartTimestamp);
+            LOGGER.info(
+                "Get upstreamStartOffset: {} for source URL: {}, topic: {}, rewind timestamp: {}",
```
huangminchn, line 860: Please add consumption task ID as log prefix; otherwise, since source topic might be RT, we wouldn't know which version this log is for.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
```diff
@@ -825,6 +825,97 @@ protected void produceToLocalKafka(
...
+              throw new VeniceException(
+                  String.format(
+                      "Failed contacting broker (%s) and no repair service available!  Aborting topic switch processing for %s. Setting upstream start offset to %d",
+                      sourceKafkaURL,
+                      sourceTopicPartition,
+                      upstreamStartOffset));
+            }
+          }
+        } else {
+          upstreamStartOffset = OffsetRecord.LOWEST_OFFSET;
```
huangminchn, line 906: Is this a valid case? Stream Reprocessing might use rewind time? Since we have a log for the previous branch anyway, could you please also add a log here?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
```diff
@@ -825,6 +825,97 @@ protected void produceToLocalKafka(
...
+        }
+      }
+      upstreamOffsetsByKafkaURLs.put(sourceKafkaURL.toString(), upstreamStartOffset);
```
huangminchn, line 909: For unreachable broker, it's put into the repair service already, we shouldn't populate this map with unreachable broker, right? Was this a bug?

huangminchn (changes requested):
Thanks a lot for the quick turnaournds, Jialin! Left some questions.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
sixpluszero: For now I think it is unexpected, but I can put a warn logging here.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestHybrid.java
sixpluszero: I copy the two tests I added to main, they all breaks as it re-consumes from beginning of the RT topic and take in bad key.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
sixpluszero: I think it is done in the below line 835. 
Also, (just for my understanding) if it is not done here, it will be done after next value being processed as well, is that right?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
sixpluszero: I think either way works. I think for now, if it is unreachable, then we will not consume from it, until it successfully subscribe to the broker, it will re-subscribe and sync the offset again. 
Personally I think it is harmless, but if you think it is good to clean up (I am not sure if there is any gotcha that I am not aware of :) ), I can also do that. What do you think?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
sixpluszero: Ehh, what do you mean by this? 
This method is the logic to calculate rewind offset based on rewind time fetched from TS .

huangminchn: You added this condition: newSourceTopic.isRealTime().
Is it intentional?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
huangminchn: I just realize that we add the unreachable broker into the map, and then later remove it from the map before subscribing:
"""
// We won't attempt to resubscribe for brokers we couldn't compute an upstream offset accurately for. We'll
      // reattempt subscription later
      // Queue up repair here:
      for (CharSequence unreachableBroker: unreachableBrokerList) {
        upstreamOffsetsByKafkaURLs.remove(unreachableBroker.toString());
      }
"""
I would recommend that we try to simply the codes as much as we can; don't add unreachable brokers into the map from the beginning, and remove the above codes in both LFSIT and AASIT.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
huangminchn: Yup, please do.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
huangminchn: I missed that line. PCS is updated.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
sixpluszero: I can do that, btw it is only AASIT.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
sixpluszero: Yeah, we only need to refer to TS to calculate the rewind offset, if we are switching into a RT leader topic as a leader.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
```diff
@@ -957,12 +964,6 @@ protected void startConsumingAsLeader(PartitionConsumptionState partitionConsump
       LOGGER.warn(
           "Failed to reach broker urls {}, will schedule retry to compute upstream offset and resubscribe!",
           unreachableBrokerList.toString());
-      // We won't attempt to resubscribe for brokers we couldn't compute an upstream offset accurately for. We'll
-      // reattempt subscription later
-      // Queue up repair here:
```
huangminchn: Nitpick: could you please modify this comment a bit and move to the above exception branch? So that readers know that by not updating the map, subscription will not happen for the unreachable brokers?

huangminchn (changes requested):

huangminchn (approved):
Thanks a lot Jialin!

PR #927, opened 2024-04-02
[admin-tool] Let admin tool deal with PubSubClientsFactory to support various type of pub sub system.

haoxu07:

Previously, admin tool only supports apache kafka clients, even it have used `PubSubClientsFactory`. In this change, we let admin tool read from system property to generate a specific implementation of `PubSubClientsFactory`, this will enable admin tool to adopt any implementation of  `PubSubClientsFactory`.


sixpluszero (approved):
Thanks for the change, it looks clean

PR #928, opened 2024-04-03
[all] Create producer factory based on the runtime configuration

sushantmane:
##  Create producer factory based on the runtime configuration 

This change removes hardcoded references to ApacheKafkaProducerAdapterFactory and
 introduces dynamic creation of pub-sub client factories based on runtime configurations.  
 

### DIFF
--- a/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
+++ b/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
@@ -131,7 +131,6 @@ import com.linkedin.venice.meta.Store;
 import com.linkedin.venice.meta.Version;
 import com.linkedin.venice.partitioner.DefaultVenicePartitioner;
 import com.linkedin.venice.partitioner.VenicePartitioner;
-import com.linkedin.venice.pubsub.adapter.kafka.producer.ApacheKafkaProducerAdapterFactory;
 import com.linkedin.venice.pushmonitor.ExecutionStatus;
 import com.linkedin.venice.schema.AvroSchemaParseUtils;
 import com.linkedin.venice.schema.writecompute.WriteComputeOperation;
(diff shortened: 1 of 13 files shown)


clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
```diff
@@ -2273,10 +2272,7 @@ void createNewStoreVersion(
   private synchronized VeniceWriter<KafkaKey, byte[], byte[]> getVeniceWriter(PushJobSetting pushJobSetting) {
     if (veniceWriter == null) {
       // Initialize VeniceWriter, TODO: passing configurable PubSubProducerAdapter
```
sixpluszero, line 2274: nit: Is this line still valid after your change?

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/PubSubClientsFactory.java
```diff
@@ -30,4 +57,60 @@ public PubSubConsumerAdapterFactory getConsumerAdapterFactory() {
...
+    }
+
+    return createInstance(className);
+  }
+
+  public static <T> T createInstance(String className) {
+    try {
+      return (T) Class.forName(className).getDeclaredConstructor().newInstance();
+    } catch (Exception e) {
+      LOGGER.error("Failed to create instance of class: {}", className, e);
```
sixpluszero, line 112: Looks like duplicate? you are anyway throwing the same error below.

sixpluszero (dismissed):
overall looks good to me. Left two minor comments, should be easy to resolve

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/PubSubClientsFactory.java
sushantmane: Yeah that's good point. It's redundant if upper layers also log this exception

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
sushantmane: Good catch. Thanks

sushantmane:
Thanks, @sixpluszero for the review!

sixpluszero (approved):

PR #929, opened 2024-04-03
[admin-tool] Allow admin tool to dump kafka topics have no store vers…

lluwm:
…ion in Venice

Today KafkaTopicDumper does not allow dump kafka topics that don't have a corresponding venice store version. For debugging purpose, dumping such kafka topics might still be useful. This change relaxes such condition and report a warning message when it happens.

### DIFF
--- a/clients/venice-admin-tool/src/main/java/com/linkedin/venice/KafkaTopicDumper.java
+++ b/clients/venice-admin-tool/src/main/java/com/linkedin/venice/KafkaTopicDumper.java
@@ -91,7 +91,7 @@ public class KafkaTopicDumper implements AutoCloseable {
   private final String latestValueSchemaStr;
   private final Schema[] allValueSchemas;
   private final Map<Integer, ValueAndDerivedSchemaData> schemaDataMap = new VeniceConcurrentHashMap<>();
-  private final boolean isChunkingEnabled;
+  private boolean isChunkingEnabled;
   private final String parentDirectory;
   private final PubSubConsumerAdapter consumer;
   private final long messageCount;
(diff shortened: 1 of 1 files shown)


clients/venice-admin-tool/src/main/java/com/linkedin/venice/KafkaTopicDumper.java
```diff
@@ -128,14 +128,20 @@ public KafkaTopicDumper(
...
+    this.isChunkingEnabled = false;
     if (Version.isATopicThatIsVersioned(topic)) {
       int version = Version.parseVersionFromKafkaTopicName(topic);
-      if (!storeInfo.getVersion(version).isPresent()) {
-        throw new VeniceException("Version: " + version + " does not exist for store: " + storeName);
+      if (storeInfo.getVersion(version).isPresent()) {
+        this.isChunkingEnabled = storeInfo.getVersion(version).get().isChunkingEnabled();
+      } else {
+        // Use the default chunking setting from store if the version does not exist.
+        this.isChunkingEnabled = storeInfo.isChunkingEnabled();
```
gaojieliu, line 138: This could be wrong as the chunking flag in store level can be different from the version level.
I wonder whether we can dynamically infer whether the chunking is enabled or not as when chunking is enabled, when chunking is enabled, regular key will have the `NON_CHUNK_KEY_SUFFIX`, so I guess we can try to compare the tailing part of key is same as the serialized bytes of `NON_CHUNK_KEY_SUFFIX` or not.

Another more complex way is to write the chunking flag into SOP message, which would need to evolve KME schema, and we don't have to do that for this debugging purpose.

sixpluszero: Or, I am wondering can we make every store/version chunking + rmd chunking enabled? For <1MB data, that's not going to change anything, is that right?

gaojieliu: chunking has some overhead as it will append some chunking suffix into the key, and in theory, we don't want to enable chunking by default as the large value is not expected in the regular stores as it would slow down the read path significantly.

PR #930, opened 2024-04-04
[fast-client] Long tail retry budget for fast client

xunyin8:
## [fast-client] Long tail retry budget for fast client
1. Introduced RetryManager to measure and calculate retry budget based on a percentage of requests.

2. The RetryManager is only applied to long tail retries and currently hard coded to be 10% of the total original request rate. This means if there were any exception that's not a 429 the retry will be triggered without going through the RetryManager. If the retry budget is exhausted then the retry task will do nothing and the request will either complete eventually (original future) or time out.

## How was this PR tested?
Existing and new tests

### DIFF
--- a/clients/venice-client/src/main/java/com/linkedin/venice/fastclient/ClientConfig.java
+++ b/clients/venice-client/src/main/java/com/linkedin/venice/fastclient/ClientConfig.java
@@ -84,6 +84,12 @@ public class ClientConfig<K, V, T extends SpecificRecord> {
    * route requests to the correct server/partition
    */
   private final GrpcClientConfig grpcClientConfig;
+  /**
+   * The time window used to calculate user traffic and corresponding long tail retry budget. The default value is 5
+   * minutes. Meaning it will use the average request occurrence rate over the 5 minutes to calculate the corresponding
+   * retry budget for the next 5 minutes and so on.
+   */
+  private final long longTailRetryBudgetEnforcementWindowInMs;
 
   private boolean projectionFieldValidation;
 
@@ -121,7 +127,8 @@ public class ClientConfig<K, V, T extends SpecificRecord> {
       String clusterDiscoveryD2Service,
       boolean useGrpc,
       GrpcClientConfig grpcClientConfig,
-      boolean projectionFieldValidation) {
+      boolean projectionFieldValidation,
+      long longTailRetryBudgetEnforcementWindowInMs) {
     if (storeName == null || storeName.isEmpty()) {
       throw new VeniceClientException("storeName param shouldn't be empty");
     }
@@ -252,6 +259,7 @@ public class ClientConfig<K, V, T extends SpecificRecord> {
     this.grpcClientConfig = grpcClientConfig;
 
     this.projectionFieldValidation = projectionFieldValidation;
+    this.longTailRetryBudgetEnforcementWindowInMs = longTailRetryBudgetEnforcementWindowInMs;
   }
 
   public String getStoreName() {
@@ -391,6 +399,10 @@ public class ClientConfig<K, V, T extends SpecificRecord> {
     return projectionFieldValidation;
   }
 
+  public long getLongTailRetryBudgetEnforcementWindowInMs() {
+    return longTailRetryBudgetEnforcementWindowInMs;
+  }
+
   public ClientConfig setProjectionFieldValidationEnabled(boolean projectionFieldValidation) {
     this.projectionFieldValidation = projectionFieldValidation;
     return this;
@@ -439,6 +451,8 @@ public class ClientConfig<K, V, T extends SpecificRecord> {
     private GrpcClientConfig grpcClientConfig = null;
 
     private boolean projectionFieldValidation = true;
+
+    private long longTailRetryBudgetEnforcementWindowInMs = 300000; // 5 minutes
 
     public ClientConfigBuilder<K, V, T> setStoreName(String storeName) {
       this.storeName = storeName;
@@ -622,6 +636,12 @@ public class ClientConfig<K, V, T extends SpecificRecord> {
       return this;
     }
 
+    public ClientConfigBuilder<K, V, T> setLongTailRetryBudgetEnforcementWindowInMs(
+        long longTailRetryBudgetEnforcementWindowInMs) {
+      this.longTailRetryBudgetEnforcementWindowInMs = longTailRetryThresholdForBatchGetInMicroSeconds;
+      return this;
+    }
+
     public ClientConfigBuilder<K, V, T> clone() {
       return new ClientConfigBuilder().setStoreName(storeName)
           .setR2Client(r2Client)
@@ -656,7 +676,8 @@ public class ClientConfig<K, V, T extends SpecificRecord> {
           .setClusterDiscoveryD2Service(clusterDiscoveryD2Service)
           .setUseGrpc(useGrpc)
           .setGrpcClientConfig(grpcClientConfig)
-          .setProjectionFieldValidationEnabled(projectionFieldValidation);
+          .setProjectionFieldValidationEnabled(projectionFieldValidation)
+          .setLongTailRetryBudgetEnforcementWindowInMs(longTailRetryBudgetEnforcementWindowInMs);
     }
 
     public ClientConfig<K, V, T> build() {
@@ -694,7 +715,8 @@ public class ClientConfig<K, V, T extends SpecificRecord> {
           clusterDiscoveryD2Service,
           useGrpc,
           grpcClientConfig,
-          projectionFieldValidation);
+          projectionFieldValidation,
+          longTailRetryBudgetEnforcementWindowInMs);
     }
   }
 }
--- a/clients/venice-client/src/main/java/com/linkedin/venice/fastclient/ClientConfig.java
+++ b/clients/venice-client/src/main/java/com/linkedin/venice/fastclient/ClientConfig.java
@@ -452,7 +452,7 @@ public class ClientConfig<K, V, T extends SpecificRecord> {
 
     private boolean projectionFieldValidation = true;
 
-    private long longTailRetryBudgetEnforcementWindowInMs = 300000; // 5 minutes
+    private long longTailRetryBudgetEnforcementWindowInMs = 60000; // 1 minute
 
     public ClientConfigBuilder<K, V, T> setStoreName(String storeName) {
       this.storeName = storeName;
--- a/clients/venice-client/src/main/java/com/linkedin/venice/fastclient/ClientConfig.java
+++ b/clients/venice-client/src/main/java/com/linkedin/venice/fastclient/ClientConfig.java
@@ -85,9 +85,9 @@ public class ClientConfig<K, V, T extends SpecificRecord> {
    */
   private final GrpcClientConfig grpcClientConfig;
   /**
-   * The time window used to calculate user traffic and corresponding long tail retry budget. The default value is 5
-   * minutes. Meaning it will use the average request occurrence rate over the 5 minutes to calculate the corresponding
-   * retry budget for the next 5 minutes and so on.
+   * The time window used to calculate user traffic and corresponding long tail retry budget. The default value is one
+   * minute. Meaning it will use the average request occurrence rate over the one minute to calculate the corresponding
+   * retry budget for the next minute and so on.
    */
   private final long longTailRetryBudgetEnforcementWindowInMs;
 
--- a/clients/venice-client/src/main/java/com/linkedin/venice/fastclient/RetriableAvroGenericStoreClient.java
+++ b/clients/venice-client/src/main/java/com/linkedin/venice/fastclient/RetriableAvroGenericStoreClient.java
@@ -9,6 +9,7 @@ import com.linkedin.venice.client.store.streaming.StreamingCallback;
 import com.linkedin.venice.compute.ComputeRequestWrapper;
 import com.linkedin.venice.exceptions.VeniceException;
 import com.linkedin.venice.fastclient.meta.RetryManager;
+import com.linkedin.venice.fastclient.stats.RetryManagerStats;
 import com.linkedin.venice.utils.concurrent.VeniceConcurrentHashMap;
 import java.util.Collections;
 import java.util.Optional;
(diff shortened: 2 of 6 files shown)


xunyin8:
Will add the following metrics to track retry budget behavior:

- The long tail retry budget value that's currently being enforced
- Token bucket stats for the long tail retry budget to monitor usage

xunyin8:
Will also add more unit tests in the `RetriableAvroGenericStoreClient` layer and have at least one integration test with a much lower enforcement window for a smoke test (since long tail retry trigger is not deterministic in integration test).

xunyin8:
> Will add the following metrics to track retry budget behavior:
> 
> * The long tail retry budget value that's currently being enforced
> * Token bucket stats for the long tail retry budget to monitor usage

Done

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/meta/RetryManager.java
```diff
@@ -0,0 +1,121 @@
...
+
+
+/**
+ * This class offers advanced client retry behaviors. Specifically enforcing a retry budget and relevant monitoring to
+ * avoid retry storm and alert users when the retry threshold is misconfigured or service is degrading.
+ */
+public class RetryManager implements Closeable {
+  private static final Logger LOGGER = LogManager.getLogger(RetryManager.class);
+  private static final int TOKEN_BUCKET_REFILL_INTERVAL_IN_SECONDS = 1;
+  private static final int TOKEN_BUCKET_CAPACITY_MULTIPLE = 5;
```
gaojieliu, line 23: Can you leave some javadoc about this field to explain the reasoning?

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/meta/RetryManager.java
```diff
@@ -0,0 +1,121 @@
...
+    }
+  }
+
+  public boolean isRetryAllowed() {
+    if (!retryBudgetEnabled.get() || retryTokenBucket.get() == null) {
+      // All retries are allowed when the feature is disabled or during the very first enforcement window when we
+      // haven't collected enough data points yet
+      return true;
+    }
+    return retryTokenBucket.get().tryConsume(1);
```
gaojieliu, line 66: As `retryTokenBucket` can change, so shall we keep a local reference to avoid the potentially race conditions?
```
TokenBucket tokenBucket = retryTokenBucket.get();
if (!retryBudgetEnabled.get() || tokenBucket == null) {
      // All retries are allowed when the feature is disabled or during the very first enforcement window when we
      // haven't collected enough data points yet
      return true;
}
return tokenBucket.tryConsume(1);
```

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/meta/RetryManager.java
```diff
@@ -0,0 +1,121 @@
...
+        long requestCountSinceLastUpdate = requestCount.getAndSet(0);
+        lastUpdateTimestamp = clock.millis();
+        // Minimum user request per second will be 1
+        long newQPS = (long) Math
+            .ceil((double) requestCountSinceLastUpdate / (double) TimeUnit.MILLISECONDS.toSeconds(elapsedTimeInMs));
+        if (previousQPS > 0) {
+          long difference = Math.abs(previousQPS - newQPS);
+          double differenceInPercentDecimal = (double) difference / (double) previousQPS;
+          if (differenceInPercentDecimal > 0.1) {
+            // Only update the retry token bucket if the change in request per seconds is more than 10 percent
```
gaojieliu, line 82: Why can't we update the token bucket all the time?
Is it an expensive operation?

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/stats/RetryManagerStats.java
```diff
@@ -0,0 +1,34 @@
...
+      }
+    }, "retry_limit_per_seconds"));
+    retriesRemaining = registerSensor(new AsyncGauge((ignored, ignored2) -> {
+      TokenBucket bucket = retryManager.getRetryTokenBucket();
+      if (bucket == null) {
+        return -1;
+      } else {
+        return bucket.getStaleTokenCount();
+      }
+    }, "retries_remaining"));
```
gaojieliu, line 32: Can we also have a metric to track denied retry requests?
I hope we can track retry stats per request type.

gaojieliu (commented):
Looks good overall, and left a few clarification comments.

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/RetriableAvroGenericStoreClient.java
```diff
@@ -315,6 +345,7 @@ private <R extends MultiKeyRequestContext<K, V>, RESPONSE> void retryStreamingMu
             savedException,
             pendingKeysFuture,
             scheduledRetryTask));
+    longTailRetryManager.recordRequest();
```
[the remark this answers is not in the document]
xunyin8: Currently retry budget is at the request level, perhaps we should maintain it at the key level and use KPS then?

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/meta/RetryManager.java
xunyin8: It's not an expensive operation. However, the default initial value for token bucket is `capacity` which is maxed out. If we keep initializing new token bucket then we are technically allowing more retry even if the traffic remains the same.

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/meta/RetryManager.java
xunyin8: Keeping a local reference just changes the behavior but doesn't really avoid any potential race conditions IMO. We are not taking a lock and blocking when consuming and and updating `retryTokenBucket`. We could get the local reference before or after `retryTokenBucket` is updated.

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/RetriableAvroGenericStoreClient.java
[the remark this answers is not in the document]
gaojieliu: I think request level is fine, and KPS level will be problematic if not considering request type.
For example, avg key count for multi-get request is 1000 and if the key based retry budget is being utlized by single-get request, it will allow a very high number of single get requests, which will be very bad.

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/meta/RetryManager.java
gaojieliu: yeah, and this code I mentioned is mostly to avoid the case that first call of `retryTokenBucket.get()` is not `null`, but the second call of `retryTokenBucket.get()` is `null`.

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/meta/RetryManager.java
gaojieliu: Emm.. 
So the initialization would assume the max cap, and I thought it should be initialized by the 0.1 * previous QPS, no? Maybe not for the first time, but for the subsequent initialization, it doesn't necessary to initialize with max cap, right?

xunyin8: I meant the max cap as in the 0.1 * previous QPS * `TOKEN_BUCKET_CAPACITY_MULTIPLE` (5).

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/ClientConfig.java
```diff
@@ -440,6 +452,8 @@ public static class ClientConfigBuilder<K, V, T extends SpecificRecord> {
 
     private boolean projectionFieldValidation = true;
 
+    private long longTailRetryBudgetEnforcementWindowInMs = 60000; // 1 minute
```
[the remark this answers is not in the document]
xunyin8: 1 minute, I think 5 minutes is too long. The traffic pattern can change multiple times within 5 minutes and not able to adjust the retry budget in a timely manner may cause issues.

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/RetriableAvroGenericStoreClient.java
```diff
@@ -230,6 +252,12 @@ public void compute(
         });
   }
 
+  @Override
+  public void close() {
+    super.close();
+    longTailRetryManager.close();
```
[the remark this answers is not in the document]
xunyin8: Good catch, let me fix that.

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/RetriableAvroGenericStoreClient.java
```diff
@@ -269,24 +297,26 @@ private <R extends MultiKeyRequestContext<K, V>, RESPONSE> void retryStreamingMu
...
-        streamingRequestExecutor.trigger(
-            retryRequestContext,
-            pendingKeys,
-            getStreamingCallback(
-                retryRequestContext,
-                finalRequestCompletionFuture,
-                savedException,
-                pendingKeysFuture,
-                null));
+        if (throwable != null || longTailRetryManager.isRetryAllowed()) {
```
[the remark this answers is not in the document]
xunyin8: The long tail scenario. If the original request is taking too long then we won't see any errors yet but we want to trigger the long tail retry because it might come back fast (before the original request). 

We always send out the original/first request. If there is no retry budget and the original request never comes back then the whole request will timeout and fail (the final future).

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/ClientConfig.java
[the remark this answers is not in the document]
gaojieliu: yes, but the javadoc mentioned something about 5mins.

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/meta/RetryManager.java
gaojieliu: What is the reason to multiple `TOKEN_BUCKET_CAPACITY_MULTIPLE`? To tolerate some random QPS spike?

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/ClientConfig.java
```diff
@@ -84,6 +84,12 @@ public class ClientConfig<K, V, T extends SpecificRecord> {
    * route requests to the correct server/partition
    */
   private final GrpcClientConfig grpcClientConfig;
+  /**
+   * The time window used to calculate user traffic and corresponding long tail retry budget. The default value is 5
```
gaojieliu, line 88: The javadoc mentions 5mins and the following config is 1min:
```
    private long longTailRetryBudgetEnforcementWindowInMs = 60000; // 1 minute
```
Are these two related?

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/RetriableAvroGenericStoreClient.java
```diff
@@ -37,7 +38,18 @@ public class RetriableAvroGenericStoreClient<K, V> extends DelegatingAvroStoreCl
   private final int longTailRetryThresholdForBatchGetInMicroSeconds;
   private final int longTailRetryThresholdForComputeInMicroSeconds;
   private final TimeoutProcessor timeoutProcessor;
+  /**
+   * The long tail retry budget is only applied to long tail retries. If there were any exception that's not a 429 the
+   * retry will be triggered without going through the long tail {@link RetryManager}. If the retry budget is exhausted
+   * then the retry task will do nothing and the request will either complete eventually (original future) or time out.
+   */
+  private RetryManager singleGetLongTailRetryManager = null;
+  private RetryManager multiGetLongTailRetryManager = null;
```
gaojieliu, line 47: Do we have a retry manager for compute request type?

gaojieliu: Or we are planning to use the multiget retry manager for compute request type?

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/meta/RetryManager.java
gaojieliu: Does this mean fast-client will allow 50% of retry all the time or just some short period and the bursty retry will still draw from the total budget, right?

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/meta/RetryManager.java
```diff
@@ -0,0 +1,142 @@
...
+   */
+  private static final int TOKEN_BUCKET_CAPACITY_MULTIPLE = 5;
+  private final AtomicBoolean retryBudgetEnabled = new AtomicBoolean();
+  private final long enforcementWindowInMs;
+  private final double retryBudgetInPercentDecimal;
+  private final AtomicLong requestCount = new AtomicLong();
+  private final AtomicReference<TokenBucket> retryTokenBucket = new AtomicReference<>(null);
+  private final ScheduledExecutorService scheduler;
+  private final Clock clock;
+  RetryManagerStats retryManagerStats;
```
gaojieliu, line 37: Can this variable be a `private` member?

gaojieliu (commented):
Left a few minor comments.

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/RetriableAvroGenericStoreClient.java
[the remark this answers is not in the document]
mynameborat: thanks for the explanation.

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/ClientConfig.java
xunyin8: Yes, good catch. FIxed.

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/RetriableAvroGenericStoreClient.java
gaojieliu: Yeah, you are correct.

gaojieliu (approved):
Thanks for the change!

PR #931, opened 2024-04-04
[controller] Start unused value schema cleanup service

majisourav99:
## Start unused value schema cleanup service

The thread to cleanup unused value schema was not started. This PR starts unused value schema cleanup service in parent controller only, 

## How was this PR tested?

CI


gaojieliu (commented):
Code change looks good.
Do we have an integration test to validate this feature?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
```diff
@@ -634,6 +632,10 @@ private void sendAdminMessageAndWaitForConsumed(String clusterName, String store
   @Override
   public void deleteValueSchemas(String clusterName, String storeName, Set<Integer> unusedValueSchemaIds) {
     Set<Integer> inuseValueSchemaIds = getInUseValueSchemaIds(clusterName, storeName);
+    if (inuseValueSchemaIds.isEmpty()) {
+      LOGGER.error("For store {} cannot find in-use schema ids.", storeName);
```
gaojieliu, line 636: Should this log be always in error level?
Such as if there is no active store versions, it is expected to have empty in-used schemas, right?
Shall we only log this in error level only the store has some active versions?

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/consumer/AdminExecutionTask.java
```diff
@@ -490,7 +490,9 @@ private void handleSetStore(UpdateStore message) {
         .setReplicationFactor(message.replicationFactor)
         .setMigrationDuplicateStore(message.migrationDuplicateStore)
         .setLatestSupersetSchemaId(message.latestSuperSetValueSchemaId)
-        .setBlobTransferEnabled(message.blobTransferEnabled);
+        .setBlobTransferEnabled(message.blobTransferEnabled)
+        .setUnusedSchemaDeletionEnabled(message.unusedSchemaDeletionEnabled)
+        .setLatestSupersetSchemaId(message.latestSuperSetValueSchemaId);
```
gaojieliu, line 495: Emmm..
Is there any prod impact caused by the missing statement?

gaojieliu (commented):
Looks good overall, left a couple of minor comments

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/consumer/AdminExecutionTask.java
majisourav99: Not yet. `setLatestSupersetSchemaId` was duplicate statement. I will remove it.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
majisourav99: Sure, we can add a check for active versions

gaojieliu (approved):
Thanks!

PR #932, opened 2024-04-05
[controller] Avoid expensive call to TM::getAllTopicRetentions in TopicCleanupService

sushantmane:
## Avoid expensive call to TM::getAllTopicRetentions in TopicCleanupService

The call to TopicManager::getAllTopicRetentions is expensive and unnecessary
for non-local region topic queries. We do not use the retention values from
remote version topics to decide whether to delete local real-time (RT) topics.
Replacing it with TM::listTopics, which is sufficient and more cost-effective.

## How was this PR tested?

CI


xunyin8 (approved):
Thanks Sushant!

sushantmane:
Thanks for the review, @xunyin8

PR #933, opened 2024-04-05
[controller] Do not let start push when future version exists

majisourav99:
## Do not let start push when future version exists

With deferred version swap enabled, if user does not roll forward to the future version, we should not let another new push start, otherwise it will accumulate many future versions and resources associated. This PR prevents such pushes.

## How was this PR tested?

integ test


services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
```diff
@@ -1227,6 +1229,22 @@ Optional<String> getTopicForCurrentPushJob(
           }
           return Optional.empty();
         } else {
+          Map<String, ControllerClient> controllerClients = getVeniceHelixAdmin().getControllerClientMap(clusterName);
+          for (Map.Entry<String, ControllerClient> entry: controllerClients.entrySet()) {
+            ControllerClient controllerClient = entry.getValue();
+            StoreResponse response = controllerClient.getStore(storeName);
```
huangminchn, line 1235: Just in case, shall we use the retry API here? ControllerClient#retryableRequest with 3 attempts?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
```diff
@@ -1227,6 +1229,22 @@ Optional<String> getTopicForCurrentPushJob(
...
+            if (!response.isError()) {
+              StoreInfo store = response.getStore();
+              Optional<Version> version = store.getVersion(versionNumber);
+              // if there is a online future version available, do not allow further pushes
+              if (version.isPresent() && version.get().getStatus().equals(ONLINE)
+                  && version.get().isVersionSwapDeferred() && store.getCurrentVersion() < versionNumber) {
+                throw new VeniceException(
```
huangminchn, line 1242: BTW, the previous branch return the topic name, shall we keep a consistent behavior here?

huangminchn (changes requested):
Thanks a lot Sourav!

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
```diff
@@ -1145,13 +1146,36 @@ Optional<String> getTopicForCurrentPushJob(
       latestTopic = Optional.of(versionTopics.get(0));
     }
 
+    Map<String, String> futureVersions = getFutureVersionsForMultiColos(clusterName, storeName);
+    for (Map.Entry<String, String> entry: futureVersions.entrySet()) {
+      ControllerClient controllerClient = getVeniceHelixAdmin().getControllerClientMap(clusterName).get(entry.getKey());
+      StoreResponse storeResponse = controllerClient.retryableRequest(5, c -> c.getStore(storeName));
```
huangminchn, line 1152: This was the two queries scenario I was trying to avoid; if "futureVersions" returned from call "getFutureVersionsForMultiColos" is not empty, why can't we just abort immediately?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
```diff
@@ -1778,7 +1802,8 @@ public Map<String, String> getFutureVersionsForMultiColos(String clusterName, St
     for (Map.Entry<String, ControllerClient> entry: controllerClients.entrySet()) {
       String region = entry.getKey();
       ControllerClient controllerClient = entry.getValue();
-      MultiStoreStatusResponse response = controllerClient.getFutureVersions(clusterName, storeName);
+      MultiStoreStatusResponse response =
+          controllerClient.retryableRequest(10, c -> c.getFutureVersions(clusterName, storeName));
       if (response.isError()) {
```
mynameborat, line 1807: Is this change relevant to the PR or is it adding more resiliency?

I wonder if this changes the behavior of the system in the case of transient issues where previously the function would return result with `IGNORED_CURRENT_VERSION` while the newer behavior upon retries can return actual results.

Do we need to be concerned with the behavior changes?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
```diff
@@ -3578,16 +3604,18 @@ private void truncateTopicsOptionally(
        * 3. the store is incremental push enabled and same incPushToRT and batch push finished
        */
       Store store = getVeniceHelixAdmin().getStore(clusterName, Version.parseStoreFromKafkaTopicName(kafkaTopic));
+      Optional<Version> version = store.getVersion(Version.parseVersionFromKafkaTopicName(kafkaTopic));
+
       boolean failedBatchPush = !incrementalPushVersion.isPresent() && currentReturnStatus == ExecutionStatus.ERROR;
       boolean incPushEnabledBatchPushSuccess = !incrementalPushVersion.isPresent() && store.isIncrementalPushEnabled();
       boolean nonIncPushBatchSuccess =
           !store.isIncrementalPushEnabled() && currentReturnStatus != ExecutionStatus.ERROR;
+      boolean isDeferredVersionSwap = version.isPresent() && version.get().isVersionSwapDeferred();
```
mynameborat, line 3613: might be simpler to rewrite it as

```
version
  .map(Version::isVersionSwapDeferred)
  .orElse(false);

```

services/venice-controller/src/test/java/com/linkedin/venice/controller/TestVeniceParentHelixAdmin.java
```diff
@@ -1465,6 +1454,26 @@ public void testStoreVersionCleanUpWithMoreVersions() {
...
   }
 
+  private void mockControllerClients() {
+    Map<String, ControllerClient> controllerClientMap = new HashMap<>();
+
+    for (int i = 0; i < 3; i++) {
+      ControllerClient client = mock(ControllerClient.class);
+      StoreResponse storeResponse = new StoreResponse();
+      Store s = TestUtils.createTestStore("s" + i, "test", System.currentTimeMillis());
+      s.setCurrentVersion(i + 4); // child region current versions 4,5,6
```
mynameborat, line 1464: I am assuming you want the current version of `i` to be offset by the number of total stores (or total number of clients) in the test.
Would be nice to make that explicit to prevent potential overlap. e.g., changes the `for loop` condition to y but forgot to make changes to the ln:1464

e.g., instead of 3, can we extract it to some meaningful constant and set the current version in ln:1464 to 
`s.setCurrentVersion(i + <CONSTANT>)`.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
mynameborat, line 1152: Is there any rationale behind using `5` as the retry count here as opposed to `10` below for the `getFutureVersions`?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
```diff
@@ -1145,13 +1146,36 @@ Optional<String> getTopicForCurrentPushJob(
       latestTopic = Optional.of(versionTopics.get(0));
     }
 
+    Map<String, String> futureVersions = getFutureVersionsForMultiColos(clusterName, storeName);
+    for (Map.Entry<String, String> entry: futureVersions.entrySet()) {
+      ControllerClient controllerClient = getVeniceHelixAdmin().getControllerClientMap(clusterName).get(entry.getKey());
+      StoreResponse storeResponse = controllerClient.retryableRequest(5, c -> c.getStore(storeName));
+      StoreInfo store = storeResponse.getStore();
+      int futureVersion = Integer.parseInt(entry.getValue());
```
mynameborat, line 1154: Can `Integer.parseInt` throw exception for invalid `getValue()`?
Or is it guaranteed to be a valid input for the function call.

mynameborat (commented):
Feel free to correct me in places where my assumptions/questions don't make sense.

services/venice-controller/src/test/java/com/linkedin/venice/controller/TestVeniceParentHelixAdmin.java
majisourav99: No, its the off by 1 mistake, its just setting the current version different in different region. I could make the regions number a constant

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
```diff
@@ -1972,6 +1986,15 @@ public void rollForwardToFutureVersion(String clusterName, String storeName, Str
...
       sendAdminMessageAndWaitForConsumed(clusterName, storeName, message);
+      Map<String, String> futureVersions = getFutureVersionsForMultiColos(clusterName, storeName);
+      int futureVersion = 0;
+      for (Map.Entry<String, String> entry: futureVersions.entrySet()) {
+        futureVersion = Integer.parseInt(entry.getValue());
+        if (futureVersion > 0) {
+          break;
+        }
+      }
+      truncateKafkaTopic(Version.composeKafkaTopic(storeName, futureVersion));
```
huangminchn, line 1997: If child controllers manage to promote the version quickly, getFutureVersion API might return nothing?

If the above race condition could happen, can we get the list of future versions first, send admin message next, and eventually truncation the parent topics?
Besides, why there could be more than 1 future version number? In theory, there is only one topic?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
```diff
@@ -3488,13 +3511,19 @@ private OfflinePushStatusInfo getOffLineJobStatus(
...
-      truncateTopicsOptionally(
-          clusterName,
-          kafkaTopic,
-          incrementalPushVersion,
-          currentReturnStatus,
-          currentReturnStatusDetails);
+      String storeName = Version.parseStoreFromKafkaTopicName(kafkaTopic);
+      int versionNum = Version.parseVersionFromKafkaTopicName(kafkaTopic);
+      boolean deferredSwap = isDeferredSwap(clusterName, storeName, versionNum);
+      if (!deferredSwap) {
```
huangminchn, line 3519: Is RollForward the only option after the push completes with the deferred version swap flag on?? Is there a cancel button? If so, we need to also truncate parent VT when they hit cancel.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
majisourav99: Good point, I will swap the order of the operations. Ideally there should not be more than 1 future version, but for  this API is for displaying future version per colo, which can have no future version for certain colo.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
majisourav99: Right, roll forward only works if deferred swap is on. No cancel button.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
majisourav99: We would rather get actual results, this change is just adding resiliency.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
majisourav99: I assumed, getStore is called more often hence higher chance of it having cached.

huangminchn: Discussed offline:
During target colo push, there is a phase where the target colo finishes the first push and no future version, while the rest of the colos haven't recieved the new push yet; at this moment, there is no future version in any child colos, and thus it's blocker for removing parent VT until we have a new state in parent controller to indicate that a push (either a regular push or single-colo push is ongoing.)

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
majisourav99: Its set to -1 for error case, so we will always parse the string to int.

huangminchn (approved):
Thanks a lot Sourav!

PR #934, opened 2024-04-05
[dvc] Add a new config to replace pid as instance name suffix

sixpluszero:
## [dvc] Add a new config to replace pid as instance name suffix
Using process pid as instance name suffix will not work for VPJ when application slice is performing large scale restart.
This PR introduces a new config as instance suffix to replace PID. And if the config is not specified, we will still use PID as fall back (but not recommended)
## How was this PR tested?

Modified existing integration test to verify the new config is adopted.


gaojieliu (commented):
Code change looks good, but the test is a little confusing as multiple applications can carry the same instance tag.
Maybe we should use something like: `app1_i001` to make it clear.

gaojieliu (approved):

PR #935, opened 2024-04-05
[controller] Adding blob transfer feature flag to store schema

ishwarya-citro:
## Adding Blob Transfer feature flag
Added a feature flag to turn on/off the blob transfer at store level
If the feature is on, it will allow snapshots of the blobs ([see PR](https://github.com/linkedin/venice/pull/914)) and enable download/upload blobs

## How was this PR tested?
CI Tests


internal/venice-common/src/main/java/com/linkedin/venice/controllerapi/UpdateStoreQueryParams.java
```diff
@@ -79,6 +79,8 @@
 import java.util.Optional;
 import java.util.stream.Collectors;
 
+import static com.linkedin.venice.controllerapi.ControllerApiConstants.*;
```
adamxchen, line 82: nit: not import all

internal/venice-common/src/main/java/com/linkedin/venice/meta/ReadOnlyStore.java
```diff
@@ -1333,6 +1333,14 @@ public boolean isUnusedSchemaDeletionEnabled() {
     return this.delegate.isUnusedSchemaDeletionEnabled();
   }
 
+  @Override
+  public void setBlobTransferEnabled(boolean blobTransferEnabled) {
+    throw new UnsupportedOperationException();
```
adamxchen, line 1338: nit: can we add some msgs?

internal/venice-common/src/main/java/com/linkedin/venice/meta/SystemStore.java
```diff
@@ -634,6 +634,14 @@ public long getMaxCompactionLagSeconds() {
     return zkSharedStore.getMaxCompactionLagSeconds();
   }
 
+  @Override
+  public void setBlobTransferEnabled(boolean blobTransferEnabled) {
+    throwUnsupportedOperationException("setBlobTransferEnabled");
```
adamxchen, line 639: nit: we can use have msg here

adamxchen (approved):
LGTM!

PR #939, opened 2024-04-11
[dvc][server]Enable leader complete status check for non Active-Active stores

m-nagarajan:
## Summary
PR [775](https://github.com/linkedin/venice/pull/775) introduced a feature for hybrid stores' followers to go ready to serve only after leader is marked completed. But it was disabled for non-AA stores as the Heartbeats (which this feature piggybacks on) are written only to local RT's but certain non-AA cases' LFSIT consumes only from parent RT.  One fix considered for these non-AA stores to get onboard to this feature was to write HB to parent RT as well. But as Parent RT is soon to be deprecated and we don't want to add more to it, this PR works around that considering the below aspects and enabling this feature for all stores except non-AA and `DRP=AGGREGATE` stores.
1. All incremental push stores should be Active-active by default.
2. All stores not on AA without incremental push except `DRP=AGGREGATE` will be reading from local RT. 

the long-term goal is to remove `DRP` completely and make all existing hybrid stores either:
1. AA: If cross region replication needed
2. non-AA: If cross region replication not needed

Also removed `LEADER_COMPLETE_STATE_UNKNOWN` and `firstHeartBeatSOSReceived` as it was introduced for deployment transition phase to start using this header. Config `SERVER_LEADER_COMPLETE_STATE_CHECK_IN_FOLLOWER_ENABLED` should be used to enable this feature after deployment.

## How was this PR tested?
GH CI

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
@@ -10,7 +10,6 @@ import static com.linkedin.davinci.kafka.consumer.LeaderFollowerStateType.STANDB
 import static com.linkedin.venice.kafka.protocol.enums.ControlMessageType.END_OF_PUSH;
 import static com.linkedin.venice.kafka.protocol.enums.ControlMessageType.START_OF_SEGMENT;
 import static com.linkedin.venice.pubsub.api.PubSubMessageHeaders.VENICE_LEADER_COMPLETION_STATE_HEADER;
-import static com.linkedin.venice.writer.LeaderCompleteState.LEADER_COMPLETE_STATE_UNKNOWN;
 import static com.linkedin.venice.writer.VeniceWriter.APP_DEFAULT_LOGICAL_TS;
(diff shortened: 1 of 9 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -1960,64 +1960,72 @@ protected boolean isHybridFollower(PartitionConsumptionState partitionConsumptio
...
+    return (getServerConfig().isLeaderCompleteStateCheckInFollowerEnabled() && this.hybridStoreConfig.isPresent()
+        && this.hybridStoreConfig.get().getDataReplicationPolicy() != DataReplicationPolicy.AGGREGATE);
+  }
+
   /**
    * Checks whether the lag is acceptable for hybrid stores
    * <p>
-   * leaderCompleteStateCheckEnabled is not used for non AA stores: as SIT reads from parent RT for non AA
-   * stores but HB is written to local RT. This should be enabled once all hybrid stores are made AA.
+   * If the instance is a hybrid standby or DaVinciClient: Also check if <br>
```
sushantmane, line 1976: Can we also add comment that leader completed check is done only if lag has caught up against VT?

sushantmane (approved):
LGTM! 

Thanks, @m-nagarajan!

PR #940, opened 2024-04-11
[admin-tool][server] Skip ACL checking for topic partition ingestion context query from admin tool for server host

haoxu07:
Topic partition ingestion context query is for debugging purpose, it will not access user' store's content, so it should not need the ACL check. 

Resolves #XXX

## How was this PR tested?

CI


huangminchn (approved):
Thanks Hao!

PR #943, opened 2024-04-15
[dvc] Split large batch-get to execute concurrently

gaojieliu:
Today, for large batch-get request, the latency will be high even the customer is using `MEMORY_BACKED_BY_DISK` mode. For example, in avg, the lookup latency is about 20us and for a request with 3k keys, the latency will be roughly 60ms, which is high and we often recommend customers to break it down into smaller chunks and execute these chunks in parallel. Now, this PR will add such support for batch-get as a default behavior.

TODO: add such support for compute request.

## How was this PR tested?
CI

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/AvroGenericDaVinciClient.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/AvroGenericDaVinciClient.java
@@ -178,11 +178,6 @@ public class AvroGenericDaVinciClient<K, V> implements DaVinciClient<K, V>, Avro
     preValidation.run();
   }
 
-  // For testing
-  void setDaVinciBackend(DaVinciBackend testDaVinciBackend) {
-    daVinciBackend = new ReferenceCounted<>(testDaVinciBackend, (ignored) -> {});
-  }
-
   @Override
   public String getStoreName() {
     return clientConfig.getStoreName();
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/AvroGenericDaVinciClient.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/AvroGenericDaVinciClient.java
@@ -114,7 +114,7 @@ public class AvroGenericDaVinciClient<K, V> implements DaVinciClient<K, V>, Avro
   private static final ExecutorService READ_CHUNK_EXECUTOR = Executors.newFixedThreadPool(
       Runtime.getRuntime().availableProcessors(),
       new DaemonThreadFactory("DaVinci_Read_Chunk_Executor"));
-  public static final int CHUNK_SPLIT_THRESHOLD = 100;
+  public static final int DEFAULT_CHUNK_SPLIT_THRESHOLD = 100;
 
   private final DaVinciConfig daVinciConfig;
   private final ClientConfig clientConfig;
@@ -334,6 +334,10 @@ public class AvroGenericDaVinciClient<K, V> implements DaVinciClient<K, V>, Avro
     return this.storeDeserializerCache;
   }
 
+  DaVinciConfig getDaVinciConfig() {
+    return this.daVinciConfig;
+  }
+
   CompletableFuture<Map<K, V>> batchGetFromLocalStorage(Iterable<K> keys) {
     // expose underlying getAll functionality.
     Map<K, V> result = new VeniceConcurrentHashMap<>();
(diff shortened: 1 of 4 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/AvroGenericDaVinciClient.java
```diff
@@ -100,6 +106,16 @@ protected boolean removeEldestEntry(Map.Entry<Schema, GenericRecord> eldest) {
 
   private static final ThreadLocal<ReusableObjects> REUSABLE_OBJECTS = ThreadLocal.withInitial(ReusableObjects::new);
 
+  /**
+   * The following two fields are used to speed up the requests with a big number of keys:
+   * 1. Split the big request into smaller chunks.
+   * 2. Execute these chunks concurrently.
+   */
+  private static final ExecutorService READ_CHUNK_EXECUTOR = Executors.newFixedThreadPool(
+      Runtime.getRuntime().availableProcessors(),
```
sixpluszero, line 115: I don't have other good default number in mind, but I am wondering if we take up all threads, will it affect the follow up execution such as Frame transformation and other app computation?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/AvroGenericDaVinciClient.java
```diff
@@ -100,6 +106,16 @@ protected boolean removeEldestEntry(Map.Entry<Schema, GenericRecord> eldest) {
...
+      new DaemonThreadFactory("DaVinci_Read_Chunk_Executor"));
+  public static final int CHUNK_SPLIT_THRESHOLD = 100;
```
majisourav99, line 117: probably can do larger threshold like 500, and it will be still under 10ms?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/StorageClass.java
```diff
@@ -1,5 +1,24 @@
 package com.linkedin.davinci.client;
 
 public enum StorageClass {
-  DISK, MEMORY_BACKED_BY_DISK,
+  /**
```
majisourav99, line 4: Thanks for adding detailed doc for these!

clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/AvroGenericDaVinciClient.java
sixpluszero: Actually how about we use a config to control the chunk split threshold?

sixpluszero (dismissed):
Overall it looks good, only one comment on the chunk threshold. Because I am a bit worried if we want to iterate on the split then it will require code change and new release.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/AvroGenericDaVinciClient.java
gaojieliu: I am trying to eliminate the config if possible.
Exposing config is not always a good practice as it might confuse customers.
100 chunk size should be a good threshold, as it would roughly take 2-3ms to execute one chunk.
10ms is typically not good enough for DVC users.

If this default won't work out, I will add a new config later.

gaojieliu:
> Also, I am wondering if we can put keys to the same partition in a same query, will it help? Because I think in memory partition calculation should be fast. (Not necessarily need to happen in this PR but just wondering if that's a good direction to pursue in the future).

The per-partition chunk will be valuable if we can leverage multiget interface of RocksDB, but last time, when I benchmarked RocksDB multi-get API against PT mode, the result is not good (weird, but don't know why).

clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/AvroGenericDaVinciClient.java
majisourav99: Minimizing config is always a good idea, but having an option to override has its benefit as well. Wondering if we make the default to 150 to match with thin-client and also eliminate unnecessary chunking for such cases.

sixpluszero:
> > Also, I am wondering if we can put keys to the same partition in a same query, will it help? Because I think in memory partition calculation should be fast. (Not necessarily need to happen in this PR but just wondering if that's a good direction to pursue in the future).
> 
> The per-partition chunk will be valuable if we can leverage multiget interface of RocksDB, but last time, when I benchmarked RocksDB multi-get API against PT mode, the result is not good (weird, but don't know why).

Sure, I remember your experiment. Yeah, we can revisit that later. No need to prioritize that for current use case onboarding.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/AvroGenericDaVinciClient.java
gaojieliu: 150 is mainly for minimizing Router retry problem and it is not really related to DaVinci.
Also this number is tightly related to the performance behavior of DaVinci.
I don't see a big difference between 100 and 150.

For DVC, the goal is as fast as possible instead of single-digit ms as all the data is in memory, and the trade-off is mainly the additional context switches and latency reduction.

If you folks insist it, I can add it to DaVinciConfig, but I guess not all the customers have the experience about how to tune it properly without consulting us.

gaojieliu: Point taken, and I will add it to DaVinciConfig to make it overridable.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/AvroGenericDaVinciClient.java
sixpluszero: Synced offline, it should be good for now. It was validated and we will validate it in the new release as well.

sixpluszero (approved):

PR #944, opened 2024-04-15
[dvc] Do not throw DIV segment MISSING when a host regains leadership after EOP

lluwm:
…ip after EOP

DIV dumps its segments information into offset record periodically or before server shut down and loads the contents back during Helix state transition (OFFLINE->STANDBY) when the host is transitioned to follower and, when it becomes the leader, copied into the leader DIV.

For hybrid store partitions (after EOP), if a host went through a LEADER -> (OFFLINE ->) FOLLOWER -> LEADER transition, e.g. because of host shutdown, its DIV dumps and reloads segments as above described. However, today, if another host became the new leader (when it was offline or follower), did its leadershipduty by consuming messages from RT topic, producing to local VT etc, then when the previous host became the leader for the 2nd time, its DIV, after loading segment info from checkpoint, only contains the stale segment info in terms of the segment number for the RT topic, thus when it starts to consume, it will throw DIV MISSING error, which in most cases a false alarm.

This false alarm can happen very often as long as for a hybrid store partition, a host shuts down, rt topic keeps getting new data and consumed by new leader, old host starts again and regains the leadership for the same rt topic.

This rb introduces a flag to indicate if a segment is constructed from a check point. When a host regains its leadership and re-consumes RT, DIV tolerates the segment gap for very first message it meets and creates a new unregistered segment (based on the contents of the impending message) to replace it.

## How was this PR tested?

- Frankenwar test
- Added new unit test
- passed GHA

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/validation/PartitionTracker.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/validation/PartitionTracker.java
@@ -22,6 +22,7 @@ import com.linkedin.venice.kafka.protocol.state.ProducerPartitionState;
 import com.linkedin.venice.kafka.validation.Segment;
 import com.linkedin.venice.kafka.validation.checksum.CheckSumType;
 import com.linkedin.venice.message.KafkaKey;
+import com.linkedin.venice.meta.Version;
 import com.linkedin.venice.offsets.OffsetRecord;
 import com.linkedin.venice.pubsub.api.PubSubMessage;
 import com.linkedin.venice.utils.CollectionUtils;
@@ -275,7 +276,8 @@ public class PartitionTracker {
        * So, when a host regains its leadership and re-consumes RT, DIV tolerates the segment gap for very first message
        * it meets and creates a new unregistered segment (based on the contents fo the impending message) to replace it.
        */
-      if (endOfPushReceived && previousSegment.isCreatedFromCheckPoint()) {
+      if (endOfPushReceived && previousSegment.isCreatedFromCheckPoint()
+          && Version.isRealTimeTopic(consumerRecord.getTopicName())) {
         return initializeNewSegment(consumerRecord, true, true);
       }
 
@@ -369,7 +371,7 @@ public class PartitionTracker {
       String errorMsgIdentifier = consumerRecord.getTopicPartition().getPubSubTopic().getName() + "-"
           + consumerRecord.getTopicPartition().getPartitionNumber() + "-" + DataFaultType.UNREGISTERED_PRODUCER;
       if (!REDUNDANT_LOGGING_FILTER.isRedundantException(errorMsgIdentifier)) {
-        logger.warn("Will {}, endOfPushReceived=true, tolerateAnyMessageType=true", scenario);
+        logger.info("Will {}, endOfPushReceived=true, tolerateAnyMessageType=true", scenario);
       }
     } else {
       throw DataFaultType.UNREGISTERED_PRODUCER.getNewException(
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/validation/PartitionTracker.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/validation/PartitionTracker.java
@@ -242,29 +242,52 @@ public class PartitionTracker {
             null,
             endOfPushReceived);
       }
-      Segment newSegment = initializeNewSegment(consumerRecord, endOfPushReceived, true);
-      return newSegment;
-    } else {
-      int previousSegmentNumber = previousSegment.getSegmentNumber();
-      if (incomingSegmentNumber == previousSegmentNumber) {
-        return previousSegment;
-      } else if (incomingSegmentNumber == previousSegmentNumber + 1 && previousSegment.isEnded()) {
-        /** tolerateAnyMessageType should always be false in this scenario, regardless of {@param endOfPushReceived} */
-        return initializeNewSegment(consumerRecord, endOfPushReceived, false);
-      } else if (incomingSegmentNumber > previousSegmentNumber) {
-        if (tolerateMissingMsgs.get()) {
-          return initializeNewSegment(consumerRecord, endOfPushReceived, true);
-        } else {
-          throw DataFaultType.MISSING.getNewException(previousSegment, consumerRecord);
-        }
-      } else if (incomingSegmentNumber < previousSegmentNumber) {
-        throw DataFaultType.DUPLICATE.getNewException(previousSegment, consumerRecord);
-      } else {
-        // Defensive code.
-        throw new IllegalStateException(
-            "This condition should never happen. " + getClass().getSimpleName() + " may have a regression.");
-      }
+      return initializeNewSegment(consumerRecord, endOfPushReceived, true);
     }
+
+    int previousSegmentNumber = previousSegment.getSegmentNumber();
+    if (incomingSegmentNumber == previousSegmentNumber) {
+      return previousSegment;
+    }
+
+    if (incomingSegmentNumber == previousSegmentNumber + 1 && previousSegment.isEnded()) {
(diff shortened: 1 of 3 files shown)


gaojieliu:
For a particular node, if it transits from leader -> offline -> follower -> leader via a restart, while the previous leader doesn't consume any messages because of 5-min delay, the newly restarted leader will tolerate missing messages (not logging) when becoming a leader even there is true missing message issue, right?
It is hard to differentiate, but just want to explore whether this is expected or not.

lluwm:
Gaojie, that's correct. Unless the host can have a global view, it doesn't really know if the gap in segment is actually caused by a true data missing or another leader's consumption. This change simply suppress the warning logging for this scenario. However, it doesn't really change the behavior of how we handle it today. With or without this change, we will create a new unregistered segment based on the contents of the impending message.

huangminchn:
> Gaojie, that's correct. Unless the host can have a global view, it doesn't really know if the gap in segment is actually caused by a true data missing or another leader's consumption. This change simply suppress the warning logging for this scenario (again we are choosing between false positive (too many false alarms) vs. false negative (no alarm for true data missing) and I kind of prefer the latter, cause thousands of false warnings could hide the true ones in terms of our responses). However, it doesn't really change the behavior of how we handle it today. With or without this change, we will create a new unregistered segment based on the contents of the impending message and following up messages will be compared against the new one, so later data missing will be discovered and reported.

To achieve the global view, we need to pass the RT producer metadata into VT, evolving KME protocol. However, if we start considering DCR where leaders deliberately drop messages, passing RT producer metadata won't be enough; leaders need to somehow pass this information to follower, for example, in the next produced message into VT, specify how many messages are dropped in purpose, and tell followers forgot about computing checksum. This can be fairly complicated.

internal/venice-common/src/main/java/com/linkedin/venice/kafka/validation/Segment.java
```diff
@@ -76,6 +76,13 @@ public class Segment {
   // record the last producer message time stamp passed within the ConsumerRecord
   private long lastRecordProducerTimestamp = -1;
 
+  /**
+   * Indicate if a segment is constructed from a check point. When a host regains its leadership and consume RT,
+   * DIV tolerates the segment gap for very first message it meets and creates a new unregistered segment to replace it.
+   * Only set to true when a segment is created from a check point or copy constructed into the leader DIV.
+   */
+  private final boolean isCreatedFromCheckPoint;
```
huangminchn, line 84: I understand the intention is to cover an edge case for RT DIV info; what would be the impact for VT DIV info?

Will it introduce new edge case for VT? Or VT has the same problem and this fix covers it?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/validation/PartitionTracker.java
```diff
@@ -242,29 +242,52 @@ private Segment trackSegment(
...
+       * the new leader (when it was offline or follower) did its leadership duty by consuming messages from RT topic,
+       * producing to local VT etc., then when previous host became leader for the 2nd time, its DIV, after loading
+       * segment from checkpoint, only contains the stale segment info in terms of the segment number for the RT topic,
+       * thus when it starts to consume, it can see segment number gaps.
+       *
+       * So, when a host regains its leadership and re-consumes RT, DIV tolerates the segment gap for very first message
+       * it meets and creates a new unregistered segment (based on the contents fo the impending message) to replace it.
+       */
+      if (endOfPushReceived && previousSegment.isCreatedFromCheckPoint()) {
+        return initializeNewSegment(consumerRecord, true, true);
```
huangminchn, line 279: If we believe this case is expected and want to avoid confusing logs, we need to change the logging inside `initializeNewSegment()` too. If the incoming message is not SOS, a warning log will be printed.

huangminchn (commented):
Thanks Lei! Looks good overall!

gaojieliu:
There are 3 purposes in DIV from my understanding:
1. Catch messages gap.
2. Filter out duplicate messages because of producer retry.
3. Checksum to validate the integrity of the messages.

So this PR is mainly about item 1, and for this particular item, I think there is an alternative while using Kafka.
Kafka is using an ever-increasing offset, and if there is no offset gap, then missing message issue won't happen.
If there is no gap in consumed offset, I assume there is no message gap, then we can safely ignore/skip the missing message exception triggered by DIV.
The RT offset will be carried over during leadership handover, so I think offset-based check should be good enough to validate whether there is a true message gap or not.
WDYT?

cc @FelixGV 
Of coz, this mechanism won't work if some pub-sub system being used doesn't support an ever-increasing offset sequence.

huangminchn:
> There are 3 purposes in DIV from my understanding:
> 
> 1. Catch messages gap.
> 2. Filter out duplicate messages because of producer retry.
> 3. Checksum to validate the integrity of the messages.
> 
> So this PR is mainly about item 1, and for this particular item, I think there is an alternative while using Kafka. Kafka is using an ever-increasing offset, and if there is no offset gap, then missing message issue won't happen. If there is no gap in consumed offset, I assume there is no message gap, then we can safely ignore/skip the missing message exception triggered by DIV. The RT offset will be carried over during leadership handover, so I think offset-based check should be good enough to validate whether there is a true message gap or not. WDYT?
> 
> cc @FelixGV Of coz, this mechanism won't work if some pub-sub system being used doesn't support an ever-increasing offset sequence.

@gaojieliu Currently DIV is created, maintained and verified within Venice's tech stack, without depending on the correctness of the pub-sub client. Checking continuity of offset is a good addition but IMHO it couldn't replace DIV; otherwise, we won't be able to detect if producers lost data internally.

FelixGV:
> this mechanism won't work if some pub-sub system being used doesn't support an ever-increasing offset sequence.

I believe this is not true in Kafka BTW... in some broker failure scenarios, offsets could go back, from what I heard, though I don't recall the details of when that would happen.

internal/venice-common/src/main/java/com/linkedin/venice/kafka/validation/Segment.java
lluwm: Good point and I think we can restrict the condition further to only apply to RT data, e.g.

`if (endOfPushReceived && previousSegment.isCreatedFromCheckPoint())`
-->
`
      if (endOfPushReceived && previousSegment.isCreatedFromCheckPoint() && Version.isRealTimeTopic(consumerRecord.getTopicName()))
`
what do you think?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/validation/PartitionTracker.java
lluwm: we can downgrade the log to the `info` level.

gaojieliu:
> > this mechanism won't work if some pub-sub system being used doesn't support an ever-increasing offset sequence.
> 
> I believe this is not true in Kafka BTW... in some broker failure scenarios, offsets could go back, from what I heard, though I don't recall the details of when that would happen.

If offset can go back, it means Kafka could maintain duplicate offsets for different messages, which I don't know whether it would work properly for the Kafka API, such as seek operation..
Anyway, this idea won't work properly as DIV is an E2E validation, the offset-based approach can only validate the consumption, not producing part.

gaojieliu:
Hi @lluwm 
IIRC, the leader replica won't persist RT DIV info into offset record during checkpoint, which means the checkpointed offset only contains the DIV for VT producers.
If the above assumption is true, it won't help to reduce the missing message logging when processing RT messages during leadership handover, right?
During a restart, the new leader will contain nothing about RT DIV info and during a rebalance: leader -> follower -> leader, the `kafkaDataIntegrityValidatorForLeaders` can keep staled DIV info for RT topic.
So can you elaborate on how this PR will help reduce the missing message logs for RT messages during leadership handover?
Maybe I missed some basic knowledge about how it works today..

lluwm:
close it as we will have a new proposal for solving this problem

PR #945, opened 2024-04-15
[controller] Do not delete true backup version during Repush

majisourav99:
## Do not delete true backup version during Repush

During repush Venice deletes the actual backup version instead of current version. 
For example some store has versions: v9, v10. When a repush starts it creates v11 and it deletes v9 assuming it is a backup version, where as repush actually copies data from v10 to v11.
Now later if use wants to rollback to backup version they can only rollback to v10 which is exactly same as v11 as v9 was deleted.

This PR fixes that. It relies on `StoreBackupVersionCleanupService` to delete the backup version v10 asynchronously. Currently it checks `controller.backup.version.metadata.fetch.cleanup.enabled` to do a safer delete of previous current version which validates there are no reads to v10 version.

## How was this PR tested?

CI

 The change in backup version definition for repush, the backup version is the current version before push instead of version before that.


services/venice-controller/src/main/java/com/linkedin/venice/controller/StoreBackupVersionCleanupService.java
```diff
@@ -224,34 +225,44 @@ private boolean validateAllServerOnCurrentVersion(Store store, String clusterNam
...
+
+    if (!isCurrentVersionFromRepush && !whetherStoreReadyToBeCleanup(store, defaultBackupVersionRetentionMs, time)) {
       // not ready to clean up backup versions yet
       return false;
     }
 
     List<Version> versions = store.getVersions();
     List<Version> readyToBeRemovedVersions = new ArrayList<>();
-    int currentVersion = store.getCurrentVersion();
+    int currentVersionNum = store.getCurrentVersion();
```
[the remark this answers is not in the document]
majisourav99: Sure

services/venice-controller/src/main/java/com/linkedin/venice/controller/StoreBackupVersionCleanupService.java
```diff
@@ -224,34 +225,44 @@ private boolean validateAllServerOnCurrentVersion(Store store, String clusterNam
...
 
-    if (currentVersion == NON_EXISTING_VERSION) {
+    if (currentVersionNum == NON_EXISTING_VERSION) {
       return false;
     }
```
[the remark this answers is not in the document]
majisourav99: Sure

services/venice-controller/src/main/java/com/linkedin/venice/controller/StoreBackupVersionCleanupService.java
```diff
@@ -224,34 +225,44 @@ private boolean validateAllServerOnCurrentVersion(Store store, String clusterNam
...
+          readyToBeRemovedVersions.add(v);
+        }
+      });
+    } else {
+      versions.forEach(v -> {
+        if (v.getNumber() < currentVersionNum) {
+          readyToBeRemovedVersions.add(v);
+        }
+      });
+    }
```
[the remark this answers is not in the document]
majisourav99: Good idea, it will make the code cleaner.

internal/venice-common/src/main/java/com/linkedin/venice/meta/VersionImpl.java
```diff
@@ -106,6 +106,12 @@ public void setAge(Duration age) {
     this.storeVersion.createdTime = System.currentTimeMillis() - age.toMillis();
   }
 
+  @JsonIgnore
+  @Override
+  public void setCreatedTime(Duration age) {
+    this.storeVersion.createdTime = age.toMillis();
+  }
+
```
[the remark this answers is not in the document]
majisourav99: Its used it a marker to prevent true backup version deletion and piggybacking on an existing field for that. Using a new field would be cleaner but seems like an overkill.

services/venice-controller/src/main/java/com/linkedin/venice/controller/StoreBackupVersionCleanupService.java
```diff
@@ -224,34 +228,29 @@ private boolean validateAllServerOnCurrentVersion(Store store, String clusterNam
...
-    if (currentVersion == NON_EXISTING_VERSION) {
-      return false;
-    }
-
     // Do not delete version unless all routers and all servers are on same current version
     if (multiClusterConfig.getControllerConfig(clusterName).isBackupVersionMetadataFetchBasedCleanupEnabled()
-        && (!validateAllRouterOnCurrentVersion(store, clusterName, currentVersion)
-            || !validateAllServerOnCurrentVersion(store, clusterName, currentVersion))) {
+        && (!validateAllRouterOnCurrentVersion(store, clusterName, currentVersionNum)
+            || !validateAllServerOnCurrentVersion(store, clusterName, currentVersionNum))) {
```
gaojieliu, line 244: These two function calls are fairly expensive, so shall we make sure there are lingering backup versions before running such validation?

services/venice-controller/src/main/java/com/linkedin/venice/controller/StoreBackupVersionCleanupService.java
```diff
@@ -286,6 +285,13 @@ protected boolean cleanupBackupVersion(Store store, String clusterName) {
     return true;
   }
 
+  private boolean readyToBeRemoved(Version v, boolean isRepush, int currentVersionNum) {
+    if (isRepush) {
+      return v.getCreatedTime() == Duration.ZERO.toMillis();
+    }
+    return v.getNumber() < currentVersionNum;
+  }
+
```
gaojieliu, line 294: I wondered whether this logic will work well or not.
Let us say we have two versions:
v1,
v2,
And v3 is the repushed version from v2.

So when cleaning up the store versions, v2 will be removed as its created time is 0, and the true backup version will never be removed as its createdTime is not `0`, right?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -2496,6 +2498,11 @@ private Pair<Boolean, Version> addVersion(
             }
             // Update ZK with the new version
             store.addVersion(version, true);
+            // update the age of the curent version to 0 for deletion in StoreBackupVersionCleanupService
+            if (isRepush && multiClusterConfigs.getControllerConfig(clusterName)
+                .isBackupVersionMetadataFetchBasedCleanupEnabled()) {
+              store.getVersion(currentVersionBeforePush).get().setCreatedTime(Duration.ZERO);
+            }
```
gaojieliu, line 2505: This may not be fully correct in all scenarios.
Let us say:
```
1. User runs a batch push job, which produces a version: v10.
2. The job succeeds in two regions with: v9, v10.
3. The job fails in one region: v9.
4. We trigger a repush based on v10.
5. Two good regions will have: v9, v10, v11 and the bad region will have v9, v11.
6. Based on this logic, the v10 in good regions will be marked with created time: 0 and the v9 in the bad region will be marked in the same way.
7. v10 will be removed from good regions and v9 will be removed from the bad region.
```
The above doesn't sound good and is not consistent.

I think we should introduce a new version-level field called re-push source version, so that we can explicitly know which versions to cleanup.
Right now, I feel the logic is a little tricky and error prone.

services/venice-controller/src/test/java/com/linkedin/venice/controller/TestStoreBackupVersionCleanupService.java
```diff
@@ -74,14 +76,18 @@ private Store mockStore(
     doReturn(latestVersionPromoteToCurrentTimestamp).when(store).getLatestVersionPromoteToCurrentTimestamp();
     doReturn(currentVersion).when(store).getCurrentVersion();
     List<Version> versionList = new ArrayList<>();
+    AtomicInteger i = new AtomicInteger();
```
gaojieliu, line 79: I think we need some E2E integration tests.

gaojieliu (changes requested):

services/venice-controller/src/main/java/com/linkedin/venice/controller/StoreBackupVersionCleanupService.java
majisourav99: Good idea, that probably skip running the service on store.versions.size == 1.

services/venice-controller/src/main/java/com/linkedin/venice/controller/StoreBackupVersionCleanupService.java
majisourav99: It will be removed in the later iteration after regular push, right?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
majisourav99: emm V9 wont be deleted in bad region as we have check for keep min number of version.

gaojieliu: Can you show me the logic regarding ^?
Also v9's created time will be updated to be `0` based on the new logic IIUC.
I think it is confusing, and it is better to make the logic clean to avoid tricky bugs.

services/venice-controller/src/main/java/com/linkedin/venice/controller/StoreBackupVersionCleanupService.java
gaojieliu: Well, what if there is no regular push in the following month? 
This comment is related to the following comment, and the main cause is that we can't differentiate which version is the source version for the repushed current version, and if we have that info, the logic will be very straightforward.
The source version cleanup will be independent from the true backup version cleanup.

services/venice-controller/src/main/java/com/linkedin/venice/controller/StoreBackupVersionCleanupService.java
```diff
@@ -139,10 +139,20 @@ CloseableHttpAsyncClient getHttpAsyncClient() {
...
+      Store store,
+      long defaultBackupVersionRetentionMs,
+      Time time,
+      int currentVersion) {
     if (store.getCurrentVersion() == NON_EXISTING_VERSION || store.getVersions().size() < 2) {
       return false;
     }
+
+    if (store.getVersion(currentVersion).get().getRepushSourceVersion() > NON_EXISTING_VERSION
+        && store.getVersions().size() > 2) {
```
gaojieliu, line 152: So shall we delete the source version even the total number of versions are `2`?

services/venice-controller/src/main/java/com/linkedin/venice/controller/StoreBackupVersionCleanupService.java
```diff
@@ -232,13 +242,14 @@ private boolean validateAllServerOnCurrentVersion(Store store, String clusterNam
...
+    if (!whetherStoreReadyToBeCleanup(store, defaultBackupVersionRetentionMs, time, currentVersion)) {
       // not ready to clean up backup versions yet
       return false;
     }
 
     List<Version> versions = store.getVersions();
-    int currentVersion = store.getCurrentVersion();
 
     // Do not delete version unless all routers and all servers are on same current version
     if (multiClusterConfig.getControllerConfig(clusterName).isBackupVersionMetadataFetchBasedCleanupEnabled()
```
gaojieliu, line 255: How can we enable this feature by default because of the new repush cleanup logic?
Otherwise, we need to delay the deletion of the source version for a while to avoid race conditions.
I guess it is good to add a delay regardless, such as 1h?

services/venice-controller/src/main/java/com/linkedin/venice/controller/StoreBackupVersionCleanupService.java
```diff
@@ -249,8 +260,19 @@ protected boolean cleanupBackupVersion(Store store, String clusterName) {
...
+    for (Version version: versions) {
+      if (version.getRepushSourceVersion() > NON_EXISTING_VERSION) {
+        repushSourceVersion = version.getRepushSourceVersion();
+      }
+    }
+    int finalRepushSourceVersion = repushSourceVersion;
+    List<Version> readyToBeRemovedVersions = versions.stream()
+        .filter(
+            v -> finalRepushSourceVersion > NON_EXISTING_VERSION
+                ? v.getNumber() == finalRepushSourceVersion
```
gaojieliu, line 273: So here, we either cleanup the source version or true backup version, which passes the delay, not both, right?
It is fine to have such logic, and please add some javadoc to describe such desired behavior, otherwise, it is a little confusing.

services/venice-controller/src/main/java/com/linkedin/venice/controller/server/CreateVersion.java
```diff
@@ -249,6 +250,8 @@ public Route requestTopicForPushing(Admin admin) {
 
         String targetedRegions = request.queryParams(TARGETED_REGIONS);
 
+        int repushSourceVersion = Integer.parseInt(request.queryParams(REPUSH_SOURCE_VERSION));
```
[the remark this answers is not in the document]
majisourav99: Sure, will add a null check. I kept it like this so that it caught all the code path not passing the param.

services/venice-controller/src/main/java/com/linkedin/venice/pushmonitor/AbstractPushMonitor.java
```diff
@@ -1008,7 +1009,10 @@ protected void handleCompletedPush(String topic) {
           e);
     }
     try {
-      storeCleaner.retireOldStoreVersions(clusterName, storeName, false, -1);
+      Store store = metadataRepository.getStore(storeName);
+      if (store.getVersion(versionNumber).get().getRepushSourceVersion() <= NON_EXISTING_VERSION) {
```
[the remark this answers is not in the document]
majisourav99: Sure.

services/venice-controller/src/main/java/com/linkedin/venice/controller/StoreBackupVersionCleanupService.java
majisourav99: If we add a delay before deleting the repush source version, I guess we can delete if total  number of versions are 2.

services/venice-controller/src/main/java/com/linkedin/venice/controller/StoreBackupVersionCleanupService.java
majisourav99: Sure will add a javadoc, its basically first deletes the repush source version, then in later iteration of the loop after threshold wait time (eg 7 days) it will delete the true backup version.

services/venice-controller/src/main/java/com/linkedin/venice/controller/StoreBackupVersionCleanupService.java
majisourav99: Sure, having some wait time regardless is good idea.

gaojieliu (approved):
Thanks for the fix!

FelixGV:
@smaji, after my rebase integrated this PR, I found three naked gets introduced by this change. I fixed them so that the code compiles, but we may wish to handle them differently than what I did. Can you please take a look at [this commit](https://github.com/linkedin/venice/pull/990/commits/59900c8f40891bc5ec0453d9e40194e2784dffbe) and LMK if you have any concerns with the way I handled these naked gets? Thanks.

PR #947, opened 2024-04-17
[dvc] Add config to return Da Vinci specific ExecutionStatus for Errors

m-nagarajan:
## Summary
* Added a config `use.da.vinci.specific.execution.status.for.error` to enable/disable returning newly added Da Vinci specific `ExecutionStatus` for Errors in Da VInci clients.
* Its disabled by default and can be enabled back again after the new code is in all components

## How was this PR tested?
GH CI

New config `use.da.vinci.specific.execution.status.for.error` to enable/disable the new statuses.


gaojieliu (approved):
Thanks!

PR #948, opened 2024-04-18
[cc] Expose partitionCount from VeniceChangelogConsumer

kcepmada:
## Summary
Exposed total partition count of store from VeniceChangelogConsumer to programmatically facilitate parallel processing (eg. VeniceIO).

## How was this PR tested?
GH CI

New method `getPartitionCount` added to experimental `VeniceChangelogConsumer` interface.

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumer.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumer.java
@@ -18,6 +18,11 @@ import java.util.concurrent.CompletableFuture;
 @Experimental
 public interface VeniceChangelogConsumer<K, V> {
   /**
+   * @return total number of store partitions
+  */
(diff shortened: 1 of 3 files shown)


nisargthakkar:
I don't believe this method belongs in the changelog client's API. Have you already discussed this change with anyone in the Venice team already? If not, let's get a thread started on our community slack.

nisargthakkar:
One thing that would fit better in the API would be to have an API called `getPartitions` that returns the Set of partitions that the consumer is currently subscribed to. 

With this API in place, the users of the client can use `subscribeAll` and then call `getPartitions().size()` to find out the partition count

ZacAttack (approved):
Talked about this internally.  Currently we're in a pretty broken state with regards to how beam clients get configured without this.  This I'm not certain is a great solution for down the road enhancements, but I don't really have a better answer for the beam connector unless we take ownership of the connector and figure out a better way to dynamically get this information in the midst of things like repartititoning.

PR #949, opened 2024-04-19
[router] Add a dedicated executor for SSL handshakes in routers

nisargthakkar:
## Add a dedicated executor for SSL handshakes in routers

Add a dedicated thread pool executor for SSL tasks in router. This same feature has been running in production and should be stable. The value of `router.client.ssl.handshake.threads` controls the number of threads that will be used. A value of 0 (default) disables the feature.

## How was this PR tested?

GH CI

### DIFF
--- a/services/venice-router/src/main/java/com/linkedin/venice/router/RouterServer.java
+++ b/services/venice-router/src/main/java/com/linkedin/venice/router/RouterServer.java
@@ -2,6 +2,7 @@ package com.linkedin.venice.router;
 
 import static com.linkedin.venice.CommonConfigKeys.SSL_FACTORY_CLASS_NAME;
 import static com.linkedin.venice.VeniceConstants.DEFAULT_SSL_FACTORY_CLASS_NAME;
+import static com.linkedin.venice.utils.concurrent.BlockingQueueType.LINKED_BLOCKING_QUEUE;
 
 import com.linkedin.alpini.base.concurrency.AsyncFuture;
 import com.linkedin.alpini.base.concurrency.TimeoutProcessor;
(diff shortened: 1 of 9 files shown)


services/venice-router/src/main/java/com/linkedin/venice/router/RouterServer.java
```diff
@@ -629,25 +630,14 @@ public AsyncFuture<LongSupplier> getLongTailRetryMilliseconds(
...
-        int clientSslHandshakeThreads = config.getClientSslHandshakeThreads();
-        int maxConcurrentClientSslHandshakes = config.getMaxConcurrentClientSslHandshakes();
-        int clientSslHandshakeAttempts = config.getClientSslHandshakeAttempts();
-        long clientSslHandshakeBackoffMs = config.getClientSslHandshakeBackoffMs();
-        if (useEpoll) {
-          sslResolverEventLoopGroup = new EpollEventLoopGroup(clientSslHandshakeThreads, sslHandshakeExecutor);
-        } else {
-          sslResolverEventLoopGroup = new NioEventLoopGroup(clientSslHandshakeThreads, sslHandshakeExecutor);
-        }
-        sslInitializer.enableResolveBeforeSSL(
```
gaojieliu, line 646: I assume this thread pool is used for DNS resolution, not for every phase of ssl handshake.
So we are confident that this feature is not required any more, right?

nisargthakkar: Yeah. This was never used. We tried enabling it once, but we could never roll it out

gaojieliu (approved):
Thanks for the change!

PR #950, opened 2024-04-20
[server][da-vinci] Bumped RocksDB dep and adopt multiget async io by default

gaojieliu:
This PR bumps up the RocksDB dep and expose a config to enable async io for multi-get and the default value is true.
rocksdb.read.async.io.enabled: default true
In theoy, with this config and posix filesystem, RocksDB multiget API will be speeded up quite a bit based on the benchmarking: https://rocksdb.org/blog/2022/10/07/asynchronous-io-in-rocksdb.html

So far, such optimization only applies to the chunk lookup for large value/rmd, and if it is proved to be more performant by checking the lookup latency for large value in the read path, we can apply such optimization in more areas:
1. DaVinci with DISK mode.
2. Ingestion code path by looking up entries in batch for AA/WC use cases.
3. Regular read path. So far, multi-get API can only be used against the same RocksDB database, so we might need some re-org of RocksDB databases if this API is truly helpful.

## How was this PR tested?
CI

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/AbstractStorageEngine.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/AbstractStorageEngine.java
@@ -496,6 +496,13 @@ public abstract class AbstractStorageEngine<Partition extends AbstractStoragePar
     });
   }
 
+  public List<byte[]> multiGet(int partitionId, List<byte[]> keys) throws VeniceException {
+    return executeWithSafeGuard(partitionId, () -> {
+      AbstractStoragePartition partition = getPartitionOrThrow(partitionId);
+      return partition.multiGet(keys);
+    });
+  }
+
   public ByteBuffer get(int partitionId, byte[] key, ByteBuffer valueToBePopulated) throws VeniceException {
     return executeWithSafeGuard(partitionId, () -> {
       AbstractStoragePartition partition = getPartitionOrThrow(partitionId);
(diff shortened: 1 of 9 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/AbstractStorageEngine.java
```diff
@@ -539,6 +546,14 @@ public byte[] getReplicationMetadata(int partitionId, byte[] key) {
     });
   }
 
+  public List<byte[]> multiGetReplicationMetadata(int partitionId, List<byte[]> keys) {
+    System.out.println("key size in multiGetReplicationMetadata:" + keys.size());
```
majisourav99, line 550: can you remove this line?

majisourav99 (dismissed):
LGTM! Do we need to add any new tests or we are relying on existing WC chunking tests for this?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/AbstractStorageEngine.java
gaojieliu: Good catch!

gaojieliu:
> LGTM! Do we need to add any new tests or we are relying on existing WC chunking tests for this?

We have existing both unit test and integration test to cover this change, so I think it is enough.

majisourav99 (approved):

PR #953, opened 2024-04-22
[dvc] Add experimental annotation to DaVinciRecordTransformer, and add back the key parameter to put

kvargha:

1. Added the `Experimental` annotation to `DaVinciRecordTransformer`, since it's still in its experimental stage.
2. Added back key as a parameter to the `put` function in `DaVinciRecordTransformer` in case if the user wants to utilize it for transformation. Updated the documentation to reflect this, which is viewable [here](https://kvargha.com/venice/docs/user_guide/read_api/da_vinci_client).

## How was this PR tested?

CI


ZacAttack (approved):
Looks good!  Thanks for the quick turn around!

PR #955, opened 2024-04-22
Add heartbeat lag metrics to change capture client

ZacAttack:
## Add heartbeat lag metrics to change capture client

This works similar to the heartbeat lag measure done in server.  This tracks the time of last consumed heartbeat from the version topic.

Resolves #XXX

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumerImpl.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumerImpl.java
@@ -113,6 +113,7 @@ public class VeniceChangelogConsumerImpl<K, V> implements VeniceChangelogConsume
   protected final ChunkAssembler chunkAssembler;
 
   protected final BasicConsumerStats changeCaptureStats;
+  protected final HeartbeatReporterThread heartbeatReporterThread;
 
   public VeniceChangelogConsumerImpl(
       ChangelogClientConfig changelogClientConfig,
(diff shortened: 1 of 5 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumerImpl.java
```diff
@@ -852,4 +883,26 @@ protected PubSubConsumerAdapter getPubSubConsumer() {
...
+    }
+
+    @Override
+    public void run() {
+      while (!Thread.interrupted()) {
+        for (Long lastHeartbeat: currentVersionLastHeartbeat.values()) {
+          changeCaptureStats.recordLag(System.currentTimeMillis() - lastHeartbeat);
+        }
+        try {
+          TimeUnit.SECONDS.sleep(60000L);
```
sixpluszero, line 899: This is 60K seconds?

ZacAttack: oh snap, good catch.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/stats/BasicConsumerStats.java
```diff
@@ -0,0 +1,20 @@
+package com.linkedin.davinci.consumer.stats;
+
+import com.linkedin.venice.stats.AbstractVeniceStats;
+import io.tehuti.metrics.MetricsRepository;
+import io.tehuti.metrics.Sensor;
+import io.tehuti.metrics.stats.Max;
+
+
+public class BasicConsumerStats extends AbstractVeniceStats {
```
xunyin8, line 9: Besides lag, should we also have consumption rate and record count for debugging purpose in case lag is high/stuck?

ZacAttack: That's an interesting idea, but not sure how best to do it.  record count/rate isn't really controlled by us.  It's really an artifact of how often the user calls poll().  I'm concerned that stats that show this is small will be distracting, and the partner will insist it's on us.  Though I think it'd be useful in so far as the partner wants to just have stats for the performance of their consumer.

xunyin8 (dismissed):

xunyin8 (approved):

PR #957, opened 2024-04-23
[server][da-vinci] Introduced a new mode to disable record-level metric when bootstrapping current version

gaojieliu:
During the analysis of DaVinci bootstrapping, we realized that metric collection takes a significant amount of cpu resources (>60%) when bootstrapping a store with many small entries. This PR adds a new config to disable record-level metrics when bootstrapping current version:
server.record.level.metrics.when.bootstrapping.current.version.enabled: default true

When disabling the above config, the record-level metrics will be disabled during bootstrapping, but the record-level metric emission will resume after all the subscribed partitions report ready to serve.

## How was this PR tested?
CI

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/config/VeniceServerConfig.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/config/VeniceServerConfig.java
@@ -98,6 +98,7 @@ import static com.linkedin.venice.ConfigKeys.SERVER_PROMOTION_TO_LEADER_REPLICA_
 import static com.linkedin.venice.ConfigKeys.SERVER_PUBSUB_CONSUMER_POLL_RETRY_BACKOFF_MS;
 import static com.linkedin.venice.ConfigKeys.SERVER_PUBSUB_CONSUMER_POLL_RETRY_TIMES;
 import static com.linkedin.venice.ConfigKeys.SERVER_QUOTA_ENFORCEMENT_ENABLED;
+import static com.linkedin.venice.ConfigKeys.SERVER_RECORD_LEVEL_METRICS_WHEN_BOOTSTRAPPING_CURRENT_VERSION_ENABLED;
 import static com.linkedin.venice.ConfigKeys.SERVER_REMOTE_CONSUMER_CONFIG_PREFIX;
 import static com.linkedin.venice.ConfigKeys.SERVER_REMOTE_INGESTION_REPAIR_SLEEP_INTERVAL_SECONDS;
 import static com.linkedin.venice.ConfigKeys.SERVER_REST_SERVICE_EPOLL_ENABLED;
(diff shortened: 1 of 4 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -3948,6 +3964,20 @@ protected boolean shouldUpdateUpstreamOffset(PubSubMessage<KafkaKey, KafkaMessag
    */
   protected abstract Set<String> maybeSendIngestionHeartbeat();
 
+  private void mayResumeRecordLevelMetricsForCurrentVersion() {
+    if (recordLevelMetricEnabled.get()) {
+      return;
+    }
+    for (PartitionConsumptionState partitionConsumptionState: partitionConsumptionStateMap.values()) {
```
sixpluszero, line 3971: This is consuming, so you are resuming after ready-to-serve for current version?

tiny nit: may->maybe as

clients/da-vinci-client/src/main/java/com/linkedin/davinci/config/VeniceServerConfig.java
```diff
@@ -758,6 +760,8 @@ public VeniceServerConfig(VeniceProperties serverProperties, Map<String, Map<Str
         serverProperties.getInt(SERVER_DEDICATED_CONSUMER_POOL_SIZE_FOR_AA_WC_LEADER, 5);
     useDaVinciSpecificExecutionStatusForError =
         serverProperties.getBoolean(USE_DA_VINCI_SPECIFIC_EXECUTION_STATUS_FOR_ERROR, false);
+    recordLevelMetricWhenBootstrappingCurrentVersionEnabled =
+        serverProperties.getBoolean(SERVER_RECORD_LEVEL_METRICS_WHEN_BOOTSTRAPPING_CURRENT_VERSION_ENABLED, true);
```
sixpluszero, line 764: Should we default it to false, as it is impacting all DVC and server?

sixpluszero (approved):
overall it is good, left two comments.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
gaojieliu: Yes, and the metrics are still useful at the steady state.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/config/VeniceServerConfig.java
gaojieliu: The default is to keep the current behavior.

PR #962, opened 2024-04-24
[router][server][alpini] Add support for IdentityParser in SslInitializer

nisargthakkar:
## Add support for IdentityParser in SslInitializer

Currently Routers and Servers log the CN to help with investigation of the callers. However, this information is insufficient and we'd like to log the service principal name instead.

## How was this PR tested?

GH CI. Will be tested in staging env


gaojieliu:
Code changes look good to me and you need to fix the unit test coverage failure.

gaojieliu (approved):

nisargthakkar:
Thanks @gaojieliu

PR #963, opened 2024-04-25
[controller] Dont call backup version cleanup service on stores without backup version

majisourav99:
## Dont call backup version cleanup service on stores without backup version

Since fetching the current version from all router and server is an expensive operation, we should not call backup version cleanup on stores without current version or backup version. Also add a keep-alive period of 1hr to the http client.

## How was this PR tested?

GHCI

### DIFF
--- a/services/venice-controller/src/main/java/com/linkedin/venice/controller/StoreBackupVersionCleanupService.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/controller/StoreBackupVersionCleanupService.java
@@ -32,8 +32,10 @@ import org.apache.http.HttpResponse;
 import org.apache.http.HttpStatus;
 import org.apache.http.client.config.RequestConfig;
 import org.apache.http.client.methods.HttpGet;
+import org.apache.http.impl.client.DefaultConnectionKeepAliveStrategy;
 import org.apache.http.impl.nio.client.CloseableHttpAsyncClient;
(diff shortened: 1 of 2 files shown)


services/venice-controller/src/main/java/com/linkedin/venice/controller/StoreBackupVersionCleanupService.java
```diff
@@ -107,6 +106,12 @@ protected StoreBackupVersionCleanupService(
     this.httpAsyncClient = HttpAsyncClients.custom()
         .setDefaultRequestConfig(RequestConfig.custom().setSocketTimeout(10000).build())
         .setSSLContext(sslFactory.map(SSLFactory::getSSLContext).orElse(null))
+        .setKeepAliveStrategy(new DefaultConnectionKeepAliveStrategy() {
```
sixpluszero, line 109: This is to make client connection alive for 1h? Just curious why do we need this.

sixpluszero (approved):

services/venice-controller/src/main/java/com/linkedin/venice/controller/StoreBackupVersionCleanupService.java
majisourav99: yeah, just an attempt to keep connection open longer

PR #964, opened 2024-04-26
[controller] Do not delete parent VT for target colo batch push

majisourav99:
## Do not delete parent VT for target colo batch push

Concurrent pushes are checked using parent VT. For batch store target colo push, it should not delete the parent VT after push finishes in target colo. It will delete the topic on completion of the rest of colo ingestion. Also updated overallStatus in VPJ only if rest of the colo push finishes.

## How was this PR tested?

GHCI


services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
```diff
@@ -3526,13 +3526,32 @@ private OfflinePushStatusInfo getOffLineJobStatus(
...
-            currentReturnStatus,
-            currentReturnStatusDetails);
+        try (AutoCloseableLock ignore = resources.getClusterLockManager().createStoreWriteLock(storeName)) {
+          ReadWriteStoreRepository repository = resources.getStoreMetadataRepository();
+          Store parentStore = repository.getStore(storeName);
+          boolean isTargetRegionPush = !StringUtils.isEmpty(targetedRegions);
+          // Truncate topic after push is in terminal state if
+          // 1. Its a hybrid store or regular push
+          // 3. If target push is enabled and its previously pushed by previous colo push (status == PUSHED)
+          if (!isTargetRegionPush || parentStore.getVersion(versionNum).get().getStatus().equals(PUSHED)
```
huangminchn, line 3539: `PUSHED` means it finishes in all 3 colos?

majisourav99: Here PUSHED means it pushed to the target colo and eventually to the other colo. If the push fails at any step, it will be updated to KILLED.

huangminchn (changes requested):

huangminchn (approved):
Thanks Sourav!

PR #966, opened 2024-04-26
[changelog] Let bootstrapping changelog consumer consumes after-image

sixpluszero:
## [changelog] Let bootstrapping changelog consumer consumes after-image

Make bootstrapping consumer also consume after-image data from version topic, instead of before-after image data from cc topic. Fixed a few small bugs during the journey, and clean up the integration tests. 
Bugs:
1. seekToChekpoint needs to be blocked by get(), otherwise consumer will jump to non-deterministic offsets and yield wrong results.
2. when persisting after-image values to local storage, we should extra bytes from ByteBuffer, instead of converting to array directly.
## How was this PR tested?

Added new integration tests and clean up old ones. Now we have 1 integration tests that cover ingestion bootstrapping consumer with/without bootstrap and 1/3 parallel consumers, which replace a few integration tests that serve the same purpose but with a lot of duplication in code. 
For the existing TestChangelogConsumer#testAAwithStoreView(), I fixed the flakiness in the test, but the test is still too big and messy, consider refactoring it in the future.

Bootstrapping changelog consumer will only consume after-image change events now.


gaojieliu:
Is this change still necessary as we are planning to deprecate after-image support?
cc @ZacAttack

sixpluszero:
> Is this change still necessary as we are planning to deprecate after-image support? cc @ZacAttack

This is the intended direction to switch from before/after to after-only image for bootstrapping changelog consumer.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangeCoordinate.java
```diff
@@ -73,7 +73,7 @@ protected String getTopic() {
     return topic;
   }
 
-  protected PubSubPosition getPosition() {
+  public PubSubPosition getPosition() {
```
ZacAttack, line 76: Do we really need to make this public?  The coordinate I think is exposed to customers, I'd rather they knew as little as possible about the internals of this object.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/InternalLocalBootstrappingVeniceChangelogConsumer.java
```diff
@@ -470,15 +466,20 @@ public CompletableFuture<Void> seekWithBootStrap(Set<Integer> partitions) {
...
+      } catch (Exception e) {
+        throw new VeniceException("Caught exception when seeking to bootstrap", e);
+      }
 
       // Poll until we've caught up completely for all subscribed partitions.
       while (bootstrapStateMap.entrySet()
           .stream()
           .anyMatch(s -> s.getValue().bootstrapState.equals(PollState.CATCHING_UP))) {
-        pollAndCatchup(5000L, ChangeCaptureView.CHANGE_CAPTURE_TOPIC_SUFFIX);
+        pollAndCatchup(5000L, "");
```
ZacAttack, line 482: Let's add a todo here similar to the PR I opened, we might down the road need to tweak this, but for now it's fine since we're extending AfterImageConsumer specifically (so theres an implication that this only works on that type of topic).

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangeCoordinate.java
sixpluszero: I think? I want to print out some offset during debugging, let me check

ZacAttack (commented):
This overall looks good, just some nits

ZacAttack (approved):
Looks good!

PR #967, opened 2024-04-29
[server] - Fix consumer leaks in TM that occur when the thread owning consumer is interrupted

sushantmane:
## Fix topic manager consumer leaks that occur when the thread owning it is interrupted
Currently, if a thread owning a consumer is interrupted, the consumer is not 
returned to the consumer pool. This occurs because the blocking calls throw an 
interrupted exception when invoked by a thread with the interrupt status set. 
In this case, it is the `LBQ::put` call in `TMDF::releaseConsumer`.

As a result, we eventually run out of consumers to execute metadata requests. 
This situation typically arises when non-blocking metrics collection threads 
cancel active executions to terminate slow queries.

## How was this PR tested?
CI

### DIFF
--- a/internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicMetadataFetcher.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicMetadataFetcher.java
@@ -1,6 +1,7 @@
 package com.linkedin.venice.pubsub.manager;
 
 import static com.linkedin.venice.pubsub.PubSubConstants.PUBSUB_OFFSET_API_TIMEOUT_DURATION_DEFAULT_VALUE;
+import static com.linkedin.venice.pubsub.manager.TopicManagerStats.SENSOR_TYPE.CONSUMER_ACQUISITION_WAIT_TIME;
 import static com.linkedin.venice.pubsub.manager.TopicManagerStats.SENSOR_TYPE.CONTAINS_TOPIC;
 import static com.linkedin.venice.pubsub.manager.TopicManagerStats.SENSOR_TYPE.GET_OFFSET_FOR_TIME;
 import static com.linkedin.venice.pubsub.manager.TopicManagerStats.SENSOR_TYPE.GET_PARTITION_LATEST_OFFSETS;
@@ -49,6 +50,7 @@ import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Supplier;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
@@ -86,6 +88,7 @@ class TopicMetadataFetcher implements Closeable {
   private final Map<PubSubTopicPartition, ValueAndExpiryTime<Long>> lastProducerTimestampCache =
       new VeniceConcurrentHashMap<>();
   private final long cachedEntryTtlInNs;
+  private final AtomicInteger consumerWaitListSize = new AtomicInteger(0);
 
   TopicMetadataFetcher(
       String pubSubClusterAddress,
@@ -125,6 +128,8 @@ class TopicMetadataFetcher implements Closeable {
         new DaemonThreadFactory("TopicMetadataFetcherThreadPool"));
     threadPoolExecutor.allowCoreThreadTimeOut(true);
 
+    stats.registerTopicMetadataFetcherSensors(this);
+
     LOGGER.info(
         "Initialized TopicMetadataFetcher for pubSubClusterAddress: {} with consumer pool size: {} and thread pool size: {}",
         pubSubClusterAddress,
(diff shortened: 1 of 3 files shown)


internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicMetadataFetcher.java
```diff
@@ -150,22 +155,25 @@ class TopicMetadataFetcher implements Closeable {
...
 
   // acquire the consumer from the pool
-  private PubSubConsumerAdapter acquireConsumer() {
+  PubSubConsumerAdapter acquireConsumer() {
     try {
-      return pubSubConsumerPool.take();
+      consumerWaitListSize.incrementAndGet();
+      long startTime = System.nanoTime();
+      PubSubConsumerAdapter consumerAdapter = pubSubConsumerPool.take();
+      stats.recordLatency(CONSUMER_ACQUISITION_WAIT_TIME, startTime);
```
lluwm, line 163: Shall we make recordLatency thread safe?

sushantmane: I believe metric code paths are already thread safe, e.g., https://github.com/tehuti-io/tehuti/blob/a50499f2be2d8fffbf0e653cada28b48a5262b19/src/main/java/io/tehuti/metrics/Sensor.java#L104

lluwm: I think was worried about `TopicManagerStats.sensorsByTypes` which is not a threadsafe data structure, but looks like we only create the contents once in the constructor and no updates to it. So, that sounds good.

One more question, who blocked `releaseConsumer` in the first place, was it `acquireConsumer`? If I understand, we should always have capacity for `releaseConsumer` to complete, right?

gaojieliu: That is correct, and Tehuti lib is thread safe.

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicMetadataFetcher.java
```diff
@@ -150,22 +155,25 @@ class TopicMetadataFetcher implements Closeable {
...
   // release the consumer back to the pool
-  private void releaseConsumer(PubSubConsumerAdapter pubSubConsumerAdapter) {
-    try {
-      pubSubConsumerPool.put(pubSubConsumerAdapter);
-    } catch (InterruptedException e) {
-      Thread.currentThread().interrupt();
-      throw new VeniceException("Interrupted while releasing pubSubConsumerAdapter", e);
+  void releaseConsumer(PubSubConsumerAdapter pubSubConsumerAdapter) {
+    if (!pubSubConsumerPool.offer(pubSubConsumerAdapter)) {
+      LOGGER.error("Failed to release pubSubConsumerAdapter back to the pool. Closing the consumer.");
```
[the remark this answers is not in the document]
sushantmane: Good catch. Let me update the log. 

> do we have logic to close this consumer elsewhere?
Aside all consumers will be closed in the end since we keep in `closeables` lists.

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicMetadataFetcher.java
```diff
@@ -150,22 +155,25 @@ class TopicMetadataFetcher implements Closeable {
   }
 
   // acquire the consumer from the pool
-  private PubSubConsumerAdapter acquireConsumer() {
+  PubSubConsumerAdapter acquireConsumer() {
     try {
-      return pubSubConsumerPool.take();
+      consumerWaitListSize.incrementAndGet();
+      long startTime = System.nanoTime();
+      PubSubConsumerAdapter consumerAdapter = pubSubConsumerPool.take();
```
gaojieliu, line 162: A safer approach from my understanding is to dynamically create new consumers if the pool doesn't have enough consumers since `consumer release` can fail.

sushantmane: I agree but this hides bugs in code if any. And since consumers are expensive, due to leaks we might end up a lot of resources which could cause another issues. 

I believe we should not have any leaks after this fix. But even if we do we should be able to detect them via newly added metrics.

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicMetadataFetcher.java
```diff
@@ -150,22 +155,25 @@ class TopicMetadataFetcher implements Closeable {
...
   // release the consumer back to the pool
-  private void releaseConsumer(PubSubConsumerAdapter pubSubConsumerAdapter) {
-    try {
-      pubSubConsumerPool.put(pubSubConsumerAdapter);
-    } catch (InterruptedException e) {
-      Thread.currentThread().interrupt();
-      throw new VeniceException("Interrupted while releasing pubSubConsumerAdapter", e);
+  void releaseConsumer(PubSubConsumerAdapter pubSubConsumerAdapter) {
+    if (!pubSubConsumerPool.offer(pubSubConsumerAdapter)) {
+      LOGGER.error("Failed to release pubSubConsumerAdapter back to the pool.");
```
gaojieliu, line 176: Is it possible that the `offer` can fail? As the total number of consumers is fixed, so in theory, it shouldn't fail.

sushantmane: yeah it shouldn't fail unless there are some logical bugs in the code.

gaojieliu (approved):

sushantmane:
Thanks @lluwm and @gaojieliu for the speedy reviews! 🙏

PR #968, opened 2024-04-29
[admin-tool] Fixed mishandling of get store response in store migration

xunyin8:
## [admin-tool] Fixed mishandling of get store response in store migration
Store migration's end migration command assumes /store returns null for non-existing store. However, it actually throws exception instead and is not handled properly which causes confusion for the user.

## How was this PR tested?
existing unit and integration tests

### DIFF
--- a/clients/venice-admin-tool/src/main/java/com/linkedin/venice/AdminTool.java
+++ b/clients/venice-admin-tool/src/main/java/com/linkedin/venice/AdminTool.java
@@ -73,6 +73,7 @@ import com.linkedin.venice.datarecovery.DataRecoveryClient;
 import com.linkedin.venice.datarecovery.EstimateDataRecoveryTimeCommand;
 import com.linkedin.venice.datarecovery.MonitorCommand;
 import com.linkedin.venice.datarecovery.StoreRepushCommand;
+import com.linkedin.venice.exceptions.ErrorType;
 import com.linkedin.venice.exceptions.VeniceException;
 import com.linkedin.venice.helix.HelixAdapterSerializer;
 import com.linkedin.venice.helix.HelixSchemaAccessor;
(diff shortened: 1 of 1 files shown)


clients/venice-admin-tool/src/main/java/com/linkedin/venice/AdminTool.java
```diff
@@ -2157,7 +2165,13 @@ private static void endMigration(CommandLine cmd) {
...
+      StoreResponse childSrcStoreResponse = entry.getValue().getStore(storeName);
+      if (childSrcStoreResponse.isError() && childSrcStoreResponse.getErrorType() != ErrorType.STORE_NOT_FOUND) {
+        System.err.println(
+            "ERROR: failed to check store " + storeName + " existence in original cluster " + srcClusterName
+                + " in fabric " + entry.getKey() + " due to error: " + childSrcStoreResponse.getError());
+      }
+      if (childSrcStoreResponse.getErrorType() != ErrorType.STORE_NOT_FOUND) {
         System.err.println(
             "ERROR: store " + storeName + " still exists in source cluster " + srcClusterName + " in fabric "
                 + entry.getKey() + ". Please try again later.");
```
huangminchn, line 2177: I understand this is out of the scope of this PR - if something goes wrong, we shouldn't continue and reset the migration flag, right?

huangminchn (approved):
Thanks Xun!

clients/venice-admin-tool/src/main/java/com/linkedin/venice/AdminTool.java
xunyin8: No, if something goes wrong here we actually don't want to reset the migration flag for the reasons below:
1. Before calling end migration we should already verified that the most if not all traffic is shifted to destination cluster and things are working. Continue to operate this way is safer.
2. We already initiated the deletion in src cluster and the store might be in a partially deleted state and shifting the traffic back is a terrible idea.

PR #969, opened 2024-04-29
[vpj] Do not mark overall status COMPLETE till all region finishes

majisourav99:
##  Do not mark overall status COMPLETE till all region finishes

The overall status should not be updated to COMPLETE till all the region are pushed when target region push is enabled.

## How was this PR tested?

GHCI


huangminchn (approved):
Thanks Sourav! Do we have test case for it?

majisourav99:
> Thanks Sourav! Do we have test case for it?

No, its bit hard to test the intermediate state of the status between the 2 pushes.

PR #970, opened 2024-04-30
[router] Added config to enable DNS resolution before SSL

huangminchn:
## Summary
Config to enable this feature:
router.resolve.before.ssl

When this config is enabled, "SslInitializer#enableSslTaskExecutor" will not be called, and the SSL handshake thread pool count will be used to construct the DNS resolution thread pool.

Besides, added two new SSL related metrics using the API from Netty: pending_ssl_handshake_count
total_failed_ssl_handshake_count

## How was this PR tested?
GHCI

### DIFF
--- a/services/venice-router/src/main/java/com/linkedin/venice/router/RouterServer.java
+++ b/services/venice-router/src/main/java/com/linkedin/venice/router/RouterServer.java
@@ -808,6 +808,9 @@ public class RouterServer extends AbstractVeniceService {
     storageNodeClient.close();
     workerEventLoopGroup.shutdownGracefully();
     serverEventLoopGroup.shutdownGracefully();
+    if (sslResolverEventLoopGroup != null) {
+      sslResolverEventLoopGroup.shutdownGracefully();
+    }
 
     dispatcher.stop();
 
--- a/services/venice-router/src/main/java/com/linkedin/venice/router/RouterServer.java
+++ b/services/venice-router/src/main/java/com/linkedin/venice/router/RouterServer.java
@@ -168,6 +168,7 @@ public class RouterServer extends AbstractVeniceService {
 
   private MultithreadEventLoopGroup workerEventLoopGroup;
   private MultithreadEventLoopGroup serverEventLoopGroup;
+  private MultithreadEventLoopGroup sslResolverEventLoopGroup;
 
   private ExecutorService workerExecutor;
   private EventThrottler routerEarlyThrottler;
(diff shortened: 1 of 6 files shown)


services/venice-router/src/main/java/com/linkedin/venice/router/stats/SecurityStats.java
```diff
@@ -49,4 +51,16 @@ public void recordNonSslRequest() {
   private void recordLiveConnectionCount() {
     this.sslLiveConnectionCount.record(secureConnectionCountSupplier.getAsInt());
   }
+
+  public void registerSslHandshakeSensors(SslInitializer sslInitializer) {
+    registerSensor(
+        new AsyncGauge(
+            (ignored1, ignored2) -> sslInitializer.getHandshakesStarted()
+                - (sslInitializer.getHandshakesSuccessful() + sslInitializer.getHandshakesFailed()),
```
xunyin8, line 59: Why are we not using `SslInitializer#getPendingHandshakes()`? Is it not reliable?

huangminchn: Right, according to Antony, this call is expensive.

xunyin8 (approved):

services/venice-router/src/main/java/com/linkedin/venice/router/RouterServer.java
```diff
@@ -639,15 +642,37 @@ public AsyncFuture<LongSupplier> getLongTailRetryMilliseconds(
...
       sslInitializer = new SslInitializer(SslUtils.toAlpiniSSLFactory(sslFactory.get()), false);
       if (config.getClientSslHandshakeThreads() > 0) {
-        ThreadPoolExecutor sslHandshakeExecutor = ThreadPoolFactory.createThreadPool(
-            config.getClientSslHandshakeThreads(),
-            "SSLHandShakeThread",
-            config.getClientSslHandshakeQueueCapacity(),
-            LINKED_BLOCKING_QUEUE);
-        new ThreadPoolStats(metricsRepository, sslHandshakeExecutor, "ssl_handshake_thread_pool");
-        sslInitializer.enableSslTaskExecutor(sslHandshakeExecutor);
+        if (!config.isResolveBeforeSSL()) {
```
nisargthakkar, line 645: Can we flip the order of `if`s? `if (config.isResolveBeforeSSL()) {...} else {...}`

nisargthakkar (approved):
I only have a nit comment. Feel free to drop it. Change LGTM

services/venice-router/src/main/java/com/linkedin/venice/router/RouterServer.java
huangminchn: Yeah, that's easier to read. Updated!

nisargthakkar (approved):

xunyin8 (approved):

PR #971, opened 2024-04-30
[vpj] Fix building or fetching dictionary for repush jobs

nisargthakkar:
## Fix building or fetching dictionary for repush jobs

Recently, in #886, we refactored some code around compression metric collection and using mapper to build dictionary. This refactoring had a bug where the dictionary building for repush jobs was skipped and if VPJ sends control messages directly, then the push would fail, as the `START_OF_PUSH` would not have any dictionary. This change fixes the issue and reenables a dictionary to be built for KIF repush as well.

## How was this PR tested?

Added an integration test. GH CI will do regression testing

### DIFF
--- a/internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestBatch.java
+++ b/internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestBatch.java
@@ -627,6 +627,37 @@ public abstract class TestBatch {
   }
 
   @Test(timeOut = TEST_TIMEOUT)
+  public void testKafkaInputBatchJobWithZstdCompression() throws Exception {
+    VPJValidator validator = (avroClient, vsonClient, metricsRepository) -> {
+      // test single get
+      for (int i = 1; i <= 100; i++) {
+        Assert.assertEquals(avroClient.get(Integer.toString(i)).get().toString(), "test_name_" + i);
+      }
+    };
+    String storeName = testBatchStore(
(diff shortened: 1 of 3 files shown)


internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestBatch.java
```diff
@@ -626,6 +626,37 @@ public void testKafkaInputBatchJob() throws Exception {
...
+    VPJValidator validator = (avroClient, vsonClient, metricsRepository) -> {
+      // test single get
+      for (int i = 1; i <= 100; i++) {
+        Assert.assertEquals(avroClient.get(Integer.toString(i)).get().toString(), "test_name_" + i);
+      }
+    };
+    String storeName = testBatchStore(
+        inputDir -> new KeyAndValueSchemas(writeSimpleAvroFileWithStringToStringSchema(inputDir)),
+        properties -> {
+          properties.setProperty(SEND_CONTROL_MESSAGES_DIRECTLY, String.valueOf(true));
```
m-nagarajan, line 640: can we test both True/False for this case?

m-nagarajan (commented):
Thanks for the quick fix. Just left one minor comment.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestBatch.java
majisourav99: Right, we should test with compression metric/KIF true-false combination.

PR #972, opened 2024-04-30
[vpj] Skip closing the SparkSession and let it be closed at app shutdown

nisargthakkar:
## Skip closing the SparkSession and let it be closed at app shutdown

In `AbstractDataWriterSparkJob`, we need a `SparkSession` and to get one, we use `SparkSession.Builder()...getOrCreate()`. This either provides an existing `SparkSession`, or creates a new one if one doesn't exist already. `SparkContext` has a `ShutdownHook` that closes the session at the end of the JVM. When VPJ is executed via `spark-submit`, it executes in the Spark driver. In this case, we do not wish to close the `SparkSession` and instead let the `ShutdownHook` take care of it. This is also okay for client-mode of Spark execution as those can be expected to be short lived. If someone needs a way to explicitly close a `SparkSession`, we can add a new function that helps achieve that. We don't have such a need right now.

## How was this PR tested?

GH CI. Tested manually as well


m-nagarajan (approved):
LGTM thanks

PR #973, opened 2024-05-01
[server] Make separate drainer for batch and hybrid store ingestions

majisourav99:
## Make separate drainer for batch and hybrid store ingestions

Make separate drainer for batch and hybrid store ingestions so that batch store ingestions does not affect hybrid store ingestions.

## How was this PR tested?

integ test

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/SeparatedStoreBufferService.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/SeparatedStoreBufferService.java
@@ -3,13 +3,9 @@ package com.linkedin.davinci.kafka.consumer;
 import com.linkedin.davinci.config.VeniceServerConfig;
 import com.linkedin.venice.kafka.protocol.KafkaMessageEnvelope;
 import com.linkedin.venice.message.KafkaKey;
-import com.linkedin.venice.pubsub.PubSubTopicPartitionImpl;
 import com.linkedin.venice.pubsub.api.PubSubMessage;
-import com.linkedin.venice.pubsub.api.PubSubTopic;
 import com.linkedin.venice.pubsub.api.PubSubTopicPartition;
-import com.linkedin.venice.utils.concurrent.VeniceConcurrentHashMap;
 import io.tehuti.metrics.MetricsRepository;
-import java.util.Map;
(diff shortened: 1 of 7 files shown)


gaojieliu (approved):
Thanks for the quick fix!

lluwm:
Can we name unsortedService and sortedService differently? Really hard to debug otherwise this PR or a later one is fine.

sixpluszero (approved):

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/SeparatedStoreBufferService.java
```diff
@@ -22,11 +18,10 @@
  */
 public class SeparatedStoreBufferService extends AbstractStoreBufferService {
   private static final Logger LOGGER = LogManager.getLogger(SeparatedStoreBufferService.class);
-  protected final StoreBufferService sortedServiceDelegate;
-  protected final StoreBufferService unsortedServiceDelegate;
+  protected final StoreBufferService sortedStoreBufferServiceDelegate;
+  protected final StoreBufferService unsortedStoreBufferServiceDelegate;
```
sixpluszero, line 22: I think everything else looks good to me, but could you for the debugging purpose consider adding a new constructor with prefix and when creating writers use the prefix for this separated drainer case? Otherwise in the log it is hard to debug.
This should be just a simple constructor change for StoreBufferService and one extra variable for prefix default to ""

sixpluszero (approved):

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/SeparatedStoreBufferService.java
lluwm: I think this is a must have before enabling the feature again, otherwise, there is no way to operate.

sixpluszero: I think it is already addressed in the code. There is suffix in the writer thread.

lluwm: then that's great

PR #974, opened 2024-05-01
[controller] Support dangling topic check and deletion in topic cleanup service.

haoxu07:

Add dangling topic checking functionality,  `TopicCleanupService` will compare topics listed from source of truth pub sub system and topics from other type of pub sub client, if there is topic could not be found by source of truth pub sub system client but can be founded by other type of pub sub client, topic will be deleted if met by times more than `CONTROLLER_DANGLING_TOPIC_OCCURRENCE_THRESHOLD_FOR_CLEANUP` by such cases:

1.  Real-time topic, if store is not hybrid and no store version carries hybrid config,  that topic is not valid. 
2. Version topic, if store does not contain that version, the topic is not valid. 
3. If there is no store available for that topic, the topic is not valid. 

Add a config for source of truth admin client class: `PUB_SUB_SOURCE_OF_TRUTH_ADMIN_ADAPTER_FACTORY_CLASS`
Add a config for consistently decide a topic is dangling: `CONTROLLER_DANGLING_TOPIC_OCCURRENCE_THRESHOLD_FOR_CLEANUP`
Add a new config to enable this feature: `CONTROLLER_DANGLING_TOPIC_CLEAN_UP_INTERVAL_SECOND`. If its value is >0 and `PUB_SUB_SOURCE_OF_TRUTH_ADMIN_ADAPTER_FACTORY_CLASS` is not the same as `PUB_SUB_ADMIN_ADAPTER_FACTORY_CLASS`,  the feature will be turned on. Otherwise, it will be disabled. 

## How was this PR tested?

CI, integration test will be in internal repo with other type of pub sub system. 

### DIFF
--- a/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerConfig.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerConfig.java
@@ -33,6 +33,7 @@ import static com.linkedin.venice.ConfigKeys.CONTROLLER_CLUSTER;
 import static com.linkedin.venice.ConfigKeys.CONTROLLER_CLUSTER_LEADER_HAAS;
 import static com.linkedin.venice.ConfigKeys.CONTROLLER_CLUSTER_REPLICA;
 import static com.linkedin.venice.ConfigKeys.CONTROLLER_CLUSTER_ZK_ADDRESSS;
+import static com.linkedin.venice.ConfigKeys.CONTROLLER_DANGLING_TOPIC_CLEAN_UP_INTERVAL_SECOND;
 import static com.linkedin.venice.ConfigKeys.CONTROLLER_DISABLED_REPLICA_ENABLER_INTERVAL_MS;
 import static com.linkedin.venice.ConfigKeys.CONTROLLER_DISABLED_ROUTES;
 import static com.linkedin.venice.ConfigKeys.CONTROLLER_DISABLE_PARENT_TOPIC_TRUNCATION_UPON_COMPLETION;
@@ -328,6 +329,8 @@ public class VeniceControllerConfig extends VeniceControllerClusterConfig {
   private final boolean useDaVinciSpecificExecutionStatusForError;
   private final PubSubClientsFactory pubSubClientsFactory;
 
+  private final long danglingTopicCleanupIntervalSeconds;
+
   public VeniceControllerConfig(VeniceProperties props) {
     super(props);
     this.adminPort = props.getInt(ADMIN_PORT);
@@ -572,6 +575,7 @@ public class VeniceControllerConfig extends VeniceControllerClusterConfig {
     this.useDaVinciSpecificExecutionStatusForError =
         props.getBoolean(USE_DA_VINCI_SPECIFIC_EXECUTION_STATUS_FOR_ERROR, false);
     this.pubSubClientsFactory = new PubSubClientsFactory(props);
+    this.danglingTopicCleanupIntervalSeconds = props.getLong(CONTROLLER_DANGLING_TOPIC_CLEAN_UP_INTERVAL_SECOND, 3600);
   }
 
   private void validateActiveActiveConfigs() {
@@ -1151,6 +1155,10 @@ public class VeniceControllerConfig extends VeniceControllerClusterConfig {
     }
 
     return outputMap;
+  }
+
+  public long getDanglingTopicCleanupIntervalSeconds() {
+    return danglingTopicCleanupIntervalSeconds;
   }
 
   /**
--- a/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerConfig.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerConfig.java
@@ -112,6 +112,7 @@ import com.linkedin.venice.authorization.DefaultIdentityParser;
 import com.linkedin.venice.client.store.ClientConfig;
 import com.linkedin.venice.controllerapi.ControllerRoute;
 import com.linkedin.venice.exceptions.VeniceException;
+import com.linkedin.venice.pubsub.PubSubAdminAdapterFactory;
 import com.linkedin.venice.pubsub.PubSubClientsFactory;
 import com.linkedin.venice.pubsub.adapter.kafka.admin.ApacheKafkaAdminAdapter;
 import com.linkedin.venice.status.BatchJobHeartbeatConfigs;
@@ -328,6 +329,8 @@ public class VeniceControllerConfig extends VeniceControllerClusterConfig {
   private final int minSchemaCountToKeep;
   private final boolean useDaVinciSpecificExecutionStatusForError;
   private final PubSubClientsFactory pubSubClientsFactory;
+
+  private final PubSubAdminAdapterFactory sourceOfTruthAdminAdapterFactory;
 
   private final long danglingTopicCleanupIntervalSeconds;
 
@@ -575,6 +578,7 @@ public class VeniceControllerConfig extends VeniceControllerClusterConfig {
     this.useDaVinciSpecificExecutionStatusForError =
         props.getBoolean(USE_DA_VINCI_SPECIFIC_EXECUTION_STATUS_FOR_ERROR, false);
     this.pubSubClientsFactory = new PubSubClientsFactory(props);
+    this.sourceOfTruthAdminAdapterFactory = PubSubClientsFactory.createSourceOfTruthAdminFactory(props);
     this.danglingTopicCleanupIntervalSeconds = props.getLong(CONTROLLER_DANGLING_TOPIC_CLEAN_UP_INTERVAL_SECOND, 3600);
   }
 
@@ -1062,6 +1066,10 @@ public class VeniceControllerConfig extends VeniceControllerClusterConfig {
 
   public PubSubClientsFactory getPubSubClientsFactory() {
     return pubSubClientsFactory;
+  }
+
+  public PubSubAdminAdapterFactory getSourceOfTruthAdminAdapterFactory() {
+    return sourceOfTruthAdminAdapterFactory;
   }
 
   /**
--- a/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerConfig.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerConfig.java
@@ -34,6 +34,7 @@ import static com.linkedin.venice.ConfigKeys.CONTROLLER_CLUSTER_LEADER_HAAS;
 import static com.linkedin.venice.ConfigKeys.CONTROLLER_CLUSTER_REPLICA;
 import static com.linkedin.venice.ConfigKeys.CONTROLLER_CLUSTER_ZK_ADDRESSS;
 import static com.linkedin.venice.ConfigKeys.CONTROLLER_DANGLING_TOPIC_CLEAN_UP_INTERVAL_SECOND;
+import static com.linkedin.venice.ConfigKeys.CONTROLLER_DANGLING_TOPIC_OCCURRENCE_THRESHOLD_FOR_CLEANUP;
 import static com.linkedin.venice.ConfigKeys.CONTROLLER_DISABLED_REPLICA_ENABLER_INTERVAL_MS;
 import static com.linkedin.venice.ConfigKeys.CONTROLLER_DISABLED_ROUTES;
 import static com.linkedin.venice.ConfigKeys.CONTROLLER_DISABLE_PARENT_TOPIC_TRUNCATION_UPON_COMPLETION;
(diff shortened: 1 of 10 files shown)


services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/TopicCleanupService.java
```diff
@@ -401,4 +446,40 @@ public static List<PubSubTopic> extractVersionTopicsToCleanup(
...
+            if (!store.isHybrid()) {
+              LOGGER.info("Will remove real-time dangling topic {}", pubSubTopic);
+              topicsTobeCleanup.add(pubSubTopic);
+            }
+          } else if (pubSubTopic.isVersionTopicOrStreamReprocessingTopic() || pubSubTopic.isViewTopic()) {
+            int versionNum = Version.parseVersionFromKafkaTopicName(pubSubTopic.getName());
+            if (!store.containsVersion(versionNum)) {
+              LOGGER.info("Will remove dangling version topic {}", pubSubTopic);
+              topicsTobeCleanup.add(pubSubTopic);
+            }
```
lluwm, line 471: We submitted a change recently to preserve VT data when DIV fatal issue happens. Can we double check that it won't be considered as dandling topics thus got deleted here?

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/TopicCleanupService.java
```diff
@@ -74,13 +82,17 @@ public class TopicCleanupService extends AbstractVeniceService {
   private boolean isRTTopicDeletionBlocked = false;
   private boolean isLeaderControllerOfControllerCluster = false;
   private long refreshQueueCycle = Time.MS_PER_MINUTE;
+  private Optional<PubSubAdminAdapter> apacheKafkaAdminAdapter;
```
adamxchen, line 85: Is there any benefit of using `Optional<>`?  If there's not much value, we can use `null` if it's absent.

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/TopicCleanupService.java
```diff
@@ -248,6 +282,17 @@ private void populateDeprecatedTopicQueue(PriorityQueue<PubSubTopic> topics) {
     if (realTimeTopicDeletionNeeded.get() && !multiDataCenterStoreToVersionTopicCount.isEmpty()) {
       refreshMultiDataCenterStoreToVersionTopicCountMap(topicsWithRetention.keySet());
     }
+
+    // Check if there is dangling topics to be deleted.
+    if (apacheKafkaAdminAdapter.isPresent()
+        && System.currentTimeMillis() - danglingTopicCleanupIntervalMs > recentDanglingTopicCleanupTime) {
+      List<PubSubTopic> pubSubTopics = collectDanglingTopics(topicsWithRetention);
+      if (!pubSubTopics.isEmpty()) {
+        LOGGER.info("Find topic discrepancy: topics not in kafka but in xinfra {}", pubSubTopics);
```
adamxchen, line 291: we should not expose the proprietary pubsub system name in OSS. And IMO it's more intuitive to say 
```
LOGGER.warn("Find topic discrepancy: {} is not present among all pubsub systems"
```

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/TopicCleanupService.java
```diff
@@ -401,4 +446,40 @@ public static List<PubSubTopic> extractVersionTopicsToCleanup(
...
+    List<PubSubTopic> topicsTobeCleanup = new ArrayList<>();
+    Set<PubSubTopic> kafkaTopics = apacheKafkaAdminAdapter.get().listAllTopics();
+    for (Map.Entry<PubSubTopic, Long> entry: pubSubTopicsRetentions.entrySet()) {
+      PubSubTopic pubSubTopic = entry.getKey();
+      if (!kafkaTopics.contains(pubSubTopic)) {
+        try {
+          String storeName = pubSubTopic.getStoreName();
+          String clusterDiscovered = admin.discoverCluster(storeName).getFirst();
+          Store store = admin.getStore(clusterDiscovered, storeName);
+          LOGGER.info("Find topic discrepancy case: {}", pubSubTopic);
```
adamxchen, line 460: I'd suggest for these bad but not the end-of-world case, we use WARN. It's not good to have discrepancy among different pubsub ystems.

services/venice-controller/src/test/java/com/linkedin/venice/controller/kafka/TestTopicCleanupService.java
```diff
@@ -217,10 +248,12 @@ public void testCleanupVeniceTopics() throws ExecutionException {
     verify(topicManager, never()).ensureTopicIsDeletedAndBlockWithRetry(getPubSubTopic(storeName2, "_rt"));
     // Delete should be blocked by remote VT
     verify(topicManager, never()).ensureTopicIsDeletedAndBlockWithRetry(getPubSubTopic(storeName3, "_rt"));
-    verify(topicCleanupServiceStats, atLeastOnce()).recordDeletableTopicsCount(5);
+    verify(topicCleanupServiceStats, atLeastOnce()).recordDeletableTopicsCount(6);
```
adamxchen, line 251: Does the existing test case has covered all 3 cases?
```
1. Real-time topic, if store is not hybrid, that topic will be deleted.
2. Version topic, if store does not contain that version, the topic will be deleted.
3. If there is no store available for that topic, the topic will be deleted.
```

sushantmane:
> * Real-time topic, if store is not hybrid, that topic will be deleted.
> * Version topic, if store does not contain that version, the topic will be deleted.
> * If there is no store available for that topic, the topic will be deleted.

So sorry for jumping in so late.

Real-time topic, if store is not hybrid, that topic will be deleted.
> It's possible that store level setting is off but one of the versions is still hybtid. (This can happen during hybrid to batch conversion). Are we handling this case?



If there is no store available for that topic, the topic will be deleted.
> We also have admin topic in one cluster.   We need to exclude it as well

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/TopicCleanupService.java
```diff
@@ -401,4 +446,40 @@ public static List<PubSubTopic> extractVersionTopicsToCleanup(
...
         .collect(Collectors.toList());
   }
+
+  private List<PubSubTopic> collectDanglingTopics(Map<PubSubTopic, Long> pubSubTopicsRetentions) {
+    List<PubSubTopic> topicsTobeCleanup = new ArrayList<>();
+    Set<PubSubTopic> kafkaTopics = apacheKafkaAdminAdapter.get().listAllTopics();
+    for (Map.Entry<PubSubTopic, Long> entry: pubSubTopicsRetentions.entrySet()) {
+      PubSubTopic pubSubTopic = entry.getKey();
+      if (!kafkaTopics.contains(pubSubTopic)) {
+        try {
```
sushantmane, line 456: Can we split this method into two private methods?
1) isStillRelevantRt/isStillValidRt
2) isStillRelevantVt/isStillValidVt

sushantmane (changes requested):

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/TopicCleanupService.java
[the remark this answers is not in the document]
haoxu07: Yeah, the expectation here for kafka admin adapter is to return non-dangling topics. So there are 2 types of admin adapter: 1) Configured one for entire controller (will talk to pub sub frontend to get topic configurations) 2) Admin client for getting all truly existed topics. Kafka admin adapter is 2nd one.

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/TopicCleanupService.java
haoxu07: Sure, that will be easier to understand.

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/TopicCleanupService.java
haoxu07: Yeah, you are right. I will remove that.

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/TopicCleanupService.java
```diff
@@ -401,4 +446,40 @@ public static List<PubSubTopic> extractVersionTopicsToCleanup(
...
+          if (e instanceof VeniceNoStoreException) {
+            LOGGER.info("No store is found for topic: {}", pubSubTopic);
+            topicsTobeCleanup.add(pubSubTopic);
+          } else {
+            LOGGER.error("Error happened during checking dangling topic: {}", pubSubTopic, e);
+          }
+        }
+      }
+    }
+    return topicsTobeCleanup;
```
sushantmane, line 483: can we keep just `topicsToCleanup`?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerConfig.java
```diff
@@ -572,6 +578,8 @@ public VeniceControllerConfig(VeniceProperties props) {
     this.useDaVinciSpecificExecutionStatusForError =
         props.getBoolean(USE_DA_VINCI_SPECIFIC_EXECUTION_STATUS_FOR_ERROR, false);
     this.pubSubClientsFactory = new PubSubClientsFactory(props);
+    this.sourceOfTruthAdminAdapterFactory = PubSubClientsFactory.createSourceOfTruthAdminFactory(props);
+    this.danglingTopicCleanupIntervalSeconds = props.getLong(CONTROLLER_DANGLING_TOPIC_CLEAN_UP_INTERVAL_SECOND, 3600);
```
sushantmane, line 582: Can we disable it by default in code but enable it in integration tests?

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/TopicCleanupService.java
haoxu07: Here dangling topic deletion only focus on the topics gap inside pub sub system component, as long as we did not try to delete the existing topics, so that topic will not be evaluated by this feature. Moreover, this feature only focus on the dangling topics are not valid based on the criteria(not vaild version, not valid real-time topic ....) anymore.

sushantmane (commented):
Thanks a lot, @haoxu07! Overall LGTM, Just two minor comments

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/TopicCleanupService.java
```diff
@@ -401,4 +444,60 @@ public static List<PubSubTopic> extractVersionTopicsToCleanup(
...
+  private List<PubSubTopic> collectDanglingTopics(Map<PubSubTopic, Long> pubSubTopicsRetentions) {
+    List<PubSubTopic> topicsToCleanup = new ArrayList<>();
+    Set<PubSubTopic> kafkaTopics = sourceOfTruthPubSubAdminAdapter.listAllTopics();
+    for (Map.Entry<PubSubTopic, Long> entry: pubSubTopicsRetentions.entrySet()) {
+      PubSubTopic pubSubTopic = entry.getKey();
+      if (!kafkaTopics.contains(pubSubTopic)) {
+        try {
+          String storeName = pubSubTopic.getStoreName();
+          if (PubSubTopicType.isAdminTopic(pubSubTopic.getName())) {
+            LOGGER.error("Skip dangling admin topic {}.", pubSubTopic);
```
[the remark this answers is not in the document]
haoxu07: This dangling topic feature is not expected to be triggered frequently, based on the config setting, my current estimation is once per hour. So there should not be that many log lines.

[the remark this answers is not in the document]
sushantmane: But why do we even need to log this line?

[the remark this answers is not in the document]
haoxu07: My point is:  just logging here even the case is very rare and if it happens it means we have a big trouble.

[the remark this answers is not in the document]
sushantmane: Got it

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/TopicCleanupService.java
```diff
@@ -471,7 +471,10 @@ private List<PubSubTopic> collectDanglingTopics(Map<PubSubTopic, Long> pubSubTop
           Store store = admin.getStore(clusterDiscovered, storeName);
           LOGGER.warn("Find topic discrepancy case: {}", pubSubTopic);
           if (!isStillValidRealtimeTopic(pubSubTopic, store) || !isStillValidVersionTopic(pubSubTopic, store)) {
-            topicsToCleanup.add(pubSubTopic);
+            if (checkIfDanglingTopicConsistentlyFound(pubSubTopic)) {
+              LOGGER.warn("Will remove consistently found dangling topic {}.", pubSubTopic);
```
sushantmane, line 475: actually we could move remove here

sushantmane (approved):
Thanks a lot, @haoxu07! This PR will help us reduce toil.  🚀

PR #975, opened 2024-05-06
[fast-client][server] Throw 403 for metadata request that does not have storage read quota enabled

xunyin8:
## [fast-client][server] Throw 403 for metadata request that does not have storage read quota enabled
1. The idea is to block new fast-client traffic that do not have storage node read quota enabled. New client version will fail fast and have a clear error message to direct the user on what's missing. Old client versions will keep failing the metadata request and the user will also see the error message once dig deeper into the error trace.

## How was this PR tested?
Unit test and integration test

### DIFF
--- a/services/venice-server/src/main/java/com/linkedin/venice/listener/ServerReadMetadataRepository.java
+++ b/services/venice-server/src/main/java/com/linkedin/venice/listener/ServerReadMetadataRepository.java
@@ -66,6 +66,11 @@ public class ServerReadMetadataRepository implements ReadMetadataRetriever {
     MetadataResponse response = new MetadataResponse();
     try {
       Store store = storeRepository.getStoreOrThrow(storeName);
+      if (!store.isStorageNodeReadQuotaEnabled()) {
+        throw new UnsupportedOperationException(
+            String
+                .format("Fast client is not enabled for store: %s, please contact Venice team for support", storeName));
+      }
       // Version metadata
       int currentVersionNumber = store.getCurrentVersion();
       if (currentVersionNumber == Store.NON_EXISTING_VERSION) {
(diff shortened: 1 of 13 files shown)


services/venice-server/src/main/java/com/linkedin/venice/listener/ServerReadMetadataRepository.java
```diff
@@ -66,6 +66,11 @@ public MetadataResponse getMetadata(String storeName) {
     MetadataResponse response = new MetadataResponse();
     try {
       Store store = storeRepository.getStoreOrThrow(storeName);
+      if (!store.isStorageNodeReadQuotaEnabled()) {
+        throw new UnsupportedOperationException(
+            String
+                .format("Fast client is not enabled for store: %s, please contact Venice team for support", storeName));
```
adamxchen, line 72: Can you remove or rephrase the second part "please contact Venice team for support"? As a open-source project, I think we should avoid using `venice team` in the codes/exception msgs. How about  "Fast client is not enabled for store: %s, please contact the service team to enable it"?

adamxchen (approved):
Looks good to me overall. Just have one nit. And for my curiosity, have we ensured all fast client would have storage read quota enabled? Like, when they onboard, will this config be turned on automatically?

services/venice-server/src/main/java/com/linkedin/venice/listener/ServerReadMetadataRepository.java
xunyin8: Good call, let me change that

xunyin8:
> Looks good to me overall. Just have one nit. And for my curiosity, have we ensured all fast client would have storage read quota enabled? Like, when they onboard, will this config be turned on automatically?

This is done manually for now but later it can and will be automated/self serve.

adamxchen (approved):
Thank you!

PR #976, opened 2024-05-07
[WIP] Utilize temp cache and skip syncOffset during CATCHING_UP phase

zx0401:
[changelog][BootstrappingVeniceChangelogConsumer] Utilize temp cache and skip syncOffset during CATCHING_UP phase to improve bootstrapping performance

- Skip syncOffset during CATCHING_UP phase
- During CATCHING_UP phase, we'll save raw data in a temporary hash map
- On completion of CATCHING_UP, we'll process records in temporary hash map and save in storage

## How was this PR tested?

- unit test

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/InternalLocalBootstrappingVeniceChangelogConsumer.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/InternalLocalBootstrappingVeniceChangelogConsumer.java
@@ -50,6 +50,7 @@ import java.nio.ByteBuffer;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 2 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/InternalLocalBootstrappingVeniceChangelogConsumer.java
```diff
@@ -372,14 +384,51 @@ protected <T> T processRecordBytes(
       PubSubTopicPartition partition,
       int readerSchemaId,
       long recordOffset) {
+    BootstrapState bootstrapState = bootstrapStateMap.get(partition.getPartitionNumber());
+    // In CATCHING_UP phase, there can be lots of updates to the same key.
+    // To improve bootstrapping performance, we'll put raw data in a temp cache (hashmap) and will
+    // process save to underlying storage on CATCHING_UP completion. Besides, we'll also skip RocksDB syncOffset.
+    if (bootstrapState.bootstrapState == PollState.CATCHING_UP) {
+      bootstrapState
```
ZacAttack, line 392: idk about this.  It seems like the cache could grow to a pretty large size.  We should probably put a size safety check in here to control flushing to disk.

PR #979, opened 2024-05-09
[vpj] Allow inc push when target region is enabled

majisourav99:
## Allow inc push when target region is enabled

Inc push is not supported for target region push. Skip target region push for inc push stores.

## How was this PR tested?

GHCI

### DIFF
--- a/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
+++ b/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
@@ -389,7 +389,8 @@ public class VenicePushJob implements AutoCloseable {
     pushJobSettingToReturn.isTargetedRegionPushEnabled = props.getBoolean(TARGETED_REGION_PUSH_ENABLED, false);
     pushJobSettingToReturn.isSystemSchemaReaderEnabled = props.getBoolean(SYSTEM_SCHEMA_READER_ENABLED, false);
     if (pushJobSettingToReturn.isIncrementalPush && pushJobSettingToReturn.isTargetedRegionPushEnabled) {
-      throw new VeniceException("Incremental push is not supported while using targeted region push mode");
+      LOGGER.warn("Incremental push is not supported while using targeted region push mode");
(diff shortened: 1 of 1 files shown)


clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
```diff
@@ -389,7 +389,8 @@ private PushJobSetting getPushJobSetting(VeniceProperties props) {
     pushJobSettingToReturn.isTargetedRegionPushEnabled = props.getBoolean(TARGETED_REGION_PUSH_ENABLED, false);
     pushJobSettingToReturn.isSystemSchemaReaderEnabled = props.getBoolean(SYSTEM_SCHEMA_READER_ENABLED, false);
     if (pushJobSettingToReturn.isIncrementalPush && pushJobSettingToReturn.isTargetedRegionPushEnabled) {
-      throw new VeniceException("Incremental push is not supported while using targeted region push mode");
+      LOGGER.warn("Incremental push is not supported while using targeted region push mode");
+      pushJobSettingToReturn.isTargetedRegionPushEnabled = false;
     }
     if (props.containsKey(TARGETED_REGION_PUSH_LIST)) {
       if (pushJobSettingToReturn.isTargetedRegionPushEnabled) {
```
nisargthakkar, line 396: We should remove this as well in that case. If a user specifies targeted region push enabled, they might also specify the targeted region push list

nisargthakkar: Github doesn't do a good job of showing the exact line. I was referring to this section:
```
    if (props.containsKey(TARGETED_REGION_PUSH_LIST)) {
      if (pushJobSettingToReturn.isTargetedRegionPushEnabled) {
        pushJobSettingToReturn.targetedRegions = props.getString(TARGETED_REGION_PUSH_LIST);
      } else {
        throw new VeniceException("Targeted region push list is only supported when targeted region push is enabled");
      }
    }
```

PR #980, opened 2024-05-10
[dvc] dvc node blob discovery

ishwarya-citro:
## dvc node blob discovery

Implemented a new API endpoint designed to facilitate the bootstrapping of nodes by identifying and listing live nodes prepared to serve the store's blobs. If the blob transfer feature is enabled for the store, it retrieves each node's hostname from the push store, given that these nodes are alive.

## How was this PR tested?
integ. test cases


clients/da-vinci-client/src/main/java/com/linkedin/davinci/repository/NativeMetadataRepository.java
```diff
@@ -580,4 +585,45 @@ public String getVeniceCluster(String storeName) {
...
+   * Retrieves the host names for live DVC nodes ready to transfer blobs based on the specified store settings.
+   * Only returns host names from nodes that have completed a full push and are active.
+   *
+   * @param storeName   the name of the store
+   * @param version     the version of the store
+   * @param partitionID the partition ID of store's partition
+   * @return a list of host names for live DVC nodes; returns an empty list if no live nodes are found or if conditions are not met
+   */
+  @Override
+  public List<String> getLiveNodeHostNamesForTransfer(String storeName, int version, int partitionID) {
```
adamxchen, line 599: I think your changes overall are heading to the right directions but it seems like they are not at the right places. In fact, you are adding the API to **Venice server** instead of routers.
IIRC, the expected workflow is:
1. When DVC clients start up, it will subscribe to some partitions
2. Upon the subscription, it sends requests to Venice router to check which node has blobs
3. Once venice router returns the response, the DVC client would send another request to that DVC node to start the blob transfer.

In this PR, we are trying to achieve #2, so this method along with many changes below should be on Venice router (they are much simpler if we move them onto venice router side IMO)

adamxchen (changes requested):

internal/venice-common/src/main/java/com/linkedin/venice/controllerapi/BlobDiscoveryResponse.java
```diff
@@ -0,0 +1,17 @@
+package com.linkedin.venice.controllerapi;
+
+import java.util.List;
+
+
+public class BlobDiscoveryResponse extends ControllerResponse {
+  private boolean isError;
```
adamxchen, line 7: We need getter/setter for `isError` too.

internal/venice-common/src/main/java/com/linkedin/venice/controllerapi/ControllerApiConstants.java
```diff
@@ -8,6 +8,7 @@ public class ControllerApiConstants {
   public static final String BATCH_JOB_HEARTBEAT_ENABLED = "batch_job_heartbeat_enabled";
 
   public static final String NAME = "store_name";
+  public static final String STORE_PARTITION = "partition";
```
adamxchen, line 11: can we call it `store_partition` so the value matches the variable name?

services/venice-router/src/main/java/com/linkedin/venice/router/MetaDataHandler.java
```diff
@@ -477,6 +491,66 @@ private void handleCurrentVersionLookup(ChannelHandlerContext ctx, VenicePathPar
...
+
+    if (StringUtils.isEmpty(storeName) || StringUtils.isEmpty(storeVersion) || StringUtils.isEmpty(storePartition)) {
+      byte[] errBody = ("Blob Discovery: missing storeName, storeVersion, or storePartition " + storeName + storeVersion
+          + storePartition).getBytes();
+      setupResponseAndFlush(BAD_REQUEST, errBody, false, ctx);
+      return;
+    }
+
+    Store store = storeRepository.getStore(storeName);
+    if (store == null || !store.isBlobTransferEnabled() || store.isHybrid()) {
```
adamxchen, line 518: `if (store == null)`?

services/venice-router/src/main/java/com/linkedin/venice/router/MetaDataHandler.java
```diff
@@ -477,6 +491,66 @@ private void handleCurrentVersionLookup(ChannelHandlerContext ctx, VenicePathPar
...
+          .getPartitionStatus(storeName, Integer.parseInt(storeVersion), Integer.parseInt(storePartition), null);
+      List<String> liveNodeHostNames = instances.entrySet()
+          .stream()
+          .map(Map.Entry::getKey)
+          .map(CharSequence::toString)
+          .filter(instanceHostName -> pushStatusStoreReader.isInstanceAlive(storeName, instanceHostName))
+          .collect(Collectors.toList());
+      response.setLiveNodeNames(liveNodeHostNames);
+    } catch (VeniceException e) {
+      response.setError(
```
adamxchen, line 547: we don't have error in `BlobDiscoveryResponse`. Can you add it?

ishwarya-citro: `BlobDiscoveryResponse` is extending from [ControllerResponse](https://jarvis.corp.linkedin.com/codesearch/result/?name=ControllerResponse.java&path=venice%2Finternal%2Fvenice-client-common%2Fsrc%2Fmain%2Fjava%2Fcom%2Flinkedin%2Fvenice%2Fcontrollerapi&reponame=linkedin%2Fvenice#64) which has the `setError()`

All of the "Responses" under `controllerapi` path extend from ControllerResponse. I.E. [MultiSchedmaIdResponse](https://jarvis.corp.linkedin.com/codesearch/result/?name=MultiSchemaIdResponse.java&path=venice%2Finternal%2Fvenice-client-common%2Fsrc%2Fmain%2Fjava%2Fcom%2Flinkedin%2Fvenice%2Fcontrollerapi&reponame=linkedin%2Fvenice) & [SystemStoreHeartbeatResponse](https://jarvis.corp.linkedin.com/codesearch/result/?name=SystemStoreHeartbeatResponse.java&path=venice%2Finternal%2Fvenice-client-common%2Fsrc%2Fmain%2Fjava%2Fcom%2Flinkedin%2Fvenice%2Fcontrollerapi&reponame=linkedin%2Fvenice)

internal/venice-common/src/main/java/com/linkedin/venice/controllerapi/ControllerApiConstants.java
ishwarya-citro: Ok. Instead of using `VERSION`, I am adding `STORE_VERSION` to be consistent

internal/venice-common/src/main/java/com/linkedin/venice/controllerapi/BlobDiscoveryResponse.java
ishwarya-citro: [see comment ](https://github.com/linkedin/venice/pull/980/files#r1600163060)

services/venice-router/src/test/java/com/linkedin/venice/router/TestMetaDataHandler.java
```diff
@@ -185,13 +218,242 @@ public void testControllerLookup() throws IOException {
...
-  // The deprecated non inclusive URL must also continue to work.
+  // The deprecated non-inclusive URL must also continue to work.
   @DataProvider(name = "controllerUrlProvider")
   public static Object[][] dataProvider() {
     // go/inclusivecode deprecated (alias="leader_controller")
     return new Object[][] { { "master_controller" }, { "leader_controller" } };
   }
 
+  @DataProvider(name = "blobDiscoveryProvider")
+  public static Object[][] blobDiscoveryDataProvider() {
```
adamxchen, line 229: I am not a big fan of this huge data provider. We use data provider because we want to examine different combination of inputs, for example, [this one](https://github.com/linkedin/venice/blob/5c0e66a203be1e9cfd0c6cad27597035543f06a9/services/venice-router/src/test/java/com/linkedin/venice/router/TestAdminOperationsHandler.java#L194). 

If I understand your intention right, we are just providing the mocked response in this data provider so it's better just to mock them in the test method itself instead of having a data provider

adamxchen (changes requested):
Can we also have one integ test to verify the API works as expected?

services/venice-router/src/test/java/com/linkedin/venice/router/TestMetaDataHandler.java
ishwarya-citro: Agree. Not a fan of a huge data provider either! 

We have dynamic response based on the inputs.

Another option I was thinking of is having a separate test method but thought this might be easier to read the test case. I can change this

ishwarya-citro:
> Can we also have one integ test to verify the API works as expected?

does calling the endpoint count as a integ test? As i think that covers the entire Scenario ?I don't see any integ tests written for new router endpoints - pl correct me if i am wrong

internal/venice-common/src/main/java/com/linkedin/venice/controllerapi/ControllerApiConstants.java
```diff
@@ -8,6 +8,8 @@ public class ControllerApiConstants {
   public static final String BATCH_JOB_HEARTBEAT_ENABLED = "batch_job_heartbeat_enabled";
 
   public static final String NAME = "store_name";
+  public static final String STORE_PARTITION = "store_partition";
+  public static final String STORE_VERSION = "store_version";
```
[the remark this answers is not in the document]
ishwarya-citro: "store_version" aligns more / is consistent with other variables names we're using; if we think "partition" only applies to a `store_partition` , i can change it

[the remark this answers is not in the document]
adamxchen: Sure, we can keep it

adamxchen:
> > Can we also have one integ test to verify the API works as expected?
> 
> does calling the endpoint count as a integ test? As i think that covers the entire Scenario ?I don't see any integ tests written for new router endpoints - pl correct me if i am wrong

Ideally, we should be writing an integ test to cover a new API too, such as [this test](https://github.com/linkedin/venice/blob/53b39bf0c8e6b78d3afa47f0393c3421ef7cbbe1/internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/router/TestRead.java#L705), which tests the `TYPE_CURRENT_VERSION` API of router. Or, you can add the integ test to [PushStatusStoreTest](https://github.com/linkedin/venice/blob/53b39bf0c8e6b78d3afa47f0393c3421ef7cbbe1/internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/PushStatusStoreTest.java#L76) (this looks like a better place to put this test.

services/venice-router/src/main/java/com/linkedin/venice/router/RouterServer.java
```diff
@@ -318,6 +324,13 @@ public RouterServer(
         config.getRefreshAttemptsForZkReconnect(),
         config.getRefreshIntervalForZkReconnectInMs());
     this.liveInstanceMonitor = new HelixLiveInstanceMonitor(this.zkClient, config.getClusterName());
+
+    D2Client d2Client = D2ClientFactory.getD2Client(config.getZkConnection(), Optional.empty());
```
gaojieliu, line 328: This is not right as `getZkConnection` is for Venice ZK.
I think we should let the backend pass an instance of `ZKClient` instead of constructing it here.

gaojieliu (changes requested):

services/venice-router/src/main/java/com/linkedin/venice/router/RouterServer.java
ishwarya-citro: @gaojieliu 

I don't believe we need an instance of `ZKClient` here. We'd need the [ZK Address](https://jarvis.corp.linkedin.com/codesearch/result/?name=VeniceRouterConfig.java&path=venice%2Fservices%2Fvenice-router%2Fsrc%2Fmain%2Fjava%2Fcom%2Flinkedin%2Fvenice%2Frouter&reponame=linkedin%2Fvenice#406) which is coming from the [router config](https://jarvis.corp.linkedin.com/codesearch/result/?name=RouterServer.java&path=venice%2Fservices%2Fvenice-router%2Fsrc%2Fmain%2Fjava%2Fcom%2Flinkedin%2Fvenice%2Frouter&reponame=linkedin%2Fvenice#207) file.   

If we can't use the `getZKConnection`, is there another way to create an instance of `D2Client`?

services/venice-router/src/main/java/com/linkedin/venice/router/MetaDataHandler.java
```diff
@@ -477,6 +504,69 @@ private void handleCurrentVersionLookup(ChannelHandlerContext ctx, VenicePathPar
...
+      byte[] errBody = (String.format(REQUEST_BLOB_DISCOVERY_ERROR_INVALID_SETTINGS, storeName)).getBytes();
+      setupResponseAndFlush(FORBIDDEN, errBody, false, ctx);
+      return;
+    }
+
+    BlobDiscoveryResponse response = new BlobDiscoveryResponse();
+    try {
+      // gets the instances for a FULL_PUSH for the store's version and partitionId
+      // gets the instance's hostnames from its keys & filter to include only live instances
+      Map<CharSequence, Integer> instances = pushStatusStoreReader.getPartitionStatus(
```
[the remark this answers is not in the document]
ishwarya-citro: as[ per this thread](https://linkedin-randd.slack.com/archives/C06HAFXUMPZ/p1716926655209899), we will revisit this after the changes have been made in future.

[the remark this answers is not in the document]
FelixGV: That's right, let's not post links to LinkedIn's internal slack. That being said, this may be a good opportunity to remind that we have the open source slack available to us, and it may be preferable to move all conversations about this project to a new public channel in the open source slack (:

internal/venice-common/src/main/java/com/linkedin/venice/controllerapi/BlobDiscoveryResponse.java
adamxchen, line 1: Actually, can we move this class to `com.linkedin.venice.routerapi` under `internal/venice-client-common`? They are not controller API but router API. We were not following.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/router/TestRead.java
```diff
@@ -121,11 +158,56 @@ protected VeniceClusterWrapper getVeniceCluster() {
...
       return;
     }
+    if ("testBlobDiscovery".equals(method.getName())) {
+      setUpForBlobTransfer();
+    } else {
+      setUpTestRead();
+    }
+  }
+
+  public void setUpForBlobTransfer() {
```
adamxchen, line 173: The idea of setting up totally different things for a test case in a single test file doesn't sound great to me. If a new setup is required, let's move to a new integ test file instead of complicating this file.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/router/TestRead.java
```diff
@@ -753,4 +836,157 @@ public void testRequestUsageMetric() {
     assertEquals(metricsRepository.metrics().get(requestUsageMetric).value(), 1.0d);
     assertEquals(metricsRepository.metrics().get(badRequestMetric).value(), 1.0d);
   }
+
+  @Test(timeOut = 10 * Time.MINUTES_PER_HOUR)
```
adamxchen, line 840: DO we really need 10 mins to run this test?

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/router/TestRead.java
```diff
@@ -753,4 +836,157 @@ public void testRequestUsageMetric() {
...
+  public void testBlobDiscovery() throws Exception {
+    if (!isTestEnabled()) {
+      return;
+    }
+
+    assertTrue(clusterNames.length > 1, "Insufficient clusters for this test to be meaningful");
```
adamxchen, line 846: Why does the test require more than 1 cluster?

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/router/TestRead.java
```diff
@@ -753,4 +836,157 @@ public void testRequestUsageMetric() {
...
+        }
+        Assert.assertEquals(
+            response.getStatusLine().getStatusCode(),
+            HttpStatus.SC_OK,
+            "Failed to get resource state for " + storeVersionName + ". Response: " + responseBody);
+        ObjectMapper mapper = ObjectMapperFactory.getInstance();
+        BlobDiscoveryResponse blobDiscoveryResponse =
+            mapper.readValue(responseBody.getBytes(), BlobDiscoveryResponse.class);
+        Assert.assertEquals(blobDiscoveryResponse.getLiveNodeHostNames().size(), 0);
+        LOGGER.info(responseBody);
```
adamxchen, line 975: we don't need log here

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/router/TestRead.java
```diff
@@ -753,4 +836,157 @@ public void testRequestUsageMetric() {
...
+          responseBody = IOUtils.toString(bodyStream);
+        }
+        Assert.assertEquals(
+            response.getStatusLine().getStatusCode(),
+            HttpStatus.SC_OK,
+            "Failed to get resource state for " + storeVersionName + ". Response: " + responseBody);
+        ObjectMapper mapper = ObjectMapperFactory.getInstance();
+        BlobDiscoveryResponse blobDiscoveryResponse =
+            mapper.readValue(responseBody.getBytes(), BlobDiscoveryResponse.class);
+        Assert.assertEquals(blobDiscoveryResponse.getLiveNodeHostNames().size(), 0);
```
adamxchen, line 974: Please also add a TODO here since we need to come back and improve the test. A response contains 0 node is not sufficient

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/router/TestRead.java
```diff
@@ -753,4 +836,157 @@ public void testRequestUsageMetric() {
...
+        LOGGER.info(responseBody);
+      } catch (Exception e) {
+        fail("Unexpected exception", e);
+      }
+    } finally {
+      D2ClientUtils.shutdownClient(daVinciD2);
```
adamxchen, line 980: this could be a part of `teardown` process of a test suite

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/router/TestRead.java
ishwarya-citro: created a new file

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/router/TestRead.java
ishwarya-citro: nope. Used that to test out the time out which no longer happens

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/router/TestBlobDiscovery.java
```diff
@@ -0,0 +1,230 @@
+package com.linkedin.venice.router;
+
+import static com.linkedin.venice.ConfigKeys.*;
```
adamxchen, line 3: not import all

adamxchen (approved):
Thank you for the hard work

services/venice-router/src/main/java/com/linkedin/venice/router/RouterServer.java
gaojieliu: I saw the code got checked in and the zk zddress in `router config` file is for Venice Zookeeper, which is different from D2 Zookeeper, and I assume here we want to use D2 Zookeeper Address to create a D2 client to access push status store, right?
The idea in the original comment is that it is easy to get a D2 client from the internal factory, and if that is not easy for testing, we will need to introduce an additional config to hold D2 Zookeeper address.

PR #982, opened 2024-05-13
[common][server] Bug fixes and clean of ACL handlers for GRPC

mynameborat:
## Summary
- Consolidate boiler plate code in ACL handlers for access validations
- Fix code path that throws exception in GRPC and close the calls with the appropriate error messages
- Fix the flow that invokes `onMessage` twice in GRPC call intercept to prevent exceptions
- Rename `GrpcSSLUtils` to `GrpcUtils` to accommodate general utility methods
- Add reflection service to GRPC server to interact with Grpcurl
- Reuse the header for skipping ACLs from router since the store ACL handling is performed at the router for non FC clients
- [code readability] Move public methods together followed by overridable and protected methods followed by package private and private methods

## How was this PR tested?
- Unit tests
- Integration tests
- Deployment to certification cluster to validate the flow using sample venice application


internal/venice-common/src/main/java/com/linkedin/venice/acl/handler/StoreAclHandler.java
```diff
@@ -251,74 +131,158 @@ public <ReqT, RespT> ServerCall.Listener<ReqT> interceptCall(
...
-          }
-        } catch (AclException e) {
+          X509Certificate clientCert = extractGrpcClientCert(call);
           String client = Objects.requireNonNull(call.getAttributes().get(Grpc.TRANSPORT_ATTR_REMOTE_ADDR)).toString();
-          String errLine = String.format("%s requested %s %s", client, method, storeName);
 
-          if (accessController.isFailOpen()) {
-            LOGGER.warn("Exception occurred! Access granted: {} {}", errLine, e);
+          if (VeniceSystemStoreUtils.isSystemStore(storeName)
+              || hasAccess(client, "", clientCert, storeName, method, errorHandler)) {
```
xunyin8, line 153: Looks like URI is only used for informative purpose but why are we passing empty string `""` for URI?

xunyin8 (commented):
Mostly looks good to me, just one question/concern.

internal/venice-common/src/main/java/com/linkedin/venice/acl/handler/StoreAclHandler.java
mynameborat: We don't have the request in case of GRPC. I initially left it empty for this reason. However, maybe we can populate it with just the authority (host w/ port) although that isn't the same as URI we get on the other path with netty.

What do you think?

mynameborat:
@xunyin8 I think i narrowed down the issue with mockito-inline upgrade which doesn't play well with the bytebuddy we pull in transitively. For now, reverting the mockito dependency upgrade and removed the test that requires mocking static utility.

Will have a separate PR to fix the mocking issue (for mocking static and final classes)

internal/venice-common/src/main/java/com/linkedin/venice/acl/handler/StoreAclHandler.java
xunyin8: I see. Yeah just the authority is good too, better than nothing.

xunyin8 (approved):
LGTM, feel free to merge once the tests are passing.

PR #984, opened 2024-05-15
[server] Lag monitor should not expect corresponding version always exists in ZK

xunyin8:
## [server] Lag monitor should not expect corresponding version always exists in ZK
1. updateLagMonitor was implemented to expect the version object to never be null. This assumption is wrong because when we are deleting a version the version object is removed from ZK first before the state transition occurs on the server from STANDBY->OFFLINE. Add the option for updateLagMonitor to accept null version if it's called for removeLagMonitor.

2. Separated the logging for null store and null version to be explicit and not logging store objects.

## How was this PR tested?
New unit test and existing unit+integration tests

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/LeaderFollowerPartitionStateModel.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/LeaderFollowerPartitionStateModel.java
@@ -122,7 +122,11 @@ public class LeaderFollowerPartitionStateModel extends AbstractPartitionStateMod
       if (isRegularStoreCurrentVersion) {
         waitConsumptionCompleted(resourceName, notifier);
       }
-      updateLagMonitor(message.getResourceName(), heartbeatMonitoringService::addFollowerLagMonitor, false);
+      updateLagMonitor(
+          message.getResourceName(),
+          heartbeatMonitoringService::addFollowerLagMonitor,
+          false,
+          messageToString(message));
     });
   }
 
@@ -134,7 +138,11 @@ public class LeaderFollowerPartitionStateModel extends AbstractPartitionStateMod
      * where a slice doesn't have a replicating leader.  While this state transition executes, there should be no
      * other leader in the slice.
      */
-    updateLagMonitor(message.getResourceName(), heartbeatMonitoringService::addLeaderLagMonitor, false);
+    updateLagMonitor(
+        message.getResourceName(),
+        heartbeatMonitoringService::addLeaderLagMonitor,
+        false,
+        messageToString(message));
     executeStateTransition(
         message,
         context,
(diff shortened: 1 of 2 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/LeaderFollowerPartitionStateModel.java
```diff
@@ -133,7 +134,7 @@ public void onBecomeLeaderFromStandby(Message message, NotificationContext conte
      * where a slice doesn't have a replicating leader.  While this state transition executes, there should be no
      * other leader in the slice.
      */
-    updateLagMonitor(message.getResourceName(), heartbeatMonitoringService::addLeaderLagMonitor);
+    updateLagMonitor(message.getResourceName(), heartbeatMonitoringService::addLeaderLagMonitor, false);
```
sushantmane, line 137: A race between inflight ST and version retirement is possible. If version is deleted first, then we'll consider `isNullVersionValid` as false which may not be true?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/LeaderFollowerPartitionStateModel.java
```diff
@@ -196,23 +197,33 @@ public void onBecomeOfflineFromError(Message message, NotificationContext contex
...
+            getPartition());
         return;
       }
+      if (version == null) {
+        if (isNullVersionValid) {
+          // During version deletion, the version will be deleted from ZK prior to servers perform resource deletion.
+          // It's valid to have null version when trying to remove lag monitor for the deleted resource.
+          version = new VersionImpl(storeName, storeVersion, "");
+        } else {
+          logger.error(
```
sushantmane, line 221: We need `return` here otherwise we'll get NPE. 

Maybe we can restructure this as below?
```java
if (version == null && !isNullVersionValid) {
    logger.error("Failed to get version for resource: {}-{}. Will not update lag monitor.", resourceName, getPartition());
    return;
}

if (version == null) {
    // During version deletion, the version will be deleted from ZK prior to servers perform resource deletion.
    // It's valid to have null version when trying to remove lag monitor for the deleted resource.
    version = new VersionImpl(storeName, storeVersion, "");
}
```

clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/LeaderFollowerPartitionStateModel.java
```diff
@@ -196,23 +197,33 @@ public void onBecomeOfflineFromError(Message message, NotificationContext contex
...
+              "Failed to get version for resource: {}-{}. Will not update lag monitor.",
```
sushantmane, line 222: Can we please log state transition during which this update lag monitor was called?

sushantmane (commented):
Thanks, @xunyin8, for the quick turnaround!

clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/LeaderFollowerPartitionStateModel.java
xunyin8: Good catch

clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/LeaderFollowerPartitionStateModel.java
xunyin8: Will do.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/LeaderFollowerPartitionStateModel.java
xunyin8: In this scenario I think we should just log the error and do nothing instead of trying to add leader lag monitor for a store version that will soon be deleted (LEADER/STANDBY->OFFLINE). Such race should be rare too, we can make incremental changes. If this really becomes common or a problem we should implement a better mechanism for this.

sushantmane: That sounds reasonable. Thank!

sushantmane (approved):
LGTM, Thanks, @xunyin8!

PR #985, opened 2024-05-15
[server] Feature to tune level0 compaction tuning for read-write leader

gaojieliu:
In the ingestion path, we often tune level0 compaction to optimize read/write/space optimization.
When AA/WC is enabled, the leader replica will read the same database while ingesting, which is different from standby replicas. To optimize the read perf of leader replica, this PR exposes the following config:
rocksdb.level0.compaction.tuning.for.read.write.leader.enabled: default false rocksdb.level0.file.num.compaction.trigger.for.read.write.leader: default 10 rocksdb.level0.slowdown.writes.trigger.for.read.write.leader: default 20 rocksdb.level0.stops.writes.trigger.for.read.write.leader: default 40

Essentially, there are 3 sets of level0 compaction tunings:
1. Leader replicas for AA/WC stores.
2. Future version.
3. Current version.

The main reason we would like to only optimize the leader replica for AA/WC stores:
1. Standby replicas don't read at ingestion path.
2. Try to keep the write/space amplication factor for standby replicas as more aggressive level0 compaction tuning is expensive.

## How was this PR tested?
CI

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/AbstractStorageEngine.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/AbstractStorageEngine.java
@@ -195,10 +195,6 @@ public abstract class AbstractStorageEngine<Partition extends AbstractStoragePar
           "StoragePartitionConfig should contain the right partition id: " + partitionId + ", but got "
               + partitionConfig.getPartitionId());
     }
-    if (!containsPartition(partitionId)) {
-      LOGGER.warn("Partition {}_{} was removed before adjusting.", storeVersionName, partitionId);
-      return;
-    }
     LOGGER.info("Storage partition adjustment got triggered by: {} with config: {}", mode, partitionConfig);
     AbstractStoragePartition partition = getPartitionOrThrow(partitionId);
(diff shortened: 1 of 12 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/AbstractStorageEngine.java
```diff
@@ -171,26 +179,30 @@ public AbstractStoragePartition getMetadataPartition() {
...
+      StoragePartitionAdjustmentTrigger mode,
+      StoragePartitionConfig partitionConfig) {
     validateStoreName(partitionConfig);
-    int partitionId = partitionConfig.getPartitionId();
+    if (partitionId != partitionConfig.getPartitionId()) {
+      throw new VeniceException(
+          "StoragePartitionConfig should contain the right partition id: " + partitionId + ", but got "
+              + partitionConfig.getPartitionId());
+    }
+    if (!containsPartition(partitionId)) {
```
majisourav99, line 198: Previously we used to throw exception for partition not found in following call to `getPartitionOrThrow`. Why do we want to silently return now?

majisourav99 (approved):

PR #986, opened 2024-05-15
[test] Fix for cross-test interrupt contamination

FelixGV:
There is an issue either in Gradle or TestNG where a test which sets the interrupt flag on the current thread can cause a later test to fail, due to checking for the interrupt flag (e.g. in sleep and syncrhonization-related primitives).

This commit fixes this by clearing the interrupt flag via the VeniceTestListener::onTestStart listener. Also registered this listener to unit tests (it was previously used only for integ tests), and added a minimal unit test which deterministically reproduces this issue (in the absence of the listener).

Miscellaneous:

- Added calls to super in all functions of VeniceTestListener, as recommended in the parent class' JavaDoc.

## How was this PR tested?
GHCI, including a new test.

### DIFF
--- a/internal/venice-test-common/src/main/java/com/linkedin/venice/testng/VeniceTestListener.java
+++ b/internal/venice-test-common/src/main/java/com/linkedin/venice/testng/VeniceTestListener.java
@@ -12,6 +12,8 @@ public class VeniceTestListener extends TestListenerAdapter {
   @Override
   public void onTestStart(ITestResult result) {
     super.onTestStart(result);
+
+    /** N.B.: {@link Thread#interrupted()} not only retrieves, but also clears, the interrupt flag */
     boolean interruptCleared = Thread.interrupted();
     LOGGER.info(
         "\n\n######## TEST ######## {}({}){} - STARTED",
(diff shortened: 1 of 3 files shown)


internal/venice-test-common/src/main/java/com/linkedin/venice/testng/VeniceTestListener.java
```diff
@@ -11,14 +11,18 @@ public class VeniceTestListener extends TestListenerAdapter {
 
   @Override
   public void onTestStart(ITestResult result) {
+    super.onTestStart(result);
+    boolean interruptCleared = Thread.interrupted();
```
sushantmane, line 15: Can we just mentioned here that  `Thread.interrupted() clears interrupt status?

FelixGV: Sure, one sec...

sushantmane: Thanks!

sushantmane (approved):
Great find! Thanks a lot, @FelixGV!

PR #987, opened 2024-05-16
[internal][client] Support GRPC port discovery for non-local environment

mynameborat:
## Summary
- [Functional] Introduce logic to use fallback Grpc port for netty server address -> Grpc server mapping
- [Refactor] Modularize request handling and remove boiler plate in `GrpcTransportClient`
- [Bug fix] Fix integration tests that was failing due to #975 
- [Clean up] Remove some validations, added few invariants check and reorganize some depending on the mode (local vs integration/production)
- [Clean up] Rename fields to remove type information
- [Test] Add unit tests for `GrpcTransportClient` class

## How was this PR tested?
- Unit test and integration tests
- Deployed client changes locally with test application

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -110,7 +110,7 @@ ext.libraries = [
     mail: 'javax.mail:mail:1.4.4',
     mapreduceClientCore: "org.apache.hadoop:hadoop-mapreduce-client-core:${hadoopVersion}",
     mapreduceClientJobClient: "org.apache.hadoop:hadoop-mapreduce-client-jobclient:${hadoopVersion}",
-    mockito: 'org.mockito:mockito-inline:4.11.0',
+    mockito: 'org.mockito:mockito-core:5.2.0',
     netty: 'io.netty:netty-all:4.1.74.Final',
     oss: 'org.sonatype.oss:oss-parent:7',
     pulsarClient: "${pulsarGroup}:pulsar-client:${pulsarVersion}",
--- a/build.gradle
+++ b/build.gradle
@@ -110,7 +110,7 @@ ext.libraries = [
     mail: 'javax.mail:mail:1.4.4',
     mapreduceClientCore: "org.apache.hadoop:hadoop-mapreduce-client-core:${hadoopVersion}",
     mapreduceClientJobClient: "org.apache.hadoop:hadoop-mapreduce-client-jobclient:${hadoopVersion}",
-    mockito: 'org.mockito:mockito-core:4.11.0',
+    mockito: 'org.mockito:mockito-inline:4.11.0',
     netty: 'io.netty:netty-all:4.1.74.Final',
     oss: 'org.sonatype.oss:oss-parent:7',
     pulsarClient: "${pulsarGroup}:pulsar-client:${pulsarVersion}",
--- a/build.gradle
+++ b/build.gradle
@@ -110,7 +110,7 @@ ext.libraries = [
     mail: 'javax.mail:mail:1.4.4',
     mapreduceClientCore: "org.apache.hadoop:hadoop-mapreduce-client-core:${hadoopVersion}",
     mapreduceClientJobClient: "org.apache.hadoop:hadoop-mapreduce-client-jobclient:${hadoopVersion}",
-    mockito: 'org.mockito:mockito-core:5.2.0',
+    mockito: 'org.mockito:mockito-core:4.11.0',
     netty: 'io.netty:netty-all:4.1.74.Final',
     oss: 'org.sonatype.oss:oss-parent:7',
     pulsarClient: "${pulsarGroup}:pulsar-client:${pulsarVersion}",
--- a/clients/venice-client/src/main/java/com/linkedin/venice/fastclient/transport/GrpcTransportClient.java
+++ b/clients/venice-client/src/main/java/com/linkedin/venice/fastclient/transport/GrpcTransportClient.java
@@ -39,7 +39,6 @@ public class GrpcTransportClient extends InternalTransportClient {
   private static final Logger LOGGER = LogManager.getLogger(GrpcTransportClient.class);
   private static final String STORAGE_ACTION = "storage";
   private static final String GRPC_ADDRESS_FORMAT = "%s:%s";
-  private static final String GRPC_PORT = "23900";
   private final VeniceConcurrentHashMap<String, ManagedChannel> serverGrpcChannels;
   private final Map<String, String> nettyServerToGrpcAddress;
   // we cache stubs to avoid creating a new stub for each request, improves performance
@@ -48,10 +47,18 @@ public class GrpcTransportClient extends InternalTransportClient {
   private final TransportClient r2TransportClientForNonStorageOps; // used for non-storage related requests
   private final ChannelCredentials channelCredentials;
 
+  private final int port;
+
   public GrpcTransportClient(GrpcClientConfig grpcClientConfig) {
     this(
         new R2TransportClient(grpcClientConfig.getR2Client()),
+        /*
+         * General principle to not mutate input parameters and wrapping it up using a new map since this
+         * class updates the map with inferred grpc address to prevent redundant computation for mapping netty
+         * to grpc address
+         */
         new HashMap<>(grpcClientConfig.getNettyServerToGrpcAddress()),
+        grpcClientConfig.getPort(),
         grpcClientConfig.getSslFactory());
   }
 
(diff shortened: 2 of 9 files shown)


clients/venice-client/src/main/java/com/linkedin/venice/fastclient/transport/GrpcTransportClient.java
```diff
@@ -22,209 +24,278 @@
...
-  public CompletableFuture<TransportClientResponse> get(String requestPath, Map<String, String> headers) {
-    return handleRequest(requestPath, headers, null, true);
+  /**
+   * Maps the given server address to GRPC server address. Typically, the mapped server address only differs in the
+   * port. Fetch the GRPC server address for a given server address if available in the #nettyServerToGrpcAddress
+   * map if not, strip the port from the original server address and replace it with the grpc port.
+   * <b>Note:</b> We have the GRPC port hard coded and operate with static port assignment for GRPC
+   */
+  @VisibleForTesting
+  String getGrpcAddressFromServerAddress(String serverAddress) {
```
xunyin8, line 146: I didn't realize how static the mapping needed to be. This means if we'd like to configure the port to something different than the fallback `GRPC_PORT` we need to provide a config map to the client that contains ALL the servers that's potentially reachable by that client (could be multiple clusters). That doesn't sound very appealing... if that's the case shall we also make the fallback configurable in case we want to change the server side hardcode to something that's not `23900`.

xunyin8 (commented):
LGTM, just some minor concerns around configuring the GRPC ports

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/transport/GrpcTransportClient.java
mynameborat: Agreed. I thought about it and was going to put #1 in this PR given its short term. While I was going do option 2 or 3 in the mid/long term if things look good as part of taking this to production.

My only reason for abandoning option 1 was introducing a new client configuration and the debt that tags along in terms of deprecating it if we are going to jump to option 2 or 3. 

How easy with venice when it comes to deprecating configurations? Would piggy backing the existing `nettyServerToGrpcAddress` be flexible enough at the same time prevent introducing public facing configurations as part of client?

build.gradle
```diff
@@ -109,7 +109,7 @@ ext.libraries = [
     mail: 'javax.mail:mail:1.4.4',
     mapreduceClientCore: "org.apache.hadoop:hadoop-mapreduce-client-core:${hadoopVersion}",
     mapreduceClientJobClient: "org.apache.hadoop:hadoop-mapreduce-client-jobclient:${hadoopVersion}",
-    mockito: 'org.mockito:mockito-core:4.11.0',
```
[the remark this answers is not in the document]
mynameborat: mockito-inline allows mocking final classes and others(static e.g.,) which are limitations in mockit-core

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/transport/GrpcTransportClient.java
```diff
@@ -22,209 +24,278 @@
...
-  private final GrpcClientConfig grpcClientConfig;
-  private final SSLFactory sslFactory;
-  private ChannelCredentials channelCredentials = InsecureChannelCredentials.create();
+  private final ChannelCredentials channelCredentials;
 
   public GrpcTransportClient(GrpcClientConfig grpcClientConfig) {
-    this.grpcClientConfig = grpcClientConfig;
+    this(
+        new R2TransportClient(grpcClientConfig.getR2Client()),
+        new HashMap<>(grpcClientConfig.getNettyServerToGrpcAddress()),
```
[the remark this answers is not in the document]
mynameborat: `GrpcTransportClient` uses this map to update / cache the mapping for the ones that we compute. I didn't want to mutate the map that is passed through the configuration as a general practice.

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/transport/GrpcTransportClient.java
```diff
@@ -22,209 +24,278 @@
...
+    public VeniceGrpcStreamObserver(CompletableFuture<TransportClientResponse> responseFuture) {
+      Preconditions.checkArgument(!responseFuture.isDone());
+      this.responseFuture = responseFuture;
     }
 
-    public CompletableFuture<TransportClientResponse> post() {
-      if (!request.getIsBatchRequest()) {
-        throw new UnsupportedOperationException("Not a batch get request, use get() instead");
+    @Override
+    public void onNext(VeniceServerResponse value) {
```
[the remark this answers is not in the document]
mynameborat: That is correct. Here is the java doc that describes the guarantees

https://github.com/grpc/grpc-java/blob/8aaace12eb8697a2dd841d2261b0051a1414f0c8/stub/src/main/java/io/grpc/stub/StreamObserver.java#L47

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/transport/GrpcTransportClient.java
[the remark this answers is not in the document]
gaojieliu: I see.
Can you leave some javadoc for ^?

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/transport/GrpcTransportClient.java
gaojieliu: Client config for GRPC port doesn't sound great even for the first customer as we are leaking backend config into client.
One workaround is to maintain a global GRPC port config, which will be managed by us and if we want to change it, we will change the global config instead of asking everyone update it, and this can work with your current code change.
Global config for GRPC port is not great as option 2/3 as the customer still needs to do a new deployment to pick up the new port, but it might be acceptable as we hide it from the customer and still manage it in a centralized fashion for the time being.

mynameborat: I was mostly referring to introducing a client config within `ClientConfig` be it internal. I think I am on the same page is making `23900` configurable through some means. Do you guys think we can leverage the existing map that we already use to pass this port to the `GrpcTransportClient`? If so, we can get away from introducing another configuration within venice repo but still populate the map externally using a global configuration keeping the internal contract more flexible for deprecation.

gaojieliu: Shall we just have a single config for the GPRC port instead of a map?
The assumption is that all the instances will use the same port in the same cluster.
And this would only be used by the benchmarking, not the prod use case, right?

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/transport/GrpcTransportClient.java
```diff
@@ -22,209 +24,286 @@
...
+   * Maps the given server address to GRPC server address. Typically, the mapped server address only differs in the
+   * port. Fetch the GRPC server address for a given server address if available in the #nettyServerToGrpcAddress
+   * map if not, strip the port from the original server address and replace it with the grpc port.
+   * <b>Note:</b> We have the GRPC port hard coded and operate with static port assignment for GRPC
+   */
+  @VisibleForTesting
+  String getGrpcAddressFromServerAddress(String serverAddress) {
+    Preconditions.checkState(StringUtils.isNotEmpty(serverAddress), "Server address cannot be empty ");
+
+    nettyServerToGrpcAddress.computeIfAbsent(serverAddress, nettyAddress -> {
```
xunyin8, line 157: In most cases this method is just returning `serverAddress` + GRPC port. I guess one reason I can think of for still keeping the `nettyServerToGrpcAddress` map is to give us flexibility in case we have a few server hosts with different GRPC port.

xunyin8 (approved):
LGTM, just one small question.

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/transport/GrpcTransportClient.java
mynameborat: actually, the map only exists for the purpose of integration tests where ew bring up multiple venice servers and we need to map the netty server address to grpc server address (ports are dynamically generated).

Once, we add port discovery through metadata the map will go away

PR #988, opened 2024-05-16
[server] Revert lossy rewind criterion to EOP from completion reporting

sushantmane:
## Revert lossy rewind criterion to EOP from completion reporting
The PR: #846 modified the method responsible for handling lossy rewind checks: it updated
the function to suppress the lossy rewind issue only after reporting completion, rather
than upon receiving the EOP. It is important to note that the status of reported completion
is not permanent and can change during state transitions. As a result of this change, there
is an increased likelihood of ingestion stalls / failures due to lossy rewinds, compared
to the previous method based on the EOP signal.

This PR reverts the check function to its original state, where the lossy rewind issue is
suppressed after reporting the EOP signal, as it was before pull request #846.
This fix is intended to avoid serving stale data. The only downside is that data
for a few keys might be inconsistent, but this issue also exists with the completion-based
check. Since this is a short-term solution, this is okay. We're working on a solution
that allows us to ignore data from illegitimate topic writers.


gaojieliu (approved):
Thanks for the change!

sushantmane:
Thanks a lot, @gaojieliu for speedy review! 🙏

PR #990, opened 2024-05-17
[dvc][server][controller][router] Removed Optional from Store API

FelixGV:
The Store::getVersion API used to return an Optional<Version>, now it instead returns a @Nullable Version. Also added a getVersionOrThrow API which can throw StoreVersionNotFoundException, a new exception type.

Lots of refactoring to adapt to the change, though these should be non-functional changes.

Significant changes:

- Some usages of getVersion in the consumer client were doing naked .get() so these have been replaced by getVersionOrThrow.

- Refactored the HelixCustomizedViewOfflinePushRepository to make all changes to the resourceToPartitionCountMap go through fewer private functions, and fixed some naked .get() bug there as well.

- Deleted the ExpiringSet and its usage in ReadQuotaEnforcementHandler.

- Removed the dependency on Kafka (the broker) from most modules, since this should
  be a test-only dependency now.

Miscellaneous test code changes:

- Removed nearly all mocking of the deprecated function PubSubProducerAdapter::getNumberOfPartitions and replaced it with VeniceWriterOptions::setPartitionCount.

- Added timeouts to all TestVeniceHelixAdminWithSharedEnvironment tests.

- Added the ability to override the duration of some long operations in HelixReadOnlySchemaRepository and PubSubConstants, only for tests. This reduces venice-common unit tests by ~6 minutes.

## How was this PR tested?

GHCI.


services/venice-server/src/main/java/com/linkedin/venice/listener/ReadQuotaEnforcementHandler.java
```diff
@@ -55,7 +52,6 @@ public class ReadQuotaEnforcementHandler extends SimpleChannelInboundHandler<Rou
   private final String thisNodeId;
   private final AggServerQuotaUsageStats stats;
   private final Clock clock;
-  private final ExpiringSet<String> noBucketStores = new ExpiringSet<>(30, TimeUnit.SECONDS);
```
ZacAttack, line 58: I'm not sure I understand what about this refactor allows us to remove this?  Or... is just useless?

services/venice-controller/src/main/java/com/linkedin/venice/controller/ErrorPartitionResetTask.java
```diff
@@ -110,11 +111,15 @@ public void run() {
   }
 
   private void resetApplicableErrorPartitions(Store store) {
-    int currentVersion = store.getCurrentVersion();
-    if (currentVersion == Store.NON_EXISTING_VERSION || !store.getVersion(currentVersion).isPresent()) {
+    int currentVersionNumber = store.getCurrentVersion();
```
ZacAttack, line 114: Ok.  I 'think' this shouldn't be the case, but is there a chance that there isn't a current version?  I think even in the presence of no version we default the version number to 0

internal/venice-test-common/src/main/java/com/linkedin/venice/pubsub/PubSubConstantsOverrider.java
```diff
@@ -0,0 +1,14 @@
+package com.linkedin.venice.pubsub;
+
+import java.time.Duration;
+
+
+public class PubSubConstantsOverrider {
```
ZacAttack, line 6: Test utility comment or maybe make package private?

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicMetadataFetcher.java
```diff
@@ -1,6 +1,6 @@
 package com.linkedin.venice.pubsub.manager;
 
-import static com.linkedin.venice.pubsub.PubSubConstants.PUBSUB_OFFSET_API_TIMEOUT_DURATION_DEFAULT_VALUE;
+import static com.linkedin.venice.pubsub.PubSubConstants.*;
```
ZacAttack, line 3: tut tut

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/InternalLocalBootstrappingVeniceChangelogConsumer.java
```diff
@@ -9,7 +9,7 @@
 import static com.linkedin.davinci.store.rocksdb.RocksDBServerConfig.ROCKSDB_LEVEL0_STOPS_WRITES_TRIGGER_WRITE_ONLY_VERSION;
 import static com.linkedin.davinci.store.rocksdb.RocksDBServerConfig.ROCKSDB_PLAIN_TABLE_FORMAT_ENABLED;
 import static com.linkedin.venice.ConfigKeys.DATA_BASE_PATH;
-import static com.linkedin.venice.pubsub.PubSubConstants.PUBSUB_OFFSET_API_TIMEOUT_DURATION_DEFAULT_VALUE;
+import static com.linkedin.venice.pubsub.PubSubConstants.*;
```
ZacAttack, line 12: obligatory star import comment (stupid intellij)

clients/venice-admin-tool/src/main/java/com/linkedin/venice/KafkaTopicDumper.java
```diff
@@ -1,7 +1,7 @@
 package com.linkedin.venice;
 
 import static com.linkedin.venice.chunking.ChunkKeyValueTransformer.KeyType.WITH_VALUE_CHUNK;
-import static com.linkedin.venice.pubsub.PubSubConstants.PUBSUB_OFFSET_API_TIMEOUT_DURATION_DEFAULT_VALUE;
+import static com.linkedin.venice.pubsub.PubSubConstants.*;
```
ZacAttack, line 4: nit: remove *

internal/venice-common/src/main/java/com/linkedin/venice/helix/HelixCustomizedViewOfflinePushRepository.java
```diff
@@ -128,22 +129,9 @@ protected void onCustomizedViewDataChange(RoutingTableSnapshot routingTableSnaps
...
-
-          if (!store.getVersion(version).isPresent()) {
-            LOGGER.warn("Version not found in store for resource: {}.", resourceName);
-            continue;
-          }
-          partitionCount = store.getVersion(version).get().getPartitionCount();
-          resourceToPartitionCountMap.put(resourceName, partitionCount);
+        int partitionCount = getPartitionCount(resourceName);
+        if (partitionCount == STORE_VERSION_PARTITION_COUNT_NOT_FOUND) {
+          continue;
```
ZacAttack, line 134: Are we no longer interested in logging the above cases?

Admittedly, I'm not sure where we'd be in a position where we'd have a customized view entry and no discernable store config/version.  My guess is it could spring up if we were mid version deletion?  Possible, but not very interesting I guess.

ZacAttack: Nevermind, I see you just moved them to the function.

ZacAttack (commented):
Looks really good.  So much cleaner without all those dumb get()'s.  Just some nit comments.

services/venice-controller/src/main/java/com/linkedin/venice/controller/ErrorPartitionResetTask.java
ZacAttack: ah wait nm.  Doesn't matter, the check below covers this.

services/venice-server/src/main/java/com/linkedin/venice/listener/ReadQuotaEnforcementHandler.java
FelixGV: In my assessment it's useless. It serves only for logging, and we already have a metric for it. If we care about this, we ought to alert on the metric, and if we don't care about it, then we don't need the log anyway 😅

internal/venice-test-common/src/main/java/com/linkedin/venice/pubsub/PubSubConstantsOverrider.java
FelixGV: I added some JavaDoc here.

ZacAttack (dismissed):

ZacAttack (approved):
Looks good

FelixGV:
Thanks for the review @ZacAttack! I merged this change.

PR #991, opened 2024-05-17
[server] Remove consumer action queue inside StoreIngestionTask

haoxu07:
This change is for removing the consumer action queue inside `StoreIngestionTask`. Previously other thread will enqueue the consumer actions, and then `StoreIngestionTask` thread will process these consumer actions. 

-  `SUBSCRIBE` : triggered by helix transition from OFFLINE to STANDBY, `StoreIngestionTask` thread for processIngestionException.
- `UNSUBSCRIBE`: triggered by helix transition from STANDBY to OFFLINE,  `StoreBufferDrainer` thread (ready to serve check and div error handling), `StoreIngestionTask` thread for processIngestionException.
- `STANDBY_TO_LEADER`:  triggered by helix transition from STANDBY to LEADER
- `LEADER_TO_STANDBY`: triggered by helix transition from LEADER to STANDBY
- `KILL`:  `missingSOPCheckExecutor` thread. 
-  `RESET_OFFSET`:  triggered by helix transition from OFFLINE to DROPPED.

By removing the queue, each thread will be responsible for execution of the action and handling exceptions. Except for `KILL` Operation,  KILL operation will still be processed by `StoreIngestionTask` thread.  There are benefits:
1. State transition thread will be blocked until these operations finished: `standby_to_leader`, `leader_to_standby`. So that means there will only be one helix thread interacting with `StoreIngestionTask` for a specific version topic partition. So it is not possible a `standby_to_leader` considered to be finished, but subscribe operation as leader is not finished. 
2. Parallel execution of consumer action execution will improve the speed. 

Introduce a new config for enabling this feature so that we can roll back safely. 
Besides, `UNSUBSCRIBE` operation called from `StoreBufferService#Drainer` would cause problem as it will need to wait drainer to finish remain records.  Without introducing a new thread for executing the unsub operation, the drainer will be stuck. 

Resolves #XXX

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
@@ -107,6 +107,8 @@ import com.linkedin.venice.utils.Utils;
 import com.linkedin.venice.utils.VeniceProperties;
 import com.linkedin.venice.utils.concurrent.VeniceConcurrentHashMap;
 import com.linkedin.venice.utils.lazy.Lazy;
+import com.linkedin.venice.utils.locks.AutoCloseableLock;
+import com.linkedin.venice.utils.locks.ResourceAutoClosableLockManager;
 import it.unimi.dsi.fastutil.objects.Object2IntMap;
 import java.io.Closeable;
 import java.io.IOException;
@@ -135,6 +137,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.BooleanSupplier;
 import java.util.function.Consumer;
 import java.util.function.Function;
@@ -326,6 +329,10 @@ public abstract class StoreIngestionTask implements Runnable, Closeable {
   private final Runnable runnableForKillIngestionTasksForNonCurrentVersions;
   protected final AtomicBoolean recordLevelMetricEnabled;
 
+  protected final ResourceAutoClosableLockManager partitionLockManager;
+
+  private final AtomicBoolean killedActionReceived;
+
   public StoreIngestionTask(
       StoreIngestionTaskFactory.Builder builder,
       Store store,
@@ -360,6 +367,7 @@ public abstract class StoreIngestionTask implements Runnable, Closeable {
     // partitionConsumptionStateMap could be accessed by multiple threads: consumption thread and the thread handling
     // kill message
     this.partitionConsumptionStateMap = new VeniceConcurrentHashMap<>();
+    this.partitionLockManager = new ResourceAutoClosableLockManager<>(ReentrantLock::new);
 
     /**
      * Use the max of the two values to avoid the case where the producer state max age is set to a value that is less
@@ -382,6 +390,7 @@ public abstract class StoreIngestionTask implements Runnable, Closeable {
     this.versionedIngestionStats = builder.getVersionedStorageIngestionStats();
     this.isRunning = new AtomicBoolean(true);
     this.emitMetrics = new AtomicBoolean(true);
+    this.killedActionReceived = new AtomicBoolean(false);
     this.readOnlyForBatchOnlyStoreEnabled = storeConfig.isReadOnlyForBatchOnlyStoreEnabled();
 
     this.storeBufferService = builder.getStoreBufferService();
@@ -555,10 +564,7 @@ public abstract class StoreIngestionTask implements Runnable, Closeable {
    */
   public synchronized void subscribePartition(PubSubTopicPartition topicPartition, boolean isHelixTriggeredAction) {
     throwIfNotRunning();
-    partitionToPendingConsumerActionCountMap
-        .computeIfAbsent(topicPartition.getPartitionNumber(), x -> new AtomicInteger(0))
-        .incrementAndGet();
-    consumerActionsQueue.add(new ConsumerAction(SUBSCRIBE, topicPartition, nextSeqNum(), isHelixTriggeredAction));
+    processConsumerActionOrEnqueue(new ConsumerAction(SUBSCRIBE, topicPartition, nextSeqNum(), isHelixTriggeredAction));
   }
 
   public synchronized CompletableFuture<Void> unSubscribePartition(PubSubTopicPartition topicPartition) {
@@ -572,13 +578,9 @@ public abstract class StoreIngestionTask implements Runnable, Closeable {
       PubSubTopicPartition topicPartition,
       boolean isHelixTriggeredAction) {
     throwIfNotRunning();
-    partitionToPendingConsumerActionCountMap
-        .computeIfAbsent(topicPartition.getPartitionNumber(), x -> new AtomicInteger(0))
-        .incrementAndGet();
     ConsumerAction consumerAction =
         new ConsumerAction(UNSUBSCRIBE, topicPartition, nextSeqNum(), isHelixTriggeredAction);
-
-    consumerActionsQueue.add(consumerAction);
+    processConsumerActionOrEnqueue(consumerAction);
     return consumerAction.getFuture();
   }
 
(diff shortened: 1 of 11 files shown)


FelixGV:
Are you still working on this?

haoxu07:
> Are you still working on this?

Yeah, about to make it pubic to review.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -618,9 +621,7 @@ public boolean hasAnySubscription() {
   public synchronized void resetPartitionConsumptionOffset(PubSubTopicPartition topicPartition) {
     throwIfNotRunning();
     amplificationFactorAdapter.execute(topicPartition.getPartitionNumber(), subPartition -> {
-      partitionToPendingConsumerActionCountMap.computeIfAbsent(subPartition, x -> new AtomicInteger(0))
-          .incrementAndGet();
-      consumerActionsQueue.add(
+      processConsumerActionOrEnqueue(
           new ConsumerAction(
               RESET_OFFSET,
```
lluwm, line 626: In old implementation, there seems to have some special handling of `RESET_OFFSET` in StoreIngestionTask.internalClose(), in case that StoreIngestionTask is killed, but `RESET_OFFSET` is in queue but not yet processed. Shall we also handle this case?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -1636,6 +1642,58 @@ private void processConsumerActions(Store store) throws InterruptedException {
     }
   }
 
+  protected void processConsumerActionOrEnqueue(ConsumerAction action) {
+    if (action == null) {
+      return;
+    }
+    if (serverConfig.isProcessConsumerActionWithoutEnqueueEnabled()) {
```
lluwm, line 1649: do you mean `if ( ! serverConfig.isProcessConsumerActionWithoutEnqueueEnabled())`?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -327,6 +330,10 @@ public abstract class StoreIngestionTask implements Runnable, Closeable {
   private final Runnable runnableForKillIngestionTasksForNonCurrentVersions;
   protected final AtomicBoolean recordLevelMetricEnabled;
 
+  protected final ResourceAutoClosableLockManager partitionLockManager;
+
+  private boolean killedActionReceived = false;
```
lluwm, line 335: Shall we declare `killedActionReceived` to be `AtomicBoolean` or `volatile`?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -2217,7 +2275,9 @@ public void processConsumerRecord(
       if (needToUnsub) {
         errorMessage += ". Consumption will be halted.";
         statusReportAdapter.reportError(Collections.singletonList(partitionConsumptionState), errorMessage, e);
-        unSubscribePartition(new PubSubTopicPartitionImpl(versionTopic, faultyPartition));
+        // As unsubscription needs to wait for all records drilled and then proceed, wait here will cause endless wait.
+        Runnable runnable = () -> unSubscribePartition(new PubSubTopicPartitionImpl(versionTopic, faultyPartition));
+        CompletableFuture.runAsync(runnable);
```
lluwm, line 2280: Endless wait is not good, but if we run it async, it has a chance that the task could be run very late, maybe even after the ingestion task is killed. I don't know if that could cause some issues

huangminchn: It's great that you take this part into consideration and did a thorough check! Just to double check my understanding:
Previously the unsubscribe is also done asynchronously by the SIT thread; with the current change, if the new feature is on, you are concerned that the drainer thread will instead do the unsubscribe work for SIT and slow down drainer? If so, can we only use the common pool in Executor when the new feature flag is on?

clients/da-vinci-client/src/test/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTaskTest.java
```diff
@@ -2071,37 +2070,6 @@ public void testKillConsumption(AAConfig aaConfig) throws Exception {
     }
   }
 
-  @Test(dataProvider = "aaConfigProvider")
-  public void testKillActionPriority(AAConfig aaConfig) throws Exception {
```
huangminchn, line 2075: Is it possible to set "SERVER_PROCESS_CONSUMER_ACTION_WITHOUT_ENQUEUE_ENABLE" as false and keep this test?

clients/da-vinci-client/src/test/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTaskTest.java
```diff
@@ -2004,14 +2003,14 @@ public void testCompleteCalledWhenUnsubscribeAfterBatchPushDisabled(AAConfig aaC
 
     runTest(Utils.setOf(PARTITION_FOO), () -> {
       Store mockStore = mock(Store.class);
-      storeIngestionTaskUnderTest.unSubscribePartition(fooTopicPartition);
```
huangminchn, line 2007: If we set "SERVER_PROCESS_CONSUMER_ACTION_WITHOUT_ENQUEUE_ENABLE" to false and keep the unsubscribe action here, will the test pass? Just want to double check, we can make this test work for the new config.

clients/da-vinci-client/src/test/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTaskTest.java
```diff
@@ -2144,21 +2112,29 @@ public void testDataValidationCheckPointing(SortedInput sortedInput, AAConfig aa
...
-            relevantPartitions.stream()
-                .forEach(
-                    partition -> storeIngestionTaskUnderTest
-                        .unSubscribePartition(new PubSubTopicPartitionImpl(pubSubTopic, partition)));
-            relevantPartitions.stream()
-                .forEach(
-                    partition -> storeIngestionTaskUnderTest
-                        .subscribePartition(new PubSubTopicPartitionImpl(pubSubTopic, partition)));
+
+            Thread infiniteLockThread =
```
huangminchn, line 2135: This is to simulate the async process, right? But why name it as infiniteLockThread?

clients/da-vinci-client/src/test/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTaskTest.java
```diff
@@ -2190,10 +2166,8 @@ public void testDataValidationCheckPointing(SortedInput sortedInput, AAConfig aa
       relevantPartitions.stream().forEach(partition -> {
         PubSubTopicPartition pubSubTopicPartition =
             new PubSubTopicPartitionImpl(pubSubTopicRepository.getTopic(topic), partition);
-        verify(mockLocalKafkaConsumer, timeout(LONG_TEST_TIMEOUT).atLeast(totalNumberOfConsumptionRestarts + 1))
+        verify(mockLocalKafkaConsumer, timeout(LONG_TEST_TIMEOUT).atLeast(1))
             .subscribe(eq(pubSubTopicPartition), anyLong());
-        verify(mockLocalKafkaConsumer, timeout(LONG_TEST_TIMEOUT).atLeast(totalNumberOfConsumptionRestarts))
-            .unSubscribe(eq(pubSubTopicPartition));
```
huangminchn, line 2196: The previous unsubscribe action is also done in the SIT thread, why that can complete but not with the "infiniteLockThread"?

huangminchn (changes requested):
Thanks for the hard word, Hao!

haoxu07:
Based on discussion, we decide to stop removing CAQ and decide to do PR https://github.com/linkedin/venice/pull/1034 first.

PR #993, opened 2024-05-17
[server] clean up DIV metrics and fix broken ones

adamxchen:
## Summary

1. This PR fixes a bug in DIV metrics. It mistakenly used nanoseconds for a method only take milliseconds.
2. Moved some ingestion latency metrics from `DIVStats` to `IngestionStats`, since they are not relevant to DIV.
3. Added a new drainer DIV completion time metrics to give us finer view on internal processing stage.
4. Made some changes to `DIVStatsGauge` so it follows the same style as `IngestionStatsGauge`
5. Fix a bug that `producer_callback_latency ` mistakenly uses the wrong start time so its value is not reliable actually

## How was this PR tested?

GHCI


clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -2255,7 +2255,7 @@ protected DelegateConsumerRecordResult delegateConsumerRecord(
         versionedDIVStats.recordLeaderDIVCompletionTime(
             storeName,
             versionNumber,
-            LatencyUtils.getElapsedTimeInMs(beforeProcessingPerRecordTimestampNs),
+            LatencyUtils.getLatencyInMS(beforeProcessingPerRecordTimestampNs),
```
FelixGV, line 2258: These two function names are too confusing... We should think of a more clear name... and rename them.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -2800,6 +2800,11 @@ private int internalProcessConsumerRecord(
         }
         if (recordLevelMetricEnabled.get()) {
           versionedDIVStats.recordSuccessMsg(storeName, versionNumber);
+          versionedDIVStats.recordDrainerDIVCompletionTime(
+              storeName,
+              versionNumber,
+              LatencyUtils.getElapsedTimeInMs(currentTimeMs),
+              currentTimeMs);
```
FelixGV, line 2807: Is this useful? We used to have a DIV latency metric and we removed it since a profiling showed that DIV itself was taking less CPU time than the metric recording time... In other words... it was such a vanishingly small number that it did not matter to instrument it. Unless we're planning to add significant complexity to the DIV logic and fear that its runtime will increase significantly, then I would like to avoid adding one more hot path metric, of which we already have too many...

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/DIVStats.java
```diff
@@ -12,21 +12,17 @@
...
 public class DIVStats {
   private final MetricConfig metricConfig = new MetricConfig();
-  private final WritePathLatencySensor producerSourceBrokerLatencySensor;
-  private final WritePathLatencySensor sourceBrokerLeaderConsumerLatencySensor;
-  private final WritePathLatencySensor producerLocalBrokerLatencySensor;
-  private final WritePathLatencySensor localBrokerFollowerConsumerLatencySensor;
-  private final WritePathLatencySensor leaderProducerCompletionLatencySensor;
+
   private final WritePathLatencySensor leaderDIVCompletionLatencySensor;
+  private final WritePathLatencySensor drainerDIVCompletionLatencySensor;
```
FelixGV, line 17: Instead of adding one more DIV latency sensor, I would suggest deleting the other one.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/DIVStatsReporter.java
```diff
@@ -21,53 +21,46 @@ public DIVStatsReporter(MetricsRepository metricsRepository, String storeName) {
...
+        new DIVStatsGauge(this, () -> sensorFunction.apply(getStats()).getAvg(), sensorBaseName + "_latency_avg_ms"));
     registerSensor(
-        new DIVStatsCounter(sensorBaseName + "_latency_max_ms", this, () -> sensorFunction.apply(getStats()).getMax()));
+        new DIVStatsGauge(this, () -> sensorFunction.apply(getStats()).getMax(), sensorBaseName + "_latency_max_ms"));
   }
 
-  protected static class DIVStatsCounter extends AsyncGauge {
-    DIVStatsCounter(String metricName, DIVStatsReporter reporter, DoubleSupplier supplier) {
+  protected static class DIVStatsGauge extends AsyncGauge {
+    DIVStatsGauge(DIVStatsReporter reporter, DoubleSupplier supplier, String metricName) {
```
FelixGV, line 63: Is this change useful? I believe it will cause a rename of metrics, right? While we might not have dashboards built on these particular metrics, we should still avoid changing metric names unless there is a good reason for it...

And likewise for the parameter order change, is it useful? If the new order is much cleaner, then sure, we can, but if it's equivalent, then IDK if it's worth muddying the understanbility of the git history for it?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/IngestionStats.java
```diff
@@ -78,6 +78,13 @@ public class IngestionStats {
   private final Int2ObjectMap<Sensor> regionIdToHybridBytesConsumedSensorMap;
   private final Int2ObjectMap<Sensor> regionIdToHybridRecordsConsumedSensorMap;
   private final Int2ObjectMap<Sensor> regionIdToHybridAvgConsumedOffsetSensorMap;
+
+  // write path latency sensors
+  private final WritePathLatencySensor producerSourceBrokerLatencySensor;
+  private final WritePathLatencySensor sourceBrokerLeaderConsumerLatencySensor;
+  private final WritePathLatencySensor producerLocalBrokerLatencySensor;
+  private final WritePathLatencySensor localBrokerFollowerConsumerLatencySensor;
+  private final WritePathLatencySensor leaderProducerCompletionLatencySensor;
```
FelixGV, line 87: I think it makes sense to have those here... that's better organized than having them inside DIV stats... thanks.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/IngestionStatsReporter.java
```diff
@@ -97,24 +98,6 @@ protected void registerStats() {
             0,
             VERSION_TOPIC_END_OFFSET_REWIND_COUNT));
 
-    // System store mostly operates on hybrid partial updates so batch metrics are not useful.
-    if (!VeniceSystemStoreUtils.isUserSystemStore(storeName)) {
```
FelixGV, line 101: So it is intentional to add these metrics back for sys stores, right?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/IngestionStatsReporter.java
```diff
@@ -295,6 +303,21 @@ protected void registerConditionalStats() {
...
+    registerSensor(
+        new IngestionStatsGauge(
+            this,
+            () -> sensorFunction.apply(getStats()).getAvg(),
+            sensorBaseName + "_latency_avg_ms"));
+    registerSensor(
+        new IngestionStatsGauge(
+            this,
+            () -> sensorFunction.apply(getStats()).getAvg(),
+            sensorBaseName + "_latency_max_ms"));
```
FelixGV, line 318: We pass avg into the max?

FelixGV (commented):
Left several comments, including many minor ones... please take a look.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
adamxchen: @huangminchn recently had some investigation and suspect the DIV in drainer might be the bottleneck and that's why we are adding this here. 
@huangminchn  can you please chime in and see if it's necessary?

huangminchn: Yes, our investigation showed that the drainer spent ~4+ ms to process one single message in one of the certification chaos test; rocksDB put only took a few micro seconds, and there was not too many rocksDB compaction happening; GC looked good. The only bottleneck I could think of is DIV processing; I would like to confirm this.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/DIVStats.java
huangminchn: It's okay to keep only the drainer one. For messages that go through producer, we actually run DIV twice for the same message, not sure if we should call that a tech debt or bug. In any case, keeping only the drainer DIV latency should cover all cases.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/DIVStatsReporter.java
huangminchn: Yeah, it will change the metric name.
Checking whether these metrics are used in alert dashboard is tedious but doable. Use the metric platform command to list all our alert dashboard and grep for "DIVStatsCounter"

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/IngestionStatsReporter.java
huangminchn: No, the new location of this code still has the same condition check; it's in line 140/123.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/DIVStatsReporter.java
adamxchen: I changed it because I want to have some consistency for `DIVStatsGauge` and `IngestionStatsGauge`. They are most frequently used subclasses of `AsyncGauge` and they have a totally reversed parameter order, which looks odd to me. Since I also moved some metrics from `DIVStatsReporter` to `IngestionStatsReporter`, I think it's better to make them consistent

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/DIVStats.java
adamxchen: plz see [my comment](https://github.com/linkedin/venice/pull/993#discussion_r1607339060) above to see if it makes sense to you.. I'd like to keep the one in the consumer thread too.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/DIVStatsReporter.java
adamxchen: And yes I've checked, we haven't set up any alerts on DIV metrics yet so there's no impact on the dashboard.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
adamxchen: i am going to rename `getLatencyInMS` to `getElapsedTimeFromNSToMS` since it takes nano-seconds as the input and `getElapsedTimeInMs` to `getElapsedTimeFromMsToMs` as it takes milli-seconds as the input. Both of methods have output in MS.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/IngestionStatsReporter.java
adamxchen: Thanks for catching it! Corrected!

adamxchen:
> Looks good overall! Thanks Adam. Please take a look at the comments from Felix, especially the naming changes and passing "avg" stat into "max" sensor.

@huangminchn  @FelixGV  please review again when you have time. Due to the renaming, there are a lot of files involved. They were done by Intellij refactor feature so they should be safe changes.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/DIVStats.java
huangminchn: SGTM! No concerns.

huangminchn (changes requested):

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
```diff
@@ -533,7 +533,7 @@ protected void processMessageAndMaybeProduceToKafka(
               mergeConflictResult.getRmdRecord());
         }
         CompletableFuture.allOf(viewWriterFutures).whenCompleteAsync((value, exception) -> {
-          hostLevelIngestionStats.recordViewProducerLatency(LatencyUtils.getLatencyInMS(preprocessingTime));
+          hostLevelIngestionStats.recordViewProducerLatency(LatencyUtils.getElapsedTimeFromNSToMS(preprocessingTime));
```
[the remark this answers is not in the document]
adamxchen: Yep! Looks like it! Good eyes. I will check and correct them

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
FelixGV: I don't understand... does this metric measure just the DIV execution time? Or does it measure some other scope of work (of which the last step is DIV completion, perhaps?)...? If it is the former, then I don't believe there is value in that metric, as I stated above. If it is the latter, that may be fine, but then we should rename the metric to more accurately match what is it that it is measuring.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/DIVStats.java
FelixGV: There should be no difference in implementation between the DIV that runs in the drainer and the one in the consumer... so we should not expect them to have significantly different latency (other than due to noise, e.g. a GC pause that strikes at exactly the right moment to get counted during DIV's execution latency...). So if there is any hunch that DIV execution might be slow, then let's take a look at the existing metric we have for the leader consumption DIV... My expectation is that we will reconfirm that DIV is extremely fast and not worth having a metric for (but I could be wrong, of course). In any case, I still don't see the point of adding one more DIV metric.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
FelixGV: Ok SG, thanks.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/DIVStats.java
adamxchen: > so we should not expect them to have significantly different latency

That's a valid point. The reason why we would like to have two DIV metrics is we want to eliminate the noise and have high fidelity metrics data since data would be more accurate when recording happens in the same thread.

There are some nuances between these two metrics. Original intention of introducing DIV completion metrics is to measure the latency from the start of processing a record until the DIV validation completes (not only for DIV process alone)

For, `leader completion DIV`, we have some operations that's unrelated to DIV happen first, for example, processing SOP, checking schema id availability and other validation work. `drainer completion DIV` is much simpler process and only DIV validation happens, so they are not identical.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
adamxchen: That's the latter actually! The naming isn't very good.. I will rename them!

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -2285,7 +2285,7 @@ protected DelegateConsumerRecordResult delegateConsumerRecord(
             consumerRecord,
             isEndOfPushReceived,
             partitionConsumptionState);
-        versionedDIVStats.recordLeaderDIVCompletionTime(
+        versionedDIVStats.recordLeaderProcessToDIV(
```
FelixGV, line 2288: Thanks for iterating on this metric name as well.

So the scope of this metric is to perform various metadata checks and bookkeeping, beginning from the start of taking a record out of the consumed batch, up until just before the point where we'll do something more significant with it (e.g. pass it to the drainer for local persistence or to the producer). The one thing which may take a significant time during this process appears to be the `SIT::waitReadyToProcessRecord` function, which will block until the availability of any schema that has never been seen yet, and then attempt to use that schema to deserialize and potentially decompress the incoming payload. If this is the phenomenon we're interested in, then I wonder if we should add a metric just for this, rather than have a metric which covers several loosely defined steps...

FelixGV (commented):
Thanks for iterating Adam. I still feel like the two "DIV" metrics are of dubious value. Please take a look at my latest comments. We can chat offline as well if you'd like.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -2826,7 +2826,7 @@ private int internalProcessConsumerRecord(
         }
         if (recordLevelMetricEnabled.get()) {
           versionedDIVStats.recordSuccessMsg(storeName, versionNumber);
-          versionedDIVStats.recordDrainerDIVCompletionTime(
+          versionedDIVStats.recordDrainerProcessToDIV(
```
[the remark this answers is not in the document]
adamxchen: Synced up offline with @FelixGV and @huangminchn .

For now, rename metrics to be `leader_preprocessing_latency` , which measures multiple validation and DIV latency, and `internal_preprocessing_latency`, which measures mainly DIV latency for now, such that they are intuitive to understand.

Per @huangminchn , we need `internal_preprocessing_latency` as a concrete evidence to confirm whether DIV currently is the bottleneck for a special case. We can remove it if we find it not useful. I also plan to remove more metrics from the hot path as a follow-up PR.

huangminchn (changes requested):
Thanks Adam!

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/IngestionStatsReporter.java
```diff
@@ -172,6 +174,20 @@ protected void registerStats() {
...
+              () -> getStats().getLeaderPreprocessingLatencyMax(),
+              0,
+              LEADER_PREPROCESSING_LATENCY + "_max"));
+
+      registerSensor(
+          new IngestionStatsGauge(
+              this,
+              () -> getStats().getInternalPreprocessingLatencyMax(),
+              0,
+              INTERNAL_PREPROCESSING_LATENCY + "_max"));
```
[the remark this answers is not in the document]
adamxchen: Added

huangminchn (approved):
Thanks a lot Adam!

PR #994, opened 2024-05-17
[controller] Check only parent colo store defered swap config

majisourav99:
## Check only parent colo store defered swap config

During batch pushes controller queries all child controller to fetch store config to validate the deferred version swap flag. This could be flaky if the some colo fails to respond in timely manner (it does 5 retries). Since the parent colo has the required store config, this PR checks only that instead of doing a expensive remote call to the child controller.

## How was this PR tested?

GHCI


huangminchn (approved):
Thanks a lot Sourav! LGTM.

PR #995, opened 2024-05-20
[client][internal] Fix tests related to PR#975

mynameborat:
## Summary
- Fix tests failures due to #975 
- Extracting the test fixes from #987 
- Reset storage node read quota enabled flag prior to each test in `FastClientIndividualFeatureConfigurationTest`

## How was this PR tested?
ran the failing tests locally in intellij and terminal 


xunyin8:
As we discussed offline `testSNQuotaNotEnabled` is the trouble maker here together with the fact that the SN read quota flag is only set during `setUp()` which is only annotated with `BeforeClass`. We need to reset the flag after each test.

adamxchen:
They were not failing in the previous tests because they are flaky, correct?

xunyin8 (approved):

mynameborat:
> They were not failing in the previous tests because they are flaky, correct?

@adamxchen There is some flakiness around `testSNQuotaNotEnabled` depending on the order of execution of tests in the class.

xunyin8 (approved):

PR #996, opened 2024-05-20
[davinci] PartitionTracker code cleanup

lluwm:
This change cleans up the code in PartitionTracker based on IDE warnings and there should have no semantics or functionality modifications.

## How was this PR tested?
passed CI tests.

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/validation/PartitionTracker.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/validation/PartitionTracker.java
@@ -46,13 +46,13 @@ import org.apache.logging.log4j.Logger;
  * This class maintains state about all the upstream producers for a given partition.
  * It keeps track of the last segment, last sequence number and incrementally computed
  * checksum for each producer (identified by a producer GUID).
- *
+ * <p>
  * This class is thread safe. Locking is at the granularity of producers. Multiple
  * threads can process records from the same partition concurrently.
- *
+ * <p>
  * This class also encapsulates the capability to clear expired state, in the functions
  * which take in the maxAgeInMs parameter:
- *
+ * <p>
  * - {@link #clearExpiredStateAndUpdateOffsetRecord(OffsetRecord, long)}
  * - {@link #setPartitionState(OffsetRecord, long)}
  */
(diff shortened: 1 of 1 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/validation/PartitionTracker.java
```diff
@@ -567,17 +566,15 @@ private void validateSequenceNumber(
         logger.error(
             "Encountered missing data message within the log compaction time window. Error msg: {}",
             dataMissingException.getMessage());
-        if (errorMetricCallback.isPresent()) {
-          errorMetricCallback.get().execute(dataMissingException);
-        }
+        errorMetricCallback.ifPresent(divErrorMetricCallback -> divErrorMetricCallback.execute(dataMissingException));
```
FelixGV, line 569: Nitpick: this style of code instantiates a closure object, while the original form does not. Since DIV is used on the hot path, we do care about minimizing allocation, and so I think it's better to use a more traditional code style, even if slightly more verbose...

clients/da-vinci-client/src/main/java/com/linkedin/davinci/validation/PartitionTracker.java
```diff
@@ -604,9 +601,7 @@ public void checkMissingMessage(
       logger.error(
           "Encountered a missing segment. This is unacceptable even if log compaction kicks in. Error msg: {}",
           missingSegment.getMessage());
-      if (errorMetricCallback.isPresent()) {
-        errorMetricCallback.get().execute(missingSegment);
-      }
+      errorMetricCallback.ifPresent(divErrorMetricCallback -> divErrorMetricCallback.execute(missingSegment));
```
FelixGV, line 604: Same nit as above...

FelixGV (commented):
Thanks Lei! Looks good overall. I just left two minor nitpicks...

clients/da-vinci-client/src/main/java/com/linkedin/davinci/validation/PartitionTracker.java
lluwm: This is a good point, will change it back. Thanks for reviewing this change, Felix!

clients/da-vinci-client/src/main/java/com/linkedin/davinci/validation/PartitionTracker.java
lluwm: Sure.

FelixGV (approved):
Thanks Lei!

PR #999, opened 2024-05-22
[vpj] Fix file system on spark push job

modithah:

The current spark job is incapable of reading from other file systems because it takes the default file system. This adds the fixes on such cases enabling to read from any file system. Related to https://github.com/linkedin/venice/pull/711

## How was this PR tested?

Tested on s3 which fails on the current code with error

```
java.lang.IllegalArgumentException: Wrong FS: s3a://.../.../a.avro, expected: file:///
```

Which now succeeds


FelixGV (approved):
Nice, thanks a lot!

PR #1002, opened 2024-05-25
[controller] Muted exception from a log in VHA::hasFatalDataValidationError

FelixGV:
There is only one code path calling this function and it is part of deleting a store-version. Therefore, it is expected that a store-version would not be found and there is no point in logging an exception for it.

## How was this PR tested?
GHCI.

### DIFF
--- a/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
@@ -3325,7 +3325,7 @@ public class VeniceHelixAdmin implements Admin, StoreCleaner {
       OfflinePushStatus offlinePushStatus = pushMonitor.getOfflinePushOrThrow(topicName);
       return offlinePushStatus.hasFatalDataValidationError();
     } catch (VeniceException e) {
-      LOGGER.warn("Failed to get offline push status for topic: {}", topicName, e);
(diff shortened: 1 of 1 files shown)


services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -3325,7 +3325,7 @@ boolean hasFatalDataValidationError(PushMonitor pushMonitor, String topicName) {
       OfflinePushStatus offlinePushStatus = pushMonitor.getOfflinePushOrThrow(topicName);
       return offlinePushStatus.hasFatalDataValidationError();
     } catch (VeniceException e) {
-      LOGGER.warn("Failed to get offline push status for topic: {}", topicName, e);
+      LOGGER.warn("Failed to get offline push status for topic: {}. It might not exist anymore.", topicName);
```
lluwm, line 3328: Felix, thanks for fixing the exception logging. I agree that the printout of the stack trace in this call path is unnecessary.

lluwm (approved):

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
FelixGV: Thanks for reviewing!

PR #1003, opened 2024-05-28
[server] Add retries for topic existence checks in SIT

sushantmane:
## Add retries for topic existence checks in SIT
Currently, ingestion errors out if an exception is thrown during topic existence checks, even for transient issues. This PR adds retries to mitigate this problem. Additionally, it includes a check to skip the compaction enabled verification for RT topics in SIT::validateMessage.

### DIFF
--- a/internal/venice-common/src/main/java/com/linkedin/venice/pubsub/api/PubSubMessage.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/pubsub/api/PubSubMessage.java
@@ -35,6 +35,10 @@ public interface PubSubMessage<K, V, OFFSET> {
     return getTopicPartition().getPubSubTopic().getName();
   }
 
+  default PubSubTopic getTopic() {
+    return getTopicPartition().getPubSubTopic();
+  }
+
   default int getPartition() {
     return getTopicPartition().getPartitionNumber();
   }
--- a/internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicMetadataFetcher.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicMetadataFetcher.java
@@ -22,6 +22,7 @@ import com.linkedin.venice.pubsub.api.PubSubMessage;
 import com.linkedin.venice.pubsub.api.PubSubMessageDeserializer;
 import com.linkedin.venice.pubsub.api.PubSubTopic;
 import com.linkedin.venice.pubsub.api.PubSubTopicPartition;
+import com.linkedin.venice.pubsub.api.exceptions.PubSubClientRetriableException;
 import com.linkedin.venice.pubsub.api.exceptions.PubSubOpTimeoutException;
 import com.linkedin.venice.pubsub.api.exceptions.PubSubTopicDoesNotExistException;
 import com.linkedin.venice.stats.StatsErrorCode;
@@ -187,8 +188,9 @@ class TopicMetadataFetcher implements Closeable {
       if (containsTopicCached(pubSubTopicPartition.getPubSubTopic())) {
         return;
       }
-    } catch (Exception e) {
-      // in case there is any exception, we will retry the operation
+    } catch (PubSubClientRetriableException e) {
+      // in case there is any retriable exception, we will retry the operation
+      LOGGER.debug("Failed to check if topic exists: {}", pubSubTopicPartition.getPubSubTopic(), e);
     }
 
     boolean topicExists = RetryUtils.executeWithMaxAttempt(
--- a/internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicMetadataFetcher.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicMetadataFetcher.java
@@ -187,11 +187,14 @@ class TopicMetadataFetcher implements Closeable {
     if (pubSubTopicPartition.getPartitionNumber() < 0) {
       throw new IllegalArgumentException("Invalid partition number: " + pubSubTopicPartition.getPartitionNumber());
     }
-    ValueAndExpiryTime<Boolean> cachedValue = topicExistenceCache.get(pubSubTopicPartition.getPubSubTopic());
-    // if the topic exists based on the cache, return
-    if (cachedValue != null && cachedValue.getValue()) {
-      return;
+    try {
+      if (containsTopicCached(pubSubTopicPartition.getPubSubTopic())) {
+        return;
+      }
+    } catch (Exception e) {
+      // in case there is any exception, we will retry the operation
     }
+
     boolean topicExists = RetryUtils.executeWithMaxAttempt(
         () -> containsTopic(pubSubTopicPartition.getPubSubTopic()),
         3,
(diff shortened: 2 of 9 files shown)


internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicMetadataFetcher.java
```diff
@@ -183,9 +183,14 @@ void validateTopicPartition(PubSubTopicPartition pubSubTopicPartition) {
...
     }
-    if (containsTopicCached(pubSubTopicPartition.getPubSubTopic())) {
-      return;
+    try {
+      if (containsTopicCached(pubSubTopicPartition.getPubSubTopic())) {
+        return;
+      }
+    } catch (Exception e) {
+      // in case there is any exception, we will retry the operation
     }
```
mynameborat, line 192: log the exception here. It doesn't look like a highly perf sensitive code path as we already have cache and unless there is something seriously wrong with the fetcher flow.

IMO, The retries without any information on why we do retry and how long is going to be hard to debug.

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/api/PubSubMessage.java
```diff
@@ -35,6 +35,10 @@ default String getTopicName() {
     return getTopicPartition().getPubSubTopic().getName();
   }
 
+  default PubSubTopic getTopic() {
+    return getTopicPartition().getPubSubTopic();
+  }
+
```
mynameborat, line 41: Is this really needed? Looks like there are few methods that already seem redundant without any concrete need. e.g., 

`getTopicName` and `getTopic`; ideally if have a way to get topic information everything about the topic should be abstracted and shouldn't be exposed as another getter.

Similar to pattern is also followed for partition where we have `getTopicPartition` and `getPartition` which are both operating on the object TopicPartition. 

I don't see a benefit why callers can't do this indirection directly as opposed exposing it through a getter (didn't notice any concrete need to be overridden or implemented by the implementors of the interface).

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicMetadataFetcher.java
```diff
@@ -183,9 +183,14 @@ void validateTopicPartition(PubSubTopicPartition pubSubTopicPartition) {
...
       throw new IllegalArgumentException("Invalid partition number: " + pubSubTopicPartition.getPartitionNumber());
     }
-    if (containsTopicCached(pubSubTopicPartition.getPubSubTopic())) {
-      return;
+    try {
+      if (containsTopicCached(pubSubTopicPartition.getPubSubTopic())) {
+        return;
+      }
+    } catch (Exception e) {
+      // in case there is any exception, we will retry the operation
```
mynameborat, line 191: Shouldn't we classify what is recoverable vs irrecoverable exceptions here to deem that we can succeed in retrying the same flow? Is that possible?

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicMetadataFetcher.java
sushantmane: >log the exception here. It doesn't look like a highly perf sensitive code path as we already have cache and unless there is something seriously wrong with the fetcher flow.

I don't think this is good idea since this method is called frequently. 


>as we already have cache and unless there is something seriously wrong with the fetcher flow.
Could you please expand this?

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/api/PubSubMessage.java
sushantmane: Let's not confuse leaky abstractions with convenience methods. 

- `getTopicName` returns the name of the topic as a `String`.
- `getTopic` returns an instance of type `PubSubTopic`.

The same applies for `PubSubTopicPartition`.

> Is this really needed?

No but it helps in reducing verbose pattern: `record.getTopicPartition().getPubSubTopic()`  to `record.getTopic()` without any downsides. 

_That being said, I do think both `PubSubTopic` and `PubSubTopicPartition` abstractions are unnecessary. Just using `String` and `int` would be sufficient_

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicMetadataFetcher.java
sushantmane: Yeah we already have that distinction. Will update

sushantmane: Done

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/api/PubSubMessage.java
mynameborat: >Let's not confuse leaky abstractions with convenience methods.
I differ here. I don't see these as convenience methods. As far as this interface is considered these are seen as contracts. If at all they are convenience methods, they shouldn't part of an interface as the implementors can choose to evolve and override and cause changes in assumptions of these representations.

It is not in the nature of this interface to hold convenience methods and utility classes are more of a fit. Typically utility classes are final and non extensible enforcing one and only behavior of the method thereby making it easy to rationalize the behavior as opposed overrides making it more complex to identify and reason about the behavior and its whereabouts (especially when its not needed).

Additionally, `TopicPartition` and `PubSubTopic` are the contracts this interface works with and exposes. It is not the inner contracts of how partition is represented within `TopicPartition` and how topic is represented as a string within `PubSubTopic`. I don't see why this interface should be responsible for handling evolutions within its contracts `TopicPartition` and `PubSubTopic`.

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicMetadataFetcher.java
mynameborat: >Could you expand this?

I still don't have full context. But glancing `TopicMetadataFetcher` and `TopicManager`, it seems some callers of this method `validateTopic...` themselves have some sort of retries. That made me come to conclusion that on a happy path, you wouldn't log anything and on a failure scenario, you would log the exception and the cadence of log spam would be tied the retry mechanisms employed by the callers (which is typically, exponential backoff) and thus preventing log spam.

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/api/PubSubMessage.java
sushantmane: >they shouldn't part of an interface as the implementors can choose to evolve and override and cause changes in assumptions of these representations.

Even if this is the case, API by contract needs to return an instance of type `PubSubTopic`.

sushantmane: I believe this can be simplified as follows:

* Each message belongs to a topic.
* Each message is assigned to a specific partition within that topic.
* This info about messages should be easily retrievable from the message itself.

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicMetadataFetcher.java
sushantmane: Your observation is valid; however, the number of failures can often be significant. This is because each node hosts multiple partitions, and this code path is called several times a minute for these partitions, either for metrics collection or other internal bookkeeping. When things go wrong it doesn't take a long time to fill up logs with redundant info.

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/api/PubSubMessage.java
mynameborat: we had a sync offline about this. This was more of an observation and comment on the pattern and lay of the land of the usage in the current class. However, I understand its not within the scope of PR to necessarily act on any of the above discussions.

For the context of the PR, feel free to treat it as a non-blocker conversation and merge it.

mynameborat (approved):

sushantmane:
Thanks a lot, @bharathkk!  🙏  I appreciate all the thorough discussions, and yes, by and large, I think I agree with you.

PR #1005, opened 2024-05-28
[CI] Update some Github Action versions

adamxchen:
## Summary

Our javadoc publish workflow is broken because Github updated the default git version and some commands/workflows are no longer working.
This PR updates some out-of-dated Github Action versions

## How was this PR tested?

CI

https://github.com/adamxchen/venice/actions/runs/9276409190/job/25523408597

### DIFF
--- a/.github/workflows/VeniceCI-CompatibilityTests.yml
+++ b/.github/workflows/VeniceCI-CompatibilityTests.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 60
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         with:
           fetch-depth: 0
       - name: Set up JDK
@@ -58,7 +58,7 @@ jobs:
     outputs:
       alpini_touched: ${{ steps.check_alpini_files_changed.outputs.alpini }}
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         with:
           # Checkout as many commits as needed for the diff
           fetch-depth: 2
@@ -69,7 +69,7 @@ jobs:
           filters: |
             alpini:
               - 'internal/alpini/**'
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         if: steps.check_alpini_files_changed.outputs.alpini == 'true'
         with:
           fetch-depth: 0
@@ -128,7 +128,7 @@ jobs:
     outputs:
       alpini_touched: ${{ steps.check_alpini_files_changed.outputs.alpini }}
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         with:
           # Checkout as many commits as needed for the diff
           fetch-depth: 2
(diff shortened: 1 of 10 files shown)


mynameborat:
Thanks for the fix! 
Wonder why we change the coverage threshold in this PR? I'd suggest doing that in a different PR where we can have more details into why that is pursued and how we landed on the number we chose? e.g., analytics from past PRs on coverage percentage etc

Additionally, keeps the motive of the PR to one thing.

.github/workflows/VeniceCI-CompatibilityTests.yml
```diff
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 60
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
```
mynameborat, line 17: Can we extract out the version into a variable to keep the changes minimal for the future edits? e.g., would something like this https://docs.geoserver.org/2.21.x/en/user/styling/ysld/reference/variables.html help extract out the actions or the version into a variable?

FelixGV: Some of these files are generated by our gradle build, so we could inject the variable there also. In any case, I don't think this changes very often. So IDK if it's worth it 🤷‍♂️

build.gradle
```diff
@@ -516,19 +516,20 @@ subprojects {
...
-          minBranches = project.ext.has('diffCoverageThreshold') ? project.ext.diffCoverageThreshold : 0.40
+          minBranches = project.ext.has('diffCoverageThreshold') ? project.ext.diffCoverageThreshold : 0.45
           failOnViolation = true
         }
       }
 
-      task logDiffCoverage {
+      task logCoverage {
         doLast {
+          parseJacocoXml("$buildDir/reports/jacoco/test/jacocoTestReport.xml")
```
FelixGV, line 526: Just curious, what does this do?

mynameborat: The diff coverage is mostly useful in highlighting the coverage gaps introduced by the changes as part of the PR.
My guess is, this provides the state of coverage for the entire code base based on our tests.

.github/workflows/VeniceCI-CompatibilityTests.yml
mynameborat: ah.. You can ignore the comment for now. I do see there are plenty of workflows and steps that are copied across the yaml we have. Maybe this can be a nice to do clean up to consolidate how these workflows are in general put together and modified.

Perhaps, finding a way to clean up and organize these systematically would be nicer.

build.gradle
adamxchen: Yes, that's correct. `test/jacocoTestReport.xml` tells us the overall status for that module and `diffCoverage/test.xml` tells the coverage for the diff.

Since we don't integrate with some code coverage github action yet (since most of they only support line coverage), I figure it might be good to print out the overall status for people to take a look if they are interested.

FelixGV (approved):
Thanks Adam!

build.gradle
FelixGV: Regarding making the code cov easier to browse online, I was thinking we could have a separate repo to host the jacoco reports, and push these files to that other repo as part of the publish step, then that other repo could render those report files. Just a random idea (out of scope of this PR, obviously...)...

PR #1006, opened 2024-05-28
[CI] bump the code coverage

adamxchen:

This PR bumps the diff code coverage from 40% to 45%.

## Background

We used to have lots of integ tests and few unit tests. Since 2023, we invested some effort to enforce unit test coverage   so people would write more unit tests to catch some bugs that are expensive to test in a integ or cannot be tested by a integ test. The goal was aggressively set to 80% but it hurt devs' velocity to deliver codes since we have too many files that don't have a good test coverage. To get this rolled up slowly, we started with 33% and bumped it to 40% after several months. This PR is a part of effort to continuously raise the bar.

## How was this PR tested?

N/A

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -516,19 +516,20 @@ subprojects {
         }
 
         violationRules {
-          minBranches = project.ext.has('diffCoverageThreshold') ? project.ext.diffCoverageThreshold : 0.40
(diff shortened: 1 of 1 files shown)


mynameborat (approved):
Thanks!

PR #1007, opened 2024-05-29
[controller] Fix a bug in deleteOneStoreVersion

lluwm:
In VHA::deleteOneStoreVersion, there is a bug that trying to get the push status after the push monitor is stopped. It is introduced in PR d7997d782. This PR fixes the issue by calling hasFatalDataValidationError before stopMonitorOfflinePush.

## How was this PR tested?
Passed CI

### DIFF
--- a/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
@@ -3263,7 +3263,11 @@ public class VeniceHelixAdmin implements Admin, StoreCleaner {
       deleteHelixResource(clusterName, resourceName);
       LOGGER.info("Killing offline push for: {} in cluster: {}", resourceName, clusterName);
       killOfflinePush(clusterName, resourceName, true);
+
+      // Check DIV error in the push status before stopping the monitor.
+      boolean hasFatalDataValidationError = hasFatalDataValidationError(pushMonitor, resourceName);
       stopMonitorOfflinePush(clusterName, resourceName, true, isForcedDelete);
+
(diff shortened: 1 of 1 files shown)


majisourav99 (approved):
LGTM

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -3263,16 +3263,19 @@ private void deleteOneStoreVersion(String clusterName, String storeName, int ver
       deleteHelixResource(clusterName, resourceName);
       LOGGER.info("Killing offline push for: {} in cluster: {}", resourceName, clusterName);
       killOfflinePush(clusterName, resourceName, true);
+
+      // Check DIV error in the push status before stopping the monitor.
+      boolean hasFatalDataValidationError = hasFatalDataValidationError(pushMonitor, resourceName);
```
mynameborat, line 3268: Looks like the resources on which the check is made is different across two code path? Is that intended?
The description only calls about moving out the place where this check happens but not necessarily changing the resource itself. e.g., Old one uses `versionTopicName` while this one uses `resourceName`. 

Are both same?

lluwm: Yeah, good point and I also noticed that. Thus, I took a look at `deleteVersionFromStoreRepository` then realized that returned `deletedVersion` is Optional whose version number is either `versionNumber` or `empty`. 

Then, if consider that when we used the returned boolean `hasFatalDataValidationError` , it already passed the condition `deletedVersion.isPresent()`, then both value have to be the same.

lluwm:
Thanks for the review, @majisourav99 and @mynameborat! Let me know if there is still comments on this PR. Otherwise, I will proceed to commit it.

PR #1010, opened 2024-05-31
[router] Delete dictionary for retired versions instead of existing versions

huangminchn:
## Summary
Current there is a bug where future for dictionary downloads of existing versions would be removed, but instead, the intention is to delete for retired versions. This PR fixed the bug.

A major change:
Previously it's the ZK callback thread that sends the requests to retrieve dictionary; now it's the router dictionary processing threads that send the requests and wait for response. This is the only way to make testing feasible and meaningful.

## How was this PR tested?
Added tests to ensure existing versions will not be removed from dictionary download futures during store config change callback.


ZacAttack (dismissed):

services/venice-router/src/test/java/com/linkedin/venice/router/api/TestDictionaryRetrievalService.java
```diff
@@ -3,15 +3,14 @@
 import static com.linkedin.venice.router.api.DictionaryRetrievalService.MAX_DICTIONARY_DOWNLOAD_DELAY_TIME_MS;
 import static com.linkedin.venice.router.api.DictionaryRetrievalService.MIN_DICTIONARY_DOWNLOAD_DELAY_TIME_MS;
 import static org.mockito.ArgumentMatchers.any;
-import static org.mockito.Mockito.anyInt;
-import static org.mockito.Mockito.doReturn;
-import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.*;
```
sushantmane, line 6: could you please replace this with specific imports?

huangminchn: Yup, reverted.

sushantmane (approved):
Thanks, @huangminchn!

PR #1012, opened 2024-06-03
[changelog] Add bootstrap listener implementation to after image changelog consumer

sixpluszero:
## [changelog] Add bootstrap listener implementation to after image changelog consumer
This PR adds bootstrap listener on regular changelog consumer and as well as specific record support.
User can override the listener bootstrap completed notification behavior to perform any action

## How was this PR tested?

Added an integration test to verify the listener is working.


FelixGV (commented):
Thanks Jialin. I left several minor comments. Can you also check the unit test coverage please...?

sixpluszero:
Hi @FelixGV  thanks for the review! As discussed with our customer, we think this listener implementation is too complicated for single user to use and is an overkill. Instead a simple change in #1016 should just work for them. I believe that new PR is fairly easily and has addressed some of the comment you posted - list Java doc. If possible could you take a look at that PR?
I plan to close the current PR now.

PR #923, opened 2024-04-01
[server][da-vinci] Switch to dedicated producers from shared producer in test cases

sushantmane:
## Switch to dedicated producers from shared producer in test cases
Use Dedicated Producers in Test Cases, Remove Shared Producer Code

This commit switches test cases from using a shared producer, which is not
supported in production, to dedicated producers. The shared producer code
has been deleted as it is currently not useful.

## How was this PR tested?
CI

## Does this PR introduce any user-facing changes?

- [x] No. You can skip the rest of this section.
- [ ] Yes. Make sure to explain your proposed changes and call out the behavior change.

PR #1014, opened 2024-06-04
[dvc][server][vpj][producer][controller] Refactored VeniceWriter's close

FelixGV:
The endAllSegments function will now write to all segments in parallel, and wait once at the end, rather than block for each one sequentially.

The close function now honors the venice.writer.close.timeout.ms config strictly. Previously, it could wait longer than this during close.

Deleted the following unused config from the VeniceWriter: venice.writer.sleep.time.ms.when.topic.missing

Deleted unnecessary overloads in ApacheKafkaProducerAdapter. In particular, stopped relying on the flush with timeout API, because it only exists in LinkedIn's fork of Kafka, and thus prevents moving to the Apache fork of Kafka.

Tweaked the synchronization to prevent race conditions when calling close multiple times. It should now be impossible to receive a RejectedExecutionException due to the threadPoolExecutor having been already shutdown.

Miscellaneous:

- TestRestartServerAfterDeletingSstFilesWithActiveActiveIngestion now has more robust resource closing logic.

## How was this PR tested?
GHCI.

See commit message for deleted config and tweaked config semantic.

### DIFF
--- a/internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
@@ -408,6 +408,10 @@ public class VeniceWriter<K, V, U> extends AbstractVeniceWriter<K, V, U> {
 
       return CompletableFuture.supplyAsync(() -> {
         synchronized (closeLock) {
+          if (isClosed) {
+            return VeniceResourceCloseResult.ALREADY_CLOSED;
+          }
+
           logger.info("Closing VeniceWriter for topic: {}, gracefulness: {}", topicName, gracefulClose);
           long timeAtStartOfClose = System.currentTimeMillis();
           try {
@@ -441,7 +445,7 @@ public class VeniceWriter<K, V, U> extends AbstractVeniceWriter<K, V, U> {
              * N.B.: We call the overload with a topic name param here for the sake of shared producer...
              * TODO: Decide whether to keep shared producer or not...
              */
-            producerAdapter.close(topicName, timeLeftToClose);
+            producerAdapter.close(timeLeftToClose);
             OPEN_VENICE_WRITER_COUNT.decrementAndGet();
           } catch (Exception e) {
             handleExceptionInClose(e, gracefulClose, retryOnGracefulCloseFailure);
@@ -465,7 +469,7 @@ public class VeniceWriter<K, V, U> extends AbstractVeniceWriter<K, V, U> {
     try {
       if (gracefulClose && retryOnGracefulCloseFailure) {
         logger.info("Ungracefully closing the VeniceWriter for topic: {}, closeTimeOut: 0 ms", topicName);
-        producerAdapter.close(topicName, 0);
+        producerAdapter.close(0);
         OPEN_VENICE_WRITER_COUNT.decrementAndGet();
       }
     } catch (Exception ex) {
@@ -1235,7 +1239,11 @@ public class VeniceWriter<K, V, U> extends AbstractVeniceWriter<K, V, U> {
   public void closePartition(int partition) {
     if (segments[partition] != null) {
       logger.debug("Closing partition: {} in VeniceWriter.", partition);
-      endSegment(partition, true);
+      try {
+        endSegment(partition, true).get();
+      } catch (InterruptedException | ExecutionException e) {
+        handleControlMessageProducingException(e, ControlMessageType.END_OF_SEGMENT);
+      }
     }
   }
 
--- a/internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
@@ -56,7 +56,6 @@ import com.linkedin.venice.utils.DaemonThreadFactory;
 import com.linkedin.venice.utils.ExceptionUtils;
 import com.linkedin.venice.utils.LatencyUtils;
 import com.linkedin.venice.utils.Time;
-import com.linkedin.venice.utils.Timer;
 import com.linkedin.venice.utils.Utils;
 import com.linkedin.venice.utils.VeniceProperties;
 import com.linkedin.venice.utils.VeniceResourceCloseResult;
@@ -108,8 +107,6 @@ public class VeniceWriter<K, V, U> extends AbstractVeniceWriter<K, V, U> {
       VENICE_WRITER_CONFIG_PREFIX + "replication.metadata.chunking.enabled";
   public static final String MAX_ATTEMPTS_WHEN_TOPIC_MISSING =
       VENICE_WRITER_CONFIG_PREFIX + "max.attemps.when.topic.missing";
-  public static final String SLEEP_TIME_MS_WHEN_TOPIC_MISSING =
-      VENICE_WRITER_CONFIG_PREFIX + "sleep.time.ms.when.topic.missing";
   /**
    * A negative value or 0 will disable the feature.
    */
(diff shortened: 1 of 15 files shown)


internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/pubsub/adapter/kafka/producer/ApacheKafkaProducerAdapterITest.java
```diff
@@ -313,7 +313,7 @@ public void testProducerCloseCanUnblockSendMessageCallerThread(boolean doFlush)
       countDownLatch.await();
       // Still wait for some time to make sure blocking sendMessage is inside kafka before closing it.
       Utils.sleep(50);
-      producerAdapter.close(5000, doFlush);
```
sushantmane, line 316: should we rename `doFlush` to `unboundedTimeout`?

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
```diff
@@ -485,19 +429,32 @@ public CompletableFuture<VeniceResourceCloseResult> closeAsync(
...
+            long timeLeftToClose = 0; // N.B.: 0 indicates an ungraceful close
+            if (gracefulClose) {
+              /** N.B.: We measure the time left to try to adhere strictly to the configured {@link closeTimeOutInMs} */
+              long timeSinceStartOfClose = System.currentTimeMillis() - timeAtStartOfClose;
+              timeLeftToClose = Math.max(0, closeTimeOutInMs - timeSinceStartOfClose);
+            }
+
+            /**
+             * N.B.: We call the overload with a topic name param here for the sake of shared producer...
+             * TODO: Decide whether to keep shared producer or not...
```
sushantmane, line 442: We've [deleted shared producer code](https://github.com/linkedin/venice/pull/923) sometime back. I think we can remove topic name from the close call

sushantmane (commented):
Looks great. Can we also replace `broadcastControlMessage` with the same concurrent approach?

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
```diff
@@ -1760,30 +1717,68 @@ public void sendControlMessage(
...
+    }
+  }
+
+  /**
+   * Main function for sending control messages. Synchronization is most minimal and there is no error checking.
+   *
+   * See also:
+   * {@link #sendControlMessageWithRetriesForNonExistentTopic(ControlMessage, int, Map, PubSubProducerCallback, LeaderMetadataWrapper)}
+   */
+  public CompletableFuture<PubSubProduceResult> sendControlMessage(
```
sushantmane, line 1758: I think we need to ensure that the callers (e.g. broadcastControlMessage) wait on this future to keep the old behavior.

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
```diff
@@ -1994,18 +1989,25 @@ private Segment startSegment(int partition) {
...
+      handleControlMessageProducingException(e, ControlMessageType.END_OF_SEGMENT);
+    }
   }
 
   /**
    * @param partition in which to end the current segment
+   * @return A CompletableFuture which either contains a result if a EOS was sent, or null if none needed to be sent.
    */
-  public void endSegment(int partition, boolean finalSegment) {
+  public CompletableFuture<PubSubProduceResult> endSegment(int partition, boolean finalSegment) {
```
sushantmane, line 2010: There are some places that expect this call to be a blocking call. Should be replace it them with `future.get()`?

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
```diff
@@ -460,19 +405,18 @@ public CompletableFuture<VeniceResourceCloseResult> closeAsync(
       if (isClosed) {
         return CompletableFuture.completedFuture(VeniceResourceCloseResult.ALREADY_CLOSED);
       }
-    }
 
-    return CompletableFuture.supplyAsync(() -> {
-      synchronized (closeLock) {
-        if (isClosed) {
-          return VeniceResourceCloseResult.ALREADY_CLOSED;
-        }
```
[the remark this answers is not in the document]
sushantmane: I think the following check is still there in the code
 `if (isClosed) {
        return CompletableFuture.completedFuture(VeniceResourceCloseResult.ALREADY_CLOSED);
      }
```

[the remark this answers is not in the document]
Comment (lluwm) [reply internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java:469]: Right, but we only set `isClosed = true` when the segments and producer are fully closed in the task thread asynchronously. If there are concurrent closeAsync arrived, then sounds like it is possible that we can have multiple tasks to close the same producer.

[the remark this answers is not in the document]
Comment (sushantmane) [reply internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java:469]: I think we should move back the following check inside ` CompletableFuture.supplyAsync` block and remove outer check 

```
        if (isClosed) {
          return VeniceResourceCloseResult.ALREADY_CLOSED;
        }
```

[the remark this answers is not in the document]
sushantmane: I think we should move back `isClosed` check inside ` CompletableFuture.supplyAsync` block and remove the outer one. 

For RejectedExecutionException I think we can put it inside try-catch block and swallow the exception as double close is not expected anyway

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/pubsub/adapter/kafka/producer/ApacheKafkaProducerAdapterITest.java
FelixGV: Calling `close` with timeout zero is the already-supported way to do the equivalent of `!doFlush`. But I am not sure what your suggestion is? Can you elaborate?

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
FelixGV: Sure, I can do that, then.

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
FelixGV: Sure, I will take a look at each one.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/pubsub/adapter/kafka/producer/ApacheKafkaProducerAdapterITest.java
sushantmane: My suggestion is mainly about renaming the variable to something like `isUnboundedTimeout` in the test method, since we don't have an **explicit** flush anymore in the close method.

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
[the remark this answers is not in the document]
FelixGV: Sure, I don't mind adding back the `if (isClosed)` check inside the closure. I think we do need to keep the outer one, though, because we want to avoid submitting a future into a shutdown `threadPoolExecutor`.

sushantmane (approved):
LGTM! Thanks a lot, @FelixGV !

lluwm (approved):
Thank you and it looks good to me!

PR #1015, opened 2024-06-04
[fast-client] IndexOutOfBoundsException in getReplicas due to bad casting

xunyin8:
## [fast-client] IndexOutOfBoundsException in getReplicas due to bad casting
There is a bug in our getReplicas code for both routing strategies:  HelixScatterGatherRoutingStrategy and LeastLoadedClientRoutingStrategy where we accidentally cast requestId (a long) into an int and this can cause integer overflow and return negative index.

int startPos = (int) requestId % replicaCnt;

instead of 

int startPos = (int) (requestId % replicaCnt);

## How was this PR tested?
New unit tests


nisargthakkar (approved):
LGTM. Great find, Xun

FelixGV (approved):
Thanks for the fix and the unit tests!

PR #1016, opened 2024-06-04
[changelog] Add isCaughtUp() API to VeniceChangelogConsumer interface

sixpluszero:
## [changelog] Add isCaughtUp() API to VeniceChangelogConsumer interface
Instead of implementing a listener as in #1012 , a simpler solution is just to provide an API that user can call to check whether all subscriptions have caught up.
Also, add specific record support for regular consumer

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumer.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumer.java
@@ -198,6 +198,14 @@ public interface VeniceChangelogConsumer<K, V> {
   Collection<PubSubMessage<K, ChangeEvent<V>, VeniceChangeCoordinate>> poll(long timeoutInMs);
 
   /**
+   * Checks whether all subscribed partitions are caught up during bootstrap.
+   * @param offsetThresholdTimestamp Offset timestamp in millisecond. If a partition's (startTimestamp - heartbeatTimestamp)
+   *                                 is smaller or equal to offsetThresholdTimestamp, we consider this partition is caught up.
+   * @return True if all subscribed partitions have caught up.
+   */
+  boolean isCaughtUp(long offsetThresholdTimestamp);
+
+  /**
    * Release the internal resources.
    */
   void close();
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumer.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumer.java
@@ -198,8 +198,8 @@ public interface VeniceChangelogConsumer<K, V> {
   Collection<PubSubMessage<K, ChangeEvent<V>, VeniceChangeCoordinate>> poll(long timeoutInMs);
 
   /**
-   * Checks whether all subscribed partitions are caught up during bootstrap. If a partition's (startTimestamp - heartbeatTimestamp)
-   * is smaller or equal to 1 min (which is 1 Heartbeat stale), we consider this partition is caught up.
+   * Checks whether all subscribed partitions are caught up during bootstrap. If a partition's (currentTimestamp - latestMessageTimestamp)
+   * is smaller or equal to 1 min, we consider this partition is caught up.
    * @return True if all subscribed partitions have caught up.
    */
   boolean isCaughtUp();
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumer.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumer.java
@@ -198,12 +198,11 @@ public interface VeniceChangelogConsumer<K, V> {
   Collection<PubSubMessage<K, ChangeEvent<V>, VeniceChangeCoordinate>> poll(long timeoutInMs);
 
   /**
-   * Checks whether all subscribed partitions are caught up during bootstrap.
-   * @param offsetThresholdTimestamp Offset timestamp in millisecond. If a partition's (startTimestamp - heartbeatTimestamp)
-   *                                 is smaller or equal to offsetThresholdTimestamp, we consider this partition is caught up.
+   * Checks whether all subscribed partitions are caught up during bootstrap. If a partition's (startTimestamp - heartbeatTimestamp)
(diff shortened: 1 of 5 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumer.java
```diff
@@ -197,6 +197,14 @@ public interface VeniceChangelogConsumer<K, V> {
    */
   Collection<PubSubMessage<K, ChangeEvent<V>, VeniceChangeCoordinate>> poll(long timeoutInMs);
 
+  /**
+   * Checks whether all subscribed partitions are caught up during bootstrap.
+   * @param offsetThresholdTimestamp Offset timestamp in millisecond. If a partition's (startTimestamp - heartbeatTimestamp)
+   *                                 is smaller or equal to offsetThresholdTimestamp, we consider this partition is caught up.
```
FelixGV, line 203: I am very confused by this... is this an offset threshold or a timestamp threshold? It's better not to mix these terms in the same param name.

Also, I am confused about `startTimestamp - heartbeatTimestamp`... that would be a negative value, right?

Can we eliminate that param completely?

sixpluszero: Yeah I can update the naming. 
The idea is that the customer wants to be able to know if all the partitions are caught up by comparing HB TS with consumer startup TS. 
The passed in param in the API is a threshold, meaning  startTimestamp - heartbeatTimestamp < threshold we are good to announce bootstrap completed. 

With we continue to ingest, after bootstrap, the number in the map will the negative, but that should be fine as user will not need them anymore

FelixGV: That is confusing, IMHO... let's say I pass zero into this param, but the consumer is quite behind for some reason. An hour after starting up, the consumer finally gets to the point where it is consuming the messages from an hour ago. At that point `isCaughtUp(0)` would finally return true. But is this useful? Why not eliminate the param altogether, and say that `isCaughtUp()` returns true if the most recent timestamp is <1m old, or something like that?

sixpluszero: I get what you mean, you think it is not necessary to have the param but use 1 min as constant.

There is only exactly one use case on this, our current observation is the bootstrap time should be around 1min. But certain partitions have not enough write volume.  The customer want to test with different parameter to see if certain param speed up the bootstrap. They potentially can tolerate messages like 5 mins old or something, so bootstrap can go even a bit faster. @ZacAttack you probably have context to this, maybe could you also chime in on this param thing?

sixpluszero: Also @zx0401 if you want to share anything from client side.

(Because I am ....... kind of convinced by what Felix mentioned above, a constant 1min threshold should be good for caught up logic)

sixpluszero: Checked with the partner, they are fine with constant threshold. That makes this easy. I am going to address the issue and update the PR.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumer.java
```diff
@@ -198,12 +198,11 @@ public interface VeniceChangelogConsumer<K, V> {
   Collection<PubSubMessage<K, ChangeEvent<V>, VeniceChangeCoordinate>> poll(long timeoutInMs);
 
   /**
-   * Checks whether all subscribed partitions are caught up during bootstrap.
-   * @param offsetThresholdTimestamp Offset timestamp in millisecond. If a partition's (startTimestamp - heartbeatTimestamp)
-   *                                 is smaller or equal to offsetThresholdTimestamp, we consider this partition is caught up.
+   * Checks whether all subscribed partitions are caught up during bootstrap. If a partition's (startTimestamp - heartbeatTimestamp)
```
FelixGV, line 201: As I tried to say in my last comment, relying on the `startTimestamp` is confusing... I think the formula should be:

```
current timestamp - timestamp of the last consumed message < 1 minute
```

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumerImpl.java
```diff
@@ -580,6 +579,21 @@ protected Collection<PubSubMessage<K, ChangeEvent<V>, VeniceChangeCoordinate>> i
     return pubSubMessages;
   }
 
+  boolean maybeUpdatePartitionToBootstrapMap(
+      PubSubMessage<KafkaKey, KafkaMessageEnvelope, Long> message,
+      PubSubTopicPartition pubSubTopicPartition,
+      long startTimestamp) {
+    ControlMessage controlMessage = (ControlMessage) message.getValue().getPayloadUnion();
+    if (controlMessage.controlMessageType == START_OF_SEGMENT.getValue()
+        && Arrays.equals(message.getKey().getKey(), KafkaKey.HEART_BEAT.getKey())) {
```
FelixGV, line 588: I don't think we must restrict this to HBs only... any recent-enough message should be able to kick us into the caught up state.

FelixGV (commented):
Thanks for consulting with the partner and for iterating on this. I have some followup comments...

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumer.java
sixpluszero: Our original discussion was concerning about chasing moving target for current TS with data message in less frequently updated partition will lead to slowness in catchup.

But I guess with 1 min fixed constant I guess it is all fine, I can adjust that.

FelixGV: If what we actually want to build is to measure if we caught up to the start time (or to some arbitrary time), then I think we cannot call this API `isCaughtUp`, because the meaning is too different, IMHO...

FelixGV (approved):
Thanks Jialin!

PR #1017, opened 2024-06-04
[router] Set default config value to limit long tail retries to 2

majisourav99:
##  Update config to limit long tail retries to 2

Current default for retry limit per route for long tail retry is -1, which is equivalient to unlimited retries. During high traffic or slow server responses, that could lead to retry storm which will flood the servers with many requests. This PR limits per route maximum retry limit to 2. Also reduces the default threshold for pending connection per server to 100, so that router does not send too many requests to an already slow host.

1. `venice.router.long.tail.retry.max.route.for.multi.keys.req` to 2 to limit retry storm.

2. `venice.router.unhealthy.pending.connection.threshold.per.host` to 100 to protect the server better.

## How was this PR tested?

CI

### DIFF
--- a/services/venice-router/src/main/java/com/linkedin/venice/router/VeniceRouterConfig.java
+++ b/services/venice-router/src/main/java/com/linkedin/venice/router/VeniceRouterConfig.java
@@ -247,7 +247,7 @@ public class VeniceRouterConfig {
     smartLongTailRetryEnabled = props.getBoolean(ROUTER_SMART_LONG_TAIL_RETRY_ENABLED, true);
     smartLongTailRetryAbortThresholdMs = props.getInt(ROUTER_SMART_LONG_TAIL_RETRY_ABORT_THRESHOLD_MS, 100);
     // Default: -1 means this feature is not enabled.
-    longTailRetryMaxRouteForMultiKeyReq = props.getInt(ROUTER_LONG_TAIL_RETRY_MAX_ROUTE_FOR_MULTI_KEYS_REQ, -1);
+    longTailRetryMaxRouteForMultiKeyReq = props.getInt(ROUTER_LONG_TAIL_RETRY_MAX_ROUTE_FOR_MULTI_KEYS_REQ, 2);
     maxKeyCountInMultiGetReq = props.getInt(ROUTER_MAX_KEY_COUNT_IN_MULTIGET_REQ, 500);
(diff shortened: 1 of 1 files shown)


services/venice-router/src/main/java/com/linkedin/venice/router/VeniceRouterConfig.java
```diff
@@ -247,7 +247,7 @@ private void checkProperties(VeniceProperties props) {
     smartLongTailRetryEnabled = props.getBoolean(ROUTER_SMART_LONG_TAIL_RETRY_ENABLED, true);
     smartLongTailRetryAbortThresholdMs = props.getInt(ROUTER_SMART_LONG_TAIL_RETRY_ABORT_THRESHOLD_MS, 100);
     // Default: -1 means this feature is not enabled.
-    longTailRetryMaxRouteForMultiKeyReq = props.getInt(ROUTER_LONG_TAIL_RETRY_MAX_ROUTE_FOR_MULTI_KEYS_REQ, -1);
+    longTailRetryMaxRouteForMultiKeyReq = props.getInt(ROUTER_LONG_TAIL_RETRY_MAX_ROUTE_FOR_MULTI_KEYS_REQ, 2);
```
KaiSernLim, line 250: I wonder if it would be better to set the limit to a higher number, such as 5.

Going from unlimited retries to 2 retries feels a bit forceful. But of course, it all depends on whether there are requests that would've typically resolved after 2+ retries that we would now see producing errors, which I have no idea about. We definitely don't want to encounter another, so maybe 2 is the right number here.

m-nagarajan: We always can override these configs for setups which serves in-memory and support high batch key count. For others, this config being 1 or 2 is good as anything more might be counterproductive (i,e retries are to help the long tails and thats supposed to help with servers that are busy or in bad state, keep retrying in those scenarios are going to put the system in a bad state).

m-nagarajan (approved):
Thanks @majisourav99

services/venice-router/src/main/java/com/linkedin/venice/router/VeniceRouterConfig.java
majisourav99: yes, with higher retry limit for a large batch size they servers are going to pounded even more and causing further retry storm.

PR #1018, opened 2024-06-05
[all] Make getValueSchemaId APIs match the canonical schemas

nisargthakkar:
## Make `getValueSchemaId` APIs match the canonical schemas

Today, `getValueSchemaId` APIs in `SchemaReader`, `SchemaData` and `ReadOnlySchemaRepository` match the schemas exactly (with the exception of `HelixReadWriteSchemaRepository`). These APIs are generally used to get a schema id to write along with the data. In such cases, we only care about canonical schemas.

The exception is where we want to register a schema with a doc change. In our codebase, this schema check is not handled via `getValueSchemaId` APIs and happens in the `VeniceHelixAdmin#addValueSchema` code paths which handle this case explicitly.

This PR makes `getValueSchemaId` use canonical schema match in all places.

## How was this PR tested?

Updated tests. GH CI

### DIFF
--- a/clients/venice-thin-client/src/main/java/com/linkedin/venice/client/schema/RouterBackedSchemaReader.java
+++ b/clients/venice-thin-client/src/main/java/com/linkedin/venice/client/schema/RouterBackedSchemaReader.java
@@ -67,6 +67,7 @@ public class RouterBackedSchemaReader implements SchemaReader {
   private final Optional<Schema> readerSchema;
   private volatile Schema keySchema;
   private final Map<Integer, SchemaEntry> valueSchemaEntryMap = new VeniceConcurrentHashMap<>();
+  private final Cache<Schema, Integer> valueSchemaToCanonicalSchemaId = Caffeine.newBuilder().maximumSize(1000).build();
   private final Cache<Schema, Integer> canonicalValueSchemaMapR = Caffeine.newBuilder().maximumSize(1000).build();
   private final Map<Integer, DerivedSchemaEntry> valueSchemaIdToUpdateSchemaEntryMap = new VeniceConcurrentHashMap<>();
   private final AtomicReference<SchemaEntry> latestValueSchemaEntry = new AtomicReference<>();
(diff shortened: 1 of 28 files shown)


clients/venice-thin-client/src/main/java/com/linkedin/venice/client/schema/RouterBackedSchemaReader.java
```diff
@@ -211,16 +212,18 @@ public Integer getLatestValueSchemaId() throws VeniceClientException {
 
   @Override
   public int getValueSchemaId(Schema schema) {
-    if (valueSchemaMapR.getIfPresent(schema) == null) {
+    String canonicalSchemaStr = AvroCompatibilityHelper.toParsingForm(schema);
```
sixpluszero, line 215: Hmm, should we move this path into below when original schema -> id is not found? Otherwise each schema id lookup will have to parse the schema? 
Like, the reserve mapping should be original schema -> value schema ID still..

nisargthakkar: That does make sense. Let me think about it

clients/venice-thin-client/src/main/java/com/linkedin/venice/client/schema/RouterBackedSchemaReader.java
```diff
@@ -726,4 +736,20 @@ private Schema preemptiveSchemaVerification(Schema writerSchema, String writerSc
...
+    Schema canonicalSchema = AvroSchemaParseUtils.parseSchemaFromJSONLooseValidation(canonicalSchemaStr);
+
+    cacheValueAndCanonicalSchemas(valueSchema, canonicalSchema, valueSchemaId);
+  }
+
+  private void cacheValueAndCanonicalSchemas(Schema valueSchema, Schema canonicalSchema, int valueSchemaId) {
+    valueSchemaToCanonicalSchemaId.put(valueSchema, valueSchemaId);
+
+    Integer cachedCanonicalSchemaId = canonicalValueSchemaMapR.getIfPresent(canonicalSchema);
+    if (cachedCanonicalSchemaId == null || cachedCanonicalSchemaId < valueSchemaId) {
```
sixpluszero, line 751: What is this case that cached schema id < value schema id? I am a bit confused on this.

sixpluszero (approved):
Overall looks good to me, I just left a question about a condition. If that's good then it should be good.

clients/venice-thin-client/src/main/java/com/linkedin/venice/client/schema/RouterBackedSchemaReader.java
nisargthakkar: I wanted to have a case where we prioritize newer schemas in case the canonical schemas are same for multiple value schemas

sixpluszero: OK got it. Thanks!
(Would be nice to have 1 line doc to describe this so others won't get confused about this check condition, as it is not very obvious from the context.)

nisargthakkar: Yeah. Good idea. I just updated it

sixpluszero (approved):

PR #1019, opened 2024-06-05
[common][client-common] high level blob transfer manager interface

adamxchen:

Add high level blob transfer manager interface as it need to be shared with other teammates now to unblock their work.

## How was this PR tested?

N/A


ishwarya-citro (dismissed):

internal/venice-client-common/src/main/java/com/linkedin/venice/exceptions/VeniceResourceNotFoundException.java
```diff
@@ -0,0 +1,14 @@
+package com.linkedin.venice.exceptions;
+
+/**
+ * Class for all Venice exceptions that are triggered by resources not being found.
+ */
+public class VeniceResourceNotFoundException extends VeniceException {
```
FelixGV, line 6: Can we be a bit more precise about what the intended usage of this new exception type is? For example, how does it compare and contrast with:

```
VeniceNoStoreException
StoreVersionNotFoundException
```

Would one of the above exceptions match the use case intended for this new one?

Or should the above exceptions extend from this new one?

internal/venice-common/src/main/java/com/linkedin/venice/blobtransfer/BlobTransferManager.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+public interface BlobTransferManager<T> extends AutoCloseable {
+  /**
+   * Get the blobs for the given storeName and partition
+   * @param storeName
+   * @param partition
+   * @return the InputStream of the blob
+   * @throws VeniceResourceNotFoundException
+   */
+  CompletionStage<InputStream> get(String storeName, int partition) throws VeniceResourceNotFoundException;
```
FelixGV, line 23: `InputStream` is an abstract class. Is this what we want here? If there is a need for different blob implementations to provide different types of input streams, should we have an additional generic type in the interface, e.g. `I extends InputStream`? Or should we constrain it to a specific type, e.g. a `BufferedInputStream`?

FelixGV (commented):
I have a couple of comments on the interface design... please take a look. Thanks!

internal/venice-client-common/src/main/java/com/linkedin/venice/exceptions/VeniceResourceNotFoundException.java
adamxchen: Yes, I've considered using them but the naming isn't a good fit since this is a partition-level.
Actually it can be a good idea to extend `VeniceNoStoreException` to `StoreVersionNotFoundException` and `StorePartitionNotFoundException`(the new one going to throw here). Will make some changes

internal/venice-common/src/main/java/com/linkedin/venice/blobtransfer/BlobTransferManager.java
adamxchen: This's still a TBD honestly. I borrowed the idea from Ambry since it's simple to start with.
I don't have enough information to determine whether we need generic type (does it worth extra code complexity) or use a specific type (`BufferedInputStream`).  The chance is we may use some customized class in the future for optimization.

adamxchen: I think I'm leaning to `BufferedInputStream` now since it should be sufficient for quite some time

adamxchen: Updated.

internal/venice-client-common/src/main/java/com/linkedin/venice/exceptions/StorePartitionNotFoundException.java
```diff
@@ -0,0 +1,10 @@
+package com.linkedin.venice.exceptions;
+
+/**
+ * Class for all Venice exceptions that are triggered by resources not being found.
```
mynameborat, line 4: I see `VeniceNoStoreException` to denote absence of the store and any extensions of it would be representative of extensions of `store`.

e.g., I see `StorepartitionNotFoundException` as an exception being thrown by the server when a query comes for a store and partition that is not hosted by it. However, that doesn't necessarily mean the store is absent, rather the store partition is not hosted by the server. 

If the store was indeed was absent, i would expect `VeniceNoStoreException` at that point. 

Can you please clarify what this exception denotes? The java document points to resources not found but doesn't talk about what those resources are?

internal/venice-client-common/src/main/java/com/linkedin/venice/exceptions/StoreVersionNotFoundException.java
```diff
@@ -1,6 +1,6 @@
 package com.linkedin.venice.exceptions;
 
-public class StoreVersionNotFoundException extends VeniceException {
+public class StoreVersionNotFoundException extends VeniceNoStoreException {
```
mynameborat, line 3: Thinking a bit more about this, I wonder if `VeniceStoreException` might be more appropriate as these extensions are not a by product of store missing rather different.

what do you think?

internal/venice-common/src/main/java/com/linkedin/venice/blobtransfer/BlobTransferManager.java
```diff
@@ -0,0 +1,37 @@
...
+   * @param storeName
+   * @param partition
+   * @return the type of the object returned from the underlying blob client to indicate the upload status
+   */
+  CompletionStage<T> put(String storeName, int partition);
+
+  /**
+   * Close the blob transfer manager and related resources
+   */
+  void close() throws Exception;
```
mynameborat, line 36: Do we need an `init` method? or `start`?

internal/venice-common/src/main/java/com/linkedin/venice/blobtransfer/BlobTransferManager.java
```diff
@@ -0,0 +1,37 @@
...
+   */
+  CompletionStage<BufferedInputStream> get(String storeName, int partition) throws StorePartitionNotFoundException;
+
+  /**
+   * Put the blob for the given storeName and partition
+   * @param storeName
+   * @param partition
+   * @return the type of the object returned from the underlying blob client to indicate the upload status
+   */
+  CompletionStage<T> put(String storeName, int partition);
```
mynameborat, line 31: Do we need a `delete` method?

internal/venice-common/src/main/java/com/linkedin/venice/blobtransfer/BlobTransferManager.java
mynameborat: Isn't `InputStream` better in the sense the underlying interactions with concrete blob service may return an `InputStream` which isn't buffered and keeping it as-is is more flexible?

Agree, that we should do `CompletionStage<? extends InputStream>` instead of `CompletionStage<InputStream>`

internal/venice-client-common/src/main/java/com/linkedin/venice/exceptions/StorePartitionNotFoundException.java
adamxchen: I missed to update the documents. I could update it to `Class for all Venice exceptions that are triggered by store partitions not being found.`

I think there are some nuances on missing stores, missing versions and missing partitions. 

> If the store was indeed was absent, i would expect VeniceNoStoreException at that point.

This sounds a right idea, so we can throw `VeniceNoStoreException` and depending on whether that's missing, we can throw the exception accordingly

internal/venice-common/src/main/java/com/linkedin/venice/blobtransfer/BlobTransferManager.java
adamxchen: We may need it in the future (if we use a 3rd party blob storage) but that's not something we plan for now.

internal/venice-common/src/main/java/com/linkedin/venice/blobtransfer/BlobTransferManager.java
adamxchen: Sure, we can add that

internal/venice-common/src/main/java/com/linkedin/venice/blobtransfer/BlobTransferManager.java
adamxchen: Oh I missed this comment. 
> concrete blob service may return an InputStream which isn't buffered 

For P2P large file transfer, I think input stream should always be buffered or similarly optimized for better performance and Ambry does something similar but they do only return `InputStream` for simplicity reason I guess.
I don't have strong preference on the generic type but would using wildcard be safe here? I don't feel the casting is safe here especially for `inputStream`

mynameborat: there is still type restrictions imposed with the `? extends InputStream` just that we aren't explicitly taking a type through interface definition. You probably don't require any casting.

mynameborat (approved):

PR #1021, opened 2024-06-07
[vpj] Fix user-friendly partial update mode in Spark

nisargthakkar:
## Fix user-friendly partial update mode in Spark

Incremental push has a user-friendly mode for partial update so users don't have to know the full update schema. VPJ would generate the full update request based on the input data's schema and the fields in that schema.

This feature is not implemented properly in Spark and causes deserialization exception in servers. This is quite critical because the Server cannot recover from this anymore without the RT data being truncated. This is also a data-loss.

## How was this PR tested?

Added a new unit test. Updated `PartialUpdateTest`. GH CI for regression tests.

### DIFF
--- a/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/spark/input/hdfs/VeniceHdfsInputPartitionReader.java
+++ b/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/spark/input/hdfs/VeniceHdfsInputPartitionReader.java
@@ -1,17 +1,7 @@
 package com.linkedin.venice.hadoop.spark.input.hdfs;
 
-import static com.linkedin.venice.hadoop.VenicePushJobConstants.DEFAULT_KEY_FIELD_PROP;
-import static com.linkedin.venice.hadoop.VenicePushJobConstants.DEFAULT_VALUE_FIELD_PROP;
-import static com.linkedin.venice.hadoop.VenicePushJobConstants.ETL_VALUE_SCHEMA_TRANSFORMATION;
-import static com.linkedin.venice.hadoop.VenicePushJobConstants.FILE_KEY_SCHEMA;
-import static com.linkedin.venice.hadoop.VenicePushJobConstants.FILE_VALUE_SCHEMA;
-import static com.linkedin.venice.hadoop.VenicePushJobConstants.KEY_FIELD_PROP;
-import static com.linkedin.venice.hadoop.VenicePushJobConstants.SCHEMA_STRING_PROP;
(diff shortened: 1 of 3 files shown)


clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/spark/input/hdfs/VeniceHdfsInputPartitionReader.java
```diff
@@ -46,25 +35,15 @@ protected VeniceRecordIterator createRecordIterator(VeniceProperties jobConfig,
...
-      String fileKeySchema = jobConfig.getString(FILE_KEY_SCHEMA);
-      String fileValueSchema = jobConfig.getString(FILE_VALUE_SCHEMA);
-      VeniceVsonRecordReader recordReader =
-          new VeniceVsonRecordReader(fileKeySchema, fileValueSchema, keyFieldStr, valueFieldStr);
+      VeniceVsonRecordReader recordReader = new VeniceVsonRecordReader(jobConfig);
       return new VeniceVsonFileIterator(fs, filePath, recordReader);
     }
 
-    // Input data is not in Vson format. It must be Avro.
-    Schema fileSchema = AvroCompatibilityHelper.parse(jobConfig.getString(SCHEMA_STRING_PROP));
```
sixpluszero, line 65: It seems like the fromProps implementation is doing slightly more strict parsing check, which I think it is fine:`    Schema dataSchema = AvroSchemaParseUtils.parseSchemaFromJSON(
        props.getString(SCHEMA_STRING_PROP),
        props.getBoolean(EXTENDED_SCHEMA_VALIDITY_CHECK_ENABLED, DEFAULT_EXTENDED_SCHEMA_VALIDITY_CHECK_ENABLED));
`
Orginal code is doing loose check.
Personally I think it is OK because it is data schema. Just poin this out, If you think it is fine as well, just close this comment.

sixpluszero (approved):
Leave one comment to confirm. But thank you for the quick fix. Other looks good to me!

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/spark/input/hdfs/VeniceHdfsInputPartitionReader.java
nisargthakkar: I think it is fine because the VPJ driver would have already done this same schema check. Also, this is the same code that MR mode is running: https://github.com/linkedin/venice/blob/main/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/mapreduce/datawriter/map/VeniceAvroMapper.java#L14

PR #1023, opened 2024-06-08
[build] Bump minor version when patch versions reach 100

nisargthakkar:
## Bump minor version when patch versions reach 100

Managing too many patch versions is hard. Bump the minor version when the patch version reaches 100.

## How was this PR tested?

Tested locally:
```shell
$ ./make_tag.py
Using remote: upstream
From github.com:linkedin/venice
 * branch                main       -> FETCH_HEAD
From github.com:linkedin/venice
 * branch                main       -> FETCH_HEAD
From github.com:linkedin/venice
 * branch                main       -> FETCH_HEAD
Current branch tagBumpBuild100 is up to date.
New tag is 0.5.0. Continue? [y/N]:  
```

### DIFF
--- a/make_tag.py
+++ b/make_tag.py
@@ -54,8 +54,14 @@ def get_next_version(bump_major, bump_minor, remote):
         return format_version(major, minor, 0)
 
     build = version_ints[2]
+    build += 1
 
-    return format_version(major, minor, build + 1)
+    # If build number is 100, increment minor and reset build
+    if build >= 100:
+        minor += 1
+        build = 0
+
+    return format_version(major, minor, build)
 
 
 def get_remote():


FelixGV:
Why do this...? Version bumps should be driven by semver, not by numerology... the number of digits should not matter 🤔 ...

nisargthakkar:
@FelixGV I kind of agree with you, but this seemed to be the consensus during VS3. We can discuss more and decide on this. I can revert it in the meantime

FelixGV:
If we feel the current minor version has gone on for too long, we can certainly bump it manually, now, once. We did accumulate a lot of changes since the last minor bump. But automating it based on number of digits is not right, IMHO.

PR #1026, opened 2024-06-11
[controller] Do not call metadata store in parent controller

majisourav99:
## Do not call metadata store in parent controller for value schema query

During unused schema cleanup, parent controller sends admin message to all admin including parent admin as well. Since the unused schema info is stored in meta store which is not available in parent cluster it would fail the admin message. This PR skips the query for parent colo and queries the schema set only from child controller.

## How was this PR tested?

integ test.

### DIFF
--- a/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
@@ -3925,6 +3925,11 @@ public class VeniceHelixAdmin implements Admin, StoreCleaner {
       }
       int previousVersion = store.getCurrentVersion();
       store.setCurrentVersion(futureVersion);
+      LOGGER.info(
+          "Rolling forward current version {} to version {} in store {}",
+          previousVersion,
+          futureVersion,
+          storeName);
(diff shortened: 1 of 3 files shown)


services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -4168,6 +4175,10 @@ public StoreMetaValue getMetaStoreValue(StoreMetaKey metaKey, String storeName)
 
   @Override
   public Set<Integer> getInUseValueSchemaIds(String clusterName, String storeName) {
+    if (isParent()) {
```
xunyin8, line 4178: I'm not sure if this work-around is good enough. This means the un-used schemas will be deleted in child fabrics but still linger around in the parent which will still be visible on Nuage right?

majisourav99: emm no. It wil still be deleted in parent, the fix just skips the query in parent colo.

xunyin8: I see, thanks for the explanation

xunyin8 (approved):

PR #1027, opened 2024-06-11
[controller] [admin-tool] Max Record Size Config

KaiSernLim:
## Summary

Large records can cause performance issues, and this new config setting is an important precursor to further work for the Handling Large Records project.

**Note:** The difference between "Venice" records and Kafka records (aka Kafka events) is that Venice records contain one Venice key + value pair and can be spread across 1 or more Kafka events (which have a `~1MB` size limit) with chunking.

**Note:** This setting (`MAX_RECORD_SIZE_BYTES`) applies to all records and is the largest limit. It is separate from the setting determining when chunking is required (`MAX_SIZE_FOR_USER_PAYLOAD_PER_MESSAGE_IN_BYTES`).

Basically:
```
Records without Chunking < MAX_SIZE_FOR_USER_PAYLOAD_PER_MESSAGE_IN_BYTES (~1MB default) < Records with Chunking < MAX_RECORD_SIZE_BYTES (10MB Default)
```

### Changes

1. **[Config]** Added a new store-level config setting `maxRecordSizeBytes`. This will be used to decide if a batch push job should fail or if nearline consumption should be paused due the ingestion of records that are too large.
    1. The default value of `maxRecordSizeBytes` is `-1`. By leaving the limit as unset by default, it is easier for us to add a global config in the future should we deem necessary.
    2. This default value of `maxRecordSizeBytes` will be changed to `10MB` in-memory when it reaches its eventual destination at `VeniceWriter`, even if the config is set to `-1`.
    3. The adjustment is at the store-level, as we anticipate only a few stores could surpass the threshold and require an increased limit.
2. **[Integration Test]** Added tests `testUpdateMaxRecordSize()` and `testChunkingMaxRecordSize()` in `VeniceParentHelixAdminTest`. The former verifies that the config is read and set properly, and the latter 
verifies that the default value of `maxRecordSizeBytes` will be `10MB` if chunking is enabled.
3. **[Refactor]** Refactored tests involving "setting a value and verifying the value was set" into a reusable function `testUpdateConfig()`. All you need to do is provide `Consumer` lambdas for setting a value and verifying the value, which will hopefully make testing configs very straightforward.

## How was this PR tested?

Integration test in CI.


clients/venice-admin-tool/src/main/java/com/linkedin/venice/AdminTool.java
```diff
@@ -1154,6 +1154,11 @@ static UpdateStoreQueryParams getUpdateStoreQueryParams(CommandLine cmd) {
     booleanParam(cmd, Arg.UNUSED_SCHEMA_DELETION_ENABLED, p -> params.setUnusedSchemaDeletionEnabled(p), argSet);
     booleanParam(cmd, Arg.BLOB_TRANSFER_ENABLED, p -> params.setBlobTransferEnabled(p), argSet);
 
+    longParam(cmd, Arg.MAX_RECORD_SIZE_BYTES, params::setMaxRecordSizeBytes, argSet);
+    if (params.getChunkingEnabled().orElse(false) && params.getMaxRecordSizeBytes().orElse(-1L) == -1L) {
+      params.setMaxRecordSizeBytes(10 * 1024 * 1024);
```
KaiSernLim, line 1159: @lluwm Do you know a good place I could put the `10 MB` constant?

lluwm: My thinking is that since this is a constant related to a store config, how about defining it in `com.linkedin.venice.utils.StoreUtils`?

lluwm (commented):
Looks good to me! Please add a bit more unit tests to pass the code coverage verification.

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
```diff
@@ -113,6 +113,20 @@ public class VeniceWriter<K, V, U> extends AbstractVeniceWriter<K, V, U> {
...
+   * Large records can cause performance issues, so this setting is used to detect and prevent them. Not to be
+   * confused with Kafka record size (which is the ~1MB limit {@link MAX_SIZE_FOR_USER_PAYLOAD_PER_MESSAGE_IN_BYTES}
+   * is designed to comply with). Venice records refer to a Venice key-value pair, which can be spread across 1+ Kafka
+   * records / events. Basically: Record without Chunking < 1 MB < Record with Chunking < Max Record Size
+   *
+   * 1. If a batch push data contains records larger than this setting, the push job will fail.
+   * 2. If a nearline job contains records larger than this setting, consumption will be paused and manual intervention
+   * will be necessary.
+   */
+  // public static final String MAX_RECORD_SIZE_BYTES = VENICE_WRITER_CONFIG_PREFIX + "max.record.size.bytes";
```
KaiSernLim, line 128: These lines have to be commented out, otherwise the code coverage tool is going to complain. They will be used in the upcoming changes.

lluwm (approved):
It looks good! Thank you @KaiSernLim !

PR #1029, opened 2024-06-13
[server] Remove Server Ingestion Isolation implementation

sixpluszero:
## [server] Remove Server Ingestion Isolation implementation
This PR removes Venice server ingestion isolation related implementation, as our experiment indicates it does not help achieving intended isolation and ingestion performance we want. It also removes related server tests.
It does not touch Da Vinci ingestion isolation as it is still actively used.

**Interface changes**: Since we only have II for Da Vinci now, this PR consolidate VeniceIngestionBackend, DaVinciIngestionBackend and IngestionBackendBase 3 interface classes into 1 single interface IngestionBackend. 
## How was this PR tested?

Existing tests.


majisourav99 (approved):
Pure red PR!  BTW there are isolated.* configs, are those all would be used in DVC backend or we could clean some of them as well ?

sixpluszero:
> Pure red PR! BTW there are isolated.* configs, are those all would be used in DVC backend or we could clean some of them as well ?

This is shared between DVC and Server, so we cannot remove it. I have removed all server config's isolated config last week, but if there is anything lingering anyone can feel free to remove them.

PR #1030, opened 2024-06-13
[vpj] Remove extra call to PubSub to fetch partition count for data writer job

nisargthakkar:

Remove extra call to PubSub to fetch partition count for data writer job. There was a job failure in production where the VeniceWriter tried to query PubSub to fetch the partition count during the PartitionWriter phase. Since the partition count is always available before the DataWriter job starts, we can specify it explicitly and reduce one call to the PubSub system.

Also, improve docs and logging for shared temp directory for VPJ

## How was this PR tested?

GH CI

### DIFF
--- a/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/task/datawriter/AbstractPartitionWriter.java
+++ b/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/task/datawriter/AbstractPartitionWriter.java
@@ -364,6 +364,7 @@ public abstract class AbstractPartitionWriter extends AbstractDataWriterTask imp
             .setChunkingEnabled(chunkingEnabled)
             .setRmdChunkingEnabled(rmdChunkingEnabled)
             .setTime(SystemTime.INSTANCE)
+            .setPartitionCount(getPartitionCount())
             .setPartitioner(partitioner)
             .build();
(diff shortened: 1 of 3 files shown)


clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/task/datawriter/AbstractPartitionWriter.java
```diff
@@ -364,6 +364,7 @@ private VeniceWriter<byte[], byte[], byte[]> createBasicVeniceWriter() {
             .setChunkingEnabled(chunkingEnabled)
             .setRmdChunkingEnabled(rmdChunkingEnabled)
             .setTime(SystemTime.INSTANCE)
+            .setPartitionCount(getPartitionCount())
```
adamxchen, line 367: is this line to remove the extra call?

nisargthakkar: Yeah. https://github.com/linkedin/venice/blob/main/internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java#L341-L347

adamxchen (approved):
🚀

PR #1032, opened 2024-06-13
[router] Do not mark storage node fast for TOO_MANY_REQUESTS(429)

majisourav99:
## Do not mark storage node fast for TOO_MANY_REQUESTS(429)

When smart long tail retry is enabled, each storage node that responds is marked as fast and is used for subsequent retries. 

However, if a storage node returns a TOO_MANY_REQUESTS (429) status code, it indicates that the node has exceeded its quota and retrying would be futile. 
Retrying in this scenario would increase the load on already overwhelmed servers, leading to higher latency and degraded performance. This PR prevents retries against storage nodes that return a 429 status code, improving overall system efficiency.

## How was this PR tested?

CI


services/venice-router/src/main/java/com/linkedin/venice/router/api/VeniceDispatcher.java
```diff
@@ -157,7 +157,8 @@ public void dispatch(
           throw throwable;
         }
 
-        if (statusCode < HttpStatus.SC_INTERNAL_SERVER_ERROR) {
+        // Do not mark storage node fast for 429 status code
+        if (statusCode < HttpStatus.SC_INTERNAL_SERVER_ERROR && statusCode != HttpStatus.SC_TOO_MANY_REQUESTS) {
```
huangminchn, line 161: Great catch! Could you please write a test for it?

huangminchn (changes requested):

services/venice-router/src/main/java/com/linkedin/venice/router/api/VeniceDispatcher.java
majisourav99: Added integ test for it

xunyin8 (approved):

huangminchn (approved):
Thanks a lot Sourav!

PR #1033, opened 2024-06-14
[server][davinci] Created interface for Blob discovery and added a server side implementation

sebas-inf:
## [server][davinci] Created interface for Blob discovery and added a server side implementation

This PR adds the interface for Blob discovery and adds a server implementation

## How was this PR tested?

Added unit tests


clients/da-vinci-client/src/main/java/com/linkedin/davinci/blob/BlobResponse.java
```diff
@@ -0,0 +1,51 @@
+package com.linkedin.davinci.blob;
+
+import java.util.List;
+import java.util.Objects;
+
+
+public class BlobResponse {
```
sixpluszero, line 7: Let's rename BlobResponse to BlobPeersDiscoveryResponse

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blob/BlobResponse.java
```diff
@@ -0,0 +1,51 @@
...
+
+  public void setMessage(String message) {
+    this.message = message;
+  }
+
+  public String getMessage() {
+    return this.message;
+  }
+
+  public void setPartitionValues(List<String> partitionURLs) {
```
sixpluszero, line 30: Let's also rename this to setDiscoveryResult, also the below getter

sixpluszero: also partitionURLs->hostNameList

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blob/BlobFinder.java
```diff
@@ -0,0 +1,9 @@
+package com.linkedin.davinci.blob;
+
+public interface BlobFinder {
+
+  // This method will look through the partitions for the store and version provided until it finds the partition
+  // requested, it will then return the URL of the instances that are ready to serve in the partition.
+  BlobResponse findBlob(String storeName, int version, int partitionId);
```
sixpluszero, line 7: Can you change this API from findBlob to discoverBlobPeers

sixpluszero (commented):
Thank you for your change! Overall it looks good, let's adjust some naming to make them look more meaningful.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blob/BlobResponse.java
sebas-inf: Made the changes

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blob/BlobResponse.java
sebas-inf: Made the changes

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blob/BlobFinder.java
sebas-inf: Made the changes

services/venice-server/src/main/java/com/linkedin/venice/listener/ServerReadMetadataRepository.java
```diff
@@ -36,13 +36,15 @@ public class ServerReadMetadataRepository implements ReadMetadataRetriever {
   private final ServerMetadataServiceStats serverMetadataServiceStats;
   private final ReadOnlyStoreRepository storeRepository;
   private final ReadOnlySchemaRepository schemaRepository;
+  // Duplicate this file and make it so that it only retrieves the customizedViewRepository
```
sixpluszero, line 39: Can you remove these comments as well?

sixpluszero (dismissed):
LGTM! Thank you for your first PR in Venice!

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blob/BlobFinder.java
```diff
@@ -0,0 +1,9 @@
+package com.linkedin.davinci.blob;
```
adamxchen, line 1: nit: do you think we can move these files to `venice-common` package?

adamxchen: Specifically `BlobFinder`, `BlobPeersDiscoveryResponse`, as I plan to use them in `venice-common`

sixpluszero: Hmm, but I think they are all for Da Vinci and Server functionality right? Or did you place everything blob related in venice-common already?

adamxchen: > Or did you place everything blob related in venice-common already?
Yes, I thought this module might be used by vpj so i think it was more appropriate to have them in `venice-common`. That said, I can move them to `da-vinci-client` package tho.

sixpluszero: OK, for simplicity, let us move all of them in the venice-common for now, so don't need to write duplicate unit test to pass coverage

sebas-inf: I moved them to venice-common

sixpluszero (approved):

internal/venice-common/src/main/java/com/linkedin/venice/blob/ServerBlobFinder.java
```diff
@@ -0,0 +1,54 @@
...
+    BlobPeersDiscoveryResponse response = new BlobPeersDiscoveryResponse();
+    try {
+      String currentVersionResource = Version.composeKafkaTopic(storeName, version);
+      // Get the partition assignments for the current version
+      List<String> hostNames = new ArrayList<>();
+      for (Partition partition: customizedViewRepository.getPartitionAssignments(currentVersionResource)
+          .getAllPartitions()) {
+        if (partition.getId() == partitionId) {
+          for (Instance instance: partition.getReadyToServeInstances()) {
+            String host = instance.getHost();
```
ZacAttack, line 35: @adamxchen Should we not specify the protocol port here somehow?  Otherwise we'd need to have some kind of client server agreement set in cfg2, which is a bit contrary to how discovery works in other contexts.

@sixpluszero?

sixpluszero: Our internal discussion leaves it outside of this interface. There will be a BlobManger thing that takes in this implementation, and there it is suppose to know port.
But I believe this client-server talking will be very much similar to II main-fork process. We will have to specify cfg for port instead of hardcode it somewhere. Both side should setup Netty server and client to talk to each other with the specific allocated port.

sixpluszero: Essentially, this interface does only one thing, look up the peers from different sources. If you think discovery is not a good name, we can adjust it later. I don't have a strong opinion in API naming, but just to mean that this API returns list of hostname having this specific resource.

PR #1034, opened 2024-06-14
[server] Pass ingestion role change for store version from StoreIngestionTask to AggKafkaConsumerService.

haoxu07:

This PR introduced a new class `TopicPartitionReplicaRole` to bundle information about particular topic partition: 
1. `Version` role: `future`/`current`/`backup`
2.  Workload type: Active Active or Write Compute due to its nature of heavy processing needed. 

And a new config `SERVER_RESUBSCRIPTION_TRIGGERED_BY_VERSION_INGESTION_CONTEXT_CHANGE_ENABLED` to turn on/off the resubscription.

`StoreIngestionTask` thread will periodically check the store's version role changed and store version's workload type changed. If changed, it will trigger resubscription by graceful un-subscription/ re-subscription for all partitions to let bottom `AggKafkaConsumerService` to find corresponding consumer thread pool for them. In the future `KafkaConsumerService` aware of store version role during subscription, there will be more information with `TopicPartitionReplicaRole`  passed with `aggKafkaConsumerService#subscribe`.  

In the future, we could also bundle more information to this object to pass into `KafkaConsumerService` to let it do better consumer pool allocation. With proper allocation of consumer thread pool, we could achieve prioritization on certain type of traffic (e.g. we could allocate larger consumer thread for current hybrid leader ingestion).

A unit test was added to verify the subscription/subscription happened for local and remote ingestion, by intentionally triggering version role change during ingestion.

## How was this PR tested?

CI test


clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/TopicPartitionReplicaRole.java
```diff
@@ -0,0 +1,41 @@
+package com.linkedin.davinci.kafka.consumer;
+
+import com.linkedin.venice.pubsub.api.PubSubTopic;
+import com.linkedin.venice.pubsub.api.PubSubTopicPartition;
+
+
+public class TopicPartitionReplicaRole {
```
ZacAttack, line 7: Let's add some javadoc?  What is this for?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/TopicPartitionReplicaRole.java
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * This class is for passing the information regarding the role of the partition replica on that host to
+ *  {@link AggKafkaConsumerService} to achieve finer granularity of consumer assignment. Those information should be
+ *  triggered by helix state transition and properly managed by {@link StoreIngestionTask} to be sent to
+ *  {@link AggKafkaConsumerService}. Currently, the role of future/current/backup for version and leader/follower for
+ *  partition are supported. We could add more information regarding this partition replica if needed in the future.
+ */
+
+public class TopicPartitionReplicaRole {
```
sixpluszero, line 15: Overall it looks good, but I am thinking an alternative way to implement this info. I see you create this class as a more stateful representation of Replica, but when you are using this, you only use VersionRole alone and don't use other info wrapped inside, like you still check leader or not using PCS info, instead of isLeaderReplica() API you created. 

Do you think it would be better either to make this class stronger representation via making this role settable and you always use this one object throughout the lifecycle of a Replica? or maybe even just simplify this to only define a VersionRole and pass it and use it as you are doing right now, instead of wrapping with other info?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -556,6 +563,23 @@ public synchronized void subscribePartition(PubSubTopicPartition topicPartition)
     subscribePartition(topicPartition, true);
   }
 
+  // TODO: ensure thread safety for version role change.
+  public synchronized void versionRoleChangeToTriggerResubscribe(TopicPartitionReplicaRole.VersionRole versionRole) {
```
sixpluszero, line 567: a bit confused here, would it be triggered to all backup/current/future SIT to do the re-sub?

sixpluszero (commented):
Thanks Hao for the change! I did one quick pass, and left two comments, overall it looks ok but just want to check with you about the way new class is defined.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/TopicPartitionReplicaRole.java
sushantmane: I agree with Jialin - we should make this as a proper Replica abstraction or just trim it down to bare minimum required fields.

haoxu07: So first, this class is not for stateful representation of replica, but for a simplified interface for bottem `KafkaConsumerService` by passing information of topic partition replica, isLeader, isCurrentVersion. So we could include these 3 information in one class, or we could just pass 3 into the interface. And it is possible that in the future we would have more information to let `KafkaConsumerService`  know. Regarding stateful replica abstraction, I think now PCS already played that role.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
haoxu07: It would trigger all SIT to do re-sub, when we detect a role change. So you think this is too costly?

sixpluszero: No, I think I kind of realized they are all necessary at least for now.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/TopicPartitionReplicaRole.java
gaojieliu: Shall we change it to something called `TopicPartitionInfo` or `TopicPartitionIngestionContext` to more a little generic as we may add more information later?
From the usage, it is certainly not a stateful object, but rather a collection of params passing to `KCS`.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ConsumerAction.java
```diff
@@ -27,6 +27,8 @@ public class ConsumerAction implements Comparable<ConsumerAction> {
   private boolean isHelixTriggeredAction = true;
   private CompletableFuture<Void> future = new CompletableFuture<>();
 
+  private TopicPartitionReplicaRole.VersionRole versionRole;
+
   public ConsumerAction(
```
gaojieliu, line 32: Maybe in the near future, we can refactor this `ConsumerAction` class as it becomes a little messy now since some of the params are not useful or required in certain actions. For example, this new param: `versionRole` is only useful for some of the actions, such as `kill` might not need this param.
During the refactoring, we can consider to use `inheritance` to specify different behaviors for different `ConsumerAction` type.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ConsumerActionType.java
```diff
@@ -9,7 +9,7 @@ public enum ConsumerActionType {
    * KILL action has higher priority than others, so that once KILL action is added to the action queue,
    * we will process it immediately to avoid doing throw-away works.
    */
-  KILL(2), STANDBY_TO_LEADER(1), LEADER_TO_STANDBY(1);
+  KILL(2), STANDBY_TO_LEADER(1), LEADER_TO_STANDBY(1), VERSION_ROLE_CHANGE_TO_TRIGGER_RESUBSCRIBE(1);
```
gaojieliu, line 12: I feel `VERSION_ROLE_CHANGE_TO_TRIGGER_RESUBSCRIBE` is too specific as it includes concrete action into the `type`, and the concrete actions can change in the future, so I feel `VERSION_ROLE_CHANG` is good enough.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -473,6 +480,17 @@ protected void processConsumerAction(ConsumerAction message, Store store) throws
...
+      case VERSION_ROLE_CHANGE_TO_TRIGGER_RESUBSCRIBE:
+        partitionConsumptionState = partitionConsumptionStateMap.get(partition);
+        if (partitionConsumptionState == null) {
+          LOGGER.info(
+              "Consumer action: {} is skipped for replica: {}, because PCS is null and the partition may have been unsubscribed.",
+              message,
+              Utils.getReplicaId(topicName, partition));
+          return;
+        }
+        resubscribe(partitionConsumptionState, message.getVersionRole());
```
gaojieliu, line 492: This action is quite expensive, and I think we should introduce a feature flag to control this feature, and when the feature flag is off, `SIT` shouldn't even receive such event.
This feature flag will serve two purposes (at least):
1. Gradual ramp.
2. Quick rollback if this feature is problematic.

gaojieliu (commented):
The code change looks good overall, and I left some minor comments.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
haoxu07: Ideally, this action could only be triggered when version role changed for all store versions. Even it is for all partitions, but it is not very frequent during the store version life cycle.  For safety, we could introduce a host level flag to turn on and off, I will add it.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -473,6 +480,17 @@ protected void processConsumerAction(ConsumerAction message, Store store) throws
          */
         endSegment(partition);
         break;
+      case VERSION_ROLE_CHANGE_TO_TRIGGER_RESUBSCRIBE:
```
huangminchn, line 483: BTW, shouldn't this be handled as a common action in SIT? I don't see any logic regarding leader follower role yet.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -1014,6 +1040,11 @@ private long getLastConsumedMessageTimestamp(int partition) {
   }
 
   protected boolean shouldNewLeaderSwitchToRemoteConsumption(PartitionConsumptionState partitionConsumptionState) {
+    LOGGER.info(
+        "PCS: {}, isConsumingFromRemoteVersionTopic: {}, isLeaderConsumingRemoteRealTimeTopic: {}",
+        partitionConsumptionState,
+        isConsumingFromRemoteVersionTopic(partitionConsumptionState),
+        isLeaderConsumingRemoteRealTimeTopic(partitionConsumptionState));
```
huangminchn, line 1047: Was this for debugging during implementation? If no, let's use the task name prefix in the log.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -3506,6 +3538,68 @@ protected Set<String> maybeSendIngestionHeartbeat() {
...
   }
 
+  /**
+   *  Resubscribe operation by passing new version role and partition role to {@link AggKafkaConsumerService}. The action
+   *  for leader and follower replica will be handled differently.
+   */
+  private void resubscribe(
+      PartitionConsumptionState partitionConsumptionState,
+      TopicPartitionReplicaRole.VersionRole versionRoleFromConsumerAction) throws InterruptedException {
+    if (isLeader(partitionConsumptionState)) {
```
huangminchn, line 3548: We put leader flag inside TopicPartitionReplicaRole but sill use PCS to determine leader role? Which flag should be the source of truth when resubscription happens?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -3506,6 +3538,68 @@ protected Set<String> maybeSendIngestionHeartbeat() {
...
+
+  protected void resubscribeAsFollower(
+      PartitionConsumptionState partitionConsumptionState,
+      TopicPartitionReplicaRole.VersionRole versionRoleFromConsumerAction) throws InterruptedException {
+    int partition = partitionConsumptionState.getPartition();
+    consumerUnSubscribe(versionTopic, partitionConsumptionState);
+    waitForAllMessageToBeProcessedFromTopicPartition(
+        new PubSubTopicPartitionImpl(versionTopic, partition),
+        partitionConsumptionState);
+    LOGGER.info(
```
huangminchn, line 3563: Shall we put the ingestion task name as prefix?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -3506,6 +3538,68 @@ protected Set<String> maybeSendIngestionHeartbeat() {
...
+      resubscribeAsFollower(partitionConsumptionState, versionRoleFromConsumerAction);
+    }
+  }
+
+  protected void resubscribeAsFollower(
+      PartitionConsumptionState partitionConsumptionState,
+      TopicPartitionReplicaRole.VersionRole versionRoleFromConsumerAction) throws InterruptedException {
+    int partition = partitionConsumptionState.getPartition();
+    consumerUnSubscribe(versionTopic, partitionConsumptionState);
+    waitForAllMessageToBeProcessedFromTopicPartition(
```
huangminchn, line 3560: Timeout exception might happen here, which is not fatal previously when server does unsubscription for standby to offline transition or when server does graceful shutdown; however, I don't think exception is okay during resubscription, right?
Let's think about it thoroughly; if unsubscription fails somehow, regardless of the error types, do we fail this partition or continue with the subscription in the new pool?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -3506,6 +3538,68 @@ protected Set<String> maybeSendIngestionHeartbeat() {
...
+    LOGGER.info(
+        "Follower replica: {} unsubscribe finished, prepare for future re-subscribe.",
+        partitionConsumptionState.getReplicaId());
+    PubSubTopicPartition followerTopicPartition = new PubSubTopicPartitionImpl(versionTopic, partition);
+    TopicPartitionReplicaRole topicPartitionReplicaRole =
+        new TopicPartitionReplicaRole(false, versionRoleFromConsumerAction, followerTopicPartition, versionTopic);
+    long latestProcessedLocalVersionTopicOffset = partitionConsumptionState.getLatestProcessedLocalVersionTopicOffset();
+    consumerSubscribe(topicPartitionReplicaRole, latestProcessedLocalVersionTopicOffset, localKafkaServer);
+    LOGGER.info(
+        "Follower replica: {} resubscribe to {}",
```
huangminchn, line 3572: Let's put "offset" in the log to make it more readable.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -469,6 +474,8 @@ public StoreIngestionTask(
         this::pauseConsumption,
         this::resumeConsumption);
     this.storeRepository.registerStoreDataChangedListener(this.storageUtilizationManager);
+    this.storeVersionRoleChangedListener = new StoreVersionRoleChangedListener(this, store, versionNumber);
+    this.storeRepository.registerStoreDataChangedListener(storeVersionRoleChangedListener);
```
huangminchn, line 478: I don't think this is the correct way to use ZK callback. Any store change will trigger callbacks in all registered listeners, meaning that changes in Store A will trigger listener for store B's SIT with the current implementation; and current implementation will end up adding too many listeners.

I believe the correct way is to build only one listener in ingestion service, when it receives changes for a specific store, invoke version change API in its SITs.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreVersionRoleChangedListener.java
```diff
@@ -0,0 +1,69 @@
...
+      this.versionRole = newVersionRole;
+      this.storeIngestionTask.versionRoleChangeToTriggerResubscribe(newVersionRole);
+    }
+  }
+
+  private TopicPartitionReplicaRole.VersionRole getStoreVersionRole(Store store) {
+    try {
+      int currentVersionNumber = store.getCurrentVersion();
+      if (currentVersionNumber < versionNumber) {
+        return TopicPartitionReplicaRole.VersionRole.FUTURE;
```
huangminchn, line 58: If a future version push already fail but it remains in the version list of the store in ZK, will this trigger resubscription?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -3506,6 +3538,68 @@ protected Set<String> maybeSendIngestionHeartbeat() {
     return failedPartitions;
   }
 
+  /**
+   *  Resubscribe operation by passing new version role and partition role to {@link AggKafkaConsumerService}. The action
+   *  for leader and follower replica will be handled differently.
+   */
+  private void resubscribe(
+      PartitionConsumptionState partitionConsumptionState,
+      TopicPartitionReplicaRole.VersionRole versionRoleFromConsumerAction) throws InterruptedException {
```
huangminchn, line 3547: Can we do a shortcuit and skip the resubscribe if version role doesn't change? For example, a store config change like increasing quota will trigger store change callback; in this case, SIT's version role doesn't change, and there is no point doing resubscription.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -327,6 +328,10 @@ public abstract class StoreIngestionTask implements Runnable, Closeable {
   private final Runnable runnableForKillIngestionTasksForNonCurrentVersions;
   protected final AtomicBoolean recordLevelMetricEnabled;
 
+  private final StoreVersionRoleChangedListener storeVersionRoleChangedListener;
+
+  protected TopicPartitionReplicaRole.VersionRole versionRole;
```
huangminchn, line 333: What is its initial value?
The very first subscription action triggered by Helix is using this value as SOT.

huangminchn (changes requested):
Discussed with Hao offline; there are 2 more high level issues to look into:
1. For batch-only stores, there is no need for resubscription. When future ingestion is done and it becomes current version, it doesn't need any consumer from the current version pool.
2. Whether there could be a race where SIT might close itself after unsubscribing to all hosted partitions.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
huangminchn: @lluwm Could you please help check if we need to set the unsubscribe flag in PCS here?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
gaojieliu: For this, I assume these two should be consistent considering consumer actions will be handled sequentially, and it is good to leave some javadoc to make that clear.

gaojieliu:
> For batch-only stores, there is no need for resubscription. When future ingestion is done and it becomes current version, it doesn't need any consumer from the current version pool.
I assume SIT would unsubscribe right after reporting completion for batch-only store, so this action for batch-only stores will be an NO-OP as there won't be active subscriptions.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
haoxu07: Actually, we also did similar thing during `LEADER_TO_STANDBY` consumer action processing. It basically did the same thing to unsub and then subscribe to the different topic, and it is also executed by SIT thread. Fail logic should be similar here, we fail the partition and this version on this host.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
haoxu07: Here this class is not stateful but more like to bundle a lot of information to be passed to KCS.  We construct new `TopicPartitionReplicaRole` before subscribe in this resubscribe operation.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
haoxu07: Actually, we cached the `verisonRole` inside the  `storeVersionRoleChangedListener` and compare new role with previous one to see if issuing `VersionRoleChange` Consumer Action.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/TopicPartitionReplicaRole.java
sixpluszero: No I mean in this case, the only thing you are passing into is the VersionRole. What I meant is if you don't need to keep it as stateful object that you can maintain the state for a Replica abstraction (which I think we eventually may/should), then you can just keep the VersionRole Enum as other information in this class you created are not using it in the code.... It is certainly not wrong, just I think it can be simplified. We can come back to this comment later.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
haoxu07: The initial value of this versionRole is set by `storeVersionRoleChangedListener` per store version initially in the constructor. If we decided to move `storeVersionRoleChangedListener` into `KafkaIngestionService`, I think we may need to add logic to decide the role initially.

huangminchn:
> > For batch-only stores, there is no need for resubscription. When future ingestion is done and it becomes current version, it doesn't need any consumer from the current version pool.
> 
> I assume SIT would unsubscribe right after reporting completion for batch-only store, so this action for batch-only stores will be an NO-OP as there won't be active subscriptions.

@gaojieliu I believe the store change listener would still send out the new consumer action (haven't checked the latest diff yet). Could you please also take a look at my comment of the listener?  https://github.com/linkedin/venice/pull/1034#discussion_r1690588199

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
gaojieliu: Another way would be let `SIT` thread periodically check the version status and if there is a change, invoke version status change API.
The subscription change is not time sensitive, so it doesn't have to react to ZK change immediately.

haoxu07: I think let SIT thread periodically check the version status would simplify the logic. It will not incur more threads and simplify the logic, and probably we do not need to enqueue consumer action.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
haoxu07: I think SIT LOGGER will take ingestion task name by default like:
`2024-07-29 18:04:03 INFO [LeaderFollowerStoreIngestionTask] [venice-SIT-TestTopic_1c7b57bc6eb9a_4b836876_v1] Leader replica: TestTopic_1c7b57bc6eb9a_4b836876_v1-1 unsubscribe finished for future resubscribe.`

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -3505,6 +3510,61 @@ protected Set<String> maybeSendIngestionHeartbeat() {
...
+        "Follower replica: {} resubscribe to offset: {}",
+        partitionConsumptionState.getReplicaId(),
+        latestProcessedLocalVersionTopicOffset);
+  }
+
+  protected void resubscribeAsLeader(PartitionConsumptionState partitionConsumptionState) throws InterruptedException {
+    OffsetRecord offsetRecord = partitionConsumptionState.getOffsetRecord();
+    PubSubTopic leaderTopic = offsetRecord.getLeaderTopic(pubSubTopicRepository);
+    int partition = partitionConsumptionState.getPartition();
+    consumerUnSubscribe(leaderTopic, partitionConsumptionState);
```
huangminchn, line 3549: Just curious, if we don't pass in the previous version role, how do we know which pool to unsubscribe from?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -1344,6 +1377,7 @@ public void run() {
 
       while (isRunning()) {
         Store store = storeRepository.getStoreOrThrow(storeName);
+        updateIngestionRoleIfStoreChanged(store);
```
huangminchn, line 1380: Why do this as the first step? Is there any particular reason?

huangminchn (changes requested):

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
haoxu07: We do not need this information to unsubscribe. For each host, there will only one unique <version_topic, topic_partition> to be subscribed at any time, no matter which consumer pool it belongs to.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
haoxu07: No particular reason, but just want to get changed version role or workloadType information asap to update internal state. As changed information inside SIT could affect following subscriptions executed in this loop.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -1273,6 +1284,28 @@ protected void checkIngestionProgress(Store store) throws InterruptedException {
     Thread.sleep(readCycleDelayMs);
   }
 
+  protected void updateIngestionRoleIfStoreChanged(Store store) throws InterruptedException {
+    PartitionReplicaIngestionContext.VersionRole newVersionRole =
+        PartitionReplicaIngestionContext.getStoreVersionRole(versionTopic, store);
+    PartitionReplicaIngestionContext.WorkloadType newWorkloadType =
+        PartitionReplicaIngestionContext.getWorkloadType(versionTopic, store);
+    if (serverConfig.isResubscriptionTriggeredByVersionIngestionContextChangeEnabled() && isHybridMode()) {
+      if (!newVersionRole.equals(versionRole) || !newWorkloadType.equals(workloadType)) {
```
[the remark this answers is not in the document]
haoxu07: As we discussed, our "writeComputationEnabled": true is store level config.  Regarding workload type we supported in this code change, AA_OR_WRITE_COMPUTE could change during ingestion.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionRoleChangedListener.java
```diff
@@ -0,0 +1,5 @@
+package com.linkedin.davinci.kafka.consumer;
+
+public class StoreIngestionRoleChangedListener {
```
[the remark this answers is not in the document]
haoxu07: Fogot to delete. We should delete it.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaConsumerServiceDelegator.java
```diff
@@ -120,12 +120,13 @@ public long getMaxElapsedTimeMSSinceLastPollInConsumerPool() {
 
   @Override
   public void startConsumptionIntoDataReceiver(
-      PubSubTopicPartition topicPartition,
+      PartitionReplicaIngestionContext partitionReplicaIngestionContext,
```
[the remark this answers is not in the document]
haoxu07: `PartitionReplicaIngestionContext` is for passing information to guide consumer pool allocation.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
```diff
@@ -1411,7 +1416,7 @@ Runnable buildRepairTask(
           getTopicPartitionOffsetByKafkaURL(sourceKafkaUrl, sourceTopicPartition, rewindStartTimestamp);
 
       // Subscribe (unsubscribe should have processed correctly regardless of remote broker state)
-      consumerSubscribe(sourceTopicPartition, upstreamOffset, sourceKafkaUrl);
+      consumerSubscribe(sourceTopicPartition, true, upstreamOffset, sourceKafkaUrl);
```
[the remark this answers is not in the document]
haoxu07: Discussed offline, as we do not need isLeader boolean flag for `PartitionReplicaIngestionContext` now, so we just removed this from it. Bottem KafkaConsumerService could infer from topic name to know it is the leader or not.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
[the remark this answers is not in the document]
haoxu07: Temporarily make workloadType a final inside SIT.

clients/da-vinci-client/src/test/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTaskTest.java
```diff
@@ -3689,6 +3659,169 @@ public void testLeaderShouldSubscribeToCorrectVTOffset() {
...
+        resubscriptionOffsetForRemoteRT.size() + resubscriptionOffsetForLocalRT.size();
+
+    vtWriter.broadcastStartOfPush(new HashMap<>());
+
+    // Produce batchMessagesNum messages to local Venice version topic
+    produceRecordsUsingSpecificWriter(localVeniceWriter, 0, batchMessagesNum, this::getNumberedKeyForPartitionBar);
+
+    // Set two observers for both local and remote consumer thread, these observers will trigger resubscription by
+    // setting
+    // the version role to Backup when the offset reaches the specified value.
```
huangminchn, line 3762: Thanks for adding the comments!

huangminchn (approved):
Thanks for the hard work Hao!

PR #1036, opened 2024-06-16
[vpj] Use native Spark readers for reading from Avro and VSON files

nisargthakkar:
## Use native Spark readers for reading from Avro and VSON files

Currently, we implement a custom input data source to read Avro and VSON files from HDFS. This input data source isn't optimized for large files (i.e. files larger than HDFS block size) and is not aware of blocks and location of those on HDFS. This PR modifies the input logic to use Spark's native Avro and SequenceFile input formats that should be much more optimal for files larger than the block size on HDFS.

Additionally, internally at LinkedIn, we have another input format that reads from a Spark table. This commit adds a lot of the base code that can be used to add arbitrary Spark dataframes as supported input formats. The remaining gap is the schema validation done before the DataWriterComputeJob.

## How was this PR tested?

Unit tests. GHCI regression tests pass


clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/spark/datawriter/jobs/DataWriterSparkJob.java
```diff
@@ -29,30 +31,59 @@ protected Dataset<Row> getUserInputDataFrame() {
...
     }
-    return dataFrameReader.load();
+  }
+
+  private Dataset<Row> getAvroDataFrame(SparkSession sparkSession, PushJobSetting pushJobSetting) {
+    Dataset<Row> df =
+        sparkSession.read().format("avro").option("pathGlobFilter", GLOB_FILTER_PATTERN).load(pushJobSetting.inputURI);
+
+    // Transforming the input data format
+    df = df.map((MapFunction<Row, Row>) (Row record) -> {
```
FelixGV, line 46: Had no idea this syntax was possible for a map closure...!

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/spark/utils/RowToAvroConverter.java
```diff
@@ -0,0 +1,475 @@
...
+ * accordance with the following resources:
+ * <ul>
+ *   <li><a href="https://spark.apache.org/docs/latest/sql-data-sources-avro.html">Spark Avro data source documentation</a></li>
+ *   <li><a href="https://avro.apache.org/docs/1.11.1/spec.html">Avro specification</a></li>
+ *   <li><a href="https://github.com/apache/spark/blob/master/connector/avro/src/main/scala/org/apache/spark/sql/avro/AvroSerializer.scala">Spark's internal Catalyst type to Avro bytes implementation</a></li>
+ * </ul>
+ *
+ * Spark's implementation is not ideal to be used directly for two reasons:
+ * <ul>
+ *   <li>It cannot handle complex unions in the version of Spark that we use (3.1.1). The support was added in 3.4.0.</li>
```
FelixGV, line 62: And I suppose there are reasons why we cannot upgrade to 3.4 yet?

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/spark/utils/RowToAvroConverter.java
```diff
@@ -0,0 +1,475 @@
...
+ *   <li>It converts directly to Avro binary that we need to deserialize, and that incurs an additional serde cost.</li>
```
FelixGV, line 63: Good to know... and I guess that is still the case even in 3.4?

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/task/datawriter/AbstractPartitionWriter.java
```diff
@@ -576,7 +554,8 @@ protected void detectAndHandleDuplicateKeys(
           identicalValuesToKeyCount++;
           if (shouldPrint) {
             shouldPrint = false;
-            LOGGER.warn(printDuplicateKey(keyBytes));
```
FelixGV, line 579: Do you think it's better to remove this, or is this just a shortcut as part of iterating on this change and you're planning to add it back after resolving some blocker...?

FelixGV (commented):
Nice work! Just did a superficial pass, but I might have missed some stuff. I think there are some test failures in any case so it probably needs a little more iteration before merging...

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/spark/utils/RowToAvroConverter.java
nisargthakkar: Yeah, it's the same in 3.4 too. We might still choose to go that route because of reduced maintenance overhead on our part

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/spark/utils/RowToAvroConverter.java
nisargthakkar: We use 3.1.1 because that's what we base the internal version at LI off of. When adding Spark support, I wanted to minimize the variables. I can try to bump to newer versions, but the transitive dependencies and stuff might not be happy. However, it is a bullet that we might need to bite

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/task/datawriter/AbstractPartitionWriter.java
nisargthakkar: The reason to remove this is it was failing due to configs missing. It is easy to have this impl, but I think it is not the right thing to log keys or values. Either of these can have private information that should not be logged. In addition to this, I have never seen a case where we relied on this to get the relevant info.

It is easy to use external analysis tools like Hive/Spark/Trino to get this info from the input data

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/spark/datawriter/jobs/DataWriterSparkJob.java
nisargthakkar: Even I didn't realize this `(Row record)` implementation. Apparently, it is not necessary to have the type. Reliance on IDEs has made us dumber.

The `(MapFunction<Row, Row>)` is a type casting of the lambda, that we need to do because of some Scala/Java weirdness

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/spark/utils/RowToAvroConverter.java
FelixGV: Bah, it's fine to keep it at 3.1.1 for now I guess.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/spark/datawriter/jobs/DataWriterSparkJob.java
```diff
@@ -55,4 +85,53 @@ protected Dataset<Row> getUserInputDataFrame() {
...
+      final byte[] inputKeyBytes = recordReader.getKeyBytes(recordAvroWrapper, null);
+      final byte[] inputValueBytes = recordReader.getValueBytes(recordAvroWrapper, null);
+
+      return new GenericRowWithSchema(new Object[] { inputKeyBytes, inputValueBytes }, DEFAULT_SCHEMA);
+    }, RowEncoder.apply(DEFAULT_SCHEMA));
+
+    return df;
+  }
+
+  @Deprecated
```
adamxchen, line 118: I wonder whether we can drop the vson support. We are adding a new method that's marked as deprecated right away

nisargthakkar: There are some stores that still use Vson input

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/spark/input/hdfs/VeniceHdfsInputPartitionReaderFactory.java
```diff
@@ -7,6 +7,7 @@
 import org.apache.spark.sql.connector.read.PartitionReaderFactory;
 
 
+// TODO: Explore FilePartitionReaderFactory
```
adamxchen, line 10: nit: can you provide a bit more context here why we need to explore it or maybe we can just create an internal ticket to track the improvement?

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestBatch.java
```diff
@@ -1409,7 +1411,7 @@ public void testKafkaInputBatchJobSucceedsWhenSourceTopicIsEmpty() throws Except
   }
 
   @Test(timeOut = TEST_TIMEOUT, dataProvider = "True-and-False", dataProviderClass = DataProviderUtils.class)
-  public void testBatchJobSnapshots(Boolean isKakfaPush) throws Exception {
```
adamxchen, line 1412: good eye 😄  It still took me a second scan to find where spelled wrong

adamxchen (approved):
LGTM. A few nits. I think our integ tests have pretty extensive testing around Avro and VSON formats so we should be good but we probably gotta do more in pre-production env

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/spark/input/hdfs/VeniceHdfsInputPartitionReaderFactory.java
nisargthakkar: I think I just forgot to remove this. I was exploring on this front previously

nisargthakkar:
> but we probably gotta do more in pre-production env
Yeah, it'll have to be somewhat of a slow ramp

adamxchen (approved):

PR #1037, opened 2024-06-17
[vpj][controller] Fix mismatch in per colo status reporting and overall status priorities to include DVC error codes

m-nagarajan:
## Summary
1. per region status returned by controller holds `ExecutionStatus` data, but is read as `PushJobDetailsStatus`. Have been working so far coincidently based on the resulting statuses, but adding new DVC statues uncovered this issue. Added a method to transform `ExecutionStatus` to `PushJobDetailsStatus` after reading in VPJ.
2. `VeniceHelixAdmin.STATUS_PRIORITIES` didn't include the new DVC error status resulting in these returning as ERROR and these new statuses were not getting checkpointed. Added these and added unit tests to check for proper returns.

## How was this PR tested?
GH CI

### DIFF
--- a/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
+++ b/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
@@ -1762,6 +1762,36 @@ public class VenicePushJob implements AutoCloseable {
     return null;
   }
 
+  /** Transform per colo {@link ExecutionStatus} to per colo {@link PushJobDetailsStatus} */
+  protected static PushJobDetailsStatus getPerColoPushJobDetailsStatusFromExecutionStatus(
+      ExecutionStatus executionStatus) {
+    switch (executionStatus.getRootStatus()) {
+      case NOT_CREATED:
+      case NEW:
+      case NOT_STARTED:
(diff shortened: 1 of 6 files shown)


clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
```diff
@@ -1772,16 +1802,17 @@ private void updatePushJobDetailsWithColoStatus(Map<String, String> coloSpecific
           // Don't bother updating the completed colo's status
           .filter(coloEntry -> !completedColos.contains(coloEntry.getKey()))
           .forEach(coloEntry -> {
-            int status = PushJobDetailsStatus.valueOf(coloEntry.getValue()).getValue();
+            ExecutionStatus executionStatus = ExecutionStatus.valueOf(coloEntry.getValue());
```
majisourav99, line 1805: can we wrap this in try-catch to capture `IllegalArgumentException` if any.

m-nagarajan: There is already a try catch logging exception for this. Resolving after discussing offline.

majisourav99 (approved):
LGTM! Thanks Manoj

PR #1039, opened 2024-06-18
[client][davinci]Blob Snapshot Manager

sebas-inf:

This PR adds the snapshot functionality for Hybrid Stores 

## How was this PR tested?

Designed some unit test for it


clients/da-vinci-client/src/main/java/com/linkedin/davinci/blob/BlobFinder.java
```diff
@@ -0,0 +1,9 @@
+package com.linkedin.davinci.blob;
+
+public interface BlobFinder {
+
+  // This method will look through the partitions for the store and version provided until it finds the partition
+  // requested, it will then return the URL of the instances that are ready to serve in the partition.
```
KaiSernLim, line 6:
```suggestion
  /**
   * This method will look through the partitions for the store and version provided until it finds the partition
   * requested, it will then return the URL of the instances that are ready to serve in the partition.
   */
```
I think it would be better to use docstring comments for classes and methods.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blob/ServerBlobFinder.java
```diff
@@ -0,0 +1,54 @@
...
+          .getAllPartitions()) {
+        if (partition.getId() == partitionId) {
+          for (Instance instance: partition.getReadyToServeInstances()) {
+            String host = instance.getHost();
+            partitionValues.add(host);
+          }
+          break;
+        }
+      }
+      response.setPartitionValues(partitionValues);
```
KaiSernLim, line 41:
```suggestion
      String topic = Version.composeKafkaTopic(storeName, version);
      Partition partition = customizedViewRepository.getPartition(topic);
      if (partition != null) {
        response.addPartition(partition);
      }
```
Does this work? (`addPartition()` would need to be implemented)

clients/da-vinci-client/src/test/java/com/linkedin/davinci/blob/BlobSnapshotManagerTest.java
```diff
@@ -0,0 +1,80 @@
...
+
+  @Test
+  public void testHybridSnapshot() throws RocksDBException {
+    RocksDB mockRocksDB = mock(RocksDB.class);
+    Checkpoint mockCheckpoint = mock(Checkpoint.class);
+    BlobSnapshotManager blobSnapshotManager = spy(new BlobSnapshotManager(BASE_PATH, SNAPSHOT_RETENTION_TIME));
+    doReturn(mockCheckpoint).when(blobSnapshotManager).createCheckpoint(mockRocksDB);
+    doNothing().when(mockCheckpoint)
+        .createCheckpoint(
+            BASE_PATH + "/" + STORE_NAME + "/" + RocksDBUtils.getPartitionDbName(STORE_NAME, PARTITION_ID));
```
KaiSernLim, line 34: This `BASE_PATH + "/" + STORE_NAME + "/" + RocksDBUtils.getPartitionDbName(STORE_NAME, PARTITION_ID)` seems to be used a lot in this file, and it does not depend on anything inside the tests. Perhaps make it a constant at the top?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blob/BlobSnapshotManager.java
```diff
@@ -0,0 +1,116 @@
...
+  // Updates the snapshot of the hybdrid store
+  private void updateHybridSnapshot(RocksDB rocksDB, String topicName, int partitionId) {
+    String fullPathForPartitionDBSnapshot = RocksDBUtils.composeSnapshotDir(this.basePath, topicName, partitionId);
+    if (fullPathForPartitionDBSnapshot.isEmpty()) {
+      return;
+    }
+    try {
+      Checkpoint checkpoint = createCheckpoint(rocksDB);
+
+      LOGGER.info("Start creating snapshots in directory: {}", fullPathForPartitionDBSnapshot);
```
KaiSernLim, line 55:
```suggestion
      LOGGER.info("Started creating snapshots in directory: {}", fullPathForPartitionDBSnapshot);
```
super nit: Either this or just `"Creating snapshots..."`?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blob/BlobSnapshotManager.java
```diff
@@ -0,0 +1,116 @@
...
+    long snapshotTimestamp = snapShotTimestamps.get(topicName).get(partitionId);
+    return System.currentTimeMillis() - snapshotTimestamp > this.snapshotRetentionTime;
+  }
+
+  // Updates the snapshot of the hybdrid store
+  private void updateHybridSnapshot(RocksDB rocksDB, String topicName, int partitionId) {
+    String fullPathForPartitionDBSnapshot = RocksDBUtils.composeSnapshotDir(this.basePath, topicName, partitionId);
+    if (fullPathForPartitionDBSnapshot.isEmpty()) {
+      return;
+    }
```
KaiSernLim, line 51: It doesn't seem like this case is possible, but I do feel like the name of the method does not suggest that nothing happens if the snapshot directory cannot be composed?
I feel like there should at least be a warning or exception in this case.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blob/BlobSnapshotManager.java
```diff
@@ -0,0 +1,116 @@
...
+  // Checks if the current snapshot of the partition is stale
+  private boolean isSnapshotStale(String topicName, int partitionId) {
+    if (snapShotTimestamps.get(topicName) == null || snapShotTimestamps.get(topicName).get(partitionId) == null) {
+      return true;
+    }
+    long snapshotTimestamp = snapShotTimestamps.get(topicName).get(partitionId);
+    return System.currentTimeMillis() - snapshotTimestamp > this.snapshotRetentionTime;
+  }
+
+  // Updates the snapshot of the hybdrid store
```
KaiSernLim, line 46:
```suggestion
  /**
   * Updates the snapshot of the hybrid store
   */
```
nit:

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blob/BlobSnapshotManager.java
```diff
@@ -0,0 +1,116 @@
...
+  public int getConcurrentUsers(RocksDB rocksDB, String topicName, int partitionId) {
+    if (rocksDB == null || topicName == null) {
+      throw new IllegalArgumentException("RocksDB instance and topicName cannot be null");
+    }
+    if (!concurrentSnapshotUsers.containsKey(topicName)) {
+      return 0;
+    }
+    if (!concurrentSnapshotUsers.get(topicName).containsKey(partitionId)) {
+      return 0;
+    }
```
KaiSernLim, line 113: nit: These two cases could be combined. But also, it would probably be better to create a helper for this?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blob/BlobSnapshotManager.java
```diff
@@ -0,0 +1,116 @@
...
+
+  @VisibleForTesting
+  protected Checkpoint createCheckpoint(RocksDB rocksDB) {
+    return Checkpoint.create(rocksDB);
+  }
+
+  private static final Logger LOGGER = LogManager.getLogger(BlobSnapshotManager.class);
+  HashMap<String, Map<Integer, AtomicLong>> concurrentSnapshotUsers;
+  HashMap<String, HashMap<Integer, Long>> snapShotTimestamps;
+  Long snapshotRetentionTime;
```
KaiSernLim, line 27:
```suggestion
  long snapshotRetentionTime;
```
super nit:

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blob/BlobSnapshotManager.java
```diff
@@ -0,0 +1,116 @@
...
+    this.concurrentSnapshotUsers = new HashMap<>();
+  }
+
+  // Checks if the current snapshot of the partition is stale
+  private boolean isSnapshotStale(String topicName, int partitionId) {
+    if (snapShotTimestamps.get(topicName) == null || snapShotTimestamps.get(topicName).get(partitionId) == null) {
+      return true;
+    }
+    long snapshotTimestamp = snapShotTimestamps.get(topicName).get(partitionId);
+    return System.currentTimeMillis() - snapshotTimestamp > this.snapshotRetentionTime;
```
KaiSernLim, line 43:
```suggestion
    long snapshotTimestamp = getSnapshotTimestamp(this.snapshotTimestamps);
    if (snapshotTimestamp == null) {
      return true;
    }
    return System.currentTimeMillis() - snapshotTimestamp > this.snapshotRetentionTime;
```
It feels like you're dancing around the fact that these two `HashMap` data structures `concurrentSnapshotUsers` and `snapShotTimestamps` are 2-dimensional, and individually doing error handling at every juncture.
I feel like this is the best example of when helper methods would be effective. Something like `getSnapshotTimestamp(this.snapshotTimestamps)`, which would either return `null` or throw an exception which would be handled by this function? I feel that this would help this entire file out a lot.

KaiSernLim:
```suggestion
    try {
      long snapshotTimestamp = snapShotTimestamps.get(topicName).get(partitionId);
      return System.currentTimeMillis() - snapshotTimestamp > this.snapshotRetentionTime;
    } except (NoSuchElementException e) {
      return true;
    }
```

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blob/ServerBlobFinder.java
sebas-inf: Managed to implement the abstraction

clients/da-vinci-client/src/test/java/com/linkedin/davinci/blob/BlobSnapshotManagerTest.java
sebas-inf: Done

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blob/BlobSnapshotManager.java
sebas-inf: Should I just do a logge.warn("No path for snapshot provided") or should I throw an error?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blob/BlobSnapshotManager.java
sebas-inf: Combined them since its only used in this case

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blob/BlobFinder.java
```diff
@@ -0,0 +1,9 @@
+package com.linkedin.davinci.blob;
+
```
sixpluszero, line 2: Hmm, why do we duplicate this file here?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blob/BlobResponse.java
```diff
@@ -0,0 +1,51 @@
+package com.linkedin.davinci.blob;
```
sixpluszero, line 1: same here, and below file.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blob/BlobSnapshotManager.java
```diff
@@ -0,0 +1,116 @@
...
+
+public class BlobSnapshotManager {
+  private final String basePath;
+
+  @VisibleForTesting
+  protected Checkpoint createCheckpoint(RocksDB rocksDB) {
+    return Checkpoint.create(rocksDB);
+  }
+
+  private static final Logger LOGGER = LogManager.getLogger(BlobSnapshotManager.class);
```
sixpluszero, line 24: nit: can we move these private and protected method to the bottom. Our format usually goes with constructor method, then public method then protected/default/private.
Also the above visible for testing method don't need to be protected, it can just go with no modifier (default)

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blob/BlobSnapshotManager.java
```diff
@@ -0,0 +1,116 @@
...
+  HashMap<String, Map<Integer, AtomicLong>> concurrentSnapshotUsers;
+  HashMap<String, HashMap<Integer, Long>> snapShotTimestamps;
+  Long snapshotRetentionTime;
+
+  // Initialized the snapshot manger for this specific dbDir
+  public BlobSnapshotManager(String basePath, long snapshotRetentionTime) {
```
sixpluszero, line 32: We should instead take the RocksDB object here during initialization, instead of passing it every time when we call the public method below, because I don't expect it will be very handy for the future BlobTransferManger to pass it in every time a transfer request comes in.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blob/BlobSnapshotManager.java
```diff
@@ -0,0 +1,116 @@
...
+    } catch (RocksDBException e) {
+      throw new VeniceException(
+          "Received exception during RocksDB's snapshot creation in directory " + fullPathForPartitionDBSnapshot,
+          e);
+    }
+  }
+
+  // Checks if the snapshot is stale, if it is and no one is using it, it updates the snapshot, otherwise it increases
+  // the count of people using the snapshot
+  public void getHybridSnapshot(RocksDB rocksDB, String topicName, int partitionId) {
```
sixpluszero, line 79: Nit: let's rename it to something like `maybeUpdateSnapshot`.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blob/BlobSnapshotManager.java
sebas-inf: Implemented the simplification you suggested

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blob/BlobSnapshotManager.java
```diff
@@ -0,0 +1,116 @@
...
+    return System.currentTimeMillis() - snapshotTimestamp > this.snapshotRetentionTime;
+  }
+
+  // Updates the snapshot of the hybdrid store
+  private void updateHybridSnapshot(RocksDB rocksDB, String topicName, int partitionId) {
+    String fullPathForPartitionDBSnapshot = RocksDBUtils.composeSnapshotDir(this.basePath, topicName, partitionId);
+    if (fullPathForPartitionDBSnapshot.isEmpty()) {
+      return;
+    }
+    try {
```
sixpluszero, line 61: If I understand the API correctly, the checkpoint API expect the path to be non exist. If there is old snapshot in the same path, it will fail. We should figure out if the path exist, and then try to delete and recreate the snapshot.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blob/BlobSnapshotManager.java
```diff
@@ -0,0 +1,116 @@
...
+public class BlobSnapshotManager {
+  private final String basePath;
+
+  @VisibleForTesting
+  protected Checkpoint createCheckpoint(RocksDB rocksDB) {
+    return Checkpoint.create(rocksDB);
+  }
+
+  private static final Logger LOGGER = LogManager.getLogger(BlobSnapshotManager.class);
+  HashMap<String, Map<Integer, AtomicLong>> concurrentSnapshotUsers;
```
sixpluszero, line 25: Let's use Map interface, instead of HashMap. Map Should be enough for our case.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blob/BlobSnapshotManager.java
```diff
@@ -0,0 +1,116 @@
...
+  HashMap<String, HashMap<Integer, Long>> snapShotTimestamps;
+  Long snapshotRetentionTime;
+
+  // Initialized the snapshot manger for this specific dbDir
+  public BlobSnapshotManager(String basePath, long snapshotRetentionTime) {
+    this.basePath = basePath;
+    this.snapshotRetentionTime = snapshotRetentionTime;
+    this.snapShotTimestamps = new HashMap<>();
```
sixpluszero, line 35: Here can we use VeniceConcurrentHashMap for these two data structures? It will achieve best possible thread safety. Even for now in the first iteration we can do the thread safety guarding in the synchronized level, it will be good to use this concurrent hash map implementation.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blob/BlobSnapshotManager.java
```diff
@@ -0,0 +1,116 @@
...
+
+  // Checks if the snapshot is stale, if it is and no one is using it, it updates the snapshot, otherwise it increases
+  // the count of people using the snapshot
+  public void getHybridSnapshot(RocksDB rocksDB, String topicName, int partitionId) {
+    if (rocksDB == null || topicName == null) {
+      throw new IllegalArgumentException("RocksDB instance and topicName cannot be null");
+    }
+
+    String fullPathForPartitionDBSnapshot = RocksDBUtils.composeSnapshotDir(this.basePath, topicName, partitionId);
+    if (!snapShotTimestamps.containsKey(topicName)) {
```
sixpluszero, line 84: nit:
Let's use snapshotTimestamps.putIfAbsent() API, same for concurrenSnapshotUser()

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blob/BlobSnapshotManager.java
sebas-inf: Ended up doing a logger error

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blob/BlobSnapshotManager.java
```diff
@@ -0,0 +1,116 @@
...
+      } else {
+        concurrentSnapshotUsers.get(topicName).put(partitionId, new AtomicLong(1));
+      }
+      LOGGER.info("Retrieved preexisting snapshot in directory: {}", fullPathForPartitionDBSnapshot);
+    } catch (Exception e) {
+      throw new VeniceException("Error getting snapshot", e);
+    }
+  }
+
+  public int getConcurrentUsers(RocksDB rocksDB, String topicName, int partitionId) {
```
sixpluszero, line 114: Please remove the rocksDB here.
Let's move all the RocksDB check in the constructor phase.

If you want to get a count, then a better way is call it like this:
private static final AtomicLong EMPTY_COUNTER  = new AtomicLong(0)

concurrentSnapshotUsers.getOrDefault(topicName, Collections.emptyMap()).getOrDefault(partitionId, EMPTY_COUNTER).get()

sixpluszero (commented):
Thanks for the PR, I left some comments. Please feel free to take a look.  I will probably leave a few more later...

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blob/BlobSnapshotManager.java
```diff
@@ -0,0 +1,124 @@
...
+      snapShotTimestamps.put(topicName, new HashMap<>());
+      concurrentSnapshotUsers.put(topicName, new HashMap<>());
+    }
+    if (isSnapshotStale(topicName, partitionId)) {
+      if (concurrentSnapshotUsers.get(topicName).get(partitionId) == null
+          || concurrentSnapshotUsers.get(topicName).get(partitionId).get() == 0) {
+        updateHybridSnapshot(rocksDB, topicName, partitionId);
+        snapShotTimestamps.get(topicName).put(partitionId, System.currentTimeMillis());
+      }
+      if (getConcurrentUsers(rocksDB, topicName, partitionId) == 0) {
```
sixpluszero, line 94: I found there is an issue around this: There is no way we decrease the counter of the snapshot transfer user. We need to think about this. Essentially, BlobTransferManger knows, maybe can we add another API called by BlobTransferManger to invoke to decrease the counter by 1?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blob/BlobSnapshotManager.java
```diff
@@ -0,0 +1,124 @@
...
+        concurrentSnapshotUsers.get(topicName).put(partitionId, new AtomicLong(1));
+      } else {
+        concurrentSnapshotUsers.get(topicName).get(partitionId).incrementAndGet();
```
sixpluszero, line 97: Also, this method does not seem to be thread-safe. But, I'd suggest we address other comments first, and eventually we come back to add the safeguard for thread safety.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blob/BlobSnapshotManager.java
sixpluszero: I am wrong, the RocksDB instance is per topic partition.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blob/BlobResponse.java
```diff
@@ -0,0 +1,51 @@
...
+import java.util.List;
+import java.util.Objects;
+
+
+public class BlobResponse {
+  private boolean isError;
+
+  private String message;
+
+  private List<String> partitionURLs;
```
ishwarya-citro, line 12: What are `partitionURLs`? Are they the urls of the host?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blob/BlobResponse.java
```diff
@@ -0,0 +1,51 @@
+package com.linkedin.davinci.blob;
+
+import java.util.List;
+import java.util.Objects;
+
+
+public class BlobResponse {
+  private boolean isError;
+
+  private String message;
```
ishwarya-citro, line 10: what does `message` contain?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blob/BlobResponse.java
```diff
@@ -0,0 +1,51 @@
...
+
+  public void setMessage(String message) {
+    this.message = message;
+  }
+
+  public String getMessage() {
+    return this.message;
+  }
+
+  public void setPartitionValues(List<String> partitionURLs) {
```
ishwarya-citro, line 30: the method name might be better if it matches the field it is setting to. I.e. `setPartitionURLs`

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blob/BlobResponse.java
```diff
@@ -0,0 +1,51 @@
...
+  public boolean equals(Object obj) {
+    if (obj == this) {
+      return true;
+    }
+    if (!(obj instanceof BlobResponse)) {
+      return false;
+    }
+    BlobResponse other = (BlobResponse) obj;
+    return other.isError == this.isError && Objects.equals(other.message, this.message) && other.partitionURLs != null
+        && other.partitionURLs.equals(this.partitionURLs);
```
ishwarya-citro, line 48: we might get NullPointException during the `partitionURLs` comparisons. Can you look into `Object.equals` which will handle the nulls

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blob/ServerBlobFinder.java
```diff
@@ -0,0 +1,54 @@
...
+import com.linkedin.venice.meta.Partition;
+import com.linkedin.venice.meta.Version;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+
+public class ServerBlobFinder implements BlobFinder {
```
ishwarya-citro, line 15: can we rename to `BlobFinderImpl`?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blob/BlobResponse.java
ishwarya-citro: I see that message contains errorMessages. In that case, can we rename this to `errorMessage`?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blob/ServerBlobFinder.java
```diff
@@ -0,0 +1,54 @@
...
+          break;
+        }
+      }
+      response.setPartitionValues(partitionValues);
+    } catch (VeniceException e) {
+      response.setError(true);
+      String errorMsg =
+          "Error finding blob for store: " + storeName + ", version: " + version + ", partitionId: " + partitionId;
+      response.setMessage(errorMsg + ".\n Error: " + e.getMessage());
+      LOGGER.warn(errorMsg, e);
```
ishwarya-citro, line 47: would this be `LOGGER.error` instead?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blob/ServerBlobFinder.java
```diff
@@ -0,0 +1,54 @@
...
+            String host = instance.getHost();
+            partitionValues.add(host);
+          }
+          break;
+        }
+      }
+      response.setPartitionValues(partitionValues);
+    } catch (VeniceException e) {
+      response.setError(true);
+      String errorMsg =
```
ishwarya-citro, line 44: can we use `String.format` to construct this error message?

ishwarya-citro (changes requested):
Thanks for the PR! 

Also, please remove the duplicate files

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blob/BlobResponse.java
sebas-inf: Made some of those changes in a previous PR

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blob/BlobResponse.java
sebas-inf: its now called setDiscoveryResult

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blob/ServerBlobFinder.java
sebas-inf: made the change

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blob/BlobResponse.java
sebas-inf: eliminated that logic in my previous pr

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blob/BlobSnapshotManager.java
sebas-inf: Implemented the functionality to decrease counter, just not sure how to check if someone is no longer using the snapshot to apply it

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blob/ServerBlobFinder.java
sebas-inf: I believe that we should keep the name as is since it used to be blob finder in the previous PR (PR 1033) but we added the server part since it this deals with the server aspect

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blob/BlobSnapshotManager.java
```diff
@@ -0,0 +1,131 @@
...
+      LOGGER.info("Retrieved preexisting snapshot in directory: {}", fullPathForPartitionDBSnapshot);
+    } catch (Exception e) {
+      throw new VeniceException("Error getting snapshot", e);
+    }
+  }
+
+  public void decreaseConcurrentUserCount(String topicName, int partitionID) {
+    if (concurrentSnapshotUsers.containsKey(topicName)
+        && concurrentSnapshotUsers.get(topicName).containsKey(partitionID)) {
+      if (concurrentSnapshotUsers.get(topicName).get(partitionID).get() > 0) {
```
sixpluszero, line 115: If the logic is correct, we should not call this before we increment the count.
Maybe can you throw a VeniceException if the result is below 0?

Also, a format suggestion is, can we do this to avoid wrapping layers of if:
` someMap = concurrentSnapshotUsers.get(topicName);
 if (someMap == null) {
  throw new VeniceException("xxxx");
 }
 someAtomicLong = someMap.get(partitionId);
 if (someAtomicLong == null) {
  throw new VeniceException("xxxx");
 }
 result = someAtomicLong.decrementAndGet();
 if (result < 0) {
  throw new VeniceException("xxxx");
 }

`

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blob/BlobSnapshotManager.java
sixpluszero: I think given the fact that we believe snapshot operations are fast, we can use a single lock object to guard against the 2 operations: (1) Check and maybe increment the snapshot user count (2) Decrease the snapshot user count.

One example in the codebase is: DiskHealthCheckService. Can you check if you can follow the usage there to use a single lock to guard against the operations?


After you are done, can you write a unit test about this behavior?
In order to test it properly, feel free to add setter/getter method with default modifier to set/get the snapshot map for your testing purpose.

internal/venice-common/src/main/java/com/linkedin/venice/blob/ServerBlobFinder.java
```diff
@@ -26,29 +23,21 @@ public BlobPeersDiscoveryResponse discoverBlobPeers(String storeName, int versio
     BlobPeersDiscoveryResponse response = new BlobPeersDiscoveryResponse();
     try {
       String currentVersionResource = Version.composeKafkaTopic(storeName, version);
-      // Get the partition assignments for the current version
```
sixpluszero, line 29: It looks like there are still a lot of changes here unrelated to this PR, did you rebase it or it is intended?

Especially for this one, as a simple POJO class "BlobPeersDiscoveryResponse", ideally we should only have setter/getter class in that class, we should not plant other complex logic there. Could you share if there is any particular reason to do so? If no, I'd suggest let's revert it back and move the addPartition() logic back into the ServerBlobFinder implementation where it should have such processing logic.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blob/BlobSnapshotManager.java
sixpluszero: As of now, we should not be worrying about whether someone is no longer using this. The caller: BlobTransferManger, is supposed to take care of this and invoke this method when the transfer is completed. We just need to make sure snapshot creation is thread safe.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blob/BlobSnapshotManager.java
```diff
@@ -0,0 +1,131 @@
...
+  protected Checkpoint createCheckpoint(RocksDB rocksDB) {
+    return Checkpoint.create(rocksDB);
+  }
+
+  private static final Logger LOGGER = LogManager.getLogger(BlobSnapshotManager.class);
+
+  /**
+   * Constructor for the BlobSnapshotManager
+   */
+  public BlobSnapshotManager(String basePath, long snapshotRetentionTime) {
```
sixpluszero, line 33: I think we need to make a config in ConfigKeys, but maybe we can leave it for now and do it later when we are integrating this logic into the BlobTransferManager class.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blob/BlobSnapshotManager.java
sixpluszero: By saying 

> BlobTransferManger, is supposed to take care of this

I mean, when we are integrating hybrid snapshot logic into BlobTransferManager, we should make sure it is invoked properly.

sixpluszero (changes requested):
Thank you for your change! I left a few more comments. Please feel free to take a look. For thread-safety, I suggest we start it as simple as we can, using a coarse granularity lock is a good starting point, and if we think performance is a concern, then we should do tune it further.

internal/venice-common/src/main/java/com/linkedin/venice/blob/ServerBlobFinder.java
sebas-inf: Changed logic back

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blob/BlobSnapshotManager.java
sebas-inf: implemented that logic

internal/venice-common/src/main/java/com/linkedin/venice/blob/ServerBlobFinder.java
```diff
@@ -26,29 +23,21 @@ public BlobPeersDiscoveryResponse discoverBlobPeers(String storeName, int versio
...
       }
-      response.setDiscoveryResult(hostNames);
     } catch (VeniceException e) {
       response.setError(true);
-      String errorMsg =
-          "Error finding blob for store: " + storeName + ", version: " + version + ", partitionId: " + partitionId;
-      response.setMessage(errorMsg + ".\n Error: " + e.getMessage());
-      LOGGER.warn(errorMsg, e);
+      String errorMsg = String
+          .format("Error finding blob for store: %s, version: %d, partitionId: %d", storeName, version, partitionId);
```
ishwarya-citro, line 35: Thank you for changing it to `String.format`!

another comment: I think it should be `Error finding peers for blob transfer in store:...` instead of `Error finding blob for store:...`
 (sorry, didn't catch before)

internal/venice-common/src/test/java/com/linkedin/venice/blob/ServerBlobFinderTest.java
```diff
@@ -3,6 +3,8 @@
 import static org.mockito.Mockito.doReturn;
 import static org.mockito.Mockito.mock;
 
+import com.linkedin.venice.blobtransfer.BlobPeersDiscoveryResponse;
```
sixpluszero, line 6: Unused import?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blob/BlobSnapshotManager.java
```diff
@@ -0,0 +1,139 @@
...
+      return 0;
+    }
+    return concurrentSnapshotUsers.get(topicName).get(partitionId).get();
+  }
+
+  /**
+   * Checks if the current snapshot of the partition is stale
+   */
+  private boolean isSnapshotStale(String topicName, int partitionId) {
+    if (!snapShotTimestamps.containsKey(topicName) || !snapShotTimestamps.get(topicName).containsKey(partitionId)) {
```
sixpluszero, line 102: I think we missed one more thing - check if the store is hybrid or not. If the store is batch only, we should not do useless checkpoint update.

What we can do is: Take in a ReadOnlyStoreRepository in the constructor, and in this method, check if store's hybrid config is null or not, if it is null, then it is a batch store, then you should just return false.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blob/BlobSnapshotManager.java
```diff
@@ -0,0 +1,139 @@
...
+   * Constructor for the BlobSnapshotManager
+   */
+  public BlobSnapshotManager(String basePath, long snapshotRetentionTime) {
+    this.basePath = basePath;
+    this.snapshotRetentionTime = snapshotRetentionTime;
+    this.snapShotTimestamps = new VeniceConcurrentHashMap<>();
+    this.concurrentSnapshotUsers = new VeniceConcurrentHashMap<>();
+  }
+
+  public void setSnapShotTimestamps(Map<String, Map<Integer, Long>> snapShotTimestamps) {
```
sixpluszero, line 38: Should this be a public method?
I think this should only be used in test.
If so, could you remove the `public` modifier, and move it down?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blob/BlobSnapshotManager.java
```diff
@@ -0,0 +1,139 @@
...
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.rocksdb.Checkpoint;
+import org.rocksdb.RocksDB;
+import org.rocksdb.RocksDBException;
+
+
+public class BlobSnapshotManager {
```
sixpluszero, line 20: Can you add a simple JavaDoc for this class? Basically sharing what's the functionality and intention of this class doing.  Format would simply be like this:

  /**
   * THE JAVA DOC 
   */

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blob/BlobSnapshotManager.java
sixpluszero: Also, could you add a simple unit test about this? Say if you test against a batch store, then the updateHybridSnapshot invocation count should be 0, even you intentionally set the snapshot TS to be stale.

sixpluszero (commented):
Thank you for your change. Overall it looks much better now. I left 2-3 comments, which should be pretty easy to resolve. Once it is all resolved, PR should be good to go.

internal/venice-common/src/test/java/com/linkedin/venice/blob/ServerBlobFinderTest.java
sebas-inf: No, its used, it also gives me an error if i delete it

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blob/BlobSnapshotManager.java
sebas-inf: will do that

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blob/BlobSnapshotManager.java
```diff
@@ -0,0 +1,139 @@
+package com.linkedin.davinci.blob;
```
adamxchen, line 1: Can you please also move this to a new package `blobtransfer`? I think it's consistent to call it `blobtransfer` everywhere.

sebas-inf:
@adamxchen I think the conflicts may be due to me changing the package names for my previous files, do you know how to resolve the conflicts?

KaiSernLim:
> @adamxchen I think the conflicts may be due to me changing the package names for my previous files, do you know how to resolve the conflicts?

I think the conflicts are due to #1041 being merged. One way you can resolve the conflicts is by merging `main` into your feature branch:

```java
git checkout main
git pull --rebase upstream main
git checkout BlobSnapshotManager
git merge main
```

Personally, I prefer squashing into one commit and then rebasing onto `main` but that may be a lot more complicated.

`ServerBlobFinderTest` is not under the `blobtransfer` package yet btw.

adamxchen:
> > @adamxchen I think the conflicts may be due to me changing the package names for my previous files, do you know how to resolve the conflicts?
> 
> I think the conflicts are due to #1041 being merged. One way you can resolve the conflicts is by merging `main` into your feature branch:
> 
> ```java
> git checkout main
> git pull --rebase upstream main
> git checkout BlobSnapshotManager
> git merge main
> ```
> 
> Personally, I prefer squashing into one commit and then rebasing onto `main` but that may be a lot more complicated.
> 
> `ServerBlobFinderTest` is not under the `blobtransfer` package yet btw.

Yes, please pull the changes from `main` branch as Kai suggested. Given Github allows us to squash the commit by the time of merge, a regular merge could be easier.

sebas-inf:
> > @adamxchen I think the conflicts may be due to me changing the package names for my previous files, do you know how to resolve the conflicts?
> 
> I think the conflicts are due to #1041 being merged. One way you can resolve the conflicts is by merging `main` into your feature branch:
> 
> ```java
> git checkout main
> git pull --rebase upstream main
> git checkout BlobSnapshotManager
> git merge main
> ```
> 
> Personally, I prefer squashing into one commit and then rebasing onto `main` but that may be a lot more complicated.
> 
> `ServerBlobFinderTest` is not under the `blobtransfer` package yet btw.

sixpluszero (approved):
LGTM, thank you!

ishwarya-citro (approved):

PR #1040, opened 2024-06-18
[dvc] Ensure onStartIngestionTask in DaVinciRecordTransformer (DVRT) finishes before put can be executed

kvargha:
## Summary

In the future, when we're going to support rebuilding state on startup from disk or from blob transfer, we need to ensure that rebuilding of state is finished before any puts can be executed. To achieve this, an internal version of DVRT, not exposed to the user, utilizes a CountDownLatch.

## How was this PR tested?

CI


clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -1160,9 +1161,7 @@ private void processIngestionException() {
           pauseConsumption(pubSubTopicPartition.getPubSubTopic().getName(), pubSubTopicPartition.getPartitionNumber());
           LOGGER.info(
               "Memory limit reached. Pausing consumption of topic-partition: {}",
-              Utils.getReplicaId(
-                  pubSubTopicPartition.getPubSubTopic().getName(),
-                  pubSubTopicPartition.getPartitionNumber()));
+              getReplicaId(pubSubTopicPartition.getPubSubTopic().getName(), pubSubTopicPartition.getPartitionNumber()));
```
ZacAttack, line 1164: This is kind of interesting.  Was this clean up done with a linter?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -300,6 +299,7 @@ public abstract class StoreIngestionTask implements Runnable, Closeable {
   protected final ChunkAssembler chunkAssembler;
   private final Optional<ObjectCacheBackend> cacheBackend;
   private final DaVinciRecordTransformer recordTransformer;
+  private InternalDaVinciRecordTransformer internalRecordTransformer;
```
ZacAttack, line 302: Do we even need to downcast like this?  Also, do we need both a reference to recordTransformer and internalRecordTransformer?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
kvargha: Yes. I think my IntelliJ did this automatically. Should I undo it?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
kvargha: I'm unclear what downcast means here, but you're right we only need one reference to recordTransformer.

ZacAttack: downcast in code means that we're using a stricter derived class vs. a base class (DavinciRecordTransformer vs. InternalDavinciRecordTransformer).  It's not technically wrong to do it like this and there are scenarios where it makes sense to do so, but theres nothing about the interface of InternalDVRT that we rely on, and it can make things a bit more inflexible when it comes to mocking/testing/changing the code.

ZacAttack (approved):
looks good

PR #1041, opened 2024-06-19
[common] Add Netty based P2P lib for blob transfer

adamxchen:
## Summary

This PR introduces the Netty-based client and server to transfer files among peers. 
* It's still a bit rough and needs some improvements (bunch of TODOs left) to provide better monitoring and tune the performance. 
* This PR uses NIO implementation as much as possible to leverage zero-copy. We need to do some performance testing to find out how's that compared to regular I/O implementation. 
* One side note is zero copy seems require non-SSL so it's likely we don't get every benefit from it if using in production where SSL is always required, but the chance is we can still get something out of it 🤞 
* Moved some codes from `blob` to `blobtransfer` @sebas-inf 

The basic workflow of the client and server is described below:
1.  Client connects to the server and sends GET request to server to request blobs.
2. Server validates the request.
3. If request is valid, server will be sending files to client in chunks (default is 8192, and we can tune it if necessary).
4. For the last chunk, server sends a `DefaultLastHttpContent` and `EMPTY_LAST_CONTENT` to indicate the end of a file transfer so when client receives `DefaultLastHttpContent`, it would know it's the end of file and would reset state then prepare for the next file. (One thing unclear to me is how's `EMPTY_LAST_CONTENT` is handled by some internal Netty stuff because it's not actually passed to `P2PFileTransferClientHandler`)
5. After every file has been transferred, the server sends a final response with a special header telling the client the transfer is done
6. Client completes the future and close the connection

## How was this PR tested?

CI


internal/venice-common/src/main/java/com/linkedin/venice/blobtransfer/P2PBlobTransferManager.java
```diff
@@ -21,4 +23,8 @@ public interface P2PBlobTransferManager<T> extends BlobTransferManager<T> {
   default CompletionStage<T> put(String storeName, int version, int partition) {
     throw new VeniceException("PUT method is not supported for P2P Blob Transfer");
   }
+
+  CompletionStage<? extends InputStream> get(String storeName, int version, int partition)
```
adamxchen, line 27: this is useless. I don't know why I have it at the first place. Will remove it in next reivision

internal/venice-common/src/test/java/com/linkedin/venice/blobtransfer/client/TestP2PFileTransferClientHandler.java
```diff
@@ -0,0 +1,208 @@
...
+
+  @Test
+  public void testFileSizeMismatchTransfer() {
+    // response
+    DefaultHttpResponse response = new DefaultHttpResponse(HttpVersion.HTTP_1_1, HttpResponseStatus.OK);
+    response.headers().add("Content-Disposition", "filename=\"test_file.txt\"");
+    response.headers().add("Content-Length", "5");
+    // content 1
+    // length 1
+    HttpContent chunk1 = new DefaultHttpContent(Unpooled.copiedBuffer("0", CharsetUtil.UTF_8));
```
ishwarya-citro, line 102: can we re-write this comment, something like `creates http content with actual file data of length 1`

internal/venice-common/src/test/java/com/linkedin/venice/blobtransfer/client/TestP2PFileTransferClientHandler.java
```diff
@@ -0,0 +1,208 @@
...
+    } catch (Exception e) {
+      Assert.assertTrue(e.getCause() instanceof VeniceException);
+      Assert.assertEquals(e.getCause().getMessage(), "No file name specified in the response for test_store_v1-0");
+    }
+  }
+
+  @Test
+  public void testFileSizeMismatchTransfer() {
+    // response
+    DefaultHttpResponse response = new DefaultHttpResponse(HttpVersion.HTTP_1_1, HttpResponseStatus.OK);
```
ishwarya-citro, line 97: can we re-write comment,  ie. `creates http response with expected content length of 5`

ishwarya-citro: same w/ other tests

ishwarya-citro (dismissed):

internal/venice-common/src/main/java/com/linkedin/venice/blobtransfer/client/P2PFileTransferClientHandler.java
```diff
@@ -0,0 +1,166 @@
...
+import java.nio.file.Paths;
+import java.nio.file.StandardOpenOption;
+import java.util.concurrent.CompletionStage;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+
+public class P2PFileTransferClientHandler extends SimpleChannelInboundHandler<HttpObject> {
```
adamxchen, line 33: This one is not threadsafe. Working on some changes now

internal/venice-common/src/main/java/com/linkedin/venice/blobtransfer/NettyP2PBlobTransferManager.java
```diff
@@ -0,0 +1,72 @@
+package com.linkedin.venice.blobtransfer;
```
sebas-inf, line 1: I made a few modifications that ishwarya asked for in the blob transfer file when it was in the blob package, when i push will it make a copy in the blob package that will then need to be transferred to the blobtransfer package or will we have to manually move it to the blob transfer package after my commit is merged?

internal/venice-common/src/main/java/com/linkedin/venice/blobtransfer/client/P2PFileTransferClientHandler.java
sebas-inf: What things make this file not thread safe and what should be done to make it thread safe?

internal/venice-common/src/main/java/com/linkedin/venice/blobtransfer/NettyP2PBlobTransferManager.java
adamxchen: This PR is not merged yet so there's no conflict yet. However, if you want to move your files from `blob` to `blobtransfer` manually in your PR, it'd be good too. I can pull your change and resolve conflicts if any.

sixpluszero (commented):
Thanks for the change. I left a few comments that I did this morning. Regarding Netty part I admit I forget almost everything about it, I need to take a closer look later.

internal/venice-common/src/main/java/com/linkedin/venice/blobtransfer/NettyP2PBlobTransferManager.java
```diff
@@ -0,0 +1,72 @@
...
+    BlobPeersDiscoveryResponse response = peerFinder.discoverBlobPeers(storeName, version, partition);
+    if (response == null || response.isError()) {
+      throw new VenicePeersNotFoundException("Failed to obtain the peers for the requested blob");
+    }
+    List<String> discoverPeers = response.getDiscoveryResult();
+    if (discoverPeers == null || discoverPeers.isEmpty()) {
+      throw new VenicePeersNotFoundException("No peers found for the requested blob");
+    }
+    try {
+      String chosenHost = discoverPeers.get(0);
```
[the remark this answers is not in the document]
adamxchen: Yes, I've actually left one todo in the `catch` block. I can move it up.

internal/venice-common/src/main/java/com/linkedin/venice/blobtransfer/NettyP2PBlobTransferManager.java
sebas-inf: sounds good, I'll do that now just to be safe

internal/venice-common/src/main/java/com/linkedin/venice/blobtransfer/NettyP2PBlobTransferManager.java
```diff
@@ -0,0 +1,72 @@
...
+      inputStream = nettyClient.get(chosenHost, storeName, version, partition);
+    } catch (InterruptedException e) {
+      LOGGER.error("Failed to fetch file from remote peer. Error: {}", e.getMessage());
+      // TODO: add some retry logic
+      throw new StorePartitionNotFoundException(storeName, version, partition);
```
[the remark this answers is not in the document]
adamxchen: It makes sense. The `InterruptedException` will be thrown if the client fails to connect to the server so the `StorePartitionNotFoundException` is not appropriate. I'd change it to `VenicePeersNotFoundException` since failure to connect is a part of `PeersNotFound` issue I suppose. Please lmk if you think it still doesn't sound right.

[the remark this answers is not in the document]
sixpluszero: yeah for now I guess it is fine. But maybe we will revisit it when we are adding throttling logic into it, but maybe it is still valid by then.

internal/venice-common/src/main/java/com/linkedin/venice/blobtransfer/NettyP2PBlobTransferManager.java
```diff
@@ -0,0 +1,72 @@
...
+  @Override
+  public void start() throws Exception {
+    blobTransferService.start();
+  }
+
+  @Override
+  public CompletionStage<InputStream> get(String storeName, int version, int partition)
+      throws VeniceNoStoreException, VenicePeersNotFoundException {
+    CompletionStage<InputStream> inputStream;
+    BlobPeersDiscoveryResponse response = peerFinder.discoverBlobPeers(storeName, version, partition);
```
ishwarya-citro, line 47: FYI - this is not getting from the router; although, it doesn't matter in this PR. I can send out a new one

sixpluszero (approved):
Overall the functionality part looks good to me! But since it is lib PR so there seems to be no integration test. Maybe in the follow up integration test we can use this to do a real e2e transfer from one DVC to the other.

ishwarya-citro (approved):
Thank you for the PR!

adamxchen:
> Overall the functionality part looks good to me! But since it is lib PR so there seems to be no integration test. Maybe in the follow up integration test we can use this to do a real e2e transfer from one DVC to the other.

yes, that's the goal for this Q.

PR #1042, opened 2024-06-20
[controller][server][client][test] Adds a new checksum type ADHASH for data integrity validation

lluwm:
[controller][server][client][test] Adds a new checksum type ADHASH for data integrity validation

This PR adds a new checksum type ADHASH for Venice's DIV. ADHASH adopts the
ideal of incremental hash from AdHash and leverages the CRC32 as the underlying
hashing algorithm.

CRC32 is commonly used in networks and storage devices for detecting data corruption.
The purpose of this algorithm is not to protect against intentional
changes (cryptographic hashing), but rather to catch accidents like network errors
and disk write errors, etc. The emphasis of CRC32 is more on speed than on security.

For testing, this PR extends all the existing checksum and DIV tests to include the
new ADHASH type.

## How was this PR tested?
- Extends existing related tests to include the new type.
- Enabled `ADHASH` as the default `DEFAULT_CHECK_SUM_TYPE` in VeniceWriter for testing and all tests were [passed](https://github.com/lluwm/venice/commit/86931b881513a8c05e2f33f5ddb819cf0a115ffb). 

### DIFF
--- a/internal/venice-common/src/main/java/com/linkedin/venice/kafka/validation/checksum/AdHash.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/kafka/validation/checksum/AdHash.java
@@ -5,7 +5,7 @@ import java.util.zip.CRC32;
 
 
 /**
- * This incremental digest is used to keep track of the checksum of the data incrementally.
+ * This class AdHash is used to keep track of the checksum of the data incrementally.
  * It adopts the idea of AdHash that:
  * " A message x' which I want to hash may be a simple modification of a message x which I previously
  *   hashed. If I have already computed the hash f(x) of x then, rather than re-computing f(x') from
@@ -21,15 +21,15 @@ import java.util.zip.CRC32;
  *   (https://dl.acm.org/doi/abs/10.1145/3578358.3591328)
  */
 
-public class IncDigestCheckSum extends CheckSum {
+public class AdHash extends CheckSum {
   private volatile long incrementalDigest;
   private final CRC32 crc32 = new CRC32();
 
-  public IncDigestCheckSum() {
+  public AdHash() {
     incrementalDigest = 0;
   }
 
-  public IncDigestCheckSum(byte[] encodedState) {
+  public AdHash(byte[] encodedState) {
     incrementalDigest = ByteUtils.readLong(encodedState, 0);
   }
 
@@ -63,7 +63,7 @@ public class IncDigestCheckSum extends CheckSum {
 
   @Override
   public CheckSumType getType() {
-    return CheckSumType.INCDIGEST;
+    return CheckSumType.ADHASH;
   }
 
   @Override
--- a/internal/venice-common/src/main/java/com/linkedin/venice/kafka/validation/checksum/IncDigestCheckSum.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/kafka/validation/checksum/IncDigestCheckSum.java
@@ -1,0 +1,73 @@
+package com.linkedin.venice.kafka.validation.checksum;
+
+import com.linkedin.venice.utils.ByteUtils;
+import java.util.zip.CRC32;
+
+
+/**
+ * This incremental digest is used to keep track of the checksum of the data incrementally.
+ * It adopts the idea of AdHash that:
+ * " A message x' which I want to hash may be a simple modification of a message x which I previously
+ *   hashed. If I have already computed the hash f(x) of x then, rather than re-computing f(x') from
+ *   scratch, I would like to just quickly 'update' the old hash value f(x) to the new value f(x').
+ *   An incremental hash function is one that permits this. "
+ *
+ * See the paper for details:
+ *   A New Paradigm for Collision-free Hashing: Incrementality at Reduced Cost.
+ *   (https://cseweb.ucsd.edu/~daniele/papers/IncHash.pdf)
+ *
+ * Its adoption in ZooKeeper:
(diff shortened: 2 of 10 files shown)


internal/venice-common/src/main/java/com/linkedin/venice/kafka/validation/checksum/IncDigestCheckSum.java
sushantmane, line 1: `IncDigestCheckSum` the name is bit redundant IMO.  Can we call rename it to `AdHash`?

lluwm: We can surely do that. Not good at naming at all.

sushantmane (commented):
Thanks, @lluwm! 
Should we add the following E2E tests:
1. Set checksum type to `ADHASH` in VPJ and run a full push
2. Set checksum type to `ADHASH` on servers and then write RT data with different writers. One writer will use to `ADHASH`  and another one will use `MD5`

lluwm:
> Thanks, @lluwm! Changes LGTM! Should we add the following E2E tests:
> 
> 1. Set checksum type to `ADHASH` in VPJ and run a full push
> 2. Set checksum type to `ADHASH` on servers and then write RT data with different writers. One writer will use to `ADHASH`  and another one will use `MD5`

Yeah, we can add them for catching any regressions in the future and the 2nd one sounds a bit more interesting. Just for the information, I mentioned in the testing part that I have tested it more throughly by enabling it as the default checksum and passing all our existing tests, but for catching regression, it makes sense to add some E2E tests.

sushantmane:
> ust for the information, I mentioned in the testing part that I have tested it more throughly by enabling it as the default checksum and passing all our existing tests, but for catching regression, it makes sense to add some E2E tests.

Thanks, @lluwm! IMO it would be good to have  E2E tests that will executed on each CI run.

internal/venice-common/src/main/java/com/linkedin/venice/kafka/validation/checksum/AdHash.java
```diff
@@ -0,0 +1,73 @@
...
+ *   A New Paradigm for Collision-free Hashing: Incrementality at Reduced Cost.
+ *   (https://cseweb.ucsd.edu/~daniele/papers/IncHash.pdf)
+ *
+ * Its adoption in ZooKeeper:
+ *   Verify, And Then Trust: Data Inconsistency Detection in ZooKeeper.
+ *   (https://dl.acm.org/doi/abs/10.1145/3578358.3591328)
+ */
+
+public class AdHash extends CheckSum {
+  private volatile long incrementalDigest;
```
sushantmane, line 25: Does it need to be volatile?

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/CheckSumTest.java
sushantmane, line 1: I think with this particular test, we're testing two different checksums on two different stores. My suggestion was to have just one store with RT producers, using AdHash on one and CRC32 on the other.

internal/venice-common/src/main/java/com/linkedin/venice/kafka/validation/checksum/AdHash.java
lluwm: We can read this value in a different thread e.g. SIT thread. Please check `StoreIngestionTask.checkLongRunningTaskState`

sushantmane: In `SIT::cloneProducerStates`?

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/CheckSumTest.java
sushantmane: > Essentially, what we want to test is that even if a server is configured with AdHash checksum, it still capable of handing MD5 and AdHash messages correctly. Is my understanding right?
Correct but this is just one case. The second case is where a store is getting data with two diff  checksum algos and still works without any issues (esp when servers subscribe to RT topics from the middle of segment/non-sos message)

internal/venice-common/src/main/java/com/linkedin/venice/kafka/validation/checksum/AdHash.java
lluwm: correct.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/CheckSumTest.java
lluwm: My thinking is that checksum works at the segment level (from SOS to EOS), so on the server side, it doesn't really matter if the two segments are from the same RT or different RTs. I feel what's important is that it is able to notice segment with different checksum types and continue to verify them with the checksum it indicates in the message correctly. So that why I think essentially first case and the second case should be the same and it's fine to verify either of them.

However, that being said, we can easily add the second case if you prefer. Shouldn't be very hard given what we have now.

sushantmane: Yeah, let's include the second case as it addresses both scenarios: a single store ingesting data with two different checksum type RT producers, and the server's capability to handle two types of checksums simultaneously.

sushantmane (approved):
LGTM! Thanks, @lluwm!

PR #1043, opened 2024-06-20
[server] Do not consult data replication policy for lag measurement if store is AA enabled

sushantmane:
## Do not consult data replication policy for lag measurement if store is AA enabled

For AA-enabled stores, it is crucial to consider lag from all fabrics to avoid premature version  
swaps in some colos. This PR removes any dependency on DRP for lag measurement if store 
 is AA enabled.

## How was this PR tested?

CI

### DIFF
--- a/internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestActiveActiveReplicationForIncPush.java
+++ b/internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestActiveActiveReplicationForIncPush.java
@@ -13,7 +13,10 @@ import static com.linkedin.venice.hadoop.VenicePushJobConstants.DEFAULT_VALUE_FI
 import static com.linkedin.venice.hadoop.VenicePushJobConstants.INCREMENTAL_PUSH;
 import static com.linkedin.venice.hadoop.VenicePushJobConstants.SEND_CONTROL_MESSAGES_DIRECTLY;
 import static com.linkedin.venice.hadoop.VenicePushJobConstants.SOURCE_GRID_FABRIC;
+import static com.linkedin.venice.meta.DataReplicationPolicy.ACTIVE_ACTIVE;
+import static com.linkedin.venice.meta.DataReplicationPolicy.NON_AGGREGATE;
 import static com.linkedin.venice.utils.TestUtils.assertCommand;
+import static com.linkedin.venice.utils.TestUtils.createAndVerifyStoreInAllRegions;
 import static com.linkedin.venice.utils.TestUtils.waitForNonDeterministicAssertion;
 import static com.linkedin.venice.utils.TestWriteUtils.getTempDataDirectory;
 import static org.testng.Assert.assertEquals;
@@ -40,9 +43,12 @@ import com.linkedin.venice.utils.TestWriteUtils;
 import com.linkedin.venice.utils.Time;
 import com.linkedin.venice.utils.Utils;
 import java.io.File;
+import java.util.LinkedHashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.Properties;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Function;
 import org.apache.avro.Schema;
@@ -61,14 +67,16 @@ public class TestActiveActiveReplicationForIncPush {
 
   private static final int NUMBER_OF_CHILD_DATACENTERS = 3;
   private static final int NUMBER_OF_CLUSTERS = 1;
+  private static final String PARENT = "parent";
   private String[] clusterNames;
   private String parentRegionName;
   private String[] dcNames;
 
   private List<VeniceMultiClusterWrapper> childDatacenters;
   private VeniceTwoLayerMultiRegionMultiClusterWrapper multiRegionMultiClusterWrapper;
-
-  PubSubBrokerWrapper veniceParentDefaultKafka;
+  private PubSubBrokerWrapper veniceParentDefaultKafka;
+  private Map<String, ControllerClient> controllerClients;
+  private String clusterName;
 
   @BeforeClass(alwaysRun = true)
   public void setUp() {
@@ -107,10 +115,27 @@ public class TestActiveActiveReplicationForIncPush {
     dcNames = multiRegionMultiClusterWrapper.getChildRegionNames().toArray(new String[0]);
 
     veniceParentDefaultKafka = multiRegionMultiClusterWrapper.getParentKafkaBrokerWrapper();
+
+    String parentControllerUrls = multiRegionMultiClusterWrapper.getControllerConnectString();
+    Function<Integer, String> connectionString = i -> childDatacenters.get(i).getControllerConnectString();
+    clusterName = this.clusterNames[0];
+    ControllerClient parentControllerClient = new ControllerClient(clusterName, parentControllerUrls);
+    ControllerClient dc0ControllerClient = new ControllerClient(clusterName, connectionString.apply(0));
+    ControllerClient dc1ControllerClient = new ControllerClient(clusterName, connectionString.apply(1));
+    ControllerClient dc2ControllerClient = new ControllerClient(clusterName, connectionString.apply(2));
+    controllerClients = new LinkedHashMap<>(4);
+    controllerClients.put(PARENT, parentControllerClient);
+    controllerClients.put(dcNames[0], dc0ControllerClient);
+    controllerClients.put(dcNames[1], dc1ControllerClient);
+    controllerClients.put(dcNames[2], dc2ControllerClient);
   }
 
   @AfterClass(alwaysRun = true)
   public void cleanUp() {
+    for (ControllerClient controllerClient: controllerClients.values()) {
+      controllerClient.close();
+    }
+
     multiRegionMultiClusterWrapper.close();
   }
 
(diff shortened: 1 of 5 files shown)


services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
```diff
@@ -2423,16 +2423,18 @@ public void updateStore(String clusterName, String storeName, UpdateStoreQueryPa
...
-          LOGGER.info(
-              "Data replication policy was not explicitly set when converting store to hybrid store: {}."
-                  + " Setting it to active-active replication policy.",
-              storeName);
+      }
 
-          updatedHybridStoreConfig.setDataReplicationPolicy(DataReplicationPolicy.ACTIVE_ACTIVE);
-          updatedConfigsList.add(DATA_REPLICATION_POLICY);
-        }
+      if (setStore.activeActiveReplicationEnabled && !hybridBufferReplayPolicy.isPresent()) {
```
huangminchn, line 2428: Why change the condition from `!hybridDataReplicationPolicy.isPresent()` to `!hybridBufferReplayPolicy.isPresent()`? I don't understand this part; was it a bug? Could you please share some details in the PR description?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
```diff
@@ -2423,16 +2423,18 @@ public void updateStore(String clusterName, String storeName, UpdateStoreQueryPa
...
-          LOGGER.info(
-              "Data replication policy was not explicitly set when converting store to hybrid store: {}."
-                  + " Setting it to active-active replication policy.",
-              storeName);
+      }
 
-          updatedHybridStoreConfig.setDataReplicationPolicy(DataReplicationPolicy.ACTIVE_ACTIVE);
-          updatedConfigsList.add(DATA_REPLICATION_POLICY);
-        }
+      if (setStore.activeActiveReplicationEnabled && !hybridDataReplicationPolicy.isPresent()) {
```
sixpluszero, line 2428: Could you add a simple integration test to guard against this behavior change?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
sushantmane: That was mistake. Updated the commit message. Thanks!

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
sushantmane: sure

sushantmane: Done

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
```diff
@@ -2413,50 +2413,89 @@ public void updateStore(String clusterName, String storeName, UpdateStoreQueryPa
...
-      }
       // When turning off hybrid store, we will also turn off incremental store config.
       if (storeBeingConvertedToBatch && setStore.incrementalPushEnabled) {
         setStore.incrementalPushEnabled = false;
         updatedConfigsList.add(INCREMENTAL_PUSH_ENABLED);
       }
 
+      /*
+       * If active-active replication mode is being disabled, and store is still a hybrid store,
+       * the store is not a system store, then
```
huangminchn, line 2478: Do we need to check system store in this case?

sushantmane: Yes we do and it's part of the condition

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestActiveActiveReplicationForIncPush.java
```diff
@@ -120,138 +145,110 @@ public void cleanUp() {
...
+      job.run();
+      Assert.assertEquals(job.getKafkaUrl(), childDatacenters.get(2).getKafkaBrokerWrapper().getAddress());
     }
+    // Run inc push with source fabric preference taking effect.
+    try (VenicePushJob job = new VenicePushJob("Test push job incremental with NR + A/A from dc-2", propsInc1)) {
+      job.run();
+      Assert.assertEquals(job.getKafkaUrl(), childDatacenters.get(2).getKafkaBrokerWrapper().getAddress());
+    }
+
+    // Verify
```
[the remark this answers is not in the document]
sushantmane: There are no changes to this code path. Not sure why it shows changes here but not in diff mode

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
```diff
@@ -2413,50 +2413,89 @@ public void updateStore(String clusterName, String storeName, UpdateStoreQueryPa
...
+      // Update active-active replication config based on the input if present, otherwise use the current value.
       setStore.activeActiveReplicationEnabled = activeActiveReplicationEnabled
           .map(addToUpdatedConfigList(updatedConfigsList, ACTIVE_ACTIVE_REPLICATION_ENABLED))
           .orElseGet(currStore::isActiveActiveReplicationEnabled);
-      // Enable active-active replication automatically when batch user store being converted to hybrid store and
-      // active-active replication is enabled for all hybrid store via the cluster config
-      if (storeBeingConvertedToHybrid && !setStore.activeActiveReplicationEnabled && !currStore.isSystemStore()
+      // Update incremental push config.
+      setStore.incrementalPushEnabled =
+          incrementalPushEnabled.map(addToUpdatedConfigList(updatedConfigsList, INCREMENTAL_PUSH_ENABLED))
```
sixpluszero, line 2423: I am wondering, if you move it here, would it be the case that it will enable inc push if admin tool ask it to do so, regardless of A/A or hybrid config it has? Should we allow this?

sushantmane: True. It will also allow us to specifically disable IncPush and keep AA.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
huangminchn: I mean whether we should remove the system store check; I think if we are disabling AA for system stores, we should update the DRP config too.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
```diff
@@ -2413,50 +2413,89 @@ public void updateStore(String clusterName, String storeName, UpdateStoreQueryPa
...
+          && updatedHybridStoreConfig != null && !currStore.isSystemStore()) {
+        if (!hybridDataReplicationPolicy.isPresent()
+            && updatedHybridStoreConfig.getDataReplicationPolicy() == DataReplicationPolicy.ACTIVE_ACTIVE) {
+          LOGGER.info(
+              "Disabling active-active replication mode for store: {}. Setting data replication policy to NON_AGGREGATE",
+              storeName);
+          updatedHybridStoreConfig.setDataReplicationPolicy(DataReplicationPolicy.NON_AGGREGATE);
+          updatedConfigsList.add(DATA_REPLICATION_POLICY);
+        }
+        if (setStore.incrementalPushEnabled && clusterConfig.enabledIncrementalPushForHybridActiveActiveUserStores()) {
```
sixpluszero, line 2492: I think if currentStore.incrementalPushEnabled is enough right?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
sushantmane: We can but IDK if we should 😄

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
sushantmane: If we want to ensure that both AA and IncPush are fully dependent on one another, regardless of the `enabledIncrementalPushForHybridActiveActiveUserStores` configuration, then yes.

sixpluszero: Yeah we should ditch the clusterconfig, but what I want to say is should we look at currentStore, instead of setStore.

sushantmane: IMO setStore.incrementalPushEnabled captures both new and old state
```
      setStore.incrementalPushEnabled =
          incrementalPushEnabled.map(addToUpdatedConfigList(updatedConfigsList, INCREMENTAL_PUSH_ENABLED))
              .orElseGet(currStore::isIncrementalPushEnabled);
```

huangminchn (approved):
Thanks a lot for cleaning up the tech debt, Sushant!

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
sushantmane: This code was removed.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
sushantmane: This change was removed.

PR #1044, opened 2024-06-21
[router] Fix NPE in ScatterGatherRequestHandlerImpl

atcurtis:

- CompletableFuture.thenCompose requires functions that do not return null.
- The return from findPartitionName is a CompletionStage (async).
- dedup partition names after retrieving partition names for each key.
- only call appendError once when there are multiple partition keys.

Resolves #XXX

## How was this PR tested?
Not yet tested ;-)

### DIFF
--- a/internal/alpini/router/alpini-router-base/src/main/java/com/linkedin/alpini/router/ScatterGatherRequestHandlerImpl.java
+++ b/internal/alpini/router/alpini-router-base/src/main/java/com/linkedin/alpini/router/ScatterGatherRequestHandlerImpl.java
@@ -936,8 +936,7 @@ public abstract class ScatterGatherRequestHandlerImpl<H, P extends ResourcePath<
       complete = complete.thenCompose(aVoid -> {
         List<CompletableFuture<String>> list = new ArrayList(part.getPartitionKeys().size());
         for (K partitionKey : part.getPartitionKeys()) {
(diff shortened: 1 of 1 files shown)


internal/alpini/router/alpini-router-base/src/main/java/com/linkedin/alpini/router/ScatterGatherRequestHandlerImpl.java
```diff
@@ -932,27 +933,32 @@ protected CompletionStage<?> appendErrorForEveryKey(
...
-                  responses,
-                  status,
-                  contentMsg.append(", RoutingPolicy=").append(roles).toString(),
-                  ex);
-              return null;
-            });
-      }
+      complete = complete.thenCompose(aVoid -> {
+        List<CompletableFuture<String>> list = part.getPartitionKeys()
+            .stream()
```
nisargthakkar, line 938: Can we please change to a non-streaming code? We want to avoid it in the hotpath due to the high amount of garbage it generates

atcurtis: redone without `stream()`

ZacAttack (approved):
Looks good!

PR #1045, opened 2024-06-21
[controller] Deprecate ACTIVE_ACTIVE and NONE data replication policies 

sushantmane:
# Deprecate ACTIVE_ACTIVE data replication policy

Deprecate the ACTIVE_ACTIVE data replication policy, as it is no longer necessary  
for lag measurements. Additionally, deprecate the NONE data replication policy.

## How was this PR tested?

CI

### DIFF
--- a/internal/venice-common/src/main/java/com/linkedin/venice/meta/DataReplicationPolicy.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/meta/DataReplicationPolicy.java
@@ -6,8 +6,9 @@ import java.util.Map;
 
 
 /**
- * Enums of the policies used to decide how real-time samza data is replicated.
- * Note: Data replication policy does not play any role for AA enabled hybrid stores.
+ * The data replication policy determines which fabric to use for sending real-time writes from clients and
+ * where to direct servers for real-time data consumption. If active-active replication is enabled,
+ * the data replication policy is bypassed, as writes are sent to the local fabric.
  */
 public enum DataReplicationPolicy {
   /**
--- a/internal/venice-common/src/main/java/com/linkedin/venice/meta/DataReplicationPolicy.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/meta/DataReplicationPolicy.java
@@ -30,6 +30,7 @@ public enum DataReplicationPolicy {
    * Samza job per colo pushes to local real-time topic. Leader SNs replicate data from real-time topic in all colos to
    * local version topic.
    */
+  @Deprecated
   ACTIVE_ACTIVE(3);
 
   private final int value;
--- a/internal/venice-common/src/main/java/com/linkedin/venice/meta/DataReplicationPolicy.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/meta/DataReplicationPolicy.java
@@ -7,6 +7,7 @@ import java.util.Map;
 
 /**
  * Enums of the policies used to decide how real-time samza data is replicated.
+ * Note: Data replication policy does not play any role for AA enabled hybrid stores.
  */
 public enum DataReplicationPolicy {
   /**
@@ -19,17 +20,9 @@ public enum DataReplicationPolicy {
    */
   AGGREGATE(1),
 
-  /**
-   * This enum value is used in 2 cases:
-   *   1. batch-only stores since it has no real-time data replication.
-   *   2. incremental push stores that do not have Samza jobs.
-   */
+  @Deprecated
   NONE(2),
 
-  /**
-   * Samza job per colo pushes to local real-time topic. Leader SNs replicate data from real-time topic in all colos to
-   * local version topic.
-   */
   @Deprecated
   ACTIVE_ACTIVE(3);
 
--- a/internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/DataRecoveryTest.java
+++ b/internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/DataRecoveryTest.java
@@ -311,7 +311,6 @@ public class DataRecoveryTest {
                   storeName,
                   new UpdateStoreQueryParams().setHybridRewindSeconds(10)
                       .setHybridOffsetLagThreshold(2)
-                      .setHybridDataReplicationPolicy(DataReplicationPolicy.NON_AGGREGATE)
                       .setNativeReplicationEnabled(true)
                       .setActiveActiveReplicationEnabled(true)
                       .setPartitionCount(1))
--- a/internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/DataRecoveryTest.java
+++ b/internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/DataRecoveryTest.java
@@ -311,7 +311,7 @@ public class DataRecoveryTest {
                   storeName,
                   new UpdateStoreQueryParams().setHybridRewindSeconds(10)
                       .setHybridOffsetLagThreshold(2)
-                      .setHybridDataReplicationPolicy(DataReplicationPolicy.ACTIVE_ACTIVE)
+                      .setHybridDataReplicationPolicy(DataReplicationPolicy.NON_AGGREGATE)
                       .setNativeReplicationEnabled(true)
                       .setActiveActiveReplicationEnabled(true)
                       .setPartitionCount(1))
--- a/internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/ingestionHeartbeat/IngestionHeartBeatTest.java
+++ b/internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/ingestionHeartbeat/IngestionHeartBeatTest.java
@@ -120,11 +120,8 @@ public class IngestionHeartBeatTest {
 
   @DataProvider
   public static Object[][] AAConfigAndIncPushAndDRPProvider() {
-    return DataProviderUtils.allPermutationGenerator(
-        (permutation) -> true,
-        DataProviderUtils.BOOLEAN,
-        DataProviderUtils.BOOLEAN,
-        DataReplicationPolicy.values());
+    return DataProviderUtils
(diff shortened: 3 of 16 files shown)


internal/venice-common/src/main/java/com/linkedin/venice/meta/DataReplicationPolicy.java
```diff
@@ -30,6 +30,7 @@ public enum DataReplicationPolicy {
    * Samza job per colo pushes to local real-time topic. Leader SNs replicate data from real-time topic in all colos to
    * local version topic.
    */
+  @Deprecated
   ACTIVE_ACTIVE(3);
```
huangminchn, line 34: Removing this should be fine for ZK. The problem might be in the historical update-store commands in the admin topics.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/DataRecoveryTest.java
```diff
@@ -311,7 +311,7 @@ public void testHybridAADataRecovery() throws Exception {
                   storeName,
                   new UpdateStoreQueryParams().setHybridRewindSeconds(10)
                       .setHybridOffsetLagThreshold(2)
-                      .setHybridDataReplicationPolicy(DataReplicationPolicy.ACTIVE_ACTIVE)
+                      .setHybridDataReplicationPolicy(DataReplicationPolicy.NON_AGGREGATE)
```
huangminchn, line 314: Why not use NONE?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
```diff
@@ -2423,15 +2422,6 @@ public void updateStore(String clusterName, String storeName, UpdateStoreQueryPa
...
         updatedConfigsList.add(ACTIVE_ACTIVE_REPLICATION_ENABLED);
-        if (!hybridDataReplicationPolicy.isPresent()) {
-          LOGGER.info(
-              "Data replication policy was not explicitly set when converting store to hybrid store: {}."
-                  + " Setting it to active-active replication policy.",
-              storeName);
-
-          updatedHybridStoreConfig.setDataReplicationPolicy(DataReplicationPolicy.ACTIVE_ACTIVE);
-          updatedConfigsList.add(DATA_REPLICATION_POLICY);
-        }
```
huangminchn, line 2434: Great! Good to see this clean up!

huangminchn (commented):
Thanks for the efforts on cleaning up the codes!

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/DataRecoveryTest.java
sushantmane: The following check won't pass with NONE.
```
    if (pushType.equals(PushType.STREAM)
        && store.getHybridStoreConfig().getDataReplicationPolicy().equals(DataReplicationPolicy.NONE)) {
      throw new VeniceHttpException(
          HttpStatus.SC_BAD_REQUEST,
          "requesting topic for streaming writes to store " + store.getName()
              + " which is configured to have a hybrid data replication policy "
              + store.getHybridStoreConfig().getDataReplicationPolicy(),
          ErrorType.BAD_REQUEST);
    }
```


I wonder if we should just replace it with `    if (pushType.equals(PushType.STREAM)
        && (store.getHybridStoreConfig().getDataReplicationPolicy().equals(DataReplicationPolicy.NONE) && !store.isAaEnabled())) {`

internal/venice-common/src/main/java/com/linkedin/venice/meta/DataReplicationPolicy.java
sushantmane: I think removing this might result in errors during dser of znodes?

nisargthakkar: Let's leave it in. If we remove it, we might end up reusing the enum number too

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/DataRecoveryTest.java
nisargthakkar:
```
  /**
   * This enum value is used in 2 cases:
   *   1. batch-only stores since it has no real-time data replication.
   *   2. incremental push stores that do not have Samza jobs.
   */
  NONE(2),
```

It doesn't make sense for batch-only stores as they won't have a hybrid config (where DRP lives). So, it's only the 2nd one.

Let's not increase the scope of DRP NONE. I think we should be able to remove that too. But it is most-likely out of scope of this change

internal/venice-common/src/main/java/com/linkedin/venice/meta/DataReplicationPolicy.java
sushantmane: Yeah agree

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/DataRecoveryTest.java
sushantmane: >Let's not increase the scope of DRP NONE. I think we should be able to remove that too. But it is most-likely out of scope of this change

Do you recommend using NON_AGG over NONE whenever possible?

nisargthakkar: Reading through the PR again, it looks okay to change DRP to NONE, and have the handling you suggested originally. I seem to prefer that over `NON_AGGREGATE` for AA stores 😅. Can you please update the javadoc of DRP NONE also?

internal/venice-common/src/main/java/com/linkedin/venice/meta/DataReplicationPolicy.java
nisargthakkar, line 34: Comment here just for reference.

I couldn't find the code change to the lag measurement for AA stores and `NON_AGGREGATE` DRP. Is that expected?

sushantmane: We removed use of DRPs for the lag measurement of AA enabled stores

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/DataRecoveryTest.java
sushantmane: sure

internal/venice-common/src/main/java/com/linkedin/venice/meta/DataReplicationPolicy.java
```diff
@@ -7,6 +7,7 @@
 
 /**
  * Enums of the policies used to decide how real-time samza data is replicated.
+ * Note: Data replication policy does not play any role for AA enabled hybrid stores.
```
adamxchen, line 10: nit: The DRP often cause some confusions on me so can you please add more notes here? So, the default DRP for AA store would be `NON_AGGREGATE`? And do we have validation if AA store is updated with `AGGREGATE` as DRP?
BTW, can we also update some internal runbooks to reflect this?

sushantmane: > The DRP often cause some confusions on me so can you please add more notes here? So, the default DRP for AA store would be NON_AGGREGATE? 
yes. But it will not have any effect

>And do we have validation if AA store is updated with AGGREGATE as DRP?
If store is AA enabled DRP is of no use. I don't follow what you are saying here. Could you add more details? Thanks


>BTW, can we also update some internal runbooks to reflect this?
sure

sushantmane: Done

adamxchen: Understood. I thought DPR is required to be  `NON_AGGREGATE` when AA is on so we need to do some validation in case it's set to `AGGREGATE`. Since it doesn't play a role at all, then it's not needed

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/DataRecoveryTest.java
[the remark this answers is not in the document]
sushantmane: Good point. 

> Is it just for reasons that it is deprecated? 


Yes

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/ingestionHeartbeat/IngestionHeartBeatTest.java
```diff
@@ -121,14 +120,11 @@ public void cleanupStore() {
...
-    return DataProviderUtils.allPermutationGenerator((permutation) -> {
-      boolean isActiveActiveEnabled = (boolean) permutation[1];
-      DataReplicationPolicy dataReplicationPolicy = (DataReplicationPolicy) permutation[3];
-      if (isActiveActiveEnabled && dataReplicationPolicy != ACTIVE_ACTIVE) {
-        return false;
-      }
-      return true;
-    }, DataProviderUtils.BOOLEAN, DataProviderUtils.BOOLEAN, DataReplicationPolicy.values());
+    return DataProviderUtils.allPermutationGenerator(
+        (permutation) -> true,
```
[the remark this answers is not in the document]
sushantmane: But here we don't need array. Just one val?

nisargthakkar (approved):
LGTM. Thanks a lot

sushantmane:
Thanks a lot, @nisargthakkar, @huangminchn, and @adamxchen!

PR #1046, opened 2024-06-23
Query blob discovery bootstrap

ishwarya-citro:
## Query the blob discovery during bootstrapping

During the bootstrapping process of DVC, it now calls the blob discovery to attempt to bootstrap after downloading the blobs from a peer node. Then, it continues consumption from Kafka.

Resolves [VENG-11145](https://jira01.corp.linkedin.com:8443/browse/VENG-11145)

## How was this PR tested?
Unit Tests


internal/venice-common/src/main/java/com/linkedin/venice/blobtransfer/DvcBlobFinder.java
```diff
@@ -0,0 +1,74 @@
...
+ */
+public class DvcBlobFinder implements BlobFinder {
+  private static final Logger LOGGER = LogManager.getLogger(DvcBlobFinder.class);
+  private static final String TYPE_BLOB_DISCOVERY = "TYPE_BLOB_DISCOVERY";
+  private static final String ERROR_DISCOVERY_MESSAGE =
+      "Error finding DVC peers for blob transfer in store: %s, version: %d, partition: %d";
+  private final TransportClient transportClient;
+  private final String routerUrl;
+
+  public DvcBlobFinder(TransportClient transportClient, String routerUrl) {
```
adamxchen, line 29: Do we need `routerUrl`? When `TransportClient` is constructed, it's supposed to have the url already. See `HttpTransportClient` as one example

internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
```diff
@@ -1739,6 +1739,14 @@ private ConfigKeys() {
   // Config to control how much percentage of DVC replica instances are allowed to be offline before failing VPJ push.
   public static final String DAVINCI_PUSH_STATUS_SCAN_MAX_OFFLINE_INSTANCE_RATIO =
       "davinci.push.status.scan.max.offline.instance.ratio";
+  // Denotes whether DVC is allowed to bootstrap via a peer-to-peer blob transfer
+  public static final String DAVINCI_P2P_BLOB_TRANSFER_ENABLED = "davinci.p2p.blob.transfer.enabled";
```
adamxchen, line 1743: Now I am thinking we don't need this config. We've added `blobTransferEnabled` to the version config, and it has implied that this config would apply to the store-version only. 
For server p2p, it should use a cluster-wide config since it doesn't much sense to me that we only enable server p2p to a few stores. @sebas-inf  @sixpluszero

internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
```diff
@@ -1739,6 +1739,14 @@ private ConfigKeys() {
...
+  // Port used by peer-to-peer transfer service
+  public static final String DAVINCI_P2P_BLOB_TRANSFER_PORT = "davinci.p2p.blob.transfer.port";
+  // Port used to connect to remote host for file transfer
+  public static final String DAVINCI_P2P_FILE_TRANSFER_PORT = "davinci.p2p.file.transfer.port";
+
+  public static final String DAVINCI_P2P_BLOB_TRANSFER_BASE_DIR = "davinci.p2p.blob.transfer.base.dir";
```
adamxchen, line 1749: Let's remove it

internal/venice-common/src/main/java/com/linkedin/venice/blobtransfer/DvcBlobFinder.java
```diff
@@ -0,0 +1,74 @@
...
+        return handleError(ERROR_DISCOVERY_MESSAGE, storeName, version, partition, e);
+      }
+    }).exceptionally(throwable -> handleError(ERROR_DISCOVERY_MESSAGE, storeName, version, partition, throwable));
+
+    return futureResponse.join();
+  }
+
+  private String buildUriForBlobDiscovery(String storeName, int version, int partition) {
+    List<NameValuePair> queryParams = new ArrayList<>();
+    queryParams.add(new BasicNameValuePair("store_name", storeName));
```
adamxchen, line 53: Since we have defined these parameter in the `ControllerApiConstants`, let's use them

clients/da-vinci-client/src/main/java/com/linkedin/davinci/VersionBackend.java
```diff
@@ -325,8 +342,11 @@ synchronized CompletableFuture<Void> subscribe(ComplementSet<Integer> partitions
         partitionFutures.computeIfAbsent(partition, k -> CompletableFuture.completedFuture(null));
       } else {
         partitionFutures.computeIfAbsent(partition, k -> new CompletableFuture<>());
-        // AtomicReference of storage engine will be updated internally.
-        backend.getIngestionBackend().startConsumption(config, partition);
+        try {
+          dbBootstrapper.bootstrapDatabase(storeName, versionNumber, partition);
+        } catch (Exception e) {
```
adamxchen, line 347: What do you expect to do when exception is caught?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/bootstrap/DvcP2PRocksDbBootstrapper.java
```diff
@@ -0,0 +1,115 @@
...
+      Boolean isHybridStore) {
+    int p2pBlobTransferPort = serverConfig.getDvcP2pBlobTransferPort();
+    int p2pFileTransferPort = serverConfig.getDvcP2pFileTransferPort();
+    String routerUrl = backend.getClientConfig().getVeniceURL();
+
+    this.p2pBlobTransferBaseDir = serverConfig.getDvcP2pBlobTransferBaseDir();
+    this.isDvcP2pBlobTransferEnabled = serverConfig.isDvcP2pBlobTransferEnabled();
+    this.veniceStoreVersionConfig = veniceStoreVersionConfig;
+    this.backend = backend;
+    this.isBlobTransferEnabled = isBlobTransferEnabled;
```
adamxchen, line 45: line 45 and line 42 can create some confusion. Plz see my comment below

clients/da-vinci-client/src/main/java/com/linkedin/davinci/bootstrap/AbstractBootstrapper.java
```diff
@@ -0,0 +1,40 @@
...
+
+
+public abstract class AbstractBootstrapper implements RocksDbBootstrapper {
+  static {
+    RocksDB.loadLibrary();
+  }
+
+  private static final Logger LOGGER = LogManager.getLogger(AbstractBootstrapper.class);
+
+  protected void verifyBootstrap(String newDbDir) throws RocksDBException {
```
adamxchen, line 18: This is method is a bit confusing. Why do we need to open the rocksdb to verify? If we are able to bootstrap from it, it means a SUCCESS.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/bootstrap/DvcP2PRocksDbBootstrapper.java
```diff
@@ -0,0 +1,115 @@
...
+    } else {
+      bootstrapFromBlobs(storeName, versionNumber, partitionId);
+    }
+  }
+
+  @Override
+  protected void bootstrapFromBlobs(String storeName, int versionNumber, int partitionId)
+      throws VeniceBootstrapException {
+    try {
+      p2pBlobTransferManager.start();
```
adamxchen, line 82: We don't expect to start this manager multiple times right?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/bootstrap/DvcP2PRocksDbBootstrapper.java
```diff
@@ -0,0 +1,115 @@
...
+   * If it fails to bootstrap from blobs at anypoint, it falls back to consuming data from Kafka
+   *
+   * @param storeName - store Name
+   * @param versionNumber - store version
+   * @param partitionId - partition ID for the store & version
+   * @throws VeniceBootstrapException
+   */
+  @Override
+  public void bootstrapDatabase(String storeName, int versionNumber, int partitionId) throws VeniceBootstrapException {
+    if (!isDvcP2pBlobTransferEnabled || !isBlobTransferEnabled || isHybridStore) {
```
adamxchen, line 71: See my other comment plz.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/bootstrap/DvcP2PRocksDbBootstrapper.java
```diff
@@ -0,0 +1,115 @@
...
+          .whenComplete((inputStream, throwable) -> handleP2PTransferCompletion(inputStream, throwable, partitionId));
+    } catch (Exception e) {
+      LOGGER.error("Error occurred during bootstrap from blobs: {}", e.getMessage());
+      bootstrapFromKafka(partitionId);
+    }
+  }
+
+  private void handleP2PTransferCompletion(InputStream inputStream, Throwable throwable, int partitionId) {
+    if (throwable != null) {
+      LOGGER.error("Error occurred during p2p transfer: {}", throwable.getMessage());
```
adamxchen, line 95: We also need to clean up the corrupted state left by the p2p transfer

clients/da-vinci-client/src/main/java/com/linkedin/davinci/bootstrap/DvcP2PRocksDbBootstrapper.java
```diff
@@ -0,0 +1,115 @@
...
+      bootstrapFromKafka(partitionId);
+      return;
+    }
+
+    try {
+      // P2P transfer is successful, attempt to bootstrap from blobs
+      verifyBootstrap(p2pBlobTransferBaseDir);
+      p2pBlobTransferManager.close();
```
adamxchen, line 103: What if we are handling two requests? Do we close the manager when one is done and another one is still in progress?

adamxchen (changes requested):

internal/venice-common/src/main/java/com/linkedin/venice/blobtransfer/DvcBlobFinder.java
ishwarya-citro: Yes, i saw that but need to pass the `routerUrl` separately because `routerUrl` is protected in `HttpTransportClient`.

internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
ishwarya-citro: would we want a "davinci_ambry_blob_transfer_enabled" as a feature flag in future? if so, maybe we should consider keeping this.

internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
ishwarya-citro: would we want a "davinci_ambry_blob_transfer_enabled" as a feature flag in future? if so, maybe we should consider keeping this.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/VersionBackend.java
ishwarya-citro: `bootstrapDatabase` attempts to bootstrap from blobs and if there's an exception at that point, it falls back to kafka. It looks like StoreIngestions handles other exceptions so I am just logging here. Is there anything else we should be doing here?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/bootstrap/AbstractBootstrapper.java
ishwarya-citro: To verify that the DB can read the entities. Could there be any issues with the data that prevents rocksdb to read/open it? If not / if we don't need this, i can remove

internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
ishwarya-citro: i think i'll remove it.

internal/venice-common/src/main/java/com/linkedin/venice/blobtransfer/DvcBlobFinder.java
adamxchen: No. My point is since `routerUrl` is already required in some `TransportClient`. Then it's not needed here when sending a request.
https://github.com/linkedin/venice/blob/0ddf4f2e116cd8b14c553b6056abc718c73725a0/clients/venice-thin-client/src/main/java/com/linkedin/venice/client/store/transport/HttpTransportClient.java#L149

clients/da-vinci-client/src/main/java/com/linkedin/davinci/bootstrap/DvcP2PRocksDbBootstrapper.java
```diff
@@ -0,0 +1,115 @@
...
+  private void handleP2PTransferCompletion(InputStream inputStream, Throwable throwable, int partitionId) {
+    if (throwable != null) {
+      LOGGER.error("Error occurred during p2p transfer: {}", throwable.getMessage());
+      bootstrapFromKafka(partitionId);
+      return;
+    }
+
+    try {
+      // P2P transfer is successful, attempt to bootstrap from blobs
+      verifyBootstrap(p2pBlobTransferBaseDir);
```
adamxchen, line 109: When `verifyBootstrap` has some issues, we just do some loggings so the transfer is always mistakenly treated as "success". Is this true?

internal/venice-common/src/main/java/com/linkedin/venice/blobtransfer/DvcBlobFinder.java
ishwarya-citro: The routerUrl is setting from here 
https://github.com/linkedin/venice/blob/main/clients/venice-thin-client/src/main/java/com/linkedin/venice/client/store/ClientFactory.java#L159

and we need to build params -> 
https://github.com/linkedin/venice/pull/1046/files#diff-f7f48dd230ec7f6940dc3dbcd1e2f4334a3a3211fe62f277ee50bc39ad1ebc67R55

maybe we should sync on this as I am not clear on the above explanation

internal/venice-common/src/main/java/com/linkedin/venice/blobtransfer/BlobTransferManager.java
```diff
@@ -42,6 +43,16 @@ CompletionStage<? extends InputStream> get(String storeName, int version, int pa
...
 
+  /**
+   * Deletes the files associated with the specified store, version, and partition when a blob transfer fails.
+   *
+   * @param storeName the name of the store
+   * @param version the version number of the store
+   * @param partition the partition ID
+   * @return a CompletableFuture that completes when the deletion process is done
+   */
+  CompletableFuture<T> delete(String storeName, int version, int partition);
```
adamxchen, line 54: As we discussed offline, we don't need this method as this method means to delete blobs files on the remote (either blobs on peers or on a blob store), not for cleaning up blobs on local. We can have a util method somewhere and call that method

clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/isolated/IsolatedIngestionServer.java
```diff
@@ -719,7 +726,19 @@ private void initializeIsolatedIngestionServer() {
...
+    int fileTransferPort = serverConfig.getDvcP2pFileTransferPort();
+    String rocksDBPath = serverConfig.getRocksDBPath();
+
+    blobTransferManager = new NettyP2PBlobTransferManager(
+        new P2PBlobTransferService(blobTransferPort, rocksDBPath),
+        new NettyFileTransferClient(fileTransferPort, rocksDBPath),
+        new DvcBlobFinder(ClientFactory.getTransportClient(clientConfig)));
+    blobTransferManager.start();
+
+    ingestionBackend =
```
adamxchen, line 740: Enabling II doesn't mean we should enabled P2P functionality. Our current setup is if the P2P is enabled in the version level config, then after the batch job, the host(server or DVC) creates the snapshot. So if we want to bring this to II, we need to check the version flag. Given we are sunsetting II in some cases, we don't have to add these codes to II for simplicity.

plus

you have created the p2p manager in `IsolatedingestionBackend` so I think we don't need to create here on more time

clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
```diff
@@ -272,6 +277,15 @@ public DaVinciBackend(
       ingestionService.start();
       ingestionService.addIngestionNotifier(ingestionListener);
 
+      int blobTransferPort = backendConfig.getDvcP2pBlobTransferPort();
```
adamxchen, line 280: Can we also check the version-level flag to confirm P2P is enabled before creating the blobTransferManager?

adamxchen (commented):
Haven't finished my review yet but have some initial comments

internal/venice-test-common/src/main/java/com/linkedin/venice/utils/TestWriteUtils.java
```diff
@@ -653,6 +655,8 @@ public static Properties defaultVPJPropsWithD2Routing(
 
     props.put(VENICE_DISCOVER_URL_PROP, String.format("d2://%s", controllerServiceName));
     props.put(SOURCE_GRID_FABRIC, childRegionNamesToZkAddress.entrySet().iterator().next().getKey());
+    props.put(DAVINCI_P2P_BLOB_TRANSFER_PORT, TestUtils.getFreePort());
```
adamxchen, line 658: We don't need these configs here as they are not useful for VPJ

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/fastclient/FastClientDaVinciClientCompatTest.java
```diff
@@ -116,6 +119,8 @@ private DaVinciClient<String, TestValueSchema> setupDaVinciClient(String storeNa
             .put(CLIENT_USE_SYSTEM_STORE_REPOSITORY, true)
             .put(CLIENT_USE_DA_VINCI_BASED_SYSTEM_STORE_REPOSITORY, false)
             .put(DATA_BASE_PATH, dataPath)
+            .put(DAVINCI_P2P_BLOB_TRANSFER_PORT, TestUtils.getFreePort())
```
adamxchen, line 122: I see we are adding these configs to many test files. Do we really use them in these files? Can we only add them if needed?

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/DaVinciClientMemoryLimitTest.java
```diff
@@ -75,7 +77,7 @@
 
 
 public class DaVinciClientMemoryLimitTest {
-  private static final int TEST_TIMEOUT = 180_000;
+  private static final int TEST_TIMEOUT = 240_000;
```
adamxchen, line 80: let's also find out why test was hanging before increasing the time out

internal/venice-common/src/main/java/com/linkedin/venice/blobtransfer/server/P2PBlobTransferService.java
```diff
@@ -62,6 +70,23 @@ public boolean startInner() throws Exception {
     return true;
   }
 
+  public CompletableFuture<Void> deleteFiles(String storeName, int version, int partition) {
```
adamxchen, line 73: We can have a util method in `RocksDBUtils` to clean up the corrupted partition

clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/DefaultIngestionBackend.java
```diff
@@ -50,13 +65,79 @@ public void startConsumption(VeniceStoreVersionConfig storeConfig, int partition
...
+  private CompletionStage<Void> bootstrapFromBlobs(Store store, int versionNumber, int partitionId) {
+    if (!store.isBlobTransferEnabled() || store.isHybrid()) {
+      return CompletableFuture.completedFuture(null);
+    }
+
+    String storeName = store.getName();
+    CompletionStage<InputStream> p2pTransfer = blobTransferManager.get(storeName, versionNumber, partitionId);
+    CompletableFuture<Void> result = new CompletableFuture<>();
+
+    blobTransferScheduler.schedule(() -> {
```
adamxchen, line 99: you don't need a scheduler service. `CompletionStage.toCompletableFuture ` -> `get(long timeout, TimeUnit)` ([doc](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html#get-long-java.util.concurrent.TimeUnit-))

clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/DefaultIngestionBackend.java
```diff
@@ -50,13 +65,79 @@ public void startConsumption(VeniceStoreVersionConfig storeConfig, int partition
...
+   * any exceptions), it deletes the partially downloaded blobs, and eventually falls back to bootstrapping from Kafka.
+   * Blob transfer should be enabled to boostrap from blobs, and it currently only supports batch-stores.
+   */
+  private CompletionStage<Void> bootstrapFromBlobs(Store store, int versionNumber, int partitionId) {
+    if (!store.isBlobTransferEnabled() || store.isHybrid()) {
+      return CompletableFuture.completedFuture(null);
+    }
+
+    String storeName = store.getName();
+    CompletionStage<InputStream> p2pTransfer = blobTransferManager.get(storeName, versionNumber, partitionId);
```
adamxchen, line 96: Can we add log here to tell we are start doing blob transfer?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/DefaultIngestionBackend.java
```diff
@@ -50,13 +65,79 @@ public void startConsumption(VeniceStoreVersionConfig storeConfig, int partition
...
+    CompletableFuture<Void> result = new CompletableFuture<>();
+
+    blobTransferScheduler.schedule(() -> {
+      if (!result.isDone()) {
```
adamxchen, line 100: You are using this `result` incorrectly and we don't need it at all. This `result` has nothing to do with our p2p future. We should chain this with `p2pTransfer`

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/fastclient/FastClientDaVinciClientCompatTest.java
ishwarya-citro: Yes. The tests were breaking w/o them

clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/DefaultIngestionBackend.java
```diff
@@ -50,13 +66,61 @@ public void startConsumption(VeniceStoreVersionConfig storeConfig, int partition
...
+        p2pFuture.get(30, TimeUnit.MINUTES);
+      } catch (Exception e) {
+        LOGGER.error(
+            "Failed bootstrapping from blobs for store {}, version {}, partition {}",
+            storeName,
+            versionNumber,
+            partitionId,
+            e);
+        RocksDBUtils.deleteBlobFiles(baseDir, storeName, versionNumber, partitionId)
+            .thenAccept(deleteResult -> p2pFuture.completeExceptionally(e))
```
adamxchen, line 115: you can just call `p2pFuture.cancel(true)` to complete it with a `CancellationException` (though it has no effect on the underlying implementation). We also have to close the channel to prevent future operation but we can leave this out of this PR.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/DefaultIngestionBackend.java
```diff
@@ -50,13 +66,61 @@ public void startConsumption(VeniceStoreVersionConfig storeConfig, int partition
...
+    CompletionStage<InputStream> p2pTransfer = blobTransferManager.get(storeName, versionNumber, partitionId);
+    CompletableFuture<InputStream> p2pFuture = p2pTransfer.toCompletableFuture();
+
     LOGGER
-        .info("Completed starting consumption in ingestion service for store {} partition {}", storeVersion, partition);
+        .info("Bootstrapping from blobs for store {}, version {}, partition {}", storeName, versionNumber, partitionId);
+
+    CompletableFuture.runAsync(() -> {
+      try {
+        p2pFuture.get(30, TimeUnit.MINUTES);
```
adamxchen, line 106: I think your implementation is on the right direction but it's a bit overcomplicated and not easy to follow.

I think we can do the followings:
```
// do something
CompletableFuture<InputStream> p2pFuture = p2pTransfer.toCompletableFuture();
// do something
return CompletableFuture.runAsync(() -> {
    try {
            p2pFuture.get(30, TimeUnit.MINUTES);
    } catch (Exception e) {
       // just delete. 
       RocksDBUtils.deleteBlobFiles(...)
      p2pFuture.cancel(true);
      // TODO: we also need to close the channel but let's do it later
    }
}

```

clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/DefaultIngestionBackend.java
```diff
@@ -50,13 +66,61 @@ public void startConsumption(VeniceStoreVersionConfig storeConfig, int partition
...
+      } catch (Exception e) {
+        LOGGER.error(
```
adamxchen, line 108: can we make it a warning since we can still fall back to regular consumption?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/DefaultIngestionBackend.java
ishwarya-citro: Thanks!

internal/venice-common/src/main/java/com/linkedin/venice/blobtransfer/server/P2PBlobTransferService.java
```diff
@@ -32,6 +33,7 @@ public class P2PBlobTransferService extends AbstractVeniceService {
 
   public P2PBlobTransferService(int port, String baseDir) {
     this.port = port;
+    this.baseDir = baseDir;
```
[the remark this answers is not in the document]
ishwarya-citro: No. I forgot to remove it. Thanks

internal/venice-common/src/main/java/com/linkedin/venice/blobtransfer/BlobFinder.java
```diff
@@ -1,6 +1,9 @@
 package com.linkedin.venice.blobtransfer;
 
-public interface BlobFinder {
+import java.io.Closeable;
+
+
+public interface BlobFinder extends Closeable {
```
adamxchen, line 6: nit: it's better to have AutoClosable

internal/venice-common/src/main/java/com/linkedin/venice/store/rocksdb/RocksDBUtils.java
```diff
@@ -95,4 +99,24 @@ public static int extractTempRMDSSTFileNo(String fileName) {
...
+   * @param storeName the name of the store
+   * @param version the version number of the store
+   * @param partition the partition ID
+   */
+  public static void deletePartitionDir(String baseDir, String storeName, int version, int partition) {
+    Path path = null;
+    try {
+      path = Paths.get(baseDir, storeName, String.valueOf(version), String.valueOf(partition));
+      if (Files.exists(path)) {
+        Files.walk(path).sorted(Comparator.reverseOrder()).map(Path::toFile).forEach(File::delete);
```
adamxchen, line 115: Sry. One thing I missed to review last time, in the p2p client, it would create the directory if not exist. Does this method also clean up the directory or just files? We need to delete the partition directory as if nothing happens

internal/venice-common/src/test/java/com/linkedin/venice/store/rocksdb/TestRocksDBUtils.java
```diff
@@ -0,0 +1,112 @@
...
+
+
+public class TestRocksDBUtils {
+  @Test
+  public void testDeletePartitionDir() throws IOException {
+    Path baseDir = null;
+
+    try {
+      // tmp directory "rocksdb/storeName/1/1/{file1.txt|file2.txt}"
+      baseDir = Files.createTempDirectory("rocksdb");
```
adamxchen, line 22: You don't need these duplicated codes if you can use   @BeforeMethod and   @AfterMethod to do setup and clean up accordingly.

internal/venice-common/src/main/java/com/linkedin/venice/blobtransfer/BlobFinder.java
ishwarya-citro: thanks - why is AutoClosable better?

internal/venice-common/src/main/java/com/linkedin/venice/store/rocksdb/RocksDBUtils.java
ishwarya-citro: if the directory is `storeName/5/6/file2.txt` , it would delete `5/6/file2.txt`. Does this answer your question?

internal/venice-common/src/main/java/com/linkedin/venice/blobtransfer/BlobFinder.java
adamxchen: https://stackoverflow.com/questions/13141302/implements-closeable-or-implements-autocloseable

adamxchen (approved):
Thanks a lot for the hard work!

internal/venice-common/src/test/java/com/linkedin/venice/store/rocksdb/TestRocksDBUtils.java
```diff
@@ -0,0 +1,85 @@
...
+    // assert version 5, partition 6 should not exist
+    assertFalse(Files.exists(baseDir.resolve("storeName_v5/storeName_v5_6")));
+  }
+
+  @Test
+  public void testDeletePartitionDir_EmptyFiles() throws IOException {
+    Files.createDirectories(baseDir.resolve("storeName_v2/storeName_v2_2"));
+    // assert the temp base directory does exist
+    assertTrue(Files.exists(baseDir));
+    deletePartitionDir(baseDir.toString(), "storeName", 2, 2);
```
adamxchen, line 83: do we need to verify anything here?

ishwarya-citro: we verify it here: https://github.com/linkedin/venice/pull/1046/files#diff-7252ecd0256005cddbe041ee69e321f2bcfb955f7f8a6bd71d4e3e72b82a8368R31

adamxchen (approved):

internal/venice-common/src/main/java/com/linkedin/venice/blobtransfer/DvcBlobFinder.java
```diff
@@ -0,0 +1,81 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+
+/**
+ * DvcBlobFinder discovers live DaVinci peer nodes to facilitate blob transfers necessary for bootstrapping the database
+ */
+public class DvcBlobFinder implements BlobFinder {
+  private static final Logger LOGGER = LogManager.getLogger(DvcBlobFinder.class);
+  private static final String TYPE_BLOB_DISCOVERY = "TYPE_BLOB_DISCOVERY";
```
adamxchen, line 27: this looks not right.

PR #1049, opened 2024-06-25
[controller] Delete backup version before start of new push in target regions

majisourav99:
## Delete backup version before starting of new push when target region push is enabled.

When target region push is enabled, after pushing to the target region, rest of the regions do not delete the backup version before the start of the push. This increases the disk usage during the duration of the push to 3x of the store size (backup + current + future versions).  This PR deletes the backup version before starting ingestion in non-target regions.

## How was this PR tested?

integ test.


huangminchn (approved):
Thanks a lot Sourav!

PR #1050, opened 2024-06-26
[all] Upgrade helix lib dependency to 1.4.1

majisourav99:
## Upgrade helix lib dependency to 1.4.0

Also upgrades zookeeper to 3.6.3. The exclusions are needed to provide JDK8 compatible APIs for various groups.

## How was this PR tested?

GHCI

### DIFF
--- a/clients/da-vinci-client/build.gradle
+++ b/clients/da-vinci-client/build.gradle
@@ -17,19 +17,19 @@ dependencies {
     exclude group: 'xerces', module: 'xerces' // resolve conflicting transitive dependency
     exclude group: 'com.fasterxml.jackson.core', module: 'jackson-core'// resolve conflicting transitive dependency
   }
-  implementation('org.apache.helix:helix-core:1.4.0:jdk8') {
+  implementation('org.apache.helix:helix-core:1.4.1:jdk8') {
     exclude group: 'org.apache.helix'
   }
-  implementation('org.apache.helix:helix-common:1.4.0:jdk8')  {
+  implementation('org.apache.helix:helix-common:1.4.1:jdk8')  {
     exclude group: 'org.apache.helix'
   }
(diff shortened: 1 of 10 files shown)


adamxchen:
IIRC, did we say `1.4.0` is not good so we gotta wait for another release?

majisourav99:
> IIRC, did we say `1.4.0` is not good so we gotta wait for another release?

Yes, 1.4.0 is not good, waiting for another release. Let me close this for now

clients/da-vinci-client/build.gradle
```diff
@@ -17,13 +17,24 @@ dependencies {
     exclude group: 'xerces', module: 'xerces' // resolve conflicting transitive dependency
     exclude group: 'com.fasterxml.jackson.core', module: 'jackson-core'// resolve conflicting transitive dependency
   }
-
+  implementation('org.apache.helix:helix-core:1.4.1:jdk8') {
```
arjun4084346, line 20: Please help me understand this statement.
we want `org.apache.helix:helix-core` but we also do not want any of `org.apache.helix` ?

majisourav99: Right, previously we used only helix-core as well and only the used library are ported to be JDK8 compatible.

sixpluszero (approved):
implementation('org.apache.helix:helix-common:1.4.1:jdk8')  {
    exclude group: 'org.apache.helix'
  }
I read your comment so looks like this is intended change.
Given it is a big version bump, we need to make sure cert/EI is running fine before moving into production.

majisourav99:
> implementation('org.apache.helix:helix-common:1.4.1:jdk8') { exclude group: 'org.apache.helix' } I read your comment so looks like this is intended change. Given it is a big version bump, we need to make sure cert/EI is running fine before moving into production.

Yes, it is indeed very big bump and needs proper validation before rolling out.

PR #1052, opened 2024-06-27
[vpj] [controller] Enforce Record Size Limit on Batch Push Jobs

KaiSernLim:
## Summary

**NOTE:** This PR adds the code for enforcing record size limits, but no push jobs will fail by default until the controller config `controller.default.max.record.size.bytes` is adjusted and deployed for each cluster.

1. When users try to push Venice records larger than `maxRecordSizeBytes` (introduced in #1027), the Batch Push Job will fail, reporting a `RecordSizeTooLarge` error. This is the same error that is emitted when `chunkingEnabled=false` and a user tries to push a record surpassing the `1MB` Kafka record size limit (handled by `maxSizeForUserPayloadPerMessageInBytes`).
2. The default is unlimited / unset (`-1`), but the store-level config `maxRecordSizeBytes` can be adjusted on the store itself.
3. The intended way of changing the setting fleetwide is by modifying the controller config `controller.default.max.record.size.bytes` per cluster. The controller will backfill the `maxRecordSizeBytes` if the store-level config is absent (`-1`) ([link](https://github.com/linkedin/venice/pull/1052#discussion_r1669466912)).
    1. The default value for this controller backfill will be `100MB`, which is an arbitrary high number that will not be reached.

### Changes

1. **[Functional]** `VeniceWriter` will now throw `RecordTooLargeException` if `largeRecordsAllowed=false` and it tries to `put()` a record which has a size that is larger than the store-level setting `maxRecordSizeBytes`. Currently, only batch push jobs should set the value to an actual value.
2. **[Functional]** `VenicePushJob` added support for the `maxRecordSizeBytes` and `largeRecordsAllowed` settings and will pass these settings down to the `DataWriterJob` who is the one that actually touches the data.
3. **[Functional]** Venice Controller will backfill the value of `maxRecordSizeBytes` with a controller config `defaultMaxRecordSizeBytes` (set by `controller.default.max.record.size.bytes`) when the `getStore()` API is called. The intended use case is for us to be able to set it on the controller-side to control the value fleetwide. The default is set to `100MB` by default which should never be hit.
4. **[Unit Test]** Added unit tests for test coverage:
    1. `testGetVeniceWriter()` and `testUpdatePushJobDetailsWithJobDetailsRecordTooLarge()` in `VenicePushJobTest`
    2. `testPutTooLargeRecord()` in `VeniceWriterUnitTest`
    3. `testGetStore()` in `StoreRoutesTest`
5. **[Integration Test]**  Added integration tests which should hopefully mean that the changes are working:
    1. `testStoreWithTooLargeValues()` in `TestBatch`
    2. `testPushJobDetailsRecordTooLarge()` in `PushJobDetailsTest`

## How was this PR tested?
Added integration tests and unit tests.

See the **Summary** section.

### DIFF
--- a/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/PushJobSetting.java
+++ b/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/PushJobSetting.java
@@ -114,7 +114,6 @@ public class PushJobSetting implements Serializable {
   public Map<String, String> partitionerParams;
   public boolean chunkingEnabled;
   public boolean rmdChunkingEnabled;
-  public boolean largeRecordsAllowed;
   public int maxRecordSizeBytes;
   public String kafkaSourceRegion;
   public transient RepushInfoResponse repushInfoResponse;
--- a/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/PushJobSetting.java
+++ b/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/PushJobSetting.java
@@ -114,6 +114,8 @@ public class PushJobSetting implements Serializable {
   public Map<String, String> partitionerParams;
   public boolean chunkingEnabled;
   public boolean rmdChunkingEnabled;
+  public boolean largeRecordsAllowed;
+  public int maxRecordSizeBytes;
   public String kafkaSourceRegion;
   public transient RepushInfoResponse repushInfoResponse;
 
--- a/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
+++ b/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
@@ -5,7 +5,6 @@ import static com.linkedin.venice.ConfigKeys.KAFKA_BOOTSTRAP_SERVERS;
 import static com.linkedin.venice.ConfigKeys.KAFKA_PRODUCER_DELIVERY_TIMEOUT_MS;
 import static com.linkedin.venice.ConfigKeys.KAFKA_PRODUCER_REQUEST_TIMEOUT_MS;
 import static com.linkedin.venice.ConfigKeys.KAFKA_PRODUCER_RETRIES_CONFIG;
-import static com.linkedin.venice.ConfigKeys.LARGE_RECORDS_ALLOWED;
 import static com.linkedin.venice.ConfigKeys.VENICE_PARTITIONERS;
 import static com.linkedin.venice.VeniceConstants.DEFAULT_SSL_FACTORY_CLASS_NAME;
 import static com.linkedin.venice.hadoop.VenicePushJobConstants.ALLOW_DUPLICATE_KEY;
@@ -379,7 +378,6 @@ public class VenicePushJob implements AutoCloseable {
     pushJobSettingToReturn.batchNumBytes = props.getInt(BATCH_NUM_BYTES_PROP, DEFAULT_BATCH_BYTES_SIZE);
     pushJobSettingToReturn.isIncrementalPush = props.getBoolean(INCREMENTAL_PUSH, false);
     pushJobSettingToReturn.isDuplicateKeyAllowed = props.getBoolean(ALLOW_DUPLICATE_KEY, false);
-    pushJobSettingToReturn.largeRecordsAllowed = props.getBoolean(LARGE_RECORDS_ALLOWED, true);
     pushJobSettingToReturn.enablePushJobStatusUpload = props.getBoolean(PUSH_JOB_STATUS_UPLOAD_ENABLE, false);
     pushJobSettingToReturn.controllerRetries = props.getInt(CONTROLLER_REQUEST_RETRY_ATTEMPTS, 1);
     pushJobSettingToReturn.controllerStatusPollRetries = props.getInt(POLL_STATUS_RETRY_ATTEMPTS, 15);
@@ -515,10 +513,6 @@ public class VenicePushJob implements AutoCloseable {
           props.getBoolean(USE_MAPPER_TO_BUILD_DICTIONARY, DEFAULT_USE_MAPPER_TO_BUILD_DICTIONARY);
       pushJobSettingToReturn.compressionMetricCollectionEnabled =
           props.getBoolean(COMPRESSION_METRIC_COLLECTION_ENABLED, DEFAULT_COMPRESSION_METRIC_COLLECTION_ENABLED);
-    }
-
-    if (pushJobSettingToReturn.isSourceKafka || pushJobSettingToReturn.isSourceETL) {
-      pushJobSettingToReturn.largeRecordsAllowed = true; // safer to simply allow large records on repush
     }
 
     // Compute-engine abstraction related configs
@@ -2347,14 +2341,10 @@ public class VenicePushJob implements AutoCloseable {
     Properties veniceWriterProperties = new Properties();
     veniceWriterProperties.put(KAFKA_BOOTSTRAP_SERVERS, pushJobSetting.kafkaUrl);
     veniceWriterProperties.put(VeniceWriter.MAX_ELAPSED_TIME_FOR_SEGMENT_IN_MS, -1);
+    veniceWriterProperties.put(VeniceWriter.MAX_RECORD_SIZE_BYTES, pushJobSetting.maxRecordSizeBytes);
     if (props.containsKey(VeniceWriter.CLOSE_TIMEOUT_MS)) { /* Writer uses default if not specified */
       veniceWriterProperties.put(VeniceWriter.CLOSE_TIMEOUT_MS, props.getInt(VeniceWriter.CLOSE_TIMEOUT_MS));
     }
-    if (props.containsKey(VeniceWriter.LARGE_RECORDS_ALLOWED)) {
-      veniceWriterProperties
-          .put(VeniceWriter.LARGE_RECORDS_ALLOWED, props.getBoolean(VeniceWriter.LARGE_RECORDS_ALLOWED));
-    }
-    veniceWriterProperties.put(VeniceWriter.MAX_RECORD_SIZE_BYTES, pushJobSetting.maxRecordSizeBytes);
     if (pushJobSetting.sslToKafka) {
       veniceWriterProperties.putAll(sslProperties.get());
     }
@@ -2720,11 +2710,12 @@ public class VenicePushJob implements AutoCloseable {
           storeResponse.getStore().isSchemaAutoRegisterFromPushJobEnabled();
       pushJobSetting.isChunkingEnabled = storeResponse.getStore().isChunkingEnabled();
       pushJobSetting.isRmdChunkingEnabled = storeResponse.getStore().isRmdChunkingEnabled();
-      pushJobSetting.maxRecordSizeBytes = storeResponse.getStore().getMaxRecordSizeBytes();
       pushJobSetting.storeCompressionStrategy = storeResponse.getStore().getCompressionStrategy();
       pushJobSetting.isStoreWriteComputeEnabled = storeResponse.getStore().isWriteComputationEnabled();
       pushJobSetting.isStoreIncrementalPushEnabled = storeResponse.getStore().isIncrementalPushEnabled();
       pushJobSetting.hybridStoreConfig = storeResponse.getStore().getHybridStoreConfig();
+      final boolean isRepush = pushJobSetting.isSourceKafka || pushJobSetting.isSourceETL; // safer to simply allow them
+      pushJobSetting.maxRecordSizeBytes = (isRepush) ? -1 : storeResponse.getStore().getMaxRecordSizeBytes();
     }
     return pushJobSetting.storeResponse;
   }
@@ -2755,7 +2746,6 @@ public class VenicePushJob implements AutoCloseable {
     propKeyValuePairs.add(
         "Total input data file size: " + ((double) inputFileDataSize / 1024 / 1024)
             + " MB. This could be the size of compressed data if the underlying filesystem compresses it");
-    propKeyValuePairs.add("Large Records Allowed: " + pushJobSetting.largeRecordsAllowed);
     propKeyValuePairs.add("Max Venice Record Size: " + pushJobSetting.maxRecordSizeBytes);
     propKeyValuePairs.add("Is Chunking Enabled: " + pushJobSetting.chunkingEnabled);
     propKeyValuePairs.add("Is Replication Metadata Chunking Enabled: " + pushJobSetting.rmdChunkingEnabled);
--- a/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
+++ b/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
@@ -2322,6 +2322,7 @@ public class VenicePushJob implements AutoCloseable {
           new VeniceWriterOptions.Builder(pushJobSetting.topic).setUseKafkaKeySerializer(true)
               .setPartitioner(partitioner)
               .setPartitionCount(pushJobSetting.partitionCount)
+              .setMaxRecordSizeBytes(pushJobSetting.maxRecordSizeBytes)
               .build();
       VeniceWriter<KafkaKey, byte[], byte[]> newVeniceWriter = veniceWriterFactory.createVeniceWriter(vwOptions);
       LOGGER.info("Created VeniceWriter: {}", newVeniceWriter);
@@ -2332,20 +2333,19 @@ public class VenicePushJob implements AutoCloseable {
 
   private synchronized Properties getVeniceWriterProperties(PushJobSetting pushJobSetting) {
     if (veniceWriterProperties == null) {
-      veniceWriterProperties = createVeniceWriterProperties(pushJobSetting);
+      veniceWriterProperties = createVeniceWriterProperties(pushJobSetting.kafkaUrl, pushJobSetting.sslToKafka);
     }
     return veniceWriterProperties;
   }
 
-  private Properties createVeniceWriterProperties(PushJobSetting pushJobSetting) {
+  private Properties createVeniceWriterProperties(String kafkaUrl, boolean sslToKafka) {
     Properties veniceWriterProperties = new Properties();
-    veniceWriterProperties.put(KAFKA_BOOTSTRAP_SERVERS, pushJobSetting.kafkaUrl);
+    veniceWriterProperties.put(KAFKA_BOOTSTRAP_SERVERS, kafkaUrl);
     veniceWriterProperties.put(VeniceWriter.MAX_ELAPSED_TIME_FOR_SEGMENT_IN_MS, -1);
-    veniceWriterProperties.put(VeniceWriter.MAX_RECORD_SIZE_BYTES, pushJobSetting.maxRecordSizeBytes);
     if (props.containsKey(VeniceWriter.CLOSE_TIMEOUT_MS)) { /* Writer uses default if not specified */
       veniceWriterProperties.put(VeniceWriter.CLOSE_TIMEOUT_MS, props.getInt(VeniceWriter.CLOSE_TIMEOUT_MS));
     }
-    if (pushJobSetting.sslToKafka) {
+    if (sslToKafka) {
       veniceWriterProperties.putAll(sslProperties.get());
     }
     veniceWriterProperties.setProperty(
@@ -2715,8 +2715,9 @@ public class VenicePushJob implements AutoCloseable {
       pushJobSetting.isStoreIncrementalPushEnabled = storeResponse.getStore().isIncrementalPushEnabled();
       pushJobSetting.hybridStoreConfig = storeResponse.getStore().getHybridStoreConfig();
       pushJobSetting.maxRecordSizeBytes = storeResponse.getStore().getMaxRecordSizeBytes();
-      if (pushJobSetting.maxRecordSizeBytes != -1 && (pushJobSetting.isSourceKafka || pushJobSetting.isSourceETL)) {
-        pushJobSetting.maxRecordSizeBytes = -1; // safer to simply allow large records on repush
+      final boolean isRepush = pushJobSetting.isSourceKafka || pushJobSetting.isSourceETL;
+      if (pushJobSetting.maxRecordSizeBytes != VeniceWriter.DEFAULT_MAX_RECORD_SIZE_BYTES && isRepush) {
+        pushJobSetting.maxRecordSizeBytes = VeniceWriter.DEFAULT_MAX_RECORD_SIZE_BYTES; // safer to allow on repush
         final String repushJobType = (pushJobSetting.isSourceKafka) ? "Kafka" : "ETL";
         LOGGER.info("Setting max record size to unlimited for {} repush job", repushJobType);
       }
--- a/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
+++ b/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
@@ -2716,8 +2716,8 @@ public class VenicePushJob implements AutoCloseable {
       pushJobSetting.hybridStoreConfig = storeResponse.getStore().getHybridStoreConfig();
       pushJobSetting.maxRecordSizeBytes = storeResponse.getStore().getMaxRecordSizeBytes();
       final boolean isRepush = pushJobSetting.isSourceKafka || pushJobSetting.isSourceETL;
-      if (pushJobSetting.maxRecordSizeBytes != VeniceWriter.DEFAULT_MAX_RECORD_SIZE_BYTES && isRepush) {
-        pushJobSetting.maxRecordSizeBytes = VeniceWriter.DEFAULT_MAX_RECORD_SIZE_BYTES; // safer to allow on repush
+      if (isRepush && pushJobSetting.maxRecordSizeBytes != VeniceWriter.UNLIMITED_MAX_RECORD_SIZE) {
+        pushJobSetting.maxRecordSizeBytes = VeniceWriter.UNLIMITED_MAX_RECORD_SIZE; // safer to allow on repush
         final String repushJobType = (pushJobSetting.isSourceKafka) ? "Kafka" : "ETL";
         LOGGER.info("Setting max record size to unlimited for {} repush job", repushJobType);
       }
--- a/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
+++ b/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
@@ -5,6 +5,7 @@ import static com.linkedin.venice.ConfigKeys.KAFKA_BOOTSTRAP_SERVERS;
 import static com.linkedin.venice.ConfigKeys.KAFKA_PRODUCER_DELIVERY_TIMEOUT_MS;
 import static com.linkedin.venice.ConfigKeys.KAFKA_PRODUCER_REQUEST_TIMEOUT_MS;
 import static com.linkedin.venice.ConfigKeys.KAFKA_PRODUCER_RETRIES_CONFIG;
+import static com.linkedin.venice.ConfigKeys.LARGE_RECORDS_ALLOWED;
 import static com.linkedin.venice.ConfigKeys.VENICE_PARTITIONERS;
 import static com.linkedin.venice.VeniceConstants.DEFAULT_SSL_FACTORY_CLASS_NAME;
 import static com.linkedin.venice.hadoop.VenicePushJobConstants.ALLOW_DUPLICATE_KEY;
@@ -378,6 +379,7 @@ public class VenicePushJob implements AutoCloseable {
     pushJobSettingToReturn.batchNumBytes = props.getInt(BATCH_NUM_BYTES_PROP, DEFAULT_BATCH_BYTES_SIZE);
     pushJobSettingToReturn.isIncrementalPush = props.getBoolean(INCREMENTAL_PUSH, false);
     pushJobSettingToReturn.isDuplicateKeyAllowed = props.getBoolean(ALLOW_DUPLICATE_KEY, false);
+    pushJobSettingToReturn.largeRecordsAllowed = props.getBoolean(LARGE_RECORDS_ALLOWED, true);
     pushJobSettingToReturn.enablePushJobStatusUpload = props.getBoolean(PUSH_JOB_STATUS_UPLOAD_ENABLE, false);
     pushJobSettingToReturn.controllerRetries = props.getInt(CONTROLLER_REQUEST_RETRY_ATTEMPTS, 1);
     pushJobSettingToReturn.controllerStatusPollRetries = props.getInt(POLL_STATUS_RETRY_ATTEMPTS, 15);
@@ -513,6 +515,10 @@ public class VenicePushJob implements AutoCloseable {
           props.getBoolean(USE_MAPPER_TO_BUILD_DICTIONARY, DEFAULT_USE_MAPPER_TO_BUILD_DICTIONARY);
       pushJobSettingToReturn.compressionMetricCollectionEnabled =
           props.getBoolean(COMPRESSION_METRIC_COLLECTION_ENABLED, DEFAULT_COMPRESSION_METRIC_COLLECTION_ENABLED);
+    }
+
+    if (pushJobSettingToReturn.isSourceKafka || pushJobSettingToReturn.isSourceETL) {
+      pushJobSettingToReturn.largeRecordsAllowed = true; // safer to simply allow large records on repush
     }
 
     // Compute-engine abstraction related configs
@@ -1710,7 +1716,7 @@ public class VenicePushJob implements AutoCloseable {
    *
    * @return Error message if there is any error detected in the reporter counter and {@code null} otherwise
    */
-  private String updatePushJobDetailsWithJobDetails(DataWriterTaskTracker dataWriterTaskTracker) {
+  String updatePushJobDetailsWithJobDetails(DataWriterTaskTracker dataWriterTaskTracker) {
     // Quota exceeded
     final long totalInputDataSizeInBytes =
         dataWriterTaskTracker.getTotalKeySize() + dataWriterTaskTracker.getTotalValueSize();
@@ -1747,11 +1753,16 @@ public class VenicePushJob implements AutoCloseable {
     final long recordTooLargeFailureCount = dataWriterTaskTracker.getRecordTooLargeFailureCount();
     if (recordTooLargeFailureCount > 0) {
       updatePushJobDetailsWithCheckpoint(PushJobCheckpoints.RECORD_TOO_LARGE_FAILED);
-      String errorMessage = String.format(
+
+      // maxSizeForUserPayloadPerMessageInBytes < 1MB (Kafka record limit); errors appear when chunking is not enabled
+      // maxRecordSizeBytes can be much larger and should take effect only once chunking is enabled
+      final int recordSizeLimit = (pushJobDetails.chunkingEnabled)
+          ? getVeniceWriter(pushJobSetting).getMaxRecordSizeBytes()
+          : getVeniceWriter(pushJobSetting).getMaxSizeForUserPayloadPerMessageInBytes();
+      final String errorMessage = String.format(
           "Input data has at least %d records that exceed the maximum record limit of %s",
           recordTooLargeFailureCount,
-          generateHumanReadableByteCountString(
-              getVeniceWriter(pushJobSetting).getMaxSizeForUserPayloadPerMessageInBytes()));
+          generateHumanReadableByteCountString(recordSizeLimit));
       return errorMessage;
     }
     return null;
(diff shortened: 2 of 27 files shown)


nisargthakkar:
Hi @KaiSernLim . I haven't gone through the changes, but I would like to call out one edge case we should handle - repush jobs should probably still allow writing large records

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
```diff
@@ -889,12 +902,16 @@ public CompletableFuture<PubSubProduceResult> put(
...
-        + replicationMetadataPayloadSize > maxSizeForUserPayloadPerMessageInBytes) {
-      if (isChunkingEnabled) {
+
+    /**
+     * {@link RecordTooLargeException} will be thrown unless the record size fits within one of the following categories:
+     * Records without Chunking < ~1MB < Records with Chunking < MAX_RECORD_SIZE_BYTES
+     */
+    int veniceRecordSize = serializedKey.length + serializedValue.length + replicationMetadataPayloadSize;
+    if (isChunkingNeededForRecord(veniceRecordSize) /* ~1MB default */) {
+      if (isChunkingEnabled && !isRecordTooLarge(veniceRecordSize) /* 10MB default */) {
```
nisargthakkar, line 914: We don't use inline `/* */` comment syntax. To keep consistency, can we replace it with `//` comment syntax at the end of the line?

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
```diff
@@ -1747,11 +1751,17 @@ private String updatePushJobDetailsWithJobDetails(DataWriterTaskTracker dataWrit
...
+      // maxRecordSizeBytes can be much larger and should take effect only once chunking is enabled
+      final int recordSizeLimit = (pushJobDetails.chunkingEnabled)
+          ? getVeniceWriter(pushJobSetting).getMaxRecordSizeBytes()
+          : getVeniceWriter(pushJobSetting).getMaxSizeForUserPayloadPerMessageInBytes();
+      final String errorMessage = String.format(
+          "Input data has at least %d records that exceed the maximum %s record limit of %s",
           recordTooLargeFailureCount,
-          generateHumanReadableByteCountString(
-              getVeniceWriter(pushJobSetting).getMaxSizeForUserPayloadPerMessageInBytes()));
+          (pushJobDetails.chunkingEnabled) ? "Venice" : "Kafka",
```
nisargthakkar, line 1765: Let's not confuse the users with different error messages. We don't really need the system name here

clients/venice-push-job/src/test/java/com/linkedin/venice/hadoop/VenicePushJobTest.java
```diff
@@ -877,6 +881,49 @@ public void testGetPerColoPushJobDetailsStatusFromExecutionStatus() {
...
+      setPushJobSettingDefaults(pushJobSetting);
+      vpj.setInputStorageQuotaTracker(mock(InputStorageQuotaTracker.class));
+      final DataWriterTaskTracker dataWriterTaskTracker = mock(DataWriterTaskTracker.class);
+      doReturn(1L).when(dataWriterTaskTracker).getRecordTooLargeFailureCount();
+
+      // The value of chunkingEnabled should dictate the error message returned
+      final String errorMessage = vpj.updatePushJobDetailsWithJobDetails(dataWriterTaskTracker);
+      final int latestCheckpoint = pushJobDetails.pushJobLatestCheckpoint;
+      Assert.assertTrue(errorMessage.contains((chunkingEnabled) ? "Venice" : "Kafka"), errorMessage);
+      Assert.assertTrue(errorMessage.contains((chunkingEnabled) ? "10.0 MiB" : "950.0 KiB"), errorMessage);
```
nisargthakkar, line 904: Instead of splitting this across two asserts, can we check once for the composed error message? It will help with readability

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
```diff
@@ -86,6 +87,7 @@
 import static com.linkedin.venice.utils.AvroSupersetSchemaUtils.validateSubsetValueSchema;
 import static com.linkedin.venice.utils.ByteUtils.generateHumanReadableByteCountString;
 
+import com.google.common.annotations.VisibleForTesting;
```
nisargthakkar, line 90: We don't use Guava directly, but instead prefer to set a comment if we want to

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
```diff
@@ -378,6 +380,7 @@ private PushJobSetting getPushJobSetting(VeniceProperties props) {
     pushJobSettingToReturn.batchNumBytes = props.getInt(BATCH_NUM_BYTES_PROP, DEFAULT_BATCH_BYTES_SIZE);
     pushJobSettingToReturn.isIncrementalPush = props.getBoolean(INCREMENTAL_PUSH, false);
     pushJobSettingToReturn.isDuplicateKeyAllowed = props.getBoolean(ALLOW_DUPLICATE_KEY, false);
+    pushJobSettingToReturn.largeRecordsAllowed = props.getBoolean(LARGE_RECORDS_ALLOWED, true);
```
nisargthakkar, line 383: Can't we infer this from the store config (`!= -1`)? Why do we need an explicit config?

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
```diff
@@ -889,12 +902,16 @@ public CompletableFuture<PubSubProduceResult> put(
...
-
     int replicationMetadataPayloadSize = putMetadata == null ? 0 : putMetadata.getSerializedSize();
     isChunkingFlagInvoked = true;
-    if (serializedKey.length + serializedValue.length
-        + replicationMetadataPayloadSize > maxSizeForUserPayloadPerMessageInBytes) {
-      if (isChunkingEnabled) {
+
+    /**
+     * {@link RecordTooLargeException} will be thrown unless the record size fits within one of the following categories:
+     * Records without Chunking < ~1MB < Records with Chunking < MAX_RECORD_SIZE_BYTES
```
nisargthakkar, line 910: This comment is confusing. The `~1MB < Records with Chunking < MAX_RECORD_SIZE_BYTES` is not always true. Records in stores with chunking can be smaller than 1 MB too.

gaojieliu:
> Hi @KaiSernLim . I haven't gone through the changes, but I would like to call out one edge case we should handle - repush jobs should probably still allow writing large records

Just a side question, do we need to cover the repush from ETL as well?

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/PushJobSetting.java
```diff
@@ -114,6 +114,8 @@ public class PushJobSetting implements Serializable {
   public Map<String, String> partitionerParams;
   public boolean chunkingEnabled;
   public boolean rmdChunkingEnabled;
+  public boolean largeRecordsAllowed = true;
+  public int maxRecordSizeBytes = -1;
```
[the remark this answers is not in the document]
gaojieliu: Yeah, also `-1` indicates that the large records are allowed, right?

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
```diff
@@ -2327,38 +2340,36 @@ private synchronized VeniceWriter<KafkaKey, byte[], byte[]> getVeniceWriter(Push
...
-    if (props.containsKey(KAFKA_PRODUCER_DELIVERY_TIMEOUT_MS)) {
-      veniceWriterProperties
-          .setProperty(KAFKA_PRODUCER_DELIVERY_TIMEOUT_MS, props.getString(KAFKA_PRODUCER_DELIVERY_TIMEOUT_MS));
-    } else {
-      veniceWriterProperties.setProperty(KAFKA_PRODUCER_DELIVERY_TIMEOUT_MS, Integer.toString(Integer.MAX_VALUE));
-    }
+          .put(VeniceWriter.LARGE_RECORDS_ALLOWED, props.getBoolean(VeniceWriter.LARGE_RECORDS_ALLOWED));
+    }
+    // TODO: difference between venicewriter and non-vw properties?
+    veniceWriterProperties.put(VeniceWriter.MAX_RECORD_SIZE_BYTES, pushJobSetting.maxRecordSizeBytes);
```
gaojieliu, line 2360: Can we collapse these two configs into one?
When `VeniceWriter.MAX_RECORD_SIZE_BYTES` is `-1`, VeniceWriter would treat the max limit as unlimited.

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
```diff
@@ -147,9 +148,9 @@ public class VeniceWriter<K, V, U> extends AbstractVeniceWriter<K, V, U> {
   public static final int DEFAULT_MAX_SIZE_FOR_USER_PAYLOAD_PER_MESSAGE_IN_BYTES = 950 * 1024;
 
   /**
-   * The default value of -1 is overwritten with 10 MB, if not specified in the VeniceProperties from the constructor.
+   * The default value of -1 from the store is overwritten with a new default of 10 MB.
    */
-  // public static final int DEFAULT_MAX_RECORD_SIZE_BYTES = 10 * 1024 * 1024;
+  public static final int DEFAULT_MAX_RECORD_SIZE_BYTES = 10 * 1024 * 1024;
```
gaojieliu, line 153: I wonder whether this can be controlled by the backend or not.
For example, we can introduce a controller config: `controller.default.max.record.size.bytes` and Controller can backfill the max record size if the store-level config is absent (-1) in StoreResponse when VenicePushJob is asking for it.
With this way, when we need to change the default max size, we only need to deploy a config change to the Controller and we can control the default max config per cluster.

gaojieliu (changes requested):

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/PushJobSetting.java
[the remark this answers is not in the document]
KaiSernLim: Removed.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
```diff
@@ -2942,4 +2956,14 @@ private static void runPushJob(String jobId, Properties props) {
...
     this.dataWriterComputeJob = dataWriterComputeJob;
   }
+
+  @VisibleForTesting
+  protected void setInputStorageQuotaTracker(InputStorageQuotaTracker inputStorageQuotaTracker) {
+    this.inputStorageQuotaTracker = inputStorageQuotaTracker;
+  }
+
+  @VisibleForTesting
+  protected PushJobDetails getPushJobDetails() {
```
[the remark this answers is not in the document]
KaiSernLim: Done.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
KaiSernLim: Removed.

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
KaiSernLim: Done.

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
KaiSernLim: Changed to `Chunking Not Needed < ~1MB < Chunking Needed < MAX_RECORD_SIZE_BYTES`.

By `Records with Chunking`, the chunking is referring to records that are already chunked rather than pertaining to whether chunking is enabled on the store, but maybe the new comment is more clear?

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
```diff
@@ -2327,38 +2338,36 @@ private synchronized VeniceWriter<KafkaKey, byte[], byte[]> getVeniceWriter(Push
...
-    if (props.containsKey(KAFKA_PRODUCER_DELIVERY_TIMEOUT_MS)) {
+    if (props.containsKey(VeniceWriter.LARGE_RECORDS_ALLOWED)) {
       veniceWriterProperties
-          .setProperty(KAFKA_PRODUCER_DELIVERY_TIMEOUT_MS, props.getString(KAFKA_PRODUCER_DELIVERY_TIMEOUT_MS));
-    } else {
-      veniceWriterProperties.setProperty(KAFKA_PRODUCER_DELIVERY_TIMEOUT_MS, Integer.toString(Integer.MAX_VALUE));
-    }
+          .put(VeniceWriter.LARGE_RECORDS_ALLOWED, props.getBoolean(VeniceWriter.LARGE_RECORDS_ALLOWED));
+    }
+    // TODO: difference between venicewriter and non-vw properties?
```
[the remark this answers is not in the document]
KaiSernLim: Removed.

But I did want to come back to this `TODO` to clarify how these string constants work.
1. So this `max_record_size_bytes` is a store-level config, so it's present in `ControllerApiConstants`.
2. But then it also needs to be in `VeniceWriter`, so there's `venice.writer.max.record.size.bytes`.
3. But it's also part of `VenicePushJob`, so maybe there should be one in `VenicePushJobConstants` as well? But it's just set in `VenicePushJob.getStoreResponse()` without any usage of string constants.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
KaiSernLim: Removed.

I actually wanted this mainly for our debugging purposes to determine whether the failure is due to the `~1MB` limit or the new `maxRecordSizeBytes` limit.

I find it a lot clearer when the system name is mentioned, otherwise we would have to look at the record limit number in the error message to determine which version of the failure is being hit.

clients/venice-push-job/src/test/java/com/linkedin/venice/hadoop/VenicePushJobTest.java
KaiSernLim: Irrelevant since [the system name was removed](https://github.com/linkedin/venice/pull/1052#discussion_r1669385417). I originally checked for the full error message, but I felt that it was slightly harder to read because it was so long.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
KaiSernLim: [See this comment.](https://github.com/linkedin/venice/pull/1052#discussion_r1669462089)

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
KaiSernLim: As you and @nisargthakkar noticed, it's not needed in the context of just this PR and enforcing the limit on Batch Push Jobs.

This feature flag is to protect against `VeniceWriter` unintentionally enforcing limits on nearline jobs, which will be handled in a later PR.

1. The conversion of the unset value (`-1`) into the default limit (currently `10MB`) happens in the `VeniceWriter`, because both the batch and nearline job code paths will go through it, and that's also where the check happens ([link](https://github.com/linkedin/venice/blob/4ba73c94710e8fad78e4f82c58f70a2b1a5b8bfc/internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java#L896)).
2. It also comes with the benefit of having a feature flag so we can easily turn off and on the feature if we need to do so in the future.

The other way to do it would be to set the default limit individually inside the `VenicePushJob` and `Consumer`, but I felt that the feature flag would be a clean approach that has added benefits.


As mentioned in your original design doc:

> Venice will have a default max allowed record size when chunking is enabled, such as 5MB (for example).

I noticed that there is already a `~1MB` limit ([`maxSizeForUserPayloadPerMessageInBytes`](https://github.com/linkedin/venice/blob/4ba73c94710e8fad78e4f82c58f70a2b1a5b8bfc/internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java#L896)) to handle the Kafka size limit, and if chunking is not enabled then an exception will be thrown. So if there are records that need to be chunked, chunking should be enabled and the default max allowed record size should always take effect after the `~1MB` limit.

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
KaiSernLim: I would be okay with implementing this, but I do find it a bit odd to have a configurable setting that's supposed to be a `"default"`.

Alternatively, I'm hoping we can figure out a good default value that won't need to be changed in the future. We are still waiting on the assembled value size metric to be rolled out to production, so we can see how large the records really are:

- https://github.com/linkedin/venice/pull/1027

If we were to go with this approach, I would prefer this to be done in a follow-up PR to keep this PR focused.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
```diff
@@ -1747,11 +1753,17 @@ private String updatePushJobDetailsWithJobDetails(DataWriterTaskTracker dataWrit
     final long recordTooLargeFailureCount = dataWriterTaskTracker.getRecordTooLargeFailureCount();
     if (recordTooLargeFailureCount > 0) {
       updatePushJobDetailsWithCheckpoint(PushJobCheckpoints.RECORD_TOO_LARGE_FAILED);
-      String errorMessage = String.format(
-          "Input data has at least %d records that exceed the maximum record limit of %s",
+
+      // maxSizeForUserPayloadPerMessageInBytes < 1MB (Kafka record limit); errors appear when chunking is not enabled
+      // maxRecordSizeBytes can be much larger and should take effect only once chunking is enabled
+      final int recordSizeLimit = (pushJobDetails.chunkingEnabled)
```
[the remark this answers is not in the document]
KaiSernLim: That's a good catch. Maybe it would be better to set the default limit in `VenicePushJob` as opposed to [centrally in `VeniceWriter`](https://github.com/linkedin/venice/pull/1052#discussion_r1672703648) then.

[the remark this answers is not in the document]
gaojieliu: I think the default limit should be controlled by via a backend config in Controller, and the config can be in cluster level.
So when Controller returns the store response, it can backfill the record size limit by the config if the value is `-1`.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
gaojieliu: It is fine if it can make the logic easier, but I don't like to hard-code the max limit in `VeniceWriter` and it needs to be configurable as we may change it and apply it globally for the stores, whose store config value is `-1`.

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
gaojieliu: Well, this is more like a existing practice.
For example, we do have a config to control the max batch-get limit and by default, it is `-1` in store level, and we control the true default value via a cluster-level config, and in most of the clusters, we set it to be `150`.
We are evaluating whether we can bump the default up to some higher number, and it will be easy to do that as we just need to deploy a config change without changing the code or touching every store.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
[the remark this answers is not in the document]
KaiSernLim: Will do.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
KaiSernLim: With that approach, I will remove the feature flag and the two code path locations (`VenicePushJob` for batch and `Consumer` for nearline) will read the default value from the controller and that's how the limit will be enforced.

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
KaiSernLim: Ah, if it's an existing practice then there are still many more things such as this that I need to familiarize myself with. I'll work on these changes right now, but I feel like creating a separate PR for that and merging that in first? I don't want to overcomplicate this PR.

Ideally it should've been done in the previous PR for just adding the max limit, but I misunderstood in our earlier conversation:
- https://github.com/linkedin/venice/pull/1027

KaiSernLim: Nevermind, I'll just keep it in this PR. I thought the changes would span many files like for the original setting:
- https://github.com/linkedin/venice/pull/1027

But I found [this commit](https://github.com/linkedin/venice/commit/9612ff6f92aa6128384724eecb91f19ea4414b57) and used it as an example, and only a few files were touched.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
```diff
@@ -2716,6 +2714,8 @@ private StoreResponse getStoreResponse(String storeName, boolean refresh) {
       pushJobSetting.isStoreWriteComputeEnabled = storeResponse.getStore().isWriteComputationEnabled();
       pushJobSetting.isStoreIncrementalPushEnabled = storeResponse.getStore().isIncrementalPushEnabled();
       pushJobSetting.hybridStoreConfig = storeResponse.getStore().getHybridStoreConfig();
+      final boolean isRepush = pushJobSetting.isSourceKafka || pushJobSetting.isSourceETL; // safer to simply allow them
+      pushJobSetting.maxRecordSizeBytes = (isRepush) ? -1 : storeResponse.getStore().getMaxRecordSizeBytes();
```
gaojieliu, line 2718: Nit,
Shall we leave an info logs for this behavior?

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
gaojieliu: Do we still need this?
When it is `-1`, it indicates that the max record size is infinite, right?

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
```diff
@@ -304,7 +304,11 @@ public VeniceWriter(
     this.isChunkingEnabled = params.isChunkingEnabled();
     this.isChunkingSet = true;
     this.isRmdChunkingEnabled = params.isRmdChunkingEnabled();
-    // this.maxRecordSizeBytes = props.getInt(MAX_RECORD_SIZE_BYTES, DEFAULT_MAX_RECORD_SIZE_BYTES);
+    int maxRecordSizeBytesFromProps = props.getInt(MAX_RECORD_SIZE_BYTES, DEFAULT_MAX_RECORD_SIZE_BYTES);
+    if (maxRecordSizeBytesFromProps == -1) {
+      maxRecordSizeBytesFromProps = DEFAULT_MAX_RECORD_SIZE_BYTES;
```
gaojieliu, line 309: Is this logic still required after Controller backfills the default value in `StoreResponse`?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerConfig.java
```diff
@@ -586,6 +589,7 @@ public VeniceControllerConfig(VeniceProperties props) {
     this.danglingTopicCleanupIntervalSeconds = props.getLong(CONTROLLER_DANGLING_TOPIC_CLEAN_UP_INTERVAL_SECOND, -1);
     this.danglingTopicOccurrenceThresholdForCleanup =
         props.getInt(CONTROLLER_DANGLING_TOPIC_OCCURRENCE_THRESHOLD_FOR_CLEANUP, 3);
+    this.defaultMaxRecordSizeBytes = props.getInt(CONTROLLER_DEFAULT_MAX_RECORD_SIZE_BYTES, -1);
```
gaojieliu, line 592: I think we should put a valid default, such as `100MB` instead of `-1` here, and maybe we should put a sanity check when loading the config to prevent negative values or the default value should be at least `1MB`.

internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
```diff
@@ -2161,4 +2161,10 @@ private ConfigKeys() {
    */
   public static final String CONTROLLER_DANGLING_TOPIC_OCCURRENCE_THRESHOLD_FOR_CLEANUP =
       "controller.dangling.topic.occurrence.threshold.for.cleanup";
+
+  /**
+   * Controller config for the default value of {@link com.linkedin.venice.writer.VeniceWriter#maxRecordSizeBytes}.
+   * Only used in batch push and nearline jobs.
+   */
+  public static final String CONTROLLER_DEFAULT_MAX_RECORD_SIZE_BYTES = "controller.default.max.record.size.bytes";
```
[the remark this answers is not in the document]
KaiSernLim: This config should still be used on the partial update path where the Consumer re-assembles the chunked record and applies the partial update right? That's what I was referring to for nearline jobs, and I will rephrase the comment to say partial update path only.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerConfig.java
KaiSernLim: Changed default value to `100MB` and added check for `< 1 MB` to throw `VeniceException`.

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
KaiSernLim: Removed. Missed `VeniceWriter`.

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
KaiSernLim: Oh I missed `VeniceWriter` when removing stuff from `VenicePushJob`. Changed default value to `-1` now.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
KaiSernLim: Are you talking about only the `isRepush` code path? I feel like it should be a certain `-1` for that case?

If you're talking about for all cases, then I'm concerned it might be too verbose, but I'm not knowledgeable about this.

internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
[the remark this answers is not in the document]
gaojieliu: `Nearline job` typically means the Samza/Flink jobs, where large put/update is banned and in Venice Server ingestion path, which would consume the records produced by `nearline job`, large records are allowed.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
gaojieliu: yes, for the repush path.

KaiSernLim: Okay added the comment:
```java
if (pushJobSetting.maxRecordSizeBytes != -1 && (pushJobSetting.isSourceKafka || pushJobSetting.isSourceETL)) {
  pushJobSetting.maxRecordSizeBytes = -1; // safer to simply allow large records on repush
  final String repushJobType = (pushJobSetting.isSourceKafka) ? "Kafka" : "ETL";
  LOGGER.info("Setting max record size to unlimited for {} repush job", repushJobType);
}
```

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
```diff
@@ -319,6 +320,11 @@ public VeniceWriter(
                 + " is true and the topic is Version Topic");
       }
     }
+    if (maxRecordSizeBytes != -1 && maxSizeForUserPayloadPerMessageInBytes > maxRecordSizeBytes) {
+      throw new VeniceException(
+          MAX_SIZE_FOR_USER_PAYLOAD_PER_MESSAGE_IN_BYTES + '=' + maxSizeForUserPayloadPerMessageInBytes
+              + " cannot be set higher than " + MAX_RECORD_SIZE_BYTES + '=' + maxRecordSizeBytes);
```
nisargthakkar, line 326: nit: What do you think of this kind of error message?
```
venice.writer.max.size.for.user.payload.per.message.in.bytes (<maxSizeForUserPayloadPerMessageInBytes>) cannot be set higher than venice.writer.max.record.size.bytes (<maxRecordSizeBytes>)
```

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
```diff
@@ -1588,7 +1599,8 @@ private String getSizeReport(int serializedKeySize, int serializedValueSize, int
     return "Key size: " + serializedKeySize + " bytes, " + "Value size: " + serializedValueSize + " bytes, "
         + "Replication Metadata size: " + replicationMetadataPayloadSize + " bytes, " + "Total payload size: "
         + (serializedKeySize + serializedValueSize + replicationMetadataPayloadSize) + " bytes, "
-        + "Max available payload size: " + maxSizeForUserPayloadPerMessageInBytes + " bytes.";
+        + "Max available payload size: " + maxSizeForUserPayloadPerMessageInBytes + " bytes"
+        + ((maxRecordSizeBytes == -1) ? "" : ", Max Venice record size: " + maxRecordSizeBytes + " bytes") + ".";
```
nisargthakkar, line 1603: Checking - Do we need one more set of parenthesis for the `else` part? When `maxRecordSizeBytes == -1`, what will this string resolve to?
```
Max available payload size: XYZ bytes.
```
Or 
```
Max available payload size: XYZ bytes, -1 bytes.
```

nisargthakkar (commented):
LGTM. A couple of nits, but they could be dropped as they are subjective

nisargthakkar (dismissed):

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestBatch.java
```diff
@@ -986,7 +987,23 @@ public void testLargeValues() throws Exception {
...
     int minSize = 1024 * 1024; // 1MB apiece
-    validatePerStoreMetricsRange(storeName, ASSEMBLED_RECORD_VALUE_SIZE_IN_BYTES, minSize, MAX_RECORD_VALUE_SIZE);
+    validatePerStoreMetricsRange(storeName, ASSEMBLED_RECORD_VALUE_SIZE_IN_BYTES, minSize, LARGE_RECORD_VALUE_SIZE);
+  }
+
+  /** Test that values that are too large will fail the push job only when the limit is enforced. */
+  @Test(dataProvider = "True-and-False", dataProviderClass = DataProviderUtils.class, timeOut = TEST_TIMEOUT)
+  public void testStoreWithTooLargeValues(boolean enforceLimit) throws Exception {
+    final int wayTooLargeRecordValueSize = 11 * 1024 * 1024; // 11 MB
+    final int maxRecordSizeBytesForTest = (enforceLimit) ? 10 * 1024 * 1024 : -1; // -1 means no limit
```
nisargthakkar, line 997: Can we replace `1024 * 1024` with `ByteUtils.BYTES_PER_MB` for readability?

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
KaiSernLim: I tested it and it actually works without the parenthesis, but I would add the parenthesis just to be clearer.

In any case, it's been revised to:
```java
+ ", Max record size: "
+ ((maxRecordSizeBytes == UNLIMITED_MAX_RECORD_SIZE) ? "unlimited" : maxRecordSizeBytes) + " bytes.";
```

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestBatch.java
KaiSernLim: Omg I've been looking for that for so long! Replaced.

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
KaiSernLim: Changed.

I'm used to the `=` for the tokenization / parsing of logs, but the `()` is totally fine too.

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriterOptions.java
```diff
@@ -127,6 +143,9 @@ private void addDefaults() {
       if (time == null) {
         time = SystemTime.INSTANCE;
       }
+      if (maxRecordSizeBytes == null) {
+        maxRecordSizeBytes = -1;
```
gaojieliu, line 147: Nit,
Shall we replace `-1` by `VeniceWriter.UNLIMITED_MAX_RECORD_SIZE`?

services/venice-controller/src/main/java/com/linkedin/venice/controller/server/StoresRoutes.java
```diff
@@ -327,6 +327,10 @@ public void internalHandle(Request request, StoreResponse veniceResponse) {
         if (storeInfo.getBackupVersionRetentionMs() < 0) {
           storeInfo.setBackupVersionRetentionMs(admin.getBackupVersionDefaultRetentionMs());
         }
+        // This is the only place the default value of maxRecordSizeBytes is set for StoreResponse for VPJ and Consumer
+        if (storeInfo.getMaxRecordSizeBytes() < 0) {
```
gaojieliu, line 331: Nit,
If this logic is being deployed later than VPJ, VPJ will receive `-1` for this field, right?
I wondered if StoreResponse doesn't contain this field at all because of the deploying order, what would VPJ receive? I guess it might receive `0` as `0` is the default value for `Integer` if the json response doesn't contain such field, so maybe we should update the default value of `StoreInfo#maxRecordSizeBytes` to be `VeniceWriter#UNLIMITED_MAX_RECORD_SIZE`?

gaojieliu (commented):
Code change looks good to me and I just left two more minor comments and please take a look to see whether additional change is required or not.

lluwm (commented):
@KaiSernLim, thanks for putting up and iterating on this change and it looks good to me! Please address comments from other reviewers.

services/venice-controller/src/main/java/com/linkedin/venice/controller/server/StoresRoutes.java
KaiSernLim: Nice catch. Fixed.

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriterOptions.java
KaiSernLim: Nice catch. Fixed.

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriterOptions.java
```diff
@@ -69,6 +70,10 @@ public boolean isRmdChunkingEnabled() {
     return rmdChunkingEnabled;
   }
 
+  public Integer getMaxRecordSizeBytes() {
```
sixpluszero, line 73: nit only, let's resolve it in the upcoming nearline part of PR: we can just use int instead of Integer, same goes to the variable definition

sixpluszero (approved):
Looks good to me. Thank you for your change! There is a single nit point but we can do that in future PR.

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriterOptions.java
KaiSernLim: Changed to `int` and set default value in `Builder` rather than using `Integer` in the nearline changes.

gaojieliu (approved):
Thanks!

PR #1054, opened 2024-06-27
[test] Fixed unit test related to multi threading

sebas-inf:
## Summary

Fixed a unit test to ensure its consistency

## How was this PR tested?

Changes unit test and ran it a few times to ensure consistent results


KaiSernLim (approved):

adamxchen:
@sebas-inf  let's squash commit msgs so it would make the git history clean

PR #1055, opened 2024-07-05
[server][dvc] tweak logging for server

adamxchen:

When the host has a few hundreds or even thousands of stores, this log line spam the log file easily.

## How was this PR tested?

n/a


KaiSernLim (approved):

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBStorageEngineFactory.java
```diff
@@ -280,7 +280,7 @@ public synchronized Set<String> getPersistedStoreNames() {
     File databaseDir = new File(rocksDBPath);
     if (databaseDir.exists() && databaseDir.isDirectory()) {
       String[] storeDirs = databaseDir.list();
-      LOGGER.info("Found the following RocksDB databases: {}", Arrays.toString(storeDirs));
```
majisourav99, line 283: may be this one by default could be helpful to find how many DB the server found ?

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/PubSubClientsFactory.java
```diff
@@ -106,10 +106,10 @@ private static <T> T createFactory(
     String className;
     if (properties.containsKey(configKey)) {
       className = properties.getString(configKey);
-      LOGGER.info("Creating pub-sub {} adapter factory instance for class: {}", factoryType, className);
```
haoxu07, line 109: Seems this is not for rocksdb. Is this too much in production?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBStorageEngineFactory.java
adamxchen: I'm not sure about this.. How are we going to use this information tho? The reason why I wanted to remove this is when a single host have too many replica, this log line would be huge and is hard to parse.

adamxchen: The good thing is it only send out once for startup. Let me know what you think. We can keep it if you find it can be helpful for debugging.

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/PubSubClientsFactory.java
adamxchen: Updated the title. Just some logging change from minor issues I've noticed during oncall. When a same pubsub system is used, say Kafka in this cluster, this log would repeat multiple times according to the number of stores it have. For example, if a host has 1000 replicas and it uses Kafka, then we would have 
```
 [PubSubClientsFactory] [main] Creating pub-sub PRODUCER adapter factory instance with default class: com.linkedin.venice.pubsub.adapter.kafka.producer.ApacheKafkaProducerAdapterFactory
 [PubSubClientsFactory] [main] Creating pub-sub CONSUMER adapter factory instance with default class: com.linkedin.venice.pubsub.adapter.kafka.consumer.ApacheKafkaConsumerAdapterFactory
 [PubSubClientsFactory] [main] Creating pub-sub ADMIN adapter factory instance with default class: com.linkedin.venice.pubsub.adapter.kafka.admin.ApacheKafkaAdminAdapterFactory
```
Repeating 1000 times, which isn't really useful IMO.

adamxchen: If you believe there's a good value to keep it, I can make this a `DEBUG`

haoxu07: Ideally, it should not be that many occurrences, as this is a factory class.

sushantmane: >Repeating 1000 times, which isn't really useful IMO.

That's not good. Do you remember if you saw this on controller or servers?

sushantmane: I checked the code and it should be one log line per client per host

adamxchen: > (For producer's it could be equal to VW factory instances per host)

I recall @haoxu07 did some search on pre-production env. He did observe this line of log is spamming due to the number of `VeniceServerConfig`. @haoxu07  can you please share your findings? I can make the log DEBUG if it has some values

sushantmane: It does help in figuring out what type of clients we are using but we can infer that using other means as well. So it should be ok to make it debug I guess

PR #1056, opened 2024-07-10
[admin-tool] Metadata migration from source ZK to destination ZK

bonytoni:
## Metadata migration from source ZK to destination ZK

- Implemented AdminTool command to migrate Venice-specific metadata from a source ZooKeeper to a destination ZooKeeper.
- Completed manual testing with ei-ltx1 ZooKeeper server and migrated some of its metadata to a local Apache ZooKeeper server in standalone mode.
- Reading and writing `data` for ZNode works!

**Future steps:**
- Folders on ZooKeeper contain `ACLs`, `Stat`, and `Context` so we should look into migrating this information as well.
- We tested migration from a live ZooKeeper server -> local Apache ZooKeeper server. We will test migrating metadata from local Apache ZooKeeper server -> live ZooKeeper server in future end-to-end testing.
- Currently, we are building a tree of paths that we want to migrate from the source ZK, then converting it to a list. We will refactor this into one step in future iterations without having duplicate functionality.

## How was this PR tested?

Unit Tests, Integration Tests, Manual Testing

### DIFF
--- a/clients/venice-admin-tool/src/main/java/com/linkedin/venice/ZkCopier.java
+++ b/clients/venice-admin-tool/src/main/java/com/linkedin/venice/ZkCopier.java
@@ -49,8 +49,10 @@ public class ZkCopier {
     if (!basePath.startsWith("/")) {
       throw new VeniceException("Base path must start with a forward slash (/)");
     }
-    TreeNode srcZkPathsTree = getZkClientPathsTree(srcZkClient, clusterNames, basePath);
-    List<String> destZkPathsList = getVenicePathsFromTree(srcZkPathsTree, clusterNames, basePath);
+    if (!srcZkClient.exists(basePath)) {
+      throw new VeniceException("Base path does not exist in source ZK");
+    }
+    List<String> destZkPathsList = getZkClientPathsList(srcZkClient, clusterNames, basePath);
     srcZkClient.setZkSerializer(new ByteArraySerializer());
     destZkClient.setZkSerializer(new ByteArraySerializer());
     LOGGER.info("Starting Migration");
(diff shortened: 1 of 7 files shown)


clients/venice-admin-tool/src/main/java/com/linkedin/venice/ZkCopier.java
```diff
@@ -1,31 +1,57 @@
...
+      String destZkAddress,
+      Set<String> clusterNames,
+      String basePath) {
+    if (!basePath.startsWith("/")) {
+      throw new VeniceException("Base path must start with a forward slash (/)");
+    }
+    ZkClient srcZkClient = ZkClientFactory.newZkClient(srcZkAddress);
+    TreeNode srcZkPathsTree = getZkClientPathsTree(srcZkClient, clusterNames, basePath);
+    List<String> srcZkPathsList = pathsTreeToList(srcZkPathsTree);
+    List<String> destZkPathsList = getVenicePaths(srcZkPathsList, clusterNames, basePath);
```
nisargthakkar, line 49: We should be able to remove one tree -> list -> tree round trip here through some small refactoring

internal/venice-common/src/main/java/com/linkedin/venice/zk/VeniceZkPaths.java
```diff
@@ -13,4 +19,8 @@ public class VeniceZkPaths {
   public static final String STORES = "Stores";
   public static final String STORE_CONFIGS = "storeConfigs";
   public static final String STORE_GRAVEYARD = "StoreGraveyard";
+  private static final Set<String> CLUSTER_ZK_PATHS_MODIFIABLE = new HashSet<>(
+      Arrays.asList(ADMIN_TOPIC_METADATA, EXECUTION_IDS, PARENT_OFFLINE_PUSHES, ROUTERS, STORE_GRAVEYARD, STORES));
+  /** Set of all Venice-managed ZooKeeper cluster paths */
```
nisargthakkar, line 24: Can we move this comment before `CLUSTER_ZK_PATHS_MODIFIABLE` and add a blank like before the comment to signify that this is different from the variables listed above

nisargthakkar (commented):
I've not had time to take a full look. I left a couple of minor comments though

clients/venice-admin-tool/src/main/java/com/linkedin/venice/ZkCopier.java
```diff
@@ -1,31 +1,69 @@
...
+  public static void migrateVenicePaths(
+      ZkClient srcZkClient,
+      ZkClient destZkClient,
+      Set<String> clusterNames,
+      String basePath) {
+    if (!basePath.startsWith("/")) {
+      throw new VeniceException("Base path must start with a forward slash (/)");
+    }
+    TreeNode srcZkPathsTree = getZkClientPathsTree(srcZkClient, clusterNames, basePath);
+    List<String> destZkPathsList = getVenicePathsFromTree(srcZkPathsTree, clusterNames, basePath);
```
nisargthakkar, line 53: There seems to be duplicate logic in `getZkClientPathsTree` and `getVenicePathsFromTree`. `getZkClientPathsTree` is building a tree of only the zk paths that we expect to copy over. With that tree, I don't think we would expect `getVenicePathsFromTree` to return anything different than all the nodes in `srcZkPathsTree`, right?

bonytoni: Is building the `srcZkPathsTree` with `getZkClientPathsTree()` and then traversing it again to build the list using `pathsTreeToList()` is inefficient since it is passes through the tree twice?
I wrote a new method `getZkClientPathsList` to directly build the list of paths that we need from the ZkClient.

clients/venice-admin-tool/src/main/java/com/linkedin/venice/ZkCopier.java
```diff
@@ -110,6 +159,54 @@ static TreeNode buildRequiredPathsTree(Set<String> clusterNames, String basePath
...
+            getZkClientPathsListHelper(zkClient, paths, sb);
+          }
+          sb.delete(sb.lastIndexOf("/" + venicePath), sb.length());
+        }
+      }
+      sb.delete(sb.lastIndexOf("/" + cluster), sb.length());
+    }
+    Collections.sort(paths);
+    return paths;
+  }
```
nisargthakkar, line 193: A couple of comments in this function:
1. Can we avoid using a single StringBuilder to orchestrate this whole logic, and also do away with the `deletes`? It is hard to make a mental model of things if there is only one object that get mutated quite frequently.
2. Instead of checking `zkClient.exists` multiple times, each of which would incur a remote query to Zk, we should do `zkClient.getChildren`, and then from that list, pick and choose the stuff that we need.

clients/venice-admin-tool/src/main/java/com/linkedin/venice/ZkCopier.java
```diff
@@ -110,6 +159,54 @@ static TreeNode buildRequiredPathsTree(Set<String> clusterNames, String basePath
...
+
+  /**
+   * Recursively traverse the ZK client and add each path to the list.
+   * @Note: This method directly modifies the list {@code paths} in the parent method {@code getZkClientPathsList()}
+   */
+  private static void getZkClientPathsListHelper(ZkClient zkClient, List<String> paths, StringBuilder sb) {
```
nisargthakkar, line 199: I suppose the comment in the previous function about not using the same `StringBuilder` also transitively change this method's implementation. Still calling it out as it is not easy to make sense of this method just by taking a quick look at it.

Also, let's rename `sb` to something more representative

clients/venice-admin-tool/src/main/java/com/linkedin/venice/ZkCopier.java
nisargthakkar: Actually, I would prefer if we reused as much logic between the two as possible. It is harder to maintain two pieces of code that essentially are supposed to do the same thing. Going through the tree twice is okay

clients/venice-admin-tool/src/main/java/com/linkedin/venice/ZkCopier.java
bonytoni: Previous iteration was better https://github.com/linkedin/venice/pull/1056#discussion_r1683600582

clients/venice-admin-tool/src/main/java/com/linkedin/venice/ZkCopier.java
bonytoni: Previous iteration was better https://github.com/linkedin/venice/pull/1056#discussion_r1683600582

nisargthakkar (approved):
LGTM. Thanks @bonytoni

PR #1057, opened 2024-07-11
[da-vinc-client] DVC aggregates push status updates across partitions

huangminchn:
## Summary
This PR enables the DVC to write batching push status updates across all partitions on a node.

During batch pushes, one DVC instance will send two events to the push statuses updates to
push status system stores:
one event indicating the push has started in this DVC node;
and another event indicating that the push has completed in all partitions assigned to this DVC node so far.

A config to control this behavior:
write.batching.push.status - by default, it's disabled. Will update default value after enough testing.

The batching only applies to batch push part, it doesn't affect incremental pushes so far.

## How was this PR tested?
Added integration tests.


clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
```diff
@@ -576,8 +578,35 @@ protected void reportPushStatus(
...
+        versionBackend.updatePartitionStatus(partition, status);
+        if (status == ExecutionStatus.COMPLETED) {
+          if (!versionBackend.isBatchPushEndSignalSent()
+              && versionBackend.areAllPartitionsOnSameTerminalStatus(status)) {
+            pushStatusStoreWriter.writeVersionLevelPushStatus(
+                version.getStoreName(),
+                version.getNumber(),
+                status,
+                versionBackend.getTrackedPartitions());
+            versionBackend.batchPushEndSignalSent();
```
mynameborat, line 592: Consider the scenario where 
1. Get a signal for partition 1 started
2. Get a signal for partition 1 completed
3. Get a signal for partition 2 started
4. Get a signal for partition 2 completed

The above can result in partially sending the completed status w/ only partition 1 information but none for partition 2. 

Is it the above possible? Is there a way to know the all the partitions ahead in time during the `versionBackend` creation?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/config/VeniceServerConfig.java
```diff
@@ -763,6 +765,7 @@ public VeniceServerConfig(VeniceProperties serverProperties, Map<String, Map<Str
         serverProperties.getInt(SERVER_DEDICATED_CONSUMER_POOL_SIZE_FOR_AA_WC_LEADER, 5);
     useDaVinciSpecificExecutionStatusForError =
         serverProperties.getBoolean(USE_DA_VINCI_SPECIFIC_EXECUTION_STATUS_FOR_ERROR, false);
+    writeBatchingPushStatus = serverProperties.getBoolean(WRITE_BATCHING_PUSH_STATUS, false);
```
gaojieliu, line 768: Shall we prefix this by `DaVinci`? So far, this feature seems to be DaVinci specific.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
huangminchn: Sync up offline; the race condition has been discussed before, users could subscribe to a new partition anytime; I will try to add a 5 minutes delay before sending the COMPLETE event.

It will only work for batch-push by using the future version subscription callback to send push status event; the current approach can be extended to support incremental push status updates in future. We will keep the current implementation.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/notifier/PushStatusDelayedUpdateTask.java
```diff
@@ -0,0 +1,128 @@
...
+
+  public void batchPushStartSignalSent() {
+    this.batchPushStartSignalSent = true;
+  }
+
+  public void batchPushEndSignalSent() {
+    this.batchPushEndSignalSent = true;
+  }
+
+  public void updatePartitionStatusAndMaybeSendBatchingStatus(
```
gaojieliu, line 55: It seems this function can be invoked by multiple threads concurrently, so shall we add any safeguard here?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/notifier/PushStatusDelayedUpdateTask.java
```diff
@@ -0,0 +1,128 @@
...
+      int partition,
+      ExecutionStatus status,
+      PushStatusStoreWriter pushStatusStoreWriter) {
+    partitionStatus.put(partition, status);
+    if (status == ExecutionStatus.COMPLETED) {
+      if (!isBatchPushEndSignalSent() && areAllPartitionsOnSameTerminalStatus(status)) {
+        // Create a delayed task which wakes up after 30 seconds and check whether all partitions are still on
+        // COMPLETED status. If so, send the COMPLETE status to the push status store.
+        // If there is already a task scheduled, cancel it and schedule a new one.
+        scheduleDelayedTaskToWriteBatchingStatus(status, () -> {
```
gaojieliu, line 65: I wondered whether we need to need this a little more complex logic here.
I am thinking the following way:
1. This function would only collect the per-partition push status.
2. The scheduler would just check the partition status map every 30 seconds (for example), and if all partitions are completed, then send out the completed status, otherwise send out the start status if it hasn't been sent out yet.
3. The scheduler will shutdown itself after sending out `completed` status to the backend.

This might be slightly less efficient than the on-demand trigger, but the overhead should be very minimal, but I feel the logic will be simpler to reason.

WDYT?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/config/VeniceServerConfig.java
huangminchn: Yup, done!

clients/da-vinci-client/src/main/java/com/linkedin/davinci/notifier/PushStatusDelayedUpdateTask.java
huangminchn: I can add synchronize keyword here; currently there is only one thread handling the status report; could you please check this in DaVinciBackend:
`private final ExecutorService ingestionReportExecutor = Executors.newSingleThreadExecutor();`

clients/da-vinci-client/src/main/java/com/linkedin/davinci/notifier/PushStatusDelayedUpdateTask.java
huangminchn: Agree. It's easier to reason. However, 3 might be reverted when we start to support batching incremental push status.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/notifier/PushStatusDelayedUpdateTask.java
```diff
@@ -0,0 +1,128 @@
...
+ */
+public class PushStatusDelayedUpdateTask {
+  private static final Logger LOGGER = LogManager.getLogger(PushStatusDelayedUpdateTask.class);
+  private Version version;
+  private boolean batchPushStartSignalSent;
+  private boolean batchPushEndSignalSent;
+  private final long daVinciWriteCompleteEventDelayInMs;
+  private final Map<Integer, ExecutionStatus> partitionStatus = new VeniceConcurrentHashMap<>();
+  // Executor for scheduling tasks
+  private final ScheduledExecutorService scheduler = Executors.newScheduledThreadPool(1);
```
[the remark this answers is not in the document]
huangminchn: Good point. Will do.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/notifier/PushStatusDelayedUpdateTask.java
gaojieliu: I see.
I think adding `synchronized` is better as this thread pool size can be changed in the future.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/notifier/PushStatusDelayedUpdateTask.java
gaojieliu: Sure, we can do that when supporting inc push status batching.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/notifier/PushStatusDelayedUpdateTask.java
huangminchn: We probably don't need that anymore after the new change; please double check.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/VersionBackend.java
```diff
@@ -98,6 +102,13 @@ public class VersionBackend {
     this.compressor = Lazy.of(
         () -> backend.getCompressorFactory().getCompressor(version.getCompressionStrategy(), version.kafkaTopicName()));
     backend.getVersionByTopicMap().put(version.kafkaTopicName(), this);
+    long daVinciPushStatusCheckIntervalInMs = this.config.getClusterProperties()
```
gaojieliu, line 105: So we don't want to use a `getter` here?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/notifier/DaVinciPushStatusUpdateTask.java
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  /**
+   * Get the partition id set that is being tracked
+   */
+  public Set<Integer> getTrackedPartitions() {
+    return partitionStatus.keySet();
+  }
+
+  public boolean areAllPartitionsOnSameTerminalStatus(ExecutionStatus status) {
```
gaojieliu, line 92: For this, do we want to look at the `VersionBackend#partitionFutures` to make sure every subscribed partition has reported the status before marking job is completed?

gaojieliu (commented):
Left two minor comments.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/VersionBackend.java
```diff
@@ -98,6 +102,13 @@ public class VersionBackend {
...
+        .getLong(DAVINCI_PUSH_STATUS_CHECK_INTERVAL_IN_MS, TimeUnit.SECONDS.toMillis(30));
+    this.daVinciPushStatusUpdateTask = new DaVinciPushStatusUpdateTask(
```
gaojieliu, line 107: Shall we only spin up this task when the write batch push status is enabled?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/VersionBackend.java
huangminchn: I didn't realize that config inherits VeniceServerConfig; changed to use `getter` function.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/notifier/DaVinciPushStatusUpdateTask.java
huangminchn: Yeah, updated in the latest diff.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/notifier/DaVinciPushStatusUpdateTask.java
```diff
@@ -0,0 +1,115 @@
...
+          ExecutionStatus.STARTED,
+          getTrackedPartitions());
+      batchPushStartSignalSent();
+    } else if (!isBatchPushEndSignalSent() && areAllPartitionsOnSameTerminalStatus(ExecutionStatus.COMPLETED)
+        && areAllPartitionFuturesCompletedSuccessfully.get()) {
+      pushStatusStoreWriter.writeVersionLevelPushStatus(
+          version.getStoreName(),
+          version.getNumber(),
+          ExecutionStatus.COMPLETED,
+          getTrackedPartitions());
```
gaojieliu, line 83: What if some DaVinci partition ingestion failed, so this update task will never send `ERROR` status, which is also a terminal status?
Maybe we should check whether all partition status hit terminal status or not, which can be `COMPLETED` or `ERROR`, and then decide the aggregated status and then send it to the push status system store.

huangminchn: Good catch! Added failure path handling.

PR #1058, opened 2024-07-11
[router][common][fast-client] Add RetryManager to venice-router

xunyin8:
## [router][common][fast-client] Add RetryManager to venice-router for long tail retry
1. Move RetryManager to venice-common so it can be used by both fast-client and venice-router. Refactored the constructor a bit so in Venice router we can initialize instances of RetryManager per store (max 2, single-get and multi-get) without  initializing multiple schedulers. 

2. RetryManager for fast-client has store level resource isolation because a fast-client instance is mapped to a single Venice store. RetryManager for venice-router in this change is also made to have store level resource isolation.

3. Updated the fast-client RetryManagerStats with store name in metric prefix to avoid conflict/confusion since a user host could have multiple instances of fast-client for multiple stores. 

## How was this PR tested?
Unit tests and integration tests

### DIFF
--- a/services/venice-router/src/main/java/com/linkedin/venice/router/RouterServer.java
+++ b/services/venice-router/src/main/java/com/linkedin/venice/router/RouterServer.java
@@ -39,6 +39,7 @@ import com.linkedin.venice.helix.SafeHelixManager;
 import com.linkedin.venice.helix.ZkRoutersClusterManager;
 import com.linkedin.venice.meta.ReadOnlySchemaRepository;
 import com.linkedin.venice.meta.ReadOnlyStoreRepository;
+import com.linkedin.venice.meta.RetryManager;
 import com.linkedin.venice.pushstatushelper.PushStatusStoreReader;
 import com.linkedin.venice.read.RequestType;
 import com.linkedin.venice.router.api.DictionaryRetrievalService;
@@ -193,6 +194,10 @@ public class RouterServer extends AbstractVeniceService {
   private VeniceJVMStats jvmStats;
 
   private final AggHostHealthStats aggHostHealthStats;
+
+  private RetryManager singleGetRetryManager;
+
+  private RetryManager multiGetRetryManager;
 
   public static void main(String args[]) throws Exception {
     if (args.length != 1) {
@@ -518,13 +523,25 @@ public class RouterServer extends AbstractVeniceService {
         config.getClusterName(),
         compressorFactory,
         metricsRepository);
+    singleGetRetryManager = new RetryManager(
+        metricsRepository,
+        "single-get-long-tail-retry-manager",
+        config.getLongTailRetryBudgetEnforcementWindowInMs(),
+        config.getSingleGetLongTailRetryBudgetPercentDecimal());
+    multiGetRetryManager = new RetryManager(
+        metricsRepository,
+        "multi-get-long-tail-retry-manager",
+        config.getLongTailRetryBudgetEnforcementWindowInMs(),
+        config.getMultiGetLongTailRetryBudgetPercentDecimal());
     VenicePathParser pathParser = new VenicePathParser(
         versionFinder,
         partitionFinder,
         routerStats,
         metadataRepository,
         config,
-        compressorFactory);
+        compressorFactory,
+        singleGetRetryManager,
+        multiGetRetryManager);
 
     MetaDataHandler metaDataHandler = new MetaDataHandler(
         routingDataRepository,
@@ -862,6 +879,12 @@ public class RouterServer extends AbstractVeniceService {
     }
     if (heartbeat != null) {
       heartbeat.stopInner();
+    }
+    if (singleGetRetryManager != null) {
+      singleGetRetryManager.close();
+    }
+    if (multiGetRetryManager != null) {
+      multiGetRetryManager.close();
     }
   }
 
--- a/services/venice-router/src/main/java/com/linkedin/venice/router/RouterServer.java
+++ b/services/venice-router/src/main/java/com/linkedin/venice/router/RouterServer.java
@@ -39,7 +39,6 @@ import com.linkedin.venice.helix.SafeHelixManager;
 import com.linkedin.venice.helix.ZkRoutersClusterManager;
 import com.linkedin.venice.meta.ReadOnlySchemaRepository;
 import com.linkedin.venice.meta.ReadOnlyStoreRepository;
-import com.linkedin.venice.meta.RetryManager;
 import com.linkedin.venice.pushstatushelper.PushStatusStoreReader;
 import com.linkedin.venice.read.RequestType;
 import com.linkedin.venice.router.api.DictionaryRetrievalService;
@@ -112,6 +111,8 @@ import java.util.Optional;
 import java.util.TreeMap;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Consumer;
@@ -195,9 +196,7 @@ public class RouterServer extends AbstractVeniceService {
 
   private final AggHostHealthStats aggHostHealthStats;
 
-  private RetryManager singleGetRetryManager;
-
-  private RetryManager multiGetRetryManager;
+  private ScheduledExecutorService retryManagerExecutorService;
 
   public static void main(String args[]) throws Exception {
     if (args.length != 1) {
@@ -523,16 +522,9 @@ public class RouterServer extends AbstractVeniceService {
         config.getClusterName(),
         compressorFactory,
         metricsRepository);
-    singleGetRetryManager = new RetryManager(
-        metricsRepository,
-        "single-get-long-tail-retry-manager",
-        config.getLongTailRetryBudgetEnforcementWindowInMs(),
-        config.getSingleGetLongTailRetryBudgetPercentDecimal());
-    multiGetRetryManager = new RetryManager(
-        metricsRepository,
-        "multi-get-long-tail-retry-manager",
-        config.getLongTailRetryBudgetEnforcementWindowInMs(),
-        config.getMultiGetLongTailRetryBudgetPercentDecimal());
+
+    retryManagerExecutorService = Executors.newScheduledThreadPool(3);
+
     VenicePathParser pathParser = new VenicePathParser(
         versionFinder,
         partitionFinder,
@@ -540,8 +532,8 @@ public class RouterServer extends AbstractVeniceService {
         metadataRepository,
         config,
         compressorFactory,
-        singleGetRetryManager,
-        multiGetRetryManager);
+        metricsRepository,
+        retryManagerExecutorService);
 
     MetaDataHandler metaDataHandler = new MetaDataHandler(
         routingDataRepository,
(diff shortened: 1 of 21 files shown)


majisourav99 (commented):
Can we add some tests in router to validate the feature?

xunyin8:
> Can we add some tests in router to validate the feature?

Yes, that's WIP since will do a bit of refactor to achieve store level resource isolation for RetryManager.

xunyin8:
> > Can we add some tests in router to validate the feature?
> 
> Yes, that's WIP since will do a bit of refactor to achieve store level resource isolation for RetryManager.

Done, ready for review.

services/venice-router/src/main/java/com/linkedin/venice/router/VeniceRouterConfig.java
```diff
@@ -393,6 +400,12 @@ private void checkProperties(VeniceProperties props) {
     perStoreRouterQuotaBuffer = props.getDouble(ROUTER_PER_STORE_ROUTER_QUOTA_BUFFER, 1.5);
     httpClientOpensslEnabled = props.getBoolean(ROUTER_HTTP_CLIENT_OPENSSL_ENABLED, true);
     identityParserClassName = props.getString(IDENTITY_PARSER_CLASS, DefaultIdentityParser.class.getName());
+    singleGetLongTailRetryBudgetPercentDecimal =
+        props.getDouble(ROUTER_SINGLE_GET_LONG_TAIL_RETRY_BUDGET_PERCENT_DECIMAL, 0.1);
```
majisourav99, line 404: Are these the default for fast client ? Do you think we should have higher limit for router ? Is it possible to make it dependent on actual QPS being used, like for 10% for 1M qps, but 50% for 1k QPS ?

xunyin8: 10% is the default for fast client. What is the logic behind a higher retry budget for lower KPS especially with store level isolation? There might be un-utilized resource when overall KPS is low but in such scenario we should expect less retries. Based on today's metrics the amount of long tail retry is relatively small in a healthy cluster.

services/venice-router/src/main/java/com/linkedin/venice/router/api/path/VeniceMultiKeyPath.java
```diff
@@ -230,14 +240,19 @@ public int getLongTailRetryMaxRouteForMultiKeyReq() {
...
+    // 1. retry attempts < longTailRetryMaxRouteForMultiKeyReq (if set).
+    // 2. we still have retry budget according to retry manager.
+    boolean longTailRetryAllowed;
     if (longTailRetryMaxRouteForMultiKeyReq == -1) {
       // feature is disabled
-      return true;
-    }
-    if (longTailRetryMaxRouteForMultiKeyReq <= 0) {
-      return false;
+      longTailRetryAllowed = true;
```
majisourav99, line 249: nit: can we just `return retryManager.isRetryAllowed();` here
same for next if, `return false` ..

services/venice-router/src/main/java/com/linkedin/venice/router/VeniceRouterConfig.java
majisourav99: I was thinking compared to overall KPS in the router. eg ie for stores with 1k KPS and 1M qps, both should not be doing same 10% retries ?

xunyin8: I'm still not sure if I fully understand the concern here. The 10% retry quota is the default limit of how much we are willing to perform long tail retry based on the user's long tail retry threshold config. It has nothing to do how much we plan/project to perform retry because we don't have full control of that. If this 10% quota is exhausted either:
1. User's long tail retry threshold is too aggressive (they can set it to 0ms and it becomes speculative query).
2. We have issues in our servers (due to regression or under provisioned).

Setting different quota for different workload without sufficient supporting data will just complicate the behavior and potentially mask the two signals above.

services/venice-router/src/main/java/com/linkedin/venice/router/api/path/VeniceMultiKeyPath.java
majisourav99: No, rather removing `longTailRetryAllowed`

xunyin8: Oh I see, my bad. Done.

services/venice-router/src/main/java/com/linkedin/venice/router/api/path/VeniceMultiKeyPath.java
```diff
@@ -230,14 +240,18 @@ public int getLongTailRetryMaxRouteForMultiKeyReq() {
 
   @Override
   public boolean isLongTailRetryAllowedForNewRoute() {
+    // We are keeping both features for now. Retry is only allowed if both conditions pass:
+    // 1. retry attempts < longTailRetryMaxRouteForMultiKeyReq (if set).
+    // 2. we still have retry budget according to retry manager.
     if (longTailRetryMaxRouteForMultiKeyReq == -1) {
       // feature is disabled
```
majisourav99, line 247: nit: update or remove the stale comment?

services/venice-router/src/main/java/com/linkedin/venice/router/VeniceRouterConfig.java
majisourav99: Discussed offline with Xun, we could have the current setup for now and see if in future we would need per store configs to control different retry behavior.

majisourav99 (commented):
Thanks Xun! BTW do we need to start with this feature disabled and then slowly ramp ? with the default config 0.1, looks like this feature is enabled.

services/venice-router/src/main/java/com/linkedin/venice/router/api/path/VeniceMultiKeyPath.java
```diff
@@ -230,14 +240,18 @@ public int getLongTailRetryMaxRouteForMultiKeyReq() {
...
-      return true;
-    }
-    if (longTailRetryMaxRouteForMultiKeyReq <= 0) {
+      return retryManager.isRetryAllowed();
+    } else if (longTailRetryMaxRouteForMultiKeyReq <= 0) {
       return false;
+    } else {
+      return (currentAllowedRetryRouteCnt.incrementAndGet() <= longTailRetryMaxRouteForMultiKeyReq)
+          && retryManager.isRetryAllowed();
```
gaojieliu, line 253: So the retry manager is enforcing in Route, which would result in some confusion:
1. Router is counting the original request.
2. Retry is being counted per route.
In theory, that is not fair or may not work as expected.
I think we need to align these two measurements, and ideally #1 should count routes of the original request instead.
The total number of requests is critical when evaluating retry storm.

gaojieliu (changes requested):

xunyin8:
> Thanks Xun! BTW do we need to start with this feature disabled and then slowly ramp ? with the default config 0.1, looks like this feature is enabled.

Good call, let me set it to 0 (disabled) by default and we can ramp it slowly and adjust the value if needed in case SLA is at risk.

services/venice-router/src/main/java/com/linkedin/venice/router/RouterServer.java
```diff
@@ -518,13 +525,19 @@ public boolean startInner() throws Exception {
         config.getClusterName(),
         compressorFactory,
         metricsRepository);
+
+    retryManagerExecutorService =
```
gaojieliu, line 529: Quick question:
how do we pick up the thread number: `3` here?
Does it need to be tunable?

services/venice-router/src/main/java/com/linkedin/venice/router/api/VenicePathParser.java
```diff
@@ -160,8 +196,17 @@ public VenicePath parseResourceUri(String uri, HTTP_REQUEST request) throws Rout
             pathHelper.getKey(),
             uri,
             partitionFinder,
-            routerStats);
+            routerStats,
+            singleGetRetryManager);
       } else if (VeniceRouterUtils.isHttpPost(method)) {
+        RetryManager multiGetRetryManager = routerMultiGetRetryManagers.computeIfAbsent(
```
gaojieliu, line 202: It seems we would use the same retry manager for both compute and multi-get requests, right?
That would be fine, but we need to call that out explicitly, and rename the variable to be `multiKeyRetryManager`?
If you agree, we need to change all the related occurrences.

services/venice-router/src/main/java/com/linkedin/venice/router/VeniceRouterConfig.java
```diff
@@ -393,6 +400,12 @@ private void checkProperties(VeniceProperties props) {
     perStoreRouterQuotaBuffer = props.getDouble(ROUTER_PER_STORE_ROUTER_QUOTA_BUFFER, 1.5);
     httpClientOpensslEnabled = props.getBoolean(ROUTER_HTTP_CLIENT_OPENSSL_ENABLED, true);
     identityParserClassName = props.getString(IDENTITY_PARSER_CLASS, DefaultIdentityParser.class.getName());
+    singleGetLongTailRetryBudgetPercentDecimal =
+        props.getDouble(ROUTER_SINGLE_GET_LONG_TAIL_RETRY_BUDGET_PERCENT_DECIMAL, 0.0);
```
gaojieliu, line 404: If the default value is `0`, it means this feature is disabled or all the retry will be rejected?

services/venice-router/src/main/java/com/linkedin/venice/router/api/path/VeniceMultiKeyPath.java
```diff
@@ -230,14 +240,18 @@ public int getLongTailRetryMaxRouteForMultiKeyReq() {
...
-      // feature is disabled
-      return true;
-    }
-    if (longTailRetryMaxRouteForMultiKeyReq <= 0) {
+      // max route feature is disabled
+      return retryManager.isRetryAllowed();
+    } else if (longTailRetryMaxRouteForMultiKeyReq <= 0) {
       return false;
+    } else {
+      return (currentAllowedRetryRouteCnt.incrementAndGet() <= longTailRetryMaxRouteForMultiKeyReq)
```
gaojieliu, line 252: This is slightly different from my understanding..

Let us say the original request contains 1000 keys and it is equally divided into 10 routes, so each route would take 100 keys.
If Router retries one slow route, which contains 100 keys, based on the current mechanism, the retry will be counted as `1`. If the route retry fanout size is 10, I think we should count it as `10` instead of `1`, as the scattered requests are the ones being sent to the storage node, and retry storm is mainly about the small requests sent to the storage nodes.

And this logic can be put in `VeniceDelegateMode`, where we would know the fanout size for both original request and the retry request for each Route.
Let me know if this is not clear.

gaojieliu (commented):
Left a few more comments/questions.

services/venice-router/src/main/java/com/linkedin/venice/router/api/VenicePathParser.java
xunyin8: Right, that's the intention for FC too once we start supporting compute with multiple keys. Will make that change.

services/venice-router/src/main/java/com/linkedin/venice/router/VeniceRouterConfig.java
xunyin8: Default value of 0 will disable the RetryManger
```
    if (enforcementWindowInMs <= 0 || retryBudgetInPercentDecimal <= 0) {
      retryBudgetEnabled.set(false);
    }
```

services/venice-router/src/main/java/com/linkedin/venice/router/RouterServer.java
xunyin8: 3 is arbitrary, will make a config for this.

gaojieliu (approved):
Thanks a lot for adding this to Router!

gaojieliu (approved):

PR #1060, opened 2024-07-12
remove unnecessary blob transfer feature flags from tests

ishwarya-citro:
## Removing unnecessary feature flag from test case

Some flags (DAVINCI_P2P_BLOB_TRANSFER_PORT) were added to this [PR](https://github.com/linkedin/venice/pull/1046). However, they aren't needed.

## How was this PR tested?

Ran the integration tests


adamxchen (approved):
Thanks a lot!

PR #1063, opened 2024-07-15
Revert "[server][da-vinci] Bumped RocksDB dep and adopt multiget async io by default (#950)

gaojieliu:

This reverts commit 02591a119534377cbd1d993e8276d3548d6d3669.

One important partner discovers some TTL DB issue with this version and that particular feature is not being used by Venice.
In the future, we will roll-forward to pick up RocksDB-9.x, which has the fix and add back this feature during the 9.x version upgrade.

## How was this PR tested?
CI


sixpluszero (approved):
LGTM as to unblock other progress.

PR #1064, opened 2024-07-15
[admin-tool] Make end-date arg optional in query-kafka-topic

nisargthakkar:
## Make `--end-date` arg optional in `query-kafka-topic` in Admin tool

This PR makes the `--end-date` arg optional, and if it is not specified, or if the value is later than the current time, then the end offset of the topic is used to get the offset.

## How was this PR tested?

Tested manually


ZacAttack (approved):
Looks good, thanks!

PR #1065, opened 2024-07-15
[controller] Avoid NPE in DIV check during kill-job

nisargthakkar:
## Avoid NPE in DIV check during kill-job

Today, we see NPE during "kill-job" while checking for fatal DIV errors. This commit explicitly checks for a `null` and avoids the NPE

## How was this PR tested?

GHCI

### DIFF
--- a/internal/venice-common/src/main/java/com/linkedin/venice/pushmonitor/PartitionStatus.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/pushmonitor/PartitionStatus.java
@@ -89,7 +89,7 @@ public class PartitionStatus implements Comparable<PartitionStatus> {
 
   public boolean hasFatalDataValidationError() {
     for (ReplicaStatus replicaStatus: replicaStatusMap.values()) {
-      if (ExecutionStatus.isError(replicaStatus.getCurrentStatus())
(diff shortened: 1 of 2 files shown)


sixpluszero (approved):

internal/venice-common/src/main/java/com/linkedin/venice/pushmonitor/PartitionStatus.java
```diff
@@ -89,7 +89,7 @@ public List<StatusSnapshot> getReplicaHistoricStatusList(String instanceId) {
 
   public boolean hasFatalDataValidationError() {
     for (ReplicaStatus replicaStatus: replicaStatusMap.values()) {
-      if (ExecutionStatus.isError(replicaStatus.getCurrentStatus())
+      if (ExecutionStatus.isError(replicaStatus.getCurrentStatus()) && replicaStatus.getIncrementalPushVersion() != null
```
FelixGV, line 92: Thanks for the catch. Is it unit testable?

nisargthakkar: Yeah. It's easy. Added

majisourav99: nit: can you use StringUtils::isEmpty instead of null check?

sixpluszero (approved):

PR #1066, opened 2024-07-15
[controller] Check deferred version swap flag before lingering push

majisourav99:
## Check deferred version swap flag before checking lingering push

If users set deferred version swap for a push, they could decide to swap the current version at much later time than 24hr of push timeout. Starting next push after 24hr would cause the previously pushed version to be killed even though the push has finished successfully. This PR fixes that by failing new pushes till the previously pushed version is swapped to current version.

## How was this PR tested?

integ test. 

### DIFF
--- a/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
@@ -1505,10 +1505,13 @@ public class VeniceParentHelixAdmin implements Admin {
             pushJobId,
             storeName);
       } else {
+        String msg = version.isVersionSwapDeferred()
+            ? ". There is already a future version " + version.getNumber() + " exists for the store " + storeName
+                + " please make that version current before starting a next push."
+            : ". An ongoing push with pushJobId " + existingPushJobId + " and topic " + currentPushTopic.get()
+                + " is found and it must be terminated before another push can be started.";
         VeniceException e = new ConcurrentBatchPushException(
-            "Unable to start the push with pushJobId " + pushJobId + " for store " + storeName
-                + ". An ongoing push with pushJobId " + existingPushJobId + " and topic " + currentPushTopic.get()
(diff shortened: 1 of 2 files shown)


services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
```diff
@@ -1460,7 +1460,8 @@ public Version incrementVersionIdempotent(
       boolean isExistingPushJobARepush = Version.isPushIdRePush(existingPushJobId);
       boolean isIncomingPushJobARepush = Version.isPushIdRePush(pushJobId);
 
-      if (getLingeringStoreVersionChecker()
+      // If version swap is enabled, do not check for lingering push as user may swap at much later time.
+      if (!version.isVersionSwapDeferred() && getLingeringStoreVersionChecker()
```
xunyin8, line 1464: IIUC the expected behavior after the change is that:
1. Old push will be prevented from getting killed automatically since it will not be picked up by the lingering store version checker.
2. User will get the concurrent push error message and needs to either complete the swap of previous version or kill the previous push to proceed.

Questions:
After the change it still throws a generic concurrent push error message. Do users have enough information to take the right action? Should we add a bit more code to differentiate scenarios for our users? The parent controller should have enough information to determine if the previous push is waiting for version swap or not, right?

majisourav99: Right, your understanding is correct. I could update the message in concurrent exception to indicate what exactly is going on. I have added the following log in parent controller, but that was not useful as vpj wont be showing it. 
`"There is already future version {} exists for store {}, please wait till the future version is made current."`

xunyin8: I see, at least it will leave some trace in logs and help the operator figure out what happened.

xunyin8 (approved):

PR #1068, opened 2024-07-17
[router] Fix MultiSchemaResponse fetch in metadata handler

majisourav99:
## Fix MultiSchemaResponse fetch in metadata handler

In `handleValueSchemaLookup` it allocated the schema response into a array of size number of schemas and sets them using the schema id as index into the array.  But if there are schemas which are deleted then those insertions would throw `java.lang.ArrayIndexOutOfBoundsException`. This PR fixes that.
 
## How was this PR tested?

unit test


gaojieliu (approved):
Thanks for the fix.

PR #1069, opened 2024-07-17
[controller] Helix resources should be removed after a VPJ kill job

ishwarya-citro:
When VPJ initiates the kill-job command, the kill-job commands didn't remove the Helix resource. Added logic to include the helix resource deletion whenever a VPJ fails.

Everytime a VPJ runs the `killOfflinePush` job, it does not remove the helix resources. As part of the job, it calls the killOfflinePush but does not call the method to remove the helix resource. 

Changes follow an example of it doing the killOfflinePush & removing helix resource correctly like [here](https://github.com/linkedin/venice/blob/main/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java#L3276) and [here](https://github.com/linkedin/venice/blob/main/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java#L3274). 

## How was this PR tested?
unit tests 

### DIFF
--- a/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
@@ -6544,6 +6544,11 @@ public class VeniceHelixAdmin implements Admin, StoreCleaner {
       return;
     }
 
+    if (!isSafeToKillOfflinePush(clusterName, kafkaTopic, isForcedKill)) {
+      LOGGER.info("Aborting killOfflinePush for: {} in cluster: {}", kafkaTopic, clusterName);
+      return;
+    }
+
     deleteHelixResource(clusterName, kafkaTopic);
 
     LOGGER.info("Killing offline push for: {} in cluster: {}", kafkaTopic, clusterName);
@@ -6611,7 +6616,7 @@ public class VeniceHelixAdmin implements Admin, StoreCleaner {
     }
 
     StatusMessageChannel messageChannel = resources.getMessageChannel();
-    // As we should already have retry outside of this function call, so we do not need to retry again inside.
+    // As we should already have retry outside this function call, so we do not need to retry again inside.
     int retryCount = 1;
     // Broadcast kill message to all of storage nodes assigned to given resource. Helix will help us to only send
     // message to the live instances.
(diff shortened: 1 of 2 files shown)


nisargthakkar:
Hi @ishwarya-citro. We should not include non-public information in open-source PRs. Can you remove the ticket reference?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -6526,6 +6524,11 @@ public void killOfflinePush(String clusterName, String kafkaTopic, boolean isFor
       LOGGER.info("Resource: {} doesn't exist, kill job will be skipped", kafkaTopic);
       return;
     }
+
+    deleteHelixResource(clusterName, kafkaTopic);
```
nisargthakkar, line 6528: This seems like a harmful change because it deletes helix resources without validating if it's not the current version, etc. Can you share some more details about the issue that was noticed?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -3269,9 +3269,6 @@ private void deleteOneStoreVersion(String clusterName, String storeName, int ver
         return;
       }
       String resourceName = Version.composeKafkaTopic(storeName, versionNumber);
-      LOGGER.info("Deleting helix resource: {} in cluster: {}", resourceName, clusterName);
```
majisourav99, line 3272: We should not remove this as kill pushjob could be called without calling delete store version.

majisourav99 (commented):
Please do not add internal jira links in open source PR.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
ishwarya-citro: Part of the `killOfflinePush` will delete the helix resourced. So, this is not removed but moved

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
ishwarya-citro: Everytime a VPJ runs the `killOfflinePush` job, it does not remove the helix resources. As part of the job, it calls the killOfflinePush but does not call the method to remove the helix resource. 

Example of it doing the killOfflinePush & removing helix resource correctly : https://github.com/linkedin/venice/blob/main/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java#L3276

https://github.com/linkedin/venice/blob/main/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java#L3274

I see that we do have version validates & i will add this to the deleteHelixResource()

majisourav99: You need to query the child controllers to validate none of them are using the killjob topic as the current version.
Also I am thinking what is the harm in keeping the helix resources intact till the time when the store versions are deleted (which happens later in some background thread).

ishwarya-citro: @majisourav99  made the changes requested. Pl lmk if this looks ok. 

also, in another [location](https://github.com/linkedin/venice/pull/1069/files#diff-36114654b45af0adf03521759f0301f267c4b453321c47a4d1c1c4726abddca2L3273), it doesn't keep helix resource intact until store versions are deleted. Is it ok to follow the same here?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -6610,6 +6637,47 @@ public void killOfflinePush(String clusterName, String kafkaTopic, boolean isFor
...
+   * If the child controllers are using the killJob topic as their current version, the killOfflinePush job should be aborted
+   */
+  public boolean isSafeToKillOfflinePush(String clusterName, String kafkaTopic, boolean isForcedKill) {
+    if (isForcedKill) {
+      return true;
+    }
+
+    String storeName = Version.parseStoreFromKafkaTopicName(kafkaTopic);
+    int versionNumber = Version.parseVersionFromKafkaTopicName(kafkaTopic);
+    Map<String, ControllerClient> controllerClients = getControllerClientMap(clusterName);
```
majisourav99, line 6650: Move this to parent helix controller and skill checking the current version in parent as parent controller store current version is not properly maintained.

kvargha:
Closing PR as Helix resources do get cleaned up when receiving a kill job message, which was observed during my testing. The child controllers invoke `VeniceHelixAdmin#deleteOneStoreVersion` which calls `VeniceHelixAdmin#deleteHelixResource`

PR #1070, opened 2024-07-17
[controller] Merge VeniceControllerConfig and VeniceControllerClusterConfig

nisargthakkar:
## Merge `VeniceControllerConfig` and `VeniceControllerClusterConfig`

Today, in controllers, we have two classes of configs which serve the same purpose, and it is unnecessary to maintain both. This commit merges `VeniceControllerConfig` and `VeniceControllerClusterConfig`.

## How was this PR tested?

GH CI


services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerMultiClusterConfig.java
sushantmane, line 1: Maybe out of scope of this PR but I'm wondering if you would be able to address the following issue:

In current setup, to get common configs we pick random cluster's controller configs. I wonder if we can move away from this and have a dedicated section for it or use some fixed order so that common configs are predictable?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerConfig.java
```diff
@@ -107,52 +170,72 @@
...
 
 
 /**
  * Configuration which is specific to a Venice controller.
- *
- * It's quite confusing to have both {@link VeniceControllerConfig} and
- * {@link VeniceControllerClusterConfig}. TODO: remove one of them
  */
-public class VeniceControllerConfig extends VeniceControllerClusterConfig {
+public class VeniceControllerConfig {
```
sushantmane, line 226: Can we retain cluster in the name since configs can vary depending on the cluster? Maybe `VeniceClusterControllerConfig`?

sushantmane (commented):
Overall LGTM. Left few comments.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerConfig.java
nisargthakkar: Sure, just the number of files that change blows up a lot 😅

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerMultiClusterConfig.java
nisargthakkar: Yes. I have been thinking of this for a while, but the overhead is that we need to go and setup the configs first, and it is tricky to know if we got it right or not. So, I'd leave that for another day

sushantmane: haha I agree 😄

sushantmane (approved):
LGTM! Thanks, @nisargthakkar!

PR #1071, opened 2024-07-18
[venice-thin-client] Allow refresh while fetching value from router

majisourav99:
## Allow refresh while fetching value from router

When fetching schemas from routers in a thin client, any transient issues can cause requests to fail, leading the client to save a null schema for the schema ID, which results in further request failures. This PR addresses the issue by forcing a refresh during schema fetches and increasing both the retry count and delay for router requests.

## How was this PR tested?

unit test

### DIFF
--- a/clients/venice-thin-client/src/main/java/com/linkedin/venice/client/schema/RouterBackedSchemaReader.java
+++ b/clients/venice-thin-client/src/main/java/com/linkedin/venice/client/schema/RouterBackedSchemaReader.java
@@ -187,7 +187,9 @@ public class RouterBackedSchemaReader implements SchemaReader {
 
   @Override
   public Schema getValueSchema(int id) {
-    SchemaEntry valueSchemaEntry = maybeFetchValueSchemaEntryById(id, false);
+    // Should call with refresh as any router transient error can lead to null schema stored in the map
+    // `valueSchemaEntryMap`
+    SchemaEntry valueSchemaEntry = maybeFetchValueSchemaEntryById(id, true);
     if (!isValidSchemaEntry(valueSchemaEntry)) {
       LOGGER.warn("Got null value schema from Venice for store: {} and id: {}", storeName, id);
       return null;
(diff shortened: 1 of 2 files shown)


sixpluszero (dismissed):
Thanks for the fix, I think it is an important one

clients/venice-thin-client/src/test/java/com/linkedin/venice/client/schema/RouterBackedSchemaReaderTest.java
```diff
@@ -161,14 +161,14 @@ public void testGetValueSchemaWhenNotExists()
       Mockito.verify(mockClient, Mockito.timeout(TIMEOUT).times(2)).getRaw(Mockito.anyString());
       Schema cachedSchema = schemaReader.getValueSchema(3);
       Assert.assertNull(cachedSchema);
-      Mockito.verify(mockClient, Mockito.timeout(TIMEOUT).times(2)).getRaw(Mockito.anyString());
+      Mockito.verify(mockClient, Mockito.timeout(TIMEOUT).times(3)).getRaw(Mockito.anyString());
```
sixpluszero, line 164: wait but why this adds one more count for each of this call?

majisourav99: For missing schema, now it does cache the null, so it will refresh again.

sixpluszero: Oh I see. Thanks!

clients/venice-thin-client/src/main/java/com/linkedin/venice/client/schema/RouterBackedSchemaReader.java
```diff
@@ -475,6 +477,7 @@ private SchemaEntry maybeFetchValueSchemaEntryById(int valueSchemaId, boolean fo
         return NOT_EXIST_VALUE_SCHEMA_ENTRY;
       } else {
         valueSchemaEntryMap.put(valueSchemaId, entry);
+        shouldRefreshLatestValueSchemaEntry.compareAndSet(false, true);
```
sixpluszero, line 480: I am struggling to think if this is not necessary for now. But right now I think adding this will not introduce harm as even if forceRefresh is true and if it is intermediate error and this will only add one time latest schema query.

sixpluszero (approved):
Thanks for adding the fix!

sixpluszero (approved):

PR #1072, opened 2024-07-18
 [controller] Remove stale kill messages from destination cluster before migration

sushantmane:
##  Remove stale kill ingestion messages from destination cluster's participant store before starting store migration

A store migration attempts can fail or be aborted, leading to `KILL` messages being added to the participant system store. These messages are not removed automatically, causing subsequent migration attempts to fail during ingestion.

To resolve this, this PR removes the kill message from the participant system store for the store version being migrated.

## How was this PR tested?
CI


haoxu07 (commented):
Thanks for digging into this! One general comment, is it possible to reproduce this issue through `TestStoreMigration`?

sushantmane:
> Thanks for digging into this! One general comment, is it possible to reproduce this issue through `TestStoreMigration`?


I added a test that reproduces this issue: _If we comment out the stale KILL message clearing logic when adding a version during store migration, the test fails; if we uncomment it, the test succeeds._ I’ve also added another test to verify the logic independently of the store migration flow.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -1520,6 +1522,60 @@ public void migrateStore(String srcClusterName, String destClusterName, String s
...
+    }
+    KillPushJob killPushJobMessage = (KillPushJob) value.messageUnion;
+    LOGGER.info(
+        "Deleting kill message for store-version: {} in cluster: {}. Message age: {}",
+        versionTopicName,
+        clusterName,
+        System.currentTimeMillis() - killPushJobMessage.getTimestamp());
+    deleteParticipantStoreKillMessage(clusterName, versionTopicName);
+
+    // wait for kill message to be removed
```
xunyin8, line 1556: Since `clearIngestionKillMessageAndVerify` will be executed in admin topic (AdminExecutionTask -> admin.addVersionAndStartIngestion). Can we rely on the admin retry and perform an immediate check and fail fast with a `VeniceRetriableException` so while we are waiting for the kill message to be deleted we are not blocking the admin topic from executing other admin messages? This is how we wait for topic creation today.

xunyin8 (changes requested):
Thanks Sushant, most of the change looks good just one major concern.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
sushantmane: This means we can end up sending multiple clear/delete kill messages unless we do some bookkeeping. Is that right?

xunyin8: Right, as we discussed offline I don't think bookkeeping is needed here since this is an edge case and a couple of harmless duplicate delete when handling an edge case is not worth the effort to perform bookkeeping.

sushantmane: I've updated PR (we'll spend at most 1min45 secs in clearing KILL message before throwing exception). Could you please take another look? Thanks.

sushantmane:
Thanks for the review, @xunyin8 & @haoxu07!  I've addressed comments. Could you take another look? Thanks!

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -1520,6 +1522,61 @@ public void migrateStore(String srcClusterName, String destClusterName, String s
...
+    RetryUtils.executeWithMaxAttemptAndExponentialBackoff(
+        () -> deleteParticipantStoreKillMessage(clusterName, versionTopicName),
+        1,
+        Duration.ofSeconds(1),
+        Duration.ofSeconds(3),
+        Duration.ofSeconds(30),
+        Collections.singletonList(Exception.class));
+
+    // wait for kill message to be removed
+    RetryUtils.executeWithMaxAttemptAndExponentialBackoff(() -> {
```
xunyin8, line 1561: I understand that the first `RetryUtils. executeWithMaxAttemptAndExponentialBackoff` is useful to reduce the likelihood of race where we try to migrate store again when the previous attempt just failed (kill is written but still on its way to be reflected). However, I think we can remove `executeWithMaxAttemptAndExponentialBackoff` or reduce the attempt & duration for verification of the delete and just rely on admin channel retry.

sushantmane: Since we've MT admin channel and assuming this is not frequent event, should 1m45sec be a concern?

xunyin8: But why wait for up to a minute here when we know we can give up the thread to process other admin messages and come back and check again later. If we are worried that an idle admin channel could end up spamming the topic with lots delete kill messages then we can still keep this but reduce the attempt and duration.

sushantmane: I believe we should give reasonable time to finish the op. We can certainly reduce the total time.

xunyin8 (approved):
LGTM Thanks Sushant!

sushantmane:
Thanks a lot, @xunyin8 for the review and suggestions! 🙏

PR #1073, opened 2024-07-18
[dvc][test] Fix missing parts for blob transfer and add one incomplete integ test

adamxchen:
## Summary

There are some missing pieces for DVC blob transfer and I've fixed all known issues. However, the new integ test is incomplete due to one DVC characteristics. Given we don't have more bandwidth for this project, we will commit now and do hand-over if needed, although I will still attempt to complete it when I have some spare time. 
A list of items included in this PR:
1. Fixed a bug in `DVCBlobFiner` which used the wrong request path
2. Fixed the issue in `TestBlobDiscovery` that no DVC peer was discovered. Turned out we miss a config called `PUSH_STATUS_STORE_ENABLED` otherwise DVC will not report status
3. Used `DVCBlobFiner` in `TestBlobDiscovery` instead of the apache http client and verified it's working as expected.
4. Fixed a bug that a blob-transfer flag in store/version level config wasn't propagated to low-level `VeniceStoreVersionConfig`. It should be propagated otherwise the underlying rocksdb will not create snapshots
5. ~~The BlobTranferManager initialization was not at the right place. It was originally initialized in the `DaVinciBackend` but turned out we allow DVC subscribe to multiple stores and versions so we have to init the manager in `StoreBackend` during startup or in `VersionBackend` at the time of subscribing to a new version that has this flag enabled.~~ Decided to use a host-level config to bootstrap it so there's no need to introduce convoluted logic for it. 
6. Enhance the logics for blob transfer manager to make sure if there's any exceptions such as connection issue, startup issue, timeout issue, they should be swallowed and logged and fall back to the default pubsub consumption w/o interrupting the application.

## How was this PR tested?

A newly added incomplete integ test `testBlobP2PTransferAmongDVC`. I've been using this integ test to verified the workflow multiple times (as that's why I can catch these issues fixed above). The reason why it's incomplete is we need to set up a different JVM for a second DVC clients in order to test it properly and the work for this is TBD. In the same JVM, DVC clients would share the same `DavinciBackend` object by design, which means they can only use one config thus it makes mimicking two DVC clients talking to each other more difficult since they need to use different configs to set up ports differently.

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
@@ -531,7 +531,7 @@ public class DaVinciBackend implements Closeable {
   public synchronized StoreBackend getStoreOrThrow(String storeName) {
     StoreBackend storeBackend = storeByNameMap.get(storeName);
     if (storeBackend == null) {
-      storeBackend = new StoreBackend(this, storeName, clientConfig);
+      storeBackend = new StoreBackend(this, storeName);
       storeByNameMap.put(storeName, storeBackend);
     }
     return storeBackend;
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
@@ -31,10 +31,7 @@ import com.linkedin.davinci.store.AbstractStorageEngine;
 import com.linkedin.davinci.store.cache.backend.ObjectCacheBackend;
 import com.linkedin.davinci.store.cache.backend.ObjectCacheConfig;
 import com.linkedin.venice.blobtransfer.BlobTransferManager;
-import com.linkedin.venice.blobtransfer.DvcBlobFinder;
-import com.linkedin.venice.blobtransfer.NettyP2PBlobTransferManager;
-import com.linkedin.venice.blobtransfer.client.NettyFileTransferClient;
-import com.linkedin.venice.blobtransfer.server.P2PBlobTransferService;
+import com.linkedin.venice.blobtransfer.BlobTransferUtil;
 import com.linkedin.venice.client.schema.StoreSchemaFetcher;
 import com.linkedin.venice.client.store.ClientConfig;
 import com.linkedin.venice.client.store.ClientFactory;
@@ -91,7 +88,6 @@ import org.apache.logging.log4j.Logger;
 
 public class DaVinciBackend implements Closeable {
   private static final Logger LOGGER = LogManager.getLogger(DaVinciBackend.class);
-  private final BlobTransferManager blobTransferManager;
   private final VeniceConfigLoader configLoader;
   private final SubscriptionBasedReadOnlyStoreRepository storeRepository;
   private final ReadOnlySchemaRepository schemaRepository;
@@ -110,6 +106,8 @@ public class DaVinciBackend implements Closeable {
   private IngestionBackend ingestionBackend;
   private final AggVersionedStorageEngineStats aggVersionedStorageEngineStats;
   private final boolean useDaVinciSpecificExecutionStatusForError;
+  private final ClientConfig clientConfig;
+  private BlobTransferManager<Void> blobTransferManager;
 
   public DaVinciBackend(
       ClientConfig clientConfig,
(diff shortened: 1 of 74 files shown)


internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/router/TestBlobDiscovery.java
```diff
@@ -183,46 +164,32 @@ public void setUp() {
     }
   }
 
-  @AfterTest(alwaysRun = true)
+  @AfterTest
   public void tearDown() {
     D2ClientUtils.shutdownClient(daVinciD2);
-    Utils.closeQuietlyWithErrorLogged(multiRegionMultiClusterWrapper);
+    Utils.closeQuietlyWithErrorLogged(multiClusterVenice);
   }
```
nisargthakkar, line 171: Is this change expected? This would leak running services

clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/DefaultIngestionBackend.java
```diff
@@ -164,13 +179,48 @@ public void setStorageEngineReference(
     }
   }
 
+  /**
+   * Create blob transfer manager and attach it to the ingestion backend since the creation of blob transfer manager is
+   * ad-hoc, depending on whether there are stores enabled with blob transfer.
+   * @param blobTransferManager
+   */
+  @Override
+  public void attachBlobTransferManager(BlobTransferManager blobTransferManager) {
```
sixpluszero, line 188: Do you think it would be simpler if we just always create a blob transfer manager to the ingestion backend even if we might not use it? (can still be turned off by a client config override) so that we don't need these logics here and there to do ad-hoc addition?

If you think this is actually more simple or legit, please don't hesitate to share

sixpluszero: I mean I do see you added a few more logics to try to create the manager based on if any of the workload is p2p enabled and such, so I am thinking if just creating it based on a config (default on) and use it when a ingestion task is current and p2p enabled.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
```diff
@@ -110,6 +106,8 @@ public class DaVinciBackend implements Closeable {
   private IngestionBackend ingestionBackend;
   private final AggVersionedStorageEngineStats aggVersionedStorageEngineStats;
   private final boolean useDaVinciSpecificExecutionStatusForError;
+  private final ClientConfig clientConfig;
+  private BlobTransferManager<Void> blobTransferManager;
```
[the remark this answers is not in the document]
adamxchen: Blobs will be file types. Depending on the blob system we use, they have the different type on PUT AP. For instance, it can be nothing for P2P, UUID for some systems, token for another system. That's why I originally wanted to introduce the type to transfer manager.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/DefaultIngestionBackend.java
adamxchen: I think this is a good point actually.. Adding the manager in a Ad-hoc style does introduce some unnecessary complexity. Let me work on this.

adamxchen: I will make it default 'off' for now since we are experimenting it

sixpluszero (approved):
Thanks for the iteration! Overall it looks good to me, left 2 comments to double check and should be easy to tackle.

clients/da-vinci-client/src/test/java/com/linkedin/davinci/ingestion/DefaultIngestionBackendTest.java
```diff
@@ -16,39 +16,66 @@
...
   private BlobTransferManager blobTransferManager;
+  @Mock
   private DefaultIngestionBackend ingestionBackend;
+  @Mock
   private VeniceStoreVersionConfig storeConfig;
+  @Mock
   private AbstractStorageEngine storageEngine;
+  @Mock
   private ReadOnlyStoreRepository metadataRepo;
+  @Mock
```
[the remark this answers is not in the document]
adamxchen:
```
@Mock
VeniceStoreVersionConfig storeConfig;
```
is same as 
`VeniceStoreVersionConfig storeConfig = mock(VeniceStoreVersionConfig.class);`

Just a way to make it less verbose

clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/DefaultIngestionBackend.java
```diff
@@ -84,34 +90,43 @@ public void startConsumption(VeniceStoreVersionConfig storeConfig, int partition
...
+      CompletableFuture<InputStream> p2pFuture =
+          blobTransferManager.get(storeName, versionNumber, partitionId).toCompletableFuture();
+      LOGGER.info(
+          "Bootstrapping from blobs for store {}, version {}, partition {}",
+          storeName,
+          versionNumber,
+          partitionId);
+      return CompletableFuture.runAsync(() -> {
+        try {
+          p2pFuture.get(30, TimeUnit.MINUTES);
```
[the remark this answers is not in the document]
adamxchen: Ack'ed. Yes, ideally we should have a config or figure out a way to configure this automatically like we do for partition.

PR #1075, opened 2024-07-19
[controller] Checking parent controller region state before handling requests for all routes

bonytoni:
## Checking parent controller region state before handling requests for all routes

- Implemented two region states in `ParentControllerRegionState`
  - `ACTIVE` - parent controller serves requests
  - `PASSIVE` - parent controller rejects requests
- `VeniceParentControllerRegionStateHandler` is a wrapper class for `Route` that checks the state of the region of the parent controller before handling requests. This class was used in `AdminSparkServer` to wrap routes in:
  - `AdminCommandExecutionRoutes`
  - `AdminTopicMetadataRoutes`
  - `ClusterDiscovery`
  - `ClusterRoutes`
  - `ControllerRoutes`
  - `CreateStore`
  - `CreateVersion`
  - `DataRecoveryRoutes`
  - `JobRoutes`
  - `MigrationRoutes`
  - `NewClusterBuildOutRoutes`
  - `NodesAndReplicas`
  - `RoutersClusterConfigRoutes`
  - `SchemaRoutes`
  - `SkipAdminRoute`
  - `StoragePersonaRoutes`
  - `StoresRoutes`
  - `VersionRoute`
- Added config key `CONTROLLER_PARENT_REGION_STATE` in `ConfigKeys` to retrieve the config value from the backend

**Next steps:** Implement the parent controller region state config in the backend

## How was this PR tested?

Unit Tests, Integration Tests

### DIFF
--- a/internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
@@ -1067,6 +1067,17 @@ public class ConfigKeys {
   public static final String CONTROLLER_PARENT_MODE = "controller.parent.mode";
 
   /**
+   * This config specifies the state of the region of the parent controller.
+   *
+   * The region can be in one of the following states:
+   * ACTIVE: the parent controller in the region is serving requests.
+   * PASSIVE: the parent controller in the region is rejecting requests.
+   *
+   * By default, this is set to ACTIVE.
+   */
+  public static final String CONTROLLER_PARENT_REGION_STATE = "controller.parent.region.state";
+
+  /**
    * This config is used to control how many errored topics we are going to keep in parent cluster.
    * This is mostly used to investigate the Kafka missing message issue.
    * If the issue gets resolved, we could change this config to be '0'.
--- a/internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/controller/server/TestAdminSparkServer.java
+++ b/internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/controller/server/TestAdminSparkServer.java
@@ -7,7 +7,6 @@ import com.linkedin.venice.ConfigKeys;
 import com.linkedin.venice.LastSucceedExecutionIdResponse;
 import com.linkedin.venice.common.VeniceSystemStoreType;
 import com.linkedin.venice.controller.Admin;
-import com.linkedin.venice.controller.ParentControllerRegionState;
 import com.linkedin.venice.controller.VeniceHelixAdmin;
 import com.linkedin.venice.controllerapi.AdminCommandExecution;
 import com.linkedin.venice.controllerapi.ControllerApiConstants;
@@ -99,7 +98,6 @@ public class TestAdminSparkServer extends AbstractTestAdminSparkServer {
     extraProperties
         .put(ConfigKeys.CONTROLLER_DEFAULT_MAX_RECORD_SIZE_BYTES, Integer.toString(TEST_MAX_RECORD_SIZE_BYTES));
     extraProperties.put(ConfigKeys.MIN_NUMBER_OF_UNUSED_KAFKA_TOPICS_TO_PRESERVE, Integer.toString(1));
-    extraProperties.put(ConfigKeys.CONTROLLER_PARENT_REGION_STATE, ParentControllerRegionState.ACTIVE.name());
     super.setUp(false, Optional.empty(), extraProperties);
   }
 
--- a/internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/controller/server/TestAdminSparkServer.java
+++ b/internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/controller/server/TestAdminSparkServer.java
@@ -7,6 +7,7 @@ import com.linkedin.venice.ConfigKeys;
 import com.linkedin.venice.LastSucceedExecutionIdResponse;
 import com.linkedin.venice.common.VeniceSystemStoreType;
 import com.linkedin.venice.controller.Admin;
+import com.linkedin.venice.controller.ParentControllerRegionState;
 import com.linkedin.venice.controller.VeniceHelixAdmin;
 import com.linkedin.venice.controllerapi.AdminCommandExecution;
 import com.linkedin.venice.controllerapi.ControllerApiConstants;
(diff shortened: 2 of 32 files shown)


services/venice-controller/src/main/java/com/linkedin/venice/controller/Admin.java
```diff
@@ -755,6 +756,12 @@ void updateRoutersClusterConfig(
    */
   boolean isParent();
 
+  /**
+   * Return the state of the parent controller in its fabric
+   * @return {@code ACTIVE} if the parent controller is actively serving requests in its fabric, else {@code PASSIVE}
```
nisargthakkar, line 761: Instead of `{@code ACTIVE}`, can you change it to `{@link ACTIVE}`. Same in other places

internal/venice-common/src/main/java/com/linkedin/venice/ParentControllerState.java
```diff
@@ -0,0 +1,11 @@
+package com.linkedin.venice;
+
+/**
+ * Enum representing the state of a parent controller in a fabric. (i.e., ParentController in prod-lva1 ACTIVE while ParentController in prod-ltx1 PASSIVE)
+ */
+public enum ParentControllerState {
```
nisargthakkar, line 6: I think we should rename this enum to `ParentControllerRegionState`

services/venice-controller/src/main/java/com/linkedin/venice/controller/server/VeniceParentControllerStateHandler.java
```diff
@@ -0,0 +1,44 @@
...
+  public VeniceParentControllerStateHandler(Admin admin, Route route) {
+    this.admin = admin;
+    this.route = route;
+  }
+
+  @Override
+  public Object handle(Request request, Response response) throws Exception {
+    ParentControllerState parentControllerState = admin.getParentControllerState();
+    boolean isParent = admin.isParent();
+    if (parentControllerState != ACTIVE || !isParent) {
```
nisargthakkar, line 34: This check is incorrect for child controllers and would reject all API request there. It should be 
```java
if (isParent && parentControllerState != ACTIVE) {
    ...
}
```

services/venice-controller/src/main/java/com/linkedin/venice/controller/server/VeniceParentControllerStateHandler.java
```diff
@@ -0,0 +1,44 @@
...
+      ControllerResponse responseObject = new ControllerResponse();
+      response.type(HttpConstants.JSON);
+      response.status(HttpStatus.SC_FORBIDDEN);
```
nisargthakkar, line 37: `SC_FORBIDDEN` is probably not the right status code. `SC_BAD_REQUEST` is a bit too generic, but might still be a better option

internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
```diff
@@ -1126,6 +1126,9 @@ private ConfigKeys() {
   /** Whether current controller is parent or not */
   public static final String CONTROLLER_PARENT_MODE = "controller.parent.mode";
 
+  /** Whether the parent controller is actively serving requests in its fabric. By default, this is ACTIVE. */
+  public static final String CONTROLLER_PARENT_STATUS = "controller.parent.status";
```
nisargthakkar, line 1130: Let's keep consistency b/w the config and what is used in variables
```java
String CONTROLLER_PARENT_REGION_STATE = "controller.parent.region.state";
```

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerClusterConfig.java
```diff
@@ -710,6 +713,7 @@ public VeniceControllerClusterConfig(VeniceProperties props) {
     this.topicCreationThrottlingTimeWindowMs =
         props.getLong(TOPIC_CREATION_THROTTLING_TIME_WINDOW_MS, 10 * Time.MS_PER_SECOND);
     this.parent = props.getBoolean(CONTROLLER_PARENT_MODE, false);
+    this.parentControllerState = ParentControllerState.valueOf(props.getString(CONTROLLER_PARENT_STATUS, "ACTIVE"));
```
nisargthakkar, line 716: instead of using `"ACTIVE"` as a string, can we use `ACTIVE.name()`?

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/controller/server/TestBackupControllerResponse.java
```diff
@@ -27,6 +30,8 @@ public class TestBackupControllerResponse {
   @Test
   public void backupControllerThrows421() throws Exception {
     String clusterName = "backupControllerThrows421";
+    Properties extraProperties = new Properties();
+    extraProperties.put(ConfigKeys.CONTROLLER_PARENT_REGION_STATE, ParentControllerRegionState.ACTIVE.name());
```
nisargthakkar, line 34: We shouldn't need this for child controllers. Also, we shouldn't need to update so many tests since we made the default value as `ACTIVE`, right?

nisargthakkar (approved):
LGTM. Thanks Tony!

Left one query. If you think it doesn't apply, feel free to resolve it and merge the PR

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/controller/server/TestBackupControllerResponse.java
bonytoni: This is correct. Removed from pr and test passes without it.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/controller/server/TestAdminSparkServer.java
```diff
@@ -98,6 +99,7 @@ public void setUp() {
     extraProperties
         .put(ConfigKeys.CONTROLLER_DEFAULT_MAX_RECORD_SIZE_BYTES, Integer.toString(TEST_MAX_RECORD_SIZE_BYTES));
     extraProperties.put(ConfigKeys.MIN_NUMBER_OF_UNUSED_KAFKA_TOPICS_TO_PRESERVE, Integer.toString(1));
+    extraProperties.put(ConfigKeys.CONTROLLER_PARENT_REGION_STATE, ParentControllerRegionState.ACTIVE.name());
```
nisargthakkar, line 102: Can you try removing this as well? I also see that the config in `TestBackupControllerResponse.java` is also still there.

nisargthakkar (commented):
Looking quite good. Just the two test-related changes needed

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/controller/server/TestAdminSparkServer.java
bonytoni: Pushed a commit with properties removed from the two test files.

nisargthakkar (approved):
LGTM. Thanks @bonytoni !

PR #1076, opened 2024-07-23
[controller][admin-tool][vpj][test] Add a multi-region config to controllers and cleanup many configs

nisargthakkar:
## Add a `multi.region` config to controllers and cleanup many other configs

1. Use multi.region config to control certain features instead of controlling them via explicit configs.
    * Native-replication - Enabled in multi-region mode. Disabled in single-region mode.
    * Admin channel consumption - Enabled in multi-region mode. Disabled in single-region mode.
        * This is still allowed to be disabled via LiveConfig for store migration purposes.
    * Controller allowing BATCH push via API - Disabled in multi-region mode. Enabled in single-region mode.
    * Active-active replication enabled on a controller - Enabled in multi-region mode. Disabled in single-region mode.
2. Commands and code to enable and disable Native-replication for a cluster has been removed as that is the only mode.
3. The following configs are now obsolete, and removed:
    * enable.native.replication.for.batch.only
    * enable.native.replication.for.hybrid
    * enable.native.replication.as.default.for.batch.only
    * enable.native.replication.as.default.for.hybrid
    * enable.active.active.replication.as.default.for.batch.only.store
    * admin.topic.remote.consumption.enabled
    * child.controller.admin.topic.consumption.enabled
        * Only removed from controller config
        * It is still used in LiveConfigs during store migration
    * active.active.enabled.on.controller
    * controller.enable.batch.push.from.admin.in.child
4. The following configs were unused and now removed:
    * kafka.min.log.compaction.lag.ms
    * topic.cleanup.send.concurrent.delete.requests.enabled
    * topic.deletion.status.poll.interval.ms
    * topic.creation.throttling.time.window.ms
    * topic.manager.kafka.operation.timeout.ms
    * admin.consumption.timeout.minute
    * amplification.factor
    * server.ingestion.isolation.metric.request.timeout.seconds

## How was this PR tested?

GH CI


internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/controller/AbstractTestVeniceHelixAdmin.java
```diff
@@ -263,9 +246,9 @@ VeniceHelixAdmin getFollower(List<VeniceHelixAdmin> admins, String cluster) {
   /**
    * Participant store should be set up by child controller.
    */
-  void verifyParticipantMessageStoreSetup() {
+  private void verifyParticipantMessageStoreSetup() {
     String participantStoreName = VeniceSystemStoreUtils.getParticipantStoreNameForCluster(clusterName);
-    TestUtils.waitForNonDeterministicAssertion(5, TimeUnit.SECONDS, () -> {
+    TestUtils.waitForNonDeterministicAssertion(10, TimeUnit.SECONDS, () -> {
```
mynameborat, line 251: Any reason to not keep it 5? Just wondering if this will increase the tests times and if this necessary

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/controller/TestEnablingSSLPushInVeniceHelixAdminWithIsolatedEnvironment.java
```diff
@@ -23,7 +24,7 @@ public class TestEnablingSSLPushInVeniceHelixAdminWithIsolatedEnvironment extend
 
   @BeforeMethod(alwaysRun = true)
   public void setUp() throws Exception {
-    setupCluster(false);
```
mynameborat, line 26: Is this needed? I don't see the previous version of the constructor deprecated. Why no leave it as is?

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/controller/TestVeniceHelixAdminWithIsolatedEnvironment.java
```diff
@@ -42,7 +42,7 @@
 public class TestVeniceHelixAdminWithIsolatedEnvironment extends AbstractTestVeniceHelixAdmin {
   @BeforeMethod(alwaysRun = true)
   public void setUp() throws Exception {
-    setupCluster(false);
+    setupCluster(false, new MetricsRepository());
```
mynameborat, line 45: same as above.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/controller/TestFabricBuildout.java
```diff
@@ -29,9 +27,17 @@
...
 import org.testng.annotations.Test;
 
 
+/**
+ * Note: These tests use an unsupported mode during setup - by creating stores and running pushes
+ * to them directly in child regions. This is not how new regions will be added. We need the test setup to support
+ * adding blank regions so that we can simulate the true fabric buildout process. Because of this, I've disabled these
+ * tests for now.
+ */
+@Ignore
```
mynameborat, line 40: Can you create a issue in github or JIRA and attach the todo?

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/controller/TestVeniceHelixAdminWithSharedEnvironment.java
```diff
@@ -101,7 +101,6 @@ public class TestVeniceHelixAdminWithSharedEnvironment extends AbstractTestVenic
   @BeforeClass(alwaysRun = true)
   public void setUp() throws Exception {
     setupCluster(true, metricsRepository);
-    verifyParticipantMessageStoreSetup();
```
mynameborat, line 104: Is this intended? can you help me understand why this is not needed anymore?

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/controller/AbstractTestVeniceHelixAdmin.java
nisargthakkar: This is to keep parity with removed code in the `setupCluster`. Although, I think 5 seconds should also be okay. I can try reverting it.

The runtime increase is not significant to lead to a increase of overall test time. Only in failure cases will it go till 10 seconds. If the condition is met earlier than 5 seconds, there will be no impact to the runtime.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/controller/TestEnablingSSLPushInVeniceHelixAdminWithIsolatedEnvironment.java
nisargthakkar: It is removed in `AbstractTestVeniceHelixAdmin`. I was touching that file, and felt it was not worth having 3 constructors as they were quite basic, and were test-only. I can revert it, though if you feel otherwise. I don't have strong opinions for this

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/controller/TestVeniceHelixAdminWithIsolatedEnvironment.java
nisargthakkar: It is removed in `AbstractTestVeniceHelixAdmin`. Again, I don't have any concerns in reverting it. Let me know whatever works.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/controller/TestVeniceHelixAdminWithSharedEnvironment.java
nisargthakkar: Yes. `setupCluster` in `AbstractTestVeniceHelixAdmin` already calls this method internally. So, this was essentially a duplicate check

docker/venice-controller/single-dc-configs/controller.properties
```diff
@@ -22,7 +22,7 @@ enable.hybrid.push.ssl.whitelist=false
 default.partition.size=100
 min.active.replica=1
 kafka.replication.factor=1
-child.cluster.allowlist=false
+child.cluster.allowlist=
```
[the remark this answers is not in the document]
nisargthakkar: This will end up in `Utils.parseCommaSeparatedStringToList` where this is the updated code:
```
if (StringUtils.isEmpty(rawString)) {
  return Collections.emptyList();
}
```

Also, this config is not used in child-controllers. It is only ever used in parent controllers. The reason I had to include this change is that there is a new code that checks if this config is set and infers it as a multi-region mode. Also, the value of `false` is absurd for this field, and hence, also incorrect.

Ideally, this config shouldn't be set, but I still wanted this left behind to reduce the number of changes, and also show that it is actually backwards-compatible if this was configured correctly in the first place. I think backwards-compatibility for misconfigured stuff can be a tricky topic and I think it is okay to break compatibility in such cases.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/controller/TestFabricBuildout.java
nisargthakkar: https://github.com/linkedin/venice/issues/1082

internal/venice-common/src/main/java/com/linkedin/venice/utils/Utils.java
```diff
@@ -461,15 +461,14 @@ public static boolean verifyTransition(ExecutionStatus newStatus, ExecutionStatu
...
 
   public static List<String> parseCommaSeparatedStringToList(String rawString) {
-    String[] strArray = rawString.split(",\\s*");
-    if (strArray.length < 1) {
-      throw new VeniceException("Invalid input: " + rawString);
+    if (StringUtils.isEmpty(rawString)) {
+      return Collections.emptyList();
     }
-    return Arrays.asList(strArray);
+    return Arrays.asList(rawString.split(",\\s*"));
```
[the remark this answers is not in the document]
nisargthakkar: When checking the uses of this function, I realized that it is no longer used. So, I've removed it now. I've replaced it's uses with `parseCommaSeparatedStringToSet`.

However, the comment is still valid as there could be a change in behavior. So, I inspected the uses of this function and have now verified that the "emptiness" of the Set is handled gracefully.

There was another change in behavior with this change for A/A configs. Previously, users would set `active.active.replication.enabled.on.controllers`, and if this is set, then we looked for AA configs like `active.active.real.time.source.fabric.list` and threw an exception. Now, with this change, since `active.active.replication.enabled.on.controllers` is no longer a valid config, I've made the default behavior of non-specified `active.active.real.time.source.fabric.list` to use the value of `native.replication.fabric.allowlist` - and this was indirectly a required config previously since in multi-region mode, NR is the only supported mode now.

docker/venice-controller/single-dc-configs/controller.properties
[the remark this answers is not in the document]
mynameborat: I see.. agree false was anyways a bad default in the first place. Sounds good 👍

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/controller/TestEnablingSSLPushInVeniceHelixAdminWithIsolatedEnvironment.java
mynameborat: in that case, can you also remove the constructor that is not needed.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/controller/TestVeniceHelixAdminWithIsolatedEnvironment.java
mynameborat: Feel free to resolve the similar comment after responding to original thread. Marking this as resolved.

internal/venice-common/src/main/java/com/linkedin/venice/utils/Utils.java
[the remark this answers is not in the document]
mynameborat: Thanks for checking.

mynameborat:
Looks good to me. If you can remove the unused constructor that would be good.

mynameborat (approved):

nisargthakkar:
Thanks a lot @mynameborat

PR #1009, opened 2024-05-30
[server] Fully Assembled Record Size Metric

KaiSernLim:
## Summary

Venice Fully Assembled Record Size Metric

#### Problem

We don't have visibility on how large customers' records can be, which can impact our system. It also helps to potentially identify a hot key and huge records which can affect the server ingestion performance.

This PR adds metrics for total value size for fully assembled records. There is an existing `record_value_size_in_bytes` metric, but that's for each chunk rather than the full size of the record.

#### Changes

1. [Functional] Added a new per-store metric called `assembled_record_value_size_in_bytes` at the drainer stage. This will record the assembled / original size of larger / chunked records by recording the `size` field of the `ChunkedValueManifest`.
    1. The `size` field is sufficient, because the size of the fully assembled record is known before the `VeniceWriter` performs chunking.
    2. The metric is added to the drainer (which is after any partial update reassembling work), so all large records should be covered.
3. [Refactor] Moved `TestChunkingUtils` into the `venice-test-common` project, which is hopefully a more easily accessible test utils location and can be helpful in more tests in the future.
4. [Refactor] Moved several helper functions from `TestKafkaTopicDumper` into `TestChunkingUtils`, in hopes that they can be more easily reused in creating chunked `PubSubMessage` objects.
5. [Clean Up] Added a test util `prependSchemaId()` to help with prepending the schema id into the put value bytes, which is useful in multiple other tests.
6. [Unit Test] Added a unit test with `testAssembledValueSizeSensor()` in `TestStoreIngestionTask` where a chunked record and its manifest are sent to the drainer, and the metrics should be populated.
7. [Integration Test] Added a check with `testLargeValues()` in `TestBatch` that the metric is populated if chunking is enabled when sending large records and that the metric is not populated, otherwise.

## How was this PR tested?

Unit test and integration test.

## Does this PR introduce any user-facing changes?

- [X] No. You can skip the rest of this section.
- [ ] Yes. Make sure to explain your proposed changes and call out the behavior change.

PR #1077, opened 2024-07-23
[server] Ratio Metric for Max Record Size

KaiSernLim:
## Summary

Added a metric to alert customers when they are nearing the record size limit by providing the percentage of the record limit reached.

### Changes
1. **[Metric]** Added the metric `assembled_record_size_ratio` which is a ratio of the record size (for records that need chunking, similar to #1009) divided by the `maxRecordSizeBytes` record size limit.
2. **[Metric]** Replaced the metric `assembled_record_value_size_in_bytes` with `assembled_record_size_in_bytes` to pair with `assembled_record_size_ratio`, so we have both metrics that are measuring record size.
3. **[Functional]** Stopped counting RMD size towards the `maxRecordSizeBytes` limit (added in #1052), because it's an internal component and a user's write operation should not be failed because of it.
4. **[Config]** Renamed the controller config `controller.default.max.record.size.bytes` to just `default.max.record.size.bytes`, which will be shared between server and controller.
5. **[Metric]** Added the metric `assembled_rmd_size_in_bytes` (similar to #1009) to track and monitor the RMD size as it contributes to overall record size.
6. **[Functional]** Added field `isRmdChunked` to `StoreIngestionTask` in order to distinguish between chunked and non-chunked RMD payloads.
7. **[Unit Test]** Extended `testAssembledValueSizeSensor()` in `StoreIngestionTaskTest` to test the values of `assembled_record_size_ratio` and `assembled_rmd_size_in_bytes`.
8. **[Integration Test]** Modified `testStoreWithTooLargeValues()` in `TestBatch` to verify that the metric is being populated with the right values.
9. **[Integration Test]** Added a check that the `assembled_rmd_size_in_bytes` metric is being populated in `testActiveActivePartialUpdateWithCompression()` within `PartialUpdateTest`.
10. **[Test]** Minor test util changes in `MetricsUtils` and `ChunkingTestUtils`.

## How was this PR tested?
GHCI

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/config/VeniceServerConfig.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/config/VeniceServerConfig.java
@@ -10,6 +10,7 @@ import static com.linkedin.venice.ConfigKeys.DAVINCI_P2P_BLOB_TRANSFER_SERVER_PO
 import static com.linkedin.venice.ConfigKeys.DA_VINCI_CURRENT_VERSION_BOOTSTRAPPING_QUOTA_BYTES_PER_SECOND;
 import static com.linkedin.venice.ConfigKeys.DA_VINCI_CURRENT_VERSION_BOOTSTRAPPING_QUOTA_RECORDS_PER_SECOND;
 import static com.linkedin.venice.ConfigKeys.DA_VINCI_CURRENT_VERSION_BOOTSTRAPPING_SPEEDUP_ENABLED;
+import static com.linkedin.venice.ConfigKeys.DEFAULT_MAX_RECORD_SIZE_BYTES;
 import static com.linkedin.venice.ConfigKeys.DIV_PRODUCER_STATE_MAX_AGE_MS;
 import static com.linkedin.venice.ConfigKeys.ENABLE_GRPC_READ_SERVER;
 import static com.linkedin.venice.ConfigKeys.ENABLE_SERVER_ALLOW_LIST;
@@ -140,6 +141,8 @@ import static com.linkedin.venice.ConfigKeys.UNREGISTER_METRIC_FOR_DELETED_STORE
 import static com.linkedin.venice.ConfigKeys.UNSORTED_INPUT_DRAINER_SIZE;
 import static com.linkedin.venice.ConfigKeys.USE_DA_VINCI_SPECIFIC_EXECUTION_STATUS_FOR_ERROR;
 import static com.linkedin.venice.pubsub.PubSubConstants.PUBSUB_TOPIC_MANAGER_METADATA_FETCHER_CONSUMER_POOL_SIZE_DEFAULT_VALUE;
+import static com.linkedin.venice.utils.ByteUtils.BYTES_PER_MB;
+import static com.linkedin.venice.utils.ByteUtils.generateHumanReadableByteCountString;
 
 import com.linkedin.davinci.helix.LeaderFollowerPartitionStateModelFactory;
 import com.linkedin.davinci.kafka.consumer.KafkaConsumerService;
@@ -480,6 +483,7 @@ public class VeniceServerConfig extends VeniceClusterConfig {
   private final long daVinciCurrentVersionBootstrappingQuotaRecordsPerSecond;
   private final long daVinciCurrentVersionBootstrappingQuotaBytesPerSecond;
   private final boolean resubscriptionTriggeredByVersionIngestionContextChangeEnabled;
+  private final int defaultMaxRecordSizeBytes;
 
   public VeniceServerConfig(VeniceProperties serverProperties) throws ConfigurationException {
     this(serverProperties, Collections.emptyMap());
@@ -794,6 +798,12 @@ public class VeniceServerConfig extends VeniceClusterConfig {
         serverProperties.getSizeInBytes(DA_VINCI_CURRENT_VERSION_BOOTSTRAPPING_QUOTA_BYTES_PER_SECOND, -1);
     resubscriptionTriggeredByVersionIngestionContextChangeEnabled =
         serverProperties.getBoolean(SERVER_RESUBSCRIPTION_TRIGGERED_BY_VERSION_INGESTION_CONTEXT_CHANGE_ENABLED, false);
+    defaultMaxRecordSizeBytes = serverProperties.getInt(DEFAULT_MAX_RECORD_SIZE_BYTES, 100 * BYTES_PER_MB);
+    if (defaultMaxRecordSizeBytes < BYTES_PER_MB) {
+      throw new VeniceException(
+          DEFAULT_MAX_RECORD_SIZE_BYTES + ": " + defaultMaxRecordSizeBytes + " must be at least "
+              + generateHumanReadableByteCountString(BYTES_PER_MB));
+    }
   }
 
   long extractIngestionMemoryLimit(
@@ -1402,6 +1412,10 @@ public class VeniceServerConfig extends VeniceClusterConfig {
 
   public String getIdentityParserClassName() {
     return identityParserClassName;
+  }
+
+  public int getDefaultMaxRecordSizeBytes() {
+    return defaultMaxRecordSizeBytes;
   }
 
   public String getRocksDBPath() {
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/config/VeniceServerConfig.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/config/VeniceServerConfig.java
@@ -2,6 +2,7 @@ package com.linkedin.davinci.config;
 
 import static com.linkedin.davinci.ingestion.utils.IsolatedIngestionUtils.INGESTION_ISOLATION_CONFIG_PREFIX;
 import static com.linkedin.davinci.store.rocksdb.RocksDBServerConfig.ROCKSDB_TOTAL_MEMTABLE_USAGE_CAP_IN_BYTES;
+import static com.linkedin.venice.ConfigConstants.DEFAULT_MAX_RECORD_SIZE_BYTES_BACKFILL;
 import static com.linkedin.venice.ConfigKeys.AUTOCREATE_DATA_PATH;
 import static com.linkedin.venice.ConfigKeys.BLOB_TRANSFER_MANAGER_ENABLED;
 import static com.linkedin.venice.ConfigKeys.DATA_BASE_PATH;
@@ -798,7 +799,8 @@ public class VeniceServerConfig extends VeniceClusterConfig {
         serverProperties.getSizeInBytes(DA_VINCI_CURRENT_VERSION_BOOTSTRAPPING_QUOTA_BYTES_PER_SECOND, -1);
     resubscriptionTriggeredByVersionIngestionContextChangeEnabled =
         serverProperties.getBoolean(SERVER_RESUBSCRIPTION_TRIGGERED_BY_VERSION_INGESTION_CONTEXT_CHANGE_ENABLED, false);
-    defaultMaxRecordSizeBytes = serverProperties.getInt(DEFAULT_MAX_RECORD_SIZE_BYTES, 100 * BYTES_PER_MB);
+    defaultMaxRecordSizeBytes =
+        serverProperties.getInt(DEFAULT_MAX_RECORD_SIZE_BYTES, DEFAULT_MAX_RECORD_SIZE_BYTES_BACKFILL);
     if (defaultMaxRecordSizeBytes < BYTES_PER_MB) {
       throw new VeniceException(
           DEFAULT_MAX_RECORD_SIZE_BYTES + ": " + defaultMaxRecordSizeBytes + " must be at least "
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
@@ -287,7 +287,9 @@ public class LeaderFollowerStoreIngestionTask extends StoreIngestionTask {
             // .setMaxRecordSizeBytes(store.getMaxRecordSizeBytes()) // TODO: allow when nearline jobs are enforced
             .build();
     this.veniceWriter = Lazy.of(() -> veniceWriterFactory.createVeniceWriter(writerOptions));
-    this.maxRecordSizeBytes = store.getMaxRecordSizeBytes(); // TODO: move to VeniceWriter when nearline jobs enforce
+    this.maxRecordSizeBytes = (store.getMaxRecordSizeBytes() < 0) // TODO: move to VeniceWriter when nearline supported
+        ? serverConfig.getDefaultMaxRecordSizeBytes()
+        : store.getMaxRecordSizeBytes();
     this.kafkaClusterIdToUrlMap = serverConfig.getKafkaClusterIdToUrlMap();
     this.kafkaDataIntegrityValidatorForLeaders = new KafkaDataIntegrityValidator(kafkaVersionTopic);
     if (builder.getVeniceViewWriterFactory() != null && !store.getViewConfigs().isEmpty()) {
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
@@ -1901,9 +1901,14 @@ public class LeaderFollowerStoreIngestionTask extends StoreIngestionTask {
   }
 
   @Override
-  protected final void recordAssembledRecordSizeRatio(long size, long currentTimeMs) {
+  protected final double calculateAssembledRecordSizeRatio(long recordSize) {
+    return (double) recordSize / maxRecordSizeBytes;
+  }
+
+  @Override
+  protected final void recordAssembledRecordSizeRatio(double ratio, long currentTimeMs) {
     if (maxRecordSizeBytes != VeniceWriter.UNLIMITED_MAX_RECORD_SIZE) {
-      hostLevelIngestionStats.recordAssembledRecordSizeRatio((double) size / maxRecordSizeBytes, currentTimeMs);
+      hostLevelIngestionStats.recordAssembledRecordSizeRatio(ratio, currentTimeMs);
     }
   }
 
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
@@ -284,7 +284,6 @@ public class LeaderFollowerStoreIngestionTask extends StoreIngestionTask {
             .setChunkingEnabled(isChunked)
             .setRmdChunkingEnabled(version.isRmdChunkingEnabled())
             .setPartitionCount(storeVersionPartitionCount)
-            // .setMaxRecordSizeBytes(store.getMaxRecordSizeBytes()) // TODO: allow when nearline jobs are enforced
             .build();
     this.veniceWriter = Lazy.of(() -> veniceWriterFactory.createVeniceWriter(writerOptions));
     this.maxRecordSizeBytes = (store.getMaxRecordSizeBytes() < 0) // TODO: move to VeniceWriter when nearline supported
@@ -1903,9 +1902,9 @@ public class LeaderFollowerStoreIngestionTask extends StoreIngestionTask {
 
   @Override
   protected final void recordAssembledRecordSizeRatio(long size, long currentTimeMs) {
-    final double sizeRatio =
-        (maxRecordSizeBytes == VeniceWriter.UNLIMITED_MAX_RECORD_SIZE) ? 0 : (double) size / maxRecordSizeBytes;
-    hostLevelIngestionStats.recordAssembledRecordSizeRatio(sizeRatio, currentTimeMs);
+    if (maxRecordSizeBytes != VeniceWriter.UNLIMITED_MAX_RECORD_SIZE) {
+      hostLevelIngestionStats.recordAssembledRecordSizeRatio((double) size / maxRecordSizeBytes, currentTimeMs);
+    }
   }
 
   private void recordRegionHybridConsumptionStats(
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
@@ -188,6 +188,8 @@ public class LeaderFollowerStoreIngestionTask extends StoreIngestionTask {
 
   private final AtomicLong lastSendIngestionHeartbeatTimestamp = new AtomicLong(0);
 
+  protected final int maxRecordSizeBytes; // TODO: move into VeniceWriter when nearline jobs enforce max record size
+
   public LeaderFollowerStoreIngestionTask(
       StoreIngestionTaskFactory.Builder builder,
       Store store,
@@ -282,8 +284,10 @@ public class LeaderFollowerStoreIngestionTask extends StoreIngestionTask {
             .setChunkingEnabled(isChunked)
             .setRmdChunkingEnabled(version.isRmdChunkingEnabled())
             .setPartitionCount(storeVersionPartitionCount)
+            // .setMaxRecordSizeBytes(store.getMaxRecordSizeBytes()) // TODO: allow when nearline jobs are enforced
             .build();
     this.veniceWriter = Lazy.of(() -> veniceWriterFactory.createVeniceWriter(writerOptions));
+    this.maxRecordSizeBytes = store.getMaxRecordSizeBytes(); // TODO: move to VeniceWriter when nearline jobs enforce
     this.kafkaClusterIdToUrlMap = serverConfig.getKafkaClusterIdToUrlMap();
     this.kafkaDataIntegrityValidatorForLeaders = new KafkaDataIntegrityValidator(kafkaVersionTopic);
     if (builder.getVeniceViewWriterFactory() != null && !store.getViewConfigs().isEmpty()) {
@@ -1893,6 +1897,13 @@ public class LeaderFollowerStoreIngestionTask extends StoreIngestionTask {
       hostLevelIngestionStats.recordTotalFollowerBytesConsumed(processedRecordSize);
       hostLevelIngestionStats.recordTotalFollowerRecordsConsumed();
     }
+  }
+
+  @Override
+  protected final void recordAssembledRecordSizeRatio(long size, long currentTimeMs) {
+    final double sizeRatio =
+        (maxRecordSizeBytes == VeniceWriter.UNLIMITED_MAX_RECORD_SIZE) ? 0 : (double) size / maxRecordSizeBytes;
+    hostLevelIngestionStats.recordAssembledRecordSizeRatio(sizeRatio, currentTimeMs);
   }
 
   private void recordRegionHybridConsumptionStats(
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
@@ -2429,7 +2429,9 @@ public abstract class StoreIngestionTask implements Runnable, Closeable {
     hostLevelIngestionStats.recordChecksumVerificationFailure();
   }
 
-  protected abstract void recordAssembledRecordSizeRatio(long recordSize, long currentTimeMs);
+  protected abstract void recordAssembledRecordSizeRatio(double ratio, long currentTimeMs);
+
+  protected abstract double calculateAssembledRecordSizeRatio(long recordSize);
 
   public abstract long getBatchReplicationLag();
 
@@ -3243,7 +3245,8 @@ public abstract class StoreIngestionTask implements Runnable, Closeable {
             ByteBuffer rmdPayload = put.getReplicationMetadataPayload();
             boolean isValueManifest = rmdPayload != null && rmdPayload.equals(VeniceWriter.EMPTY_BYTE_BUFFER);
             if (isValueManifest) {
-              recordAssembledRecordSizeRatio(keyLen + chunkedValueManifest.getSize(), currentTimeMs);
+              double ratio = calculateAssembledRecordSizeRatio(keyLen + chunkedValueManifest.getSize());
+              recordAssembledRecordSizeRatio(ratio, currentTimeMs);
               hostLevelIngestionStats.recordAssembledValueSize(chunkedValueManifest.getSize(), currentTimeMs);
             } else { // the manifest pertains to a chunked rmd
               hostLevelIngestionStats.recordAssembledRmdSize(chunkedValueManifest.getSize(), currentTimeMs);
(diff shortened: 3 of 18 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -176,6 +176,9 @@ public abstract class StoreIngestionTask implements Runnable, Closeable {
   private static final ExecutorService SHUTDOWN_EXECUTOR_FOR_DVC =
       Executors.newFixedThreadPool(Runtime.getRuntime().availableProcessors() * 2);
 
+  protected static final ChunkedValueManifestSerializer CHUNKED_VALUE_MANIFEST_SERIALIZER =
+      new ChunkedValueManifestSerializer(true);
```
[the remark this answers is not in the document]
nisargthakkar: Yeah. There would be perf concerns too, I believe

internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
```diff
@@ -2169,7 +2169,13 @@ private ConfigKeys() {
...
-   * Only used in batch push jobs and partial updates.
+   * Only used in batch push jobs.
    */
   public static final String CONTROLLER_DEFAULT_MAX_RECORD_SIZE_BYTES = "controller.default.max.record.size.bytes";
+
+  /**
+   * Server config for the default value of {@link com.linkedin.venice.writer.VeniceWriter#maxRecordSizeBytes}.
+   * Only used in partial updates.
+   */
+  public static final String SERVER_DEFAULT_MAX_RECORD_SIZE_BYTES = "server.default.max.record.size.bytes";
```
nisargthakkar, line 2180: Can we merge both of these into one - `default.max.record.size.bytes`? Is there a case where both would diverge?

I suppose we also need to add some logic to handle record limit in Samza RP too as that is similar to batch push job and can chunk records on client-side ([code](https://github.com/linkedin/venice/blob/main/clients/venice-samza/src/main/java/com/linkedin/venice/samza/VeniceSystemProducer.java#L397-L403))

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -2370,6 +2371,10 @@ public void recordChecksumVerificationFailure() {
     hostLevelIngestionStats.recordChecksumVerificationFailure();
   }
 
+  protected void recordAssembledRecordSizeRatio(long recordSize, long currentTimeMs) {
+    // no op
+  }
```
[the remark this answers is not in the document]
KaiSernLim: Will do.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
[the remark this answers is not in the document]
KaiSernLim: I worry about that too, which is why it wasn't like this in the original PR.

I noticed that `ChunkedValueManifestSerializer#deserialize()` doesn't seem to be writing anything to the instance variables, so I thought it might be safe to initialize the instance once and reuse it many times.

I saw another class that had this serializer as `static` (`LeaderProducerCallback`, albeit I think it's a single-use class?) and I wanted to propose this. Just not sure if it's safe to do so in Java?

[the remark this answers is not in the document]
lluwm: I feel that there could still might be code path to change its internal state, e.g. `cacheDatumReader` will modify `protocolVersionToReader`. If we would like to make this change, maybe in a separate PR so that it would catch people's attention.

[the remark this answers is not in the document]
KaiSernLim: Oh I see, I'll revert it then. I just wanted to propose it in a small PR so it was easier to see.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/HostLevelIngestionStats.java
```diff
@@ -473,6 +471,10 @@ public void recordAssembledValueSize(long bytes, long currentTimeMs) {
     assembledValueSizeSensor.record(bytes, currentTimeMs);
   }
 
+  public void recordAssembledRecordSizeRatio(double ratio, long currentTimeMs) {
+    assembledRecordSizeRatioSensor.record(ratio, currentTimeMs);
```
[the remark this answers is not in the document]
KaiSernLim: I think they're actually the same:
```java
  /**
     * Record a value at a known time. This method is slightly faster than {@link #record(double)} since it will reuse
     * the time stamp.
     */
    public void record(double value, long timeMs) {
```

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/HostLevelIngestionStats.java
```diff
@@ -278,13 +280,9 @@ public HostLevelIngestionStats(
...
-    this.assembledValueSizeSensor = registerSensor(
-        ASSEMBLED_RECORD_VALUE_SIZE_IN_BYTES,
-        new Avg(),
-        new Min(),
-        new Max(),
-        TehutiUtils
-            .getPercentileStat(getName() + AbstractVeniceStats.DELIMITER + ASSEMBLED_RECORD_VALUE_SIZE_IN_BYTES));
+    this.assembledValueSizeSensor = registerSensor(ASSEMBLED_RECORD_VALUE_SIZE_IN_BYTES, avgAndMax());
+
+    this.assembledRecordSizeRatioSensor = registerSensor(ASSEMBLED_RECORD_SIZE_RATIO, avgAndMax());
```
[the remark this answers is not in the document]
KaiSernLim: I see. I agree and I think `Max` is sufficient for this new `ASSEMBLED_RECORD_SIZE_RATIO`.

internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
gaojieliu: Yeah, it is fine to group the setting as one.
Regarding Samza RP, we don't normally put a lot of configs in Samza RP as we can't deploy it as easily as VPJ.
One thing, we can do is to compare the full value size with the limit in Venice Server for Samza RP job and fail the job if it goes beyond the limit.
If that is too complex, we can skip Samza RP job for now as there are not many RP users and they are limited into one single cluster.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -1893,6 +1899,13 @@ protected void recordProcessedRecordStats(
     }
   }
 
+  @Override
+  protected final void recordAssembledRecordSizeRatio(long size, long currentTimeMs) {
+    final double sizeRatio =
```
gaojieliu, line 1904: If the size is unlimited, maybe we shouldn't record `0` all the time as it is a waste of effort.
So just call `recordAssembledRecordSizeRatio` when it is not unlimited.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -283,8 +285,12 @@ public LeaderFollowerStoreIngestionTask(
             .setChunkingEnabled(isChunked)
             .setRmdChunkingEnabled(version.isRmdChunkingEnabled())
             .setPartitionCount(storeVersionPartitionCount)
+            // .setMaxRecordSizeBytes(store.getMaxRecordSizeBytes()) // TODO: allow when nearline jobs are enforced
```
gaojieliu, line 288: I think we should remove this as the logic can be different in the future.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -3180,6 +3182,7 @@ private int processKafkaDataMessage(
                 ByteUtils.extractByteArray(put.getPutValue()), // must be done before recordTransformer changes putValue
                 AvroProtocolDefinition.CHUNKED_VALUE_MANIFEST.getCurrentProtocolVersion());
             hostLevelIngestionStats.recordAssembledValueSize(chunkedValueManifest.getSize(), currentTimeMs);
+            recordAssembledRecordSizeRatio(keyLen + chunkedValueManifest.getSize(), currentTimeMs); // TODO: RMD?
```
gaojieliu, line 3185: `RMD` is a good concern, and maybe we should emit two metrics: one for user supplied value and another one is for `RMD`, and when we setup alert, we should only set it up against the value size ratio and for RMD, we only need to emit the absolute size, no ratio is needed and for now, it will be mostly for informational purpose.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
KaiSernLim: That's a good point. It's pointless to record `0` all the time. I don't think that the value should ever be set to unlimited and the effective default value should be `100MB` though... So there should usually be values for this.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
KaiSernLim: Yeah maybe it's too early for this. Removed.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
KaiSernLim: That would be an nice idea, but I do not know how to determine whether it is RMD or not. As far as I can tell from `WritingChunkingHelper#chunkPayloadAndSend()`, the `ChunkedValueManifest` is the same for both RMD and value?

I spoke with @sixpluszero and due to the difficulty in keeping track of value manifest (which only has value size) and rmd manifest (for chunked rmd) together, it might be better to leave it as-is and recognize this limitation in the calculation of size. Since we control the threshold for alerting customers, perhaps we could make it lower to account for the missing RMD?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
gaojieliu: Well, for non chunked stores, we don't need to record this metric in theory as it is not our target, right?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -285,6 +287,9 @@ public LeaderFollowerStoreIngestionTask(
             .setPartitionCount(storeVersionPartitionCount)
             .build();
     this.veniceWriter = Lazy.of(() -> veniceWriterFactory.createVeniceWriter(writerOptions));
+    this.maxRecordSizeBytes = (store.getMaxRecordSizeBytes() < 0) // TODO: move to VeniceWriter when nearline supported
```
gaojieliu, line 290: I assume the store-level/default config are only for chunk enabled stores, right?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
gaojieliu: I think we need to differentiate the chunked payload is for value or RMD, and for RMD, we shouldn't alert the customers as RMD is the internal component to handle DCR and customers shouldn't need to be aware of that, and also today in theory, RMD can grow to be very large as it is keeping tombstones.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/HostLevelIngestionStats.java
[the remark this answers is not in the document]
nisargthakkar: Yes, but it adds the overhead of keeping track of a time in layers above and have to carry it around?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
KaiSernLim: Yes, this metric should never be recorded for non-chunked stores.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
KaiSernLim: The configs will always be set on the object, but it should only take effect for chunk-enabled stores: ([VeniceWriter](https://github.com/linkedin/venice/blob/4a20cdf03d362989bfcdf0ee22ffd7f2038fe50a/internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java#L911))

internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
KaiSernLim: > One thing, we can do is to compare the full value size with the limit in Venice Server for Samza RP job and fail the job if it goes beyond the limit.

Yeah, I'm not sure how to do that. I do see that there is a [`StoreReponse` / `StoreInfo` from the controller in `VeniceSystemProducer`](https://github.com/linkedin/venice/blob/4a20cdf03d362989bfcdf0ee22ffd7f2038fe50a/clients/venice-samza/src/main/java/com/linkedin/venice/samza/VeniceSystemProducer.java#L541)? Then to fail the job, we would just throw an `Exception`?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
KaiSernLim: I checked with @sixpluszero and I have a method of differentiating whether a manifest pertains to a value or RMD, so I should be able to separate the two now.

KaiSernLim: I do find it a bit odd that the alert on the customer side does not include RMD size, but [the condition for `RecordTooLarge` does include RMD size in being `key + value + rmd`](https://github.com/linkedin/venice/blob/4a20cdf03d362989bfcdf0ee22ffd7f2038fe50a/internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java#L909)? If the RMD size is very large, then maybe we could see `RecordTooLarge` failures before the ratio reaches high levels?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/HostLevelIngestionStats.java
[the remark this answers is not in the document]
KaiSernLim: I feel like it's widely used under this `StoreIngestionTask#internalProcessConsumerRecord()` method, and since the variable is there, I may as well use it.

I don't have a strong preference for either way.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
gaojieliu: I think we should mutate the logic in VeniceWriter, so that the limit only applies to the value record and once we have a separate metric to track the RMD size, it will be clear which part is causing issue.
RMD is the internal component, which Venice team is responsible for the size optimization, and we shouldn't fail user's write if RMD is too large from my understanding.

gaojieliu: Even we want to put up a limit against RMD size to avoid the damage to other tenants, we should use a separate feature flag to control this logic and it should be off by default and we can use an uniform limit controlled by Venice Config from the beginning.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
gaojieliu: But in `LFSIT`, we are doing the size check/ratio calculation before passing it to `VeniceWriter`, right?

internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
gaojieliu: I meant in `SIT`, Venice Server can throw an exception when the value size hits the limit, since the exception will be thrown before `EOP`, the Samza RP job will be killed.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
KaiSernLim: Understood. I'll change the logic in `VeniceWriter` to only look at `key + value` size:

```java
int veniceRecordSize = serializedKey.length + serializedValue.length + replicationMetadataPayloadSize;
if (isChunkingNeededForRecord(veniceRecordSize)) { // ~1MB default
  // RMD size is not checked because it's an internal component, and a user's write should not be failed due to it
  if (isChunkingEnabled && !isRecordTooLarge(serializedKey.length + serializedValue.length)) {
```

I think RMD is also stored in a separate column, so maybe it'll be fine not to count it at all?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
KaiSernLim: It is true that in `LFSIT`, we are doing the size check/ratio calculation on the Kafka message before passing the message to `VeniceWriter`.

Regarding configs, the store-level/default config will be retrieved from `VeniceServerConfig` by `LFSIT` and passed to `VeniceWriter` via `VeniceWriterOptions` in the constructor of `LFSIT`.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
gaojieliu: RMD size should be dealt with separately as it is invisible to the end user and if we alert it, it would cause a lot of confusions.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
KaiSernLim: > if the store doesn't have chunked enabled, this.maxRecordSizeBytes will be using the default size: 100MB even for non-chunking stores, is it right?

This is correct. 

> Maybe we shouldn't need to emit the size ratio metrics for non-chunking stores?

The metric should already not be emitted for non-chunking stores because it's dependent on detecting `ChunkedValueManifest`, which should not exist for non-chunking stores:

```java
if (put.getSchemaId() == AvroProtocolDefinition.CHUNKED_VALUE_MANIFEST.getCurrentProtocolVersion()) {
  try {
    ChunkedValueManifest chunkedValueManifest = manifestSerializer.deserialize(put.getPutValue());
    if (isValueManifest) {
      recordAssembledRecordSizeRatio();
      hostLevelIngestionStats.recordAssembledValueSize();
    } else { // the manifest pertains to a chunked rmd
      hostLevelIngestionStats.recordAssembledRmdSize();
    }
  }
}
```

clients/da-vinci-client/src/main/java/com/linkedin/davinci/config/VeniceServerConfig.java
```diff
@@ -778,6 +782,12 @@ public VeniceServerConfig(VeniceProperties serverProperties, Map<String, Map<Str
     recordLevelMetricWhenBootstrappingCurrentVersionEnabled =
         serverProperties.getBoolean(SERVER_RECORD_LEVEL_METRICS_WHEN_BOOTSTRAPPING_CURRENT_VERSION_ENABLED, true);
     identityParserClassName = serverProperties.getString(IDENTITY_PARSER_CLASS, DefaultIdentityParser.class.getName());
+    defaultMaxRecordSizeBytes = serverProperties.getInt(DEFAULT_MAX_RECORD_SIZE_BYTES, 100 * BYTES_PER_MB);
```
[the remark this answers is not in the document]
KaiSernLim: Added `DEFAULT_MAX_RECORD_SIZE_BYTES_VALUE` in `ConfigConstants` for this.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
gaojieliu: Got it, thanks.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -3203,7 +3204,14 @@ private int processKafkaDataMessage(
             ChunkedValueManifest chunkedValueManifest = manifestSerializer.deserialize(
                 ByteUtils.extractByteArray(put.getPutValue()), // must be done before recordTransformer changes putValue
                 AvroProtocolDefinition.CHUNKED_VALUE_MANIFEST.getCurrentProtocolVersion());
-            hostLevelIngestionStats.recordAssembledValueSize(chunkedValueManifest.getSize(), currentTimeMs);
+            ByteBuffer rmdPayload = put.getReplicationMetadataPayload();
+            boolean isValueManifest = rmdPayload != null && rmdPayload.equals(VeniceWriter.EMPTY_BYTE_BUFFER);
+            if (isValueManifest) {
+              recordAssembledRecordSizeRatio(keyLen + chunkedValueManifest.getSize(), currentTimeMs);
```
gaojieliu, line 3210: I just noticed this statement:
```
recordAssembledRecordSizeRatio(keyLen + chunkedValueManifest.getSize(), currentTimeMs);
```
This seems to be incorrect, as we would like to emit the ratio, not the absolute size.
Shouldn't be:
```
recordAssembledRecordSizeRatio((keyLen + chunkedValueManifest.getSize() / maxRecordSizeBytes)  , currentTimeMs);
```

Or override the implementation in `LFSIT`?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -3203,7 +3204,14 @@ private int processKafkaDataMessage(
...
+              hostLevelIngestionStats.recordAssembledValueSize(chunkedValueManifest.getSize(), currentTimeMs);
```
gaojieliu, line 3211: The above statement is measuring both key and value and this line is only measuring value.
This is confusing, and I think we need to make them consistent, and I think it is fine to only measure value size for now by skipping `keyLen` as `key` is typically small and the chunking is only for value.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
KaiSernLim: Yeah, since `maxRecordSizeBytes` needs `VeniceWriter`, the setting is only present on `LFSIT`. The implementation is overridden in `LFSIT`, but I do agree that it is confusing with the current naming:
```java
  @Override
  protected final void recordAssembledRecordSizeRatio(long size, long currentTimeMs) {
    if (maxRecordSizeBytes != VeniceWriter.UNLIMITED_MAX_RECORD_SIZE) {
      hostLevelIngestionStats.recordAssembledRecordSizeRatio((double) size / maxRecordSizeBytes, currentTimeMs);
    }
  }
```

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/HostLevelIngestionStats.java
```diff
@@ -473,6 +475,14 @@ public void recordAssembledValueSize(long bytes, long currentTimeMs) {
     assembledValueSizeSensor.record(bytes, currentTimeMs);
   }
 
+  public void recordAssembledRmdSize(long bytes, long currentTimeMs) {
+    assembledRmdSizeSensor.record(bytes, currentTimeMs);
+  }
+
+  public void recordAssembledRecordSizeRatio(double ratio, long currentTimeMs) {
```
gaojieliu, line 482: Actually, this method is causing confusions:
In `SIT`:
```
public void recordAssembledRecordSizeRatio(double ratio, long currentTimeMs)
```
In `LFSIT`:
```
  protected final void recordAssembledRecordSizeRatio(long size, long currentTimeMs) {
```

Why do we need to implement in this way?
I guess `SIT` can get the max record size and we don't need the override in `LFSIT`, and this method is wrong.

gaojieliu (commented):
I realized there are some inconsistent understanding about this max size feature and some logic is enforcing/measuring value only and some logic is applying to both key and value.
The latter measurement is more aligned with the naming: `max record size`, and I think we should be consistent in all the places regarding enforcement and measurement.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/HostLevelIngestionStats.java
KaiSernLim: Both `SIT` and `LFSIT` have `long size`. The method with `double ratio` should be in `HostLevelIngestionStats`?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
gaojieliu: Sure, that will be better.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
KaiSernLim: What is confusing about these two? They are separate metrics that both utilize value manifests.
 
`recordAssembledValueSize()` is for an existing metric measuring assembled values that was added prior to this Handling Large Records project as part of a GCN action item.

`recordAssembledRecordSizeRatio()` is for comparing the current record's size with the `maxRecordSizeBytes` setting.  I thought this project was specifically about handling large records, which consist of key + value (and possible RMD)?

I actually had confusion even over the term "record" at the beginning, because there is also the concept of a record in Kafka.

KaiSernLim: If it is fine to only measure value size, I will change the name to `recordAssembledValueSizeRatio()`.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -3980,6 +4005,10 @@ protected boolean isChunked() {
     return isChunked;
   }
 
+  void setRmdChunked(boolean isRmdChunked) {
```
gaojieliu, line 4008: Nit:
This is for testing purpose, if yes, can you leave some javadoc here?

gaojieliu: Also, can we use other ways to test this code path? I think we can mock `Version` to simulate the RMD chunking, right? Having such setter is not good unless it is absolutely necessary.

gaojieliu (commented):
Left a minor comment and other than it, the PR looks good.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
sixpluszero: Does this means we will eventually set the threshold based on pure user value size, and internally given that we know RMD size is kind of proportional to value size, the actual limit would be as much as 2x of the set up threshold?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -2428,6 +2430,50 @@ public void recordChecksumVerificationFailure() {
...
+   * Also records the ratio of assembled record size to maximum allowed size, which is intended to be used to alert
+   * customers about how close they are to hitting the size limit.
+   * @param keyLen The size of the record's key
+   * @param valueBytes {@link Put#putValue} which is expected to be a serialized {@link ChunkedValueManifest}
+   * @param rmdBytes {@link Put#replicationMetadataPayload} which can be a serialized {@link ChunkedValueManifest} if
+   * RMD chunking was enabled or just the RMD payload otherwise
+   */
+  protected void recordAssembledRecordSize(int keyLen, ByteBuffer valueBytes, ByteBuffer rmdBytes, long currentTimeMs) {
+    try {
+      // TODO: Check with Jialin on this
```
sixpluszero, line 2445: Should we remove this or if you still want to check on this point?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -2428,6 +2430,50 @@ public void recordChecksumVerificationFailure() {
...
+        return;
+      }
+
+      int rmdSize = rmdBytes.remaining();
+      if (ByteUtils.getIntHeaderFromByteBuffer(rmdBytes) == CHUNK_MANIFEST_SCHEMA_ID) {
+        byte[] rmdByteArray = ByteUtils.extractByteArray(rmdBytes);
+        ChunkedValueManifest rmdManifest = manifestSerializer.deserialize(rmdByteArray, CHUNK_MANIFEST_SCHEMA_ID);
+        rmdSize = rmdManifest.getSize();
+      }
+      hostLevelIngestionStats.recordAssembledRmdSize(rmdSize, currentTimeMs);
```
sixpluszero, line 2467: Looks like we are creating different metrics on value and RMD, will we enforce RMD size checking as well or it is just auxiliary information for oncall to check cluster/store status?

KaiSernLim: It's the latter. This RMD size will just be a separate metric just for our information.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
KaiSernLim: Good suggestion on using other ways to test. Removed this entirely.

sixpluszero (approved):
Thank you for your change. 
btw I'd suggest for the implicit RMD chunking logic implemented in code with AA enabled and chunking enabled, we should cut a ticket describe the finding and current status even though I don't think we need to act on it right now.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
KaiSernLim: Checked with @sixpluszero offline, and implemented a solution.

PR #1079, opened 2024-07-24
[da-vinci][controller] Add check to ensure version is valid before ingestion

### LINKED ISSUE #657: [Feature] Venice Push Status System store needs to have a stronger guarantee since it is a critical piece in Push Job
### Willingness to contribute

### Feature Request Proposal

Right now, Venice Push Status System Store creation seems to be a best effort in prod and some stores have it and some stores are not.

But polling DaVinci Push Status System Store is in the critical path of push job offline status polling and if push status system store is not there, the push job will never succeed.

One idea is to have a monitoring service in Controller to periodically check the healthiness of the push status system store for each Venice store and fix it automatically.

### Motivation

This is to improve write path availability.

### Details

_No response_

- [X] `Controller`: This is the control-plane for Venice. Used to create/update/query stores and their metadata.
- [X] `Da Vinci Client`: This is an embedded, stateful client that materializes store data locally.

tylerhu3:

Add check to ensure version is valid before ingestion

Resolves #657

## How was this PR tested?
Ran "./gradlew check --continue"

- [ X] No. You can skip the rest of this section.


services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
```diff
@@ -1000,8 +1002,25 @@ public void addVersionAndStartIngestion(
...
-    // Parent controller will always pick the replicationMetadataVersionId from configs.
+    // Parent controller will always pick the replicationMetadataVersionId from
+    // configs.
     final int replicationMetadataVersionId = getRmdVersionID(storeName, clusterName);
+
+    // Ensure valid version
+    if (getStore(clusterName, storeName).isDaVinciPushStatusStoreEnabled()) {
+      Set<Map.Entry<String, Integer>> versionSet = getCurrentVersionsForMultiColos(
+          clusterName,
+          SystemStore.SYSTEM_STORE_NAME_PREFIX + VeniceSystemStoreUtils.DAVINCI_PUSH_STATUS_STORE_STR + "_" + storeName)
```
nisargthakkar, line 1013: nit: This can be replaced by `VeniceSystemStoreType.getSystemStoreName(storeName)`

mynameborat:
Based on the original issue, we still only add a check here as opposed to trying to create one if it doesn't exist? The issue also suggests to try to create one if not exists and then fail if the creation fails.

Do we plan to add that in subsequent PRs? If so, I 'd clarify the description to specify it partially resolves the issue.

kvargha:
Closing as contributor is unable to continue working on this

PR #1080, opened 2024-07-24
[admin-tool]Modified integration test for blob-transfer

sebas-inf:

Modified an integration test to account for DaVinci blob transfer

## How was this PR tested?

Ran the test and it worked as expected

### DIFF
--- a/clients/venice-admin-tool/src/test/java/com/linkedin/venice/TestAdminTool.java
+++ b/clients/venice-admin-tool/src/test/java/com/linkedin/venice/TestAdminTool.java
@@ -74,13 +74,15 @@ public class TestAdminTool {
   public void testAdminUpdateStoreArg() throws ParseException, IOException {
     final String K1 = "k1", V1 = "v1", K2 = "k2", V2 = "v2", K3 = "k3", V3 = "v3";
     String[] args = { "--update-store", "--url", "http://localhost:7036", "--cluster", "test-cluster", "--store",
-        "testStore", "--rmd-chunking-enabled", "true", "--partitioner-params",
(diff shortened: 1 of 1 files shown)


clients/venice-admin-tool/src/test/java/com/linkedin/venice/TestAdminTool.java
```diff
@@ -74,13 +74,15 @@ public void testAdminToolRequiresSSLConfigFile() {
   public void testAdminUpdateStoreArg() throws ParseException, IOException {
```
haoxu07, line 74: Seems this is a unit test?

sebas-inf: I was told that it was an integration test

sixpluszero:
Let's follow the convention of the title tag to [admin-tool] as it is only touching admin tool

clients/venice-admin-tool/src/test/java/com/linkedin/venice/TestAdminTool.java
sixpluszero: Yeah I did not checked how it was implemented before I told you to add/modify an integration test, but the integration test usually spin up the whole venice cluster in the test suite and run end to end. This only creates admin tool object so it is not.

sebas-inf: oh ok

sixpluszero (approved):
Could you update the title to update "integration test" to "unit test" before you merge the PR? Thank you!

PR #1081, opened 2024-07-24
[vpj][server] Fix DELETE handling in TTL repush and allow server to auto convert value level TS to field level TS

sixpluszero:
## [vpj][server] Fix DELETE handling in TTL repush and allow server to auto convert value level TS to field level TS
This PR addresses two issues related to AAWC:
(1) SIT does not fail when it tries to merge PUT with existing value level RMD record. This can happen when user enables PartialUpdate for an AA store and performs either a restart or repush.
(2) TTL Repush job does not fail when DELETE is the latest value of a key.

## How was this PR tested?

Add new integration test and modify existing integration test.
Will also try to add new unit test for more coverage.
Now we can perform repush to create a new version or just execute a cluster wise restart to pick up the PU config change (the restart case it not ideal, there should be a future PR to refactor ParitalUpdate config to become a version aware config)

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/replication/merge/MergeConflictResolver.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/replication/merge/MergeConflictResolver.java
@@ -312,15 +312,14 @@ public class MergeConflictResolver {
       long newValueSourceOffset,
       int newValueSourceBrokerID,
       int newValueSchemaID) {
-    if (!(oldTimestampObject instanceof GenericRecord)) {
-      throw new IllegalStateException(
-          "Per-field RMD timestamp must be a GenericRecord. Got: " + oldTimestampObject + " and store name is: "
-              + storeName);
+
+    if (oldTimestampObject instanceof GenericRecord) {
+      final GenericRecord oldValueFieldTimestampsRecord = (GenericRecord) oldTimestampObject;
+      if (ignoreNewPut(oldValueSchemaID, oldValueFieldTimestampsRecord, newValueSchemaID, putOperationTimestamp)) {
+        return MergeConflictResult.getIgnoredResult();
+      }
     }
-    final GenericRecord oldValueFieldTimestampsRecord = (GenericRecord) oldTimestampObject;
-    if (ignoreNewPut(oldValueSchemaID, oldValueFieldTimestampsRecord, newValueSchemaID, putOperationTimestamp)) {
-      return MergeConflictResult.getIgnoredResult();
-    }
+
     final SchemaEntry mergeResultValueSchemaEntry =
         mergeResultValueSchemaResolver.getMergeResultValueSchema(oldValueSchemaID, newValueSchemaID);
     /**
--- a/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/input/kafka/ttl/VeniceRmdTTLFilter.java
+++ b/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/input/kafka/ttl/VeniceRmdTTLFilter.java
@@ -17,6 +17,7 @@ import com.linkedin.venice.schema.rmd.RmdUtils;
 import com.linkedin.venice.schema.rmd.RmdVersionId;
 import com.linkedin.venice.serializer.RecordDeserializer;
 import com.linkedin.venice.serializer.RecordSerializer;
+import com.linkedin.venice.utils.AvroSchemaUtils;
 import com.linkedin.venice.utils.VeniceProperties;
 import com.linkedin.venice.utils.concurrent.VeniceConcurrentHashMap;
 import java.io.IOException;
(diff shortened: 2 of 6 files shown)


clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/input/kafka/ttl/VeniceRmdTTLFilter.java
```diff
@@ -96,6 +96,14 @@ boolean filterByTTLandMaybeUpdateValue(final INPUT_VALUE value) {
     if (rmdTimestampType.equals(RmdTimestampType.VALUE_LEVEL_TIMESTAMP)) {
       return (long) rmdTimestampObject <= filterTimestamp;
     }
+    ByteBuffer valuePayload = getValuePayload(value);
+    if (valuePayload == null || valuePayload.remaining() == 0) {
+      /**
+       * We do not need check the field level TS for DELETE. If a field has TS older than TTL, it does not matter as it
+       * will still be overwritten by new PUT/UPDATE.
+       */
+      return false;
```
nisargthakkar, line 105: I think we should still filter it out for Auto-purge use-cases. We should not even keep the keys around if a record is expected to be auto-purged

sixpluszero: Yeah, I was struggling on that to make a decision because I basically think that's fine. If you think we should be strict and clear about this I can update the PR.

mynameborat: Isn't the behavior prior this is, the store can potentially drop the newer update depending on the delete timestamp in which case, the filtered record exists in the store as tombstone but now, the records will be filtered by the TTL push and the newer update from RT with older timestamp can potentially be accepted due to lack of RMD resulting in behavioral change?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/replication/merge/MergeConflictResolver.java
```diff
@@ -312,15 +312,14 @@ private MergeConflictResult mergePutWithFieldLevelTimestamp(
...
-    }
-    final GenericRecord oldValueFieldTimestampsRecord = (GenericRecord) oldTimestampObject;
-    if (ignoreNewPut(oldValueSchemaID, oldValueFieldTimestampsRecord, newValueSchemaID, putOperationTimestamp)) {
-      return MergeConflictResult.getIgnoredResult();
+
+    if (oldTimestampObject instanceof GenericRecord) {
+      final GenericRecord oldValueFieldTimestampsRecord = (GenericRecord) oldTimestampObject;
+      if (ignoreNewPut(oldValueSchemaID, oldValueFieldTimestampsRecord, newValueSchemaID, putOperationTimestamp)) {
+        return MergeConflictResult.getIgnoredResult();
+      }
```
mynameborat, line 320: Just by the virtue of not throwing an exception for non generic per-field RMD, sufficient for converting the value level TS to field level TS?

I don't have a clear picture on what takes precedence and how one overrides another by removing this check. Feel free to point me to code blocks that explain the behavior.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/input/kafka/ttl/VeniceRmdTTLFilter.java
sixpluszero: Hmm, my take is: 
1. Prior to this change, the push job will died because of exception. Not working at all. 
2. After the change, prior to Nisarg's suggestion: If final resolved value is a DELETE with higher TS than TTL, then it will be kept, and a PUT with TTL < TS < DELETE TS will still be dropped. If DELETE TS < TTL, for current code change it will be kept, but due to the fact that rewind PUT TS should be > TTL as how we configured TTL time and rewind time for TTL repush, then new PUT should be kept as expected.
3. If I take in Nisarg's change, it will try its best to remove DELETE with TS < TTL as well. This should look cleaner in theory.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/replication/merge/MergeConflictResolver.java
sixpluszero: Yes, there is a code below to actually perform auto conversion:
createOldValueAndRmd in which it calls:
    // RMD record should contain a per-field timestamp and it should use the RMD schema generated from
    // mergeResultValueSchema.
    oldRmdRecord = convertToPerFieldTimestampRmd(oldRmdRecord, oldValueRecord);

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/input/kafka/ttl/VeniceRmdTTLFilter.java
sixpluszero: Addressed the issue in the latest commit. 
I thought about iterating over RMD timestamp fields but that looks not clean. 
Since it is offline flow logic, I ended up creating a default record with old DELETE timestamp for it to go through the below DELETE flow with TTL timestamp, which essentially the same. It is only used to determine whether such record should be kept or not, this default value won't be emitted as the result.

nisargthakkar (approved):
LGTM

PR #1083, opened 2024-07-25
[server] Fix a race condition in SIT graceful shutdown workflow

lluwm:
In deferred write mode, a running checksum is calculated by drainer for every key/value pair from Kafka PUT message and then stored in PartitionConsumptionState.expectedSSTFileChecksum.

When we gracefully close a StoreIngestionTask, SIT thread calls syncOffset to checkpoint current state and the underlying RocksDBSstFileWriter who executes the sync call also calculates checksum of all the key/value pair stored in the input sstFilePath and then verifies if the checksum matches with the value stored in PartitionConsumptionState.

Because the above actions happen in different threads, a race can happen when they run concurrently and a VeniceChecksumException will be thrown in SIT.

To fix this issue, this PR doesn't allow SIT thread to call syncOffset directly. Instead, we let SIT thread to insert a SYNC_OFFSET command node into the drainer queue and let the same drainer thread to execute the SYNC_OFFSET action so that we don't have any race.

SIT thread then waits for command to finish within a certain time period. If times out, SIT can invalidate the command if it hasn't been executed yet. Otherwise, if the command is already in execution, SIT continues to wait for it to finish.

## How was this PR tested?

CI tests passed.


adamxchen:
> To fix this issue, this PR doesn't allow SIT thread to call syncOffset directly. Instead, we let SIT thread to insert a SYNC_OFFSET command node into the drainer queue and let the same drainer thread to execute the SYNC_OFFSET action so that we don't have any race.

Do it mean we add the SYNC_OFFSET by the time of shutdown? How can we guarantee SYNC_OFFSET will must be executed?  (I don't recall if we would drain the consumer action queue by the time of graceful shutdown. Can you please confirm this?)

lluwm:
> Do it mean we add the SYNC_OFFSET by the time of shutdown?

That's correct.

> How can we guarantee SYNC_OFFSET will must be executed? (I don't recall if we would drain the queue by the time of graceful shutdown. Can you please confirm this?)

SIT doesn't guarantee it must be executed. It will do its best by waiting for some time and hope SYNC_OFFSET will be executed and finished. Even if it doesn't sync the most recent offset, we would consume from the earlier checkpoint, which is not efficient, but it shouldn't affect the correctness, if I understand.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreBufferService.java
```diff
@@ -516,6 +544,58 @@ public boolean equals(Object o) {
...
+
+    @Override
+    public int hashCode() {
+      return super.hashCode();
+    }
+
+    @Override
+    public boolean equals(Object o) {
+      return super.equals(o);
+    }
```
gaojieliu, line 596: Since these methods are not doing anything special, we can remove them, right?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreBufferService.java
```diff
@@ -211,6 +212,18 @@ private static void processRecord(
...
+  private static void processCommand(
+      CommandQueueNode cmd,
+      StoreIngestionTask ingestionTask,
+      PartitionConsumptionState pcs) {
+    // We only support SYNC_OFFSET command for now.
+    if (cmd.getCommandType() != CommandQueueNode.CommandType.SYNC_OFFSET) {
+      throw new VeniceException("Unsupported command type: " + cmd.getCommandType());
+    }
+
+    cmd.executeSync(() -> ingestionTask.updateOffsetMetadataAndSyncOffset(pcs));
```
gaojieliu, line 224: Nit,
shall we log an `INFO` message here?

gaojieliu (commented):

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -1450,6 +1449,55 @@ public void run() {
     }
   }
 
+  private void waitForSyncOffsetCmd(StoreBufferService.QueueNode node, PubSubTopicPartition topicPartition)
```
gaojieliu, line 1452: I think it is better to let `execSyncOffsetCommandAsync` to return a customized `CompletableFuture` and the future.cancel will invalidate the scheduled command.
`StoreBufferService.QueueNode` is more like some implementation detail and normally `SIT` shouldn't use it directly.
WDYT?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreBufferService.java
lluwm: Sure, will do.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreBufferService.java
lluwm: I gave a try to remove it and it turns out that SpotBug had an opinion:

> This class extends a class that defines an equals method and adds fields, but doesn't define an equals method itself. Thus, equality on instances of this class will ignore the identity of the subclass and the added fields. Be sure this is what is intended, and that you don't need to override the equals method. Even if you don't need to override the equals method, consider overriding it anyway to document the fact that the equals method for the subclass just return the result of invoking super.equals(o).

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
lluwm: I think the only difference is that in this implementation, `invalidate` needs to acquire a lock first and this lock enforces  a point that if the `runnable` is already executing, then `invalidate` call has to wait for `runnable` to finish. The is useful as it will prevent the SIT thread from continuing to `close`. Concurrent SIT `close` and `runnable` executing is bad because the internal state of SIT is not reliable anymore  and `runnable` (syncOffset) might not work properly. 

```
    public synchronized void executeSync(Runnable runnable) {
      if (valid) {
        runnable.run();
      }
      cmdExecutedFuture.complete(null);
    }
```
I feel like it is not quite easy to achieve the above if we return a `Future`, but maybe I am wrong here.

lluwm (commented):
Thanks @gaojieliu for the review! Replied to your comments inline.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
gaojieliu: The Future can be a **customized** `Future` implementation, which can keep a reference to whatever field is required.
To the user of `StoreBufferService`, it shouldn't need to worry about the internal implementation detail, such as `QueueNode`, which we don't expose it from the beginning.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreBufferService.java
gaojieliu: Sure, we can keep them.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
lluwm: I must missed the `customized` keyword in your first comment :-). Yeah, I also don't like the idea to expose the internal `QueueNode` out, but didn't think of an alternative. It's a good idea with a customized future and I will give a try. Thanks for suggestion!

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreBufferService.java
```diff
@@ -516,6 +546,66 @@ public boolean equals(Object o) {
...
+    public void executeSync(Runnable runnable) {
+      synchronized (cmdExecutedFuture.getLock()) {
+        if (cmdExecutedFuture.isDone() || cmdExecutedFuture.isCancelled()) {
+          LOGGER.warn(
+              "Command {} for {} in drainer queue is already done or cancelled",
+              commandType,
+              getConsumerRecord().getTopicPartition());
+          return;
+        }
+        runnable.run();
```
gaojieliu, line 589: Should we wrap it up by `try/catch` here with proper logging?
```
try {
    runnable.run();
    cmdExecutedFuture.complete(null);
} catch (Exception e) {
    cmdExecutedFuture.completeExceptionally(e);
}

lluwm: agree and it looks in a better shape.

gaojieliu (approved):
Thanks!

PR #1084, opened 2024-07-26
[controller] Delete leaking backup versions from repush

majisourav99:
## Delete leaking backup versions from repush

   If there are still leaking versions due to consecutive repush and intermittent failed push, there could be versions with repushSourceVersion which does not match the current version. This PR will delete those versions after backup retention period.

## How was this PR tested?

CI

### DIFF
--- a/services/venice-controller/src/main/java/com/linkedin/venice/controller/StoreBackupVersionCleanupService.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/controller/StoreBackupVersionCleanupService.java
@@ -276,8 +276,9 @@ public class StoreBackupVersionCleanupService extends AbstractVeniceService {
                 : v.getNumber() < currentVersion)
         .collect(Collectors.toList());
 
-    // If there are still leaking versions due to consecutive repushes, there could be versions with repushSourceVersion
-    // does not match current version, delete them after backup retention period.
+    // If there are still leaking versions due to consecutive repushes with some version failing in other fabric,
+    // there could be versions with repushSourceVersion does not match current version, delete them after backup
+    // retention period.
     if (readyToBeRemovedVersions.isEmpty()) {
       for (Version version: versions) {
         if (version.getNumber() < currentVersion && store.getLatestVersionPromoteToCurrentTimestamp()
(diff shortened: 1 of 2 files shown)


services/venice-controller/src/main/java/com/linkedin/venice/controller/StoreBackupVersionCleanupService.java
```diff
@@ -276,6 +276,16 @@ protected boolean cleanupBackupVersion(Store store, String clusterName) {
...
+    // If there are still leaking versions due to consecutive repushes, there could be versions with repushSourceVersion
+    // does not match current version, delete them after backup retention period.
+    if (readyToBeRemovedVersions.isEmpty()) {
+      for (Version version: versions) {
+        if (version.getNumber() < currentVersion && store.getLatestVersionPromoteToCurrentTimestamp()
+            + defaultBackupVersionRetentionMs < time.getMilliseconds()) {
+          readyToBeRemovedVersions.add(version);
+        }
+      }
+    }
```
sixpluszero, line 288: Could you add one integration to reproduce this behavior and validate the fix?

services/venice-controller/src/main/java/com/linkedin/venice/controller/StoreBackupVersionCleanupService.java
```diff
@@ -276,6 +276,16 @@ protected boolean cleanupBackupVersion(Store store, String clusterName) {
                 : v.getNumber() < currentVersion)
         .collect(Collectors.toList());
 
+    // If there are still leaking versions due to consecutive repushes, there could be versions with repushSourceVersion
+    // does not match current version, delete them after backup retention period.
+    if (readyToBeRemovedVersions.isEmpty()) {
+      for (Version version: versions) {
+        if (version.getNumber() < currentVersion && store.getLatestVersionPromoteToCurrentTimestamp()
```
huangminchn, line 283: My understanding is that this is the same as removing the above filtering?

mynameborat:
The failure scenario is when an ongoing push fails and there is another push triggered which will result in two new versions? I am not fully clear about the bug. 

Shouldn't both of the pushes have the same `repushSourceVersion`?

services/venice-controller/src/main/java/com/linkedin/venice/controller/StoreBackupVersionCleanupService.java
majisourav99: As discussed offline, there is a subtle difference if we remove the above filtering which would lead to leaking 2 backup version for the whole retention period.

huangminchn:
> The failure scenario is when an ongoing push fails and there is another push triggered which will result in two new versions? I am not fully clear about the bug.
> 
> Shouldn't both of the pushes have the same `repushSourceVersion`?

The repush failed in the remote colo but succeeded in the source colo; then the next repush used the version number in the source colo as source version.

huangminchn (approved):
Thanks Sourav!

PR #1087, opened 2024-07-30
[dvc][doc] Create MVP for DaVinciRecordTransformer

kvargha:
## Summary

1. Added `storeRecordsInDaVinci` boolean to the constructor to detect if a user wants to store their transformed records in DaVinci or another storage of their choice.
2. Created `DaVinciRecordTransformerConfig` to be used by `DaVinciConfig`, so we have access to Output Value Class and Schema without needing to create a dummy record transformer outside of `StoreIngestionTask`.
3. Added a recovery method to bootstrap from VT or from local state on startup. If blob transfer is enabled, an error will be thrown since it's not currently supported with the record transformer.
4. Created `AbstractStorageIterator` and a RocksDB implementation to be used by the recovery method.
5. Added logic to detect if transformer logic has changed on startup, utilizing the hash of the class bytecode. If transformer logic has changed, that means local state is stale and we need to bootstrap directly from the VT.
6. Added logic to create a DVRT specific Avro deserializer.
7. Split `put` into `transform` and `processPut`, so that when we're trying to bootstrap from local state, we don't try to transform the data again as it already has been transformed. Also created a wrapper method `transformAndProcessPut` when both need to be called in `StoreIngestionTask`.
8. Created `DaVinciRecordTransformerResult`, to handle cases where the user doesn't transform the value or wants to skip it.
9. Added delete logic in `StoreIngestionTask`.
10. Throw an exception when a user tries to use Ingestion Isolation with DVRT. This can cause issues if a user's transformer is stateful, and it will cause the storage engine to be deleted every time the client comes online. 
11. Updated [record transformer doc](https://kvargha.com/venice/docs/user_guide/read_api/da_vinci_client#record-transformer) with new logic.
12. Made Venice repo a link in the [workspace setup doc](https://kvargha.com/venice/docs/dev_guide/how_to/workspace_setup#fork-the-venice-repository).

## How was this PR tested?

Added more unit and integration tests.


FelixGV:
The thing about doing a hash of the bytecode sounds pretty interesting, but I wonder how reliable it is... I suspect that it probably has _extremely_ low chances of false negative (the code changed but we fail to detect it) but that it might have a somewhat large probability of false positives (the code did not change but we think it did). The latter could be due to differences in JDK versions or flags used for compilation... Maybe it's not a big deal, but IDK...

Also, coming back to false negatives, those may still exist if the scope of the checksumming is not large enough, e.g. the class we checksum did not change but it calls into another class which did change.

Overall, I am not sure how I feel about this mechanism, and I think we may want to have a config to opt out of this check. Possibly even: this should be opt in to begin with (i.e. off by default) until we can gain a better understanding of its limitations.

kvargha:
> The thing about doing a hash of the bytecode sounds pretty interesting, but I wonder how reliable it is... I suspect that it probably has _extremely_ low chances of false negative (the code changed but we fail to detect it) but that it might have a somewhat large probability of false positives (the code did not change but we think it did). The latter could be due to differences in JDK versions or flags used for compilation... Maybe it's not a big deal, but IDK...
> 
> Also, coming back to false negatives, those may still exist if the scope of the checksumming is not large enough, e.g. the class we checksum did not change but it calls into another class which did change.
> 
> Overall, I am not sure how I feel about this mechanism, and I think we may want to have a config to opt out of this check. Possibly even: this should be opt in to begin with (i.e. off by default) until we can gain a better understanding of its limitations.

From my testing, the hash would change under these circumstances:
1. Change in JDK version.
2. Any changes made within the user's implementation (I found this to be very reliable).

Circumstances where it didn't change:
1. Change in version numbers of imported packages that's used by the user's implementation.
2. Code changes in functions imported.
3. Changes in the abstract class.

I think having it off by default and having the user enable it makes sense.

We could have a global variable that's initially set to false, and add another constructor where a user can pass in a boolean of whether or not to enable it. We could call the variable `detectChangesInLogic`. Thoughts @ZacAttack?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
```diff
@@ -247,6 +250,19 @@ public DaVinciBackend(
       cacheBackend = cacheConfig
           .map(objectCacheConfig -> new ObjectCacheBackend(clientConfig, objectCacheConfig, schemaRepository));
 
+      String storeName = clientConfig.getStoreName();
+
+      if (getRecordTransformer != null && getRecordTransformer.apply(0) != null) {
```
ZacAttack, line 255: What is the meaning of doing apply(0) here?  This seems kind of sketchy?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
```diff
@@ -247,6 +250,19 @@ public DaVinciBackend(
...
+        storeRepository.refreshOneStore(storeName);
+        Store store = storeRepository.getStoreOrThrow(storeName);
+        int version = store.getCurrentVersion();
+
+        DaVinciRecordTransformer clientRecordTransformer = getRecordTransformer.apply(version);
+        recordTransformer = new BlockingDaVinciRecordTransformer(
```
ZacAttack, line 261: This will cache a recordtransformer at 'version' but how does it cycle out as versions change?  This backend object I believe doesn't get recreated on new version creation.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/DaVinciRecordTransformer.java
```diff
@@ -96,10 +163,112 @@ public final ByteBuffer getValueBytes(Schema schema, V value) {
...
+   * Bootstraps the client after it comes online and {@link #onStart()} completes.
+   */
+  public final void onRecovery(
+      AbstractStorageEngine storageEngine,
+      StoreBackend storeBackend,
+      BlobTransferManager blobTransferManager,
+      List<Integer> partitions,
+      Optional<Version> version) {
+
+    if (blobTransferManager != null) {
```
ZacAttack, line 224: Can you elaborate for me a bit on why this must be the case?  Looking at the below code, it either bootstraps from VT or local storage.  It would seem that if blob transfer is available, it would just create the local storage and then we would read that.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
```diff
@@ -457,9 +474,16 @@ private synchronized void bootstrap() {
...
+      AbstractStorageEngine storageEngine = storageService.getStorageEngine(versionTopic);
+      aggVersionedStorageEngineStats.setStorageEngine(versionTopic, storageEngine);
       StoreBackend storeBackend = getStoreOrThrow(storeName);
-      storeBackend.subscribe(ComplementSet.newSet(partitions), Optional.of(version));
+
+      if (recordTransformer != null) {
+        recordTransformer
+            .onRecovery(storageEngine, storeBackend, blobTransferManager, partitions, Optional.of(version));
+      } else {
+        storeBackend.subscribe(ComplementSet.newSet(partitions), Optional.of(version));
```
ZacAttack, line 485: Why are we 'else'ing here?  It seems we're doing so because we want all steps in onREcovery to come before storeBackend.subscribe is called.  But that would just happen anyway by virtue of .onRecovery being called before subscribe unless theres something I'm missing.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/StoreBackend.java
```diff
@@ -128,7 +128,7 @@ public CompletableFuture<Void> subscribe(ComplementSet<Integer> partitions) {
     return subscribe(partitions, Optional.empty());
   }
 
-  synchronized CompletableFuture<Void> subscribe(
+  public synchronized CompletableFuture<Void> subscribe(
```
ZacAttack, line 131: Hrm.  Does this need to be public?  Can it just be protected?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/AvroGenericDaVinciClient.java
```diff
@@ -175,6 +176,13 @@ protected AvroGenericDaVinciClient(
     this.managedClients = managedClients;
     this.icProvider = icProvider;
     this.chunkingAdapter = chunkingAdapter;
+
+    this.recordTransformer = daVinciConfig.isRecordTransformerEnabled() ? daVinciConfig.getRecordTransformer(0) : null;
```
ZacAttack, line 180: again, this is kind of sketchy.  Why are we doing a call with version zero?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/AvroGenericDaVinciClient.java
```diff
@@ -175,6 +176,13 @@ protected AvroGenericDaVinciClient(
...
+
+    if (this.recordTransformer != null) {
+      this.clientConfig.setSpecificValueClass(recordTransformer.getOutputValueClass());
```
ZacAttack, line 183: This might be kind of confusing for users.  Users that specify a specificValue class will have that config silently overridden.  We should either avoid that, or at least log that it's getting overwritten.

I think it might be more prudent to actually compare these.  If they're set to different values throw an error, or, check to see if specificvalueclass is null or not.

Also, can we double check how specificVAlueClass is used?  I seem to recall that this decides whether or not we use a Generic or Specific client? (I might be wrong)

clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/AvroGenericDaVinciClient.java
```diff
@@ -646,7 +654,8 @@ protected GenericRecordChunkingAdapter getGenericRecordChunkingAdapter() {
     return GenericRecordChunkingAdapter.INSTANCE;
   }
 
-  private D2ServiceDiscoveryResponse discoverService() {
+  // Visible for testing
+  public D2ServiceDiscoveryResponse discoverService() {
```
ZacAttack, line 658: Can we make this protected?

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/DaVinciClientTest.java
```diff
@@ -234,7 +234,8 @@ public void testBatchStore(DaVinciConfig clientConfig) throws Exception {
         backendConfig)) {
       DaVinciClient<Integer, Object> clientWithRecordTransformer =
           factory.getAndStartGenericAvroClient(storeName1, clientConfig.setRecordTransformerFunction((storeVersion) -> {
-            TestRecordTransformer recordTransformer = new TestRecordTransformer(storeVersion);
+            TestRecordTransformer recordTransformer = null;
```
ZacAttack, line 237: nit: we don't need to split declaration and assignment.

internal/venice-common/src/main/java/com/linkedin/venice/serialization/avro/AvroSpecificStoreDeserializerCache.java
```diff
@@ -27,6 +27,10 @@ public AvroSpecificStoreDeserializerCache(
     this(id -> schemaRepository.getValueSchema(storeName, id).getSchema(), valueClass);
   }
 
+  public AvroSpecificStoreDeserializerCache(Schema valueSchema, Class<V> valueClass) {
```
ZacAttack, line 30: This... is kind of sketchy.  It's sketchy enough that I think we should extend this class to include it perhaps.  This is creating an instance of the deserializer cache that will only return one version of the value schema regardless of what id you pass to it.  I need to wrap my head around this PR more to understand what this is for I think.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/DaVinciRecordTransformer.java
```diff
@@ -96,10 +163,112 @@ public final ByteBuffer getValueBytes(Schema schema, V value) {
...
+    } catch (IOException e) {
+      throw new VeniceException("Failed to get classHash", e);
+    }
+  }
+
+  /**
+   * @return true if the transformation logic has changed since the last time the class was loaded
+   */
+  // Visible for testing
+  public final boolean hasTransformationLogicChanged(int classHash) {
```
ZacAttack, line 192: protected as well?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/DaVinciRecordTransformer.java
```diff
@@ -1,26 +1,46 @@
...
+import com.linkedin.venice.blobtransfer.BlobTransferManager;
+import com.linkedin.venice.exceptions.VeniceException;
+import com.linkedin.venice.meta.Version;
+import com.linkedin.venice.serializer.AvroGenericDeserializer;
 import com.linkedin.venice.serializer.AvroSerializer;
+import com.linkedin.venice.utils.ComplementSet;
 import com.linkedin.venice.utils.lazy.Lazy;
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.FileReader;
```
ZacAttack, line 15: So we're bringing in quite a lot of imports and logic to this interface.  Theres nothing technically wrong with this.

HOWEVER, I think it's distracting.  This interface is weather we like it or not, going to be a bit of user facing documentation.  To that end, let's see if we can slim down the amount of stuff the user has to look at when implementing this.  I think it'd be good if we had a protected/final/static object here that encapsulated the complicated logic a bit more.  That way this interfaces final methods just invoke object.recover/object.checkForChanges, and the user isn't confronted with all this stuff.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -3322,6 +3316,17 @@ private int processKafkaDataMessage(
...
           delete = ((Delete) leaderProducedRecordContext.getValueUnion());
         }
+
+        if (recordTransformer != null) {
+          SchemaEntry keySchema = schemaRepository.getKeySchema(storeName);
+          Lazy<Object> lazyKey = Lazy.of(() -> deserializeAvroObjectAndReturn(ByteBuffer.wrap(keyBytes), keySchema));
+          Object record = recordTransformer.processDelete(lazyKey);
+
+          if (record != null) {
+            break;
```
ZacAttack, line 3326: Is this right?  Is there not a scenario here where we'd delete from the local storage engine?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/AbstractStoragePartition.java
```diff
@@ -175,4 +176,8 @@ public void deleteWithReplicationMetadata(byte[] key, byte[] metadata) {
   public long getRmdByteUsage() {
     throw new VeniceUnsupportedOperationException("getRmdByteUsage");
   }
+
+  public RocksIterator getRocksDBIterator() {
```
ZacAttack, line 180: We cannot do this like this.  AbstractStorage* is meant to abstract the underlying storage.  It shouldn't leak Rocksdb interfaces through it.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/AbstractStorageEngine.java
```diff
@@ -742,4 +743,8 @@ public void suppressLogs(boolean suppressLogs) {
   public boolean hasMemorySpaceLeft() {
     return true;
   }
+
+  public RocksIterator getRocksDBIterator(int partitionId) {
```
ZacAttack, line 747: ditto here.  We should not leak rocksdb concretions through these interfaces.

ZacAttack:
I have finished my first two passes over this.  Good progress, but there are some significant gotchas that have to be addressed.  The first and most primary one being that I do not believe this implementation will work with version pushes.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
kvargha: The delete method originally returned an object before I took over. Maybe there was some motivation that some clients wanted to keep some objects in storage? If not, I can go ahead and change the return type to void.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/DaVinciRecordTransformer.java
kvargha: Blob transfer would require some modifications to its API to get it working with the record transformer, such as passing in the classHash into the blob transfer's request body. However, since work on the blob transfer has been paused, I was advised by Manu to throw an error in the meantime before work is resumed on it.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
kvargha: The intention behind this structure was that if blob transfer was to work with the record transformer, we would need to create a custom subscribe method that would pass in the record transformer so that it can later be referenced when blob transfer is invoked.

But since blob transfer has been paused, I can go ahead and remove the else and subscribe call within onRecovery.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
kvargha: Could we add a check [here](https://github.com/linkedin/venice/blob/95bf948550f51b17801204bc27d4de3b26774cb8/clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/AvroGenericDaVinciClient.java#L714) that `getRecordTransformer` isn't null, so that a new `DaVinciBackend` is created every time `initBackend` is called?

ZacAttack: I don't think we should do that, the code is intended to have DavinciBackend as a singleton object.

I think maybe you should take a look at VersionBackend instead.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/DaVinciRecordTransformer.java
ZacAttack: The hash is stored in rocksdb right?  In which case, why would it need to be part of the api?  It should just transfer

clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
ZacAttack: Is this still the case even with the notion that the hash is stored in rocksdb itself and therefore doesn't need to be in the interface?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/DaVinciRecordTransformer.java
kvargha: Right now, the hash is being written directly to a directory on disk, but it would simplify things if it was just written to RocksDB.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
kvargha: If the hash was stored in RocksDB, then we would actually need subscribe to complete first before we start reading from RocksDB and calling processPut.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/DaVinciRecordTransformer.java
ZacAttack: Ok.  Can you size that task?  I know we don't have a ton of official cycles we can commit here.  If it's a good amount of refactoring to move the hash to rocksdb then I'm content to just leave this as a TODO item.

kvargha: I think it should be a fairly small task. Is not just calling `storageEngine::put` and `storageEngine::get`?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
kvargha: There was a bug in my code that would cause `getRecordTransformer.apply(n)` to return null in some cases. I've addressed it and that check is no longer needed.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/AbstractStoragePartition.java
kvargha: Should I create an `AbstractStorageIterator` then create a RocksDB implementation of it?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
kvargha: By moving `onRecovery` to SIT, bootstrap will complete before `onRecovery` is executed.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/AvroGenericDaVinciClient.java
kvargha: What should the desired behavior be if specificValueClass is null, move forward with a generic client?

ZacAttack: yes

clients/da-vinci-client/classHash-1.txt
```diff
@@ -0,0 +1 @@
+-382233033
```
ZacAttack, line 1: Is this needed for testing?  Could we exclude it?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/AvroGenericDaVinciClient.java
ZacAttack: I'm not sure I follow.  Couldn't we just pass it along via configs?

kvargha: When we're creating the [AvroSpecificStoreDeserializerCache](https://github.com/kvargha/venice/blob/14fef3146d21050013ebff53975cea26a48b99f8/clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/AvroGenericDaVinciClient.java#L797-L799), right now we pass in the ValueClass from the RecordTransformer into the constructor. 

In order to access the ValueClass, we have to initialize a RecordTransformer to extract it.

Since `AvroGenericDaVinciClient` doesn't have access to the RecordTransformer that `SIT` creates, I have to create a dummy one here.

[This](https://github.com/linkedin/venice/blob/2cbd14f3d63fcd22b2ddb60460a431ec6cfc62b4/clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/DaVinciRecordTransformer.java#L267-L273) is the function inside RecordTransformer that returns the ValueClass. Is there a better way to do it?

clients/da-vinci-client/classHash-1.txt
kvargha: No, this was committed by mistake. Thank you for catching this.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/DaVinciRecordTransformer.java
```diff
@@ -29,77 +36,161 @@ public abstract class DaVinciRecordTransformer<K, V, O> {
...
+   * @return the transformed record
    */
-  public final ByteBuffer getValueBytes(Schema schema, V value) {
-    ByteBuffer transformedBytes = ByteBuffer.wrap(new AvroSerializer(schema).serialize(value));
-    ByteBuffer newBuffer = ByteBuffer.allocate(Integer.BYTES + transformedBytes.remaining());
-    newBuffer.putInt(1);
-    newBuffer.put(transformedBytes);
-    newBuffer.flip();
-    return newBuffer;
+  public final O transformAndProcessPut(Lazy<K> key, Lazy<V> value) {
```
kvargha, line 128: Reading about the behavior of the key and value in [SIT](https://github.com/kvargha/venice/blob/91ba18590c7b742b6b73ced7e351602c705740b9/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java#L3316-L3321), I don't think value needs to be lazy since the operation of assembling has already occurred. Right now, it doesn't really do anything.

Thoughts @ZacAttack?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/AvroGenericDaVinciClient.java
ZacAttack: Could we just have the getRecordTransformer also take a schema argument that it gets from the davinciConfig?  That way the user specifies it in the config and then it's accessible at this level.

I really really think that instantiating something that is user defined frivolously is a bad idea because we don't know what the user will have tied to this interface.  It could be very expensive like database instances or file directories.

ZacAttack: Rereading this I think I'm more and more in favor of that just adding an etnry to the davinci config for the output schema, that way we don't need the member here.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/DaVinciRecordTransformer.java
ZacAttack: I'm not sure that's exactly right is it?  The point of the lazy is to handle not having to do value deserialization or chunking unless the user needs it.  The code today acts like a follower node should which is just storing key/values to bytes, and the point of the lazy is to give users the ability to read the value as a serialized/assembled record if they need to process it for their logic.

Deserializing twice is a problem.... but could be solved with a lazy.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/AvroGenericDaVinciClient.java
kvargha: I completely agree. Not a big fan of this hack.

In this file we also need the output value class. Since we're no longer going to instantiate the recordTransformer here, this also needs to be added to the DaVinciConfig. Is that ok?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/DaVinciRecordTransformer.java
kvargha: > Deserializing twice is a problem.... but should be solved via the lazy right? The lazy function should deserialize once and then on successive calls to get it will just return a cached value.
Yes, you're right.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/AvroGenericDaVinciClient.java
kvargha: ToDo: Create record transformer config that's a subconfig of davinciconfig.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
kvargha: After discussing verbally, if we want the user to proceed with deletion in DaVinci they should return null. If they are deleting in an external storage they should return an object.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/DaVinciRecordTransformer.java
kvargha: Added a todo to make chunking lazy.

kvargha: Just to document the discussion made during the meeting. We need to think about what to do with the lazy value, since the parameter for transform is lazy, and processPut is also lazy. Should transform return a lazy value, or should we create a lazy/non-lazy API? TBD.

ZacAttack: To document this a bit further, the 4 scenarios we need to make sure we have correctness for our:

1. User deserialized the value and altered it
2. User deserialized the value and didn't alter it
3. User did not deserialize the value and just passed it long to the storage engine
4. User did not deserialize the value and passed along something else entirely

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -3503,20 +3509,21 @@ private int processKafkaDataMessage(
...
+          ByteBuffer assembledObject = chunkAssembler.bufferAndAssembleRecord(
               consumerRecord.getTopicPartition(),
               putSchemaId,
               keyBytes,
               valueBytes,
               consumerRecord.getOffset(),
-              Lazy.of(() -> new AvroGenericDeserializer<>(valueSchema, valueSchema)),
+              null,
               putSchemaId,
               compressor.get());
```
kvargha, line 3528: We should add an integration test that tests chunking.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/DaVinciRecordTransformerUtility.java
```diff
@@ -0,0 +1,137 @@
...
+      AbstractStorageIterator iterator = storageEngine.getIterator(partition);
+      for (iterator.seekToFirst(); iterator.isValid(); iterator.next()) {
+        byte[] keyBytes = iterator.key();
+        byte[] valueBytes = iterator.value();
+        Lazy<K> lazyKey = Lazy.of(() -> getKeyDeserializer().deserialize(ByteBuffer.wrap(keyBytes)));
+        Lazy<O> lazyValue = Lazy.of(() -> getOutputValueDeserializer().deserialize(ByteBuffer.wrap(valueBytes)));
+        recordTransformer.processPut(lazyKey, lazyValue);
+      }
+    }
+  }
```
kvargha, line 112: We should add an integration test for this.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/DaVinciRecordTransformer.java
```diff
@@ -29,77 +34,176 @@ public abstract class DaVinciRecordTransformer<K, V, O> {
...
+   * Lifecycle event triggered when the current {@link #storeVersion} is being retired.
+   * Use this method to perform cleanup operations such as closing database connections or dropping tables.
    *
    * By default, it performs no operation.
-   *
    */
-  public void onEndIngestionTask() {
+  public void onEnd() {
     return;
   }
```
kvargha, line 117: Since we're abstracting away the ingestion task from the user, I feel like `onEnd` isn't descriptive enough about what the function does. I'm thinking `onShutdown` would be better.

kvargha (commented):
ToDo: Throw error if Ingestion Isolation is turned on with DVRT.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/DaVinciRecordTransformerUtility.java
```diff
@@ -0,0 +1,161 @@
...
+      // Bootstrap from local storage
+      AbstractStorageIterator iterator = storageEngine.getIterator(partition);
+      for (iterator.seekToFirst(); iterator.isValid(); iterator.next()) {
+        byte[] keyBytes = iterator.key();
+        byte[] valueBytes = iterator.value();
+        Lazy<K> lazyKey = Lazy.of(() -> getKeyDeserializer().deserialize(ByteBuffer.wrap(keyBytes)));
+        Lazy<O> lazyValue = Lazy.of(() -> {
+          ByteBuffer valueByteBuffer = ByteBuffer.wrap(valueBytes);
+          // Skip schema id
+          valueByteBuffer.position(Integer.BYTES);
```
kvargha, line 123: Key won't always be an integer, right?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/DaVinciRecordTransformer.java
kvargha: Renaming to `onEndVersionIngestion`

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestStringRecordTransformer.java
```diff
@@ -0,0 +1,65 @@
...
+    return Schema.create(Schema.Type.STRING);
+  }
+
+  public DaVinciRecordTransformerResult<String> transform(Lazy<Integer> key, Lazy<String> value) {
+    String valueStr = convertUtf8ToString(value.get());
+    String transformedValue = valueStr + "Transformed";
+    return new DaVinciRecordTransformerResult<>(DaVinciRecordTransformerResult.Result.TRANSFORMED, transformedValue);
+  }
+
+  public void processPut(Lazy<Integer> key, Lazy<String> value) {
```
ZacAttack, line 33: random: Shouldn't this @Override?

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/DaVinciClientTest.java
```diff
@@ -239,50 +239,315 @@ public void testConcurrentGetAndStart() throws Exception {
...
+      // Test single-get access
+      for (int k = 1; k <= numKeys; ++k) {
+        Object valueObj = clientWithRecordTransformer.get(k).get();
+        String expectedValue = "a" + k + "Transformed";
+        assertEquals(valueObj.toString(), expectedValue);
+        assertEquals(recordTransformer.get(k), expectedValue);
+      }
+
+      // Mimic restart
+      clientWithRecordTransformer.close();
```
ZacAttack, line 329: Just to make sure I understand the implication of this test, does this run through on disk state on restart?  Or.... does it actually reingest?

kvargha: The former. I used the debugger to verify that it's actually reading from disk and not re-ingesting.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/DaVinciClientTest.java
```diff
@@ -239,50 +239,315 @@ public void testConcurrentGetAndStart() throws Exception {
...
+        assertEquals(valueObj.toString(), expectedValue);
+        assertEquals(recordTransformer.get(k), expectedValue);
+      }
+
+      clientWithRecordTransformer.unsubscribeAll();
+    }
+  }
+
+  @Test(timeOut = TEST_TIMEOUT, dataProvider = "dv-client-config-provider", dataProviderClass = DataProviderUtils.class)
+  public void testRecordsInDaVinciDisabledRecordTransformer(DaVinciConfig clientConfig) throws Exception {
```
ZacAttack, line 420: I don't understand this test.  What exactly is this case??

kvargha: This tests the case where a user passes false into the [constructor](https://github.com/kvargha/venice/blob/beb821f72cea7536f5e2c1844d8e2b733ffd284e/clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/DaVinciRecordTransformer.java#L50) for `storeRecordsInDaVinci`.

That means nothing should be stored in DaVinci.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/AvroGenericDaVinciClient.java
```diff
@@ -657,7 +667,7 @@ protected GenericRecordChunkingAdapter getGenericRecordChunkingAdapter() {
     return GenericRecordChunkingAdapter.INSTANCE;
   }
 
-  private D2ServiceDiscoveryResponse discoverService() {
+  D2ServiceDiscoveryResponse discoverService() {
```
ZacAttack, line 670: protected?  Or /* package private */ ?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/AvroGenericDaVinciClient.java
```diff
@@ -795,12 +806,28 @@ public synchronized void start() {
       this.keyDeserializer = FastSerializerDeserializerFactory.getFastAvroGenericDeserializer(keySchema, keySchema);
       this.genericRecordStoreDeserializerCache =
           new AvroStoreDeserializerCache(daVinciBackend.get().getSchemaRepository(), getStoreName(), true);
-      this.storeDeserializerCache = clientConfig.isSpecificClient()
-          ? new AvroSpecificStoreDeserializerCache<>(
+
+      if (clientConfig.isSpecificClient()) {
+        if (daVinciConfig.isRecordTransformerEnabled()) {
+          if (recordTransformerConfig.getOutputValueClass() != clientConfig.getSpecificValueClass()) {
```
ZacAttack, line 812: Let's do a test case through these, I think looking back at the integration test cases the transformed type is always the same.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/DaVinciClientTest.java
kvargha: Updated the function name to be clearer and added additional comments.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/AvroGenericDaVinciClient.java
kvargha: I have a couple of [unit tests](https://github.com/kvargha/venice/blob/83a88bd89be67696ca39ae513689941fbe265ea8/clients/da-vinci-client/src/test/java/com/linkedin/davinci/client/AvroGenericDaVinciClientTest.java#L109-L128) that go over this. Unless you mean changing the types? Ex: Int -> String?

ZacAttack: Yeah, let's do a type change

ZacAttack (dismissed):
This looks really good.  I have the mention about the test case, but this PR has gone on quite a while.  If you think you can knock it out quickly let's get it added on, otherwise I think this ready.  Great work!

ZacAttack (approved):

PR #1088, opened 2024-07-30
[controller] Add retry thread for serviceDiscoveryAnnouncers that fail to register

bonytoni:
## Add retry thread for serviceDiscoveryAnnouncers that fail to register

In future iterations, service discovery announcers can fail registration if it is announcing to a mismatched d2 zk host. This PR contains the logic to retry registration in case of failure.
- `AsyncRetryingServiceDiscoveryAnnouncer` is an implementation of `ServiceDiscoveryAnnouncer` contains logic for registering and unregistering the announcers as well as starting and stopping the thread to asynchronously register failed announcers.
- `ServiceDiscoveryAnnouncerRetryTask` contains the task to retry registering failed announcers.
- Updated `VeniceController` with new implementation of registering and unregistering
- Added config key `SERVICE_DISCOVERY_REGISTRATION_RETRY_MS` in `ConfigKeys` which is the time interval to retry registration for failed announcers. By default, this is set to 30 seconds.
  - Added constant to `VeniceControllerClusterConfig` and `VeniceControllerMultiClusterConfig`

## How was this PR tested?

Unit Tests


services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerClusterConfig.java
```diff
@@ -504,6 +505,7 @@ public class VeniceControllerClusterConfig {
   private final boolean errorLeaderReplicaFailOverEnabled;
 
   private final Set<String> childDatacenters;
+  private final Long retryRegisterServiceDiscoveryAnnouncerMS;
```
nisargthakkar, line 508: Let's change the `Long` to `long`

internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
```diff
@@ -203,6 +203,12 @@ private ConfigKeys() {
    */
   public static final String DEPRECATED_TOPIC_RETENTION_MS = "deprecated.topic.retention.ms";
 
+  /**
+   * Time interval to retry registering service discovery announcers that failed to register. By default, this is set to 30000L ms.
+   */
+  public static final String RETRY_REGISTER_SERVICE_DISCOVERY_ANNOUNCER_MS =
+      "retry.register.service.discovery.announcer.ms";
```
nisargthakkar, line 210: Let's also rename to `service.discovery.registration.retry.ms`

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerClusterConfig.java
```diff
@@ -899,6 +901,8 @@ public VeniceControllerClusterConfig(VeniceProperties props) {
     this.danglingTopicCleanupIntervalSeconds = props.getLong(CONTROLLER_DANGLING_TOPIC_CLEAN_UP_INTERVAL_SECOND, -1);
     this.danglingTopicOccurrenceThresholdForCleanup =
         props.getInt(CONTROLLER_DANGLING_TOPIC_OCCURRENCE_THRESHOLD_FOR_CLEANUP, 3);
+    this.retryRegisterServiceDiscoveryAnnouncerMS =
+        props.getLong(RETRY_REGISTER_SERVICE_DISCOVERY_ANNOUNCER_MS, 30000L);
```
nisargthakkar, line 905: Let's use `30 * Time.MS_PER_SECOND` to help with readability

services/venice-controller/src/main/java/com/linkedin/venice/controller/ServiceDiscoveryAnnouncerHelper.java
```diff
@@ -0,0 +1,53 @@
...
+import com.linkedin.venice.servicediscovery.ServiceDiscoveryAnnouncer;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.BlockingQueue;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+
+public class ServiceDiscoveryAnnouncerHelper {
```
nisargthakkar, line 12: Instead of having a `ServiceDiscoveryAnnouncerHelper` class, can we add a new `ServiceDiscoveryAnnouncer` implementation that wraps a list of other `ServiceDiscoveryAnnouncer` objects?

Like `RetryingServiceDiscoveryAnnouncer` or `AsyncRetryingServiceDiscoveryAnnouncer`?

Let's also move it to the same folder as `ServiceDiscoveryAnnouncer`. There is nothing Controller-specific in this class

internal/venice-common/src/main/java/com/linkedin/venice/servicediscovery/ServiceDiscoveryAnnouncerRetryTask.java
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  @Override
+  public void run() {
+    while (true) {
+      ServiceDiscoveryAnnouncer announcer = null;
+      try {
+        LOGGER.info("Thread for retry task sleeping for {} ms", serviceDiscoveryRegistrationRetryMS);
+        Thread.sleep(serviceDiscoveryRegistrationRetryMS);
+        LOGGER.info("Thread for retry task woke up");
```
nisargthakkar, line 32: Let's remove some verbosity from this logging. I don't think we need these logging.

Also, we should put the sleep after the `take` since these are two different threads and `.take` is a blocking call.

internal/venice-common/src/test/java/com/linkedin/venice/servicediscovery/TestServiceDiscoveryAnnouncerRetryTask.java
```diff
@@ -0,0 +1,102 @@
...
+    } catch (Exception e) {
+      Assert.fail("The method call should not throw an exception.");
+    }
+  }
+
+  private void putTestToSleep(long ms) {
+    LOGGER.info("Test sleeping for {} ms", ms);
+    sleep(ms);
+    LOGGER.info("Test woke up");
+  }
```
nisargthakkar, line 101: We don't need this additional function and logging. Let's just replace this with `Utils.sleep` directly

nisargthakkar (commented):
Mostly LGTM. Thanks

internal/venice-common/src/main/java/com/linkedin/venice/servicediscovery/ServiceDiscoveryAnnouncerRetryTask.java
```diff
@@ -0,0 +1,56 @@
...
+    this.serviceDiscoveryRegistrationRetryMS = serviceDiscoveryRegistrationRetryMS;
+  }
+
+  /**
+   * If the retry queue is empty, the task will wait until there is an available element in the queue.
+   * Each service discovery announcer will retry registration every {@link ServiceDiscoveryAnnouncerRetryTask#serviceDiscoveryRegistrationRetryMS} milliseconds.
+   */
+  @Override
+  public void run() {
+    while (true) {
```
nisargthakkar, line 33: One minor optimization could be that we could change this while true to `!serviceDiscoveryAnnouncerRetryQueue.isEmpty()` - that way, the thread can exit if there are no more failures

nisargthakkar (commented):
One nit. But it looks quite good now!

nisargthakkar (approved):

PR #1089, opened 2024-07-30
[da-vinci] Added a speedup throttler when DaVinci is bootstrapping current versions

gaojieliu:
When working with DaVinci customers, we often tune the ingestion params to speed up the bootstrapping, but we don't want to use the same setup at the steady state since the high ingestion throughput would affect the performance of the read path. This PR introduces a way to dynamically switch ingestion throttlers based on the ingestion status.
Essentially, this new feature would switch to the speedup throttler when it detects any current version bootstrapping and otherwise, it will continue to use the regular throttler.
To optimize the bootstrapping perf, we can bump up the following pools together with this feature:
1. Consumer pool.
2. Drainer/writer pool. As the GC issue is mainly triggered by the ingestion speed, we can safely increase the pool size.

New configs:
da.vinci.current.version.bootstrapping.speedup.enabled : default false da.vinci.current.version.bootstrapping.quota.records.per.second : default -1 da.vinci.current.version.bootstrapping.quota.bytes.per.second : default -1

## How was this PR tested?
CI

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/IngestionThrottler.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/IngestionThrottler.java
@@ -1,0 +1,151 @@
+package com.linkedin.davinci.kafka.consumer;
+
+import static java.lang.Thread.currentThread;
+
+import com.linkedin.alpini.base.concurrency.Executors;
+import com.linkedin.davinci.config.VeniceServerConfig;
+import com.linkedin.venice.throttle.EventThrottler;
+import com.linkedin.venice.utils.DaemonThreadFactory;
+import java.io.Closeable;
+import java.io.IOException;
+import java.util.Map;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Supplier;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+
+/**
+ * This throttler has the following functionality:
+ * 1. When running in DaVinci mode, if there are active current version bootstrapping with speedup mode is on, this
+ *    ingestion throttler will switch to speedup throttler.
+ * 2. Otherwise, this class will switch to regular throttler.
+ *
+ * This throttler is supposed to be adaptive throttler to speed up the DaVinci bootstrapping and fall back to the regular
+ * mode when the bootstrapping is done.
+ */
+public class IngestionThrottler implements Closeable {
+  private final Logger LOGGER = LogManager.getLogger(IngestionThrottler.class);
+
+  private final static int CURRENT_VERSION_BOOTSTRAPPING_DEFAULT_CHECK_INTERVAL = 30;
+  private final static TimeUnit CURRENT_VERSION_BOOTSTRAPPING_DEFAULT_CHECK_TIMEUNIT = TimeUnit.SECONDS;
+
+  private final ScheduledExecutorService eventThrottlerUpdateService;
+
+  private volatile EventThrottler finalRecordThrottler;
+  private volatile EventThrottler finalBandwidthThrottler;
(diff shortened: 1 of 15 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/IngestionThrottler.java
```diff
@@ -0,0 +1,151 @@
...
+      EventThrottler speedupBandwidthThrottler = new EventThrottler(
+          serverConfig.getDaVinciCurrentVersionBootstrappingQuotaBytesPerSecond(),
+          serverConfig.getKafkaFetchQuotaTimeWindow(),
+          "current_version_speedup_kafka_consumption_bandwidth",
+          false,
+          EventThrottler.BLOCK_STRATEGY);
+      this.eventThrottlerUpdateService =
+          Executors.newSingleThreadScheduledExecutor(new DaemonThreadFactory("Ingestion_Event_Throttler_update"));
+      this.eventThrottlerUpdateService.scheduleAtFixedRate(() -> {
+        Map<String, StoreIngestionTask> ongoingStoreIngestionTaskMap = ongoingIngestionTaskMapSupplier.get();
```
sixpluszero, line 88: Should we exit here if we see isUsingSpeedupThrottler is set to false? Once all current version are bootstrapped we don't need to do this check anymore. I don't know if we can cancel the scheduling inside the schedule logic, but at least we can skip all these logics.

sixpluszero: Or is it valid because we HAVE TO periodically check, since there might be edge case that this check runs first before bootstrap() or active ingestion comes in and if the order does not play well we will not trigger the speed up throttler...

sixpluszero:
> To optimize the bootstrapping perf, we can bump up the following pools together with this feature:
> 
> Consumer pool.
> Drainer/writer pool. As the GC issue is mainly triggered by the ingestion speed, we can safely increase the pool size.

Just to confirm, this PR only address 1, is that right?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/IngestionThrottler.java
gaojieliu: The logic was designed on purpose in this way to handle the runtime current version, which shouldn't be very common.
Another case is that the customer might bootstrap stores one by one, so we don't want to miss any of them.
The check is quite cheap, so keeping it shouldn't introduce much overhead.

gaojieliu:
> > To optimize the bootstrapping perf, we can bump up the following pools together with this feature:
> > Consumer pool.
> > Drainer/writer pool. As the GC issue is mainly triggered by the ingestion speed, we can safely increase the pool size.
> 
> Just to confirm, this PR only address 1, is that right?

The consumer pool/drainer pool tuning can be done without this change, and this PR is mainly to offer a dynamic throttler.

sixpluszero:
> > > To optimize the bootstrapping perf, we can bump up the following pools together with this feature:
> > > Consumer pool.
> > > Drainer/writer pool. As the GC issue is mainly triggered by the ingestion speed, we can safely increase the pool size.
> > 
> > 
> > Just to confirm, this PR only address 1, is that right?
> 
> The consumer pool/drainer pool tuning can be done without this change, and this PR is mainly to offer a dynamic throttler.

Oh you mean just adjusting the pool count in the config. I see..

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/IngestionThrottler.java
sixpluszero: OK, I think that makes sense to me

sixpluszero (approved):
I think the change is clean to me. Thank you! We need to add global config as well

PR #1091, opened 2024-07-31
[controller] Harden update-store workflow

nisargthakkar:
## Harden update-store workflow

This PR unifies the `UpdateStore` logic between child controller and parent controllers. We've faced many issues due to the structure of the code (especially `VeniceParentHelixAdmin`) where a set of store-update operations puts it into a non-healthy state. These are all the changes in this PR:
1. Add a new class `UpdateStoreUtils` that is called by both parent and child controllers to apply the store update and perform the necessary validations.
2. Perform store config validation after applying all configs. This helps check for config compatibility. The following validations are performed currently.
    1. Batch-only:
        1. Inc push is not allowed
        2. WC is not allowed
        3. A/A is not allowed
    2. Hybrid:
        1. Rewind time in seconds cannot be negative
        2. Both offset lag threshold and producer time lag threshold cannot be negative
        4. In multi-region mode, inc push is not allowed for `NON_AGGREGATE` data replication policy
        5. `ACTIVE_ACTIVE` data replication policy is only supported when Active-Active replication is enabled
        6. Partition count must be greater than 0
    3. In single-region mode, incremental push allowed when the store is hybrid
    4. In multi-region mode, incremental push is allowed when the store is hybrid and either of the following are true:
        1. Store has active-active replication enabled
        2. Store doesn't have active-active replication, but `DataReplicationPolicy` is `AGGREGATE`
    5. Storage quota cannot be less than 0
    6. Read quota cannot be less than 0
    7. Read quota cannot be larger than "router count * max_per_router_quota"
    8. Store cannot be active-active replication enabled if it is not also native-replication enabled
    9. Active-Active is not supported when amplification factor is more than 1
    10. Write-compute is not supported when amplification factor is more than 1
    11. Store must have a partitioner-config set
    12. Validate if the partitioner can be built with the specified partitioner configs
    13. Latest superset value schema id must map to a valid value schema id
    14. Max compaction lag < min compaction lag
    15. ETL Proxy user must be set. (We might want to deprecate this and remove the corresponding tests, but we have plans to get KIF-based ETL that could benefit from the ETLConfig in store configs)
4. Added a new concept of "primary controller". This is the controller that is allowed to perform inferred updates. In a multi-region mode, this is the parent controller. In a single-region mode, this is a child controller.
5. Previously each "feature" update incurred an update to the store metadata on Zk - and the corresponding watchers would get invoked. This change replaces all those updates with a single Zk update, which is done after the resulting configs are validated.
6. The `ordinal` from `BackupStrategy` enum was being used to write to the admin channel. This is problematic when the enums evolve. Added a `getValue` function and used that instead.
8. A new config: `controller.external.superset.schema.generation.enabled` has been added to replace `controller.parent.external.superset.schema.generation.enabled` because external superset schema generation must be allowed in single-region mode also. `controller.parent.external.superset.schema.generation.enabled` has been marked deprecated, but it has not been completely removed yet for backward compatibility reasons.
9. Due to the changes to inferred configs, a large chunk of tests started failing as they had incorrect setups. Fixed most of such test setups.

Some side-effects of this change are:
1. Controller now checks if the store schema can support WC at the time of enabling the WC config. There were a few tests that also used unsupported value schemas for WC.
3. Persona validation was only present in parent controller in multi-region mode. This now works in single-region mode as well
4. `SupersetSchemaGeneratorWithCustomProp` had a bug where if the first schema has a custom prop, the future superset schema generation would fail as Avro doesn't allow overriding custom props. This got caught as the update store logic now also tries creating superset schema if a store enabled RC or WC, or if it previously had a superset schema.

There are a few other changes that we should do, but are not done in this PR:
All major operations should only be allowed on the parent controller - Create a store, delete a store, add schemas. We should exclude some system stores from this check like we have for the check allowing batch push to admin in child

# Recommendation for Reviewers
I recommend going through the changes in at least two passes. In the first pass, look through all the stuff that has been purely deleted. (Skip `ParentControllerConfigUpdateUtils` as that has been renamed to `PrimaryControllerConfigUpdateUtils` and it's contents partially moved to `UpdateStoreUtils`. So, GitHub doesn't detect is as a renaming). While doing this pass, you can also glance through various small changes that do not need much pondering over.

In the second pass, follow this review order:
1. `VeniceControllerClusterConfig`, `VeniceControllerConfig`, `VeniceControllerMultiClusterConfig`
2. `VeniceControllerService`
3. `Admin`
5. `UpdateStoreUtils`
7. `PrimaryControllerConfigUpdateUtils`
8. `AdminUtils`
10. `UpdateStoreWrapper` 
11. `VeniceHelixAdmin`
12. `VeniceParentHelixAdmin`
13. `SupersetSchemaGeneratorWithCustomProp`
14. All the test modifications
15. All newly added tests

## How was this PR tested?

Added new tests. Modified existing tests. GH CI


services/venice-controller/src/main/java/com/linkedin/venice/controller/util/UpdateStoreUtils.java
```diff
@@ -0,0 +1,1173 @@
...
+    Optional<Integer> maxRecordSizeBytes = params.getMaxRecordSizeBytes();
+    Optional<Boolean> unusedSchemaDeletionEnabled = params.getUnusedSchemaDeletionEnabled();
+    Optional<Boolean> blobTransferEnabled = params.getBlobTransferEnabled();
+
+    owner.map(addToUpdatedConfigs(updatedConfigs, OWNER)).ifPresent(updatedStore::setOwner);
+    readability.map(addToUpdatedConfigs(updatedConfigs, ENABLE_READS)).ifPresent(updatedStore::setEnableReads);
+    writeability.map(addToUpdatedConfigs(updatedConfigs, ENABLE_WRITES)).ifPresent(updatedStore::setEnableWrites);
+    partitionCount.map(addToUpdatedConfigs(updatedConfigs, PARTITION_COUNT)).ifPresent(updatedStore::setPartitionCount);
+    largestUsedVersionNumber.map(addToUpdatedConfigs(updatedConfigs, LARGEST_USED_VERSION_NUMBER))
+        .ifPresent(updatedStore::setLargestUsedVersionNumber);
```
FelixGV, line 220: Nitpick: This stuff looks a bit like boilerplate to me... I think we could simplify further, along the lines of:

```
  private static <T> void addToUpdatedConfigs(Set<CharSequence> updatedConfigs, String configName, Optional<T> param, Consumer<T> paramSetter) {
    if (param.isPresent()) {
      updatedConfigs.add(configName);
      paramSetter.accept(param.get());
    }
  }
```

And then:

```
  addToUpdatedConfigs(updatedConfigs, LARGEST_USED_VERSION_NUMBER, largestUsedVersionNumber, updatedStore::setLargestUsedVersionNumber);
```

It might still line wrap, depending on the length of the params we pass in, but the map/ifPresent ritual would be gone, which seems clearer to me...

FelixGV (commented):
Thanks! This looks like a good change! I didn't have time to look at it fully yet, but I left one minor comment so far.

services/venice-controller/src/main/java/com/linkedin/venice/controller/util/UpdateStoreUtils.java
nisargthakkar: Let me try it. One more place that could benefit from some boilerplate removal is `VeniceParentHelixAdmin#updateStore`. Can you recommend some suggestions there as well?

services/venice-controller/src/main/java/com/linkedin/venice/controller/init/SystemStoreInitializationHelper.java
```diff
@@ -161,6 +162,11 @@ public static void setupSystemStore(
     }
 
     if (updateStoreQueryParams != null && updateStoreCheckSupplier.apply(store)) {
+      if (store.getPartitionCount() == 0 && !updateStoreQueryParams.getPartitionCount().isPresent()) {
+        updateStoreQueryParams
+            .setPartitionCount(multiClusterConfigs.getControllerConfig(clusterName).getMinNumberOfPartitions());
+      }
+
```
mynameborat, line 169: What if updateStoreQueryParams exists and the partition count is set to invalid count? e.g., <= minNumberOfPartitions? Is that possible?

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/consumer/AdminExecutionTask.java
```diff
@@ -426,8 +426,6 @@ private void handleSetStoreCurrentVersion(SetStoreCurrentVersion message) {
     String storeName = message.storeName.toString();
     int version = message.currentVersion;
     admin.setStoreCurrentVersion(clusterName, storeName, version);
-
-    LOGGER.info("Set store: {} version to {} in cluster: {}", storeName, version, clusterName);
```
mynameborat, line 430: why do we want to remove this?

services/venice-controller/src/main/java/com/linkedin/venice/controller/util/UpdateStoreUtils.java
```diff
@@ -0,0 +1,1215 @@
+package com.linkedin.venice.controller.util;
+
+import static com.linkedin.venice.controllerapi.ControllerApiConstants.ACCESS_CONTROLLED;
+import static com.linkedin.venice.controllerapi.ControllerApiConstants.ACTIVE_ACTIVE_REPLICATION_ENABLED;
+import static com.linkedin.venice.controllerapi.ControllerApiConstants.AMPLIFICATION_FACTOR;
+import static com.linkedin.venice.controllerapi.ControllerApiConstants.AUTO_SCHEMA_REGISTER_FOR_PUSHJOB_ENABLED;
+import static com.linkedin.venice.controllerapi.ControllerApiConstants.BACKUP_STRATEGY;
```
mynameborat, line 7: The convention (noticed in other places in the codebase as well) is to wildcard static imports.
Would make it less verbose and more readable.

mynameborat (commented):
I did a pass. Would recommend having another pair of eyes especially on the following files due to my limited context in the code base
* VeniceParentHelixAdmin
* VeniceHelixAdmin
* UpdateStoreUtils

internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
```diff
@@ -339,8 +339,15 @@ private ConfigKeys() {
       "controller.store.graveyard.cleanup.sleep.interval.between.list.fetch.minutes";
 
   /**
-   * Whether the superset schema generation in Parent Controller should be done via passed callback or not.
+   * Whether the superset schema generation in Primary Controller should be done via passed callback or not.
    */
+  public static final String CONTROLLER_EXTERNAL_SUPERSET_SCHEMA_GENERATION_ENABLED =
+      "controller.external.superset.schema.generation.enabled";
```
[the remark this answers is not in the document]
nisargthakkar: Yes. The "external superset schema generation" feature allows us to plug in customized classes that will be used to decide on what the resulting superset schema contains. Ideally, this should work for single-region and multi-region mode. The current implementation, however only works in multi-region mode. So, the `parent` terminology is being removed.

internal/venice-common/src/main/java/com/linkedin/venice/helix/StoragePersonaRepository.java
```diff
@@ -179,15 +179,15 @@ private void deleteStores(List<String> storeNames) {
...
       return null;
+    }
     return getPersona(personaName);
   }
 
   private boolean isStoreSetValid(StoragePersona persona, Optional<Store> additionalStore) {
     Set<String> setToValidate = new HashSet<>();
-    if (additionalStore.isPresent())
-      setToValidate.add(additionalStore.get().getName());
+    additionalStore.ifPresent(store -> setToValidate.add(store.getName()));
```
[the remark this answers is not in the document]
nisargthakkar: Yeah, it's only a few. Maybe 1-2 files. I added them here because they were problematic when reading this code while making the main change

internal/venice-common/src/main/java/com/linkedin/venice/meta/HybridStoreConfig.java
```diff
@@ -27,4 +27,9 @@ public interface HybridStoreConfig extends DataModelBackedStructure<StoreHybridC
   BufferReplayPolicy getBufferReplayPolicy();
 
   HybridStoreConfig clone();
+
+  default boolean isHybrid() {
+    return this.getRewindTimeInSeconds() >= 0 && (this.getOffsetLagThresholdToGoOnline() >= 0
+        || this.getProducerTimestampLagThresholdToGoOnlineInSeconds() >= 0);
+  }
```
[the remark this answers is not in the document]
nisargthakkar: Yeah, I can remove it. This was essentially copy-pasted from another class

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/ingestionHeartbeat/IngestionHeartBeatTest.java
```diff
@@ -109,15 +112,6 @@ public void setUp() {
...
 
-  @AfterTest(alwaysRun = true)
-  public void cleanupStore() {
-    String parentControllerUrl = parentController.getControllerUrl();
-    try (ControllerClient parentControllerClient =
-        new ControllerClient(multiRegionMultiClusterWrapper.getClusterNames()[0], parentControllerUrl)) {
-      parentControllerClient.disableAndDeleteStore(storeName);
-    }
-  }
-
```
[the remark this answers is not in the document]
nisargthakkar: Kind of. The tests in this class were failing due to the new changes. So, while fixing this, I noticed that it works weirdly.

It uses a global variable just for the purposes of cleanups after the test runs. This deleteion can also fail and cause test flakiness. So, I removed it because it wasn't really needed

services/venice-controller/src/main/java/com/linkedin/venice/controller/Admin.java
```diff
@@ -755,6 +762,13 @@ void updateRoutersClusterConfig(
    */
   boolean isParent();
 
+  /**
+   * The "Primary Controller" term is used to refer to whichever controller is the main controller in a Venice set-up.
+   * In a multi-region deployment, the primary controller is the parent controller.
+   * In a single-region deployment, the primary controller is the only controller.
+   */
+  boolean isPrimary();
+
```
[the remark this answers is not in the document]
nisargthakkar: Actually, this check can be used without having knowledge of multi-region vs single-region. It is what abstracts the need for admin to know about single-region/multi-region setup

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/consumer/AdminExecutionTask.java
nisargthakkar: This line moved to `VeniceHelixAdmin#setStoreCurrentVersion` - and will be logged even when there is single-region mode.

services/venice-controller/src/main/java/com/linkedin/venice/controller/init/SystemStoreInitializationHelper.java
nisargthakkar: This will get validated in `UpdateStoreUtils#validateStorePartitionCountUpdate`

services/venice-controller/src/main/java/com/linkedin/venice/controller/util/UpdateStoreUtils.java
nisargthakkar: We prefer not to use wildcards. The IDEs make it easy to collapse the imports, so, this is not too problematic. The ones that exist need to be removed.

services/venice-controller/src/main/java/com/linkedin/venice/controller/Admin.java
```diff
@@ -1010,4 +1024,21 @@ default void clearInstanceMonitor(String clusterName) {
...
+  /**
+   * Return the current superset schema generator for the given cluster.
+   */
+  SupersetSchemaGenerator getSupersetSchemaGenerator(String clusterName);
+
+  /**
+   * Return the multi-cluster configs for the controller.
+   */
+  VeniceControllerMultiClusterConfig getMultiClusterConfigs();
```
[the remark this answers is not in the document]
nisargthakkar: I think if this was a widely used interface, it makes sense. But this is not externally settable, and only has two implementations: `VeniceHelixAdmin` and `VeniceParentHelixAdmin`

So, this works, IMO. WDYT?

services/venice-controller/src/main/java/com/linkedin/venice/controller/util/AdminUtils.java
```diff
@@ -0,0 +1,113 @@
...
+
+    // A/A can always support incremental push
+    if (activeActiveReplicationEnabled) {
+      return true;
+    }
+
+    DataReplicationPolicy dataReplicationPolicy = hybridStoreConfig.getDataReplicationPolicy();
+    return dataReplicationPolicy == DataReplicationPolicy.AGGREGATE
+        || dataReplicationPolicy == DataReplicationPolicy.NONE;
+  }
```
mynameborat, line 112: Should we leverage this in the `UpdateStoreUtils` for validation of scenarios for incremental push allowed? Seems redudant checks

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
```diff
@@ -3057,25 +2760,71 @@ private SchemaEntry addValueSchemaEntry(
...
+      supersetSchemaCreation.supersetSchema = supersetSchemaMeta;
+      supersetSchemaCreation.supersetSchemaId = supersetSchemaId;
+
+      AdminOperation message = new AdminOperation();
+      message.operationType = AdminMessageType.SUPERSET_SCHEMA_CREATION.getValue();
+      message.payloadUnion = supersetSchemaCreation;
+
+      sendAdminMessageAndWaitForConsumed(clusterName, storeName, message);
+    } finally {
+      releaseAdminMessageLock(clusterName, storeName);
```
mynameborat, line 2826: Can we guarantee that we don't run into scenario where the admin is stuck and this never executes? Is there some sort of timed lock releases?

mynameborat (commented):
Looks good. I mostly made sure that the checks and description in the summary matches w/ the code changes.

Final comment son strategy of releasing the changes
1. Do we have branch based releases for complex changes/features?
2. It feels merging this w/ master and finding issues during certification can be potentially difficult to deal/mitigate as surface area of the changes is large and reverting such large commit accurately (assuming other changes pile in) might be tricky.

How do plan to certify this w/ our current certification flow?

services/venice-controller/src/main/java/com/linkedin/venice/controller/util/UpdateStoreUtils.java
```diff
@@ -0,0 +1,1239 @@
...
+        updatedConfigs,
+        BLOB_TRANSFER_ENABLED,
+        blobTransferEnabled,
+        updatedStore::setBlobTransferEnabled);
+    addToUpdatedConfigs(
+        updatedConfigs,
+        STORAGE_NODE_READ_QUOTA_ENABLED,
+        storageNodeReadQuotaEnabled,
+        updatedStore::setStorageNodeReadQuotaEnabled);
+    addToUpdatedConfigs(updatedConfigs, REGULAR_VERSION_ETL_ENABLED, regularVersionETLEnabled, regularVersionETL -> {
```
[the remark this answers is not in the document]
nisargthakkar: There is some potential of adding blank likes b/w logical groups, but I don't think it is needed here since almost all configs are logically different. It is also easier to modify when adding new configs

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
nisargthakkar: This is carried over from the current way we run this. Also, there is a timeout in `sendAdminMessageAndWaitForConsumed`

services/venice-controller/src/main/java/com/linkedin/venice/controller/util/UpdateStoreUtils.java
```diff
@@ -0,0 +1,1239 @@
...
+    if (!partitionerClass.isPresent() && !partitionerParams.isPresent() && !amplificationFactor.isPresent()) {
+      return oldStore.getPartitionerConfig();
+    }
+
+    PartitionerConfig originalPartitionerConfig;
+    if (oldStore.getPartitionerConfig() == null) {
+      originalPartitionerConfig = new PartitionerConfigImpl();
+    } else {
+      originalPartitionerConfig = oldStore.getPartitionerConfig();
+    }
```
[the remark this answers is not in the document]
nisargthakkar: We generally prefer to avoid using `Optional` where possible. Let me check

services/venice-controller/src/main/java/com/linkedin/venice/controller/util/UpdateStoreUtils.java
```diff
@@ -0,0 +1,1239 @@
...
+    VeniceControllerClusterConfig clusterConfig = admin.getHelixVeniceClusterResources(clusterName).getConfig();
+
+    int newPartitionCount = updatedStore.getPartitionCount();
+    if (newPartitionCount < 0) {
+      String errorMessage = errorMessagePrefix + "Partition count: " + newPartitionCount + " should NOT be negative";
+      LOGGER.error(errorMessage);
+      throw new VeniceHttpException(HttpStatus.SC_BAD_REQUEST, errorMessage, ErrorType.INVALID_CONFIG);
+    }
+
+    if (updatedStore.isHybrid() && newPartitionCount == 0) {
```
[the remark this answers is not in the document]
nisargthakkar: Yes. For batch-only stores, a `partition count = 0` means that each push will compute a new partition count based on the store's assigned quota and our configs of how much each partition should hold. Both of these could change between versions.

services/venice-controller/src/main/java/com/linkedin/venice/controller/util/UpdateStoreUtils.java
```diff
@@ -0,0 +1,1239 @@
...
+      Set<String> regionsFilter = parseRegionsFilterList(params.getRegionsFilter().get());
+      if (checkRegionFilter && !regionsFilter.contains(multiClusterConfigs.getRegionName())) {
+        LOGGER.info(
+            "UpdateStore command will be skipped for store: {} in cluster: {}, because the region filter is {}"
+                + " which doesn't include the current region: {}",
+            storeName,
+            clusterName,
+            regionsFilter,
+            multiClusterConfigs.getRegionName());
+        return null;
```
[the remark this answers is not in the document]
nisargthakkar: That is correct. `null` means no change. We generally prefer `null` over `Optional` as it avoids garbage and that reduces GC

services/venice-controller/src/main/java/com/linkedin/venice/controller/util/AdminUtils.java
nisargthakkar: Yeah, I'll update it

nisargthakkar:
> Final comments on the strategy about releasing the changes
> 
> 1. Do we have branch based releases for complex changes/features?


No, we don't. I think since the integration tests didn't have to change apart from fixing invalid configs, I can fix all stores to make sure that they have valid configs before we merge this PR. There is a concern that this impacts more than just LI, but I think no one else is using most of the advanced features that have added validations. Also, this won't impact any reads/writes, but control paths would get impacted.


> 2. It feels merging this w/ master and finding issues during certification can be potentially difficult to deal/mitigate as surface area of the changes is large and reverting such large commit accurately (assuming other changes pile in) might be tricky.


That is correct. With such a large change that modifies how update store is basically done, this is a risk I think is worth taking. There are lots of behavioral changes that make it impractical to be made in a way we can flip it on and off through feature flags. I don't think we have added any change that blocks a revert - so we have that option open.


> 
> How do plan to certify this w/ our current certification flow?

This will get merged in our code and will only get deployed in the certification cluster. It will not be exposed to user clusters and user stores until we certify the change

PR #1092, opened 2024-08-01
[router] Drain in-flight requests for graceful router shutdown

majisourav99:
## Drain in-flight requests for graceful router shutdown 

During router shutdown after D2 denouncement, it waits 30s to drain all in-flight requests, but it might not be enough or deterministic as there could still be in-flight requests between router<->server which would fail when the client is closed. This PR adds a wait and validates that all in-flight requests have been drained. 
## How was this PR tested?

CI

### DIFF
--- a/internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/integration/utils/VeniceRouterWrapper.java
+++ b/internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/integration/utils/VeniceRouterWrapper.java
@@ -145,7 +145,7 @@ public class VeniceRouterWrapper extends ProcessWrapper implements MetricsAware 
           .put(ROUTER_HTTPASYNCCLIENT_CONNECTION_WARMING_LOW_WATER_MARK, 1)
           .put(ROUTER_MAX_OUTGOING_CONNECTION, 10)
           // To speed up test
-          .put(ROUTER_NETTY_GRACEFUL_SHUTDOWN_PERIOD_SECONDS, 0)
+          .put(ROUTER_NETTY_GRACEFUL_SHUTDOWN_PERIOD_SECONDS, 1)
           .put(MAX_READ_CAPACITY, DEFAULT_PER_ROUTER_READ_QUOTA)
           .put(SYSTEM_SCHEMA_CLUSTER_NAME, clusterName)
           .put(ROUTER_STORAGE_NODE_CLIENT_TYPE, StorageNodeClientType.APACHE_HTTP_ASYNC_CLIENT.name())
--- a/internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/integration/utils/VeniceRouterWrapper.java
+++ b/internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/integration/utils/VeniceRouterWrapper.java
@@ -145,7 +145,7 @@ public class VeniceRouterWrapper extends ProcessWrapper implements MetricsAware 
           .put(ROUTER_HTTPASYNCCLIENT_CONNECTION_WARMING_LOW_WATER_MARK, 1)
           .put(ROUTER_MAX_OUTGOING_CONNECTION, 10)
           // To speed up test
-          .put(ROUTER_NETTY_GRACEFUL_SHUTDOWN_PERIOD_SECONDS, 10)
+          .put(ROUTER_NETTY_GRACEFUL_SHUTDOWN_PERIOD_SECONDS, 0)
           .put(MAX_READ_CAPACITY, DEFAULT_PER_ROUTER_READ_QUOTA)
           .put(SYSTEM_SCHEMA_CLUSTER_NAME, clusterName)
           .put(ROUTER_STORAGE_NODE_CLIENT_TYPE, StorageNodeClientType.APACHE_HTTP_ASYNC_CLIENT.name())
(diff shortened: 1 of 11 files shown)


services/venice-router/src/main/java/com/linkedin/venice/router/api/VeniceDispatcher.java
```diff
@@ -411,4 +411,25 @@ public void run() {
       }
     }
   }
+
+  public boolean hasInFlightRequests() {
+    AggRouterHttpRequestStats stats = routerStats.getStatsByType(RequestType.SINGLE_GET);
```
gaojieliu, line 416: By default, Tehuti is using 30s as the sliding window, which means if there is one request in the first second, the metric value won't be 0 until 30s later, which can potentially delay the Router shutdown, and today, the window is fixed.
I wondered whether we should introduce a new metric to capture the inflight requests for all kind of requests and the time-window is configurable, so that we can better control the delay duration.
What do you think?

majisourav99: We could introduce the new metric to track total in-flight for all type of requests with shorter period, but saving extra few seconds during router shutdown which are generally fast, at the cost of more frequent metric collection may not be worth it?

majisourav99: Also total extra time added due to this change is limited to 10x10s. We may actually increase it bit to say 5x30s to account for metric collection delay.

gaojieliu: Short window is not more frequent metric collections, and it is just trying to use a short window and we don't need to report this metric. 
The meta point is that we are using the existing metrics in a different way, which is not optimized for the inflight requests check purpose, and the new metric will help that and we can make the time window configuragble, and it won't affect other existing metrics.

Today, in the steady state, each Router may only take 10 seconds to shutdown, and with the new logic, it can increase the deployment time (no idea how much), and if it does increase it a lot, we don't have a way to tune it.

Router layer is normally using a large fleet.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/integration/utils/VeniceRouterWrapper.java
```diff
@@ -145,7 +145,7 @@ static StatefulServiceProvider<VeniceRouterWrapper> generateService(
           .put(ROUTER_HTTPASYNCCLIENT_CONNECTION_WARMING_LOW_WATER_MARK, 1)
           .put(ROUTER_MAX_OUTGOING_CONNECTION, 10)
           // To speed up test
-          .put(ROUTER_NETTY_GRACEFUL_SHUTDOWN_PERIOD_SECONDS, 0)
+          .put(ROUTER_NETTY_GRACEFUL_SHUTDOWN_PERIOD_SECONDS, 10)
```
gaojieliu, line 148: Will this change slowdown the existing integration tests greatly?
We don't need to do graceful shutdown in the integration test env unless the test is validating this particular feature, right? When there is a need, the test can override this config.

services/venice-router/src/main/java/com/linkedin/venice/router/stats/RouterHttpRequestStats.java
```diff
@@ -71,6 +80,8 @@ public class RouterHttpRequestStats extends AbstractVeniceHttpStats {
   private Sensor keySizeSensor;
   private final String systemStoreName;
 
+  private static final AtomicInteger totalInflightRequest = new AtomicInteger();
```
gaojieliu, line 83: Do we still need this?

services/venice-router/src/main/java/com/linkedin/venice/router/stats/RouterHttpRequestStats.java
```diff
@@ -19,10 +21,17 @@
...
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 
 
 public class RouterHttpRequestStats extends AbstractVeniceHttpStats {
+  private static final MetricConfig METRIC_CONFIG = new MetricConfig().timeWindow(10, TimeUnit.SECONDS);
+  private static final MetricsRepository localMetricRepo = new MetricsRepository(METRIC_CONFIG);
+  private final static Sensor totalInflightRequestSensor = localMetricRepo.sensor("total_inflight_request");
+  static {
+    totalInflightRequestSensor.add("total_inflight_request_count", new Max());
```
gaojieliu, line 33: I assume `Max` would work, but if we want to measure the QPS, we typically would use `Rate` instead.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/integration/utils/VeniceRouterWrapper.java
majisourav99: Sure, I can override this for the failing tests.

services/venice-router/src/main/java/com/linkedin/venice/router/stats/RouterHttpRequestStats.java
majisourav99: Good catch! I will remove it

gaojieliu (approved):

services/venice-router/src/main/java/com/linkedin/venice/router/stats/RouterHttpRequestStats.java
majisourav99: Replaced with Rate, but Max would work as well.

gaojieliu (approved):

PR #1094, opened 2024-08-02
[da-vinci][router] Fix for router related issues on DaVinci Blob-Transfer

sebas-inf:

This PR is to resolved the router related issues on the DaVinci blob transfer. Changed TransportClient to AvroGenericStoreClientImpl due to uri issues. Then changed the filter in MetaDataHandlers to filter for only completed peers instead of live peers. Lastly, added a constructor for the d2Client in RouterServer.

## How was this PR tested?

Ran the testBlobP2PTransferAmongDVC in DaVinciClientTest and github ic


internal/venice-common/src/main/java/com/linkedin/venice/blobtransfer/DaVinciBlobFinder.java
```diff
@@ -22,32 +27,46 @@
...
+  private byte[] executeRequest(String requestPath) {
+    byte[] response;
+    try {
+      response = RetryUtils.executeWithMaxAttempt(
+          () -> ((CompletableFuture<byte[]>) storeClient.getRaw(requestPath)).get(),
+          3,
+          Duration.ofSeconds(5),
+          Collections.singletonList(ExecutionException.class));
+    } catch (Exception e) {
+      throw new VeniceException("Failed to fetch schema from path " + requestPath, e);
```
adamxchen, line 63: Why do we need to throw exception related to schema?

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/DaVinciClientTest.java
```diff
@@ -1002,7 +1002,7 @@ public void testReadCompressedData(CompressionStrategy compressionStrategy) thro
    * For the local P2P testing, need to setup two different directories and ports for the two Da Vinci clients in order
    * to avoid conflicts.
    */
-  @Test(timeOut = 2 * TEST_TIMEOUT, enabled = false)
+  @Test(timeOut = 2 * TEST_TIMEOUT, enabled = true)
```
adamxchen, line 1005: This doesn't work as expected when testing on local so we need to keep it `false`

clients/venice-admin-tool/src/test/java/com/linkedin/venice/TestAdminTool.java
```diff
@@ -74,13 +74,15 @@ public void testAdminToolRequiresSSLConfigFile() {
   public void testAdminUpdateStoreArg() throws ParseException, IOException {
     final String K1 = "k1", V1 = "v1", K2 = "k2", V2 = "v2", K3 = "k3", V3 = "v3";
     String[] args = { "--update-store", "--url", "http://localhost:7036", "--cluster", "test-cluster", "--store",
-        "testStore", "--rmd-chunking-enabled", "true", "--partitioner-params",
```
sixpluszero, line 77: Is this part of the change PR? Could you update your forked repo and update again?

internal/venice-common/src/main/java/com/linkedin/venice/blobtransfer/DaVinciBlobFinder.java
sixpluszero: I think because he copied from the SchemaFetcher. Please update the exception message to our purpose.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/DaVinciClientTest.java
sixpluszero: Should we find some multi-cluster test class and move this to the test class? Let me try original test in some other cluster as I expect without cluster discovery, the raw transport client should fail on other cluster...

internal/venice-common/src/main/java/com/linkedin/venice/blobtransfer/DaVinciBlobFinder.java
```diff
@@ -22,32 +27,46 @@
...
+  private final AvroGenericStoreClientImpl storeClient;
 
-  public DvcBlobFinder(TransportClient transportClient) {
-    this.transportClient = transportClient;
+  public DaVinciBlobFinder(AvroGenericStoreClientImpl storeClient) {
+    this.storeClient = storeClient;
   }
 
   @Override
   public BlobPeersDiscoveryResponse discoverBlobPeers(String storeName, int version, int partition) {
```
sixpluszero, line 42: Can you add another logic in the caller of this discoverBlobPeers and iterate through all the available peers? And please print out the peer for better future debugging purpose. For now I think we can iterate all of them before real prod onboarding, but in the future stage, we should set a limit of how many peers we can try before we give up.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/DaVinciClientTest.java
adamxchen: That's a good point actually. We can do it

services/venice-router/src/main/java/com/linkedin/venice/router/RouterServer.java
```diff
@@ -274,7 +274,8 @@ public RouterServer(
       List<ServiceDiscoveryAnnouncer> serviceDiscoveryAnnouncers,
       Optional<DynamicAccessController> accessController,
       Optional<SSLFactory> sslFactory,
-      MetricsRepository metricsRepository) {
+      MetricsRepository metricsRepository,
+      D2Client d2Client) {
```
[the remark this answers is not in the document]
sixpluszero: I think the old one is in above lines. 
But yes, backward compatibility is important.


Null is fine as original code was broken anyways, wait but if null works for integration test you wrote, does it mean the 2nd client was not doing blob transfer as all?

sixpluszero:
The integration test G is broken, could you check the run log and fix the issue as well?

services/venice-router/src/main/java/com/linkedin/venice/router/RouterServer.java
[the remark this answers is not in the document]
adamxchen: I was checking the constructor on the production side, it's using this one.

> Null is fine as original code was broken anyways, wait but if null works for integration test you wrote, does it mean the 2nd client was not doing blob transfer as all?

I am not following actually. The main issue for the current integ test is the 2nd client can discover the 1st client but couldn't do blob transfer due to port conflict issue, but it still has a valid D2 client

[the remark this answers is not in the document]
sixpluszero: OK, thanks for the explanation, let's not work on integration test in this PR. We can come back later on this. Server is relatively easy to test p2p in integration test though.

internal/venice-common/src/main/java/com/linkedin/venice/blobtransfer/DaVinciBlobFinder.java
sixpluszero: No I mean in the caller of this method, for now after discovery, the current logic only try for the first peer it sees, however, it could be a dead instance or some other issues (for example in the future certain node hit throttling limit). Can you iterate through and try to fetch from them, once you found one working, you can quit the loop.

sixpluszero: And let's not log all the peers in the list for now. Only log the peer hostname you try in the loop.

sebas-inf: please double check if the logic i implemented would work, it passes the local tests

internal/venice-common/src/main/java/com/linkedin/venice/blobtransfer/NettyP2PBlobTransferManager.java
```diff
@@ -37,25 +41,36 @@ public void start() throws Exception {
...
-      inputStream = nettyClient.get(chosenHost, storeName, version, partition);
-    } catch (Exception e) {
-      throw new VenicePeersNotFoundException("The connection to peers failed", e);
+
+    for (String peer: discoverPeers) {
+      try {
+        // instanceName comes as a format of <hostName>_<applicationPort>
+        String chosenHost = peer.split("_")[0];
+        CompletionStage<InputStream> result = nettyClient.get(chosenHost, storeName, version, partition);
+        InputStream inputStream = result.toCompletableFuture().get();
```
sixpluszero, line 61: There is no point to wrap original inputStream with different layer of futures... Just catch the exception here, if error is thrown that means connection is broken, the node might just not exist. For other types of errors I think we can deal with them in the follow up PR.
And you are returning a different type a future, which is different from the method signature, IDK if that even compiles.

Can you just find the first host that does not fail the connection and return it? (Plus print the chosen host as well?)

sebas-inf: Did i do what you meant in this new push?

internal/venice-common/src/main/java/com/linkedin/venice/blobtransfer/NettyP2PBlobTransferManager.java
```diff
@@ -46,16 +49,19 @@ public CompletionStage<InputStream> get(String storeName, int version, int parti
...
+    for (String peer: discoverPeers) {
+      try {
+        // TODO: add some retry logic or strategy to choose the peers differently in case of failure
+        // instanceName comes as a format of <hostName>_<applicationPort>
+        String chosenHost = peer.split("_")[0];
+        inputStream = nettyClient.get(chosenHost, storeName, version, partition);
+        LOGGER.info("Chosen host: {}", chosenHost);
+        return inputStream;
+      } catch (Exception e) {
+        throw new VenicePeersNotFoundException("The connection to peers failed", e);
```
sixpluszero, line 61: No so in this case, why are you still throwing Exception? We can just log the exception with WARN, and continue, until we find out the one that actually connects. Take a look at how this nettyClient.get() is implemented, you will see how it tries to establish connection to the host and what it will do if it fails..

internal/venice-common/src/main/java/com/linkedin/venice/blobtransfer/NettyP2PBlobTransferManager.java
```diff
@@ -46,16 +49,19 @@ public CompletionStage<InputStream> get(String storeName, int version, int parti
...
+      }
     }
-
-    return inputStream;
+    throw new VenicePeersNotFoundException("Failed to obtain the blob from the peers");
```
sixpluszero, line 64: If you reach this stage, then it means all the peers are not valid, can you update the log message to reflect that?

internal/venice-common/src/main/java/com/linkedin/venice/blobtransfer/NettyP2PBlobTransferManager.java
sebas-inf: addressed in latest commit i believe

services/venice-router/src/main/java/com/linkedin/venice/router/MetaDataHandler.java
```diff
@@ -552,12 +553,44 @@ private void handleBlobDiscovery(ChannelHandlerContext ctx, VenicePathParserHelp
           Integer.parseInt(storeVersion),
           Integer.parseInt(storePartition),
           Optional.empty());
+
+      if (!instances.isEmpty()) {
+        LOGGER.info(
```
sixpluszero, line 558: This seems to be redundant logging from your debugging session. We can log when no instances are found, but when there are instances found, for production use cases, there might be thousands of replicas for certain Da Vinci use cases. Let's not log it there, or at least, just log the number of the instances.

Also, can you only log these once? Try to merge with below logging line. 
Also, nit: Please update the name to liveNodeHostNames to something like readyToServeNodeHostNames, and update the corresponding log message, as we are for now not checking whether it is alive instance, but rather checking its push readiness.

sebas-inf: please double check the logging changes

sixpluszero (dismissed):
LGTM, thank you for the fix! Next step is to really try it out in the test environment after we release this.

internal/venice-common/src/main/java/com/linkedin/venice/blobtransfer/DaVinciBlobFinder.java
```diff
@@ -22,32 +27,48 @@
...
  */
-public class DvcBlobFinder implements BlobFinder {
-  private static final Logger LOGGER = LogManager.getLogger(DvcBlobFinder.class);
+public class DaVinciBlobFinder implements BlobFinder {
+  private static final Logger LOGGER = LogManager.getLogger(DaVinciBlobFinder.class);
   private static final String TYPE_BLOB_DISCOVERY = "blob_discovery";
   private static final String ERROR_DISCOVERY_MESSAGE =
       "Error finding DVC peers for blob transfer in store: %s, version: %d, partition: %d";
-  private final TransportClient transportClient;
+  private final AvroGenericStoreClientImpl storeClient;
```
sixpluszero, line 35: Can you change the class to `AbstractAvroStoreClient` ? As we don't need specific implementation class as the variable definition, interface is enough. And in the test, you are performing a type casting which is not valid.

internal/venice-common/src/main/java/com/linkedin/venice/blobtransfer/DaVinciBlobFinder.java
```diff
@@ -22,32 +27,48 @@
...
+  private byte[] executeRequest(String requestPath) {
+    byte[] response;
+    try {
+      response = RetryUtils.executeWithMaxAttempt(
+          () -> ((CompletableFuture<byte[]>) storeClient.getRaw(requestPath)).get(),
+          3,
+          Duration.ofSeconds(5),
+          Collections.singletonList(ExecutionException.class));
+    } catch (Exception e) {
+      throw new VeniceException("Failed to fetch blob peers from path " + requestPath, e);
```
adamxchen, line 65: This would throw exception at line 44, then it need to be handled by the caller of `discoverBlobPeers`. This is different than the previous behavior where all exceptions are supposed to be caught and return a `BlobPeersDiscoveryResponse` has `Error` set.

Regardless whether we want to explicitly throw exceptions or catch them and return a POJO, I think we should make the behavior consistent in this method

internal/venice-common/src/main/java/com/linkedin/venice/blobtransfer/NettyP2PBlobTransferManager.java
```diff
@@ -46,16 +49,19 @@ public CompletionStage<InputStream> get(String storeName, int version, int parti
...
-      inputStream = nettyClient.get(chosenHost, storeName, version, partition);
-    } catch (Exception e) {
-      throw new VenicePeersNotFoundException("The connection to peers failed", e);
+    for (String peer: discoverPeers) {
+      try {
+        // TODO: add some retry logic or strategy to choose the peers differently in case of failure
+        // instanceName comes as a format of <hostName>_<applicationPort>
+        String chosenHost = peer.split("_")[0];
+        inputStream = nettyClient.get(chosenHost, storeName, version, partition);
+        LOGGER.info("Chosen host: {}", chosenHost);
```
adamxchen, line 58: how about moving it above line 57? When there's something wrong happen at line 57, the line 58 would not get executed. though eventually It would also get logged at line 61

internal/venice-common/src/main/java/com/linkedin/venice/blobtransfer/DaVinciBlobFinder.java
sebas-inf: sounds good, I have changed some of the logic in my latest commit so please let me know if it is what you meant

internal/venice-common/src/main/java/com/linkedin/venice/blobtransfer/BlobTransferUtil.java
```diff
@@ -32,10 +33,12 @@ public static BlobTransferManager<Void> getP2PBlobTransferManagerAndStart(
       String baseDir,
       ClientConfig clientConfig) {
     try {
+      AvroGenericStoreClientImpl storeClient =
```
sixpluszero, line 36: We should not casting type randomly anywhere, could you also fix this?

internal/venice-common/src/main/java/com/linkedin/venice/blobtransfer/DaVinciBlobFinder.java
sebas-inf: I have made the changes to use AbstractAvroStoreClient in my latest commit, please review it when you get a chance

internal/venice-common/src/main/java/com/linkedin/venice/blobtransfer/BlobTransferUtil.java
sebas-inf: Fixed in latest commit

internal/venice-common/src/test/java/com/linkedin/venice/blobtransfer/DaVinciBlobFinderTest.java
```diff
@@ -1,16 +1,11 @@
...
 
-import static org.mockito.ArgumentMatchers.anyString;
-import static org.mockito.Mockito.doThrow;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.spy;
-import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.when;
-import static org.testng.AssertJUnit.assertEquals;
-import static org.testng.AssertJUnit.assertTrue;
+import static org.mockito.ArgumentMatchers.*;
```
sixpluszero, line 3: Can you fix this * import style by going to the IntelliJ project structure -> code style->Java and update the line to start * import to high number?

internal/venice-common/src/main/java/com/linkedin/venice/blobtransfer/BlobTransferUtil.java
```diff
@@ -1,9 +1,12 @@
 package com.linkedin.venice.blobtransfer;
 
+import static com.linkedin.venice.client.store.ClientFactory.*;
```
sixpluszero, line 3: Also here.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/router/TestBlobDiscovery.java
```diff
@@ -7,6 +7,7 @@
 import static com.linkedin.venice.ConfigKeys.OFFLINE_JOB_START_TIMEOUT_MS;
 import static com.linkedin.venice.ConfigKeys.PERSISTENCE_TYPE;
 import static com.linkedin.venice.ConfigKeys.PUSH_STATUS_STORE_ENABLED;
+import static com.linkedin.venice.client.store.ClientFactory.*;
```
sixpluszero, line 10: Same here.

sixpluszero (commented):
This is a style related comment, it is easy to resolve. I will also DM you with the solution

sixpluszero (commented):
I took another pass. The failed test does tell a bunch of issues in this PR that we need to address. Please take a look and try to test it. The test itself is correct, and we need to make sure it pass.

sixpluszero (approved):
LGTM, thanks for the changes!

PR #1095, opened 2024-08-02
[server] Fixed ServerStoreAclHandler unable to process DICTIONARY query action

xunyin8:
## [server] Fixed ServerStoreAclHandler unable to process DICTIONARY query action
In venice server the resource for ServerStoreAclHandler is a version topic for STORAGE and COMPUTE query actions but it's still a store name if the query action is DICTIONARY. Always processing it as a version topic will result in an empty string as the resource name.

Added tests with a mock DynamicAccessController to verify basic request parsing and handling for ServerStoreAclHandler.

## How was this PR tested?
Added new unit tests

### DIFF
--- a/services/venice-server/src/main/java/com/linkedin/venice/listener/ServerStoreAclHandler.java
+++ b/services/venice-server/src/main/java/com/linkedin/venice/listener/ServerStoreAclHandler.java
@@ -63,10 +63,15 @@ public class ServerStoreAclHandler extends StoreAclHandler {
    */
   @Override
   protected String extractStoreName(String resourceName, QueryAction queryAction) {
-    if (QueryAction.DICTIONARY.equals(queryAction)) {
-      return resourceName;
-    } else {
-      return Version.parseStoreFromKafkaTopicName(resourceName);
+    switch (queryAction) {
+      case STORAGE:
+      case COMPUTE:
+        return Version.parseStoreFromKafkaTopicName(resourceName);
+      case DICTIONARY:
+        return resourceName;
+      default:
+        throw new IllegalArgumentException(
+            String.format("Unexpected QueryAction: %s with resource name: %s", queryAction, resourceName));
     }
   }
 
--- a/services/venice-server/src/main/java/com/linkedin/venice/listener/ServerStoreAclHandler.java
+++ b/services/venice-server/src/main/java/com/linkedin/venice/listener/ServerStoreAclHandler.java
@@ -58,11 +58,15 @@ public class ServerStoreAclHandler extends StoreAclHandler {
   }
 
   /**
-   * In Venice Server, the resource name is actually a Kafka topic name.
+   * In Venice Server, the resource name is actually a Kafka topic name for STORAGE/COMPUTE but store name for DICTIONARY.
    */
   @Override
   protected String extractStoreName(String resourceName) {
-    return Version.parseStoreFromKafkaTopicName(resourceName);
+    if (Version.isVersionTopic(resourceName)) {
+      return Version.parseStoreFromKafkaTopicName(resourceName);
+    } else {
+      return resourceName;
+    }
   }
 
   protected static boolean checkWhetherAccessHasAlreadyApproved(ChannelHandlerContext ctx) {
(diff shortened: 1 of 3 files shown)


services/venice-server/src/main/java/com/linkedin/venice/listener/ServerStoreAclHandler.java
```diff
@@ -58,11 +58,15 @@ public <ReqT, RespT> ServerCall.Listener<ReqT> interceptCall(
...
+   * In Venice Server, the resource name is actually a Kafka topic name for STORAGE/COMPUTE but store name for DICTIONARY.
    */
   @Override
   protected String extractStoreName(String resourceName) {
-    return Version.parseStoreFromKafkaTopicName(resourceName);
+    if (Version.isVersionTopic(resourceName)) {
+      return Version.parseStoreFromKafkaTopicName(resourceName);
+    } else {
+      return resourceName;
+    }
```
nisargthakkar, line 69: This kind of extraction is problematic since store names could also have `_vN` suffix. Can we have a check by also using the `QueryAction`? That way we would be stricter in what kind of values we allow.

I don't remember exactly where we saw this in the past, but I remember fixing this bug in another place. I'll see if I can find that place

nisargthakkar (changes requested):

services/venice-server/src/main/java/com/linkedin/venice/listener/ServerStoreAclHandler.java
nisargthakkar: It was this one: https://github.com/linkedin/venice/commit/8c8b6563688ee1583a84aa25515493130a0b3525

gaojieliu: This is a valid concern, so maybe we can pass the `QueryAction` to this function: `extractStoreName`, and for `DICTIONARY` request, it will use the passed `resourceName` as `storeName` directly.

services/venice-server/src/test/java/com/linkedin/venice/listener/ServerStoreAclHandlerTest.java
```diff
@@ -5,20 +5,125 @@
...
 
 
 public class ServerStoreAclHandlerTest {
+  private static final String TEST_STORE_NAME = "testStore";
+  private static final String TEST_STORE_VERSION = Version.composeKafkaTopic(TEST_STORE_NAME, 1);
+
+  /**
+   * Mock access controller to verify basic request parsing and handling for {@link ServerStoreAclHandler}
+   */
+  private class MockAccessController implements DynamicAccessController {
```
gaojieliu, line 46: Thanks a lot for adding this!

services/venice-server/src/main/java/com/linkedin/venice/listener/ServerStoreAclHandler.java
xunyin8: Thanks for pointing this out. Will add a test for this as well.

gaojieliu (dismissed):

services/venice-server/src/main/java/com/linkedin/venice/listener/ServerStoreAclHandler.java
```diff
@@ -58,11 +59,15 @@ public <ReqT, RespT> ServerCall.Listener<ReqT> interceptCall(
...
    */
   @Override
-  protected String extractStoreName(String resourceName) {
-    return Version.parseStoreFromKafkaTopicName(resourceName);
+  protected String extractStoreName(String resourceName, QueryAction queryAction) {
+    if (QueryAction.DICTIONARY.equals(queryAction)) {
+      return resourceName;
+    } else {
+      return Version.parseStoreFromKafkaTopicName(resourceName);
+    }
```
nisargthakkar, line 70: Should we explicitly handle `STORAGE`, `COMPUTE`, `DICTIONARY` and fail fast for the rest? That way, we'd enforce that this stays updated as the code evolves and new QueryActions are added

xunyin8: Sure, we will basically have a double safe to catch outdated code here. One is in the test I added and second will be this method throwing explicit error.

nisargthakkar (approved):
LGTM. Thanks @xunyin8

PR #1096, opened 2024-08-05
[server] Supporting 4 consumer pools for current version and isolation for AA WC leader.

haoxu07:
This PR is mainly for simplifying the logic of consumer pool assignment logic when subscribing one topic partition through `PartitionReplicaIngestionContext`.  Inside `KafkaConsumerServiceDelegator`, A `ConsumerPoolStrategyType` is defined to specify consumer pool allocation strategy implementation of `ConsumerPoolStrategy`.  
`AggKafkaConsumerService#subscribeConsumerFor` will pass `PartitionReplicaIngestionContext` with `VersionRole` and `WorkloadType` information to let `CurrentVersionConsumerPoolStrategy`  pick up the specific consumer pool.  Each instance of  `KakfaConsumerService` will be built inside `ConsumerPoolStrategy`.  

`CurrentVersionConsumerPoolStrategy` will support for 4 different consumer pools with these 4 configs:

1. `SERVER_CONSUMER_POOL_SIZE_FOR_CURRENT_VERSION_AA_WC_LEADER` 
2. `SERVER_CONSUMER_POOL_SIZE_FOR_NON_CURRENT_VERSION_AA_WC_LEADER`
3. `SERVER_CONSUMER_POOL_SIZE_FOR_CURRENT_VERSION_NON_AA_WC_LEADER`
4. `SERVER_CONSUMER_POOL_SIZE_FOR_NON_CURRENT_VERSION_NON_AA_WC_LEADER`

`DefaultConsumerPoolStrategy` will support 1 pool per region with original default setting.

`AAOrWCLeaderConsumerPoolStrategy` will support for 1 more pool with: `SERVER_DEDICATED_CONSUMER_POOL_SIZE_FOR_AA_WC_LEADER`. Here for config backward-compatibility, we will keep relying on `SERVER_DEDICATED_CONSUMER_POOL_FOR_AA_WC_LEADER_ENABLED` to turn on this feature,  `SERVER_CONSUMER_POOL_ALLOCATION_STRATEGY` is only useful when `SERVER_DEDICATED_CONSUMER_POOL_FOR_AA_WC_LEADER_ENABLED` is turned off.

After we validate the `CurrentVersionConsumerPoolStrategy`, we will try to clean up the configs and rely on `SERVER_CONSUMER_POOL_ALLOCATION_STRATEGY` to switch strategy. 

Besides, this change also enable re-subscritpion based on workload type change.

Resolves #XXX

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaConsumerServiceDelegator.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaConsumerServiceDelegator.java
@@ -2,6 +2,7 @@ package com.linkedin.davinci.kafka.consumer;
 
 import com.linkedin.davinci.config.VeniceServerConfig;
 import com.linkedin.davinci.ingestion.consumption.ConsumedDataReceiver;
+import com.linkedin.venice.exceptions.VeniceException;
 import com.linkedin.venice.kafka.protocol.KafkaMessageEnvelope;
 import com.linkedin.venice.message.KafkaKey;
 import com.linkedin.venice.pubsub.api.PubSubMessage;
@@ -10,7 +11,6 @@ import com.linkedin.venice.pubsub.api.PubSubTopicPartition;
 import com.linkedin.venice.utils.Pair;
 import com.linkedin.venice.utils.concurrent.VeniceConcurrentHashMap;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
@@ -31,7 +31,7 @@ import java.util.stream.Collectors;
  */
 public class KafkaConsumerServiceDelegator extends AbstractKafkaConsumerService {
   private final Function<String, Boolean> isAAWCStoreFunc;
-  // Map from a pair of <Version Topic, PubSubTopicPartition> to the consumer service assigned to it.
+  // Map from a pair of <Version Topic, Version topic/Real time topic partition> to the consumer service assigned to it.
   private final Map<Pair<PubSubTopic, PubSubTopicPartition>, KafkaConsumerService> vtTopicPartitionPairToConsumerService =
       new VeniceConcurrentHashMap<>();
   private final List<KafkaConsumerService> consumerServices;
@@ -62,7 +62,7 @@ public class KafkaConsumerServiceDelegator extends AbstractKafkaConsumerService 
     ConsumerPoolStrategyType consumerPoolStrategyType = serverConfig.getConsumerPoolStrategyType();
 
     // For backward compatibility, if the dedicated consumer pool for AA/WC leader is enabled, we will use it.
-    // TODO: Remove this block after new pooling strategy is verified in production environment.
+    // TODO: Remove this boolean check after new pooling strategy is verified in production environment.
     if (serverConfig.isDedicatedConsumerPoolForAAWCLeaderEnabled()) {
       this.consumerPoolStrategy = new AAOrWCLeaderConsumerPoolStrategy();
     } else {
@@ -71,7 +71,12 @@ public class KafkaConsumerServiceDelegator extends AbstractKafkaConsumerService 
           consumerPoolStrategy = new AAOrWCLeaderConsumerPoolStrategy();
           break;
         case CURRENT_VERSION_PRIORITIZATION:
-          consumerPoolStrategy = new CurrentVersionConsumerPoolStrategy();
+          if (serverConfig.isResubscriptionTriggeredByVersionIngestionContextChangeEnabled()) {
+            consumerPoolStrategy = new CurrentVersionConsumerPoolStrategy();
+          } else {
+            throw new VeniceException(
+                "Resubscription should be enabled with consumer pool strategy: " + consumerPoolStrategyType);
+          }
           break;
         default:
           consumerPoolStrategy = new DefaultConsumerPoolStrategy();
@@ -81,13 +86,12 @@ public class KafkaConsumerServiceDelegator extends AbstractKafkaConsumerService 
   }
 
   private KafkaConsumerService assignKafkaConsumerServiceFor(
-      PubSubTopic versionTopic,
       PartitionReplicaIngestionContext topicPartitionReplicaRole) {
     Pair<PubSubTopic, PubSubTopicPartition> versionTopicPartitionPair =
-        new Pair<>(versionTopic, topicPartitionReplicaRole.getPubSubTopicPartition());
+        new Pair<>(topicPartitionReplicaRole.getVersionTopic(), topicPartitionReplicaRole.getPubSubTopicPartition());
     return vtTopicPartitionPairToConsumerService.computeIfAbsent(
         versionTopicPartitionPair,
-        pair -> consumerPoolStrategy.delegateKafkaConsumerServiceFor(versionTopic, topicPartitionReplicaRole));
+        pair -> consumerPoolStrategy.delegateKafkaConsumerServiceFor(topicPartitionReplicaRole));
   }
 
   private KafkaConsumerService getKafkaConsumerService(PubSubTopic versionTopic, PubSubTopicPartition topicPartition) {
@@ -99,13 +103,30 @@ public class KafkaConsumerServiceDelegator extends AbstractKafkaConsumerService 
   public SharedKafkaConsumer getConsumerAssignedToVersionTopicPartition(
       PubSubTopic versionTopic,
       PubSubTopicPartition topicPartition) {
-    return getKafkaConsumerService(versionTopic, topicPartition)
-        .getConsumerAssignedToVersionTopicPartition(versionTopic, topicPartition);
+    KafkaConsumerService kafkaConsumerService = getKafkaConsumerService(versionTopic, topicPartition);
+    if (kafkaConsumerService != null) {
+      return kafkaConsumerService.getConsumerAssignedToVersionTopicPartition(versionTopic, topicPartition);
+    } else {
+      logger.error(
+          "No consumer service found for version topic {} and partition {} when fetch assigned consumer.",
+          versionTopic,
+          topicPartition);
+      return null;
+    }
   }
 
   @Override
-  public SharedKafkaConsumer assignConsumerFor(PubSubTopic versionTopic, PubSubTopicPartition topicPartition) {
-    return getKafkaConsumerService(versionTopic, topicPartition).assignConsumerFor(versionTopic, topicPartition);
+  public SharedKafkaConsumer assignConsumerFor(PubSubTopic versionTopic, PubSubTopicPartition pubSubTopicPartition) {
+    KafkaConsumerService kafkaConsumerService = getKafkaConsumerService(versionTopic, pubSubTopicPartition);
+    if (kafkaConsumerService != null) {
+      return kafkaConsumerService.assignConsumerFor(versionTopic, pubSubTopicPartition);
+    } else {
+      logger.error(
+          "No consumer service found for version topic {} and partition {} when assigning consumer.",
+          versionTopic,
+          pubSubTopicPartition);
+      return null;
+    }
   }
 
   @Override
@@ -121,6 +142,11 @@ public class KafkaConsumerServiceDelegator extends AbstractKafkaConsumerService 
     if (kafkaConsumerService != null) {
       kafkaConsumerService.unSubscribe(versionTopic, pubSubTopicPartition);
       vtTopicPartitionPairToConsumerService.remove(new Pair<>(versionTopic, pubSubTopicPartition));
+    } else {
+      logger.warn(
+          "No consumer service found for version topic {} and partition {} when unsubscribing.",
+          versionTopic,
+          pubSubTopicPartition);
     }
   }
 
@@ -152,8 +178,7 @@ public class KafkaConsumerServiceDelegator extends AbstractKafkaConsumerService 
       PartitionReplicaIngestionContext partitionReplicaIngestionContext,
       long lastReadOffset,
       ConsumedDataReceiver<List<PubSubMessage<KafkaKey, KafkaMessageEnvelope, Long>>> consumedDataReceiver) {
-    PubSubTopic versionTopic = consumedDataReceiver.destinationIdentifier();
-    assignKafkaConsumerServiceFor(versionTopic, partitionReplicaIngestionContext)
+    assignKafkaConsumerServiceFor(partitionReplicaIngestionContext)
         .startConsumptionIntoDataReceiver(partitionReplicaIngestionContext, lastReadOffset, consumedDataReceiver);
   }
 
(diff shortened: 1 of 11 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaConsumerServiceDelegator.java
```diff
@@ -83,39 +110,41 @@ public SharedKafkaConsumer assignConsumerFor(PubSubTopic versionTopic, PubSubTop
 
   @Override
   public void unsubscribeAll(PubSubTopic versionTopic) {
-    defaultConsumerService.unsubscribeAll(versionTopic);
-    if (consumerServiceForAAWCLeader != null) {
-      consumerServiceForAAWCLeader.unsubscribeAll(versionTopic);
-    }
+    consumerServices.forEach(kafkaConsumerService -> kafkaConsumerService.unsubscribeAll(versionTopic));
+    vtTopicPartitionPairToConsumerService.entrySet().removeIf(entry -> entry.getKey().getFirst().equals(versionTopic));
```
huangminchn, line 114: Looping over a small list of consumer services is kind of okay; but here, we are iterating through all topic partitions hosted on this node to find a small set of matching topic partition, it doesn't seem like a good data structure design. Why not just topic->consumerService mapping, and let the consumerService itself take care of the subscribe or unsubscribe action?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaConsumerServiceDelegator.java
```diff
@@ -25,9 +30,11 @@
  * that handling the writes before putting into the drainer queue is too expensive comparing to others.
  */
 public class KafkaConsumerServiceDelegator extends AbstractKafkaConsumerService {
-  private final KafkaConsumerService defaultConsumerService;
-  private final KafkaConsumerService consumerServiceForAAWCLeader;
   private final Function<String, Boolean> isAAWCStoreFunc;
+  // Map from a pair of <Version Topic, PubSubTopicPartition> to the consumer service assigned to it.
+  private final Map<Pair<PubSubTopic, PubSubTopicPartition>, KafkaConsumerService> vtTopicPartitionPairToConsumerService =
```
gaojieliu, line 35: https://github.com/linkedin/venice/blob/main/internal/venice-common/src/main/java/com/linkedin/venice/pubsub/api/PubSubTopicPartition.java#L7
`PubSubTopicPartition` has a reference to `PubSubTopic`, so do we still need to use `Pair` as key here?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaConsumerServiceDelegator.java
```diff
@@ -83,39 +110,41 @@ public SharedKafkaConsumer assignConsumerFor(PubSubTopic versionTopic, PubSubTop
...
   }
 
   @Override
   public void unSubscribe(PubSubTopic versionTopic, PubSubTopicPartition pubSubTopicPartition) {
-    getKafkaConsumerService(versionTopic, pubSubTopicPartition).unSubscribe(versionTopic, pubSubTopicPartition);
+    KafkaConsumerService kafkaConsumerService = getKafkaConsumerService(versionTopic, pubSubTopicPartition);
+    if (kafkaConsumerService != null) {
+      kafkaConsumerService.unSubscribe(versionTopic, pubSubTopicPartition);
+      vtTopicPartitionPairToConsumerService.remove(new Pair<>(versionTopic, pubSubTopicPartition));
+    }
```
gaojieliu, line 124: Shall we log a warn message for `else` condition?
It is unexpected, and this comment applies to all the related functions in this class.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaConsumerServiceDelegator.java
```diff
@@ -124,45 +153,162 @@ public void startConsumptionIntoDataReceiver(
...
+    @Override
+    public KafkaConsumerService delegateKafkaConsumerServiceFor(
+        PubSubTopic versionTopic,
+        PartitionReplicaIngestionContext topicPartitionReplicaRole) {
+      return defaultConsumerService;
     }
+
+    @Override
+    public List<KafkaConsumerService> getConsumerServices() {
+      return Collections.singletonList(defaultConsumerService);
```
gaojieliu, line 226: Shall we maintain a list as class-level field, so that we don't need to construct a new list for every call?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaConsumerServiceDelegator.java
```diff
@@ -124,45 +153,162 @@ public void startConsumptionIntoDataReceiver(
...
+        PartitionReplicaIngestionContext topicPartitionReplicaRole) {
+      PubSubTopicPartition topicPartition = topicPartitionReplicaRole.getPubSubTopicPartition();
+      if (isAAWCStoreFunc.apply(versionTopic.getName()) && topicPartition.getPubSubTopic().isRealTime()) {
+        return dedicatedConsumerService;
+      }
+      return defaultConsumerService;
+    }
+
+    @Override
+    public List<KafkaConsumerService> getConsumerServices() {
```
gaojieliu, line 251: Shall we maintain a list as class-level field, so that we don't need to construct a new list for every call?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaConsumerServiceDelegator.java
```diff
@@ -124,45 +153,162 @@ public void startConsumptionIntoDataReceiver(
...
+        PubSubTopic versionTopic,
+        PartitionReplicaIngestionContext topicPartitionReplicaRole) {
+      PartitionReplicaIngestionContext.VersionRole versionRole = topicPartitionReplicaRole.getVersionRole();
+      PartitionReplicaIngestionContext.WorkloadType workloadType = topicPartitionReplicaRole.getWorkloadType();
+      PubSubTopicPartition pubSubTopicPartition = topicPartitionReplicaRole.getPubSubTopicPartition();
+      PubSubTopic pubSubTopic = pubSubTopicPartition.getPubSubTopic();
+      if (versionRole.equals(PartitionReplicaIngestionContext.VersionRole.CURRENT)) {
+        if (workloadType.equals(PartitionReplicaIngestionContext.WorkloadType.AA_OR_WRITE_COMPUTE)
+            && pubSubTopic.isRealTime()) {
+          return consumerServiceForCurrentVersionAAWCLeader;
```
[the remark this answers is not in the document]
gaojieliu: `leader` is one of the magic term to differentiate different types of workload, another magic term is `current_version`.

gaojieliu (changes requested):
My change has a quite number of conflicts with this PR:
https://github.com/linkedin/venice/pull/1102
I will let you merge first as I might need to add throttler for each consumer pool.

Can we add a check to make sure the version jumping feature will be enabled together with the 4-pool strategy?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaConsumerServiceDelegator.java
haoxu07: Oh, the pair here is mainly used for <version topic, version/real-time topic partition>, the reason is version topic could be different from real-time topic.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaConsumerServiceDelegator.java
```diff
@@ -7,12 +7,17 @@
 import com.linkedin.venice.pubsub.api.PubSubMessage;
 import com.linkedin.venice.pubsub.api.PubSubTopic;
 import com.linkedin.venice.pubsub.api.PubSubTopicPartition;
+import com.linkedin.venice.utils.Pair;
```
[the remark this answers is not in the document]
haoxu07: Yeah, I am also thinking about getting rid of this. Not decided which one to go, any suggestions?

haoxu07:
> My change has a quite number of conflicts with this PR: #1102 I will let you merge first as I might need to add throttler for each consumer pool.
> 
> Can we add a check to make sure the version jumping feature will be enabled together with the 4-pool strategy?

Yeah, I also see your change. Definitely I could add check for the version jumping feature will be enabled together with the 4-pool strategy.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaConsumerServiceDelegator.java
haoxu07: You are right, better to have one for it.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaConsumerServiceDelegator.java
gaojieliu: Won't the referenced `PubSubTopic` contain the full name, so that we can differentiate whether it is a real-time topic or version topic?

haoxu07: `PubSubTopic` inside `PubSubTopicPartition` could be real-time or version topic, the problem here is: our server host may ingest 2+ same real-time topic partitions from 2+ version topics (current/backup/future). As version topic + pubsubtopicpartition could be unique for that server host.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaConsumerServiceDelegator.java
haoxu07: topic->consumerService is not correct here, as one version topic could interact with more than 1 consumer services, due to topic partition is leader or follower.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaConsumerServiceDelegator.java
gaojieliu: Got it, make sense.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaConsumerServiceDelegator.java
gaojieliu: I think @huangminchn 's comment was referring to this field: `vtTopicPartitionPairToConsumerService`.
Right now, I am not sure how important this inefficiency is as this event should be less frequent at runtime, and it will be triggered more frequently when stopping, but as different store ingestion task will trigger this function concurrently, maybe it can be fine as this function doesn't use any synchronization logic.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/integration/utils/VeniceServerWrapper.java
```diff
@@ -253,7 +255,10 @@ static StatefulServiceProvider<VeniceServerWrapper> generateService(
               pubSubBrokerWrapper.getPubSubClientsFactory().getAdminAdapterFactory().getClass().getName())
           .put(SERVER_INGESTION_HEARTBEAT_INTERVAL_MS, 5000)
           .put(SERVER_LEADER_COMPLETE_STATE_CHECK_IN_FOLLOWER_VALID_INTERVAL_MS, 5000)
-          .put(SERVER_RESUBSCRIPTION_TRIGGERED_BY_VERSION_INGESTION_CONTEXT_CHANGE_ENABLED, true);
+          .put(SERVER_RESUBSCRIPTION_TRIGGERED_BY_VERSION_INGESTION_CONTEXT_CHANGE_ENABLED, true)
+          .put(
```
gaojieliu, line 259: This will change every integration test to use the new pool strategy, while production is still using the old one.
So shall we only override in one of the integration test cases instead of all.
After we fully ramp the new pool strategy, we can switch the default.

gaojieliu (commented):
Left a few more minor comments.
The code change looks good overall.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaConsumerServiceDelegator.java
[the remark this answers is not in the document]
adamxchen: @haoxu07  can we add some documentations around `PartitionReplicaIngestionContext`? It'd create some confusion in the future.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaConsumerServiceDelegator.java
```diff
@@ -40,129 +47,319 @@ public class KafkaConsumerServiceDelegator extends AbstractKafkaConsumerService
...
   public KafkaConsumerServiceDelegator(
       VeniceServerConfig serverConfig,
       BiFunction<Integer, String, KafkaConsumerService> consumerServiceConstructor,
       Function<String, Boolean> isAAWCStoreFunc) {
 
-    this.defaultConsumerService =
-        consumerServiceConstructor.apply(serverConfig.getConsumerPoolSizePerKafkaCluster(), ""); // Empty stats suffix
+    this.serverConfig = serverConfig;
+    this.consumerServiceConstructor = consumerServiceConstructor;
+    this.isAAWCStoreFunc = vt -> storeVersionAAWCFlagMap.computeIfAbsent(vt, ignored -> isAAWCStoreFunc.apply(vt));
```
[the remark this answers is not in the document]
haoxu07: I decide to add it in my current change.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaConsumerServiceDelegator.java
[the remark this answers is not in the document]
haoxu07: I decide to introduce a new term of `TopicPartitionForIngestion` wrapping version topic and `PubSubTopicPartition` to replace deprecated the pair used here, and also plan to modify  all KafkaConsumerService interface to adopt it.

gaojieliu (approved):
Do we have a test case, which will go through these changes, so that we can observe the consumer pool change?
```
future version -> current version
current version -> backup version
non-write-compute -> write-compute
write-compute -> non-write-compute
```

haoxu07:
> The code change looks good to me. Do we have a test case, which will go through these changes, so that we can observe the consumer pool change?
> 
> ```
> future version -> current version
> current version -> backup version
> non-write-compute -> write-compute
> write-compute -> non-write-compute
> ```
We had unit test for role change triggering re-subscription and, will add an integration test in next PR for verifying these change.

PR #1099, opened 2024-08-05
[vpj][server][samza][admin-tool][controller] Added parent controller discovery through multiple D2 clients

bonytoni:
## Added parent controller discovery through multiple D2 clients

- `D2ControllerClient` - In future iterations, there could be local ZooKeeper instances in each region so we want to add support to iterate through all D2Clients to find the leader parent controller.
  - Added new constructors to accept `List<D2Client> d2Clients` or `List<String> d2ZkHosts`
  - Updated `discoverLeaderController` and `discoverCluster` to iterate through the list of `D2Client`
- Tested constructors and implementation of methods above in `TestD2ControllerClient`

## How was this PR tested?

Unit Tests

### DIFF
--- a/internal/venice-common/src/main/java/com/linkedin/venice/controllerapi/D2ControllerClient.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/controllerapi/D2ControllerClient.java
@@ -13,6 +13,7 @@ import java.nio.charset.StandardCharsets;
 import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
+import java.util.stream.Collectors;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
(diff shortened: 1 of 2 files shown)


internal/venice-common/src/main/java/com/linkedin/venice/controllerapi/D2ControllerClient.java
```diff
@@ -57,10 +62,23 @@ public D2ControllerClient(
     this.sslFactory = sslFactory;
   }
 
+  /**
+   * Discover leader controller from a D2 client.
+   * @return leader controller url
+   */
   @Override
   protected String discoverLeaderController() {
+    List<D2Client> d2Client = Collections.singletonList(this.d2Client);
```
nisargthakkar, line 71: Let's create the object once in the constructor and use it every time we need to. Creating a new List every time is not efficient.

internal/venice-common/src/main/java/com/linkedin/venice/controllerapi/D2ControllerClient.java
```diff
@@ -96,24 +114,27 @@ protected String discoverLeaderController() {
...
-      return ObjectMapperFactory.getInstance().readValue(responseBody, responseClass);
-    } catch (Exception e) {
-      throw new VeniceException("Failed to get response for url: " + requestPath + " with D2 Client", e);
+    for (D2Client d2Client: d2ClientList) {
+      try {
+        RestResponse response = D2ClientUtils.sendD2GetRequest(requestPath, d2Client);
+        String responseBody = response.getEntity().asString(StandardCharsets.UTF_8);
+        return ObjectMapperFactory.getInstance().readValue(responseBody, responseClass);
+      } catch (Exception e) {
+        LOGGER.error("Failed to get response for url: " + requestPath + " with D2 Client:" + d2Client, e);
```
nisargthakkar, line 129: Nit: Add a space after `Client:`

internal/venice-common/src/main/java/com/linkedin/venice/controllerapi/D2ControllerClient.java
```diff
@@ -96,24 +114,27 @@ protected String discoverLeaderController() {
...
+      }
     }
+    throw new VeniceException("List of D2 Clients failed to get response for url: " + requestPath);
```
nisargthakkar, line 132: Let's change the error message to the previous one. We don't need to call out if it was a List of D2 clients or a single D2 client

internal/venice-common/src/main/java/com/linkedin/venice/controllerapi/D2ControllerClient.java
```diff
@@ -96,24 +114,27 @@ protected String discoverLeaderController() {
...
   }
 
   public static D2ServiceDiscoveryResponse discoverCluster(D2Client d2Client, String d2ServiceName, String storeName) {
```
nisargthakkar, line 135: `discoverCluster` will also need a method to support a list of D2 clients. A Leader controller is a property of a cluster, and the call to fetch the leader controller info needs to be preceded by a cluster discovery call. Only adding resiliency to the leader controller call is not sufficient to provide the resiliency

nisargthakkar (approved):

PR #1101, opened 2024-08-07
[server] Use chunk manifest schema id for writer when performing updates to chunked records

ZacAttack:
## [server] Use supersetschema id for readerschema when performing updates

There is an edge case when a chunk is assembled from batch push that the result of the assembling doesn't contain the schema id in the first few bytes of the resulting record (there's even a bit of code in the ingestion task today which adds it back later in the processing pipeline).  This means that the old code parsing the schemaid in this case would get a corrupted schema id.  Looking at the code, in this path the reader schema is always the latest superset schema id anyway, and should therefore be safe to utilize for deserialization of the record (no field drops should should occur).  This has the advantage of requiring not much change, though it might be more technically correct to make passing the original schema id more explicit in the code?

Unsure what might be more preferred.

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
@@ -475,7 +475,7 @@ public class ActiveActiveStoreIngestionTask extends LeaderFollowerStoreIngestion
 
       case UPDATE:
         mergeConflictResult = mergeConflictResolver.update(
-            oldValueByteBufferProvider.get(),
+            oldValueByteBufferProvider,
             rmdWithValueSchemaID,
             ((Update) kafkaValue.payloadUnion).updateValue,
             incomingValueSchemaId,
@@ -484,7 +484,7 @@ public class ActiveActiveStoreIngestionTask extends LeaderFollowerStoreIngestion
             sourceOffset,
             kafkaClusterId,
             kafkaClusterId,
-            valueManifestContainer.getManifest());
+            valueManifestContainer);
         getHostLevelIngestionStats()
             .recordIngestionActiveActiveUpdateLatency(LatencyUtils.getElapsedTimeFromNSToMS(beforeDCRTimestampInNs));
         break;
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
@@ -474,9 +474,8 @@ public class ActiveActiveStoreIngestionTask extends LeaderFollowerStoreIngestion
         break;
 
       case UPDATE:
-        int oldValueSchemaId = oldValueProvider.get().writerSchemaId();
         mergeConflictResult = mergeConflictResolver.update(
-            oldValueByteBufferProvider,
+            oldValueByteBufferProvider.get(),
             rmdWithValueSchemaID,
             ((Update) kafkaValue.payloadUnion).updateValue,
             incomingValueSchemaId,
@@ -485,7 +484,7 @@ public class ActiveActiveStoreIngestionTask extends LeaderFollowerStoreIngestion
             sourceOffset,
             kafkaClusterId,
             kafkaClusterId,
-            oldValueSchemaId);
+            valueManifestContainer.getManifest());
         getHostLevelIngestionStats()
             .recordIngestionActiveActiveUpdateLatency(LatencyUtils.getElapsedTimeFromNSToMS(beforeDCRTimestampInNs));
         break;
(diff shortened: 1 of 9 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/replication/merge/MergeConflictResolver.java
```diff
@@ -617,32 +616,35 @@ private GenericRecord deserializeWriteComputeBytes(
...
       if (oldValueBytes == null) {
         // Value and RMD both never existed
-        newValue = AvroSchemaUtils.createGenericRecord(readerValueSchemaEntry.getSchema());
+        newValue = AvroSchemaUtils.createGenericRecord(superSetSchemaSchemaEntry.getSchema());
       } else {
         /**
-         * RMD does not exist. This means the value is written in Batch phase and does not have RMD associated. In this
-         * case, the value must be retrieved from storage engine, and is prepended with schema ID.
+         * RMD does not exist. This means the value is written in Batch phase and does not have RMD associated. Records
+         * should have the schema id in the first few bytes unless they are assembled from a chunked value. In order
```
FelixGV, line 629: So is it the case that `oldValueBytes` would be the fully assembled payload, in case of chunking? So this would never be a manifest nor a chunk?

It is concerning if the schema ID can either be present or not at the start of the `oldValueBytes`... because we need to either treat the payload as going from 0 to len -1, or from 4 to len -1, depending on the case, right?

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/PartialUpdateTest.java
```diff
@@ -491,6 +492,73 @@ public void testPartialUpdateOnBatchPushedKeys(CompressionStrategy compressionSt
     }
   }
 
+  @Test(timeOut = TEST_TIMEOUT_MS)
+  public void testPartialUpdateOnBatchPushedChunkKeys() throws IOException {
```
FelixGV, line 496: Does this repro the issue, in the absence of the main code fix?

FelixGV (commented):
A couple of questions...

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/PartialUpdateTest.java
ZacAttack: Yes

clients/da-vinci-client/src/main/java/com/linkedin/davinci/replication/merge/MergeConflictResolver.java
ZacAttack: Right.  It is never a manifest or chunk.  It's the real McCoy

sixpluszero: oldValueBytes are already retrieved via chunking adapter. It is just it does not come with schema id prepended afaik. Having it prepended is another way of fixing it, but if superset schema works then it is even simpler.

sixpluszero: Actually, I did a test, deserialization with superset schema being writer schema will fail..

ZacAttack: tweaked it to instead pass along the schemaid from the chunk manifest.

sixpluszero (approved):
LGTM. Yeah I didn't realize we have oldValueManifest ready for use :) 

Oops you have spot bug issue, but should be easy to fix...

sixpluszero (dismissed):

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
```diff
@@ -475,15 +475,16 @@ protected void processMessageAndMaybeProduceToKafka(
 
       case UPDATE:
         mergeConflictResult = mergeConflictResolver.update(
-            oldValueByteBufferProvider,
+            oldValueByteBufferProvider.get(),
```
sixpluszero, line 478: Wait, actually I realize this after I 🚢 . Why do we need to remove the lazy here? UPDATE might not need to be applied if RMD indicates it is a fully stale UPDATE?

ZacAttack: Good point.  I removed the lazy because invoking the lazy is the only way to populate the chunk manifest with a schema id.

Relooking at the code, I think the way to avoid this would be to pass along the ChunkedValueManifestcontainer entirely (instead of calling getManifest).  I'll make that tweak.

ZacAttack: Added it back.  Now everything works.

sixpluszero (approved):

PR #1102, opened 2024-08-08
[server] Added a throttler for AAWC workload of leader replica

gaojieliu:
We know AA/WC workload is expensive and in the past, we were trying to limit the consumer pool for AA/WC leader to control the resource usage, but sometimes, it is not very accurate. This code change introduces a new throttler for Leader AA/WC workload, and it will only take effect when the dedicated consumer pool for AA/WC leader feature is enabled.
New config:
server.aa.wc.leader.quota.records.per.second : default (-1)

## How was this PR tested?
CI


clients/da-vinci-client/src/main/java/com/linkedin/davinci/config/VeniceServerConfig.java
```diff
@@ -806,6 +808,7 @@ public VeniceServerConfig(VeniceProperties serverProperties, Map<String, Map<Str
           DEFAULT_MAX_RECORD_SIZE_BYTES + ": " + defaultMaxRecordSizeBytes + " must be at least "
               + generateHumanReadableByteCountString(BYTES_PER_MB));
     }
+    aaWCLeaderQuotaRecordsPerSecond = serverProperties.getInt(SERVER_AA_WC_LEADER_QUOTA_RECORDS_PER_SECOND, -1);
```
adamxchen, line 811: nit: regarding the naming, I think the `leader_throttle_records_per_second` sounds better

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaConsumerServiceDelegator.java
```diff
@@ -319,19 +324,23 @@ public class CurrentVersionConsumerPoolStrategy extends ConsumerPoolStrategy {
     public CurrentVersionConsumerPoolStrategy() {
       this.consumerServiceForCurrentVersionAAWCLeader = consumerServiceConstructor.apply(
           serverConfig.getConsumerPoolSizeForCurrentVersionAAWCLeader(),
-          CURRENT_VERSION_AAWC_LEADER_STATS_SUFFIX);
+          CURRENT_VERSION_AAWC_LEADER_STATS_SUFFIX,
```
haoxu07, line 327: Maybe we could use this ConsumerPoolType as suffix, I also planned to use it for consumption task thread name and task id str in my next PR, what do you think?

gaojieliu: yeah, we can do that. Or maybe just put the `suffix` definition into the enum definition.

haoxu07: We could use the type to replace suffix, essentially we just want to pass the type information to KCS.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/IngestionThrottler.java
```diff
@@ -69,6 +74,37 @@ public IngestionThrottler(
         false,
         EventThrottler.BLOCK_STRATEGY);
 
+    this.aaWCLeaderRecordThrottler = new EventThrottler(
```
haoxu07, line 77: Besides, wondering if it is possible that we could lazily instantiate the per-pool throttler inside this class when creating KCS, and then KCS could have `IntConsumer poolTypeThrottlerFunc` passing ingestionThrottler.maybeThrottleRecordRate(ConsumerPoolType poolType, int recordCount) to `ConsumptionTask`.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ConsumerPoolType.java
```diff
@@ -0,0 +1,16 @@
+package com.linkedin.davinci.kafka.consumer;
+
+public enum ConsumerPoolType {
+  /**
+   * The following pool types are being used when the dedicated AA/WC leader consumer pool feature is enabled.
+   */
+  AA_WC_LEADER_POOL, // For AA/WC leader
+  REGULAR_POOL, // For other kinds of workload except AA/WC leader
```
haoxu07, line 8: Nits about comment: seems REGULAR_POOL will be used for both DEFAULT and  AA_OR_WC_LEADER_DEDICATED strategy.

haoxu07 (commented):
Just some comments about code structures. Having one general question (might be not related to this change): we will try to use records quota and consumer thread number per pool to tune the performance of certain traffic, but wondering is it reasonable to define per consumer thread capacity/quota?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaConsumerServiceDelegator.java
gaojieliu: We typically don't use the capital letters in the metric and I think we should follow the convention if possible.

gaojieliu:
```
Just some comments about code structures. Having one general question (might be not related to this change): we will try to use records quota and consumer thread number per pool to tune the performance of certain traffic, but wondering is it reasonable to define per consumer thread capacity/quota?
```

No, we don't want to have too tight control of the throttling and per-consumer-thread-quota will be very hard to tune.
Also even we define a throttler per pool, we don't necessarily to config a valid value for each of them, and originally, we only want to control the AA/WC workload to contain the cpu usage and for other workload, we potentially would just use the global throttlers.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/IngestionThrottler.java
gaojieliu: Honestly, I don't think such optimization is required from perf POV and I would like to keep the code structure simple, such as keeping all the throttlers inside one class instead of spreading them in different classes.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ConsumerPoolType.java
gaojieliu: Sure

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaConsumerServiceDelegator.java
haoxu07: I think you most-recent update is what I meant.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/IngestionThrottler.java
haoxu07: Oh, this is not for perf POV, but for code simplicity. I mean we could have a general `ingestionThrottler.maybeThrottleRecordRate(ConsumerPoolType poolType, int recordCount)` to trigger per-type throttler creation inside `IngestionThtrottler` and throttle action.  So KCS constructor does not need to check different type to create specific `IntConsumer poolTypeThrottlerFunc` function. All throttles are still inside `IngestionThtrottler`.

gaojieliu: Sure, I can make the change to move the mapping logic from KCS to `IngestionThrottler`.

haoxu07 (approved):

PR #1104, opened 2024-08-08
[dvc][fc][tc] Lazily register client-side metrics 

kvargha:
## Summary

Unused metrics result in excessive WARN log messages. By registering metrics only when they are recorded, this issue is avoided.

## How was this PR tested?

CI

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/IngestionStats.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/IngestionStats.java
@@ -7,6 +7,7 @@ import com.linkedin.davinci.kafka.consumer.PartitionConsumptionState;
 import com.linkedin.davinci.kafka.consumer.StoreIngestionTask;
 import com.linkedin.venice.stats.LongAdderRateGauge;
 import com.linkedin.venice.utils.RegionUtils;
+import com.linkedin.venice.utils.lazy.Lazy;
 import io.tehuti.metrics.MetricConfig;
 import io.tehuti.metrics.MetricsRepository;
 import io.tehuti.metrics.Sensor;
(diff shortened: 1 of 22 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/IngestionStats.java
```diff
@@ -109,14 +110,14 @@ public class IngestionStats {
 
   /** Record a version-level offset rewind events for VTs across all stores. */
   private final Count versionTopicEndOffsetRewindCount = new Count();
```
ZacAttack, line 112: Do we need to do this one too?

ZacAttack:
Is there a rhyme or reason to which metrics we did this fix for?  I think it's non exhaustive?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/IngestionStats.java
```diff
@@ -207,6 +211,17 @@ public IngestionStats(VeniceServerConfig serverConfig) {
     registerSensor(localMetricRepository, OFFSET_REGRESSION_DCR_ERROR, offsetRegressionDCRErrorSensor);
     registerSensor(localMetricRepository, TOMBSTONE_CREATION_DCR, tombstoneCreationDCRSensor);
     registerSensor(localMetricRepository, IDLE_TIME, idleTimeSensor);
+
+    transformerLatencySensor = new WritePathLatencySensor(localMetricRepository, METRIC_CONFIG, TRANSFORMER_LATENCY);
```
ZacAttack, line 215: I'm confused here.  Why are some of these lazy and some not?  When do these get registered?

kvargha: Some metrics are required to be registered by some unit/integration tests by asserting that their value is 0. Should I modify the tests to assert that the sensor be null instead?

The underlying sensor inside `WritePathLatencySensor` is registered as lazy, but when a record is called on it the sensor will be registered.

ZacAttack: It's ok to edit tests if the behavior no longer matches up.  I think we should be uniform here unless theres a valid reason to not do so.

kvargha:
Closing due to OTEL migration

PR #1105, opened 2024-08-08
[da-vinci] Open RocksDB after blob transfer is completed

sebas-inf:

When doing blob transfer the files passed in would have already been somewhat initialized instead of it being a clean file. This fix reorganizes the DefaultIngestionBackend file to ensure that blob transfer will be tried first with a clean file.

## How was this PR tested?

Local build and QEI testing 

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/DefaultIngestionBackend.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/DefaultIngestionBackend.java
@@ -61,18 +61,22 @@ public class DefaultIngestionBackend implements IngestionBackend {
         Utils.waitStoreVersionOrThrow(storeVersion, getStoreIngestionService().getMetadataRepo());
     Supplier<StoreVersionState> svsSupplier = () -> storageMetadataService.getStoreVersionState(storeVersion);
     syncStoreVersionConfig(storeAndVersion.getFirst(), storeConfig);
-    AbstractStorageEngine storageEngine = storageService.openStoreForNewPartition(storeConfig, partition, svsSupplier);
-    topicStorageEngineReferenceMap.compute(storeVersion, (key, storageEngineAtomicReference) -> {
-      if (storageEngineAtomicReference != null) {
-        storageEngineAtomicReference.set(storageEngine);
-      }
-      return storageEngineAtomicReference;
-    });
 
+    long startTime = System.currentTimeMillis();
     CompletionStage<Void> bootstrapFuture =
         bootstrapFromBlobs(storeAndVersion.getFirst(), storeAndVersion.getSecond().getNumber(), partition);
 
     bootstrapFuture.whenComplete((result, throwable) -> {
+      long endTime = System.currentTimeMillis();
+      LOGGER.info("Bootstrapping duration: {} ms", endTime - startTime);
+      AbstractStorageEngine storageEngine =
+          storageService.openStoreForNewPartition(storeConfig, partition, svsSupplier);
+      topicStorageEngineReferenceMap.compute(storeVersion, (key, storageEngineAtomicReference) -> {
+        if (storageEngineAtomicReference != null) {
+          storageEngineAtomicReference.set(storageEngine);
+        }
+        return storageEngineAtomicReference;
+      });
       LOGGER.info(
           "Retrieved storage engine for store {} partition {}. Starting consumption in ingestion service",
           storeVersion,
@@ -83,6 +87,7 @@ public class DefaultIngestionBackend implements IngestionBackend {
           storeVersion,
           partition);
     });
+
   }
 
   /**
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/DefaultIngestionBackend.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/DefaultIngestionBackend.java
@@ -68,7 +68,11 @@ public class DefaultIngestionBackend implements IngestionBackend {
 
     bootstrapFuture.whenComplete((result, throwable) -> {
       long endTime = System.currentTimeMillis();
-      LOGGER.info("Bootstrapping duration: {} ms", endTime - startTime);
+      LOGGER.info(
+          "Blob Transfer duration for store name: {} version: {}: time: {} ms",
+          storeVersion,
+          storeAndVersion.getSecond().getNumber(),
+          endTime - startTime);
       AbstractStorageEngine storageEngine =
           storageService.openStoreForNewPartition(storeConfig, partition, svsSupplier);
       topicStorageEngineReferenceMap.compute(storeVersion, (key, storageEngineAtomicReference) -> {
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/DefaultIngestionBackend.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/DefaultIngestionBackend.java
@@ -59,20 +59,9 @@ public class DefaultIngestionBackend implements IngestionBackend {
     LOGGER.info("Retrieving storage engine for store {} partition {}", storeVersion, partition);
     Pair<Store, Version> storeAndVersion =
         Utils.waitStoreVersionOrThrow(storeVersion, getStoreIngestionService().getMetadataRepo());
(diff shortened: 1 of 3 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/DefaultIngestionBackend.java
```diff
@@ -61,18 +61,22 @@ public void startConsumption(VeniceStoreVersionConfig storeConfig, int partition
...
-      return storageEngineAtomicReference;
-    });
 
+    long startTime = System.currentTimeMillis();
     CompletionStage<Void> bootstrapFuture =
         bootstrapFromBlobs(storeAndVersion.getFirst(), storeAndVersion.getSecond().getNumber(), partition);
 
     bootstrapFuture.whenComplete((result, throwable) -> {
+      long endTime = System.currentTimeMillis();
+      LOGGER.info("Bootstrapping duration: {} ms", endTime - startTime);
```
adamxchen, line 71: If we are going to log the blob transfer time, can we also add the store name and version too otherwise it's hard to differentiate

adamxchen: Also, let's call it blob transfer duration

internal/venice-common/src/main/java/com/linkedin/venice/blobtransfer/client/P2PFileTransferClientHandler.java
```diff
@@ -72,9 +72,11 @@ protected void channelRead0(ChannelHandlerContext ctx, HttpObject msg) throws Ex
 
       // Parse the file name
       this.fileName = getFileNameFromHeader(response);
+
       if (this.fileName == null) {
         throw new VeniceException("No file name specified in the response for " + payload.getFullResourceName());
       }
+      LOGGER.info("Starting blob transfer for file: {}", fileName);
```
adamxchen, line 79: Can we make this `DEBUG`? It'd be a lot of logs for a single partition.

internal/venice-common/src/main/java/com/linkedin/venice/blobtransfer/client/P2PFileTransferClientHandler.java
```diff
@@ -117,7 +119,7 @@ protected void channelRead0(ChannelHandlerContext ctx, HttpObject msg) throws Ex
 
       if (content instanceof DefaultLastHttpContent) {
         // End of a single file transfer
-        LOGGER.debug("A file {} received successfully for {}", fileName, payload.getFullResourceName());
+        LOGGER.info("A file {} received successfully for {}", fileName, payload.getFullResourceName());
```
adamxchen, line 122: I'd like to keep it debug for the same reason above.

internal/venice-common/src/main/java/com/linkedin/venice/blobtransfer/NettyP2PBlobTransferManager.java
```diff
@@ -54,7 +54,6 @@ public CompletionStage<InputStream> get(String storeName, int version, int parti
         // TODO: add some retry logic or strategy to choose the peers differently in case of failure
         // instanceName comes as a format of <hostName>_<applicationPort>
         String chosenHost = peer.split("_")[0];
-        LOGGER.info("Chosen host: {}", chosenHost);
```
adamxchen, line 57: Why do we need to remove it? Can we make it debug if that's too much?

internal/venice-common/src/main/java/com/linkedin/venice/blobtransfer/client/P2PFileTransferClientHandler.java
sixpluszero: But actually I am feeling it is pretty useful for debugging purpose...I am not sure if this matters or not... given that Da Vinci only usually subscribe to a few partitions?

internal/venice-common/src/main/java/com/linkedin/venice/blobtransfer/NettyP2PBlobTransferManager.java
sebas-inf: I removed it on accident when cleaning up my debugging logs, bringing it back

clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/DefaultIngestionBackend.java
sebas-inf: will do

sixpluszero:
Can you update the PR title to describe the real purpose of this change? This "reorganize" stuff does not really make sense to others for your real PR intention...

internal/venice-common/src/main/java/com/linkedin/venice/blobtransfer/client/P2PFileTransferClientHandler.java
sixpluszero: Or maybe once we make sure this is working in local QEI deploy, we can for now still make it as DEBUG.
Though I still think the log volume won't be too many, like ~20 per replica? Or maybe if it large hybrid store it can have a lot prior to compaction...

adamxchen: > Or maybe once we make sure this is working in local QEI deploy, we can for now still make it as DEBUG

I'm cool with that. Right now, using INFO seems to be a lot easier for you to test it out. We can make it DEBUG afterwards

sebas-inf: should i put as as info or debug? as soon as I know I can push the new commit addressing the comments

internal/venice-common/src/main/java/com/linkedin/venice/blobtransfer/client/P2PFileTransferClientHandler.java
sebas-inf: back to debug in the latest iteration

internal/venice-common/src/main/java/com/linkedin/venice/blobtransfer/client/P2PFileTransferClientHandler.java
sixpluszero: Yeah, for now logging means everything. We can take another pass when we are really pushing it toward production. The next step is the performance...which I have some concern but let's commit it fix and do the test again in VSC.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/DefaultIngestionBackend.java
```diff
@@ -61,18 +61,26 @@ public void startConsumption(VeniceStoreVersionConfig storeConfig, int partition
...
-      return storageEngineAtomicReference;
-    });
 
+    long startTime = System.currentTimeMillis();
     CompletionStage<Void> bootstrapFuture =
         bootstrapFromBlobs(storeAndVersion.getFirst(), storeAndVersion.getSecond().getNumber(), partition);
 
     bootstrapFuture.whenComplete((result, throwable) -> {
+      long endTime = System.currentTimeMillis();
+      LOGGER.info(
```
sixpluszero, line 71: OK I know why this happens basically for the II + memory limiter integration test...

With current change, the call will happen within an async thread, when the error in memory limiter is thrown, the error is not caught but the push job is stuck, so the test eventually times out.

I think these two features are not major Da Vinci features and are not related to your feature.
In order to help you to unblock, I think this works:

use a Runnable to capture all the action we want to do here inside the whenComplete.
Move the blob transfer check:
 if (!storeAndVersion.getFirst().isBlobTransferEnabled() || storeAndVersion.getFirst().isHybrid() || blobTransferManager == null) 
from the below method to here,
when not doing blob transfer, directly call runnable.run(), otherwise, call runnable.run() in the whenComplete part.

It is not a prefect solution, but I think it is good enough to unblock here....

sebas-inf: launched in latest commit, please review it when you get a chance

sixpluszero (approved):

PR #1108, opened 2024-08-09
[fast-client] Added fanout size metric for FC to monitor multi-key HAR performance

xunyin8:
## Summary
[fast-client] Added fanout size metric for FC to monitor multi-key HAR performance

Added `multi_key_fanout_size` Avg and Max metric so we can monitor FC multi-key request fanout size
for HAR.

## How was this PR tested?
Existing and new integration test

### DIFF
--- a/clients/venice-client/src/main/java/com/linkedin/venice/fastclient/MultiKeyRequestContext.java
+++ b/clients/venice-client/src/main/java/com/linkedin/venice/fastclient/MultiKeyRequestContext.java
@@ -44,6 +44,8 @@ public abstract class MultiKeyRequestContext<K, V> extends RequestContext {
   final boolean isPartialSuccessAllowed;
   private boolean completed;
 
+  private int fanoutSize;
+
   MultiKeyRequestContext(int numKeysInRequest, boolean isPartialSuccessAllowed) {
     this.routeRequests = new VeniceConcurrentHashMap<>();
     this.firstRequestSentTS = new AtomicLong(-1);
@@ -165,6 +167,14 @@ public abstract class MultiKeyRequestContext<K, V> extends RequestContext {
 
   public void setRoutesForPartitionMapping(Map<Integer, Set<String>> routesForPartition) {
     this.routesForPartition = routesForPartition;
+  }
+
+  public void setFanoutSize(int fanoutSize) {
+    this.fanoutSize = fanoutSize;
+  }
+
+  public int getFanoutSize() {
+    return fanoutSize;
   }
 
   /**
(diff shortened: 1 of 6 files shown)


clients/venice-client/src/main/java/com/linkedin/venice/fastclient/StatsAvroGenericStoreClient.java
```diff
@@ -176,11 +176,13 @@ private <R> CompletableFuture<R> recordRequestMetrics(
       } else if (requestContext instanceof MultiKeyRequestContext) {
         // MultiKeyRequestContext is the superclass for ComputeRequestContext and BatchGetRequestContext
         MultiKeyRequestContext<K, V> multiKeyRequestContext = (MultiKeyRequestContext<K, V>) requestContext;
+        clientStats.recordMultiKeyFanoutSize(multiKeyRequestContext.getFanoutSize());
         if (multiKeyRequestContext.retryContext != null
             && multiKeyRequestContext.retryContext.retryRequestContext != null) {
           clientStats.recordLongTailRetryRequest();
           clientStats
               .recordRetryRequestKeyCount(multiKeyRequestContext.retryContext.retryRequestContext.numKeysInRequest);
+          clientStats.recordMultiKeyFanoutSize(multiKeyRequestContext.retryContext.retryRequestContext.getFanoutSize());
```
gaojieliu, line 185: Can we track retry fanout size separately from the original request?
The retried request typically has few keys than the original request, so I would like to understand the behavior of the retry request.

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/stats/FastClientStats.java
```diff
@@ -96,6 +96,7 @@ private FastClientStats(MetricsRepository metricsRepository, String storeName, R
         return System.currentTimeMillis() - this.cacheTimeStampInMs;
       }
     }, "metadata_staleness_high_watermark_ms"));
+    this.multiKeyFanoutSizeSensor = registerSensor("multi_key_fanout_size", new Avg(), new Max());
```
gaojieliu, line 99: Shall we just called it `fanout_size` as the request type will be the prefix of the stat name?

ZacAttack (approved):

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/MultiKeyRequestContext.java
```diff
@@ -167,6 +169,14 @@ public void setRoutesForPartitionMapping(Map<Integer, Set<String>> routesForPart
     this.routesForPartition = routesForPartition;
   }
 
+  public void setFanoutSize(int fanoutSize) {
+    this.fanoutSize = fanoutSize;
+  }
+
+  public int getFanoutSize() {
+    return fanoutSize;
+  }
```
gaojieliu, line 178: Nit,
do we need to maintain `fanoutSize` field here?
It seems we can directly use `routeRequests.size()` as the fanout size, right?

xunyin8: That was my initial implementation but that didn't work because in the `RetriableAvroGenericStoreClient` we decided to pass in a new request context instead of the original which is what the metric emission will use. This is why I added the field and setter. Alternatively we need to expose setting the `routeRequests` which is a more intrusive change IMO.
```
 requestContext.setFanoutSize(originalRequestContext.getFanoutSize());
```

gaojieliu (approved):
Thanks for the change!

PR #1113, opened 2024-08-13
[server] Fix server side connection tracking and logging

sushantmane:
## Fix server side connection tracking and logging

**Switch to channelActive/channelInactive for Accurate Connection Tracking:**
Replaced the use of channelRegistered/channelUnregistered with
channelActive/channelInactive to improve the tracking of active connections.

**Safety Enhancements:** Added safeguards to ensure connections are only tracked if
they have been correctly marked as active. This prevents multiple
channelActive/channelInactive events from being processed for the same
connection and avoids tracking connections that are marked inactive before ever
being marked as active.

**Router Principal Update:** Updated the router principal name from CN=venice-router
to venice-router and switched to using a substring match instead of equals for
more flexible connection tracking from routers.

**Log Notifier Improvement:** Enhanced log statements within LogNotifier for better
clarity and removed redundant logging to streamline output.

## How was this PR tested?

CI


FelixGV (approved):
Thanks!

PR #1116, opened 2024-08-14
[fast-client] Fixed request banding issue and request id generator

gaojieliu:
This PR fixed two issues:
1. Request banding issue and it was caused by the request-id-based replica selection while using least-loaded routing strategy, and the fix is to randomize the replicas for even distribution.
2. Only generate a new request id when trying to send out a request out.

This PR also disabled a few test cases, which rely on the deterministic replica selection while using least-loaded routing strategy and need to evaluate whether we should remove them completely or not.

This PR exposes one more config: `harClusters` and if the discovered server d2 service name fails into this category, `ClientRoutingStrategyType.HELIX_ASSISTED` will be used and if the customers would like to skip HAR routing algo, they can pass an empty set.

I also discovered a few issues when investigating the above items:
1. Fast-Client is not sending `streaming` header to Venice Server, which would result in some inefficiency for retry as Venice Server won't return non-existing keys and currently, retry logic is trying to collect keys without a response and then retry.
2. `RetriableAvroGenericStoreClient` will create two `RequestContext` for multi-key request, and one for the original request and the other one is for the retry request, but at the end, the routing related info are not copying to the passed `RequestContext`, which results in detailed metrics missing issue. It is tricky to make it right as we need to establish some consolidation mechanism.

Will follow up with these issues in a following PR.

## How was this PR tested?
CI

### DIFF
--- a/clients/venice-client/src/main/java/com/linkedin/venice/fastclient/DispatchingAvroGenericStoreClient.java
+++ b/clients/venice-client/src/main/java/com/linkedin/venice/fastclient/DispatchingAvroGenericStoreClient.java
@@ -56,6 +56,7 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentSkipListSet;
 import java.util.concurrent.Executor;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.Function;
 import org.apache.avro.Schema;
 import org.apache.avro.generic.GenericRecord;
@@ -68,6 +69,8 @@ import org.apache.logging.log4j.Logger;
  * This class is in charge of routing and serialization/de-serialization.
  */
 public class DispatchingAvroGenericStoreClient<K, V> extends InternalAvroStoreClient<K, V> {
+  private static final AtomicLong REQUEST_ID_GENERATOR = new AtomicLong();
+
   private static final Logger LOGGER = LogManager.getLogger(DispatchingAvroGenericStoreClient.class);
   private static final String URI_SEPARATOR = "/";
   private static final Executor DESERIALIZATION_EXECUTOR = AbstractAvroStoreClient.getDefaultDeserializationExecutor();
@@ -217,7 +220,7 @@ public class DispatchingAvroGenericStoreClient<K, V> extends InternalAvroStoreCl
      * might return more than required number of routes
      */
     List<String> routes = metadata.getReplicas(
-        requestContext.requestId,
+        REQUEST_ID_GENERATOR.getAndIncrement(),
         currentVersion,
         partitionId,
         requiredReplicaCount,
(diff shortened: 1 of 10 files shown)


clients/venice-client/src/main/java/com/linkedin/venice/fastclient/meta/LeastLoadedClientRoutingStrategy.java
```diff
@@ -23,20 +23,20 @@ public LeastLoadedClientRoutingStrategy(InstanceHealthMonitor instanceHealthMoni
...
     }
-    int replicaCnt = replicas.size();
-    int startPos = (int) (requestId % replicaCnt);
     List<String> availReplicas = new ArrayList<>();
-    for (int i = 0; i < replicaCnt; ++i) {
-      String replica = replicas.get((i + startPos) % replicaCnt);
+    /**
+     * For even distribution, we need to shuffle the replicas.
+     */
+    Collections.shuffle(replicas);
```
m-nagarajan, line 34: Looking at the code, `requestId` only made a difference if the pending request count is the same for multiple replicas as it immediately does a sort after this. so, when there is not much traffic, if the pending request counter numbers are 0 (or similar due to some reasons), then the shuffle comes to rescue?

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/DispatchingAvroGenericStoreClient.java
```diff
@@ -217,7 +220,7 @@ protected CompletableFuture<V> get(GetRequestContext requestContext, K key) thro
      * might return more than required number of routes
      */
     List<String> routes = metadata.getReplicas(
-        requestContext.requestId,
+        REQUEST_ID_GENERATOR.getAndIncrement(),
```
m-nagarajan, line 223: can you elaborate more on why this change is helpful?

m-nagarajan (commented):
Thanks @gaojieliu. Left a couple of questions.

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/meta/LeastLoadedClientRoutingStrategy.java
gaojieliu: Each client only has the local view, so if the qps is low, the pending request count for every node can be 0 for most of time as you mentioned.

Request-id-based replica assignment will always assign the same group of replicas when the reminder is same, which results in the consistent request banding in the backend, and reshuffle will guarantee some randomness for every request to achieve even distribution.

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/DispatchingAvroGenericStoreClient.java
gaojieliu: Currently, in `RetriableAvroGenericStoreClient`, when retry is enabled, for multi-get request, it will generate 3 RequestContexts: original one passed by the upstreaming layer, a new one for the original request and a new one for the retry request, so request-id is kind of messy-up even it would only send out two requests at most.
The request id generation in `DispatchingAvroGenericStoreClient.java` will guarantee the request id would only be bumped when there is a request to be sent out, and this also decouples request id generation from the `RequestContext` creation to be less error prone.

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/meta/LeastLoadedClientRoutingStrategy.java
m-nagarajan: Makes sense thanks.

m-nagarajan (approved):
LGTM, thanks.

PR #1117, opened 2024-08-14
[fast-client] Make multi-key retry budget enforce at key level in FC

xunyin8:
## [fast-client] Make multi-key retry budget enforce at key level in FC
Unlike routers where traffic from many thin-clients are funnelled through a relatively small set of routers, FC can have the following scenario:

1. Many FCs doing low QPS high batch size multi-key workload. e.g. 1-2 QPS with batch size of 500.

2. RetryManager will calculate to have the refill amount to be at least 1 for the token bucket.

3. If the long tail retry threshold for batch get is misconfigured and we are constantly retrying on all keys then we are not really limiting the retries to be 10% of traffic. The KPS will double for 1 QPS, 50% for 2 QPS and so on. The impact could be significant when there are many FCs with low QPS. The excessive retry could be triggered by bad configs or degrading servers. Both were the intended scenario for retry budget.

Enforcing the retry budget at key level for multi-key requests in FC for best attempt to keep retry traffic around 10% (current config) of user traffic.

## How was this PR tested?

### DIFF
--- a/clients/venice-client/src/main/java/com/linkedin/venice/fastclient/RetriableAvroGenericStoreClient.java
+++ b/clients/venice-client/src/main/java/com/linkedin/venice/fastclient/RetriableAvroGenericStoreClient.java
@@ -312,7 +312,7 @@ public class RetriableAvroGenericStoreClient<K, V> extends DelegatingAvroStoreCl
           finalRequestCompletionFuture.completeExceptionally(throwable);
           return;
         }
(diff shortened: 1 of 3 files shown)


gaojieliu (commented):
Code change looks good.
Can you write a unit test to verify the new behavior?

gaojieliu (approved):

PR #1121, opened 2024-08-15
[server][test] Add integration test for thread pool jumping by enabling/disabling AA or WC.

haoxu07:
### `PartialUpdateTest#testEnablePartialUpdateOnActiveActiveStore` change
This PR adds an integration test support for verifying the pool jumping strategy: `KafkaConsumerServiceDelegator.ConsumerPoolStrategyType.CURRENT_VERSION_PRIORITIZATION`, when version role and workload type changed.  

### `ConsumptionTask` change
We also add consumer pool name into the `ConsumptionTask`'s id string, we are able to know the consumer thread comes from a certain pool. This will ensure `TopicPartitionIngestionContext` contains the `ConsumptionTask`'s id string for certain topic partition belonging to a specific version topic. To assert the pool changed or not, we rely on previous debugging method by querying `TopicPartitionIngestionContext` from server host. 


gaojieliu (approved):
Thanks!

sixpluszero (approved):

FelixGV:
Hello @haoxu07 ... it looks like the test added in this PR might be flaky? See this example, where the modified integration test failed 5 times in a row:

https://github.com/linkedin/venice/actions/runs/10620188779/job/29439452773?pr=1147

The failures come specifically from `verifyConsumerThreadPoolFor`, which was added in this PR, but IDK if it indicates an issue with the main code or with the test code.

Would you mind taking a look please?

PR #1122, opened 2024-08-15
[test] Microbenchmark for TokenBucket

FelixGV:
We suspected that TokenBucket may be slow, though this microbenchmark indicates that it can run at ~microsecond latency even under high concurrency, and at sub-microsec under light load. So this is not the bottleneck we were looking for, but we'll keep the microbench checked in just in case we need it again in the future.

## How was this PR tested?
PR itself is just one more test.


sushantmane (approved):
LGTM! Thanks!

PR #1123, opened 2024-08-15
[server] Enforce Record Size Limit on Nearline Jobs / Partial Update

KaiSernLim:
## Summary

1. **[Functional]** Added a `catch RecordTooLargeException` clause for `VeniceWriter#put()` in `ActiveActiveStoreIngestionTask` for the expected partial update code path. When the exception is caught, consumption will be paused for the partition and a user will need to work with the Venice team to get it resolved by bumping up the size limit `maxRecordSizeBytes`.
2. **[Functional]**  In `StorageUtilizationManager`, added pausing and resuming consumption on partitions for `RecordTooLarge` issues. If the store-level config `maxRecordSizeBytes` is increased, `StorageUtilizationManager` will try to resume consumption for all partitions paused due to `RecordTooLarge`. Partitions must not be paused for either `RecordTooLarge` or `QuotaExceeded` in order to be allowed to resume consumption.
3. **[Config]** Moved `maxRecordSizeBytes` from its temporary position within `LeaderFollowerStoreIngestionTask` into its final position within `VeniceWriter`.
4. **[Unit Test]** Added to existing test `testLeaderCanSendValueChunksIntoDrainer()` in `ActiveActiveStoreIngestionTaskTest` to verify that partitions are paused when a large record is detected.
5. **[Unit Test]** `TestPausePartitionForRecordTooLage()` and `testPartialResumePartitions()` in `StorageUtilizationManagerTest` verifies that partitions are paused and resumed accordingly, even despite the interactions with Hybrid Quota.
6. **[Integration Test]** `testLargeValuesPartialUpdate()` in `PartialUpdateTest` verifies that `RecordTooLargeException` is thrown on the nearline path.

TODO: Add an integration test that covers the interaction between hybrid quota and this

## How was this PR tested?

Unit tests:
    1. `testLeaderCanSendValueChunksIntoDrainer()` in `ActiveActiveStoreIngestionTaskTest`
    2. `TestPausePartitionForRecordTooLage()` and `testPartialResumePartitions()` in `StorageUtilizationManagerTest`
Integration Tests:
    1. `testLargeValuesPartialUpdate()` in `PartialUpdateTest`

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
@@ -702,7 +702,7 @@ public class ActiveActiveStoreIngestionTask extends LeaderFollowerStoreIngestion
    * This function may modify the original record in KME and it is unsafe to use the payload from KME directly after
    * this function.
    *
-   * @param mergeConflictResult        The result of conflict resolution.
+   * @param mergeConflictResult         The result of conflict resolution.
    * @param partitionConsumptionState The {@link PartitionConsumptionState} of the current partition
    * @param key                       The key bytes of the incoming record.
    * @param consumerRecord            The {@link PubSubMessage} for the current record.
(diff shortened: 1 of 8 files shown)


gaojieliu:
Hi @KaiSernLim 
I am not sure whether we have reached consensus regarding how to deal with real-time ingestion when encountering too large record, so I suggest you to start a thread internally, and even we want to have such feature merged, I think we might need a different config to measure real-time large record as pausing consumption has severe outcome, which we need to evaluate carefully.

PR #1125, opened 2024-08-19
[fast-client][router] Fixed one slow start issue

gaojieliu:
During the fast-client experiment, I noticed that
the fast-client will take a quite long time to stablize the latency for medim/large response.
This behavior is reproducible, but the stablization periods vary. After several rounds of experiments, I realized that this can be related to the small Http2 InitialWindow/Frame settings being used by Fast-Client. Even Venice Server is using a large config value, but the handshake will choose the small ones sent by Fast-Client, H2 can automatically adjust the setting, but it will take time and the duration is non-deterministic. So this PR will use a same set of values as Venice Server to avoid this slow start issue.

## How was this PR tested?
CI and manual test.


adamxchen (approved):
👍

m-nagarajan (approved):
Thanks @gaojieliu

PR #1126, opened 2024-08-19
[vpj] Exclude jackson dependency from HDFS dependency

huangminchn:
## Summary
Also reverted a7318c1d918af1d0c53214173200f9e1c254cd9c:
[controller][router][server][da-vinci] Pull jacksonCore into venice-common (#1124)

## How was this PR tested?
./gradlew assemble

### DIFF
--- a/clients/venice-push-job/build.gradle
+++ b/clients/venice-push-job/build.gradle
@@ -31,6 +31,7 @@ dependencies {
     // Exclude transitive dependency
     exclude group: 'org.apache.avro'
     exclude group: 'javax.servlet'
+    exclude group: 'com.fasterxml.jackson'
   }
 
   implementation (libraries.apacheSparkAvro) {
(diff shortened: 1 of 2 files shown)


nisargthakkar (dismissed):

FelixGV (approved):
Thanks Min! Let's give it a shot!

clients/venice-push-job/build.gradle
```diff
@@ -31,6 +31,7 @@ dependencies {
     // Exclude transitive dependency
     exclude group: 'org.apache.avro'
     exclude group: 'javax.servlet'
+    exclude group: 'com.fasterxml.jackson'
```
nisargthakkar, line 34: From Gaojie's exploration, the offending lib is `com.fasterxml.jackson.core:jackson-databind`. So, do we need to exclude group `com.fasterxml.jackson.core` instead?

gaojieliu: @huangminchn 
You can run a depdencyInsight to see whether the offending lib is still there or not.

gaojieliu: @nisargthakkar is right, and we should do:
```
exclude group: 'com.fasterxml.jackson.core'
```

huangminchn: Yeah, updated.

gaojieliu (approved):

PR #1128, opened 2024-08-20
[vpj] Remove hadoop-hdfs dependency

nisargthakkar:
## Remove hadoop-hdfs dependency

This dependency pulls in jackson versions that have vulnerabilities. Turns out, this might not be needed. If tests pass, we can merge this instead of trying to fix the gradle stuff.

## How was this PR tested?

GH CI


nisargthakkar:
When running tests locally, I found that there is a NSME error thrown:
```
Caused by: java.lang.ClassNotFoundException: org.apache.hadoop.hdfs.DistributedFileSystem
```

PR #1131, opened 2024-08-21
[server] [controller] [admin-tool] Max Nearline Record Size Config

KaiSernLim:
## Summary

Added the nearline analogue of the store-level max record size config from #1027 so that we will be able to bump the record size limit for nearline jobs without affecting the record size limit for batch jobs.

### Changes

**[Config]** Added a new store-level config setting `maxNearlineRecordSizeBytes`. This will be used to determine whether to pause consumption on nearline jobs with partial updates if the server encounters a record that is too large. If not specified (`-1`), the server config `default.max.record.size.bytes` will be backfilled (`100MB` default).

## How was this PR tested?

GHCI


clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -188,7 +188,8 @@ public class LeaderFollowerStoreIngestionTask extends StoreIngestionTask {
 
   private final AtomicLong lastSendIngestionHeartbeatTimestamp = new AtomicLong(0);
 
-  protected final int maxRecordSizeBytes; // TODO: move into VeniceWriter when nearline jobs enforce max record size
```
lluwm, line 191: If I understand, this is the max record size for batch data, can you explain a bit why we can remove it here?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -1902,12 +1904,12 @@ protected void recordProcessedRecordStats(
 
   @Override
   protected final double calculateAssembledRecordSizeRatio(long recordSize) {
-    return (double) recordSize / maxRecordSizeBytes;
+    return (double) recordSize / maxNearlineRecordSizeBytes;
```
lluwm, line 1907: It seems that we don't differentiate batch record from the nearline data in calculating the ratio. Would that be okay? What's your opinion?

lluwm (commented):
Thanks for the PR, @KaiSernLim , it looks good to me overall, just left several comments.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
KaiSernLim: This max record size variable is only for nearline data. The one pertaining to batch data is set by the `DataWriterJob` onto `VeniceWriter` on the VPJ side ([link](https://github.com/linkedin/venice/blob/aea19bfa64b98ed9c34e23a2ef454c17f97b95a3/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/task/datawriter/AbstractPartitionWriter.java#L358)).

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
KaiSernLim: That's true. We already cannot differentiate between batch and nearline data when calculating the metric. Now there is an added difference in the record size limit between batch and nearline. The default batch limit goes through the controller path, so the server (this code path) never even gets to see it.

Maybe we can just leave that as a limitation, especially given the current low number of stores with large records. What do you think @gaojieliu?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
lluwm: Got it and thanks for the explanation.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
gaojieliu: I think it will be fine to use `maxRecordSizeBytes` (batch part) for metrics purpose, and the reason to introduce `maxNearlineRecordSizeBytes` is to avoid failing the nearline job before we have a good way to communicate with the customers.

gaojieliu: In theory, we should use the same number for both the batch job and nearline job eventually when everything is ready, and the separate nearline limit is intermediate to ease the feature rollout.

lluwm: Sounds good, please add some comments in the to clarify it.

lluwm (approved):

PR #1132, opened 2024-08-21
[Server][FC] throws exception during store migration to enforce d2 refresh

adamxchen:
## Summary

As titled, we need to throw exceptions explicitly so FC can detect the store migration and refresh d2 service as needed. How it works is basically we compare the server cluster with the store cluster set on ZK. If they are different, it means the store migration has completed in the given region, so we can throw exception to refresh the service discovery.

## How was this PR tested?

new unit tests and the integ test.


clients/venice-client/src/test/java/com/linkedin/venice/fastclient/meta/RequestBasedMetadataTest.java
```diff
@@ -370,4 +375,31 @@ public void testRequestBasedMetadataStartFailFast() throws IOException {
       Assert.assertThrows(ConfigurationException.class, requestBasedMetadata::start);
     }
   }
+
+  @Test(timeOut = TEST_TIMEOUT)
+  public void testRequestBasedMetadataStartRefreshDuringStoreMigration() throws IOException, InterruptedException {
```
adamxchen, line 380: I'll rename this. This unit test is to validate whether FC would do refresh when there's exception caught

services/venice-server/src/main/java/com/linkedin/venice/listener/ServerReadMetadataRepository.java
```diff
@@ -72,6 +72,11 @@ public MetadataResponse getMetadata(String storeName) {
                 "Fast client is not enabled for store: %s, please ensure storage node read quota is enabled for the given store",
                 storeName));
       }
+      if (store.isMigrating()) {
```
xunyin8, line 75: This check is incomplete. We only throw exception if the `store.isMigrating()` AND the cluster has changed to destination cluster. i.e. complete migration is called but we haven't ended migration yet. This is the window we should be redirecting. Too early won't do the trick.

xunyin8 (changes requested):

services/venice-server/src/main/java/com/linkedin/venice/listener/ServerReadMetadataRepository.java
```diff
@@ -72,6 +77,21 @@ public MetadataResponse getMetadata(String storeName) {
                 "Fast client is not enabled for store: %s, please ensure storage node read quota is enabled for the given store",
                 storeName));
       }
+
+      if (store.isMigrating()) {
+        // only obtain store Config when store is migrating and only throw exceptions when dest cluster is ready
+        StoreConfig storeConfig = zkStoreConfigAccessor.getStoreConfig(storeName);
```
xunyin8, line 83: Can we use `HelixReadOnlyStoreConfigRepository` instead of `ZkStoreConfigAccessor`? This `getStoreConfig` will generate a call to zk via `ZkBaseDataAccessor` which can be expensive when the server is handling many of these metadata requests. This approach won't generate any zk watchers but for something like StoreConfig that doesn't change frequently I think `HelixReadOnlyStoreConfigRepository` has the better trade-offs.

services/venice-server/src/main/java/com/linkedin/venice/listener/ServerReadMetadataRepository.java
adamxchen: @xunyin8 Something occurred to me about this logic. Say, at the stage 2, current cluster name is same as dest cluster so a refresh happens but if the `end migration` command is not sent timely, the the client actually would keep getting exceptions because isMigrating flag will only be set to false after `end migration` command is received.

`end migration` can only be called when all regions has completed migration which means if any of regions is lagging behind, the clients in other regions couldn't proceed due to this check.

adamxchen: If the server tells clients which cluster it should belongs to and client has a way to verify that (e.g. using `serverD2ServiceName` to infer the cluster name so it can choose to refresh or not.  WDYT?

xunyin8: Ok, let's take a step back and outline the exact conditions we are checking and we can go over the different cases. We can even include the case where the migration was completed but we decided to abort last second for whatever reason.

The conditions we are checking and will be throwing exception are (both needs to be true):
1. Store is migrating
2. Cluster info does not match with the cluster that the server is responding to the request. e.g. store cluster config is v-1 but the server responding is in v-0.

In the case you described, once we re-discover and start hitting server in v-1 we will no longer throw exceptions and metadata request will go through and we will be getting routing table in v-1 and traffic will be going to v-1.

It'll be tricky if we decided to abort migration after we completed migration for whatever reason. Store is no longer migrating, we reset the cluster info and deleted the store in dest cluster. This will cause an issue and I feel like we need to improve this even for other clients (TC) as well. We should also split abort into two steps like complete+end. We should only delete stores in dest or src in end and both abort and complete should just be changing the cluster info. If we do that then we will also get a window to redirect traffic back to src cluster.

adamxchen: I see why we have this misunderstanding. You are saying we check the server cluster name with the cluster set for a given store from `StoreConfig`. I thought you suggest we compare the dest cluster name with the cluster set for a given store.  The former makes more sense.

FelixGV:
For a commit which changes the server, please include `[server]` in the commit message prefix.

services/venice-server/src/main/java/com/linkedin/venice/listener/ServerReadMetadataRepository.java
gaojieliu: For the abort-migration after complete-migration, if it is tricky to support it, maybe we should just disallow such action as the operator can always finish the migration and migrate it back to the original cluster if there is a need, but I don't think this happen often.

xunyin8: I checked the code and abort migration is a two step approach already unless we provided with override flag. So I think we are good in that front too. @gaojieliu Why is it tricky? Today we just rever the cluster discovery back to original and can stay in that state without deleting the cloned store until traffic is fully shifted back.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestStoreMigration.java
```diff
@@ -532,103 +551,11 @@ private Properties createAndPushStore(String clusterName, String storeName) thro
...
-        Assert.assertFalse(storeResponse.getStore().isMigrating());
-        Assert.assertFalse(storeResponse.getStore().isMigrationDuplicateStore());
-      });
-    }
-
-  }
-
   private void readFromStore(AvroGenericStoreClient<String, Object> client) {
     int key = ThreadLocalRandom.current().nextInt(RECORD_COUNT) + 1;
     client.get(Integer.toString(key));
```
xunyin8, line 556: A bit surprised to see this, but shouldn't we get the result of the future and assert on it in case there are exceptions?

xunyin8 (commented):
LGTM just some minor comments

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestStoreMigration.java
adamxchen: Wow, didn't realize it doesn't get on the future. Will update it and see how it looks. like

adamxchen: I've updated it. It's a bit tricky to intentionally catch the exception from the client, not the `RequestBasedMetadata` so didn't add it but add one additional call

xunyin8: The client should never throw any exceptions since we are trying to make store migration transparent to our users. The underlying request metadata fetch can fail and do the re-discovery but that shouldn't impact actual read traffic from the client.

xunyin8 (approved):

adamxchen:
Enabled the automerge but forgot to squash my commit msgs 😢

PR #1133, opened 2024-08-21
[server][da-vinci] Bug fix: unsorted drainer service metrics failed to register

huangminchn:
## Summary
Both sorted and unsorted drainer services are using the same metrics name, and sorted drainer service registers metric first; as a result, unsorted drainer service fails to register because it believes the metrics already exist - inside AbstractVeniceStats#registerSensor, it calls HashMap#computeIfAbsent, so whoever comes later with the same name would fail to register.

This PR fixes the bug by using different metric name prefix: StoreBufferServiceSorted
StoreBufferServiceUnsorted

## How was this PR tested?
Added unit test to ensure two sets of metrics are registered.

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreBufferService.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreBufferService.java
@@ -80,10 +80,10 @@ public class StoreBufferService extends AbstractStoreBufferService {
     }
     this.isSorted = sorted;
     this.leaderRecordHandler = queueLeaderWrites ? this::queueLeaderRecord : StoreBufferService::processRecord;
-    String metricName = sorted ? "StoreBufferServiceSorted" : "StoreBufferServiceUnsorted";
+    String metricNamePrefix = sorted ? "StoreBufferServiceSorted" : "StoreBufferServiceUnsorted";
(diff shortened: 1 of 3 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreBufferService.java
```diff
@@ -80,8 +80,10 @@ public StoreBufferService(
     }
     this.isSorted = sorted;
     this.leaderRecordHandler = queueLeaderWrites ? this::queueLeaderRecord : StoreBufferService::processRecord;
+    String metricName = sorted ? "StoreBufferServiceSorted" : "StoreBufferServiceUnsorted";
     this.storeBufferServiceStats = new StoreBufferServiceStats(
         metricsRepository,
+        metricName,
```
haoxu07, line 86: Nit: maybe changed to `metricNamePrefix`?

huangminchn: For sure. Updated.

majisourav99 (approved):
LGTM. Thanks Min for fixing the missing metrics.

haoxu07 (approved):
Thanks for the fix!

PR #1139, opened 2024-08-23
[server] Add metric to verifying current version prioritization strategy.

haoxu07:

This PR adds 2 metrics:
1) `total_kafka_consumer_service_for_<pub sub region alias>--subscribed_partitions_num`: consumer pool level metric total subscribed topic partitions of all consumers, this could be used to verify specific pool subscription change when a new version push is happened.
2) `<store name>--resubscription_failure`: store level metric to alert on exception throwing during resubscription when ingestion role changed.

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaConsumerService.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaConsumerService.java
@@ -68,22 +68,20 @@ import org.apache.logging.log4j.Logger;
  * @see AggKafkaConsumerService which wraps one instance of this class per Kafka cluster.
  */
 public abstract class KafkaConsumerService extends AbstractKafkaConsumerService {
-  private static final int SHUTDOWN_TIMEOUT_IN_SECOND = 1;
-  private static final RedundantExceptionFilter REDUNDANT_LOGGING_FILTER =
-      RedundantExceptionFilter.getRedundantExceptionFilter();
-
-  private final ExecutorService consumerExecutor;
   protected final String kafkaUrl;
   protected final String kafkaUrlForLogger;
-  private final Logger LOGGER;
-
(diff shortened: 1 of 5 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaConsumerService.java
```diff
@@ -68,22 +68,20 @@
...
-      RedundantExceptionFilter.getRedundantExceptionFilter();
-
-  private final ExecutorService consumerExecutor;
   protected final String kafkaUrl;
   protected final String kafkaUrlForLogger;
-  private final Logger LOGGER;
-
-  protected AggKafkaConsumerServiceStats aggStats;
+  protected final ConsumerPoolType poolType;
+  protected final AggKafkaConsumerServiceStats aggStats;
```
huangminchn, line 74: Why the agg stats is inside a specific consumer service?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/HostLevelIngestionStats.java
```diff
@@ -325,6 +327,12 @@ public HostLevelIngestionStats(
         () -> totalStats.ingestionFailureSensor,
         new Count());
 
+    this.resubscriptionFailureSensor = registerPerStoreAndTotalSensor(
+        "resubscription_failure",
+        totalStats,
+        () -> totalStats.resubscriptionFailureSensor,
+        new Count());
```
huangminchn, line 334: FYI, Count metric would always increase and doesn't reset until we bounce the node, but maybe this is what we want here?

haoxu07: Cumulative failure count is sth we expect to have here.

huangminchn (approved):
Thanks Hao. LGTM!

PR #1140, opened 2024-08-26
[fc] Fix fast client multiget metrics to use the right context

m-nagarajan:
## Summary
1. Fast client multiget created 2 new contexts for the original and the retry request on top of the context passed to it which was used while recording metrics. Fix is to create a new context only for the retry and use the context passed for the original request.
2. Added additional check on the retry request context for partial response failure.
3. removed unused `isTriggeredByBatchGet` for single get

Resolves #XXX

## How was this PR tested?
GH CI

### DIFF
--- a/clients/venice-client/src/main/java/com/linkedin/venice/fastclient/InternalAvroStoreClient.java
+++ b/clients/venice-client/src/main/java/com/linkedin/venice/fastclient/InternalAvroStoreClient.java
@@ -45,50 +45,6 @@ public abstract class InternalAvroStoreClient<K, V> implements AvroGenericReadCo
     return batchGet(new BatchGetRequestContext<>(keys.size(), false), keys);
   }
 
-  /**
-   * Check for partial failures in the multi-key request context and return the exception if any.
-   * @param multiKeyRequestContext
-   * @return
-   */
-  Throwable checkBatchGetPartialFailure(MultiKeyRequestContext<K, V> multiKeyRequestContext) {
-    Throwable partialResponseException = null;
-    // check for partial failures for multi-key requests
-    if (multiKeyRequestContext.retryContext != null
-        && multiKeyRequestContext.retryContext.retryRequestContext != null) {
-      // retry is triggered
-      if (multiKeyRequestContext.isPartialSuccessAllowed) {
-        if (multiKeyRequestContext.retryContext.retryRequestContext.isCompletedSuccessfullyWithPartialResponse()) {
-          partialResponseException =
-              (Throwable) multiKeyRequestContext.retryContext.retryRequestContext.getPartialResponseException().get();
-        }
-      } else if (multiKeyRequestContext.retryContext.retryRequestContext.getPartialResponseException().isPresent()) {
-        partialResponseException =
-            (Throwable) multiKeyRequestContext.retryContext.retryRequestContext.getPartialResponseException().get();
-      }
-      if (partialResponseException != null) {
-        // if there is no exception in the retry request, everything passed, but if there is an exception in the
-        // retry request, that failure might have passed in the original request after the retry started. checking
-        // the numKeysCompleted for now.
-        int totalKeyCount = multiKeyRequestContext.numKeysInRequest;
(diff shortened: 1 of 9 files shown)


clients/venice-client/src/main/java/com/linkedin/venice/fastclient/InternalAvroStoreClient.java
```diff
@@ -45,6 +45,50 @@ public final CompletableFuture<Map<K, V>> batchGet(Set<K> keys) throws VeniceCli
...
+    Throwable partialResponseException = null;
+    // check for partial failures for multi-key requests
+    if (multiKeyRequestContext.retryContext != null
+        && multiKeyRequestContext.retryContext.retryRequestContext != null) {
+      // retry is triggered
+      if (multiKeyRequestContext.isPartialSuccessAllowed) {
+        if (multiKeyRequestContext.retryContext.retryRequestContext.isCompletedSuccessfullyWithPartialResponse()) {
+          partialResponseException =
+              (Throwable) multiKeyRequestContext.retryContext.retryRequestContext.getPartialResponseException().get();
+        }
```
gaojieliu, line 63: The logic here seems a little confusing.
Is it possible to check whether the streaming response has received all the keys or not as the first step and if not, then try to collect the exception either from the original context or the retry context?
Let us try to converge the logic to always use the partial response API as recommended in the previous comment as the current logic seems to be confusing.

gaojieliu: I guess `isPartialSuccessAllowed` flag is used as an optimization in case there is no route available for a certain partition, FC can fail fast, but I would argue such type of optimization is not good:
1. The chance of such situation is low.
2. The code becomes more complex and hurts maintainability in some way.

The extra overhead on client side in the exceptional scenario normally is acceptable IMO.

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/InternalAvroStoreClient.java
```diff
@@ -53,11 +97,14 @@ protected CompletableFuture<Map<K, V>> batchGet(BatchGetRequestContext<K, V> req
     streamingResultFuture.whenComplete((response, throwable) -> {
       if (throwable != null) {
         resultFuture.completeExceptionally(throwable);
-      } else if (!requestContext.isPartialSuccessAllowed && requestContext.getPartialResponseException().isPresent()) {
-        resultFuture.completeExceptionally(
-            new VeniceClientException("Response was not complete", requestContext.getPartialResponseException().get()));
       } else {
-        resultFuture.complete(response);
+        Throwable partialResponseException = checkBatchGetPartialFailure(requestContext);
```
gaojieliu, line 101: Originally, I was thinking that `RetriableAvroGenericStoreClient` should decide whether the request with retry enabled succeed or not and other layers should be just pass-through.
Can you explain why we need to invoke this function here?

gaojieliu (commented):
Left some comments and let us explore whether we can make the code simpler and easier to reason/maintain.

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/InternalAvroStoreClient.java
m-nagarajan: removed these newly added logics. 

`RetriableAvroGenericStoreClient->getStreamingCallback->onRecordReceived`,  `onCompletion` and `finalRequestCompletionFuture.whenComplete` does these checks, so just doing a minor change there to account for `retryContext` as well.

gaojieliu (approved):
The code change LGTM!
We can find another chance to cleanup the logic of handling partial response as from my understanding, `batch-get` is an API, which should only present in the out-most layer, which is `StatsAvroGenericStoreClient/StatsAvroSpecificStoreClient`, where it can check whether the streaming response is full or not and if it is not full, throw the saved exception, and I think this will make the logic simpler.

PR #1141, opened 2024-08-26
[vpj] [common] [da-vinci] Refactor schema source venice-common to consolidate duplicated code in ETL

mynameborat:
## Summary
- Extend `SchemaSource` to fetch and persist key schema
- Move `SchemaSource` and `HDFSSchemaSource` to `venice-common`
- Move `SchemaUtils` from `da-vinci` to `venice-common`

The changes will enable us to leverage the caching code to disk in ETL libraries and avoid duplicating RMD related logic which doesn't exist currently.

## How was this PR tested?
- Unit test


internal/venice-common/src/main/java/com/linkedin/davinci/schema/SchemaUtils.java
xunyin8, line 1: Should we remove `davinci` from the path after moving it to venice-common? Probably `internal/venice-common/src/main/java/com/linkedin/venice/schema` is a better place?

internal/venice-common/src/main/java/com/linkedin/venice/hadoop/schema/HDFSSchemaSource.java
```diff
@@ -121,6 +155,22 @@ void saveSchemaResponseToDisk(MultiSchemaResponse.Schema[] schemas, boolean isRm
     }
   }
 
+  void saveKeySchemaToDisk(SchemaResponse schema) throws IOException, IllegalStateException {
+    Preconditions.checkState(includeKeySchema, "Cannot be invoked with invalid key schema directory");
```
xunyin8, line 159: Should we also check if the `SchemaResponse` has any errors before processing it?

internal/venice-common/src/main/java/com/linkedin/venice/hadoop/schema/HDFSSchemaSource.java
```diff
@@ -121,6 +155,22 @@ void saveSchemaResponseToDisk(MultiSchemaResponse.Schema[] schemas, boolean isRm
...
+
+    Path schemaPath = new Path(keySchemaDir, String.valueOf(schema.getId()));
+    if (!fs.exists(schemaPath)) {
+      try (FSDataOutputStream outputStream = fs.create(schemaPath);
+          OutputStreamWriter outputStreamWriter = new OutputStreamWriter(outputStream, StandardCharsets.UTF_8)) {
+        outputStreamWriter.write(schema.getSchemaStr() + "\n");
+        outputStreamWriter.flush();
+      }
+    } else {
+      throw new IllegalStateException(String.format("The schema path %s already exists", schemaPath));
```
xunyin8, line 170: In what scenario can this happen? Are we the ones responsible for cleaning up the schema paths after use?

internal/venice-common/src/main/java/com/linkedin/venice/hadoop/schema/HDFSSchemaSource.java
```diff
@@ -173,6 +223,31 @@ public Map<Integer, Schema> fetchValueSchemas() throws IOException {
     return mapping;
   }
 
+  @Override
+  public Schema fetchKeySchema() throws IOException {
+    FileStatus[] fileStatus = fs.listStatus(keySchemaDir);
```
xunyin8, line 228: Should we short circuit and throw a more explicit exception in the case when `keySchemaDir ` is null and includeKeySchema is `false`?

internal/venice-common/src/main/java/com/linkedin/davinci/schema/SchemaUtils.java
mynameborat: I kept the package as is to minimize the changes. Additionally it is mostly used by davinci and the test usage is the first one outside. 
We needed to move this file to avoid dependency issue.

xunyin8: I see, was just odd to see davinci in the path for a utils library in `venice-common`

xunyin8 (approved):

PR #1142, opened 2024-08-27
[server][metrics] Add new metric which distinguishes a serving partition from a catching up one

ZacAttack:
## [server][metrics] Add new metric which distinguishes a serving partition from a catching up one

Additionally mutes the metric of whichever one isn't getting updated

Resolves #XXX

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/ingestion/heartbeat/HeartbeatMonitoringService.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/ingestion/heartbeat/HeartbeatMonitoringService.java
@@ -244,7 +244,7 @@ public class HeartbeatMonitoringService extends AbstractVeniceService {
     recordLags(
         leaderHeartbeatTimeStamps,
         ((storeName, version, region, heartbeatTs, isReadyToServe) -> versionStatsReporter
-            .recordLeaderLag(storeName, version, region, heartbeatTs, isReadyToServe)));
+            .recordLeaderLag(storeName, version, region, heartbeatTs)));
     recordLags(
         followerHeartbeatTimeStamps,
         ((storeName, version, region, heartbeatTs, isReadyToServe) -> versionStatsReporter
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/ingestion/heartbeat/HeartbeatMonitoringService.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/ingestion/heartbeat/HeartbeatMonitoringService.java
@@ -9,6 +9,8 @@ import io.tehuti.metrics.MetricsRepository;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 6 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/ingestion/heartbeat/HeartbeatMonitoringService.java
```diff
@@ -41,8 +43,8 @@ public class HeartbeatMonitoringService extends AbstractVeniceService {
   private final String localRegionName;
 
   // store -> version -> partition -> region -> timestamp
-  private final Map<String, Map<Integer, Map<Integer, Map<String, Long>>>> followerHeartbeatTimeStamps;
-  private final Map<String, Map<Integer, Map<Integer, Map<String, Long>>>> leaderHeartbeatTimeStamps;
+  private final Map<String, Map<Integer, Map<Integer, Map<String, Pair<Long, Boolean>>>>> followerHeartbeatTimeStamps;
+  private final Map<String, Map<Integer, Map<Integer, Map<String, Pair<Long, Boolean>>>>> leaderHeartbeatTimeStamps;
```
sixpluszero, line 47: nit: Overall it looks good, just this map of map of ... of map is getting more and more complicated. In the future we might consider converting to certain named class instead of Map+Pair.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/ingestion/heartbeat/HeartbeatStat.java
```diff
@@ -9,8 +9,9 @@
 
 
 public class HeartbeatStat {
-  Map<String, WritePathLatencySensor> leaderSensors = new VeniceConcurrentHashMap<>();
-  Map<String, WritePathLatencySensor> followerSensors = new VeniceConcurrentHashMap<>();
+  Map<String, WritePathLatencySensor> readyToServeLeaderSensors = new VeniceConcurrentHashMap<>();
+  Map<String, WritePathLatencySensor> readyToServeFollowerSensors = new VeniceConcurrentHashMap<>();
+  Map<String, WritePathLatencySensor> catchingUpFollowerSensors = new VeniceConcurrentHashMap<>();
```
sixpluszero, line 14: I just want to confirm my understanding -> since current version offline->standby is blocking, there is no chance leader is in (after EOP, before COMPLETED) status right? So in theory we don't need to do the same for leader.

ZacAttack: More importantly, we should NEVER squelch lag for leader for a current version.  The reason why is because so long as the leader isn't on the tail of the RT, we're lagging.  So it's important to always report this for the sake of accurate SLA reporting.

sixpluszero: That's true. Thanks for the confirmation.

PR #1143, opened 2024-08-27
[controller] Fail hybrid store conversion if a pushjob is running

majisourav99:
## Fail hybrid store conversion if a pushjob is already running

If a batch store is converted to hybrid while a push is already running, it might convert the store to hybrid when its running the push in the target colo. After the target colo push finishes, the state of the store is mismatched in target vs child colo. Then during the push to other colo which does not support hybrid store ingestion and might not have the RT topic created which would make the pushjob hang forever.  This PR blocks conversion to hybrid store during actively running pushes.

## How was this PR tested?

CI

### DIFF
--- a/internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/controller/TestClusterLevelConfigForActiveActiveReplication.java
+++ b/internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/controller/TestClusterLevelConfigForActiveActiveReplication.java
@@ -2,6 +2,7 @@ package com.linkedin.venice.controller;
 
 import static com.linkedin.venice.ConfigKeys.ENABLE_ACTIVE_ACTIVE_REPLICATION_AS_DEFAULT_FOR_HYBRID_STORE;
 import static com.linkedin.venice.utils.TestUtils.assertCommand;
+import static com.linkedin.venice.utils.TestUtils.waitForNonDeterministicPushCompletion;
 import static org.testng.Assert.assertEquals;
 import static org.testng.Assert.assertFalse;
 import static org.testng.Assert.assertTrue;
@@ -12,6 +13,7 @@ import com.linkedin.venice.integration.utils.ServiceFactory;
 import com.linkedin.venice.integration.utils.VeniceMultiRegionClusterCreateOptions;
 import com.linkedin.venice.integration.utils.VeniceTwoLayerMultiRegionMultiClusterWrapper;
 import com.linkedin.venice.meta.StoreInfo;
+import com.linkedin.venice.meta.Version;
 import com.linkedin.venice.utils.TestUtils;
 import com.linkedin.venice.utils.Time;
 import com.linkedin.venice.utils.Utils;
@@ -71,6 +73,12 @@ public class TestClusterLevelConfigForActiveActiveReplication {
 
     // Check store level Active/Active is enabled or not
     assertFalse(store.isActiveActiveReplicationEnabled());
+
+    waitForNonDeterministicPushCompletion(
+        Version.composeKafkaTopic(storeName, 1),
+        parentControllerClient,
+        30,
+        TimeUnit.SECONDS);
 
     // Convert to hybrid store
     assertCommand(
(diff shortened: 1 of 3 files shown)


services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
```diff
@@ -2279,6 +2279,16 @@ public void updateStore(String clusterName, String storeName, UpdateStoreQueryPa
       setStore.storeName = storeName;
       setStore.owner = owner.map(addToUpdatedConfigList(updatedConfigsList, OWNER)).orElseGet(currStore::getOwner);
 
+      if (!currStore.isHybrid() && (hybridRewindSeconds.isPresent() || hybridOffsetLagThreshold.isPresent())) {
```
huangminchn, line 2282: Can we only limit this to stores with target colo push enabled?

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/controller/TestClusterLevelConfigForActiveActiveReplication.java
```diff
@@ -72,6 +74,12 @@ public void testClusterLevelActiveActiveReplicationConfigForNewHybridStores() th
     // Check store level Active/Active is enabled or not
     assertFalse(store.isActiveActiveReplicationEnabled());
 
+    waitForNonDeterministicPushCompletion(
+        Version.composeKafkaTopic(storeName, 1),
+        parentControllerClient,
```
huangminchn, line 79: Why are we adding this? We didn't create topic in parent Kafka cluster in the test before?

majisourav99: These are actually failing because we were updating config while push was running.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
majisourav99: Today we do not have store config for target colo, only vpj config.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/controller/TestClusterLevelConfigForActiveActiveReplication.java
huangminchn: Then why adding a wait time here could help?

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/controller/TestClusterLevelConfigForNativeReplication.java
```diff
@@ -97,4 +105,30 @@ public void testClusterLevelNativeReplicationConfigForNewStores() {
           "dc-hybrid");
     });
   }
+
+  @Test(timeOut = TEST_TIMEOUT)
+  public void testConvertHybridDuringPushjob() {
+    String storeName = Utils.getUniqueString("test-store");
+    parentControllerClient.createNewStore(storeName, "test-owner", "\"string\"", "\"string\"");
+    parentControllerClient.requestTopicForWrites(
```
huangminchn, line 113: Thanks for adding the test! We didn't kill this push; would it affect other tests in the same file, since we are using the same store name here.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/controller/TestClusterLevelConfigForActiveActiveReplication.java
majisourav99: Wait will make sure that the push has finished, then we can convert it to hybrid.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/controller/TestClusterLevelConfigForNativeReplication.java
majisourav99: Sure, good catch

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/controller/TestClusterLevelConfigForActiveActiveReplication.java
huangminchn: I see! Thanks for clarification.

huangminchn (approved):
Thanks a lot Sourav!

PR #1145, opened 2024-08-28
[server][router] flush latency/response size metrics in Server and HB metric bug fix in Router

gaojieliu:

This PR added a few changes:
1. A new server config to tune the write buffer high watermark and a higher write buffer high watermark will allow writing more data at one time as H2 typically works with a smaller number of connections, and this change will affect HTTP/1.1 connections as well.
2. Flush latency/response size metrics in Server.
3. Fixed an issue in Router unhealthy host tracking during node swap. Essentially, if the node is not alive, Router will remove it from the unhealthy host tracking.

New server config:
server.channel.option.write.buffer.watermark.high.bytes: default 64KB (Netty default)

## How was this PR tested?
CI

### DIFF
--- a/services/venice-router/src/main/java/com/linkedin/venice/router/api/VeniceHostHealth.java
+++ b/services/venice-router/src/main/java/com/linkedin/venice/router/api/VeniceHostHealth.java
@@ -9,6 +9,7 @@ import com.linkedin.venice.router.stats.AggHostHealthStats;
 import com.linkedin.venice.router.stats.HostHealthStats;
 import com.linkedin.venice.router.stats.RouteHttpRequestStats;
 import com.linkedin.venice.utils.concurrent.VeniceConcurrentHashMap;
+import java.util.Iterator;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentSkipListSet;
@@ -59,6 +60,7 @@ public class VeniceHostHealth implements HostHealthMonitor<Instance> {
     String identifier = instance.getNodeId();
     unhealthyHosts.add(identifier);
     LOGGER.warn("Marking {} as unhealthy until it passes the next health check.", identifier);
+    cleanupUnhealthyHostsBasedOnLiveInstanceMonitor();
     aggHostHealthStats.recordUnhealthyHostCountCausedByRouterHeartBeat(unhealthyHosts.size());
   }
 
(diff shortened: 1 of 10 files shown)


services/venice-router/src/main/java/com/linkedin/venice/router/api/VeniceHostHealth.java
```diff
@@ -74,9 +76,24 @@ public void setHostAsHealthy(Instance hostname) {
       unhealthyHosts.remove(identifier);
```
FelixGV, line 76: Minor nitpick... there is no need to do `contains` followed by `remove`... we can just `remove` directly, and the return value is what was removed, if any. So the code can be:

```
if (unhealthyHosts.remove(identifier) != null) {
  // LOGGER ...
}
```

services/venice-server/src/main/java/com/linkedin/venice/listener/ListenerService.java
```diff
@@ -162,7 +163,14 @@ public ListenerService(
         .option(ChannelOption.SO_BACKLOG, nettyBacklogSize)
         .childOption(ChannelOption.SO_KEEPALIVE, true)
         .option(ChannelOption.SO_REUSEADDR, true)
-        .childOption(ChannelOption.TCP_NODELAY, true);
+        .childOption(ChannelOption.TCP_NODELAY, true)
+        // A higher buffer watermark will allow more data write for a given channel and it would be useful for H2,
+        // as H2 works with a smaller number of connections.
+        .childOption(
+            ChannelOption.WRITE_BUFFER_WATER_MARK,
```
FelixGV, line 170: Do you know what is the default value for this channel option?

I'm fine with changing Venice's default to 1 MB, but it would be good if the commit message included details about what the default was previously, in case we need to change it back...

Also, the commit message and code comment above mention that this change is useful for HTTP/2, but will it also affect HTTP/1?

FelixGV (commented):
Thanks Gaojie. The change is probably fine, but I left a minor comment (unrelated to your change) and a question. Please take a look.

services/venice-router/src/main/java/com/linkedin/venice/router/api/VeniceHostHealth.java
gaojieliu: Sure, I can make the change.

services/venice-server/src/main/java/com/linkedin/venice/listener/ListenerService.java
gaojieliu: The default value is 64KB.
I can mentioned that in the commit message.

This will affect HTTP/1 as well, and it can be more useful for H2.
Recently, we noticed a weird behavior as sometimes flush latency can go up high and go down after some period, so I would like to test whether this particular tuning is helpful or not.

gaojieliu: Maybe I should keep the default value and tune the config explicitly in the experimental env and if there is a need, we can change the default later on.

FelixGV (approved):
Thanks!

PR #1146, opened 2024-08-29
[fast-client][thin-client] Common headers and metrics logic between thin-client and fast-client

xunyin8:
## [fast-client][thin-client] Common headers and metrics logic between thin-client and fast-client

1. Fast-client and thin-client use similar headers when making requests to the router/servers. Abstracted the common logic to a static utility class `RequestHeadersProvider` to ensure optimizations like providing key count in request header for early quota rejection is also available for fast-client (this PR does not include the server change that performs the early quota rejection).

2. Although fast-client's `FastClientStats` extends `com.linkedin.venice.client.stats.ClientStats`, we currently have two similar yet different approaches when it comes to metrics emission. Thin-client emits metrics using `TrackingStreamingCallback` which wraps the user callback, specifically the `StatTrackingStreamingCallback` implementation in `StatTrackingStoreClient` together with implementations of `RecordStreamDecoder`. Fast-client on the other hand record some stats in `RequestContext` and adds a handle stage to the user callback in `StatsAvroGenericStoreClient` to perform the actual metrics emission. Today there is no true streaming between fast-client and server but ideally we could have an `AbstractRecordStreamDecoder` with common logic such as metrics emission so it can be used by fast-client as well so it doesn't try to add its own metric emission handle stage (which had gaps) to the user callback in `StatsAvroGenericStoreClient`.
For now the missing streaming response percentile latency metric is added to fast-client using existing methods by abstracting the common logic to an utility class `StreamingResponseTracker`.

## How was this PR tested?
New and existing unit and integration tests

### DIFF
--- a/clients/venice-client/src/main/java/com/linkedin/venice/fastclient/DispatchingAvroGenericStoreClient.java
+++ b/clients/venice-client/src/main/java/com/linkedin/venice/fastclient/DispatchingAvroGenericStoreClient.java
@@ -1,6 +1,5 @@
 package com.linkedin.venice.fastclient;
 
-import static com.linkedin.venice.HttpConstants.VENICE_COMPUTE_VALUE_SCHEMA_ID;
 import static com.linkedin.venice.client.store.AbstractAvroStoreClient.TYPE_COMPUTE;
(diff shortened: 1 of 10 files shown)


m-nagarajan (approved):
LGTM. Thanks @xunyin8

PR #1147, opened 2024-08-29
[dvc] Fixed DVC integration test resource leak

FelixGV:
There is a resource leak in `DaVinciClientTest::testBlobP2PTransferAmongDVC` which causes subsequent tests to fail. This is occasionally masked by the `test-retry` plugin, though the retry itself is flaky and so it sometimes fails to retry at all. In any case, it is better to not fail in the first place...

Also did some minor tweaks to the main code to make debugging issues easier in the future. This includes extra logging in `DaVinciBackend` and a slightly more precise exception message in `AvroGenericDaVinciClient`.

## How was this PR tested?
Existing tests are deflaked.

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/AvroGenericDaVinciClient.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/AvroGenericDaVinciClient.java
@@ -725,8 +725,10 @@ public class AvroGenericDaVinciClient<K, V> implements DaVinciClient<K, V>, Avro
             backend -> {
               // Ensure that existing backend is fully closed before a new one can be created.
               synchronized (AvroGenericDaVinciClient.class) {
+                logger.info("Start of " + this.getClass().getSimpleName() + "'s ref counted deleter closure.");
                 daVinciBackend = null;
                 backend.close();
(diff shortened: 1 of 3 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/AvroGenericDaVinciClient.java
```diff
@@ -766,9 +768,7 @@ public synchronized void start() {
         daVinciConfig::getRecordTransformer);
 
     try {
-      if (!getBackend().compareCacheConfig(cacheConfig)) {
-        throw new VeniceClientException("Cache config conflicts with existing backend, storeName=" + getStoreName());
```
mynameborat, line 770: Can we clean up `compareCacheConfig` method?

mynameborat (commented):
Thanks for the fix!

clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/AvroGenericDaVinciClient.java
FelixGV: Oh yeah for sure. Done in the latest commit. Thanks!

mynameborat (approved):

PR #1148, opened 2024-08-29
[router][samza] Harden up quota monitor requests

ZacAttack:
## [router][samza] Harden up quota monitor requests
Main fix is that there exists a condition in the router where during a store move, the store will have information in the routers internal store config repository, but it won't have it in the store repository after the store was deleted at the end of the store move.  This would then trigger an exception and 500 error, and would prevent producers from reinitializing their clients. This fix addresses that.

Added tests in the store migration test suite to include a running producer.  Also added some protection in the code against empty responses (which was encountered while running the debugger).

Resolves #XXX

### DIFF
--- a/clients/venice-samza/src/main/java/com/linkedin/venice/pushmonitor/RouterBasedHybridStoreQuotaMonitor.java
+++ b/clients/venice-samza/src/main/java/com/linkedin/venice/pushmonitor/RouterBasedHybridStoreQuotaMonitor.java
@@ -124,6 +124,10 @@ public class RouterBasedHybridStoreQuotaMonitor implements Closeable {
       // Get hybrid store quota status
       CompletableFuture<TransportClientResponse> responseFuture = transportClient.get(requestPath);
       TransportClientResponse response = responseFuture.get(POLL_TIMEOUT_MS, TimeUnit.MILLISECONDS);
+      if (response == null) {
+        LOGGER.error("Router was not able to get hybrid quota status! Received null response!");
+        return;
+      }
       HybridStoreQuotaStatusResponse quotaStatusResponse =
           mapper.readValue(response.getBody(), HybridStoreQuotaStatusResponse.class);
       if (quotaStatusResponse.isError()) {
--- a/internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestStoreMigration.java
+++ b/internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestStoreMigration.java
@@ -240,6 +240,13 @@ public class TestStoreMigration {
     String storeName = Utils.getUniqueString("testWithMetaSystemStore");
     createAndPushStore(srcClusterName, storeName);
 
+    // Write streaming records
+    SystemProducer veniceProducer =
+        getSamzaProducer(multiClusterWrapper.getClusters().get(srcClusterName), storeName, Version.PushType.STREAM);
+    for (int i = 11; i <= 20; i++) {
+      sendStreamingRecord(veniceProducer, storeName, i);
+    }
+
     // Meta system store is enabled by default. Check if it has online version.
     try (ControllerClient srcChildControllerClient = new ControllerClient(srcClusterName, childControllerUrl0)) {
       String systemStoreName = VeniceSystemStoreType.META_STORE.getSystemStoreName(storeName);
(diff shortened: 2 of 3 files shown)


clients/venice-samza/src/main/java/com/linkedin/venice/pushmonitor/RouterBasedHybridStoreQuotaMonitor.java
```diff
@@ -124,6 +124,10 @@ protected void checkStatus() throws ExecutionException, InterruptedException, Ti
       // Get hybrid store quota status
       CompletableFuture<TransportClientResponse> responseFuture = transportClient.get(requestPath);
       TransportClientResponse response = responseFuture.get(POLL_TIMEOUT_MS, TimeUnit.MILLISECONDS);
+      if (response == null) {
+        LOGGER.error("Router was not able to get hybrid quota status! Received null response!");
+        return;
+      }
```
mynameborat, line 130: What is the behavior with early bailout? Would we skip any quota enforcement or block the push or let the caller decide?

services/venice-router/src/main/java/com/linkedin/venice/router/MetaDataHandler.java
```diff
@@ -670,13 +670,14 @@ private void handleStreamHybridStoreQuotaStatusLookup(ChannelHandlerContext ctx,
       throws IOException {
     String storeName = helper.getResourceName();
     checkResourceName(storeName, "/" + TYPE_STREAM_HYBRID_STORE_QUOTA + "/${storeName}");
-    if (!storeConfigRepo.getStoreConfig(storeName).isPresent()) {
+    Store store = storeRepository.getStore(storeName);
```
mynameborat, line 673: Will the store config repository be cleaned eventually? Wondering if there is something more fundamental in the way store deletion occurs where the associated state isn't cleaned up consistently?

xunyin8: Store config and store repository are different entities.
Store config is one per fabric and it holds cluster discovery and migration information. i.e. current cluster, is store migrating, source cluster and destination cluster.
Store repo is one per fabric per cluster and it holds the store's config and states (versions) in that cluster.
IIUC the scenario Zac is trying to solve here is when the store is deleted from the source cluster because store migration is completed and end migration is called to cleanup the store from source cluster. The store will continue to exist in the store repo of the new cluster and the store config will exist and be updated with the dest cluster.

mynameborat: I see. Thanks for the explanation :)

clients/venice-samza/src/main/java/com/linkedin/venice/pushmonitor/RouterBasedHybridStoreQuotaMonitor.java
ZacAttack: The current behavior (sans this change even) on any error for failing to ascertain the quota enforcement state is to do nothing, it will allow the writer to continue in the current state.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestStoreMigration.java
```diff
@@ -240,6 +240,13 @@ public void testStoreMigrationWithMetaSystemStore() throws Exception {
     String storeName = Utils.getUniqueString("testWithMetaSystemStore");
     createAndPushStore(srcClusterName, storeName);
 
+    // Write streaming records
```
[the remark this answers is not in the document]
ZacAttack: It's already enabled in the store set up.

services/venice-router/src/main/java/com/linkedin/venice/router/MetaDataHandler.java
```diff
@@ -670,13 +670,14 @@ private void handleStreamHybridStoreQuotaStatusLookup(ChannelHandlerContext ctx,
...
+    if (!storeConfigRepo.getStoreConfig(storeName).isPresent() || store == null) {
```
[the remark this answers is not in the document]
ZacAttack: The producer internally has a client that upon receiving a NOT_FOUND error will reinitialize the client and do another round of store discovery.

xunyin8 (approved):
Thank you Zac!

mynameborat (approved):

PR #1149, opened 2024-08-30
 [controller] Remove older backup versions after retention period

majisourav99:
##  Remove older backup versions after retention period

If there are old lingering version and there are newer repush version in regular cadence, today Venice will not delete the old version as the `getLatestVersionPromoteToCurrentTimestamp` updated and never fall off retention period. This PR fixed that by checking version creation time and delete if its older than retention period.

### DIFF
--- a/services/venice-controller/src/main/java/com/linkedin/venice/controller/StoreBackupVersionCleanupService.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/controller/StoreBackupVersionCleanupService.java
@@ -281,8 +281,8 @@ public class StoreBackupVersionCleanupService extends AbstractVeniceService {
     // retention period.
     if (readyToBeRemovedVersions.isEmpty()) {
       for (Version version: versions) {
-        if (version.getNumber() < currentVersion && store.getLatestVersionPromoteToCurrentTimestamp()
-            + defaultBackupVersionRetentionMs < time.getMilliseconds()) {
+        if (version.getNumber() < currentVersion
+            && version.getCreatedTime() + defaultBackupVersionRetentionMs < time.getMilliseconds()) {
           readyToBeRemovedVersions.add(version);
         }
       }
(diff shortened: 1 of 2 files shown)


services/venice-controller/src/main/java/com/linkedin/venice/controller/StoreBackupVersionCleanupService.java
```diff
@@ -281,8 +281,8 @@ protected boolean cleanupBackupVersion(Store store, String clusterName) {
     // retention period.
     if (readyToBeRemovedVersions.isEmpty()) {
       for (Version version: versions) {
-        if (version.getNumber() < currentVersion && store.getLatestVersionPromoteToCurrentTimestamp()
-            + defaultBackupVersionRetentionMs < time.getMilliseconds()) {
+        if (version.getNumber() < currentVersion
+            && version.getCreatedTime() + defaultBackupVersionRetentionMs < time.getMilliseconds()) {
```
gaojieliu, line 285: I think this is problematic..
For example, one store is doing weekly pushes and the backup version will be deleted right away when the new version just becomes the current version.
I think this new logic should only apply to the stores, which are doing frequent repushes and for the stores which are not doing repushes, we should keep the current logic.

The change should be small:
```
if (version.getNumber() < currentVersion &&
    (repushSourceVersion > 0 && version.getCreatedTime() + defaultBackupVersionRetentionMs < time.getMilliseconds() 
     || repushSourceVersion <= 0 && store.getLatestVersionPromoteToCurrentTimestamp()
            + defaultBackupVersionRetentionMs < time.getMilliseconds())
 ) {
    readyToBeRemovedVersions.add(version);
}
```
We should add some javadoc around this new check.

nisargthakkar: Agree that this current check is problematic.

Where do we have the check of excluding version deletion for source version for repush? I think such logic would make more sense there

gaojieliu (approved):
Thanks for the fix.

PR #1152, opened 2024-09-03
[server][dvc] Server request/response handling overhaul

FelixGV:
This refactors many parts of server request and response handling, including how metrics are recorded.

The significant functional changes are:

1. Parallel batch gets no longer have a race condition in the way metrics are recorded. Previously, the metrics which aggregate a value for all keys in the query could be missing data points and thus under-report due to this race.

2. The configs to enable and tune parallel batch gets now apply to compute as well, which was not supported before.

3. K/V size profiling is now always enabled for single gets, since it is considered cheap enough for that workload. The config still controls whether it is enabled or not for batch get and compute.

Non-functional changes include:

1. The elimination of locking in parallel query processing.

2. The serialization of responses to bytes now happens inside each subtask, rather than sequentially after all subtasks are completed. These are then presented as a CompositeByteBuf, which may result in smaller allocations since there is no need for one large byte[] holding the whole response anymore.

3. The objects used to record stats are now specialized for each query type (single get, batch get, compute) as well as for whether K/V size profiling is enabled or not, leading to a reduction of memory footprint for workloads that don’t need certain stats. See the class hierarchy details below for more info.

4. As much as possible, when some behavior is fixed for the whole duration of the server’s runtime, a property is allocated to hold the closure to be executed according to the relevant config, rather than evaluating the immutable config in the hot path.

Below are some of the implementation details for achieving the above changes as well as to clean up some tech debt.

The ReadResponse is now split into two interfaces, each of which has its own hierarchy of implementations:

- ReadResponse, which holds the state required to achieve the functional goals of the server, as seen by the querier (i.e. populating the payload and headers). This is now moved from the DVC module to the server module. The following subclass hierarchy exists:

  - AbstractReadResponse: holds common functional state and behavior for all read requests.

    - SingleGetResponseWrapper: renamed from StorageResponseObject.

    - MultiKeyResponseWrapper: used for single-threaded batch get requests as well as subtasks of parallel batch get requests.

      - ComputeResponseWrapper: used for single-threaded compute requests as well as subtasks of parallel compute requests.

    - ParallelMultiKeyResponseWrapper: used in parallel batch get & compute requests to hold an array of MultiKeyResponseWrapper, each of which are used exclusively by one subtask. This class is then responsible for aggregating the results of all these wrappers.

- ReadResponseStats: holds the state required for recording metrics, after the response is flushed. This replaces the ReadResponse parameter passed into all chunking-related utilities, thus more clearly constraining the role of that parameter in those classes. The purpose of this interface is to accumulate stats, but it doesn’t offer any API for reading or using those stats (see below how that part is done). The following subclass hierarchy exists:

  - AbstractReadResponseStats: holds the common metrics state reported for all read responses. The class also implements ReadResponseStatsRecorder, a new interface which is responsible for recording the accumulated metric state. Each subclass is responsible for defining the recording logic pertaining to the extra state it carries.

    - MultiKeyResponseStats: holds the record count.

      - MultiGetResponseStatsWithSizeProfiling: holds K/V sizes for each record.

    - ComputeResponseStats: holds nine primitives used exclusively in read compute stats.

      - ComputeResponseStatsWithSizeProfiling: holds K/V sizes for each record.

- MultiKeyRouterRequestWrapper::getKeys now returns a List<K> rather than an Iterable<K>.

- DaVinciBackend::getStoreOrThrow now makes use of CHM::computeIfAbsent, rather than syncrhonized. This helped stabilize an integration test.

- Muted a common stacktrace in IsolatedIngestionUtils.

- ThinClientMetaStoreBasedRepository::getStoreMetaValue now has a lower timeout and retries, rather than trying only one request with the default timeout of 10 seconds. Total time spent should be the same.

## How was this PR tested?
CI, including some new tests.

The parallel batch get config now affects compute as well.

K/V size profiling will now always be enabled for single gets.


services/venice-server/src/test/java/com/linkedin/venice/listener/response/MultiKeyResponseWrapperTest.java
```diff
@@ -0,0 +1,75 @@
...
+public class MultiKeyResponseWrapperTest {
+  private static final int RECORD_COUNT = 10;
+  private static final int CHUNK_COUNT = 5;
+  private static final int CHUNK_SIZE = RECORD_COUNT / CHUNK_COUNT;
+  /** We always use the same value but it does not matter for the sake of this test... */
+  private static final ByteBuffer SERIALIZED_VALUE = ByteBuffer.wrap(new byte[] { 0, 1, 2, 3 });
+
+  @DataProvider(name = "responseWrapperProvider")
+  public static Object[][] responseWrapperProvider() {
+    MultiGetResponseWrapper multiGetResponseWrapper = new MultiGetResponseWrapper(RECORD_COUNT);
```
ZacAttack, line 25: very optional nit:  This is a bit of a wall of big scary words.  Could we add just a few comments to describe what the test cases are which are getting returned by this provider?

services/venice-server/src/main/java/com/linkedin/venice/listener/response/ParallelMultiKeyResponseWrapper.java
```diff
@@ -0,0 +1,133 @@
...
+
+    @Override
+    public void recordUnmergedMetrics(ServerHttpRequestStats stats) {
+      for (int i = 0; i < this.statsRecorders.length; i++) {
+        this.statsRecorders[i].recordUnmergedMetrics(stats);
+      }
+    }
+
+    @Override
+    public void merge(ReadResponseStatsRecorder other) {
```
xunyin8, line 130: Should we throw `UnsupportedOperationException` explicitly here? Since we never expect this to be called on a `ParallelMultiKeyResponseWrapper`.

services/venice-server/src/main/java/com/linkedin/venice/listener/response/ParallelMultiKeyResponseWrapper.java
```diff
@@ -0,0 +1,133 @@
...
+    @Override
+    public void recordMetrics(ServerHttpRequestStats stats) {
+      this.mergedStats.recordMetrics(stats);
+      recordUnmergedMetrics(stats);
+    }
+
+    @Override
+    public void recordUnmergedMetrics(ServerHttpRequestStats stats) {
+      for (int i = 0; i < this.statsRecorders.length; i++) {
+        this.statsRecorders[i].recordUnmergedMetrics(stats);
```
xunyin8, line 125: Not sure if I missed something but besides preference, any other reason to `recordUnmergedMetrics` for the first chunk as part of `recordMetrics` and have `statsRecorders` array to be off by one (excluding the first chunk)?

services/venice-server/src/test/java/com/linkedin/venice/listener/StorageReadRequestHandlerTest.java
```diff
@@ -192,6 +217,27 @@ public void cleanUp() {
...
+    final boolean configValue;
+
+    ParallelQueryProcessing(boolean configValue) {
+      this.configValue = configValue;
+    }
+  }
+
+  @DataProvider(name = "storageReadRequestHandlerParams")
+  public Object[][] storageReadRequestHandlerParams() {
+    return new Object[][] { { ParallelQueryProcessing.SEQUENTIAL, 10, ValueSize.SMALL_VALUE },
```
ZacAttack, line 235: For symmetry's sake, do we need 

ParallelQueryProcessing.SEQUENTIAL, 100, ValueSize.LARGE_VALUE
ParallelQueryProcessing.SEQUENTIAL, 100, ValueSize.SMALL_VALUE

Cases?

services/venice-server/src/test/java/com/linkedin/venice/listener/StorageReadRequestHandlerTest.java
```diff
@@ -318,6 +414,22 @@ public void testMultiGetNotUsingKeyBytes(Boolean isParallel) throws Exception {
...
       assertEquals(results.get(i), allValueStrings.get(i));
     }
+
+    /**
+     * The assertions below can catch an issue where metrics are inaccurate during parallel batch gets. This was due to
+     * a race condition which is now fixed. Unfortunately, the race is hard to hit, even on the buggy code, so the test
+     * succeeding is not a guarantee of the absence of a regression (but the test failing certainly indicates that).
+     *
+     * We are making the test fail reliably (in the presence of buggy stats handling) via a backdoor available through
+     * {@link ReadResponseLatencyInjector}.
```
ZacAttack, line 424: neat

xunyin8 (commented):
Thanks for the cleanup Felix!

services/venice-server/src/test/java/com/linkedin/venice/listener/response/MultiKeyResponseWrapperTest.java
FelixGV: Sure thing! I added some comments in a JavaDoc at the top of the function and in comments near the return statement.

services/venice-server/src/test/java/com/linkedin/venice/listener/StorageReadRequestHandlerTest.java
FelixGV: I don't think we "need" it, but why not :D ... I tweaked the params so that this is a recordCount param rather than chunkSize param to make it less error-prone, and added a couple extra permutations.

services/venice-server/src/test/java/com/linkedin/venice/listener/StorageReadRequestHandlerTest.java
FelixGV: I don't like it a lot to be honest, but I think the overhead should be negligible and it's good to have peace of mind 🤷 ...

services/venice-server/src/main/java/com/linkedin/venice/listener/response/ParallelMultiKeyResponseWrapper.java
FelixGV: The way I did it is definitely not the only way...

For example, another way could have been to keep a handle of all the statsRecorders in the array, and call `recordUnmergedMetrics` on all of them. As for the mergeable metrics, I could have created a new (empty) stats recorder and merge into that one. But a difficulty of doing it this way is that I don't necessarily have access to the constructor of the correct stats recorder type (I would have needed to wire it in, or else add one more function to the interface to get an empty instance).

I also had an earlier iteration where I added a getter for the storage execution submission wait time and I gathered an array of just these data points... this might have been slightly more lightweight but I thought the design was too brittle. Adding metrics would require more custom coding, which I thought was already a problem in the legacy code (adding a metric required adding getters and setters into `ReadResponse`, and `StatsContext` and copying each field one by one, and all that...).

So I tried to pick a general approach that would be easy to extend with new metrics. But it's certainly possible that there could be an even better way.

FelixGV:
Thanks for the reviews @ZacAttack and @xunyin8! I've addressed the comments and also rebased in the latest push.

services/venice-server/src/main/java/com/linkedin/venice/listener/response/ParallelMultiKeyResponseWrapper.java
xunyin8: I see, thanks for the detailed explanation. Was curious to understand the reasoning behind this specific approach.

xunyin8 (approved):

FelixGV:
Thanks for shipping the PR, @xunyin8! Unfortunately, the rebase had conflicts, and I had to force push again, which dismissed your 🚢 ...

ZacAttack (approved):

services/venice-server/src/test/java/com/linkedin/venice/listener/StorageReadRequestHandlerTest.java
FelixGV: Thanks for the review on PR 1152. I merged it yesterday, but this testing backdoor I left behind to reliably repro the race condition did not sit well with me... I took another stab at it here:

https://github.com/linkedin/venice/pull/1161

And I also threw in a few other cleanups that I had pulled out of the previous PR because that one was already big and I wanted to reduce its scope.

services/venice-server/src/main/java/com/linkedin/venice/listener/response/stats/ComputeResponseStats.java
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  @Override
+  public void addReadComputeLatency(double latency) {
+    this.readComputeLatency += latency;
+  }
+
+  @Override
+  public void addReadComputeDeserializationLatency(double latency) {
+    this.readComputeDeserializationLatency += latency;
```
sushantmane, line 32: This may not be `ThreadSafe` for parallel processing of compute requests, right?

PR #1154, opened 2024-09-04
[server] Refactor ReadQuotaEnforcementHandler to reduce code duplication across gRPC and Netty handlers

sushantmane:
## [server] Refactor ReadQuotaEnforcementHandler to reduce code duplication across gRPC and Netty handlers
- Extracted common quota logic into ReadQuotaEnforcementHandler::enforceQuota
- Kept handler-specific logic for gRPC and Netty in their respective methods
- Renamed GrpcErrorCodes to VeniceReadResponseStatus for shared use across HTTP and gRPC

## How was this PR tested?

CI


services/venice-server/src/main/java/com/linkedin/venice/listener/grpc/handlers/GrpcOutboundResponseHandler.java
```diff
@@ -53,7 +53,7 @@ public void processRequest(GrpcRequestContext ctx) {
     ctx.setError();
     statsContext.setResponseStatus(NOT_FOUND);
     veniceServerResponseBuilder.setData(ByteString.EMPTY);
-    veniceServerResponseBuilder.setErrorCode(GrpcErrorCodes.KEY_NOT_FOUND);
+    veniceServerResponseBuilder.setErrorCode(VeniceReadResponseStatus.KEY_NOT_FOUND);
```
xunyin8, line 56: Will this break anything? The old `GrpcErrorCodes.KEY_NOT_FOUND` has a value of 101 and the new value is now -420.

sushantmane: There are no production use cases yet, so this breaking change won't have any impact.

internal/venice-common/src/main/java/com/linkedin/venice/response/VeniceReadResponseStatus.java
```diff
@@ -0,0 +1,15 @@
+package com.linkedin.venice.response;
+
+/**
+ * Response status codes for Venice read requests.
+ */
+public class VeniceReadResponseStatus {
+  public static final int DEADLINE_EXCEEDED = -504;
```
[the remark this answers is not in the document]
sushantmane: Same here: >There are no production use cases yet, so this breaking change won't have any impact.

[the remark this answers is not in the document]
xunyin8: Let's remove this error code static field for now since it's not used anywhere.

services/venice-server/src/main/java/com/linkedin/venice/listener/grpc/handlers/GrpcOutboundResponseHandler.java
xunyin8: It might cause confusion later when we try to productionize GRPC, can we add a new static field in `VeniceReadResponseStatus` called `GRPC_KEY_NOT_FOUND` and set its value to 101?

sushantmane: >called GRPC_KEY_NOT_FOUND and set its value to 101?

QQ: Why do we need to call out gRPC in the var name?

xunyin8: You are right, we don't need to have separate error code for `KEY_NOT_FOUND` since we own the handling for this error end to end so as long as it doesn't conflict with any existing error code we should be fine.

internal/venice-common/src/main/java/com/linkedin/venice/response/VeniceReadResponseStatus.java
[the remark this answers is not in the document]
sushantmane: Done. Thanks!

xunyin8 (approved):
Thanks for the refactor.

sushantmane:
Thanks a lot, @xunyin8!

PR #1155, opened 2024-09-04
[vpj] Support AAWC TTL repush with compression

sixpluszero:
## [vpj] Support AAWC TTL repush with compression
Current AAWC TTL repush need to deserialize value and perform a TTL based delete on the record to see if it needs to be filtered. The code does not handle the case when source version has compression enabled thus deserialization can fail. This PR fixes it by building the source version compressor and fix the deserialization issue.

## How was this PR tested?
Add Compression Mode for AAWC TTL integration test.

### DIFF
--- a/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/input/kafka/ttl/VeniceRmdTTLFilter.java
+++ b/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/input/kafka/ttl/VeniceRmdTTLFilter.java
@@ -1,5 +1,8 @@
 package com.linkedin.venice.hadoop.input.kafka.ttl;
 
+import static com.linkedin.venice.hadoop.VenicePushJobConstants.KAFKA_INPUT_BROKER_URL;
+import static com.linkedin.venice.hadoop.VenicePushJobConstants.KAFKA_INPUT_SOURCE_COMPRESSION_STRATEGY;
+import static com.linkedin.venice.hadoop.VenicePushJobConstants.KAFKA_INPUT_TOPIC;
 import static com.linkedin.venice.hadoop.VenicePushJobConstants.REPUSH_TTL_POLICY;
 import static com.linkedin.venice.hadoop.VenicePushJobConstants.REPUSH_TTL_START_TIMESTAMP;
 import static com.linkedin.venice.hadoop.VenicePushJobConstants.RMD_SCHEMA_DIR;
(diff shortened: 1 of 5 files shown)


majisourav99 (commented):
Change looks good to me, can you add some tests with ZSTD compression ?

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/input/kafka/ttl/VeniceRmdTTLFilter.java
```diff
@@ -125,7 +152,11 @@ boolean filterByTTLandMaybeUpdateValue(final INPUT_VALUE value) {
         valueSerializerCache.computeIfAbsent(valueSchemaId, this::generateValueSerializer);
     RecordSerializer<GenericRecord> rmdSerializer =
         rmdSerializerCache.computeIfAbsent(rmdVersionId, this::generateRmdSerializer);
-    updateValuePayload(value, valueSerializer.serialize(valueRecord));
+    try {
+      updateValuePayload(value, sourceVersionCompressor.compress(valueSerializer.serialize(valueRecord)));
```
gaojieliu, line 156:
```
props.getBoolean(KAFKA_INPUT_COMPRESSION_BUILD_NEW_DICT_ENABLED, true);
```
Repush will always build a new dict for the repushed version, so shall we read the dictionary in the new version instead for the compression?

sixpluszero: Hmm, would that be available in the mapper / reducer phase? As far as I know this filter is invoked in both mapper and reducer today. 

Also, what'd be the difference for reading for the new dictionary? I am a bit new to ZSTD_WITH_DICT so I don't know that difference.
Also I am wondering, if dest version changed compression strategy, would it still work?

majisourav99: Using newer dictionary would give better compression, but not a must. If dest changed compression strategy, we should fail the repush?

gaojieliu: I think we can build a new compressor for the dest version as the regular push can get a proper compressor in the reducer phase.
You can check the related code.

gaojieliu (approved):
Synced offline, and we can optimize the multiple rounds of decompression/re-compression in a separate PR.

sixpluszero:
> Change looks good to me, can you add some tests with ZSTD compression ?

The newly added integration test does cover all compression

PR #1157, opened 2024-09-04
[server] SN read quota usage should return NaN for stores with no replicas

xunyin8:

If a store's partition count is small it's normal and expectd to have some hosts in the cluster to not have any replicas assigned to them. The hosts should emit NaN for the SN read quota usage metric instead of -1. Emitting -1 or 0 will mess up aggregation of metrics used for monitoring.

The NaN should be ignored by aggregators and provide clear signals in case of unexpected null rate limiter.

## How was this PR tested?

### DIFF
--- a/services/venice-server/src/main/java/com/linkedin/venice/stats/ServerQuotaUsageStats.java
+++ b/services/venice-server/src/main/java/com/linkedin/venice/stats/ServerQuotaUsageStats.java
@@ -65,7 +65,7 @@ public class ServerQuotaUsageStats extends AbstractVeniceStats {
    */
   private Double getReadQuotaUsageRatio() {
     if (tokenBucket == null) {
-      return -1d;
+      return Double.NaN;
     } else {
       return tokenBucket.getStaleUsageRatio();
     }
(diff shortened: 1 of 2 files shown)


services/venice-server/src/main/java/com/linkedin/venice/stats/ServerQuotaUsageStats.java
```diff
@@ -65,7 +65,7 @@ public void setTokenBucket(TokenBucket tokenBucket) {
    */
   private Double getReadQuotaUsageRatio() {
     if (tokenBucket == null) {
-      return -1d;
+      return Double.NaN;
```
majisourav99, line 68: Do we have a test to check if NaN is skipped by metric aggregator?

xunyin8: I'm still looking for answers for this one. Also wanted to make sure it doesn't cause any like performance issues for the aggregator.

xunyin8: Verified that the average aggregator will ignore NaN values, don't think we can have a meaningful integration test for it tho.

majisourav99 (approved):
Thanks Xun!

PR #1159, opened 2024-09-05
[server] AAWC should ignore NO_OP new field from update record

sixpluszero:
## [server] AAWC should ignore NO_OP new field from update record
If a field from update record is new field but it is NO_OP, we should skip this field, instead of letting it slip into checking TS from old TS record, which will result in exception and halt the ingestion

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/replication/merge/MergeConflictResolver.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/replication/merge/MergeConflictResolver.java
@@ -766,9 +766,11 @@ public class MergeConflictResolver {
       case PER_FIELD_TIMESTAMP:
         GenericRecord timestampRecord = (GenericRecord) oldTimestampObject;
         for (Schema.Field field: writeComputeRecord.getSchema().getFields()) {
-          if (getFieldOperationType(writeComputeRecord.get(field.pos())) != NO_OP_ON_FIELD
-              && timestampRecord.get(field.name()) == null) {
-            return false; // Write Compute tries to update a non-existing field.
+          if (timestampRecord.get(field.name()) == null) {
+            if (getFieldOperationType(writeComputeRecord.get(field.pos())) == NO_OP_ON_FIELD) {
+              continue; // New field does not perform actual update.
+            }
+            return false; // Partial update tries to update a non-existing field.
           }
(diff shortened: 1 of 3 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/replication/merge/MergeConflictResolver.java
```diff
@@ -766,9 +766,11 @@ private boolean ignoreNewUpdate(
       case PER_FIELD_TIMESTAMP:
         GenericRecord timestampRecord = (GenericRecord) oldTimestampObject;
```
majisourav99, line 767: unrelated nit: can we rename this as `oldTimestampRecord` ?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/replication/merge/MergeConflictResolver.java
```diff
@@ -766,9 +766,11 @@ private boolean ignoreNewUpdate(
...
         for (Schema.Field field: writeComputeRecord.getSchema().getFields()) {
-          if (getFieldOperationType(writeComputeRecord.get(field.pos())) != NO_OP_ON_FIELD
-              && timestampRecord.get(field.name()) == null) {
-            return false; // Write Compute tries to update a non-existing field.
+          if (!timestampRecord.hasField(field.name())) {
+            if (getFieldOperationType(writeComputeRecord.get(field.pos())) == NO_OP_ON_FIELD) {
+              continue; // New field does not perform actual update.
+            }
+            return false; // Partial update tries to update a non-existing field.
```
majisourav99, line 773: should this be true ie be ignored ?

sixpluszero: True means we should ignore the whole update.
continue means we ignore this field, but whether to ignore the whole UPDATE depends on the check of other fields.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/replication/merge/MergeConflictResolver.java
sixpluszero: Sure, I can update.

majisourav99 (approved):
Thanks a lot Jialin for the quick fix

PR #1160, opened 2024-09-05
[server] Make rate limiter configurable in quota enforcement handler

sushantmane:
## Make rate limiter configurable in quota enforcement handler
- Add configuration to allow swapping between different rate limiter types.
- Make quota enforcement interval and capacity (used in TokenBucket) configurable.
- Update TokenBucket to support refreshing at millisecond intervals.
- Some other cleanups to make code style consistent 

## How was this PR tested?

CI 


internal/venice-test-common/src/integrationTest/resources/log4j2.properties
sushantmane, line 1: Will delete it once done with code reviews

services/venice-server/src/main/java/com/linkedin/venice/listener/ReadQuotaEnforcementHandler.java
```diff
@@ -330,27 +307,77 @@ private void updateQuota(PartitionAssignment partitionAssignment) {
...
+    // If the rate limiter is already created and the quota is the same, return the existing rate limiter
+    long newQuota = (long) Math.ceil(quotaInRcu * thisNodeQuotaResponsibility);
+    if (currentRateLimiter != null && currentRateLimiter.getGlobalPermitsCount() == quotaInRcu
+        && currentRateLimiter.getThisNodeResponsibility() == thisNodeQuotaResponsibility) {
+      return currentRateLimiter;
+    }
+
+    VeniceRateLimiter newRateLimiter;
+    if (rateLimiterType == RateLimiterType.EVENT_THROTTLER_WITH_SILENT_REJECTION) {
+      LOGGER.info(
```
xunyin8, line 350: Let's not spam our logs with these `Creating x rate limiter for store version x with quota x`. If we want to know the latest rate limiter quota for a store version we can emit that as a metric. The `rateLimiterType` is a server config and won't change from store to store so if we want we can just emit it as a single log message when ReadQuotaEnforcementHandler is initialized.

services/venice-server/src/main/java/com/linkedin/venice/listener/ReadQuotaEnforcementHandler.java
```diff
@@ -330,27 +307,77 @@ private void updateQuota(PartitionAssignment partitionAssignment) {
...
+          new EventThrottler(newQuota, quotaEnforcementIntervalInMs, storeVersionName, true, REJECT_STRATEGY);
+    } else if (rateLimiterType == RateLimiterType.GUAVA_RATE_LIMITER) {
+      LOGGER.info(
+          "Creating GuavaRateLimiter rate limiter for store version: {} with quota: {}",
+          storeVersionName,
+          newQuota);
+      newRateLimiter = new GuavaRateLimiter(newQuota);
+    } else {
+      LOGGER.info(
+          "Creating TokenBucket rate limiter for store version: {} with quota: {} and thisNodeQuotaResponsibility: {} quotaEnforcementIntervalInMs: {} enforcementCapacityMultiple: {}",
```
xunyin8, line 364: Similar here, if we want to log about these specific configs regarding to token bucket type we can log it during initialization of `ReadQuotaEnforcementHandler` since these configs are also not changing from store to store.

internal/venice-common/src/main/java/com/linkedin/venice/throttle/VeniceRateLimiter.java
```diff
@@ -0,0 +1,26 @@
...
+  /**
+   * Helper methods used
+   */
+  void setGlobalPermitsCount(long totalPermits);
+
+  void setThisNodeResponsibility(double thisNodeResponsibility);
+
+  long getGlobalPermitsCount();
+
+  double getThisNodeResponsibility();
```
xunyin8, line 25: This feels a bit odd, why do the underlying rate limiter need to be aware of the node responsibility? If it's only used to determine if two rate limiter is the same (for deciding if we need to create a new rate limiter) then wouldn't be better to just have setter and getter on the post node responsibility permits count and use that for comparison instead of the global permits count + node responsibility?

xunyin8 (commented):
Thanks for the refactor/cleanup, left a few comments/concerns

services/venice-server/src/main/java/com/linkedin/venice/stats/ServerQuotaUsageStats.java
```diff
@@ -28,8 +23,6 @@ public ServerQuotaUsageStats(MetricsRepository metricsRepository, String name) {
     rejectedQPS = registerSensor("quota_rejected_request", new Rate());
     rejectedKPS = registerSensor("quota_rejected_key_count", new Rate());
     allowedUnintentionallyKPS = registerSensor("quota_unintentionally_allowed_key_count", new Count());
-    usageRatioSensor =
-        registerSensor(new AsyncGauge((ignored, ignored2) -> getReadQuotaUsageRatio(), "quota_requested_usage_ratio"));
```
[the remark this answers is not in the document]
sushantmane: I removed this metric for two reasons: 1) it seems it's not accurate since usage is usually per sec and the metric reporting happens at minute level 2) We can infer the same info using quota request rate & quota rejected rate

services/venice-server/src/main/java/com/linkedin/venice/listener/ReadQuotaEnforcementHandler.java
sushantmane: Good point

internal/venice-common/src/main/java/com/linkedin/venice/throttle/VeniceRateLimiter.java
sushantmane: Originally added this to help aide debugging but wha you're saying is good enough too

services/venice-server/src/main/java/com/linkedin/venice/stats/ServerQuotaUsageStats.java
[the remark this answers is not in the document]
sushantmane: Thanks Xun for the clarification. As discussed offline I added back quota usage ratio metric that is calculated using the formula: `quota_request_key_count / nodeQpsResponsibility`

xunyin8 (approved):

sushantmane:
Thanks a lot @xunyin8 for the review and discussions! 🙏

PR #1161, opened 2024-09-06
[server][controller][router] Minor cleanups

FelixGV:
PR #1152 added a special behavior to inject latency during the increment operation in AbstractReadResponseStats::incrementMultiChunkLargeValueCount, and this was controlled by a static boolean variable.

The current commit makes this cleaner by removing the special behavior from the main code, and instead using dependency injection in the tests only.

- Reduced the parameters required to construct a StorageReadRequestHandler.

- The HelixReadOnlySchemaRepository and HelixSchemaAccessor are now less chatty. In particular, listener registration prints one line instead of four.

- Muted the stacktrace for InterruptedExceptions in TopicCleanupService and DictionaryRetrievalService.

- Added a toString to D2Server.

Miscellaneous test changes:

- Added missing @Test annotation to ConsumerIntegrationTestWithProtocolHeader so now it should run as part of the suite.

- Tweaked the integration test bucketing, since bucket Z became the long tail again. Specifially:

  - Moved A/A tests from buckets C to D (where other A/A tests already lived).

  - Placed all consumer-related tests into bucket C (moved from D and Z).

## How was this PR tested?

CI.


internal/venice-common/src/test/java/com/linkedin/venice/helix/HelixReadOnlySchemaRepositoryTest.java
```diff
@@ -157,8 +158,8 @@ public void testForceRefreshSchemaData() {
     Assert.assertNull(schemaRepository.getSupersetSchema(storeName));
   }
 
-  @Test
-  public void getDerivedSchemaIdTest() {
+  @Test(dataProviderClass = DataProviderUtils.class, dataProvider = "Two-True-and-False")
```
sushantmane, line 161: Should we have assertions to check whether the code enabled by `wcEnabled` & `aaEnabled` was called or not?

sushantmane (commented):
Thanks, @FelixGV! Overall LGTM, Left just one minor comment

internal/venice-common/src/test/java/com/linkedin/venice/helix/HelixReadOnlySchemaRepositoryTest.java
sushantmane: Thanks

sushantmane (approved):
LGTM. Thanks, @FelixGV!

FelixGV:
Thanks @sushantmane!

PR #1162, opened 2024-09-09
[service][dvc] generate snapshot after endBatchWrite complete

jingy-li:
## Create snapshots after end batch write complete

1. After a batch write, create a partition snapshot. The snapshot will be under the directory in the "./snapshot" folder for each batch-push version. The subfiles under "./snapshot" will be used for blob transfer.
2. The blob transfer manager are initial for Service, not only DVC.

## How was this PR tested?

Run integration test at TestBatch. Snapshots are generated.

Screenshot 2024-09-06 at 1 34 31 PM

**[WIP]** plain table integration test will be added in next commit. 

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/BlobSnapshotManager.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/BlobSnapshotManager.java
@@ -161,4 +161,28 @@ public class BlobSnapshotManager {
     return Checkpoint.create(rocksDB);
   }
 
+  /**
+   * util method to create a snapshot for batch only
+   */
+  public static void createSnapshotForBatch(RocksDB rocksDB, String fullPathForPartitionDBSnapshot) {
+    LOGGER.info("Creating snapshot for batch in directory: {}", fullPathForPartitionDBSnapshot);
+
+    if (fullPathForPartitionDBSnapshot == null || fullPathForPartitionDBSnapshot.isEmpty()) {
+      LOGGER.error("Snapshot directory {} is null or empty", fullPathForPartitionDBSnapshot);
+      return;
+    }
+
+    try {
+      LOGGER.info("Start creating snapshots for batch in directory: {}", fullPathForPartitionDBSnapshot);
+
+      Checkpoint checkpoint = Checkpoint.create(rocksDB);
+      checkpoint.createCheckpoint(fullPathForPartitionDBSnapshot);
+
+      LOGGER.info("Finished creating snapshots for batch in directory: {}", fullPathForPartitionDBSnapshot);
+    } catch (RocksDBException e) {
+      throw new VeniceException(
+          "Received exception during RocksDB's snapshot creation in directory " + fullPathForPartitionDBSnapshot,
+          e);
+    }
+  }
 }
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBStoragePartition.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBStoragePartition.java
@@ -218,8 +218,7 @@ public class RocksDBStoragePartition extends AbstractStoragePartition {
           options,
           fullPathForTempSSTFileDir,
           false,
-          rocksDBServerConfig,
-          blobTransferEnabled);
+          rocksDBServerConfig);
     }
 
     /**
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBStoragePartition.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBStoragePartition.java
@@ -2,6 +2,7 @@ package com.linkedin.davinci.store.rocksdb;
 
 import static com.linkedin.davinci.store.AbstractStorageEngine.METADATA_PARTITION_ID;
 
+import com.linkedin.davinci.blobtransfer.BlobSnapshotManager;
 import com.linkedin.davinci.callback.BytesStreamingCallback;
 import com.linkedin.davinci.config.VeniceStoreVersionConfig;
 import com.linkedin.davinci.stats.RocksDBMemoryStats;
@@ -484,16 +485,12 @@ public class RocksDBStoragePartition extends AbstractStoragePartition {
      * the last SST file written is finished.
      */
     rocksDBSstFileWriter.ingestSSTFiles(rocksDB, columnFamilyHandleList);
-
-    if (blobTransferEnabled) {
-      createSnapshot();
-    }
   }
 
   @Override
   public synchronized void createSnapshot() {
     if (blobTransferEnabled) {
-      rocksDBSstFileWriter.createSnapshot(rocksDB);
+      BlobSnapshotManager.createSnapshotForBatch(rocksDB, fullPathForTempSnapshotFileDir);
     }
   }
 
--- a/clients/da-vinci-client/src/test/java/com/linkedin/davinci/store/rocksdb/RocksDBSstFileWriterTest.java
+++ b/clients/da-vinci-client/src/test/java/com/linkedin/davinci/store/rocksdb/RocksDBSstFileWriterTest.java
@@ -1,18 +1,8 @@
 package com.linkedin.davinci.store.rocksdb;
 
-import static org.mockito.Mockito.doNothing;
-import static org.mockito.Mockito.doReturn;
-import static org.mockito.Mockito.doThrow;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.spy;
-import static org.mockito.Mockito.times;
-import static org.mockito.Mockito.verify;
-import static org.testng.Assert.assertThrows;
-
 import com.linkedin.venice.exceptions.VeniceException;
 import com.linkedin.venice.kafka.validation.checksum.CheckSum;
 import com.linkedin.venice.kafka.validation.checksum.CheckSumType;
-import com.linkedin.venice.store.rocksdb.RocksDBUtils;
 import com.linkedin.venice.utils.Utils;
 import com.linkedin.venice.utils.VeniceProperties;
 import java.io.File;
@@ -22,10 +12,8 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.Optional;
 import org.apache.commons.io.FileUtils;
-import org.rocksdb.Checkpoint;
 import org.rocksdb.EnvOptions;
 import org.rocksdb.Options;
-import org.rocksdb.RocksDB;
 import org.rocksdb.RocksDBException;
 import org.testng.Assert;
 import org.testng.annotations.Test;
@@ -36,7 +24,6 @@ public class RocksDBSstFileWriterTest {
   private static final int PARTITION_ID = 0;
   private static final String DB_DIR = Utils.getUniqueTempPath("sstTest");
   private static final boolean IS_RMD = false;
-  private static final boolean IS_BLOB_TRANSFER_ENABLED = false;
   private static final RocksDBServerConfig ROCKS_DB_SERVER_CONFIG = new RocksDBServerConfig(VeniceProperties.empty());
 
   @Test
@@ -51,8 +38,7 @@ public class RocksDBSstFileWriterTest {
           new Options(),
           DB_DIR,
           IS_RMD,
-          ROCKS_DB_SERVER_CONFIG,
-          IS_BLOB_TRANSFER_ENABLED);
+          ROCKS_DB_SERVER_CONFIG);
       Map<String, String> checkpointedInfo = new HashMap<>();
 
       // Expect false as that file is not found
@@ -76,8 +62,7 @@ public class RocksDBSstFileWriterTest {
           new Options(),
           DB_DIR,
           IS_RMD,
-          ROCKS_DB_SERVER_CONFIG,
-          IS_BLOB_TRANSFER_ENABLED);
+          ROCKS_DB_SERVER_CONFIG);
       Map<String, String> checkpointedInfo = new HashMap<>();
 
       // Checkpoint that 1 sst file should be found
@@ -110,8 +95,7 @@ public class RocksDBSstFileWriterTest {
           new Options(),
           DB_DIR,
           IS_RMD,
-          ROCKS_DB_SERVER_CONFIG,
-          IS_BLOB_TRANSFER_ENABLED);
+          ROCKS_DB_SERVER_CONFIG);
       Map<String, String> checkpointedInfo = new HashMap<>();
 
       // Checkpoint that 2 sst file should be found
@@ -138,8 +122,7 @@ public class RocksDBSstFileWriterTest {
           new Options(),
           DB_DIR,
           IS_RMD,
-          ROCKS_DB_SERVER_CONFIG,
-          IS_BLOB_TRANSFER_ENABLED);
+          ROCKS_DB_SERVER_CONFIG);
       Map<String, String> checkpointedInfo = new HashMap<>();
 
       // Checkpoint -1: invalid
@@ -165,8 +148,7 @@ public class RocksDBSstFileWriterTest {
           new Options(),
           DB_DIR,
           IS_RMD,
-          ROCKS_DB_SERVER_CONFIG,
-          IS_BLOB_TRANSFER_ENABLED);
+          ROCKS_DB_SERVER_CONFIG);
       Map<String, String> checkpointedInfo = new HashMap<>();
 
       // Checkpoint that 6 sst file should be found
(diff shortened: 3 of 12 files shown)


FelixGV:
Nitpick: I imagine you intend to prefix the commit message with `[server]` rather than `[service]`?

sixpluszero (commented):
Thanks for putting up this PR. I think the core logic is good to me, just have a comment on the snapshot creation util.

clients/da-vinci-client/src/test/java/com/linkedin/davinci/store/rocksdb/RocksDBSstFileWriterTest.java
```diff
@@ -289,108 +268,6 @@ public void testOpenAndSyncWithAllValidFiles() throws IOException {
     }
   }
 
-  @Test
-  public void testCreateSnapshotWithBlobTransferDisabled() throws RocksDBException {
```
adamxchen, line 293: Since we are moving codes hence deleting these codes from old classes, can we also make sure they are added to the new classes?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBStoragePartition.java
```diff
@@ -484,16 +484,12 @@ public synchronized void endBatchWrite() {
...
-    if (blobTransferEnabled) {
-      createSnapshot();
-    }
   }
 
   @Override
   public synchronized void createSnapshot() {
     if (blobTransferEnabled) {
-      rocksDBSstFileWriter.createSnapshot(rocksDB);
+      BlobSnapshotManager.createSnapshotForBatch(rocksDB, fullPathForTempSnapshotFileDir);
```
adamxchen, line 492: I don't recall it needed `fullPathForTempSnapshotFileDir` and `fullPathForPartitionDBSnapshot`. Looks like they are thee same so we are eliminating one of them, correct?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/BlobSnapshotManager.java
```diff
@@ -161,4 +161,28 @@ protected Checkpoint createCheckpoint(RocksDB rocksDB) {
     return Checkpoint.create(rocksDB);
   }
 
+  /**
+   * util method to create a snapshot for batch only
+   */
+  public static void createSnapshotForBatch(RocksDB rocksDB, String fullPathForPartitionDBSnapshot) {
```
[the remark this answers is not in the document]
adamxchen: +1 what jialin suggested. It's great to make `BlobSnapshotManager` a central class to manage rocksdb blob creation so if we can make it generic w/o significant change, let's do so.

> Also, why it is not acceptable when the path is empty? I think empty indicates there is nothing there and we can create stuff there.

This is copied from the original logic i believe. The original logic would create snapshot directory first before creating the snapshot. If the directory isn't present, it means either the `blobTransferEnabled` feature is not enabled or there's some disk issue hence skipping the snapshotting.

adamxchen (commented):
Great work! Please take a look at comments

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBStoragePartition.java
adamxchen: We have some duplicated logics here, not your changes but from previous changes. We may address them too.

In `RocksDBStoragePartition` constructor, we first check whether `blobTransferEnabled` is enabled or not to determine whether we need to create `fullPathForTempSnapshotFileDir`. In this method, we check this flag again to determine whether we need to run `BlobSnapshotManager.createSnapshotForBatch`. In the `createSnapshotForBatch` method, we check whether `fullPathForTempSnapshotFileDir` is present or not which is actually already guarded by the logic in `RocksDBStoragePartition` constructor. I feel we could remove one of these duplicated checks.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/BlobSnapshotManager.java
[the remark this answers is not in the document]
jingy-li: `do you think you want to limit the scope first and then expand to hybrid? ` Yes, we will extend the scope to hybrid later. Will rename it to `createSnapshot` once it can be leveraged by hybrid mode. But for distinguishing, lets name it createSnapshotForBatch for now. 

`why it is not acceptable when the path is empty? should we delete everything if the folder is not clean? ` This logic is copy from the original logic. I think previously it checks the path string is a empty string. (@adamxchen Seem it is not check the directory presence, fullPathForPartitionDBSnapshot here is a string) It is a good point that we should clean old snapshot.

internal/venice-common/src/main/java/com/linkedin/venice/blobtransfer/BlobTransferUtil.java
```diff
@@ -45,7 +47,34 @@ public static BlobTransferManager<Void> getP2PBlobTransferManagerAndStart(
...
+
+  /**
+   * Get a P2P blob transfer manager for Service and start it.
+   * @param p2pTransferServerPort the port used by the P2P transfer service
+   * @param p2pTransferClientPort the port used by the P2P transfer client
+   * @param baseDir the base directory of the underlying storage
+   * @param customizedViewFuture the future of the customized view repository
+   * @return the blob transfer manager
+   */
+  public static BlobTransferManager<Void> getP2PBlobTransferManagerForServiceAndStart(
```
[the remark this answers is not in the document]
jingy-li: Updated in next commit.

internal/venice-common/src/main/java/com/linkedin/venice/blobtransfer/BlobTransferUtil.java
```diff
@@ -45,7 +47,34 @@ public static BlobTransferManager<Void> getP2PBlobTransferManagerAndStart(
...
+    try {
+      BlobTransferManager<Void> manager = new NettyP2PBlobTransferManager(
+          new P2PBlobTransferService(p2pTransferServerPort, baseDir),
+          new NettyFileTransferClient(p2pTransferClientPort, baseDir),
+          new ServerBlobFinder(customizedViewFuture));
+      manager.start();
+      return manager;
+    } catch (Exception e) {
+      // swallow the exception and continue the consumption via pubsub system
+      LOGGER.warn("Failed to start up the P2P blob transfer manager for service", e);
```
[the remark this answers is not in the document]
jingy-li: Updated in next commit.

clients/da-vinci-client/src/test/java/com/linkedin/davinci/store/rocksdb/RocksDBSstFileWriterTest.java
jingy-li: Move the unit test to RocksDBStoragePartitionTest to test the blob transfer config.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBStoragePartition.java
jingy-li: Yes, lets only use fullPathForPartitionDBSnapshot. 
It is a good point! lets remove the path check at createSnapshotForBatch.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/BlobSnapshotManager.java
```diff
@@ -163,13 +163,19 @@ protected Checkpoint createCheckpoint(RocksDB rocksDB) {
...
     LOGGER.info("Creating snapshot for batch in directory: {}", fullPathForPartitionDBSnapshot);
 
-    if (fullPathForPartitionDBSnapshot == null || fullPathForPartitionDBSnapshot.isEmpty()) {
-      LOGGER.error("Snapshot directory {} is null or empty", fullPathForPartitionDBSnapshot);
-      return;
+    // clean up the snapshot directory if it exists
+    File partitionSnapshotDir = new File(fullPathForPartitionDBSnapshot);
+    if (partitionSnapshotDir.exists()) {
+      LOGGER.info("Snapshot directory already exists, deleting old snapshots at {}", fullPathForPartitionDBSnapshot);
+      if (!partitionSnapshotDir.delete()) {
```
adamxchen, line 175: I'm a little confused on this. `File#delete` would only work for an empty directory. Do you want to delete the empty directory or the contents of it?

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/DaVinciClientTest.java
```diff
@@ -1107,6 +1108,47 @@ public void testBlobP2PTransferAmongDVC() throws Exception {
     }
   }
 
+  /**
+   * Test to verify the snapshot generation for plain table.
+   */
+  @Test(timeOut = 2 * TEST_TIMEOUT)
+  public void testDVCSnapshotGenerationForPT() throws Exception {
```
adamxchen, line 1115: we can use `DataProviderUtils#trueAndFalseProvider` to control whether this client uses `MEMORY_BACKED_BY_DISK` or `DISK`

clients/da-vinci-client/src/test/java/com/linkedin/davinci/store/rocksdb/RocksDBStoragePartitionTest.java
```diff
@@ -902,4 +904,83 @@ public void checkMemoryLimitAtDatabaseOpen() {
       removeDir(storeDir);
     }
   }
+
+  @Test
+  public void testCreateSnapshotWithBlobTransferEnable() {
```
adamxchen, line 909: nit: `testCreateSnapshotWithBlobTransferEnabled`

clients/da-vinci-client/src/test/java/com/linkedin/davinci/store/rocksdb/RocksDBStoragePartitionTest.java
```diff
@@ -902,4 +904,83 @@ public void checkMemoryLimitAtDatabaseOpen() {
...
+    String storeName = Version.composeKafkaTopic(Utils.getUniqueString("test_store"), 1);
```
adamxchen, line 910: nit: maybe we can reuse some codes for these two methods so there's less duplicated codes

clients/da-vinci-client/src/test/java/com/linkedin/davinci/store/rocksdb/RocksDBSstFileWriterTest.java
adamxchen: Hmm. I think it makes more sense to have them in `BlobSnapshotManagerTest` since the method `createSnapshotForBatch` is in the `BlobSnapshotManager`. Having them in `RocksDBStoragePartitionTest` also works but that looks like a bit convoluted given we have to prepare many things for testing. What do you think?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/BlobSnapshotManager.java
jingy-li: good call! should not use File#delete, use FileUtils#deleteDirectory

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/BlobSnapshotManager.java
```diff
@@ -172,9 +174,12 @@ public static void createSnapshotForBatch(RocksDB rocksDB, String fullPathForPar
     File partitionSnapshotDir = new File(fullPathForPartitionDBSnapshot);
     if (partitionSnapshotDir.exists()) {
       LOGGER.info("Snapshot directory already exists, deleting old snapshots at {}", fullPathForPartitionDBSnapshot);
-      if (!partitionSnapshotDir.delete()) {
-        throw new VeniceException(
-            "Failed to delete the existing snapshot directory: " + fullPathForPartitionDBSnapshot);
+      try {
+        FileUtils.deleteDirectory(partitionSnapshotDir);
+      } catch (IOException e) {
+        throw new RuntimeException(
```
adamxchen, line 180: Can we use `VeniceException`?

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/DaVinciClientTest.java
```diff
@@ -1109,10 +1110,10 @@ public void testBlobP2PTransferAmongDVC() throws Exception {
   }
 
   /**
-   * Test to verify the snapshot generation for plain table.
+   * Test to verify the snapshot generation
    */
-  @Test(timeOut = 2 * TEST_TIMEOUT)
-  public void testDVCSnapshotGenerationForPT() throws Exception {
+  @Test(dataProviderClass = DataProviderUtils.class, dataProvider = "True-and-False", timeOut = 2 * TEST_TIMEOUT)
+  public void testDVCSnapshotGeneration(boolean useDiskStorage) throws Exception {
```
adamxchen, line 1116: nit: Nice! Do you think we can delete `testBlobP2PTransferAmongDVC` or rename one to another? Both integ tests have many similarity so I'd like to have only one of them. If you'd like to do it later, that's okay too.

jingy-li: Lets keep testBlobP2PTransferAmongDVC, as it contains some logic related to transfer from DVC1 to DVC2, which should can be reused when we test transfer stage.

clients/da-vinci-client/src/test/java/com/linkedin/davinci/store/rocksdb/RocksDBSstFileWriterTest.java
jingy-li: added in both BlobSnapshotManagerTest/RocksDBStoragePartitionTest

clients/da-vinci-client/src/test/java/com/linkedin/davinci/blobtransfer/BlobSnapshotManagerTest.java
```diff
@@ -140,4 +145,42 @@ public void testSnapshotNotUpdatedWhenNotHybrid() throws RocksDBException {
...
+            spy(new BlobSnapshotManager(BASE_PATH, SNAPSHOT_RETENTION_TIME, readOnlyStoreRepository));
+        File file = spy(new File(fullSnapshotPath));
+        doReturn(false).when(file).exists();
+
+        doNothing().when(mockCheckpoint).createCheckpoint(fullSnapshotPath);
+
+        // test execute
+        blobSnapshotManager.createSnapshotForBatch(mockRocksDB, fullSnapshotPath);
+        // test verify
+        verify(mockCheckpoint, times(1)).createCheckpoint(fullSnapshotPath);
```
adamxchen, line 170: If the file doesn't exist, we'd like to verify the `deleteDirectory` is not called at all. Are we able to verify that?

clients/da-vinci-client/src/test/java/com/linkedin/davinci/blobtransfer/BlobSnapshotManagerTest.java
```diff
@@ -140,4 +145,42 @@ public void testSnapshotNotUpdatedWhenNotHybrid() throws RocksDBException {
...
+
+        // case 2: snapshot file exists
+        // test prepare
+        doReturn(true).when(file).exists();
+        fileUtilsMockedStatic.when(() -> FileUtils.deleteDirectory(file)).thenAnswer(invocation -> {
```
adamxchen, line 175: Just want to make sure I understand, this line means `FileUtils.deleteDirectory` would do nothing but does it verify we have called it once?

clients/da-vinci-client/src/test/java/com/linkedin/davinci/blobtransfer/BlobSnapshotManagerTest.java
```diff
@@ -140,4 +145,42 @@ public void testSnapshotNotUpdatedWhenNotHybrid() throws RocksDBException {
...
+          return null;
+        });
+
+        // test execute
+        blobSnapshotManager.createSnapshotForBatch(mockRocksDB, fullSnapshotPath);
+        // test verify
+        verify(mockCheckpoint, times(2)).createCheckpoint(fullSnapshotPath);
+      }
```
adamxchen, line 183: Can I ask for adding a couple more scenarios where the exception could be thrown when failed to delete snapshot directory and failed to create rocksdb snapshot?

PR #1163, opened 2024-09-09
[server] Rewrite gRPC read Service to align with Netty-Based HTTP Server

sushantmane:
## Rewrite gRPC read Service to align with Netty-Based HTTP Server
- Refactored the Netty handler to enable code reuse for gRPC handlers 
- Added missing APIs to the gRPC service to close feature gaps with the 
  HTTP implementation.
- Fixed the gRPC I/O request path threading model: previously, requests 
  were handled by the gRPC executor thread pool, but now they are processed  
  by the `StorageReadExecutor` thread pool for better performance.  
  - Deprecated the old `get` and `multiGet` gRPC services, which relied on 
  a single schema and encouraged tightly coupled code. This refactor paves 
  the way for implementing server-side streaming support for batch get 
  requests more easily.
- Simplified the request processing logic by replacing the chain of
 responsibility pattern with a single processor class.

## How was this PR tested?

WIP (UTs and E2E will be added)


internal/venice-common/src/main/java/com/linkedin/venice/response/VeniceReadResponseStatus.java
```diff
@@ -1,19 +1,40 @@
...
+  public HttpResponseStatus getHttpResponseStatus() {
+    return httpResponseStatus;
+  }
+
+  public int getCode() {
+    return httpResponseStatus.code();
+  }
+
+  public static VeniceReadResponseStatus fromCode(int code) {
+    for (VeniceReadResponseStatus status: values()) {
```
FelixGV, line 33: We want to avoid calling `values()` because it allocates a new array each time. A few alternatives:

1. You can allocate a static property which contains the result of `values()`, and iterate over the static property here.
2. You can change this function into a switch/case (would need to be manually maintained if new enum values are added, though that is probably not a big deal, and this expectation could be checked trivially via a unit test).

internal/venice-common/src/main/proto/VeniceReadService.proto
```diff
@@ -29,4 +54,126 @@ message VeniceServerResponse {
...
+  optional string errorMessage = 5;
+  string contentType = 6;
+  uint32 contentLength = 7;
+  uint32 rcu = 8;
+}
+
+message MultiGetRequest {
+  string resourceName = 1;
+  uint32 partition = 2;
+  bytes  keyBytes = 3;            // used for batch get
```
FelixGV, line 80: Does this contain the whole slew of keys, concatenated one after the other? Should we use a collection type instead?

internal/venice-common/src/main/proto/VeniceReadService.proto
```diff
@@ -29,4 +54,126 @@ message VeniceServerResponse {
...
+  optional string errorMessage = 5;
+  string contentType = 6;
+  uint32 contentLength = 7;
+  uint32 rcu = 8;
+}
+
+
+message ComputeRequest {
+  string resourceName = 1;
+  bytes payload = 2;
```
FelixGV, line 99: Same question as above...

services/venice-server/src/main/java/com/linkedin/venice/grpc/GrpcIoRequestProcessor.java
```diff
@@ -0,0 +1,212 @@
...
+  public void processRequest(GrpcRequestContext requestContext) {
+    RouterRequest request = requestContext.getRouterRequest();
+    QuotaEnforcementResult result =
+        readQuotaEnforcementHandler != null ? readQuotaEnforcementHandler.enforceQuota(request) : null;
+    // If the request is allowed, hand it off to the storage read request handler
+    if (result == null || result == ALLOWED) {
+      GrpcStorageResponseHandlerCallback callback = GrpcStorageResponseHandlerCallback.create(requestContext);
+      storageReadRequestHandler.queueIoRequestForAsyncProcessing(request, callback);
+      return;
+    }
```
FelixGV, line 49: Instead of these null checks, can we make the `ReadQuotaEnforcementHandler` never null, meaning it's either going to contain a real handler, or else a "no op quota enforcement handler" which always returns `ALLOWED`? This behavior is decided at construction time and will never change afterwards, so there is no point in having the conditional checks on the hot path. Being never null also makes the code more readable (IMHO).

services/venice-server/src/main/java/com/linkedin/venice/grpc/GrpcRequestContext.java
```diff
@@ -0,0 +1,93 @@
...
+
+
+/**
+ * We need to keep track of each request as it goes through the pipeline so that we can record the necessary metrics
+ * and separate different parts of the logic for the response. If a specific handler raises an error, we set
+ * the hasError flag to true and stop executing the rest of the pipeline excluding the stats collection.
+ */
+public class GrpcRequestContext<T> {
+  private final StreamObserver<T> responseObserver;
+  private final Class<?> responseType;
```
FelixGV, line 17: Can we use an enum instead? I think it would be cleaner when inspecting this property to be able to do a switch/case on the enum rather than a series of `if (responseType instanceof xyz)`. And if we do need the `Class` object, we can still get it out of a property of the enum.

services/venice-server/src/main/java/com/linkedin/venice/listener/StorageResponseHandlerCallback.java
```diff
@@ -0,0 +1,19 @@
+package com.linkedin.venice.listener;
+
+import com.linkedin.davinci.listener.response.ReadResponse;
+import com.linkedin.venice.response.VeniceReadResponseStatus;
+
+
+/**
+ * Callback interface to handle the response from the {@link StorageReadRequestHandler#processRequest} method.
+ */
+public interface StorageResponseHandlerCallback {
```
FelixGV, line 10: That's very good. I was hoping we could do something like this to reduce the copy pasted code, though the outline of the needed abstraction was not clear to me. Thanks for putting it together!

FelixGV (commented):
Thanks a lot Sushant! The gRPC code needs a lot of polishing and this PR looks like a very good change. I left a bunch of comments, though they are likely all minor. We also need to fix the static analysis and code coverage... Thanks again, I'm looking forward to seeing this merged in, great work!

services/venice-server/src/main/java/com/linkedin/venice/listener/request/GetRouterRequest.java
```diff
@@ -51,24 +50,34 @@ public int getKeyCount() {
...
+    return new GetRouterRequest(resourceName, partition, keyBytes, isRetryRequest, isStreamingRequest);
+  }
 
-    return new GetRouterRequest(resourceName, partition, keyBytes);
+  public static GetRouterRequest parseSingleGetGrpcRequest(SingleGetRequest singleGetRequest) {
+    String resourceName = singleGetRequest.getResourceName();
+    int partition = singleGetRequest.getPartition();
+    byte[] keyBytes = getKeyBytesFromUrlKeyString(singleGetRequest.getKey());
+    boolean isRetryRequest = singleGetRequest.getIsRetryRequest();
+    return new GetRouterRequest(resourceName, partition, keyBytes, isRetryRequest, false);
```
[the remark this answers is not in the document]
sushantmane: Yeah that's right. The old method was kept to support legacy gRPC single get service.  Added JavaDoc and `@Deprecated`  annotation.

services/venice-server/src/main/java/com/linkedin/venice/listener/request/GetRouterRequest.java
```diff
@@ -51,24 +50,34 @@ public int getKeyCount() {
...
+    if (requestParts.length != 5) {
       throw new VeniceException("Not a valid request for a STORAGE action: " + request.uri());
     }
+    // [0]""/[1]"action"/[2]"store"/[3]"partition"/[4]"key"
+    String topicName = requestParts[2];
+    int partition = Integer.parseInt(requestParts[3]);
+    byte[] keyBytes = getKeyBytesFromUrlKeyString(requestParts[4]);
+    boolean isRetryRequest = containRetryHeader(request);
+    boolean isStreamingRequest = StreamingUtils.isStreamingEnabled(request);
+    return new GetRouterRequest(topicName, partition, keyBytes, isRetryRequest, isStreamingRequest);
```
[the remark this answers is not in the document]
sushantmane: I was unaware of this implicit convention. I made the change to reduce indentation which IMO makes code easy to follow, though I understand this might be subjective. I will revert the change.

services/venice-server/src/main/java/com/linkedin/venice/listener/request/RouterRequest.java
```diff
@@ -60,11 +53,11 @@ public boolean isStreamingRequest() {
...
-    return request.headers().contains(HttpConstants.VENICE_RETRY);
-  }
-
   public boolean shouldRequestBeTerminatedEarly() {
     return requestTimeoutInNS != NO_REQUEST_TIMEOUT && System.nanoTime() > requestTimeoutInNS;
   }
+
+  public static boolean containRetryHeader(HttpRequest request) {
+    return request.headers().contains(HttpConstants.VENICE_RETRY);
+  }
```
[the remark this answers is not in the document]
sushantmane: Code move is little weird here.. I don't see reason for it😄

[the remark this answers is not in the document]
sushantmane: Moved it to `NettyUtils`

internal/venice-common/src/main/java/com/linkedin/venice/response/VeniceReadResponseStatus.java
sushantmane: > We want to avoid calling values() because it allocates a new array each time.

TIL. 

I like/prefer option 1: static property.

internal/venice-common/src/main/proto/VeniceReadService.proto
sushantmane: Yeah, I considered that, but decided to keep it similar to the HTTP request format since changing it would also require adding / modifying the multiGet API (POST) in `TransportClient`.

However, given the benefit of avoiding deserialization if we use a collection, I’m now leaning towards making the change. Do you agree we should do it in this PR?

services/venice-server/src/main/java/com/linkedin/venice/grpc/GrpcIoRequestProcessor.java
sushantmane: Yeah good point. Added `NoOpReadQuotaEnforcementHandler`

services/venice-server/src/main/java/com/linkedin/venice/listener/HttpStorageResponseHandlerCallback.java
```diff
@@ -0,0 +1,37 @@
...
+  private final ChannelHandlerContext context;
+
+  public HttpStorageResponseHandlerCallback(ChannelHandlerContext context) {
+    this.context = context;
+  }
+
+  // Factory method for creating an instance of this class.
+  public static HttpStorageResponseHandlerCallback create(ChannelHandlerContext context) {
+    return new HttpStorageResponseHandlerCallback(context);
+  }
```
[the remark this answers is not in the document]
sushantmane: No we don't. I forgot to change access modifier for the constructor. Thanks for catching it

services/venice-server/src/main/java/com/linkedin/venice/listener/StorageResponseHandlerCallback.java
sushantmane: Thanks! 🙏

services/venice-server/src/main/java/com/linkedin/venice/grpc/GrpcRequestContext.java
sushantmane: I agree, and originally it was written that way, but I switched to using responseType to avoid introducing yet another enum. However, that’s not a strong enough reason. So switching back to enum

sushantmane: Done

internal/venice-common/src/main/proto/VeniceReadService.proto
```diff
@@ -29,4 +54,126 @@ message VeniceServerResponse {
...
+
+message MultiGetRequest {
+  string resourceName = 1;
+  uint32 partition = 2;
+  bytes  keyBytes = 3;            // used for batch get
+  uint32 keyCount = 4;
+  bool isRetryRequest = 5;
+}
+
+message MultiGetResponse {
```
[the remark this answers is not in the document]
sushantmane: Right both MultiGet and Compute RPCs have this format. 


> If so, should we name it MultiKeyResponse instead?  

Yes

internal/venice-common/src/main/proto/VeniceReadService.proto
```diff
@@ -29,4 +54,126 @@ message VeniceServerResponse {
...
+  int32 statusCode = 1;
+  bytes value = 2;
```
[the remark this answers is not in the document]
sushantmane: > so that it can be written out incrementally, while all (or most?) 


I don't follow. Can you please elaborate?

services/venice-server/src/main/java/com/linkedin/venice/listener/HttpStorageResponseHandlerCallback.java
```diff
@@ -0,0 +1,37 @@
...
+  public void onReadResponse(ReadResponse readResponse) {
+    context.writeAndFlush(readResponse);
+  }
+
+  @Override
+  public void onError(VeniceReadResponseStatus readResponseStatus, String message) {
+    HttpShortcutResponse response = new HttpShortcutResponse(message, readResponseStatus.getHttpResponseStatus());
+    if (readResponseStatus == VeniceReadResponseStatus.MISROUTED_STORE_VERSION) {
+      response.setMisroutedStoreVersion(true);
+    }
```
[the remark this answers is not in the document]
sushantmane: Good catch, I missed that. This reminds me of a metrics/stats reporting gap between the two HTTP & gRPC implementations. I’m trying to think of how we can reduce the chances of mistakes by having them go through the same code path. I'll admit that at the moment I don't have a clean cut way to address all these issues but maybe I'll come up with something or if you happen to have some ideas I would love to hear them.

internal/venice-common/src/main/java/com/linkedin/venice/grpc/GrpcUtils.java
```diff
@@ -92,4 +94,19 @@ private static KeyStore loadStore(String path, char[] password, String type)
...
+   * @param body The ByteBuf to be converted to ByteString.
+   * @return The resulting ByteString.
+   */
+  public static ByteString toByteString(ByteBuf body) {
+    if (body.hasArray()) {
+      // Directly use the backing array to avoid copying
+      return ByteString.copyFrom(body.array(), body.arrayOffset() + body.readerIndex(), body.readableBytes());
+    }
+    // Fallback to nioBuffer() to handle the conversion efficiently
+    return ByteString.copyFrom(body.nioBuffer());
```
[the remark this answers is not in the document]
sushantmane: I see your point. 

>The more efficient approach might be to do response streaming, but I'm not sure what that entails.

That would be ideal and I do have plan to add true streaming API. But for now if you still think the cost of copying CompositeByteBuf is substantial then maybe we'll have to expand the scope to accomodate this change. 


> since you are refactoring the protocol buffer schemas anyway, might be to model the response as a collection rather than one big bytes field

Yeah that's also possible.

services/venice-server/src/main/java/com/linkedin/venice/listener/HttpStorageResponseHandlerCallback.java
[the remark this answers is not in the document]
sushantmane: I've one idea: maybe we should record stats in `StorageResponseHandlerCallback` callback. We can make `StorageResponseHandlerCallback` abstract class and the common metric reporting logic will reside there.

This solves the problem for IO requests (singleGet/multiGet/Compute) but not for other APIs/rpcs.  We don't report lot of metrics for them so some duplication might be ok?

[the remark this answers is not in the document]
sushantmane: Fixed this particular metric. But I'm still working on getting other metrics properly reported in gRPC path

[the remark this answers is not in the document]
FelixGV: I don't have a clear idea off the top of my head. This PR will already make the two paths much closer. It's ok if some discrepancies remain. We'll keep improving it later.

services/venice-server/src/main/java/com/linkedin/venice/grpc/GrpcRequestContext.java
```diff
@@ -0,0 +1,93 @@
...
+  private final ServerStatsContext serverStatsContext;
+
+  private RouterRequest routerRequest;
+  private ReadResponse readResponse = null;
+  private VeniceReadResponseStatus readResponseStatus;
+  private String errorMessage;
```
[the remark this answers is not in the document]
sushantmane: Currently `ReadResponse` is only used for IO requests. For all requests and for early errored requests in Netty we use `HttpShortcutResponse`.  

Ideally it would be good to have common request context for all requests but I think I'll skip it for now

services/venice-server/src/main/java/com/linkedin/venice/listener/OutboundHttpWrapperHandler.java
```diff
@@ -55,18 +55,16 @@ public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise)
     try {
       if (msg instanceof AbstractReadResponse) {
         AbstractReadResponse obj = (AbstractReadResponse) msg;
-        ServerStatsContext statsContext = statsHandler.getServerStatsContext();
-        setStats(statsContext, obj);
-
+        requestStatsRecorder.setReadResponseStats(obj.getStatsRecorder());
```
FelixGV, line 59: While I'm not opposed to renaming the `ServerStatsContext`, I think we may need to take a holistic look at the various "stats recorders" we have in play. There is a "read response stats recorder" since my last refactoring, and now we would have a "request stats recorder" as well. Maybe this is fine, or maybe it needs some further tweaking, I am not sure. But this line in particular illustrates the confusion (it seems that at the very least, the `AbstractReadResponse::getStatsRecorder` function ought to be renamed more precisely?)

sushantmane: IMHO we should merge `RequestStatsRecorder` (formerly `ServerStatsContext`)  and all of `ReadResponseStatsRecorder` into a common class to keep things simple. 

However, if we think it's not desriable I can change `AbstractReadResponse::getStatsRecorder` to `AbstractReadResponse::getReadResponseStatsRecorder` or something?

services/venice-server/src/main/java/com/linkedin/venice/listener/HttpStorageResponseHandlerCallback.java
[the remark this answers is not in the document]
sushantmane: Now we have common metrics reporting function and it seems much cleaner now IMHO.

services/venice-server/src/main/java/com/linkedin/venice/grpc/GrpcIoRequestProcessor.java
```diff
@@ -0,0 +1,284 @@
...
+        break;
+      default:
+        requestContext.setReadResponseStatus(VeniceReadResponseStatus.INTERNAL_SERVER_ERROR);
+        requestContext.setErrorMessage("Unknown quota enforcement result: " + result);
+    }
+    sendResponse(requestContext);
+  }
+
+  /**
+   * Ensures that all fields in the request context are properly set before invoking this method.
```
sushantmane, line 118: fix type: `Ensures` to `Ensure`

services/venice-server/src/main/java/com/linkedin/venice/listener/OutboundHttpWrapperHandler.java
sushantmane: I do have one question: why did we introduce `ReadResponseStatsRecorder`?

FelixGV: So... my motivation during that refactoring was that I did not like the previous structure of the code where a bunch of stats were getting recorded in the `ReadResponse`, and then had to be copied one by one into the `ServerStatsContext`, and then had to be taken from that context and passed into the various metric recording calls... I thought it was kind of tedious to add one more metric, having to wire through all these steps.

In the new code structure, stats are accumulated in a `ReadResponseStats`, which only carries setters and incrementer methods. And then a reference to that one object is passed into the context one time (not each field separately). Finally, when it's recording time, the `ReadResponseStatsRecorder` comes into play and accepts the metrics recorder object and acts on it.

I think it could be fine to pull more of the `ServerStatsContext` fields into the `ReadResponseStats`... or maybe there are more/bigger structural improvements we could do. IDK.

sushantmane: I've rename `AbstractReadResponse::getStatsRecorder` to `AbstractReadResponse::getReadResponseStatsRecorder` as suggested in the first comment. 

As for larger/broader refactoring of stats record, I would like to revisit it in the future

internal/venice-common/src/main/proto/VeniceReadService.proto
sushantmane: I started working on this, but the changes required are quite extensive due to how tightly our code is coupled with the current abstractions. You can check out the partial changes here: [commit link](https://github.com/sushantmane/venice/commit/da641dcd248d0f1908fb0736d605cf8da2763a23).

Addressing this and the remaining three comments will require significant refactoring, which I’d prefer to handle in a separate PR, as we discussed offline.

internal/venice-common/src/main/java/com/linkedin/venice/grpc/GrpcUtils.java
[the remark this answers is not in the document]
FelixGV: Dropping this comment for now, since the solution would require too much changes and this PR is already big.

internal/venice-common/src/main/java/com/linkedin/venice/meta/ServerAdminAction.java
```diff
@@ -1,8 +1,19 @@
...
 public enum ServerAdminAction {
   DUMP_INGESTION_STATE(0), DUMP_SERVER_CONFIGS(1);
 
+  private static final Map<Integer, ServerAdminAction> ADMIN_ACTION_MAP = new HashMap<>(2);
+
+  static {
+    for (ServerAdminAction action: values()) {
+      ADMIN_ACTION_MAP.put(action.getValue(), action);
+    }
+  }
```
FelixGV, line 16: FYI, for the common case of enum codes that range from 0..N monotonically, we have a cleaner solution than using a map... See `EnumUtils::getEnumValuesArray` and `VeniceEnumValue`. It is very easy to use, more performant, and will require a bit fewer lines of code than you have here...

internal/venice-common/src/main/proto/VeniceReadService.proto
```diff
@@ -1,11 +1,36 @@
...
+    option deprecated = true;
+  }
+
+  rpc batchGet(VeniceClientRequest) returns (VeniceServerResponse) {
+    option deprecated = true;
+  }
+
+  rpc singleGet(SingleGetRequest) returns (SingleGetResponse) {}
+
+  rpc multiGet(MultiGetRequest) returns (MultiKeyResponse) {}
```
FelixGV, line 19: Sorry for the confusion here... I know I am the one that asked for marking one of the APIs as deprecated, but I did not have the full understanding at the time I said that...

Since we are not yet implementing the new protocol buffers on the client-side, then effectively these old `get/batchGet` protocols are still the only valid one... the new ones are still experimental, they're not implemented end-to-end, and they might still change before becoming the new protocols we use (e.g. in case we decide to model the multi-key content as separate units).

So given all of the above, I think it's incorrect to mark the old protocols as deprecated, both here and in the Java code which use them.

Sorry again for the flip flopping...

internal/venice-common/src/main/proto/VeniceReadService.proto
FelixGV: SG, let's leave it out of scope of the current PR.

internal/venice-common/src/main/proto/VeniceReadService.proto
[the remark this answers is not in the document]
FelixGV: The idea was that the header-type content should be a prefix, and the repetitive content can be flushed to the wire incrementally. And therefore, the ordering might be better if the values are at the end... but since these new protocols are still a work in progress, it's fine to ignore for now...

FelixGV (commented):
Doing another pass on this PR... thanks for iterating on it. Looks good so far, but I'm not done going through it... I left some minor comments, and I will resume my second pass later...

sushantmane:
I'm still working on client side changes to support streaming within a single route. I've figured out structure for it and working  changes

FelixGV:
> I'm still working on client side changes to support streaming within a single route. I've figured out structure for it and working changes

Nice, looking forward to it. When you're ready with your next iteration, please rebase, push it up and LMK. I'll take another look.

gaojieliu:
Hi @sushantmane 
Do you still have plan to merge this change?
It seems to be a good change, but it is too large and do you have plan to split it into multiple ones?

sushantmane:
I'll get back to this after PubSub migration.

sushantmane:
I'll get back to this after PubSub migration.

sushantmane:
I'll get back to this after PubSub migration.

PR #1164, opened 2024-09-10
[server] Parallel processing for AA/WC workload

gaojieliu:
This code change introduces a new thread pool to process AA/WC workload, and essentially, all the AA/WC processing logic will be delegated to this new thread pool regardless of the consumer threads/pool assignment. The reason to have such feature:
1. Consumption is not the bottleneck and AA/WC processing logic is cpu intensive, so even with one single consumer, the thread pool can be largely utilized when there are a lot of AA/WC messages from this single consumer.
2. In prod, we have seen a few scenarios, which have long-tail partitions, which would slow down inc push or rebalance significantly.

This code change introduces two new configs:
server.aa.wc.workload.parallel.processing.enabled: default false server.aa.wc.workload.parallel.processing.thread.pool.size: default 8

Here is how the workflow is changed to adopt the new parallel processing pattern:
1. Consumer threads will send a list of messages belonging to the same partition to `ConsumptionTask`, which is an existing behavior.
2. When `StoreIngestionTask` (extending `ConsumptionTask`) receives a batch of messages, it would check whether all the messages are AA/WC messages and if yes, it will delegate them batch by batch to the new processing thread pool when this feature is enabled. For each batch, here are the steps: a. Lock the keys in the batch. b. Execute the AA/WC logic, and for the updates to the same key, they will be executed sequentially. c. Update the transient cache. d. Return a new compatible format: `PubSubMessageProcessedResultWrapper`, which includes both the original message and the processed result.
3. Continue the existing logic to function: `delegateConsumerRecord`.
4. When processing each message, if the message carries the processed result, use it, otherwise, continue the processing synchronously.

There is another change, which will do DIV validation/filtering before step 2, which is used to filter out duplicate messages as early as possible to reduce the unnecessary compute.

This PR also fixes a race condition in changelog view-writer and the fix is simple, and Venice Server will update the transient cache before writing to the changelog view topic, so that when next message tries to lookup the same key, it will get the latest result even the writing to the version topic is still pending.

## How was this PR tested?
CI


clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -1040,84 +1124,156 @@ protected void produceToStoreBufferServiceOrKafka(
...
+
+    if (metricsEnabled) {
+      if (totalBytesRead > 0) {
+        hostLevelIngestionStats.recordTotalBytesReadFromKafkaAsUncompressedSize(totalBytesRead);
+      }
+      if (elapsedTimeForPuttingIntoQueue.getValue() > 0) {
+        hostLevelIngestionStats.recordConsumerRecordsQueuePutLatency(
+            elapsedTimeForPuttingIntoQueue.getValue(),
+            beforeProcessingBatchRecordsTimestampMs);
       }
```
FelixGV, line 1194: Why carry this all the way here? Can we do the metrics recording directly inside `handleSingleMessage`, rather than merely setting the value into this `ValueHolder` over there?

FelixGV (commented):
I just begun taking a look at this PR and I'm not done yet... but I'm stopping for today so I'll just leave this one comment for now.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
gaojieliu: This is trying to preserve the existing semantics to report the latency per batch..
I don't mind to record metrics directly in the new function, but it would cause a change in the monitoring.

FelixGV: Ah, so this metric is per batch or per record?

gaojieliu: It is per batch today:
[elapsedTimeForPuttingIntoQueue](https://github.com/linkedin/venice/blob/main/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java#L1106)

FelixGV: Ok I see... this seems like a confusing metric semantic though xD

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -1040,84 +1124,156 @@ protected void produceToStoreBufferServiceOrKafka(
...
+    }
+
     long totalBytesRead = 0;
-    double elapsedTimeForPuttingIntoQueue = 0;
+    ValueHolder<Double> elapsedTimeForPuttingIntoQueue = new ValueHolder<>(0d);
     boolean metricsEnabled = emitMetrics.get();
     long beforeProcessingBatchRecordsTimestampMs = System.currentTimeMillis();
+
     for (PubSubMessage<KafkaKey, KafkaMessageEnvelope, Long> record: records) {
+      partitionConsumptionState = partitionConsumptionStateMap.get(topicPartition.getPartitionNumber());
```
sixpluszero, line 1158: maybe we can move this out of the loop?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -1040,84 +1124,156 @@ protected void produceToStoreBufferServiceOrKafka(
...
+    }
+    /**
+     * Process records batch by batch.
+     */
+    for (List<PubSubMessage<KafkaKey, KafkaMessageEnvelope, Long>> batch: batches) {
+      List<ReentrantLock> locks = ingestionBatchProcessor.lockKeys(batch);
+      try {
+        long beforeProcessingPerRecordTimestampNs = System.nanoTime();
+        List<PubSubMessageProcessedResultWrapper<KafkaKey, KafkaMessageEnvelope, Long>> processedResults =
+            ingestionBatchProcessor.process(
```
sixpluszero, line 1254: Please bear me with some dumb question - even though we synced offline but I am still a bit confused about the processing logic here: if there is control message such as SOS/EOS (which carries something like HB) how does batch processes process these messages? I checked the processFunction passed in, looks like it only assumes data message?

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestActiveActiveIngestionWithReadWriteLeaderCompactionTuningAndParallelProcessing.java
```diff
@@ -1,11 +1,18 @@
...
+ * Integration test to verify active/active replication when applying read-write leader compaction optimization
+ * and parallel processing of AA/WC workload.
  */
-public class TestActiveActiveIngestionWithReadWriteLeaderCompactionTuning extends TestActiveActiveIngestion {
+public class TestActiveActiveIngestionWithReadWriteLeaderCompactionTuningAndParallelProcessing
+    extends TestActiveActiveIngestion {
   @Override
   protected boolean isLevel0CompactionTuningForReadWriteLeaderEnabled() {
     return true;
   }
```
sixpluszero, line 12: This change is pretty important, is it possible that we add a new test that sends a bunch of updates with same keys? I am not sure if this scenario is covered in existing tests.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -501,13 +506,16 @@ public StoreIngestionTask(
     if (!this.recordLevelMetricEnabled.get()) {
       LOGGER.info("Disabled record-level metric when ingesting current version: {}", kafkaVersionTopic);
     }
+    this.parallelProcessingThreadPool = builder.getAaWCWorkLoadProcessingThreadPool();
```
sixpluszero, line 509: nit: Should we use AAWC instead of AaWC as it is the pattern everywhere else.

sixpluszero (commented):
Thanks for the big change! I did one pass, since this change in pipeline is big there are some confusion still so I left some comments below. Please feel free to take a look.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -293,13 +297,35 @@ public LeaderFollowerStoreIngestionTask(
               store,
               version.getNumber(),
               schemaRepository.getKeySchema(store.getName()).getSchema());
+      boolean tmpValueForHasChangeCaptureViewWriter = false;
```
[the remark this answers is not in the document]
gaojieliu: Yeah, I didn't change the location where the `VeniceViewWriter` is being initialized today, and I guess this doesn't matter.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -2121,6 +2152,59 @@ protected void recordHeartbeatReceived(
...
   }
 
+  @Override
+  protected Iterable<PubSubMessage<KafkaKey, KafkaMessageEnvelope, Long>> validateAndFilterOutDuplicateMessagesFromLeaderTopic(
+      Iterable<PubSubMessage<KafkaKey, KafkaMessageEnvelope, Long>> records,
+      PubSubTopicPartition topicPartition) {
+    PartitionConsumptionState partitionConsumptionState =
+        partitionConsumptionStateMap.get(topicPartition.getPartitionNumber());
+    if (partitionConsumptionState == null) {
+      // The partition is likely unsubscribed, will skip these messages.
```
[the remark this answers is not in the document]
gaojieliu: Sure.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/IngestionBatchProcessor.java
```diff
@@ -0,0 +1,165 @@
...
+                  partition,
+                  kafkaUrl,
+                  kafkaClusterId,
+                  beforeProcessingRecordTimestampNs,
+                  beforeProcessingBatchRecordsTimestampMs));
+        });
+      }, batchProcessingThreadPool));
+    });
+    try {
+      CompletableFuture.allOf(futureList.toArray(new CompletableFuture[0])).get();
```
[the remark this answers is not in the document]
gaojieliu: I think it should be good enough to check the overall ingestion process, which is the ultimate metric we should look at.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestActiveActiveIngestionWithReadWriteLeaderCompactionTuningAndParallelProcessing.java
gaojieliu: `TestChangelogConsumer#testAAIngestionWithStoreView` has this pattern, and that is the place I discovered the race condition with changelog view.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
gaojieliu: Sure

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
gaojieliu: Batch processing function will skip all the Control messages.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/config/VeniceServerConfig.java
```diff
@@ -895,6 +899,10 @@ public VeniceServerConfig(VeniceProperties serverProperties, Map<String, Map<Str
     if (channelOptionWriteBufferHighBytes <= 0) {
       throw new VeniceException("Invalid channel option write buffer high bytes: " + channelOptionWriteBufferHighBytes);
     }
+    aaWCWorkloadParallelProcessingEnabled =
+        serverProperties.getBoolean(SERVER_AA_WC_WORKLOAD_PARALLEL_PROCESSING_ENABLED, false);
+    aaWCWorkloadParallelProcessingThreadPoolSize =
+        serverProperties.getInt(SERVER_AA_WC_WORKLOAD_PARALLEL_PROCESSING_THREAD_POOL_SIZE, 8);
```
sixpluszero, line 905: Actually I have a question - I think correctness there is no issue, but if all the AAWC leader can submit batch of record pulled from RT, then would this be true for current / non-current version to share the same critical path and one maybe impact the other?

gaojieliu: Yes, it is correct.
For the current/non-current version differentiation, we will delegate it to the consumer pool strategy, rather than solving it here.

There are different layer of strategies, and this layer is purely for scheduling, not prioritization.
That is the way how it is designed originally: separate of concerns.

sixpluszero: Synced offline, I think for now it is not a concern as described above.

.gitignore
```diff
@@ -17,6 +17,7 @@ build/
 internal/venice-test-common/tmp/
 internal/venice-test-common/dump.complete.hprof
 internal/venice-test-common/src/jmh/generated
+internal/venice-test-common/file:/
```
sixpluszero, line 20: minor question - why do we need to add this?

gaojieliu: When running the E2E integration tests locally, I saw this folder got created every time and that is the reason I would like to exclude from here.

PR #1165, opened 2024-09-10
[common] Retry controller client request on exceptions

majisourav99:
## Retry controller client request on exceptions

Today, "ControllerClient#retryableRequest" retries when the controller returns an error response. It doesn't retry when an exception is thrown. We need to handle exceptions as an error condition and retry till the retryAttempt count is reached.

Also it misses to retry VeniceException from `getLeaderControllerUrl`  which could lead to data recovery pushes to fail on a single unsuccessful controller discovery call. 

## How was this PR tested?

GHCI


nisargthakkar (approved):
LGTM

PR #1168, opened 2024-09-11
[dvc] Exposed an API to pass an executor to CachingDaVinciClientFactory

gaojieliu:
This PR makes the read chunk executor pluggable.

## How was this PR tested?
CI


xunyin8 (approved):
Thank you Gaojie!

xunyin8 (approved):

PR #1170, opened 2024-09-11
[common][server][dvc-client] Make a true non-blocking end-offset fetcher

majisourav99:
## Create a true non-blocking version of end-offset fetcher

In recent past we have seen cases where end offset fetch call stalls for a very long time. It could happen due to multiple reasons, like general Kafka outage, or mismatched VT vs RT partition count leading to non-existed partition or other network issues. 
Even though we cache the offset value, there is a blocking call before cache is updated. This hold the lock for a very long time and subsequent calls (either from metics collection) or the ready to serve call waits forever.  Since drainer thread is shared this blocks the processing of other resources in a drainer thread leading to cluster-wide impact. This PR check for cache miss and returns immediately with some sentinel value while a nonblocking call updates the cache asynchronously. 

## How was this PR tested?

GHCI

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
@@ -2062,13 +2062,18 @@ public abstract class StoreIngestionTask implements Runnable, Closeable {
     if (offsetFromConsumer >= 0) {
       return offsetFromConsumer;
     }
-    return RetryUtils.executeWithMaxAttempt(() -> {
-      long offset = getTopicManager(kafkaUrl).getLatestOffsetCachedNonBlocking(pubSubTopic, partition);
-      if (offset == -1) {
-        throw new VeniceException("Found latest offset -1");
-      }
-      return offset;
-    }, 5, Duration.ofSeconds(1), Collections.singletonList(VeniceException.class));
+    try {
+      return RetryUtils.executeWithMaxAttempt(() -> {
+        long offset = getTopicManager(kafkaUrl).getLatestOffsetCachedNonBlocking(pubSubTopic, partition);
+        if (offset == -1) {
+          throw new VeniceException("Found latest offset -1");
+        }
+        return offset;
+      }, 5, Duration.ofSeconds(1), Collections.singletonList(VeniceException.class));
+    } catch (Exception e) {
+      LOGGER.error("Could not find latest offset for {} even after 5 retries", pubSubTopic.getName());
+      return -1;
+    }
   }
 
   protected long getPartitionOffsetLagBasedOnMetrics(String kafkaSourceAddress, PubSubTopic topic, int partition) {
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
@@ -2061,7 +2061,7 @@ public abstract class StoreIngestionTask implements Runnable, Closeable {
     if (offsetFromConsumer >= 0) {
       return offsetFromConsumer;
     }
-    return getTopicManager(kafkaUrl).getLatestOffsetCached(pubSubTopic, partition);
+    return getTopicManager(kafkaUrl).getLatestOffsetCachedNonBlocking(pubSubTopic, partition);
   }
 
   protected long getPartitionOffsetLagBasedOnMetrics(String kafkaSourceAddress, PubSubTopic topic, int partition) {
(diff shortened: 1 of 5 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -2061,7 +2062,18 @@ protected long getTopicPartitionEndOffSet(String kafkaUrl, PubSubTopic pubSubTop
...
+      return RetryUtils.executeWithMaxAttempt(() -> {
+        long offset = getTopicManager(kafkaUrl).getLatestOffsetCachedNonBlocking(pubSubTopic, partition);
+        if (offset == -1) {
+          throw new VeniceException("Found latest offset -1");
+        }
+        return offset;
+      }, 5, Duration.ofSeconds(1), Collections.singletonList(VeniceException.class));
+    } catch (Exception e) {
+      LOGGER.error("Could not find latest offset for {} even after 5 retries", pubSubTopic.getName());
+      return -1;
```
sixpluszero, line 2075: I am wondering if return -1 is expected. If the VT exists, RT does not exist, we return -1 and if the latest processed offset is initialized to 0 (is it true?) then the lag is 1, and it will not block the partition to go online?

majisourav99: In RTS code caller `measureRTOffsetLagForSingleRegion`, it does this

`if (lastOffsetInRealTimeTopic < 0) {
      if (!REDUNDANT_LOGGING_FILTER.isRedundantException("Got a negative lastOffsetInRealTimeTopic")) {
        LOGGER.warn(
            "Unexpected! Got a negative lastOffsetInRealTimeTopic ({})! Will return Long.MAX_VALUE ({}) as the lag for replica: {}.",
            lastOffsetInRealTimeTopic,
            Long.MAX_VALUE,
            pcs.getReplicaId(),
            new VeniceException("Exception not thrown, just for logging purposes."));
      }
      return Long.MAX_VALUE;
    }

FelixGV:
> mismatched VT vs RT partition count leading to non-existed partition

Does this actually happen?

majisourav99:
> > mismatched VT vs RT partition count leading to non-existed partition
> 
> Does this actually happen?

Yes, it did happen! Hybrid conversion during target push created a scenario different colo RT to be created with different partition count.

FelixGV:
> > > mismatched VT vs RT partition count leading to non-existed partition
> > 
> > 
> > Does this actually happen?
> 
> Yes, it did happen! Hybrid conversion during target push created a scenario different colo RT to be created with different partition count.

Wow, that is super messed up... is there a way we can guard against this scenario upstream?

majisourav99:
> > > > mismatched VT vs RT partition count leading to non-existed partition
> > > 
> > > 
> > > Does this actually happen?
> > 
> > 
> > Yes, it did happen! Hybrid conversion during target push created a scenario different colo RT to be created with different partition count.
> 
> Wow, that is super messed up... is there a way we can guard against this scenario upstream?

We can let SIT constructor fail if VT != RT partition count for hybrid stores

sixpluszero:
> We can let SIT constructor fail if VT != RT partition count for hybrid stores

Do you plan to include this in the PR, or you want to have a separate PR to deal with the changes?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
sixpluszero: Synced offline - I think it is ok for now.

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicMetadataFetcher.java
```diff
@@ -379,6 +379,25 @@ CompletableFuture<Long> getLatestOffsetWithRetriesAsync(PubSubTopicPartition pub
...
+    updateCacheAsync(
+        pubSubTopicPartition,
+        cachedValue,
+        latestOffsetCache,
+        () -> getLatestOffsetWithRetriesAsync(
+            pubSubTopicPartition,
+            DEFAULT_MAX_RETRIES_FOR_POPULATING_TMD_CACHE_ENTRY));
+    if (cachedValue == null) {
+      cachedValue = latestOffsetCache.get(pubSubTopicPartition);
+      if (cachedValue == null) {
```
sixpluszero, line 394: btw I have a last question - why would you do this 2 times? when it is null, meaning it is not going to get anything from the latestOffsetCache. Or do you think updateCacheAsync has the chance to update it in the instant?
IMO we should not really expect that to happen...

sixpluszero (approved):
I left a final question but overall it is good. Thank you for the fix!

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicMetadataFetcher.java
majisourav99: Its the later one, `Or do you think updateCacheAsync has the chance to update it in the instant` Its best effort, not super critical to have this.

PR #1174, opened 2024-09-12
[doc][script] Automatically verify JDK version in gradle, and specify in docs users need to set their JAVA_HOME

kvargha:

1. When users are on the wrong version, the error message when running gradle tasks can be cryptic. So verifying their JDK version before running other tasks will make it easier for users to debug.
2. The docs mention to download JDK 17, but it doesn't explicitly mention to add it to their JAVA_HOME.

## How was this PR tested?

Tested locally

Users will receive an explicit error message when using the incorrect JDK version.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -799,12 +799,11 @@ allprojects {
   task printAllDependencies(type: DependencyReportTask) {}
 }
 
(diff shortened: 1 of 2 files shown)


FelixGV (approved):
Nice! Thanks a lot Koorous!

PR #1175, opened 2024-09-13
[server][router] Handle migrated store before checking ACL

nisargthakkar:
## [server][router] Handle migrated store before checking ACL

When a store is completely migrated, but there is still some client that didn't have queries at time of migration, and hence didn't migrate to the new cluster. We fixed this issue in #843, but removed the logic to check the store existence from the ACL handler with the expectation that the next layers would redirect the requests for migrated store.

The interface of the ACL handler doesn't specify how to handle the case of a missing store. So, it is possible that the ACL handler could either:
1. Return `false` to the `hasAccess` call
2. Throw an `AclException`

Only in the second case would the request be forwarded to the next handler. This commit adds the store existence check before checking ACLs and hence, can forward the request to the next handler so that the client can be notified of the store move.

This commit unifies common store ACL handler logic into `AbstractStoreAclHandler` and splits router-specific and server-specific logic into their independent implementations: `RouterStoreAclHandler` and `ServerStoreAclHandler`

## How was this PR tested?

Added unit tests. GH CI for regression tests


internal/venice-common/src/main/java/com/linkedin/venice/acl/handler/AbstractStoreAclHandler.java
```diff
@@ -95,115 +79,55 @@ public void channelRead0(ChannelHandlerContext ctx, HttpRequest req) throws SSLP
...
-    return resourceName;
+  protected boolean isAccessAlreadyApproved(ChannelHandlerContext ctx) {
+    return false;
   }
 
-  @VisibleForTesting
-  boolean isInvalidRequest(String[] requestParts) {
-    int partsLength = requestParts.length;
-    boolean invalidRequest = false;
+  protected abstract boolean needsAclValidation(String[] requestParts);
```
xunyin8, line 116: Should we just make this method take in a `RouterResourceType`? Either way I think it'd be great if we can have a test where we pass in or construct requestParts for all enums in `RouterResourceType` to ensure we are handling things correctly and didn't miss anything.

xunyin8 (commented):
LGTM just one minor comment.

internal/venice-common/src/main/java/com/linkedin/venice/acl/handler/AbstractStoreAclHandler.java
nisargthakkar: This piece of code is shared b/w Router and Server. So, this function cannot use `RouterResourceType`. But let me see if I can do something to reduce the string parsing.

Also, the tests for this class were somewhat tedious to automate. I tried improving, but that would end up overwriting all tests for this class. So, I avoided it. I can give it another shot

internal/venice-common/src/main/java/com/linkedin/venice/acl/handler/AbstractStoreAclHandler.java
sushantmane, line 1: Instead of having `abstract` class, can we make it concrete class and pass it as dependency to Netty and gRPC handlers?

nisargthakkar: How would that work? Can you share some details/examples? From what I've been able to come up with, the concrete class would still have to handle some generics, and also need some abstractions to handle Netty and gRPC objects.

I wanted to remove duplicating logic, but that would end up bloating up the PR, and also distracts from the issue at hand.

nisargthakkar: The alternative is to use `Consumer`, `Supplier`, etc interfaces to act as the abstractions

sushantmane: >How would that work? Can you share some details/examples? From what I've been able to come up with, the concrete class would still have to handle some generics, and also need some abstractions to handle Netty and gRPC objects.

If we restructure it so that the concrete class StoreAclService only returns AccessResult, and then allow the respective Netty (server or router) or gRPC handlers to interpret the result and format the response accordingly, that approach would work well.

We are following this format in `ReadQuotaEnforcementHandler`

```java
public class StoreAclService {
  private final IdentityParser identityParser;
  private final ReadOnlyStoreRepository metadataRepository;
  private final DynamicAccessController accessController;

// this could be modified to has message in it
  public enum AccessResult {
    GRANTED,
    FORBIDDEN,
    UNAUTHORIZED
  }

  public StoreAclService(
      IdentityParser identityParser,
      DynamicAccessController accessController,
      ReadOnlyStoreRepository metadataRepository) {
    this.identityParser = identityParser;
    this.metadataRepository = metadataRepository;
    this.accessController = accessController
        .init(metadataRepository.getAllStores().stream().map(Store::getName).collect(Collectors.toList()));
    this.metadataRepository.registerStoreDataChangedListener(new AclCreationDeletionListener(accessController));
  }

  public AccessResult checkAccess(
      String uri,
      X509Certificate clientCert,
      String storeName,
      String method) {

    // uri and client could be used in logging else they are not needed
    String client = identityParser.parseIdentityFromCert(clientCert);

    try {
      if (accessController.hasAccess(clientCert, storeName, method)) {
        return AccessResult.GRANTED;
      }

      return handleAccessError(storeName);
    } catch (AclException e) {
      return handleException(storeName, e);
    }
  }

  private AccessResult handleAccessError(String storeName) throws AclException {
    if (!accessController.isFailOpen() && !accessController.hasAcl(storeName)) {
      return AccessResult.UNAUTHORIZED;
    } else {
      return AccessResult.FORBIDDEN;
    }
  }

  private AccessResult handleException(String storeName, AclException e) {
    if (accessController.isFailOpen()) {
      return AccessResult.GRANTED;
    } else {
      return AccessResult.FORBIDDEN;
    }
  }
}


public class ServerStoreNettyAclHandler extends SimpleChannelInboundHandler<HttpRequest> {
  private final StoreAclService storeAclService;

  public NettyServerStoreAclHandler(StoreAclService storeAclService) {
    this.storeAclService = storeAclService;
  }

  @Override
  public void channelRead0(ChannelHandlerContext ctx, HttpRequest req) throws SSLPeerUnverifiedException {
    X509Certificate clientCert = extractClientCert(ctx);
    String uri = req.uri();
    String storeName = extractStoreNameFromUri(uri);
    String method = req.method().name();

    // skip ACL check for certain URIs
    if (isAccessAlreadyApproved(ctx) || !needsAclValidation(uri)) {
      ReferenceCountUtil.retain(req);
      ctx.fireChannelRead(req);
      return;
    }
    
    StoreAclService.AccessResult result = storeAclService.checkAccess(uri, clientCert, storeName, method);

    if (result == StoreAclService.AccessResult.GRANTED) {
      ctx.fireChannelRead(req); // Continue if access is granted
    } else if (result == StoreAclService.AccessResult.UNAUTHORIZED) {
      NettyUtils.setupResponseAndFlush(HttpResponseStatus.UNAUTHORIZED, "ACL not found!".getBytes(), false, ctx);
    } else if (result == StoreAclService.AccessResult.FORBIDDEN) {
      NettyUtils.setupResponseAndFlush(HttpResponseStatus.FORBIDDEN, "Access denied!".getBytes(), false, ctx);
    }
    ctx.close(); // Close connection on access denial
  }
}
```

sushantmane: If you notice `StoreAclService` doesn't have any netty (rest/http) or grpc (protobuf) related code.

sushantmane: `RouterStoreNettyAclHandler` will look similar to `ServerStoreNettyAclHandler`

sushantmane: >I wanted to remove duplicating logic, but that would end up bloating up the PR, and also distracts from the issue at hand.


I understand, and I’ll leave it to your discretion. However, I do believe that the approach mentioned above would help make the code cleaner IMHO

sushantmane: > The alternative is to use Consumer, Supplier, etc interfaces to act as the abstractions 


IMHO, except in a few specific cases, using consumers, suppliers, and functions tends to significantly reduce code readability and places a heavier burden on the reader.

nisargthakkar: The suggestion about `checkAccess` is good, and I can add it regardless. I see this as an improvement over the `hasAccess` method. Let me make this change and see how things look

But my primary concern is that there is still duplication about how to handle other common things like:
* Checking for `isAccessAlreadyApproved(ctx) || !needsAclValidation(uri)`
* validating if the request type is valid or not
* Store doesn't exist

This is why I was thinking about it in terms of abstraction instead of a helper function.


> IMHO, except in a few specific cases, using consumers, suppliers, and functions tends to significantly reduce code readability and places a heavier burden on the reader.

Totally agree about this.

sushantmane: I think tiny code duplication (it's mostly in handler channelRead/intercept methods I suppose) should be ok if it means better readability.

`isAccessAlreadyApproved(ctx)` will be same for `NettyServer` and `Router` code path but will look different for gRPC. So having helper should be alright for this common path: `NettyServer` and `Router`. 

Similarly, `NettyServer` and `GrpcServer` paths will call the same helper `needsAclValidation` but router will have different version.

nisargthakkar: I tried to incorporate your suggestion, but I've still not removed the abstract class. I still thing there is value in having an abstract class for server and router as we can handle some common logic in one place. Let me know if you strongly feel that this shouldn't be done this way

I've not unified the gRPC and Netty paths much as that would need more abstractions that I don't think should be done in this PR

sushantmane: > I still thing there is value in having an abstract class for server and router as we can handle some common logic in one place.

That should be fine IMO. When we want to unify gRPC related code we can make more cleanups

services/venice-router/src/main/java/com/linkedin/venice/router/acl/RouterStoreAclHandler.java
```diff
@@ -0,0 +1,83 @@
...
+      case TYPE_ADMIN: // Access control for Admin operations are handled in AdminOperationsHandler
+      case TYPE_CURRENT_VERSION:
+      case TYPE_RESOURCE_STATE:
+      case TYPE_BLOB_DISCOVERY:
+      case TYPE_REQUEST_TOPIC:
+        return false;
+      case TYPE_STORAGE:
+      case TYPE_COMPUTE:
+        return true;
+      case TYPE_INVALID:
```
gaojieliu, line 52: There is a big list of the switch cases, so if every valid request needs to go through the list, it will be slightly inefficient, right?
If we still to this pattern, I think we should move `TYPE_STORAGE` and `TYPE_COMPUTE` up, which are more latency sensitive with more QPS, and another way is to use a `HashMap`.
WDYT?

gaojieliu (commented):
Looks good overall.
Left one minor comment.

services/venice-router/src/main/java/com/linkedin/venice/router/acl/RouterStoreAclHandler.java
nisargthakkar: I can move it around, but some resources I read online say that the compiler will make this into a lookup table on it's own if there are >5 branches

nisargthakkar: https://www.geeksforgeeks.org/switch-vs-else/

nisargthakkar: But I could get rid of the rest, and only keep `TYPE_STORAGE` and `TYPE_COMPUTE`, and let the tests handle checking the case that all types are handled already. We have this test now

nisargthakkar: This is how it would look:
```
  @Override
  protected boolean needsAclValidation(RouterResourceType resourceType) {
    if (resourceType == TYPE_STORAGE || resourceType == TYPE_COMPUTE) {
      return true;
    }

    if (resourceType == TYPE_INVALID) {
      throw new VeniceUnsupportedOperationException(resourceType.name());
    }

    return false;
  }
```

gaojieliu: Got it.
Only keeping `TYPE_STORAGE` and `TYPE_COMPUTE` sounds good to me.
This would also reduce the maintenance burden when we adding more admin/metadata endpoints in the future.

gaojieliu (approved):

PR #1176, opened 2024-09-14
[controller] Emit push job status metrics from controller

m-nagarajan:
## Summary
`VPJ` communicates with the controller to write `PushJobDetails` to the `PUSH_JOB_DETAILS_STORE_NAME` system store. This PR introduces new metrics emitted by the parent controller for push job success/failure.

Choice of Stat Type: The original idea is for each event (push job success/failure) to emit 1, which can be aggregated in the metrics receiver in any manners to calculate push job availability.
- Using `OccurrenceRate(SECOND)` gives QPS, which needs to be 60x to get the minute’s count. However, with a small number of push jobs and tight SLAs, this might lead to rounding errors.
- Using` OccurrenceRate(MINUTE)` and `Gauge` requires metric collection time to sync with metric fetch time.
- `Count` provides an unwindowed cumulative over all time. This was chosen but also needed something windowed for push job availability % calculation in the receiver. 
- So, added `CountSinceLastMeasurement` via this [Tehuti PR](https://github.com/tehuti-io/tehuti/pull/32) to provide a windowed cumulative, emitting the delta every time it’s fetched (every minute in this case) => using tehuti 0.12.2

New Metrics Added (`Count` and `CountSinceLastMeasurement`):
`batch_push_job_success`, `batch_push_job_failed_user_error`, `batch_push_job_failed_non_user_error`
`incremental_push_job_success`, `incremental_push_job_failed_user_error`, `incremental_push_job_failed_non_user_error`

Current flow is `VPJ` checks `push.job.status.upload.enable` config and sends `PushJobDetails` to `/send_push_job_details` path in `Venice-controller`, which writes it to the push job details system store. Approaches considered to emit metrics:
1. Derive success/failure in the controller and emit metrics, tying metric emission and push job details upload via `push.job.status.upload.enable` config. This config is enabled everywhere.
2. Add a new controller endpoint to emit metrics, duplicating work to talk to controllers and serialize/deserialize data.
3. Add two new controller endpoints: one to emit metrics if upload is not enabled, and one to do both (emit metrics and upload push status store) to avoid duplicate work.
4. Evolve `PushJobDetails` to hold `push.job.status.upload.enable` config and move the logic of determining success/failure status to `VenicePushJob`.

Chose approach 1 as it’s the simplest and doesn’t require deployment ordering (controllers -> VPJ) unlike other options and no schema evolution needed.

For Reviewers: Main code changes are in `VeniceHelixAdmin.java` and `PushJobStatusStats.java`.

## How was this PR tested?
GH CI


m-nagarajan:
Messed up the PR with updates. Will open a new one.

PR #1178, opened 2024-09-17
[router][venice-thin-client] Add safeguard against invalid schema ID

majisourav99:
## Add safeguard against invalid schema ID in value schema lookup

If schema repo returns invalid schema id, that fails client requests. This PR adds safeguard against such invalid schemas in meta data handler and client side schema reader.

## How was this PR tested?

CI

### DIFF
--- a/services/venice-router/src/main/java/com/linkedin/venice/router/MetaDataHandler.java
+++ b/services/venice-router/src/main/java/com/linkedin/venice/router/MetaDataHandler.java
@@ -311,7 +311,7 @@ public class MetaDataHandler extends SimpleChannelInboundHandler<HttpRequest> {
         responseObject.setSuperSetSchemaId(superSetSchemaId);
       }
       Collection<SchemaEntry> valueSchemaEntries = schemaRepo.getValueSchemas(storeName);
-      int schemaNum = valueSchemaEntries.size();
+      int schemaNum = (int) valueSchemaEntries.stream().filter(schemaEntry -> schemaEntry.getId() > 0).count();
       MultiSchemaResponse.Schema[] schemas = new MultiSchemaResponse.Schema[schemaNum];
       int index = 0;
       for (SchemaEntry entry: valueSchemaEntries) {
@@ -321,7 +321,7 @@ public class MetaDataHandler extends SimpleChannelInboundHandler<HttpRequest> {
               "Got an invalid schema id ({}) for store {} in handleValueSchemaLookup; will not include this in the {}.",
               entry.getId(),
               storeName,
-              MultiSchemaIdResponse.class.getSimpleName());
+              responseObject.getClass().getSimpleName());
           continue;
         }
         schemas[index] = new MultiSchemaResponse.Schema();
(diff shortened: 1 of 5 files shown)


services/venice-router/src/main/java/com/linkedin/venice/router/MetaDataHandler.java
```diff
@@ -316,6 +316,14 @@ private void handleValueSchemaLookup(ChannelHandlerContext ctx, VenicePathParser
       int index = 0;
       for (SchemaEntry entry: valueSchemaEntries) {
         int schemaId = entry.getId();
+        if (schemaId < 1) {
```
FelixGV, line 319: If that ever happened then the array setting below would fail with `IndexOutOfBoundsException`... I wonder if that ever happened, but it doesn't hurt to have the defensive code!

services/venice-router/src/main/java/com/linkedin/venice/router/MetaDataHandler.java
```diff
@@ -316,6 +316,14 @@ private void handleValueSchemaLookup(ChannelHandlerContext ctx, VenicePathParser
...
+          LOGGER.warn(
+              "Got an invalid schema id ({}) for store {} in handleValueSchemaLookup; will not include this in the {}.",
+              entry.getId(),
+              storeName,
+              MultiSchemaIdResponse.class.getSimpleName());
```
FelixGV, line 324: In this case, the response is of type `MultiSchemaResponse`. Instead of getting the class statically, it might be worth changing this to `responseObject.getClass().getSimpleName()` instead.

services/venice-router/src/main/java/com/linkedin/venice/router/MetaDataHandler.java
```diff
@@ -385,6 +393,14 @@ private void handleValueSchemaIdsLookup(ChannelHandlerContext ctx, VenicePathPar
     }
     Set<Integer> schemaIdSet = new HashSet<>();
     for (SchemaEntry entry: schemaRepo.getValueSchemas(storeName)) {
+      if (entry.getId() < 1) {
+        LOGGER.warn(
+            "Got an invalid schema id ({}) for store {} in handleValueSchemaIdsLookup; will not include this in the {}.",
+            entry.getId(),
+            storeName,
+            MultiSchemaIdResponse.class.getSimpleName());
```
FelixGV, line 401: Maybe change this to `responseObject.getClass().getSimpleName()` as well?

FelixGV (commented):
Thanks Sourav. Just a minor comment...

FelixGV (dismissed):
Thanks!

services/venice-router/src/main/java/com/linkedin/venice/router/MetaDataHandler.java
majisourav99: Sure

services/venice-router/src/main/java/com/linkedin/venice/router/MetaDataHandler.java
majisourav99: Good catch! added some tests as well.

FelixGV (approved):

PR #1179, opened 2024-09-17
[compat][da-vinci][test] Global RT DIV improvement (part 1): RT and VT DIV se…

lluwm:
…paration

**Problem:**
Data Integrity Validator (DIV) is used for validating and discovering data issues residing in the Kafka topics when consumed. However, without a shared global view of the RealTime (RT) topics among all leaders, DIV states for RT topics are scattered among all past leaders and without a way to pass RT DIV from leader to its successors, DIV results can be inaccurate especially during the leadership transitions.

At a high level, we proposes to replicate the leader DIV states into the local VT periodically or on-demand. Followers consume these leader DIV states and sync itself up to the leader state along the way.

**This PR:**
As the first part of the implementation for global DIV improvement, this PR contains the following:
1. introduce a new flag for the global rt div feature.
2. introduce a new realtimeTopicProducerStates in PartitionState in the local rocksdb checkpoint.

```
avro diff: 
    {
      "name": "realtimeTopicProducerStates",
      "doc": "A map that maps upstream Kafka bootstrap server url -> to a map of producer GUID -> producer state for real-time data.",
      "type": {
        "type": "map",
        "values": {
          "type": "map",
          "values": "com.linkedin.venice.kafka.protocol.state.ProducerPartitionState"
        },
        "avro.java.string": "String"
      },
      "default": {}
   }
```

4. divide today's complete DIV into two groups: VT and RT DIV,  and RT DIV is further divided by its source fabric, so that each group can be updated separately based on which topic the Kafka records come from. (without feature enabled, only VT DIV is updated).
5. Adding test to verify the read/write of an `OffsetRecord` with new the new added field.

**Future changes:**
After this PR, as the next step, we will start to implement the following:
1. In the leader replica, create a snapshot of the RT DIV  and replicate it to the local VT by using a chunking supported heartbeat messages (DIV heartbeat).
2. In follower replicas, consume the leader's RT DIV snapshot and sync itself up to the leader state along the way. (this actually needs to be checked in first to handle potential deployment rollbacks).
3. The lifecycle of the in memory RT DIV states will be managed by the leader and is shown in the following figure:
Screenshot 2024-10-02 at 9 34 08 AM

## How was this PR tested?

CI (with or without feature enabled)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -3034,6 +3041,7 @@ private int internalProcessConsumerRecord(
            * the last message of the sequence, which is not a chunk but rather the manifest.
            */
           validateMessage(
+              PartitionTracker.TopicType.VERSION_TOPIC,
               this.kafkaDataIntegrityValidator,
               consumerRecord,
```
gaojieliu, line 3046: So if we want to truly separate RT tracking from VT tracking ,the `consumerRecord` needs to be from VT.
There was an issue we discussed in the past is that for hybrid store, the `consumerRecord` in leader replica is actually the one from RT topic instead of Version Topic, so I think we need to make some logic change in `leader replica` when producing to the local drainer queue in `LeaderProducerCallback`.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -3254,7 +3263,10 @@ protected void validateMessage(
    * consumption in leader is ahead of drainer, leaders and drainers are processing messages at different paces.
    */
   protected void cloneProducerStates(int partition, KafkaDataIntegrityValidator validator) {
-    this.kafkaDataIntegrityValidator.cloneProducerStates(partition, validator);
+    this.kafkaDataIntegrityValidator
+        .cloneProducerStates(PartitionTracker.TopicType.VERSION_TOPIC, partition, validator);
+    this.kafkaDataIntegrityValidator
```
gaojieliu, line 3268: I think we need to carefully think about the future phases for this project.
In the high-level, if we don't choose eDIV, but use a full snapshot, here is the producing logic in my mind:
1. Config a time-interval.
2. When the ingestion progress passes the time interval, `SIT` will produce a full snapshot to VersionTopic per partition.

There are a few prod fabircs and with current logic, RT topic in each prod fabric will be consumed independently, that means the RT DIV can be updated while `SIT` is generating a full snapshot per partition.
There is a race condition, and to avoid it, with today's architecture, we will need to maintain a full snapshot per source fabric per partition.
Please double check in case I had some misunderstanding.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
lluwm: Thanks a lot for your review! @gaojieliu . You are right that `consumerRecord` validated in here can come from both RT and VT topics, so it is inaccurate to put all of them into the VT DIV.

My thought is that this is the `drainer DIV` and its result is inaccurate because of the DCR and ignored messages. Based on the high level design of the snapshot approach, we need to later transfer the full DIV responsibility to the `consumer DIV` which resides in the consumer thread and then gradually retire the use of `drainer DIV`. However, because this the 1st PR for global RT series, DIV heartbeat is yet implemented, so I would like to keep `drainer DIV` the way as is today (a mixed DIV state). When future work is completed,  `drainer DIV` logic can be simply bypassed, and any work that we do today to fix the `drainer DIV` could be thrown away. This is the main reason I didn't change  `drainer DIV` logic. Do you think it makes sense?

gaojieliu: With today's logic, the VT DIV gap issue before reporting completed can still happen.
Let us leave a `TODO` here in case we forget it in the future.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
lluwm: Gaojie, your understanding is correct and I think I see your concern here. Let me try if I understand it right,

The race condition could happen for AA store and when multi consumer thread updates and snapshots the same partition DIV, to address it, potentially we can:
 
1. as you suggested, divides our RT DIV further and into per colo RT DIV, so that per-source-fabric and per-partition DIV is updated by a single consumer thread. (no lock is required, but we may have per-source-fabric per-partition RT DIV HB messages) 
2. or introduce a `partition-level` wr-lock, so that any consumer threads working on same partition require to hold a w-lock when updating, and a r-lock when snapshotting. (sounds more performance overhead)

gaojieliu: 2) carries a lot of overheads and it means the updates to the same partition from different RT topics will be executed sequentially, which is the opposite to the direction of parallelizing the processing for AA/WC workload.

lluwm: Makes sense. Working on a per-fabric RT DIV change.

internal/venice-common/src/main/java/com/linkedin/venice/offsets/OffsetRecord.java
```diff
@@ -147,6 +148,33 @@ public synchronized Map<CharSequence, ProducerPartitionState> getProducerPartiti
     return this.partitionState.producerStates;
   }
 
+  public synchronized void setRealtimeTopicProducerState(
+      String kafkaUrl,
+      GUID producerGuid,
+      ProducerPartitionState state) {
+    partitionState.getRealtimeTopicProducerStates().putIfAbsent(kafkaUrl, new VeniceConcurrentHashMap<>());
```
gaojieliu, line 155: `computeIfAbsent` is better than `putIfAbsent` to avoid unnecessary object creation.

gaojieliu (commented):
If the SIT is not ingesting the leader replica, will SIT load the RT DIV or not?
Also when the replica transitions from leader replicas to follower replica, will SIT offload RT DIV from the memory and reload it when necessary?

Maybe this is not part of the scope of this PR, and some clarification will be hepful.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -2183,15 +2185,28 @@ protected Iterable<PubSubMessage<KafkaKey, KafkaMessageEnvelope, Long>> validate
...
-            partitionConsumptionState);
+        if (!isGlobalRtDivEnabled) {
+          validateMessage(
+              PartitionTracker.VERSION_TOPIC,
+              this.kafkaDataIntegrityValidatorForLeaders,
+              record,
+              isEndOfPushReceived,
+              partitionConsumptionState);
+        } else {
+          PartitionTracker.TopicType rtTopic =
```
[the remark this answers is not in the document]
lluwm: sure, done.

internal/venice-common/src/main/java/com/linkedin/venice/offsets/OffsetRecord.java
lluwm: Good idea. done.

lluwm:
> If the SIT is not ingesting the leader replica, will SIT load the RT DIV or not? Also when the replica transitions from leader replicas to follower replica, will SIT offload RT DIV from the memory and reload it when necessary?
> 
> Maybe this is not part of the scope of this PR, and some clarification will be hepful.

Will update the description part to address this comment.

gaojieliu (approved):
Thanks!

PR #1180, opened 2024-09-17
[vpj] Ignore field props check during subset schema validation.

haoxu07:

Previous issue: During subset schema validation, we will compare field strictly including the customized props. Such props could be JSON string provided by customers, without strict string order.  And when we generate superset schema, we also did not check the props difference. 

The fix is avoiding the props checking when comparing field with the same name, between subset schema and superset schema. We will deep-copy fields from subset schema and superset schema and then do equlity check. 

## How was this PR tested?

ci


m-nagarajan (dismissed):
LGTM, Thanks @haoxu07 for the fix.

xunyin8 (changes requested):
See above comments for additional test request

haoxu07:
> See above comments for additional test request



> See above comments for additional test request

Sure, let me try to add that test case.

xunyin8 (approved):

PR #1181, opened 2024-09-17
[script] Explicitly tell the user to set their JAVA_HOME if their JDK version is invalid

kvargha:

This is to tell the user explicitly how to resolve the JDK version mismatch error. 

## How was this PR tested?

Tested locally

Error message will contain new instructions.


eldernewborn (approved):
Thank you.

nisargthakkar (approved):

PR #1184, opened 2024-09-18
[controller] Superset schema generation should reflect default value updates in new value schema

sixpluszero:
## [controller] Superset schema generation should reflect default value updates in new value schema
When user updates a field default value, we should reflect it in the changes in superset schema, which is used in read-compute / partial update store. This should apply to all the value type, including union schema. 

## How was this PR tested?
Add a new integration test to validate

### DIFF
--- a/internal/venice-client-common/src/main/java/com/linkedin/venice/utils/AvroSupersetSchemaUtils.java
+++ b/internal/venice-client-common/src/main/java/com/linkedin/venice/utils/AvroSupersetSchemaUtils.java
@@ -83,7 +83,7 @@ public class AvroSupersetSchemaUtils {
       case MAP:
         return Schema.createMap(generateSuperSetSchema(existingSchema.getValueType(), newSchema.getValueType()));
       case UNION:
-        return unionSchema(existingSchema, newSchema);
+        return unionSchema(newSchema, existingSchema);
       default:
         throw new VeniceException("Super set schema not supported");
     }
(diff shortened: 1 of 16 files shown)


internal/venice-client-common/src/main/java/com/linkedin/venice/utils/AvroSupersetSchemaUtils.java
```diff
@@ -34,56 +34,56 @@ public static boolean isSupersetSchema(Schema s1, Schema s2) {
...
-   * @param existingSchema schema existing in the repo
-   * @param newSchema schema to be added.
+   * schemas and the generated schema will pick the default value from new value schema.
+   * @param newSchema schema existing in the repo
+   * @param oldSchema schema to be added.
    * @return super-set schema of existingSchema abd newSchema
    */
-  public static Schema generateSuperSetSchema(Schema existingSchema, Schema newSchema) {
-    if (existingSchema.getType() != newSchema.getType()) {
+  public static Schema generateSuperSetSchema(Schema newSchema, Schema oldSchema) {
```
m-nagarajan, line 42: I feel not changing the function definition should make the diff a bit concise. what do you think?

sixpluszero: Do you mean reverse the order back? Let me see if I can achieve that.

sixpluszero: Fixed. Could you check again?

m-nagarajan (commented):
LGTM Thanks @sixpluszero, left one minor comment.

m-nagarajan (approved):
Thanks @sixpluszero

PR #1186, opened 2024-09-18
[vpj] More reorg for the venice-push-job module

eldernewborn:
## Brought heartbeat, jobs packages into the main folder
None of them are Hadoop specific so they should not live under hadoop package.
Updated relevant tests.

## How was this PR tested?
ran the test suit for the venice-push-job module
ran build


clients/venice-push-job/src/main/java/com/linkedin/venice/input/kafka/KafkaInputFormat.java
```diff
@@ -1,8 +1,8 @@
-package com.linkedin.venice.hadoop.input.kafka;
+package com.linkedin.venice.input.kafka;
```
nisargthakkar, line 1: Would prefer keeping this in MR-specific packages. Is it possible?

clients/venice-push-job/src/main/java/com/linkedin/venice/heartbeat/DefaultPushJobHeartbeatSender.java
```diff
@@ -25,7 +25,7 @@
 
 
 @NotThreadSafe
-class DefaultPushJobHeartbeatSender implements PushJobHeartbeatSender {
+public class DefaultPushJobHeartbeatSender implements PushJobHeartbeatSender {
```
eldernewborn, line 28: Please share if there is any concern here.

clients/venice-push-job/src/main/java/com/linkedin/venice/heartbeat/DefaultPushJobHeartbeatSender.java
```diff
@@ -77,7 +77,7 @@ class DefaultPushJobHeartbeatSender implements PushJobHeartbeatSender {
   }
 
   // For test purpose.
-  VeniceWriter<byte[], byte[], byte[]> getVeniceWriter() {
+  public VeniceWriter<byte[], byte[], byte[]> getVeniceWriter() {
```
eldernewborn, line 80: Please share if there is any concerns here.

clients/venice-push-job/src/main/java/com/linkedin/venice/input/kafka/KafkaInputFormat.java
eldernewborn: Sure, I imagined this is not MR specific, I'll put it back, or split it into common and MR specific bits.

eldernewborn: moved them back.

FelixGV:
Thanks for continuing to push on this project @eldernewborn!

Just a minor nitpick: when you squash and merge, please alter the commit message so that it contains only `[vpj]`. We want the prefix tags to contain only deployable services and user-facing libraries.

Thanks!

clients/venice-push-job/src/main/java/com/linkedin/venice/heartbeat/DefaultPushJobHeartbeatSender.java
FelixGV: Usually, the APIs which are intended for tests only are left package private on purpose, so as to avoid spreading their usage outside of test code. This is not a perfect blocker of usage outside of test code, but merely relies on the fact that most code happens to be in different packages, and therefore package-private will make the API unreachable.

In this particular case, IDK if it is a big deal for the API to graduate from intended-for-tests-only to generally/publicly available, but you could assess that and share your opinion. If there is no reason for making the API generally-available, then we would probably prefer keeping the package structure the same to avoid increasing visibility...

eldernewborn: The tests are rearranged and access is set to protected now.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/hadoop/input/kafka/TestKafkaInputFormat.java
```diff
@@ -1,9 +1,9 @@
 package com.linkedin.venice.hadoop.input.kafka;
```
FelixGV, line 1: If the goal is to axe `hadoop` from package names as much as possible (is that the goal? IDK if I understand the goal of this PR, actually; but anyway, if that is the goal...) then why leave the test classes in the `hadoop` package?

The convention is that a given test class should be in the same package as the main code it intends to exercise. This enables us to limit visibility rules to package-private (i.e. no visibility modifier at all), so that APIs which need to do special dependency injection are less visible.

The `KafkaInputFormat` class has a visibility change due to this mismatch.

clients/venice-push-job/src/main/java/com/linkedin/venice/heartbeat/DefaultPushJobHeartbeatSender.java
FelixGV: This change is not necessary, since the factory calling this constructor is also in the new package.

clients/venice-push-job/src/main/java/com/linkedin/venice/heartbeat/DefaultPushJobHeartbeatSender.java
```diff
@@ -77,7 +77,7 @@ class DefaultPushJobHeartbeatSender implements PushJobHeartbeatSender {
   }
 
   // For test purpose.
-  VeniceWriter<byte[], byte[], byte[]> getVeniceWriter() {
+  protected VeniceWriter<byte[], byte[], byte[]> getVeniceWriter() {
```
FelixGV, line 80: This change is not necessary, from my understanding.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/hadoop/input/kafka/TestKafkaInputFormat.java
eldernewborn: The goal is to move non-Hadoop parts out of the Hadoop package. Nisarg pointed out that this input format is tied to MR framework and by that token, Hadoop, so it should stay within that package.

I think at this point, the main class for KafkaInputFormat resides in the same pacakge as the tests for it, both unit and integration.

FelixGV: Got it. So then `KafkaInputFormat:: getLatestOffsets` can be made package-private, rather than public.

clients/venice-push-job/src/main/java/com/linkedin/venice/heartbeat/DefaultPushJobHeartbeatSender.java
eldernewborn: fixed.

clients/venice-push-job/src/main/java/com/linkedin/venice/heartbeat/DefaultPushJobHeartbeatSender.java
eldernewborn: fixed.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/hadoop/input/kafka/TestKafkaInputFormat.java
eldernewborn: fixed.

clients/venice-push-job/src/main/java/com/linkedin/venice/pushjob/VenicePushJobConstants.java
```diff
@@ -1,11 +1,13 @@
-package com.linkedin.venice.hadoop;
+package com.linkedin.venice.pushjob;
```
nisargthakkar, line 1: Can we move it to `com.linkedin.venice.common`?

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/input/kafka/KafkaInputFormat.java
```diff
@@ -49,7 +49,7 @@ public class KafkaInputFormat implements InputFormat<KafkaInputMapperKey, KafkaI
   public static final long DEFAULT_KAFKA_INPUT_MAX_RECORDS_PER_MAPPER = 5000000L;
   private final PubSubTopicRepository pubSubTopicRepository = new PubSubTopicRepository();
 
-  protected Map<TopicPartition, Long> getLatestOffsets(JobConf config) {
```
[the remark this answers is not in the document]
eldernewborn: it's no longer needed, restored to original state.

clients/venice-push-job/src/main/java/com/linkedin/venice/pushjob/VenicePushJobConstants.java
eldernewborn: It's now renamed to vpj

nisargthakkar (approved):

PR #1187, opened 2024-09-19
[server][dvc] transfer offsetRecord and storeVersionState during blob transfer

jingy-li:
## Transfer offsetRecord and storeVersionState during blob transfer

Main code changes:
  1. For P2PFileTransferClientHandler:
      - No longer use inputStream to determine if the transfer is complete. Instead, use BlobTransferPartitionMetadata, which is also needed for updating the offsetRecord later.
      - For the responses, distinguish between file and metadata types. If it's a file type, write the file; if it's a metadata type, use it to update the metadata.
      - In bootstrapFromBlobs, once the metadata is transferred, update the offsetRecord in storageMetadataService and update the SVS via StoreVersionStateSyncer.
  2. For P2PFileTransferServerHandler:
      - After P2PFileTransferServerHandler sends the file and before sending the BLOB_TRANSFER_COMPLETED flag, the server handler sends metadata record information with the partition's latest offset and store version state, retrieved from StorageMetadataService.
 
Minor code change:
  - Move the blob transfer client/server from venice-common to davinci, because it will use storageMetadataService to get the necessary offset/SVS information. Otherwise, it will cause a cyclic dependency.

## How was this PR tested?
Unit test and integration test.
Screenshot 2024-09-19 at 10 24 46 AM


clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/client/P2PFileTransferClientHandler.java
```diff
@@ -1,12 +1,15 @@
-package com.linkedin.venice.blobtransfer.client;
+package com.linkedin.davinci.blobtransfer.client;
 
-import static com.linkedin.venice.blobtransfer.BlobTransferUtils.BLOB_TRANSFER_COMPLETED;
-import static com.linkedin.venice.blobtransfer.BlobTransferUtils.BLOB_TRANSFER_STATUS;
+import static com.linkedin.davinci.blobtransfer.BlobTransferUtils.*;
```
adamxchen, line 3: let's not import all

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/client/P2PFileTransferClientHandler.java
```diff
@@ -129,13 +170,45 @@ protected void channelRead0(ChannelHandlerContext ctx, HttpObject msg) throws Ex
...
 
+  /**
+   * Handles the metadata transfer response.
+   * @param ctx the channel context
+   * @param msg the http object
+   * @throws JsonProcessingException
+   */
+  private BlobTransferPartitionMetadata handlerMetadataTransferResponse(ChannelHandlerContext ctx, HttpObject msg)
+      throws JsonProcessingException {
+    if (msg instanceof HttpContent) {
```
adamxchen, line 187: I think we are pretty sure the metadata has to be `HttpContent` in order to process it so instead of doing the check here, I'd suggest we update the method signature to be ` handlerMetadataTransferResponse(ChannelHandlerContext ctx, HttpContent msg)`

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/client/P2PFileTransferClientHandler.java
```diff
@@ -166,15 +240,15 @@ private String getFileNameFromHeader(HttpResponse response) {
...
+    metadataFuture.toCompletableFuture().complete(metadata);
     ctx.close();
   }
 
   private void resetState() {
     outputFileChannel = null;
     fileName = null;
     fileContentLength = 0;
+    currentTransferType = null;
+    accumulatedMetadataContent.clear();
```
adamxchen, line 252: `clear()` on a ByteBuf does not release the underlying memory; it only resets the read and write indices, making the buffer appear empty without deallocating memory. If we need to release, we need to call `release`

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/client/P2PFileTransferClientHandler.java
```diff
@@ -129,13 +170,45 @@ protected void channelRead0(ChannelHandlerContext ctx, HttpObject msg) throws Ex
...
+          accumulatedMetadataContent.readBytes(contentBytes);
+          ObjectMapper objectMapper = new ObjectMapper();
+          BlobTransferPartitionMetadata metadata = objectMapper
+              .readValue(new String(contentBytes, StandardCharsets.UTF_8), BlobTransferPartitionMetadata.class);
+          LOGGER.info("Metadata transfer completed, topic {} metadata is {}", payload.getTopicName(), metadata);
+          return metadata;
+        }
+      }
+    }
+    return null;
```
adamxchen, line 209: Should we even allow `null` happening for metadata transfer?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/BlobTransferManager.java
```diff
@@ -30,7 +29,7 @@ public interface BlobTransferManager<T> extends AutoCloseable {
    * thrown, but it's wrapped inside the CompletionStage.
    */
   @Experimental
-  CompletionStage<? extends InputStream> get(String storeName, int version, int partition)
+  CompletionStage<? extends BlobTransferPartitionMetadata> get(String storeName, int version, int partition)
```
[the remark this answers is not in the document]
adamxchen: We've talked about making two calls offline which makes the logic cleaner but has some potential issues for the data gap when there's a large amount of data in a short elapsed time, so I am wondering whether it's a good idea to pass `metadataService` in the handler so handler would update it upon receiving it, so we can keep the `InputStream` as the output

clients/da-vinci-client/src/test/java/com/linkedin/davinci/blobtransfer/TestP2PFileTransferClientHandler.java
```diff
@@ -0,0 +1,391 @@
+package com.linkedin.davinci.blobtransfer;
+
+import static com.linkedin.davinci.blobtransfer.BlobTransferUtils.*;
```
adamxchen, line 3: ditto

adamxchen (commented):
Great progress. Left some initial comments. We can discuss more

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/BlobTransferManager.java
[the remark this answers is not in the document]
adamxchen: Or we still spilt the call? Get the metadata first but do not persist. Then get the data file. Given the metadata is small, we should be okay with this gap. After all data is transferred, then we persist the metadata. This looks a bit anti-intuitive tho

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/client/P2PFileTransferClientHandler.java
```diff
@@ -66,9 +78,37 @@ protected void channelRead0(ChannelHandlerContext ctx, HttpObject msg) throws Ex
       // Already end of transfer. Close the connection and completes the future
       if (response.headers().get(BLOB_TRANSFER_STATUS) != null
           && response.headers().get(BLOB_TRANSFER_STATUS).equals(BLOB_TRANSFER_COMPLETED)) {
+        LOGGER.info("Received blob transfer completed response.");
+        resetState();
         handleEndOfTransfer(ctx);
         return;
       }
+      currentTransferType = response.headers().get(BLOB_TRANSFER_TYPE);
```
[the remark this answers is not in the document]
jingy-li: The logic here is: if the msg is a file or metadata, we need to update the current type so that we can decide if to pass it to the handlerMetadataTransferResponse or handleFileTransferResponse. It has to be in the `if` statement to check if it is httpResponse. It is not related to BLOB_TRANSFER_COMPLETED.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/client/P2PFileTransferClientHandler.java
jingy-li: Yes, we need. as handleMetadataTransferContent return value need to be either null or transferred metadata. When last http content receive, return metadata, otherwise it should be null. And in updateStorePartitionMetadata, if metadata is null, it will throw exception.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/client/P2PFileTransferClientHandler.java
```diff
@@ -0,0 +1,309 @@
...
+
+  /**
+   * Given the metadata from blob transfer, update the store partition metadata in storage service.
+   */
+  public void updateStorePartitionMetadata(
+      KafkaStoreIngestionService kafkaStoreIngestionService,
+      StorageService storageService,
+      BlobTransferPartitionMetadata metadata) {
+    if (metadata == null) {
+      LOGGER.error("No metadata received for topic {}. skip updating.", payload.getTopicName());
```
adamxchen, line 291: We are not really "skip updating" 😄  We fail it

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/client/P2PFileTransferClientHandler.java
adamxchen: Hmm..I prefer we should just throw exception when there's any issue to retrieve the metadata rather than waiting until the very end, but I suppose that's fine

clients/da-vinci-client/src/test/java/com/linkedin/davinci/blobtransfer/TestP2PFileTransferServerHandler.java
```diff
@@ -157,10 +203,52 @@ public void testTransferMultipleFiles() throws IOException {
     response = ch.readOutbound();
     Assert.assertTrue(response instanceof LastHttpContent);
     // end of a file2
+
+    // start of metadata
+    response = ch.readOutbound();
+    Assert.assertTrue(response instanceof DefaultHttpResponse);
+    DefaultHttpResponse metadataResponse = (DefaultHttpResponse) response;
```
adamxchen, line 210: How about we verify the response too?

clients/da-vinci-client/src/test/java/com/linkedin/davinci/blobtransfer/TestNettyP2PBlobTransferManager.java
```diff
@@ -48,9 +62,17 @@ public void setUp() throws Exception {
...
     // intentionally use different directories for snapshot and partition so that we can verify the file transfer
-    server = new P2PBlobTransferService(port, tmpSnapshotDir.toString());
-    client = new NettyFileTransferClient(port, tmpPartitionDir.toString());
+    storageMetadataService = mock(StorageMetadataService.class);
+    server = new P2PBlobTransferService(port, tmpSnapshotDir.toString(), storageMetadataService);
+
+    storageService = mock(StorageService.class);
+    storeVersionStateSyncer = mock(BiConsumer.class);
+    Mockito.doNothing().when(storeVersionStateSyncer).accept(anyString(), Mockito.any());
+    Mockito.doReturn(storeVersionStateSyncer).when(storageService).getStoreVersionStateSyncer();
```
adamxchen, line 71: I thought `storageService` is not used? If yes, can we remove it

clients/da-vinci-client/src/test/java/com/linkedin/davinci/blobtransfer/TestP2PFileTransferClientHandler.java
```diff
@@ -0,0 +1,433 @@
...
+        AvroProtocolDefinition.PARTITION_STATE.getSerializer();
+    OffsetRecord offsetRecord = new OffsetRecord(partitionStateSerializer);
+    offsetRecord.setOffsetLag(1000L);
+    expectedMetadata.setOffsetRecord(ByteBuffer.wrap(offsetRecord.toBytes()));
+
+    ObjectMapper objectMapper = new ObjectMapper();
+    String metadataJson = objectMapper.writeValueAsString(expectedMetadata);
+    byte[] metadataBytes = metadataJson.getBytes(CharsetUtil.UTF_8);
+
+    DefaultHttpResponse metadataResponse = new DefaultHttpResponse(HttpVersion.HTTP_1_1, HttpResponseStatus.OK);
```
adamxchen, line 263: If we use `DefaultFullHttpResponse`, we don't need `metadataChunk` and `endOfTransfer`, correct?

clients/da-vinci-client/src/test/java/com/linkedin/davinci/ingestion/DefaultIngestionBackendTest.java
```diff
@@ -78,12 +79,13 @@ public void setUp() {
     when(storageService.openStoreForNewPartition(eq(storeConfig), eq(PARTITION), any())).thenReturn(storageEngine);
 
     // Create the DefaultIngestionBackend instance with mocked dependencies
-    ingestionBackend = new DefaultIngestionBackend(
-        storageMetadataService,
-        storeIngestionService,
-        storageService,
-        blobTransferManager,
-        veniceServerConfig);
+    ingestionBackend = spy(
```
adamxchen, line 82: Why do we need this change?

adamxchen (commented):
Great progress. There are some minor comments. We should be able to merge it soon

clients/da-vinci-client/src/test/java/com/linkedin/davinci/blobtransfer/TestNettyP2PBlobTransferManager.java
```diff
@@ -155,5 +186,13 @@ public void testLocalFileTransfer() throws IOException, ExecutionException, Inte
     Assert.assertTrue(Arrays.equals(Files.readAllBytes(file1), Files.readAllBytes(destFile1)));
     Assert.assertTrue(Arrays.equals(Files.readAllBytes(file2), Files.readAllBytes(destFile2)));
     Assert.assertTrue(Arrays.equals(Files.readAllBytes(file3), Files.readAllBytes(destFile3)));
+
+    // Verify the record is updated
+    Mockito.verify(storageMetadataService, Mockito.times(1))
+        .put(TEST_STORE + "_v" + TEST_VERSION, TEST_PARTITION, expectOffsetRecord);
+
+    // Verify the store version state is updated
+    Mockito.verify(storageMetadataService, Mockito.times(1))
```
adamxchen, line 195: This verifies we updated the metadata on client side. Can we also add a few lines to ensure the metadata service is used to retrieve metadata?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/client/P2PMetadataTransferHandler.java
```diff
@@ -0,0 +1,119 @@
...
+
+    ByteBuf content = msg.content();
+    byte[] metadataBytes = new byte[content.readableBytes()];
+    ObjectMapper objectMapper = ObjectMapperFactory.getInstance();
+    content.readBytes(metadataBytes);
+    BlobTransferPartitionMetadata transferredMetadata =
+        objectMapper.readValue(new String(metadataBytes, StandardCharsets.UTF_8), BlobTransferPartitionMetadata.class);
+    if (transferredMetadata == null) {
+      LOGGER.error("No transferPartitionMetadata received for topic {}. ", payload.getTopicName());
+      throw new VeniceException("No transferPartitionMetadata received for topic " + payload.getTopicName());
```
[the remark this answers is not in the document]
jingy-li: Lets call it transferredMetadata to differentiate with the metadata, which is a member variable.

adamxchen (approved):
Great process!

PR #1189, opened 2024-09-19
[fast-client] Replace per-store per-route tracking by per-route tracking

gaojieliu:
This PR is trying to reduce the total number of metrics emitted by fast-client.
Previously, Fast-Client will have the instance-level tracking for each store, which will create a lot of metrics when the application is using multiple stores, and this code change will only emit tacking metrics (qps, latency) in instance level rather than per store per instance.

This PR also removes the pending request metric, which is per store per route, and we will add back this metric in a separate PR when we improve `InstanceHealthMonitor` to be cluster-level instead of store-level.

## How was this PR tested?
CI

### DIFF
--- a/clients/venice-client/src/main/java/com/linkedin/venice/fastclient/stats/ClusterRouteStats.java
+++ b/clients/venice-client/src/main/java/com/linkedin/venice/fastclient/stats/ClusterRouteStats.java
@@ -39,7 +39,7 @@ public class ClusterRouteStats {
       String instanceName = instanceUrl;
       try {
         URL url = new URL(instanceUrl);
-        instanceName = url.getHost() + "_" + url.getPort();
+        instanceName = url.getHost();
       } catch (MalformedURLException e) {
         LOGGER.error("Invalid instance url: {}", instanceUrl);
       }
--- a/clients/venice-client/src/main/java/com/linkedin/venice/fastclient/stats/ClusterRouteStats.java
+++ b/clients/venice-client/src/main/java/com/linkedin/venice/fastclient/stats/ClusterRouteStats.java
@@ -1,0 +1,115 @@
+package com.linkedin.venice.fastclient.stats;
+
+import com.linkedin.venice.read.RequestType;
+import com.linkedin.venice.stats.AbstractVeniceHttpStats;
+import com.linkedin.venice.stats.StatsUtils;
+import com.linkedin.venice.stats.TehutiUtils;
+import com.linkedin.venice.utils.concurrent.VeniceConcurrentHashMap;
+import io.tehuti.metrics.MetricsRepository;
+import io.tehuti.metrics.Sensor;
+import io.tehuti.metrics.stats.OccurrenceRate;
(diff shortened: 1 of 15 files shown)


clients/venice-client/src/main/java/com/linkedin/venice/fastclient/stats/ClusterRouteStats.java
```diff
@@ -0,0 +1,115 @@
...
+      MetricsRepository metricsRepository,
+      String clusterName,
+      String instanceUrl,
+      RequestType requestType) {
+    String combinedKey = clusterName + "-" + instanceUrl + "-" + requestType.toString();
+    return perRouteStatMap.computeIfAbsent(combinedKey, ignored -> {
+      String instanceName = instanceUrl;
+      try {
+        URL url = new URL(instanceUrl);
+        instanceName = url.getHost() + "_" + url.getPort();
```
xunyin8, line 42: What's the reason to include port here? Asking because it might affect our expected behavior when we start using gRPC which have different port.

xunyin8 (commented):
LGTM thanks Gaojie just one small comment.

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/stats/ClusterRouteStats.java
gaojieliu: `Port` is actually not required from my understanding considering Venice Server is not sharing the hardware with other components and I can remove it.

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/stats/ClusterRouteStats.java
```diff
@@ -0,0 +1,115 @@
...
+      } catch (MalformedURLException e) {
+        LOGGER.error("Invalid instance url: {}", instanceUrl);
```
m-nagarajan, line 44: Is it intentional to go continue after logging this?

m-nagarajan (commented):
LGTM overall. Thank you.
Looks like we didn't have any tests around this. Can we add some unit tests.
Probably will need that for diff coverage anyways?

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/stats/ClusterRouteStats.java
gaojieliu: I guess so and this is mainly for defensive coding and in practice, it shouldn't happen and if it does happen, I guess it might be fine to just use the original instance url as the metric name.

xunyin8 (dismissed):

gaojieliu:
> LGTM overall. Thank you. Looks like we didn't have any tests around this. Can we add some unit tests. Probably will need that for diff coverage anyways?

Good question!
Let me see how to add some metric validation in the current test cases.

m-nagarajan (approved):
Thanks @gaojieliu

PR #1190, opened 2024-09-20
[controller] Add support for controller instance tagging

kvargha:

In order for the controller to be migrated to an ephemeral system like Kubernetes and provide isolation for cluster assignment, the controller needs to be able to dynamically set its instance tag.

This instance tag can be specified in the config by setting `controller.instance.tag.list`

## How was this PR tested?

Added new tests.

### DIFF
--- a/internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
@@ -214,8 +214,8 @@ public class ConfigKeys {
   public static final String CONTROLLER_CLUSTER_ZK_ADDRESSS = "controller.cluster.zk.address";
   // Name of the Helix cluster for controllers
   public static final String CONTROLLER_CLUSTER = "controller.cluster.name";
-  // What tag to assign to a controller instance
-  public static final String CONTROLLER_INSTANCE_TAG = "controller.instance.tag";
+  // What tags to assign to a controller instance
+  public static final String CONTROLLER_INSTANCE_TAG_LIST = "controller.instance.tag.list";
 
   /** List of forbidden admin paths */
   public static final String CONTROLLER_DISABLED_ROUTES = "controller.cluster.disabled.routes";
--- a/internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
@@ -214,6 +214,8 @@ public class ConfigKeys {
   public static final String CONTROLLER_CLUSTER_ZK_ADDRESSS = "controller.cluster.zk.address";
   // Name of the Helix cluster for controllers
   public static final String CONTROLLER_CLUSTER = "controller.cluster.name";
+  // What tag to assign to a controller instance
+  public static final String CONTROLLER_INSTANCE_TAG = "controller.instance.tag";
 
   /** List of forbidden admin paths */
   public static final String CONTROLLER_DISABLED_ROUTES = "controller.cluster.disabled.routes";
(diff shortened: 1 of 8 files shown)


internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
```diff
@@ -214,6 +214,8 @@ private ConfigKeys() {
   public static final String CONTROLLER_CLUSTER_ZK_ADDRESSS = "controller.cluster.zk.address";
   // Name of the Helix cluster for controllers
   public static final String CONTROLLER_CLUSTER = "controller.cluster.name";
+  // What tag to assign to a controller instance
+  public static final String CONTROLLER_INSTANCE_TAG = "controller.instance.tag";
```
nisargthakkar, line 218: Can we support multiple tags per instance?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerClusterConfig.java
```diff
@@ -635,6 +637,7 @@ public VeniceControllerClusterConfig(VeniceProperties props) {
      */
     this.adminCheckReadMethodForKafka = props.getBoolean(ADMIN_CHECK_READ_METHOD_FOR_KAFKA, true);
     this.controllerClusterName = props.getString(CONTROLLER_CLUSTER, "venice-controllers");
+    this.controllerInstanceTagList = props.getList(CONTROLLER_INSTANCE_TAG_LIST, new ArrayList<>());
```
nisargthakkar, line 640: Let's use `Collections.emptyList` instead

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/controller/TestVeniceHelixAdminWithSharedEnvironment.java
```diff
@@ -2054,4 +2054,34 @@ public void testRaceConditionFixForKillOfflinePushAndVersionSwap(boolean isKillO
...
+  @Test
+  public void testInstanceTagging() {
+    List<String> instanceTagList = Arrays.asList("GENERAL", "TEST");
+    String controllerClusterName = "venice-controllers";
+    int newAdminPort = controllerConfig.getAdminPort() + 1; /* Note: dummy port */
+    PropertyBuilder builder = new PropertyBuilder().put(controllerProps.toProperties()).put("admin.port", newAdminPort);
+
+    VeniceProperties newControllerProps = builder.build();
+    VeniceControllerClusterConfig newConfig = new VeniceControllerClusterConfig(newControllerProps);
+    VeniceHelixAdmin newHelixAdmin = new VeniceHelixAdmin(
```
nisargthakkar, line 2066: Since we've set the config in `AbstractTestVeniceHelixAdmin`, we shouldn't need to create a new `VeniceHelixAdmin` object

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -793,6 +794,14 @@ private synchronized void connectToControllerCluster() {
     }
     controllerClusterKeyBuilder = new PropertyKey.Builder(tempManager.getClusterName());
     helixManager = tempManager;
+
+    VeniceControllerClusterConfig controllerConfig = multiClusterConfigs.getCommonConfig();
+    List<String> instanceTagList = controllerConfig.getControllerInstanceTagList();
```
nisargthakkar, line 799: Since instance tag is an instance-level config, we should add the `getInstanceTagList` function in `multiClusterConfigs` and handle the part about fetching from common configs there?

nisargthakkar (approved):

PR #1192, opened 2024-09-21
[server] Race condition fix for re-subscription of real-time topic partitions.

haoxu07:

Recently we introduced re-subscription feature for different store versions. Leader and follower partitions from different version topics will experience re-subscription triggered by store version role change concurrently.  

Problem: 
Two `StoreIngestionTask` threads for different store versions (store version 1 and store version 2) try to do re-subscription (unsub and sub) for the same real-time topic partition concurrently. During re-subscription triggered by one store version, operation of `consumer.unSubscribe` to remove assignment of the topic partition and operation of `consumerToConsumptionTask.get(consumer).removeDataReceiver(pubSubTopicPartition)` are sequentially executed. It is possible that `StoreIngestionTask` thread for store version 2 got the same `ConsumptionTask`, but `DataReceiver` inside `ConsumptionTask` is still from store version 1). As each `ConsumptionTask` will only allow one `DataReceiver` for one particular real-time topic partition, the `DataReceiver` from store version 2 will not be able to be added to this `ConsumptionTask`.

Solution: 
Using `SharedKafkaConsumer` level lock to protect `KafkaConsumcerService` for `unSubscribe`, `SetDataReceiver`, `unSubscribeAll` to guarantee there will one real-time partition from specific store version doing `DataReceiver` related assignment change.  

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaConsumerService.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaConsumerService.java
@@ -222,6 +222,10 @@ public abstract class KafkaConsumerService extends AbstractKafkaConsumerService 
     versionTopicToTopicPartitionToConsumer.compute(versionTopic, (k, topicPartitionToConsumerMap) -> {
       if (topicPartitionToConsumerMap != null) {
         topicPartitionToConsumerMap.forEach((topicPartition, sharedConsumer) -> {
+          /**
+           * Refer {@link KafkaConsumerService#startConsumptionIntoDataReceiver} for avoiding race condition caused by
+           * setting data receiver and unsubscribing concurrently for the same topic partition on a shared consumer.
+           */
           synchronized (sharedConsumer) {
             sharedConsumer.unSubscribe(topicPartition);
             removeTopicPartitionFromConsumptionTask(sharedConsumer, topicPartition);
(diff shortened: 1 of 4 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaConsumerService.java
```diff
@@ -237,8 +239,10 @@ public void unsubscribeAll(PubSubTopic versionTopic) {
   public void unSubscribe(PubSubTopic versionTopic, PubSubTopicPartition pubSubTopicPartition) {
     PubSubConsumerAdapter consumer = getConsumerAssignedToVersionTopicPartition(versionTopic, pubSubTopicPartition);
     if (consumer != null) {
-      consumer.unSubscribe(pubSubTopicPartition);
-      consumerToConsumptionTask.get(consumer).removeDataReceiver(pubSubTopicPartition);
+      synchronized (consumer) {
```
lluwm, line 242: Could you also check  function `batchUnsubscribe` in this file? feel like it also needs a similar fix also need a similar fix.

haoxu07: As mentioned in test, `batchUnsubscribe` is only for batch-only stores, it will not unsubscribe real-time topic partition. This issue is mainly related to real-time topic partition.

lluwm: If this assumption holds for future usage, then let's make it clear in the comments that don't use `batchUnsubscribe` for unsubscribing real time topic partition. Otherwise, later who forgets about such requirements would spend another round of time and energy to debug a similar issue.

lluwm (approved):
Thanks @haoxu07 for the fix and it looks good to me!

PR #1193, opened 2024-09-23
[vpj][common][dvc] Minor logging change

majisourav99:
##  Minor logging change to add exception in the log

Add more info in logger with exception log.
 
## How was this PR tested?

CI

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaStoreIngestionService.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaStoreIngestionService.java
@@ -1195,7 +1195,8 @@ public class KafkaStoreIngestionService extends AbstractVeniceService implements
       LOGGER.error(
           "Caught exception when deserializing offset record byte array: {} for replica: {}",
           Arrays.toString(offsetRecordByteArray),
-          Utils.getReplicaId(topicName, partition));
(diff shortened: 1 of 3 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaStoreIngestionService.java
```diff
@@ -1195,7 +1195,8 @@ public void updatePartitionOffsetRecords(String topicName, int partition, ByteBu
       LOGGER.error(
           "Caught exception when deserializing offset record byte array: {} for replica: {}",
           Arrays.toString(offsetRecordByteArray),
-          Utils.getReplicaId(topicName, partition));
+          Utils.getReplicaId(topicName, partition),
```
sixpluszero, line 1198: For not-hot-path logging it is always good to put exception as well, just curious did you find this issue during real debugging experience or you just happen to see it and change it?

sixpluszero (approved):

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaStoreIngestionService.java
majisourav99: While debugging some oncall issues, there was no visibility as to why some method call fails. If being verbose you think might be an issue we can rate limit it.

PR #1194, opened 2024-09-23
[controller] Make setting instance group tags for controller cluster resources configurable

kvargha:

Make setting instance group tags for controller cluster resources configurable. This is needed to provide cluster resource level isolation.

## How was this PR tested?

Added new tests

### DIFF
--- a/internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/controller/TestHAASController.java
+++ b/internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/controller/TestHAASController.java
@@ -45,7 +45,7 @@ import org.testng.annotations.Test;
 public class TestHAASController {
   private Properties enableControllerClusterHAASProperties;
   private Properties enableControllerAndStorageClusterHAASProperties;
-  private final String instanceTag = "GENERAL";
+  private final static String instanceTag = "GENERAL";
 
   @BeforeClass
   public void setUp() {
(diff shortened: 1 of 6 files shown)


services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerMultiClusterConfig.java
```diff
@@ -283,6 +283,10 @@ public long getServiceDiscoveryRegistrationRetryMS() {
     return getCommonConfig().getServiceDiscoveryRegistrationRetryMS();
   }
 
+  public String getControllerResourceInstanceGroupTag() {
+    return getCommonConfig().getControllerResourceInstanceGroupTag();
+  }
```
nisargthakkar, line 288: We should not use this from common configs. This needs to be set at a per-cluster level

nisargthakkar (changes requested):

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/controller/TestHAASController.java
```diff
@@ -44,18 +45,38 @@
...
 
   @BeforeClass
   public void setUp() {
     enableControllerClusterHAASProperties = new Properties();
     enableControllerClusterHAASProperties.put(ConfigKeys.CONTROLLER_CLUSTER_LEADER_HAAS, String.valueOf(true));
     enableControllerClusterHAASProperties
         .put(ConfigKeys.CONTROLLER_HAAS_SUPER_CLUSTER_NAME, HelixAsAServiceWrapper.HELIX_SUPER_CLUSTER_NAME);
+    enableControllerClusterHAASProperties.put(ConfigKeys.CONTROLLER_RESOURCE_INSTANCE_GROUP_TAG, instanceTag);
+    enableControllerClusterHAASProperties.put(ConfigKeys.CONTROLLER_INSTANCE_TAG_LIST, instanceTag);
+
```
nisargthakkar, line 58: Can we add this in the test specific for it?

nisargthakkar (approved):
Code change looks good, but there are some coverage checks that are failing

nisargthakkar (approved):
LGTM. Thanks

PR #1195, opened 2024-09-24
[server] Added metrics to track batch processing perf 

gaojieliu:

This PR added the following metrics:
Host/store-version level:
batch_processing_request
batch_processing_request_size
batch_processing_request_records
batch_processing_request_latency
batch_processing_request_error

Host/Store level:
leader_produce_latency

## How was this PR tested?
CI

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/IngestionStatsReporter.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/IngestionStatsReporter.java
@@ -2,6 +2,11 @@ package com.linkedin.davinci.stats;
 
 import static com.linkedin.davinci.stats.IngestionStats.BATCH_FOLLOWER_OFFSET_LAG;
 import static com.linkedin.davinci.stats.IngestionStats.BATCH_LEADER_OFFSET_LAG;
+import static com.linkedin.davinci.stats.IngestionStats.BATCH_PROCESSING_REQUEST;
+import static com.linkedin.davinci.stats.IngestionStats.BATCH_PROCESSING_REQUEST_ERROR;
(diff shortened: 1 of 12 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/IngestionStatsReporter.java
```diff
@@ -226,6 +231,44 @@ protected void registerStats() {
...
+              this,
+              () -> getStats().getBatchProcessingRequestError(),
+              BATCH_PROCESSING_REQUEST_ERROR));
+      registerSensor(
+          new IngestionStatsGauge(
+              this,
+              () -> getStats().getBatchProcessingRequestRecords(),
+              0,
+              BATCH_PROCESSING_REQUEST_RECORDS));
+      registerSensor(
```
sixpluszero, line 248: Just a comment: This PR register a few more per store metrics which is what we are trying to avoid. Do you think it is important to have all of them? Maybe not _size/ request_ metric?

gaojieliu: These detailed metrics will help debug the ingestion performance.

Considering we have removed many dead stores, maybe these new metrics can be fine?

gaojieliu: I can remove the store-level aggregation and only keep version-level metrics for batch processing.

gaojieliu: The batch-processing metrics don't have the store-level aggregation, just host-level.

sixpluszero: OK I see. Yeah if this is the case, then it should be ok to me.

PR #1196, opened 2024-09-24
[server] Dropping unassigned partitions

### LINKED ISSUE #650: [BUG] Reconcile on disk state with assigned partitions for given resource and delete unused partitions
### Willingness to contribute

### Venice version

Observed since March in production

### System information

- **OS Platform and Distribution (e.g., Linux Ubuntu 20.0)**: Marner 5.15.111.1-1.cm2
- **JDK version**: 17

### Describe the problem

Currently the on-disk state inside the rocksDB folder of a venice server for a give resource (store version) can contain folders for partitions that are no longer assigned to the host, due to missed "partition drops".

We need to ensure that only the known (metadata ?)  or "assigned" partitions are present in the folder for each store_version , and remove the rest very quickly in order to recover disk space.

It is NOT sufficient to clean up this data on start without some introspection.  Venice relies on delayed rebalance on it's controller to avoid unnecessary bootstraps, so some examination may be needed to determine what the appropriate disk state should be. 

### Tracking information

_No response_

### Code to reproduce bug

_No response_

- [X] `Server`: This is the component that persists all the store data.
- [x] `Da Vinci Client`: This is an embedded, stateful client that materializes store data locally.

kristyelee:
The goal is to remove unassigned partitions that are on disk state

This involves storage service and engine, which upon starting (constructed), will go through folders on disk and verify status of folders
Helix's ideal state is used for consulting partitions and their assignment
The change proposed includes defining a function in VeniceServer that goes through partitions in idealState and compares to partitions in storageEngine
StorageEngine object initialization is also being updated to run the new function that performs this verification. 

Resolves #650 

## How was this PR tested?
Will write corresponding test.
EDIT: A unit test has been written.


clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -587,6 +587,19 @@ void resubscribeForAllPartitions() throws InterruptedException {
...
+   * Removes partitions that are not subscribed / not assigned.
+   */
+
+  public synchronized void removeUnsubscribedPartitions() {
+    throwIfNotRunning();
+    for (PartitionConsumptionState partitionConsumptionState: partitionConsumptionStateMap.values()) {
+      if (!partitionConsumptionState.isSubscribed()) {
+        partitionConsumptionStateMap.remove(partitionConsumptionState.getPartition(), partitionConsumptionState);
+      }
+    }
```
kvargha, line 600: I believe in the meeting we discussed that this isn't the correct way to check for partitions to be removed. Please correct me if I'm wrong.

services/venice-server/src/main/java/com/linkedin/venice/server/VeniceServer.java
```diff
@@ -42,10 +44,7 @@
 import com.linkedin.venice.listener.ServerReadMetadataRepository;
 import com.linkedin.venice.listener.ServerStoreAclHandler;
 import com.linkedin.venice.listener.StoreValueSchemasCacheService;
-import com.linkedin.venice.meta.ReadOnlyLiveClusterConfigRepository;
-import com.linkedin.venice.meta.ReadOnlySchemaRepository;
-import com.linkedin.venice.meta.ReadOnlyStoreRepository;
-import com.linkedin.venice.meta.StaticClusterInfoProvider;
+import com.linkedin.venice.meta.*;
```
kvargha, line 47: Please don't import everything. IntelliJ does this automatically, which is a bit annoying.

services/venice-server/src/main/java/com/linkedin/venice/server/VeniceServer.java
```diff
@@ -64,13 +63,13 @@
 import com.linkedin.venice.utils.Utils;
 import com.linkedin.venice.utils.lazy.Lazy;
 import io.tehuti.metrics.MetricsRepository;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.List;
-import java.util.Optional;
+import java.util.*;
```
kvargha, line 66: Same here

kvargha (changes requested):

internal/venice-common/src/main/java/com/linkedin/venice/helix/SafeHelixDataAccessor.java
```diff
@@ -82,7 +82,7 @@ public <T extends HelixProperty> boolean updateProperty(PropertyKey key, DataUpd
     return helixDataAccessor.updateProperty(key, updater, value);
   }
 
-  public <T extends HelixProperty> T getProperty(PropertyKey key) {
+  public static <T extends HelixProperty> T getProperty(PropertyKey key) {
```
kvargha, line 85: It looks like making this static is causing your tests to fail.

kvargha (changes requested):

services/venice-server/src/main/java/com/linkedin/venice/server/VeniceServer.java
```diff
@@ -714,7 +714,13 @@ private Function<String, Boolean> functionToCheckWhetherStorageEngineShouldBeKep
     return storageEngineName -> {
       String storeName = Version.parseStoreFromKafkaTopicName(storageEngineName);
 
-      AbstractStorageEngine storageEngine = storageService.getStorageEngine(storageEngineName);
```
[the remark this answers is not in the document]
kvargha: `checkWhetherStorageEngineShouldBeKeptOrNot` determines whether or not the entire `storageEngine` should be dropped, and where it's called it assumes that the `storageEngine` hasn't been opened yet.

What we should do here is modify the `StorageService` constructor to take another function that determines whether a partition should be kept or not. Ex: `checkWhetherStoragePartitionShouldBeKeptOrNot`

This new function can be invoked [here](https://github.com/linkedin/venice/blob/04aeaebb1fa56b5f392336074120ac9eb72acff1/clients/da-vinci-client/src/main/java/com/linkedin/davinci/storage/StorageService.java#L256C13-L256C55) after the `storageEngine` has been opened.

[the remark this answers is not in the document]
kristyelee: This idea makes sense. I will integrate this.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/storage/StorageService.java
```diff
@@ -88,6 +89,7 @@ public class StorageService extends AbstractVeniceService {
       boolean restoreDataPartitions,
       boolean restoreMetadataPartitions,
       Function<String, Boolean> checkWhetherStorageEngineShouldBeKeptOrNot,
+      Function<AbstractStorageEngine, Void> checkWhetherStoragePartitionsShouldBeKeptOrNot,
```
kvargha, line 92: Can we create an additional constructor that doesn't have `checkWhetherStoragePartitionsShouldBeKeptOrNot`, so if users don't want to use this, they wouldn't have to pass in a functional interface that does nothing.

kristyelee: Ok, that sounds correct; `checkWhetherStoragePartitionsShouldBeKeptOrNot` is mainly being written in VeniceServer.

services/venice-server/src/main/java/com/linkedin/venice/server/VeniceServer.java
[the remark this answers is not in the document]
kristyelee: Operations involving Helix [`HelixParticipationService`, `SafeHelixDataAccessor`] may require that services are started for `StorageService.` [requires that Venice server has started]. `SafeHelixDataAccessor` is used to obtain the ideal state as referenced in #650. Will have to check the structure here so that `checkWhetherStoragePartitionsShouldBeKeptOrNot` executes correctly.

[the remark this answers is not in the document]
kvargha: Try moving the execution of `checkWhetherStoragePartitionsShouldBeKeptOrNot` to after `HelixParticipationService` is initialized in VeniceServer.

services/venice-server/src/main/java/com/linkedin/venice/server/VeniceServer.java
```diff
@@ -601,6 +600,10 @@ public void start() throws VeniceException {
       service.start();
     }
 
+    for (AbstractStorageEngine storageEngine: storageService.getStorageEngineRepository().getAllLocalStorageEngines()) {
```
kristyelee, line 603: I have moved the function here which is after services start, so that `HelixParticipationService` is started/initialized for referencing the ideal state in function.
I have reorganized the code in `StorageService` to its initial state.

kvargha: It looks like `HelixParticipationService` is started asynchronously, and it is causing some tests to fail.

We can try using [managerFuture](https://github.com/linkedin/venice/blob/df1c0b9f3b19a7dc4fbd3113512e3a21abbc278f/services/venice-server/src/main/java/com/linkedin/venice/server/VeniceServer.java#L356C3-L356C83). This will runs things after `HelixParticipationService` has started.

Ex:
```
managerFuture.thenApply(manager -> {
      // Call checkWhetherStoragePartitionsShouldBeKeptOrNot here
});
```

kristyelee:
A few tests still have yet to pass with the changes that include the check for partition assignment. [e.g. this [test](https://github.com/linkedin/venice/actions/runs/11132688423/job/30943346267?pr=1196)] 
I have to check whether these are related to the changes I've submitted; also some of these tests seem to pass locally.

services/venice-server/src/main/java/com/linkedin/venice/server/VeniceServer.java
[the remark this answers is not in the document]
kristyelee: (Resolution is here: https://github.com/linkedin/venice/pull/1196#discussion_r1781747306; when `HelixParticipationService` is started, `managerFuture` is completed in `HelixParticipationService.`).

kristyelee:
(Suggested addition to code, that seems directly useful but is no longer / not used actively:

In the `HelixParticipationService` file, similar to the getInstance function, I also wrote a getHelixManager function. I tried using this before in access after `HelixParticipationService` is instantiated, but have changed this implementation. )

kristyelee:
> (Suggested addition to code, that seems directly useful but is no longer / not used actively:
> 
> In the `HelixParticipationService` file, similar to the getInstance function, I also wrote a getHelixManager function. I tried using this before in access after `HelixParticipationService` is instantiated, but have changed this implementation. )

(Reverted code in this file back to original state but thought I should make a note of this).

clients/da-vinci-client/src/main/java/com/linkedin/davinci/storage/StorageService.java
```diff
@@ -371,6 +375,38 @@ public synchronized AbstractStorageEngine openStore(
     return engine;
   }
 
+  public synchronized void checkWhetherStoragePartitionsShouldBeKeptOrNot(SafeHelixManager manager) {
+    if (getStorageEngineRepository() == null || manager == null) {
```
ZacAttack, line 379: nit: getStorageEngineRepository() == null isn't needed.  The getter is based off of a final member variable.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/storage/StorageService.java
```diff
@@ -371,6 +375,38 @@ public synchronized AbstractStorageEngine openStore(
...
+    for (AbstractStorageEngine storageEngine: getStorageEngineRepository().getAllLocalStorageEngines()) {
+      String storageEngineName = storageEngine.getStoreVersionName();
+      String storeName = Version.parseStoreFromKafkaTopicName(storageEngineName);
+      PropertyKey.Builder propertyKeyBuilder =
+          new PropertyKey.Builder(configLoader.getVeniceClusterConfig().getClusterName());
+      SafeHelixDataAccessor helixDataAccessor = manager.getHelixDataAccessor();
+      IdealState idealState = helixDataAccessor.getProperty(propertyKeyBuilder.idealStates(storeName));
+
+      if (idealState == null) {
+        return;
```
ZacAttack, line 391: This seems wrong.  This will abort the whole function, but we would want to iterate on the remainder of the storage engines to determine if we should keep them.  I think instead of return, we should use continue

clients/da-vinci-client/src/main/java/com/linkedin/davinci/storage/StorageService.java
```diff
@@ -371,6 +375,38 @@ public synchronized AbstractStorageEngine openStore(
...
+
+      Set<Integer> idealStatePartitionIds = new HashSet<>();
+
+      idealState.getPartitionSet().stream().forEach(partitionId -> {
+        idealStatePartitionIds.add(Integer.parseInt(partitionId));
+      });
+      Set<Integer> storageEnginePartitionIds = storageEngine.getPartitionIds();
+
+      for (Integer storageEnginePartitionId: storageEnginePartitionIds) {
+        if (idealStatePartitionIds.contains(storageEnginePartitionId)) {
```
ZacAttack, line 402: This also isn't quite right.  The ideal state is the superset of all partitions for the store.  All partitions ID's in the storageEngine will be in the superset.

We need to parse more information from the ideal state.  Specifically, for each partition that is on this node, we need to determine if the ideal state for that partition id contains the node name.

To better paint the picture, an ideal state is a big json document that looks like this:

>    "Test_Store_Migration_Demo_v1_0": {
>       "lor1-app56585.prod.linkedin.com_1690": "LEADER",
>       "lor1-app56614.prod.linkedin.com_1690": "STANDBY",
>       "lor1-app110448.prod.linkedin.com_1690": "STANDBY"
>     },
>     "Test_Store_Migration_Demo_v1_1": {
>       "lor1-app56586.prod.linkedin.com_1690": "LEADER",
>       "lor1-app71895.prod.linkedin.com_1690": "STANDBY",
>       "lor1-app111181.prod.linkedin.com_1690": "STANDBY"
>     },

Here, for partitions 1 and 0, we have three hosts assigned roles for that partition.  What we should do is, does the current host we're working with reside in these lists?

services/venice-server/src/main/java/com/linkedin/venice/server/VeniceServer.java
```diff
@@ -310,14 +312,20 @@ private List<AbstractVeniceService> createServices() {
         ? new RocksDBMemoryStats(metricsRepository, "RocksDBMemoryStats", plainTableEnabled)
         : null;
 
+    boolean whetherToRestoreDataPartitions = !isIsolatedIngestion()
```
ZacAttack, line 315: @sixpluszero , can you confirm this to be right?

I think this is PR we maybe shouldn't bother with this.  The original code has this hard coded to true, i think we should just pass true instead of potentially altering the behavior as this juncture.

services/venice-server/src/main/java/com/linkedin/venice/server/VeniceServer.java
```diff
@@ -310,14 +312,20 @@ private List<AbstractVeniceService> createServices() {
...
         veniceConfigLoader,
         storageEngineStats,
         rocksDBMemoryStats,
         storeVersionStateSerializer,
         partitionStateSerializer,
-        metadataRepo);
+        metadataRepo,
+        whetherToRestoreDataPartitions,
+        true,
+        functionToCheckWhetherStorageEngineShouldBeKeptOrNot());
```
ZacAttack, line 328: in the way that you've structured this PR it seems what you're doing is waiting for the managerFuture to complete as your means to do the local state clean up.  That's not a bad idea actually, I think that's actually pretty clever.  However, if you do that, then we don't need to mess with this at all.  Instead we can leave this code alone, and just use the shortened constructor that we were using before.

ZacAttack (changes requested):

services/venice-server/src/main/java/com/linkedin/venice/server/VeniceServer.java
kristyelee: (Revised).

clients/da-vinci-client/src/main/java/com/linkedin/davinci/storage/StorageService.java
```diff
@@ -371,6 +375,38 @@ public synchronized AbstractStorageEngine openStore(
...
+      SafeHelixDataAccessor helixDataAccessor = manager.getHelixDataAccessor();
+      IdealState idealState = helixDataAccessor.getProperty(propertyKeyBuilder.idealStates(storeName));
+
+      if (idealState == null) {
+        return;
+      }
+
+      Set<Integer> idealStatePartitionIds = new HashSet<>();
+
+      idealState.getPartitionSet().stream().forEach(partitionId -> {
```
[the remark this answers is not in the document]
kristyelee: I have updated this code and reflected this understanding.

kristyelee:
I wrote the related unit test for this PR. 
GitHub tests pass for program, except for one Integration test unrelated to my change.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/storage/StorageService.java
```diff
@@ -371,6 +376,43 @@ public synchronized AbstractStorageEngine openStore(
...
+          int partitionId = RocksDBUtils.parsePartitionIdFromPartitionDbName(partitionDbName);
+          if (!storageEnginePartitionIds.contains(partitionId)) {
+            continue;
+          }
+          for (String hostName: entry.getValue().keySet()) {
+            if (hostName.equals(instanceHostName)) {
+              keepPartition = true;
+              break;
+            }
+          }
```
kvargha, line 407: Instead of looping over every key, can we instead check if the key `instanceHostName` exists in `entry`?

clients/da-vinci-client/src/test/java/com/linkedin/davinci/storage/StorageServiceTest.java
```diff
@@ -121,4 +136,76 @@ public void testGetStoreAndUserPartitionsMapping() {
...
+    ZNRecord record = new ZNRecord("testId");
+    Map<String, Map<String, String>> mapFields = new HashMap<>();
+    Map<String, String> testPartitionZero = new HashMap<>();
+    Map<String, String> testPartitionOne = new HashMap<>();
+    testPartitionZero.put("lor1-app56585.prod.linkedin.com_1690", "LEADER");
+    testPartitionZero.put("lor1-app56614.prod.linkedin.com_1690", "STANDBY");
+    testPartitionZero.put("lor1-app110448.prod.linkedin.com_1690", "STANDBY");
+    testPartitionOne.put("lor1-app56586.prod.linkedin.com_1690", "LEADER");
+    testPartitionOne.put("lor1-app71895.prod.linkedin.com_1690", "STANDBY");
+    testPartitionOne.put("lor1-app111181.prod.linkedin.com_1690", "STANDBY");
```
kvargha, line 180: Instead of using actual LinkedIn host names, can you create general ones?

clients/da-vinci-client/src/test/java/com/linkedin/davinci/storage/StorageServiceTest.java
```diff
@@ -121,4 +136,76 @@ public void testGetStoreAndUserPartitionsMapping() {
...
+    mapFields.put("test_store_v1_0", testPartitionZero);
+    mapFields.put("test_store_v1_1", testPartitionOne);
+    record.setMapFields(mapFields);
+    when(idealState.getRecord()).thenReturn(record);
+    when(manager.getInstanceName()).thenReturn("lor1-app56586.prod.linkedin.com_1690");
```
kvargha, line 185: Same here

clients/da-vinci-client/src/test/java/com/linkedin/davinci/storage/StorageServiceTest.java
```diff
@@ -121,4 +136,76 @@ public void testGetStoreAndUserPartitionsMapping() {
...
+    storageEngineRepositoryField.set(mockStorageService, mockStorageEngineRepository);
+    when(mockStorageService.getStorageEngineRepository()).thenReturn(mockStorageEngineRepository);
+    when(mockStorageService.getStorageEngineRepository().getAllLocalStorageEngines()).thenReturn(localStorageEngines);
+    Field configLoaderField = StorageService.class.getDeclaredField("configLoader");
+    configLoaderField.setAccessible(true);
+    configLoaderField.set(mockStorageService, mockVeniceConfigLoader);
+
+    doCallRealMethod().when(mockStorageService).checkWhetherStoragePartitionsShouldBeKeptOrNot(manager);
+    mockStorageService.checkWhetherStoragePartitionsShouldBeKeptOrNot(manager);
+    Assert.assertFalse(abstractStorageEngine.containsPartition(0));
```
kvargha, line 209: Can you add this check:
`verify(abstractStorageEngine.dropPartition(0))`

clients/da-vinci-client/src/test/java/com/linkedin/davinci/storage/StorageServiceTest.java
kristyelee: Updated

clients/da-vinci-client/src/main/java/com/linkedin/davinci/storage/StorageService.java
```diff
@@ -371,6 +376,42 @@ public synchronized AbstractStorageEngine openStore(
...
+
+      if (idealState != null) {
+        Set<Integer> idealStatePartitionIds = new HashSet<>();
+        Set<String> partitionSet = idealState.getPartitionSet();
+        partitionSet.stream().forEach(partitionDbName -> {
+          idealStatePartitionIds.add(RocksDBUtils.parsePartitionIdFromPartitionDbName(partitionDbName));
+        });
+
+        Map<String, Map<String, String>> mapFields = idealState.getRecord().getMapFields();
+        for (Map.Entry<String, Map<String, String>> entry: mapFields.entrySet()) {
```
kvargha, line 401: Instead of looping over every partition in `idealState`, can we instead loop over the partitions we have locally and validate it against the `mapFields`?

Since the partition Ids we have locally are ints, we have to convert them to the format that `mapFields` has them (Test_Store_Migration_Demo_v1_0).
I think this would be the format: `storeName + localPartitionId`

clients/da-vinci-client/src/main/java/com/linkedin/davinci/storage/StorageService.java
kristyelee: I have written the comparison and retrieved the current host from `manager.getInstanceName().`

clients/da-vinci-client/src/main/java/com/linkedin/davinci/storage/StorageService.java
kristyelee: (Just to confirm, this is completed).

clients/da-vinci-client/src/main/java/com/linkedin/davinci/storage/StorageService.java
```diff
@@ -371,6 +375,31 @@ public synchronized AbstractStorageEngine openStore(
     return engine;
   }
 
+  public synchronized void checkWhetherStoragePartitionsShouldBeKeptOrNot(SafeHelixManager manager) {
+    if (manager == null) {
+      return;
+    }
+    for (AbstractStorageEngine storageEngine: getStorageEngineRepository().getAllLocalStorageEngines()) {
+      String storeName = storageEngine.getStoreVersionName();
+      Set<Integer> storageEnginePartitionIds = new HashSet<>(storageEngine.getPartitionIds());
```
kvargha, line 384: Since you're just looping over this, not need to convert it to a set. You can keep it as a list.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/storage/StorageService.java
```diff
@@ -371,6 +375,31 @@ public synchronized AbstractStorageEngine openStore(
...
+      PropertyKey.Builder propertyKeyBuilder =
+          new PropertyKey.Builder(configLoader.getVeniceClusterConfig().getClusterName());
+      SafeHelixDataAccessor helixDataAccessor = manager.getHelixDataAccessor();
+      IdealState idealState = helixDataAccessor.getProperty(propertyKeyBuilder.idealStates(storeName));
+
+      if (idealState != null) {
+        Map<String, Map<String, String>> mapFields = idealState.getRecord().getMapFields();
+        for (Integer partitionId: storageEnginePartitionIds) {
+          String partitionDbName = storeName + "_" + partitionId;
+          if (!mapFields.get(partitionDbName).containsKey(instanceHostName)) {
```
kvargha, line 395: This check assumes that `partitionDbName` is in `mapFields`. To be on the safe side can you add a safeguard here that if it isn't in the `mapFields` delete it too.

kristyelee: I have added this.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/storage/StorageService.java
```diff
@@ -371,6 +375,33 @@ public synchronized AbstractStorageEngine openStore(
...
+      IdealState idealState = helixDataAccessor.getProperty(propertyKeyBuilder.idealStates(storeName));
+
+      if (idealState != null) {
+        Map<String, Map<String, String>> mapFields = idealState.getRecord().getMapFields();
+        for (Integer partitionId: storageEnginePartitionIds) {
+          String partitionDbName = storeName + "_" + partitionId;
+          if (!mapFields.containsKey(partitionDbName)) {
+            storageEngine.dropPartition(partitionId);
+          } else if (!mapFields.get(partitionDbName).containsKey(instanceHostName)) {
+            storageEngine.dropPartition(partitionId);
```
kvargha, line 398: Instead of separating this into two if statements, can you combine the into with with an or?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/storage/StorageService.java
kristyelee: [I have mutated the set, so I have created a copy to iterate over.]

clients/da-vinci-client/src/main/java/com/linkedin/davinci/storage/StorageService.java
kristyelee: The code logic here to consult the ideal state's information of partitions and the hosts they are assigned to [in the list discussed] has been added, and comparison to current host to determine how to keep the partition has been integrated.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/storage/StorageService.java
```diff
@@ -371,6 +375,35 @@ public synchronized AbstractStorageEngine openStore(
...
+
+      if (idealState != null) {
+        Map<String, Map<String, String>> mapFields = idealState.getRecord().getMapFields();
+        for (Integer partitionId: storageEnginePartitionIds) {
+          if (storageEngine.isMetadataPartition(partitionId)) {
+            continue;
+          }
+          String partitionDbName = storeName + "_" + partitionId;
+          if (!mapFields.containsKey(partitionDbName) || (!mapFields.get(partitionDbName).containsKey(instanceHostName)
+              && !mapFields.get(partitionDbName).isEmpty())) {
```
kvargha, line 399: Although `!mapFields.get(partitionDbName).isEmpty()` may make the tests pass, this isn't exactly desirable behavior.

I think your existing logic is correct, but there is something wrong with the tests failing. The instance should be assigned to the partition, but it's not. I will try to back to this later this week.

kristyelee: For one case, ensuring the list of hosts is not empty and the current host is in the list of hosts for the partition is one case in which we delete the partition.

kvargha: It's possible for the `mapFields` to be empty for a `partitionDbName`, if any hosts haven't been assigned to it yet. However, in the tests that are failing, I can see that the host is being assigned in the ExternalView but not in the IdealState. So there's a problem with Helix not being updated in-time when we're trying to perform the cleanup.

kristyelee: Right, I understand this possibility.  
I initially thought of one possibility that coincided with the check of current host being assigned that may have made sense.

kristyelee: Commit [5e43f6f](https://github.com/linkedin/venice/commit/5e43f6f3671598deefbc6c4c3f8dea10ab0e9190) covers the change that causes the tests to pass (a check that initially could coincide). 
I have reverted the change.

kristyelee:
The check over which storage partitions to persist is moved to the shutdown phase of VeniceServer which was discussed. 

Shutdown phase contains call to `storageService.checkWhetherStoragePartitionsShouldBeKeptOrNot`. Services are also stopped when VeniceServer shuts down.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/storage/StorageService.java
```diff
@@ -371,6 +375,35 @@ public synchronized AbstractStorageEngine openStore(
...
+          new PropertyKey.Builder(configLoader.getVeniceClusterConfig().getClusterName());
+      SafeHelixDataAccessor helixDataAccessor = manager.getHelixDataAccessor();
+      IdealState idealState = helixDataAccessor.getProperty(propertyKeyBuilder.idealStates(storeName));
+
+      if (idealState != null) {
+        Map<String, Map<String, String>> mapFields = idealState.getRecord().getMapFields();
+        for (Integer partitionId: storageEnginePartitionIds) {
+          if (storageEngine.isMetadataPartition(partitionId)) {
+            continue;
+          }
```
kvargha, line 396: Looking through the code, `METADATA_PARTITION_ID` doesn't get added to `partitionList`, so it's safe to remove this.

services/venice-server/src/main/java/com/linkedin/venice/server/VeniceServer.java
```diff
@@ -625,6 +625,8 @@ public void shutdown() throws VeniceException {
         }
       }
 
+      storageService.checkWhetherStoragePartitionsShouldBeKeptOrNot(helixParticipationService.getHelixManager());
+
```
kvargha, line 629: During the meeting, Zac mentioned that we should move this back to startup.

Please change it back to how it was previously. My bad.
`managerFuture.thenApply(manager -> {
      storageEngine.checkWhetherStoragePartitionsShouldBeKeptOrNot(manager);
      return true;
    });`

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/server/VeniceServerTest.java
```diff
@@ -182,6 +183,33 @@ public void testCheckBeforeJointClusterBeforeHelixInitializingCluster() throws E
...
+      // Create a storage engine.
+      String storeName = Version.composeKafkaTopic(cluster.createStore(1), 1);
+      storageService.openStoreForNewPartition(
+          server.getVeniceServer().getConfigLoader().getStoreConfig(storeName),
+          1,
+          () -> null);
+      Assert.assertEquals(
+          repository.getAllLocalStorageEngines().size(),
+          1,
+          "We have created one storage engine for store: " + storeName);
```
kvargha, line 204: You shouldn't have to do this. `cluster.createStore` should be doing this under the hood eventually.

services/venice-server/src/main/java/com/linkedin/venice/server/VeniceServer.java
kristyelee: This has been updated accordingly.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/server/VeniceServerTest.java
```diff
@@ -182,6 +183,37 @@ public void testCheckBeforeJointClusterBeforeHelixInitializingCluster() throws E
...
+      Assert.assertTrue(server.getVeniceServer().getHelixParticipationService().isRunning());
+      Assert.assertEquals(storageService.getStorageEngine(storeName).getPartitionIds().size(), 3);
+
+      server.getVeniceServer().shutdown();
+      cluster.getControllerClient().deleteStore(storeName);
+
+      cluster.stopVeniceServer(server.getPort());
+      cluster.restartVeniceServer(server.getPort());
+      repository = server.getVeniceServer().getStorageService().getStorageEngineRepository();
+      Assert.assertEquals(repository.getAllLocalStorageEngines().size(), 1);
```
kvargha, line 212: If we've dropped all partitions, I think we should delete the entire store.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/server/VeniceServerTest.java
```diff
@@ -182,6 +183,37 @@ public void testCheckBeforeJointClusterBeforeHelixInitializingCluster() throws E
...
+      // Create a storage engine.
+      String storeName = Version.composeKafkaTopic(cluster.createStore(1), 1);
+      Assert.assertEquals(repository.getAllLocalStorageEngines().size(), 1);
+      Assert.assertTrue(server.getVeniceServer().getHelixParticipationService().isRunning());
+      Assert.assertEquals(storageService.getStorageEngine(storeName).getPartitionIds().size(), 3);
+
+      server.getVeniceServer().shutdown();
+      cluster.getControllerClient().deleteStore(storeName);
+
+      cluster.stopVeniceServer(server.getPort());
```
kvargha, line 209: Is this needed since you're already doing the shutdown on line 206?

kristyelee: Yes, the storage service cannot be obtained if I don't use `cluster.stopVeniceServer` and `cluster.restartVeniceServer.`

kvargha: I'm not sure I understand. 

Shutdown is performed on line 206, and on line 209 stop is called on the server.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/server/VeniceServerTest.java
```diff
@@ -182,6 +183,38 @@ public void testCheckBeforeJointClusterBeforeHelixInitializingCluster() throws E
...
+
+      server.getVeniceServer().shutdown();
+
+      cluster.stopVeniceServer(server.getPort());
+      cluster.restartVeniceServer(server.getPort());
+      repository = server.getVeniceServer().getStorageService().getStorageEngineRepository();
+      Assert.assertEquals(repository.getAllLocalStorageEngines().size(), 1);
+      Assert.assertEquals(storageService.getStorageEngine(storeName).getPartitionIds().size(), 0);
+
+      cluster.getControllerClient().deleteStore(storeName);
```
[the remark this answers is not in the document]
kristyelee: Removing `cluster.stopVeniceServer(server.getPort())` results in an error of not being able to restart the server; removing both `cluster.stopVeniceServer(server.getPort())` and `cluster.restartVeniceServer(server.getPort())` results in not being able to fetch the storage service.

Using `cluster.stopVeniceServer` and `cluster.restartVeniceServer` without the shutdown command works.

Should I just remove `server.getVeniceServer().shutdown()` instead? I am trying this now and running the test.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/server/VeniceServerTest.java
kvargha: Re-opening this one.

In our case, we're deleting the whole store which causes us to delete all of the partitions. So the `storageEngine` exists with 0 partitions. If this happens, we should drop the `storageEngine` too.

So `repository.getAllLocalStorageEngines().size()` should return 0.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/server/VeniceServerTest.java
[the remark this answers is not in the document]
kvargha: Yes, I think removing `server.getVeniceServer().shutdown()` is the way to go.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/server/VeniceServerTest.java
kvargha: Here's an [example](https://github.com/linkedin/venice/blob/78de20c7014f8dc1b86e3e1bb0d165fcb0e06911/clients/da-vinci-client/src/main/java/com/linkedin/davinci/storage/StorageService.java#L391-L409). If `remainingPartitions` is empty, it calls `removeStorageEngine`.

To prevent duplicate logic you could just call `dropStorePartition` and set `removeEmptyStorageEngine` to `true`.

kristyelee: The storage engine has been dropped upon removal of partitions.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/server/VeniceServerTest.java
```diff
@@ -209,6 +209,9 @@ public void testStartServerAndShutdownWithPartitionAssignmentVerification() {
       repository = server.getVeniceServer().getStorageService().getStorageEngineRepository();
       Assert.assertEquals(repository.getAllLocalStorageEngines().size(), 1);
       Assert.assertEquals(storageService.getStorageEngine(storeName).getPartitionIds().size(), 0);
+      server.getVeniceServer().getStorageService().removeStorageEngine(storeName);
```
kvargha, line 212: This logic should be done in `checkWhetherStoragePartitionsShouldBeKeptOrNot`. 

If we've deleted all partitions for a `storageEngine`, we should call `removeStorageEngine`.

kvargha: Since `removeStorageEngine`, is now called in `checkWhetherStoragePartitionsShouldBeKeptOrNot`, you can remove line 212. Once that's in I'll kick it off to Zac.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/storage/StorageService.java
```diff
@@ -371,6 +375,35 @@ public synchronized AbstractStorageEngine openStore(
...
+          if (!mapFields.containsKey(partitionDbName)
+              || !mapFields.get(partitionDbName).containsKey(instanceHostName)) {
+            storageEngine.dropPartition(partitionId);
+          }
+        }
+        if (storageEngine.getPartitionIds().isEmpty()) {
+          removeStorageEngine(storeName);
+        }
+      }
+    }
```
kvargha, line 404: `idealState` can be null if the store doesn't exist anymore. I'll leave it up to @ZacAttack whether or not we should remove the `storageEngine` in the else conditional.

ZacAttack: We should do that a it indicates a store deletion.  In which case we should clean up the disk of all partitions associated to that ideal state.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/storage/StorageService.java
```diff
@@ -371,6 +375,35 @@ public synchronized AbstractStorageEngine openStore(
     return engine;
   }
 
+  public synchronized void checkWhetherStoragePartitionsShouldBeKeptOrNot(SafeHelixManager manager) {
+    if (manager == null) {
+      return;
+    }
+    for (AbstractStorageEngine storageEngine: getStorageEngineRepository().getAllLocalStorageEngines()) {
+      String storeName = storageEngine.getStoreVersionName();
+      Set<Integer> storageEnginePartitionIds = new HashSet<>(storageEngine.getPartitionIds());
```
ZacAttack, line 384: There is one edge case that we need to be aware of here.  Venice internally assigns a metadata partition to the storage engine.  This metadata partition exists on every Venice replica.  This metadata partition will not be part of the ideal state.  If you look at AbstractStorageEngine.METADATA_PARTITION_ID you can use this to screen dropping that particular partition (as it won't be in the ideal state).

clients/da-vinci-client/src/main/java/com/linkedin/davinci/storage/StorageService.java
```diff
@@ -371,6 +375,38 @@ public synchronized AbstractStorageEngine openStore(
...
+      PropertyKey.Builder propertyKeyBuilder =
+          new PropertyKey.Builder(configLoader.getVeniceClusterConfig().getClusterName());
+      SafeHelixDataAccessor helixDataAccessor = manager.getHelixDataAccessor();
+      IdealState idealState = helixDataAccessor.getProperty(propertyKeyBuilder.idealStates(storeName));
+
+      if (idealState != null) {
+        Map<String, Map<String, String>> mapFields = idealState.getRecord().getMapFields();
+        for (Integer partitionId: storageEnginePartitionIds) {
+          if (storageEngine.isMetadataPartition(partitionId)) {
+            storageEngine.dropPartition(partitionId);
```
kvargha, line 395: We should skip over deleting the metaDataPartition.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/server/VeniceServerTest.java
```diff
@@ -205,10 +205,6 @@ public void testStartServerAndShutdownWithPartitionAssignmentVerification() {
 
       cluster.stopVeniceServer(server.getPort());
 
-      // Create new servers so partition assignment is removed for the offline participant
-      cluster.addVeniceServer(featureProperties, new Properties());
-      cluster.addVeniceServer(featureProperties, new Properties());
-
```
kvargha: These lines are needed, otherwise the partitions will still be assigned to the original server.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/server/VeniceServerTest.java
```diff
@@ -182,6 +179,42 @@ public void testCheckBeforeJointClusterBeforeHelixInitializingCluster() throws E
...
+      cluster.addVeniceServer(featureProperties, new Properties());
+      cluster.addVeniceServer(featureProperties, new Properties());
+
+      cluster.restartVeniceServer(server.getPort());
+      repository = server.getVeniceServer().getStorageService().getStorageEngineRepository();
+      StorageEngineRepository storageEngineRepository = repository;
+      TestUtils.waitForNonDeterministicAssertion(
+          3,
+          TimeUnit.SECONDS,
+          () -> Assert.assertEquals(storageEngineRepository.getAllLocalStorageEngines().size(), 0));
```
kvargha, line 214: I think this whole section needs to be wrapped in `TestUtils.waitForNonDeterministicAssertion`. I'll let @ZacAttack confirm.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/storage/StorageService.java
```diff
@@ -371,6 +375,40 @@ public synchronized AbstractStorageEngine openStore(
...
+          }
+        }
+        if (storageEngine.getPartitionIds().isEmpty()) {
+          removeStorageEngine(storeName);
+        }
+      } else {
+        removeStorageEngine(storeName);
+      }
+    }
```
kvargha, line 409: Can we please add some logging here to indicate that we're deleting partitions, storage engines? It would be helpful for debugging in the case we observe unnecessary boostrapping.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/server/VeniceServerTest.java
ZacAttack: Nah this is actually ok.  We're basically just doing the restart and waiting for what we need to happen.  In which case we need to only wait on the assert.

ZacAttack (approved):
Looks good!  Great work, thank you for all the effort!

kvargha (approved):

kvargha:
Congratulations on merging your first PR into Venice @kristyelee. Thank you for your persistence.

services/venice-server/src/main/java/com/linkedin/venice/server/VeniceServer.java
```diff
@@ -355,6 +355,11 @@ private List<AbstractVeniceService> createServices() {
       return helixData;
     });
 
+    managerFuture.thenApply(manager -> {
+      storageService.checkWhetherStoragePartitionsShouldBeKeptOrNot(manager);
+      return true;
+    });
+
```
huangminchn, line 362: What if Helix couldn't compute the new Ideal state quick enough after the host is restarted and connected? Server might delete everything and rebootstrap from scratch?

ZacAttack: So a restarted host shouldn't alter the idealstate.  idealstate will only be altered after a rebalance has occurred, which will only happen in cases where the node was down for an extended period.

PR #1197, opened 2024-09-24
[controller] Set rebalancing strategy to Waged instead of Crush when adding a resource to controller cluster

kvargha:

Based on testing using FULL_AUTO, Waged performed better than Crush and CrushEd. So we will be setting this as the default rebalancing strategy.

## How was this PR tested?

CI


nisargthakkar (approved):
LGTM

ZacAttack:
An we have some description on what dimensions 'performed better' was measured?  Like, just faster?  Better rebalance?

kvargha:
It provided a more even rebalancing. I can provide the test results on Slack if you're interested.

PR #1198, opened 2024-09-25
[controller] Add instance stoppable check based on CV states

majisourav99:
## Add instance stoppable check based on CV states 

Implement customized server stoppable check API in controller

The customized stoppable check takes a list of instances, along with a list of previously approved instances and decide which of these instances can be stopped.

For the instances that cannot be stopped, the controller returns an enum that denotes the reason for rejecting the request
     
{WILL_LOSE_DATA, MIN_ACTIVE_REPLICA_VIOLATION, ONGOING_MAINTENANCE, UNKNOWN_INSTANCE}

### DIFF
--- a/clients/venice-admin-tool/src/main/java/com/linkedin/venice/Arg.java
+++ b/clients/venice-admin-tool/src/main/java/com/linkedin/venice/Arg.java
@@ -283,7 +283,7 @@ public enum Arg {
   NEARLINE_PRODUCER_COUNT_PER_WRITER(
       "nearline-producer-count-per-writer", "npcpw", true,
       "How many producers will be used to write nearline workload in Server"
-  );
+  ),
   STORAGE_NODES("storage-nodes", "sns", true, "Input list of storage nodes to check if they can removed or not"),
   INSTANCES("instances", "in", true, "Input list of storage nodes to check if they can removed or not"),
   TO_BE_STOPPED_NODES("to-be-stopped-nodes", "tbsn", true, "List of nodes assumed to be stopped");
--- a/clients/venice-admin-tool/src/main/java/com/linkedin/venice/Arg.java
+++ b/clients/venice-admin-tool/src/main/java/com/linkedin/venice/Arg.java
@@ -284,6 +284,8 @@ public enum Arg {
       "nearline-producer-count-per-writer", "npcpw", true,
       "How many producers will be used to write nearline workload in Server"
   );
+  STORAGE_NODES("storage-nodes", "sns", true, "Input list of storage nodes to check if they can removed or not"),
+  TO_BE_STOPPED_NODES("to-be-stopped-nodes", "tbsn", true, "List of nodes assumed to be stopped");
 
   private final String argName;
   private final String first;
--- a/clients/venice-admin-tool/src/main/java/com/linkedin/venice/Arg.java
+++ b/clients/venice-admin-tool/src/main/java/com/linkedin/venice/Arg.java
@@ -283,8 +283,7 @@ public enum Arg {
   NEARLINE_PRODUCER_COUNT_PER_WRITER(
       "nearline-producer-count-per-writer", "npcpw", true,
       "How many producers will be used to write nearline workload in Server"
-  ),
-  STORAGE_NODES("storage-nodes", "sns", true, "Input list of storage nodes to check if they can removed or not"),
+  ), STORAGE_NODES("storage-nodes", "sns", true, "Input list of storage nodes to check if they can removed or not"),
   INSTANCES("instances", "in", true, "Input list of storage nodes to check if they can removed or not"),
   TO_BE_STOPPED_NODES("to-be-stopped-nodes", "tbsn", true, "List of nodes assumed to be stopped");
 
--- a/clients/venice-admin-tool/src/main/java/com/linkedin/venice/Arg.java
+++ b/clients/venice-admin-tool/src/main/java/com/linkedin/venice/Arg.java
@@ -283,9 +283,8 @@ public enum Arg {
   NEARLINE_PRODUCER_COUNT_PER_WRITER(
       "nearline-producer-count-per-writer", "npcpw", true,
       "How many producers will be used to write nearline workload in Server"
-  ), STORAGE_NODES("storage-nodes", "sns", true, "Input list of storage nodes to check if they can removed or not"),
-  INSTANCES("instances", "in", true, "Input list of storage nodes to check if they can removed or not"),
-  TO_BE_STOPPED_NODES("to-be-stopped-nodes", "tbsn", true, "List of nodes assumed to be stopped");
+  ), INSTANCES("instances", "in", true, "Input list of helix ids of nodes to check if they can removed or not"),
+  TO_BE_STOPPED_NODES("to-be-stopped-nodes", "tbsn", true, "List of helix ids of nodes assumed to be stopped");
 
   private final String argName;
   private final String first;
--- a/clients/venice-admin-tool/src/main/java/com/linkedin/venice/Arg.java
+++ b/clients/venice-admin-tool/src/main/java/com/linkedin/venice/Arg.java
@@ -285,6 +285,7 @@ public enum Arg {
       "How many producers will be used to write nearline workload in Server"
   );
   STORAGE_NODES("storage-nodes", "sns", true, "Input list of storage nodes to check if they can removed or not"),
+  INSTANCES("instances", "in", true, "Input list of storage nodes to check if they can removed or not"),
   TO_BE_STOPPED_NODES("to-be-stopped-nodes", "tbsn", true, "List of nodes assumed to be stopped");
 
   private final String argName;
--- a/internal/venice-common/src/main/java/com/linkedin/venice/controllerapi/ControllerApiConstants.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/controllerapi/ControllerApiConstants.java
@@ -31,8 +31,6 @@ public class ControllerApiConstants {
 
   public static final String INSTANCES = "instances";
   public static final String TO_BE_STOPPED_INSTANCES = "to_be_stopped_instances";
-
-  public static final String LOCKED_STORAGE_NODE_IDS = "locked_storage_node_ids";
   public static final String INSTANCE_VIEW = "instance_view";
   public static final String KEY_SCHEMA = "key_schema";
   public static final String VALUE_SCHEMA = "value_schema";
--- a/internal/venice-common/src/main/java/com/linkedin/venice/controllerapi/ControllerApiConstants.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/controllerapi/ControllerApiConstants.java
@@ -28,9 +28,9 @@ public class ControllerApiConstants {
 
   // We need the following 3 constants for Helix API aggregatedHealthStatus
   public static final String CLUSTER_ID = "cluster_id";
-
   public static final String INSTANCES = "instances";
   public static final String TO_BE_STOPPED_INSTANCES = "to_be_stopped_instances";
+
   public static final String INSTANCE_VIEW = "instance_view";
   public static final String KEY_SCHEMA = "key_schema";
   public static final String VALUE_SCHEMA = "value_schema";
@@ -244,4 +244,6 @@ public class ControllerApiConstants {
 
   public static final String NEARLINE_PRODUCER_COMPRESSION_ENABLED = "nearline_producer_compression_enabled";
   public static final String NEARLINE_PRODUCER_COUNT_PER_WRITER = "nearline_producer_count_per_writer";
+
+  public static final String AGGR_HEALTH_STATUS_URI = "/aggregatedHealthStatus";
 }
--- a/internal/venice-common/src/main/java/com/linkedin/venice/controllerapi/ControllerApiConstants.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/controllerapi/ControllerApiConstants.java
@@ -3,8 +3,6 @@ package com.linkedin.venice.controllerapi;
 public class ControllerApiConstants {
   public static final String HOSTNAME = "hostname";
   public static final String CLUSTER = "cluster_name";
-  public static final String CLUSTER_ID = "cluster_id";
-
   public static final String CLUSTER_DEST = "cluster_name_dest";
   public static final String SOURCE_GRID_FABRIC = "source_grid_fabric";
   public static final String BATCH_JOB_HEARTBEAT_ENABLED = "batch_job_heartbeat_enabled";
@@ -27,6 +25,9 @@ public class ControllerApiConstants {
   public static final String FROZEN = "frozen";
   public static final String ERROR = "error";
   public static final String STORAGE_NODE_ID = "storage_node_id"; /* host_port */
+
+  // We need the following 3 constants for Helix API aggregatedHealthStatus
+  public static final String CLUSTER_ID = "cluster_id";
 
   public static final String INSTANCES = "instances";
   public static final String TO_BE_STOPPED_INSTANCES = "to_be_stopped_instances";
--- a/internal/venice-common/src/main/java/com/linkedin/venice/controllerapi/ControllerApiConstants.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/controllerapi/ControllerApiConstants.java
@@ -25,6 +25,10 @@ public class ControllerApiConstants {
   public static final String FROZEN = "frozen";
   public static final String ERROR = "error";
   public static final String STORAGE_NODE_ID = "storage_node_id"; /* host_port */
+
+  public static final String INSTANCES = "INSTANCES";
+  public static final String TO_BE_STOPPED_INSTANCES = "TO_BE_STOPPED_INSTANCES";
+
   public static final String LOCKED_STORAGE_NODE_IDS = "locked_storage_node_ids";
   public static final String INSTANCE_VIEW = "instance_view";
   public static final String KEY_SCHEMA = "key_schema";
--- a/internal/venice-common/src/main/java/com/linkedin/venice/controllerapi/ControllerApiConstants.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/controllerapi/ControllerApiConstants.java
@@ -95,7 +95,6 @@ public class ControllerApiConstants {
   public static final String BOOTSTRAP_TO_ONLINE_TIMEOUT_IN_HOURS = "bootstrap_to_online_timeout_in_hours";
 
   public static final String INCLUDE_SYSTEM_STORES = "include_system_stores";
-  public static final String AGGR_HEALTH_STATUS_PARAM = "aggr_health_status_param";
 
   public static final String STORE_VIEW = "store_view";
   public static final String STORE_VIEW_NAME = "store_view_name";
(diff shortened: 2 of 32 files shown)


services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -6465,6 +6465,94 @@ public boolean isLeaderControllerFor(String clusterName) {
...
+      return statuses;
+    }
+
+    List<String> allInstances = helixAdminClient.getInstancesInCluster(cluster);
+    Set<Instance> liveInstances = getLiveInstanceMonitor(cluster).getAllLiveInstances();
+
+    HelixCustomizedViewOfflinePushRepository customizedViewRepo = resources.getCustomizedViewRepository();
+    ReadWriteStoreRepository storeRepo = getHelixVeniceClusterResources(cluster).getStoreMetadataRepository();
+
+    for (String instanceId: instances) {
```
nisargthakkar, line 6504: Can we unify the logic b/w this and `node_removable` checks? That one already has the functionality of `to_be_stopped_instances` through `LOCKED_STORAGE_NODE_IDS`

majisourav99: May be not in this PR.  `node_removable` does not use CV and calling it in aggregate over a batch of instances require internal logic change.

nisargthakkar: We can still abstract the main code in one method, and call it from two places. `node_removable` can pass a single-element-list to this new function.

Also, this method can accept a `RoutingDataRepository`, and `stoppable-check` can pass CV repo while `node_removable` can pass EV repo

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -6509,6 +6509,48 @@ public boolean isLeaderControllerFor(String clusterName) {
...
+    // Check for maintenance, for ongoing maintenance none can be removed.
+    HelixVeniceClusterResources resources = getHelixVeniceClusterResources(cluster);
+    MaintenanceSignal maintenanceSignal = HelixUtils.getClusterMaintenanceSignal(cluster, resources.getHelixManager());
+
+    if (maintenanceSignal != null) {
+      for (String instanceId: instances) {
+        nonStoppableInstances.put(instanceId, InstanceRemovableStatuses.NonStoppableReason.ONGOING_MAINTENANCE.name());
+      }
+      return statuses;
+    }
```
nisargthakkar, line 6540: Why do we want to handle MM as a special case?

majisourav99: Its from the doc where it mentions 

{
  "stoppableInstances": ["instance1", "instance2"],
  "nonStoppableInstancesWithReasons": {
    "instance3": "MIN_ACTIVE_REPLICA_VIOLATION",
    "instance4": "ONGOING_OPERATION"
  }
}

services/venice-controller/src/main/java/com/linkedin/venice/controller/server/ControllerRoutes.java
```diff
@@ -179,6 +190,44 @@ private Route updateKafkaTopicConfig(Admin admin, UpdateTopicConfigFunction upda
...
 
+  public Route getAggregatedHealthStatus(Admin admin) {
+    return (request, response) -> {
+      StoppableNodeStatusResponse responseObject = new StoppableNodeStatusResponse();
+      response.type(HttpConstants.JSON);
+
+      try {
+        String cluster = request.queryParams(CLUSTER_ID);
+        String instances = request.queryParams(INSTANCES);
+        String toBeStoppedInstances = request.queryParams(TO_BE_STOPPED_INSTANCES);
```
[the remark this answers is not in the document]
majisourav99: It should, check other post APIs.

services/venice-controller/src/main/java/com/linkedin/venice/controller/InstanceStatusDecider.java
```diff
@@ -63,102 +72,137 @@ static NodeRemovableResult isRemovable(
...
+      List<String> lockedNodes,
+      boolean isInstanceView) {
+    List<NodeRemovableResult> removableResults = new ArrayList<>();
+    Set<String> nonStoppableNodes = new HashSet<>();
+    RoutingDataRepository routingDataRepository = resources.getCustomizedViewRepository();
+    for (String instanceId: instanceIds) {
+      try {
+        // If instance is not alive, it's removable.
+        if (!HelixUtils.isLiveInstance(clusterName, instanceId, resources.getHelixManager())) {
+          nonStoppableNodes.add(instanceId);
```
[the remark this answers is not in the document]
majisourav99: But what is the result we should return to the caller for dead hosts ?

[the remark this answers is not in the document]
nisargthakkar: We should say dead nodes are removable. This is already what is being done in the next line in your code. But adding a node to `nonStoppableNodes` and returning a `NodeRemovableResult.removableResult` is incorrect

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
nisargthakkar: I don't think this was intended for MM. I think we can leave the MM one out, or just check the min active replica. We can skip rebalance. Let's check with Helix team on the behavior during MM

clients/venice-admin-tool/src/main/java/com/linkedin/venice/Arg.java
```diff
@@ -275,7 +275,9 @@ public enum Arg {
   ), RECOVER_CLUSTER("recover-cluster", "rc", true, "Cluster to recover from"),
   BACKUP_FOLDER("backup-folder", "bf", true, "Backup folder path"),
   DEBUG("debug", "d", false, "Print debugging messages for execute-data-recovery"),
-  BLOB_TRANSFER_ENABLED("blob-transfer-enabled", "bt", true, "Flag to indicate if the blob transfer is allowed or not");
+  BLOB_TRANSFER_ENABLED("blob-transfer-enabled", "bt", true, "Flag to indicate if the blob transfer is allowed or not"),
+  INSTANCES("instances", "in", true, "Input list of storage nodes to check if they can removed or not"),
+  TO_BE_STOPPED_NODES("to-be-stopped-nodes", "tbsn", true, "List of nodes assumed to be stopped");
```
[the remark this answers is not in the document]
majisourav99: We could. Today `LOCKED_STORAGE_NODE_IDS` is not used in hooks util, we can replace it by `TO_BE_STOPPED_NODES`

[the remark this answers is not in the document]
nisargthakkar: Sure, let's keep one and remove the other in admin tool

internal/venice-common/src/main/java/com/linkedin/venice/controllerapi/ControllerApiConstants.java
```diff
@@ -25,7 +25,12 @@ public class ControllerApiConstants {
   public static final String FROZEN = "frozen";
   public static final String ERROR = "error";
   public static final String STORAGE_NODE_ID = "storage_node_id"; /* host_port */
-  public static final String LOCKED_STORAGE_NODE_IDS = "locked_storage_node_ids";
+
+  // We need the following 3 constants for Helix API aggregatedHealthStatus
+  public static final String CLUSTER_ID = "cluster_id";
+
+  public static final String INSTANCES = "instances";
+  public static final String TO_BE_STOPPED_INSTANCES = "to_be_stopped_instances";
```
nisargthakkar, line 33: Can we put the blank line after all the 3 for better readability?

services/venice-controller/src/main/java/com/linkedin/venice/controller/InstanceRemovableStatuses.java
```diff
@@ -0,0 +1,39 @@
...
+  private Map<String, String> nonStoppableInstancesWithReasons;
+  private String redirectUrl;
+
+  public void setRedirectUrl(String redirectUrl) {
+    this.redirectUrl = redirectUrl;
+  }
+
+  public String getRedirectUrl() {
+    return redirectUrl;
+  }
```
nisargthakkar, line 18: Let's add `@JsonIgnore` annotation for these two as we don't want them to be a part of the json output

services/venice-controller/src/main/java/com/linkedin/venice/pushmonitor/AbstractPushMonitor.java
```diff
@@ -149,6 +149,25 @@ private void loadAllPushes(List<OfflinePushStatus> offlinePushStatusList) {
...
+           */
+          refreshedOfflinePushStatusList
+              .add(offlinePushAccessor.getOfflinePushStatusAndItsPartitionStatuses(offlinePushStatus.getKafkaTopic()));
+        } catch (Exception e) {
+          LOGGER.error("Could not load offline push for {}", offlinePushStatus.getKafkaTopic(), e);
+        }
+
+      }
+      offlinePushStatusList = refreshedOfflinePushStatusList;
+
```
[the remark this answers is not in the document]
majisourav99: rebase issue, will remove it.

services/venice-controller/src/main/java/com/linkedin/venice/controller/server/NodesAndReplicas.java
```diff
@@ -1,13 +1,6 @@
...
 
-import static com.linkedin.venice.controllerapi.ControllerApiConstants.CLUSTER;
-import static com.linkedin.venice.controllerapi.ControllerApiConstants.ENABLE_DISABLED_REPLICAS;
-import static com.linkedin.venice.controllerapi.ControllerApiConstants.INSTANCE_VIEW;
-import static com.linkedin.venice.controllerapi.ControllerApiConstants.LOCKED_NODE_ID_LIST_SEPARATOR;
-import static com.linkedin.venice.controllerapi.ControllerApiConstants.LOCKED_STORAGE_NODE_IDS;
-import static com.linkedin.venice.controllerapi.ControllerApiConstants.NAME;
-import static com.linkedin.venice.controllerapi.ControllerApiConstants.STORAGE_NODE_ID;
-import static com.linkedin.venice.controllerapi.ControllerApiConstants.VERSION;
+import static com.linkedin.venice.controllerapi.ControllerApiConstants.*;
```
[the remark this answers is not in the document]
majisourav99: damn intellij restarts :(

services/venice-controller/src/main/java/com/linkedin/venice/controller/server/ControllerRoutes.java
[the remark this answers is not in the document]
nisargthakkar: Can you verify this through curl or other HTTP request tooling while integration tests are running? Our implementation seems to be using `x-www-urlencoded` that sparkjava will parse as query params. I didn't find the handling for JSON inputs in Spark java documentation. So, it doesn't look like it'll work by default

services/venice-controller/src/main/java/com/linkedin/venice/controller/server/ControllerRoutes.java
```diff
@@ -179,6 +187,46 @@ private Route updateKafkaTopicConfig(Admin admin, UpdateTopicConfigFunction upda
...
+        if (!StringUtils.isEmpty(toBeStoppedInstances)) {
+          toBeStoppedInstanceList = Utils.parseCommaSeparatedStringToList(toBeStoppedInstances);
+        } else {
+          toBeStoppedInstanceList = new ArrayList<>();
+        }
+        responseObject.setCluster(cluster);
+        InstanceRemovableStatuses statuses =
+            admin.getAggregatedHealthStatus(cluster, instanceList, toBeStoppedInstanceList);
+        if (statuses.getRedirectUrl() != null) {
+          response.redirect(statuses.getRedirectUrl() + "/aggregatedHealthStatus", HttpStatus.SC_MOVED_TEMPORARILY);
```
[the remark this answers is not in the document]
nisargthakkar: In the case of 302, we shouldn't send any content in the response, right?

internal/venice-common/src/main/java/com/linkedin/venice/controllerapi/ControllerClient.java
```diff
@@ -913,6 +915,17 @@ public MultiNodeResponse listStorageNodes() {
     return request(ControllerRoute.LIST_NODES, newParams(), MultiNodeResponse.class);
   }
 
+  public StoppableNodeStatusResponse getAggregatedHealthStatus(
+      String clusterName,
+      List<String> instances,
+      List<String> toBeStoppedInstances) {
+    QueryParams params = newParams().add(CLUSTER_ID, clusterName)
+        .add(INSTANCES, String.join(",", instances))
+        .add(TO_BE_STOPPED_INSTANCES, String.join(",", toBeStoppedInstances));
```
mynameborat, line 924: Should use `putStringSet` instead of joining the collection with `,`.

Also, you'd ideally want to take `Set` as the input parameter to the API in the first place as you don't expect the `instances` to be duplicated nor `to_be_stopped_instance` to be duplicated.

internal/venice-common/src/main/java/com/linkedin/venice/utils/Utils.java
```diff
@@ -467,6 +467,13 @@ public static Set<String> parseCommaSeparatedStringToSet(String rawString) {
     return Utils.setOf(rawString.split(",\\s*"));
   }
 
+  public static List<String> parseCommaSeparatedStringToList(String rawString) {
+    if (StringUtils.isEmpty(rawString)) {
+      return Collections.emptyList();
+    }
+    return Arrays.asList(rawString.split("\\s*,\\s*"));
+  }
+
```
mynameborat, line 476: Using `Set` in the `QueryParams` would get rid of this boiler plate

services/venice-controller/src/main/java/com/linkedin/venice/controller/server/ControllerRoutes.java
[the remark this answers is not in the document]
mynameborat: It does come in as query parameters. Our implementation of executing a post request seems a bit strange with putting the `QueryParams` as part of the url and setting the `byte[]` payload if any as part of the post request body.

I'd suggest give the instances and `not_being_stopped` takes a set that can have multiple items (order of 10s -> potentially 100) as our cluster grows, it would be wise to use this the post payload to pass this information.

Check out other APIs that does this; `sendPushJobDetails` is an example on how this done; You will need to have a model and a serializer ahead and pass the serialized bytes as part of the `request` method

internal/venice-common/src/main/java/com/linkedin/venice/controllerapi/ControllerClient.java
nisargthakkar: `instances` needs to be a list as the decision of which nodes are removable depends on that ordering. I don't have any preference of set vs list for `to_be_stopped_instances`

mynameborat: Thanks for the context on ordering. The motivation was to leverage the existing collection APIs under `QueryParams`. If order does matter, if we decide to go down the `QueryParams`, would be good to add support for `putStringList`.

Although, from my discussion this afternoon, to truly support the POST request (current helix implementation), we may have to use the `byte[]` payload to put in these details which would render the above comment obsolete.

internal/venice-common/src/main/java/com/linkedin/venice/utils/Utils.java
mynameborat: Resolving this comment, as this may not even applicable

services/venice-controller/src/main/java/com/linkedin/venice/controller/server/ControllerRoutes.java
```diff
@@ -179,6 +187,42 @@ private Route updateKafkaTopicConfig(Admin admin, UpdateTopicConfigFunction upda
...
+          return AdminSparkServer.OBJECT_MAPPER.writeValueAsString(responseObject);
+        }
+
+        List<String> toBeStoppedInstanceList = statusRequest.getToBeStoppedInstances();
+        responseObject.setCluster(cluster);
+
+        InstanceRemovableStatuses statuses =
+            admin.getAggregatedHealthStatus(cluster, instanceList, toBeStoppedInstanceList);
+        if (statuses.getRedirectUrl() != null) {
+          response.redirect(statuses.getRedirectUrl() + AGGR_HEALTH_STATUS_URI, HttpStatus.SC_MOVED_TEMPORARILY);
```
nisargthakkar, line 213: I still think we should not return a response in case we're returning a `SC_MOVED_TEMPORARILY` status code

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -6536,6 +6536,37 @@ public boolean isLeaderControllerFor(String clusterName) {
...
+  @Override
+  public InstanceRemovableStatuses getAggregatedHealthStatus(
+      String cluster,
+      List<String> instances,
+      List<String> toBeStoppedInstances) {
+    InstanceRemovableStatuses statuses = new InstanceRemovableStatuses();
+
+    // If current controller is not the leader, redirect with leader URL
+    if (!isLeaderControllerFor(cluster)) {
+      statuses.setRedirectUrl(getLeaderController(cluster).getUrl(false));
```
nisargthakkar, line 6548: Let's return the port that corresponds to the type of call that the controller received - if the call was `https`, we should return `https` response. If the call was `http`, we should return a `http` response

internal/venice-common/src/main/java/com/linkedin/venice/controllerapi/ControllerRoute.java
```diff
@@ -134,6 +135,9 @@ public enum ControllerRoute {
   ROLLBACK_TO_BACKUP_VERSION(
       "/rollback_to_backup_version", HttpMethod.POST, Collections.singletonList(NAME), REGIONS_FILTER
   ),
+  AGGREGATED_HEALTH_STATUS(
+      "/aggregatedHealthStatus", HttpMethod.POST, Collections.emptyList(), AGGR_HEALTH_STATUS_PARAM
+  ),
```
nisargthakkar, line 140: Can we also and one integration test that doesn't use `ControllerClient`, but rather some basic HTTP call?

internal/venice-common/src/main/java/com/linkedin/venice/controllerapi/ControllerClient.java
```diff
@@ -913,6 +917,24 @@ public MultiNodeResponse listStorageNodes() {
...
+  public StoppableNodeStatusResponse getAggregatedHealthStatus(
+      String clusterName,
+      List<String> instances,
+      List<String> toBeStoppedInstances) throws JsonProcessingException {
+
+    AggregatedHealthStatusRequest request =
+        new AggregatedHealthStatusRequest(instances, toBeStoppedInstances, clusterName);
+    request.setToBeStoppedInstances(toBeStoppedInstances);
+    request.setInstances(instances);
+    request.setClusterId(clusterName);
```
nisargthakkar, line 929: We can remove these 3 lines now

nisargthakkar (commented):
Mostly LGTM. One last nit

nisargthakkar (approved):
LGTM. Thanks @majisourav99 !

PR #1199, opened 2024-09-26
[controller] Manual store deletion only can be done by specific user group. 

haoxu07:

This Code change adds a feature allowing manual store deletion from admin tool to a specific user group only. The newly added interface of `isAllowlistUsersForStoreDeletion` for `AccessController` will be used to check if user belong to that specific user group.  

### DIFF
--- a/internal/venice-common/src/main/java/com/linkedin/venice/acl/AccessController.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/acl/AccessController.java
@@ -49,6 +49,16 @@ public interface AccessController {
   boolean isAllowlistUsers(X509Certificate clientCert, String resource, String method);
 
   /**
+   * Check whether the client is the allowlist admin users for store deletion.
+   *
+   * @param clientCert the X509Certificate submitted by client
+   * @param resource the resource being requested;
+   * @param method the operation (GET, POST, ...) to perform against the resource
+   * @return true if the client is admin
+   */
+  boolean isAllowlistUsersForStoreDeletion(X509Certificate clientCert, String resource, String method);
+
+  /**
    * Get principal Id from client certificate.
    * @param clientCert the X509Certificate submitted by client
    * @return principal Id. (headless account name, service name, LDAP id or group id)
--- a/services/venice-controller/src/main/java/com/linkedin/venice/controller/server/AbstractRoute.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/controller/server/AbstractRoute.java
@@ -151,6 +151,23 @@ public class AbstractRoute {
   }
 
   /**
+   * Check whether the user is within the admin users allowlist.
+   */
+  protected boolean isAllowListUserForStoreDeletion(Request request) {
+    if (!isAclEnabled()) {
+      /**
+       * Grant access if it's not required to check ACL.
+       * {@link accessController} will be empty if ACL is not enabled.
+       */
+      return true;
+    }
+    X509Certificate certificate = getCertificate(request);
+
+    String storeName = request.queryParamOrDefault(NAME, STORE_UNKNOWN);
+    return accessController.get().isAllowlistUsersForStoreDeletion(certificate, storeName, HTTP_GET);
+  }
+
+  /**
    * @return whether SSL is enabled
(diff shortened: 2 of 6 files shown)


internal/venice-common/src/main/java/com/linkedin/venice/acl/AccessController.java
```diff
@@ -48,6 +48,16 @@ public interface AccessController {
...
 
+  /**
+   * Check whether the client is the allowlist admin users for store deletion.
+   *
+   * @param clientCert the X509Certificate submitted by client
+   * @param resource the resource being requested;
+   * @param method the operation (GET, POST, ...) to perform against the resource
+   * @return true if the client is admin
+   */
+  boolean isAllowlistUsersForStoreDeletion(X509Certificate clientCert, String resource, String method);
```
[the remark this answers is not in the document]
haoxu07: Sure, we could do that, but wondering why this naming is better?

services/venice-controller/src/main/java/com/linkedin/venice/controller/server/AbstractRoute.java
```diff
@@ -150,6 +150,23 @@ protected boolean isAllowListUser(Request request) {
...
+  /**
+   * Check whether the user is within the admin users allowlist for store deletion.
+   */
+  protected boolean isAllowListUserForStoreDeletion(Request request) {
+    if (!isAclEnabled()) {
+      /**
+       * Grant access if it's not required to check ACL.
+       * {@link accessController} will be empty if ACL is not enabled.
+       */
+      return true;
```
huangminchn, line 162: We want to do the opposite thing, right? If ACL is not enabled, do not allow the store deletion at all.

Besides, it's cleaner to check whether ACL is enabled in AdminSparkServer - do not register this API if ACL is not enabled:
`
    if (sslEnabled) {
      httpService.post(DELETE_STORE.getPath(),
          new VeniceParentControllerRegionStateHandler(admin, storesRoutes.deleteStore(admin)));
    }
`
The spike will be fixing the test cases; let's see how difficult that is. WDYT?

services/venice-controller/src/main/java/com/linkedin/venice/controller/server/StoresRoutes.java
```diff
@@ -506,6 +506,12 @@ public Route deleteStore(Admin admin) {
     return new VeniceRouteHandler<TrackableControllerResponse>(TrackableControllerResponse.class) {
       @Override
       public void internalHandle(Request request, TrackableControllerResponse veniceResponse) {
+        // This is to limit the manual store deletion from admin tool without https to a specific allow list users.
+        if (!isSslEnabled()
+            && !checkIsAllowListUser(request, veniceResponse, () -> isAllowListUserForStoreDeletion(request))) {
+          return;
```
huangminchn, line 512: We should fail the request and send bad request error code to user.

huangminchn (changes requested):

services/venice-controller/src/main/java/com/linkedin/venice/controller/server/StoresRoutes.java
haoxu07: The `checkedIsAllowListUser` will mark the request as Bad Request: 
```
  protected boolean checkIsAllowListUser(
      Request request,
      ControllerResponse veniceResponse,
      BooleanSupplier isAllowListUser) {
    if (!isAllowListUser.getAsBoolean()) {
      veniceResponse.setError(ACL_CHECK_FAILURE_WARN_MESSAGE_PREFIX + request.url());
      veniceResponse.setErrorType(ErrorType.BAD_REQUEST);
      return false;
    }
    return true;
  }
```

internal/venice-common/src/main/java/com/linkedin/venice/acl/AccessController.java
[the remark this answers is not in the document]
nisargthakkar: It feels more concise to me, while conveying the same information

haoxu07:
Temporarily closed this, as we decide not to pursure the permission check.

PR #1201, opened 2024-09-26
Revert "[common][server][dvc-client] Make a true non-blocking end-offset fetcher

majisourav99:
This reverts commit 262101c6968abbd47b5d2ea28e4eb9e0c4784778.


FelixGV (approved):
Thanks. Task 1000 passed in 8m, no internal retries needed.

majisourav99:
> Thanks. Task 1000 passed in 8m, no internal retries needed.

At least now we know this did cause flakiness. Need to find the RCA now.

PR #1202, opened 2024-09-26
[controller] Ensure empty string from config doesn't result in an array of an empty string for instance tag list

kvargha:

Currently, if you provide an empty string for `controller.instance.tag.list`, it will result in an array that has an empty string as an element. This isn't desirable, because the code would then assign an empty string to an instance's tag list in the controller.

This change will make sure that if there is an empty string, it will result in an empty array.

## How was this PR tested?

Added a new test case

### DIFF
--- a/internal/venice-client-common/src/main/java/com/linkedin/venice/utils/VeniceProperties.java
+++ b/internal/venice-client-common/src/main/java/com/linkedin/venice/utils/VeniceProperties.java
@@ -389,9 +389,6 @@ public class VeniceProperties implements Serializable {
     }
 
     String value = get(key);
-    if (value.isEmpty()) {
-      return defaultValue;
-    }
 
     String[] pieces = value.split("\\s*,\\s*");
     return Arrays.asList(pieces);
@@ -406,10 +403,6 @@ public class VeniceProperties implements Serializable {
       value = get(preferredKey);
     } else if (containsKey(altKey)) {
       value = get(altKey);
-    }
-
-    if (value.isEmpty()) {
-      return defaultValue;
     }
 
     String[] pieces = value.split("\\s*,\\s*");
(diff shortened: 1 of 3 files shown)


internal/venice-client-common/src/main/java/com/linkedin/venice/utils/VeniceProperties.java
```diff
@@ -389,6 +389,10 @@ public List<String> getList(String key, List<String> defaultValue) {
     }
 
     String value = get(key);
+    if (value.isEmpty()) {
+      return defaultValue;
+    }
+
```
nisargthakkar, line 395: I'm not sure this would be safe since this is used widely. We need to be careful and audit all usages of these methods

nisargthakkar:
Also, this specific change affects almost every component, and not just `controller`. Can we handle this in `VeniceControllerClusterConfig` instead?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerClusterConfig.java
```diff
@@ -654,7 +654,9 @@ public VeniceControllerClusterConfig(VeniceProperties props) {
     this.adminCheckReadMethodForKafka = props.getBoolean(ADMIN_CHECK_READ_METHOD_FOR_KAFKA, true);
     this.controllerClusterName = props.getString(CONTROLLER_CLUSTER, "venice-controllers");
     this.controllerResourceInstanceGroupTag = props.getString(CONTROLLER_RESOURCE_INSTANCE_GROUP_TAG, "");
-    this.controllerInstanceTagList = props.getList(CONTROLLER_INSTANCE_TAG_LIST, Collections.emptyList());
+    this.controllerInstanceTagList = props.getString(CONTROLLER_INSTANCE_TAG_LIST, "").isEmpty()
+        ? Collections.emptyList()
+        : props.getList(CONTROLLER_INSTANCE_TAG_LIST, Collections.emptyList());
```
nisargthakkar, line 659: This is a bit hard to read. Can we split it in two lines? One to read the string, other to make it a list?

Also, since we're just using the string to do the emptiness check, let's keep null as the default and do a null check instead

nisargthakkar (approved):

PR #1203, opened 2024-09-27
[da-vinci][controller] Filter out status reporting from the nodes, where the DaVinci app is still bootstrapping

gaojieliu:
## Summary
In Prod, we are seeing the bootstrapping nodes are blocking the new push, and this PR will let Controller ignore the status report from the nodes, which are still bootstrapping.
This feature is especially important for DaVinci apps, which are using multiple large stores and bootstrapping from an empty state can take a long time and it can delay the new push significantly.

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
@@ -603,16 +603,22 @@ public class DaVinciBackend implements Closeable {
       int partition,
       ExecutionStatus status,
       Optional<String> incrementalPushVersion) {
-    if (hasCurrentVersionBootstrapping()) {
+    VersionBackend versionBackend = versionByTopicMap.get(kafkaTopic);
+    if (hasCurrentVersionBootstrapping()
+        /**
+         * It is fine to report the status for Venice current version as the push job for current version is completed already.
+         * Another benefit is to support blob discovery for current version.
+         */
+        && getVeniceCurrentVersionNumber(Version.parseStoreFromKafkaTopicName(kafkaTopic)) != Version
+            .parseVersionFromVersionTopicName(kafkaTopic)) {
       LOGGER.info(
-          "DaVinci is still bootstrapping, so skip the status report for store version:{}, partition: {}, status: {}{}",
+          "DaVinci is still bootstrapping, so skip the status report for store version: {}, partition: {}, status: {}{}",
           kafkaTopic,
           partition,
           status,
           (incrementalPushVersion.isPresent() ? ", inc push version: " + incrementalPushVersion.get() : ""));
       return;
     }
-    VersionBackend versionBackend = versionByTopicMap.get(kafkaTopic);
     if (versionBackend != null && versionBackend.isReportingPushStatus()) {
       Version version = versionBackend.getVersion();
       if (writeBatchingPushStatus && !incrementalPushVersion.isPresent()) {
@@ -664,6 +670,11 @@ public class DaVinciBackend implements Closeable {
     } catch (VeniceNoStoreException e) {
       return null;
     }
+  }
+
+  int getVeniceCurrentVersionNumber(String storeName) {
+    Version currentVersion = getVeniceCurrentVersion(storeName);
+    return currentVersion == null ? -1 : currentVersion.getNumber();
   }
 
   private Version getVeniceLatestNonFaultyVersion(Store store, Set<Integer> faultyVersions) {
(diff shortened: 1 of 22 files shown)


FelixGV (dismissed):
Thanks, makes sense, LGTM.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
```diff
@@ -602,6 +602,13 @@ protected void reportPushStatus(
       int partition,
       ExecutionStatus status,
       Optional<String> incrementalPushVersion) {
+    if (hasCurrentVersionBootstrapping()) {
+      LOGGER.info(
+          "DaVinci is still bootstrapping, so skip the status report for store version: " + kafkaTopic + ", partition: "
+              + partition + ", status: " + status
+              + (incrementalPushVersion.isPresent() ? ", inc push version: " + incrementalPushVersion.get() : ""));
```
FelixGV, line 609: Wait, sorry, I just saw this at the last moment after submitting my previous review... can we use string interpolation here like in the rest of the code, instead of concatenation?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaStoreIngestionService.java
```diff
@@ -557,6 +557,20 @@ private static void shutdownExecutorService(ExecutorService executor, String nam
     }
   }
 
+  public boolean hasCurrentVersionBootstrapping() {
+    return hasCurrentVersionBootstrapping(topicNameToIngestionTaskMap);
+  }
+
+  public static boolean hasCurrentVersionBootstrapping(Map<String, StoreIngestionTask> ingestionTaskMap) {
+    for (Map.Entry<String, StoreIngestionTask> entry: ingestionTaskMap.entrySet()) {
```
sixpluszero, line 565: I have a question/concern: Does this work with ingestion isolation feature? Because essentially ingestion will start on forked process

gaojieliu: In theory, it should.
At the startup time, all the version ingestions will happen in the isolated process including current version until it is finished, and once it is finished and handed over to the main process, it won't matter as it is completed.

sixpluszero: But I am still thinking the case -
Say there is a II DVC, now it is bootstrapping with one current version and a future version. 
Regarding future version, it will still ingest in forked process until RTS is reached. But in the middle, it will report STARTED/END_OF_PUSH_RECEIVED and the above change in the reportPushStatus logic does block it from reporting. So ZK/system store will still register such status. And when controller is polling the job status, it will see it as a not-ready-replica and blocked on it?

sixpluszero: Actually it will not block, but it may potentially lead to too many dead instances because you also delete HB for the host, and fail fast for the job? (but if these bootstrapping nodes are small enough, then we might be good.....)

gaojieliu: delete HB won't result in dead instances and it is used to avoid dead instances..

clients/da-vinci-client/src/main/java/com/linkedin/davinci/VersionBackend.java
```diff
@@ -193,14 +193,30 @@ synchronized void tryStartHeartbeat() {
...
+          sendOutHeartbeat(backend, version);
         } catch (Throwable t) {
           LOGGER.error("Unable to send heartbeat for {}", this);
         }
       }, 0, heartbeatInterval, TimeUnit.SECONDS);
     }
   }
 
+  protected static void sendOutHeartbeat(DaVinciBackend backend, Version version) {
+    if (backend.hasCurrentVersionBootstrapping()) {
```
sixpluszero, line 205: Hmm, maybe I still get it wrong, but this to me still has the same issue towards II logic - because in the beginning, all recovery/bootstrap work will happen in forked process and main process ingestion task map may not even have that SIT. So in that case - it will still send out status and regular HB. Maybe we should sync offline today to make sure we are aligned?

sixpluszero: But overall, I like current approach better than the delete HB one, and maybe we need to add [controller] tag, as it also take some changes there..

sixpluszero (approved):
Thanks for the change and offline discussion!

gaojieliu:
Thanks @sixpluszero for the solid review!

PR #1204, opened 2024-09-27
[server][common] Allow store config repo to fetch configs only when needed

adamxchen:
## Summary

https://github.com/linkedin/venice/pull/1132 introduced additional ZK watch. The implementation of `HelixReadOnlyStoreConfigRepository` reads all store configs, regardless of which cluster, this creates an issue since it would increase the additional ZK watch by `store number x host name`. In this PR, it adds some logics to lazy fetch the store config and cache.

## How was this PR tested?

new unit test


clients/da-vinci-client/src/main/java/com/linkedin/davinci/repository/VeniceMetadataRepositoryBuilder.java
```diff
@@ -147,7 +147,6 @@ private void initServerStoreAndSchemaRepository() {
         adapter,
         clusterConfig.getRefreshAttemptsForZkReconnect(),
         clusterConfig.getRefreshIntervalForZkReconnectInMs());
-    storeConfigRepo.refresh();
```
gaojieliu, line 150: I think we should still call `refresh` to get a list of stores, and please see my following comments.

gaojieliu (commented):
I am thinking the following refactor idea, which I think will be safer:
1. Keep the existing `refresh` calls as today.
2. `HelixReadOnlyStoreConfigRepository#refresh` will place a child-change listener as today, but it will only get a list of store names, not store configs.
3. In the child-change listener, whenever there is a child change, it will only mutate the maintained store list and of coz, for deleted stores, we will remove the corresponding entry from the local store config map and corresponding zk watch.
4. We will only place a zk watch on store config node in `lazy fetch` function.
5. In `HelixReadOnlyStoreConfigRepository#getStoreConfig` function, if the requested store doesn't in the store list, return `Optional.empty()` directly, and if it is present, fetch the store config on the fly and place the zk watch to monitor data change.
6. `HelixReadOnlyStoreConfigRepository` would maintain a `storeConfigMap`, but this map will be lazily populated by `HelixReadOnlyStoreConfigRepository#getStoreConfig` function.

WDYT?

internal/venice-common/src/main/java/com/linkedin/venice/helix/HelixReadOnlyStoreConfigRepository.java
```diff
@@ -115,6 +121,15 @@ public Optional<StoreConfig> getStoreConfig(String storeName) {
...
+      config = accessor.getStoreConfig(veniceStoreName);
+      if (config != null) {
+        storeConfigMap.get().put(config.getStoreName(), config);
+        accessor.subscribeStoreConfigDataChangedListener(veniceStoreName, storeConfigChangedListener);
+      }
+    }
+
     if (config != null) {
       return Optional.of(config.cloneStoreConfig());
     } else {
```
huangminchn, line 135: There is already a branch for config being null. Shall we put the new logic here or remove it?

internal/venice-common/src/main/java/com/linkedin/venice/helix/HelixReadOnlyStoreConfigRepository.java
```diff
@@ -115,6 +121,15 @@ public Optional<StoreConfig> getStoreConfig(String storeName) {
       veniceStoreName = VeniceSystemStoreType.DAVINCI_PUSH_STATUS_STORE.extractRegularStoreName(storeName);
     }
     StoreConfig config = storeConfigMap.get().get(veniceStoreName);
+    // Lazy fetch. It should happen when the entire store config map is not loaded, and only do ad-hoc fetch
+    if (config == null) {
+      config = accessor.getStoreConfig(veniceStoreName);
+      if (config != null) {
+        storeConfigMap.get().put(config.getStoreName(), config);
+        accessor.subscribeStoreConfigDataChangedListener(veniceStoreName, storeConfigChangedListener);
```
huangminchn, line 129: I wonder if we can unsubscribe when the store is deleted from this cluster; is it feasible with the current implementation? Does it know if the store is deleted?

internal/venice-common/src/main/java/com/linkedin/venice/helix/HelixReadOnlyStoreConfigRepository.java
adamxchen: After the new logic, the config can still be null or not-null so this branch still applies. We may use ternary operator to make it look nice tho

internal/venice-common/src/main/java/com/linkedin/venice/helix/HelixReadOnlyStoreConfigRepository.java
adamxchen: IIUC, we will need the `StoreConfigAddedOrDeletedChangedListener` for this and that's only used in `refresh`.. Let me see if I can incorporate Gaojie's suggestion to refactor this into next revision

clients/da-vinci-client/src/main/java/com/linkedin/davinci/repository/VeniceMetadataRepositoryBuilder.java
adamxchen: Thanks for the great suggestions! The core idea is only place watch when there's a need to get store config, and for other stores, we just cache them to know whether they are present or not. Will update it

internal/venice-common/src/main/java/com/linkedin/venice/helix/HelixReadOnlyStoreConfigRepository.java
```diff
@@ -33,73 +34,58 @@
...
-  public HelixReadOnlyStoreConfigRepository(
-      ZkClient zkClient,
-      ZkStoreConfigAccessor accessor,
-      int refreshAttemptsForZkReconnect,
-      long refreshIntervalForZkReconnectInMs) {
+  public HelixReadOnlyStoreConfigRepository(ZkClient zkClient, ZkStoreConfigAccessor accessor) {
     this.zkClient = zkClient;
     this.accessor = accessor;
-    this.storeConfigMap = new AtomicReference<>(new HashMap<>());
+    this.loadedStoreConfigMap = new AtomicReference<>(new HashMap<>());
```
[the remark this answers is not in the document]
gaojieliu: We should use ^ as `getStoreConfig` can update `loadedStoreConfigMap` concurrently when store configs are missing from the map.

gaojieliu (commented):
Left some more comments and you need to fix the test failures...

internal/venice-common/src/main/java/com/linkedin/venice/helix/ZkStoreConfigAccessor.java
```diff
@@ -46,7 +46,7 @@ public ZkStoreConfigAccessor(
   }
 
   public List<String> getAllStores() {
-    return dataAccessor.getChildNames(ROOT_PATH, AccessOption.PERSISTENT);
+    return HelixUtils.listPathContents(dataAccessor, ROOT_PATH);
```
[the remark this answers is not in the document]
adamxchen: It provides a bit better error handling, though essentially they are same.
```
  public static <T> List<String> listPathContents(ZkBaseDataAccessor<T> dataAccessor, String path) {
    try {
      List<String> paths = dataAccessor.getChildNames(path, AccessOption.PERSISTENT);
      return paths == null ? Collections.emptyList() : paths;
    } catch (Exception e) {
      LOGGER.error("Error when listing contents in path: {}", path, e);
      throw e;
    }
  }
```

internal/venice-common/src/main/java/com/linkedin/venice/helix/HelixReadOnlyStoreConfigRepository.java
```diff
@@ -114,12 +100,20 @@ public Optional<StoreConfig> getStoreConfig(String storeName) {
...
+
+    if (availableStoreSet.get().contains(storeName)) {
+      StoreConfig config = loadedStoreConfigMap.get().get(storeName);
+      if (config == null) {
+        // lazy fetch from ZK and attach watch
+        config = accessor.getStoreConfig(veniceStoreName);
+        if (config != null) {
+          loadedStoreConfigMap.get().put(config.getStoreName(), config);
+          accessor.subscribeStoreConfigDataChangedListener(veniceStoreName, storeConfigChangedListener);
+        }
```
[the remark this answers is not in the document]
adamxchen: I'd opt-in the first approach since it avoids unnecessary ZK watch if the config is null.

internal/venice-common/src/main/java/com/linkedin/venice/helix/HelixReadOnlyStoreConfigRepository.java
```diff
@@ -114,12 +100,25 @@ public Optional<StoreConfig> getStoreConfig(String storeName) {
...
+      if (config == null) {
+        // lazy fetch from ZK and attach watch
+        config = accessor.getStoreConfig(veniceStoreName);
+        if (config != null) {
+          loadedStoreConfigMap.get().put(config.getStoreName(), config);
+          accessor.subscribeStoreConfigDataChangedListener(veniceStoreName, storeConfigChangedListener);
+          StoreConfig configToVerify = accessor.getStoreConfig(veniceStoreName);
+          if (!configToVerify.equals(config)) {
+            LOGGER.debug("Store config is changed during fetching. Reload the store config.");
+            config = configToVerify;
```
gaojieliu, line 115: If it is truly different, we need to update the map, right?

internal/venice-common/src/main/java/com/linkedin/venice/helix/HelixReadOnlyStoreConfigRepository.java
```diff
@@ -131,62 +130,70 @@ public StoreConfig getStoreConfigOrThrow(String storeName) {
...
+    public void handleDataChange(String dataPath, Object data) {
       if (!(data instanceof StoreConfig)) {
         throw new VeniceException(
             "Invalid data from zk notification. Required: StoreConfig, but get: " + data.getClass().getName());
       }
       StoreConfig config = (StoreConfig) data;
-      synchronized (storeConfigMap) {
-        Map<String, StoreConfig> map = new HashMap<>(storeConfigMap.get());
+      synchronized (loadedStoreConfigMap) {
+        Map<String, StoreConfig> map = new HashMap<>(loadedStoreConfigMap.get());
```
gaojieliu, line 194: This is not quite right.
If the `loadedStoreConfigMap` is using concurrent hashmap, we don't need this section, instead we can update the map directly, right?

gaojieliu (commented):
Left two more comments.

internal/venice-common/src/main/java/com/linkedin/venice/helix/HelixReadOnlyStoreConfigRepository.java
adamxchen: Ooops, missed that part

internal/venice-common/src/main/java/com/linkedin/venice/helix/HelixReadOnlyStoreConfigRepository.java
adamxchen: Double-checked the codes. There's no need to swap the entire map atomically so I will remove the AtomicReference and just use concurrent hashmap

internal/venice-common/src/main/java/com/linkedin/venice/helix/HelixReadOnlyStoreConfigRepository.java
adamxchen: Thanks for catching it! Also added one unit test.

internal/venice-common/src/main/java/com/linkedin/venice/helix/HelixReadOnlyStoreConfigRepository.java
```diff
@@ -114,12 +99,26 @@ public Optional<StoreConfig> getStoreConfig(String storeName) {
...
-      return Optional.of(config.cloneStoreConfig());
-    } else {
-      return Optional.empty();
+
+    if (availableStoreSet.get().contains(veniceStoreName)) {
+      StoreConfig config = loadedStoreConfigMap.get(veniceStoreName);
+      if (config == null) {
+        // lazy fetch from ZK and attach watch
+        config = accessor.getStoreConfig(veniceStoreName);
+        if (config != null) {
```
gaojieliu, line 108: Minor improvement:
Shall we log a warn message if the store is present in the store list, but the corresponding `storeConfig` is `null`?
normally, it should not happen.

internal/venice-common/src/main/java/com/linkedin/venice/helix/HelixReadOnlyStoreConfigRepository.java
```diff
@@ -131,63 +130,63 @@ public StoreConfig getStoreConfigOrThrow(String storeName) {
...
-        Map<String, StoreConfig> map = new HashMap<>(storeConfigMap.get());
-        List<String> newStores =
-            currentChildren.stream().filter(newStore -> !map.containsKey(newStore)).collect(Collectors.toList());
-
-        Set<String> deletedStores = new HashSet<>(map.keySet());
-        currentChildren.forEach(deletedStores::remove);
+    public void handleChildChange(String parentPath, List<String> currentChildren) {
+      synchronized (availableStoreSet) {
+        Set<String> storeSetSnapshot = new HashSet<>(availableStoreSet.get());
+        List<String> newStores = currentChildren.stream()
```
gaojieliu, line 158: Do we have this variable at all?

gaojieliu (commented):
The code looks good overall now, and I left some minor improvement suggestions.

internal/venice-common/src/main/java/com/linkedin/venice/helix/HelixReadOnlyStoreConfigRepository.java
adamxchen: I think it's a great idea

internal/venice-common/src/main/java/com/linkedin/venice/helix/HelixReadOnlyStoreConfigRepository.java
adamxchen: It's purely for information purpose and has some value. I will make it a `int`

adamxchen: Updated.

adamxchen: @gaojieliu  can you please take another look when you have time?

gaojieliu (approved):
Sorry for the late review, and the change LGTM!

PR #1205, opened 2024-09-28
[vpj][all] Create a new module for offline components to reduce leakage

nisargthakkar:
## Create a new module for offline components to reduce leakage

Recently, we moved some shared classes into `venice-common` to share common functionality. To do this, we added `hadoop-common` library into `venice-common`. However, this leaks `hadoop-common` into online components too which is not desired. Moved these classes into a `venice-hadoop-common` package

## How was this PR tested?

GHCI


FelixGV:
Who needs to depend on this module besides VPJ?

haoxu07 (approved):

nisargthakkar:
> Who needs to depend on this module besides VPJ?

We have an ETL module internally that would benefit from this. This class was originally copied from there into OSS. We plan to release in open-source at some point. It is just a matter of allocating some bandwidth for

FelixGV:
> > Who needs to depend on this module besides VPJ?
> We have an ETL module internally that would benefit from this. This class was originally copied from there into OSS. We plan to release in open-source at some point. It is just a matter of allocating some bandwidth for

Creating a new module just for the sake of this proprietary ETL depender does not seem quite right... can ETL depend on all of VPJ instead?

PR #1206, opened 2024-09-30
[server]  integration test for blob transfer between servers

jingy-li:
## Integration test for blob transfer between servers

- Minor bug fix for ServerBlobFinder. Previously initialize ServerBlobFinder need to wait for customizedViewRepository complete, it waits so long. Moving get customizedViewRepository when we need to find peer not in the constructor. 
- Add integration test to do 
    - Start 2 servers and a VPJ. 
    - Remove one servers metadata partition and data partition 
    - Restart server1. 
    - Verify that the re-started server successfully ingest data from server2 via blob transfer. 

## How was this PR tested?
(1) Integration test: the partition file is transferred and the metadata is transferred and updated. 
(2) Adding logs in updateStorePartitionMetadata to compare the offset record before and after the metadata update. Offset record is successfully transferred and used.  

Screenshot 2024-09-27 at 5 49 12 PM
Screenshot 2024-09-27 at 5 49 46 PM


internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/BlobP2PTransferAmongServersTest.java
```diff
@@ -0,0 +1,163 @@
...
+
+      OffsetRecord offsetServer1 =
+          server1.getVeniceServer().getStorageMetadataService().getLastOffset("test-store_v1", i);
+      OffsetRecord offsetServer2 =
+          server2.getVeniceServer().getStorageMetadataService().getLastOffset("test-store_v1", i);
+      Assert.assertEquals(offsetServer1.getLocalVersionTopicOffset(), offsetServer2.getLocalVersionTopicOffset());
+    }
+  }
+
+  public VeniceClusterWrapper initializeVeniceCluster() {
```
adamxchen, line 84: The practice we usually do it to move codes related to setup env to a method, say `setup`, then add testNG annotation to let it run before test cases so we can potentially add more test cases for this suite. Can you plz do that?

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/BlobP2PTransferAmongServersTest.java
```diff
@@ -0,0 +1,163 @@
...
+    // cleanup and restart server 1
+    FileUtils.deleteDirectory(
+        new File(RocksDBUtils.composePartitionDbDir(path1 + "/rocksdb", storeName + "_v1", METADATA_PARTITION_ID)));
+    FileUtils.deleteDirectory(new File(RocksDBUtils.composePartitionDbDir(path1 + "/rocksdb", storeName + "_v1", 0)));
+    FileUtils.deleteDirectory(new File(RocksDBUtils.composePartitionDbDir(path1 + "/rocksdb", storeName + "_v1", 1)));
+    FileUtils.deleteDirectory(new File(RocksDBUtils.composePartitionDbDir(path1 + "/rocksdb", storeName + "_v1", 2)));
+    cluster.stopAndRestartVeniceServer(server1.getPort());
+
+    // wait 2 min
+    Thread.sleep(120000);
```
adamxchen, line 68: can we use `TestUtils.waitForNonDeterministicAssertion`?

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/BlobP2PTransferAmongServersTest.java
```diff
@@ -0,0 +1,163 @@
...
+
+    // the partition files should be transferred to server 1 and offset should be the same
```
adamxchen, line 70: The concern is we don't know whether the server ingests the data via blob transfer or the kafka.  Yes, the logging can help us here but it's not a programmatic way and the chance is someone breaks the feature in the future and we don't know.

We discussed this in [another thread](https://github.com/linkedin/venice/pull/1118#discussion_r1719109319) a while ago. Do you think we can leverage that to validate?

adamxchen: discussed offline. The IDENTITY file changes so it cannot be used to verify whether partitions really happened or not. There's no reason to block this PR at least and we will come back to the topic later on.

adamxchen (approved):
Thanks!

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/BlobP2PTransferAmongServersTest.java
```diff
@@ -0,0 +1,182 @@
...
+  public void testBlobP2PTransferAmongServers() throws Exception {
+    String storeName = "test-store";
+    Consumer<UpdateStoreQueryParams> paramsConsumer = params -> params.setBlobTransferEnabled(true);
+    setUpStore(cluster, storeName, paramsConsumer, properties -> {}, true);
+
+    VeniceServerWrapper server1 = cluster.getVeniceServers().get(0);
+    VeniceServerWrapper server2 = cluster.getVeniceServers().get(1);
+
+    // very the snapshot is generated for both servers after the job is done
+    for (int i = 0; i < 3; i++) {
```
lluwm, line 64: Nit, would it be more readable if we have a descriptively named constants here?

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/BlobP2PTransferAmongServersTest.java
```diff
@@ -0,0 +1,182 @@
...
+      String snapshotPath2 = RocksDBUtils.composeSnapshotDir(path2 + "/rocksdb", storeName + "_v1", i);
+      Assert.assertTrue(Files.exists(Paths.get(snapshotPath2)));
+    }
+
+    // cleanup and restart server 1
+    FileUtils.deleteDirectory(
+        new File(RocksDBUtils.composePartitionDbDir(path1 + "/rocksdb", storeName + "_v1", METADATA_PARTITION_ID)));
+    FileUtils.deleteDirectory(new File(RocksDBUtils.composePartitionDbDir(path1 + "/rocksdb", storeName + "_v1", 0)));
+    FileUtils.deleteDirectory(new File(RocksDBUtils.composePartitionDbDir(path1 + "/rocksdb", storeName + "_v1", 1)));
+    FileUtils.deleteDirectory(new File(RocksDBUtils.composePartitionDbDir(path1 + "/rocksdb", storeName + "_v1", 2)));
```
lluwm, line 76: Nit, seems we can put it into a small loop for simplicity? Also, can we add a step to verify that files are not existent after the delete?

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/BlobP2PTransferAmongServersTest.java
```diff
@@ -0,0 +1,182 @@
...
+  @Test(singleThreaded = true)
+  public void testBlobP2PTransferAmongServers() throws Exception {
+    String storeName = "test-store";
+    Consumer<UpdateStoreQueryParams> paramsConsumer = params -> params.setBlobTransferEnabled(true);
+    setUpStore(cluster, storeName, paramsConsumer, properties -> {}, true);
+
+    VeniceServerWrapper server1 = cluster.getVeniceServers().get(0);
+    VeniceServerWrapper server2 = cluster.getVeniceServers().get(1);
+
+    // very the snapshot is generated for both servers after the job is done
```
lluwm, line 63: Typo: very-> verify?

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/BlobP2PTransferAmongServersTest.java
```diff
@@ -0,0 +1,182 @@
...
+    for (int i = 0; i < 3; i++) {
+      String snapshotPath1 = RocksDBUtils.composeSnapshotDir(path1 + "/rocksdb", storeName + "_v1", i);
+      Assert.assertTrue(Files.exists(Paths.get(snapshotPath1)));
+      String snapshotPath2 = RocksDBUtils.composeSnapshotDir(path2 + "/rocksdb", storeName + "_v1", i);
+      Assert.assertTrue(Files.exists(Paths.get(snapshotPath2)));
+    }
```
lluwm, line 69: A dumb question, do we remove snapshot path after the blob transfer is done in the design?

adamxchen: I missed this. Yes, @jingy-li please double-check and see if we need to do more clean-up here

jingy-li: In server side, we do not clean the snapshot when the transfer is done.

lluwm (approved):
Thank you for this change @jingy-li, LGTM!

PR #1207, opened 2024-09-30
[server] tweak heartbeat reporting to use completion check

ZacAttack:
## [server] tweak heartbeat reporting to use completion check
EOM

Resolves #XXX


sixpluszero (approved):
Thanks you for the quick change!

PR #1208, opened 2024-09-30
[vpj] Merge venice-hadoop-common back into VPJ

nisargthakkar:
## Merge `venice-hadoop-common` back into VPJ

`venice-hadoop-common` holds some shared classes that we want to use in our internal ETL. But in OSS, it doesn't serve a purpose other than a shared lib for `venice-push-job`. Move these classes back into `venice-push-job`

## How was this PR tested?

GHCI


FelixGV (approved):
Thanks Nisarg.

mynameborat:
I remember there was another util file `SchemaUtils`. What about that one?

mynameborat:
> I remember there was another util file `SchemaUtils`. What about that one?

Sync'd up offline on this. The `SchemaUtils` will remain in `venice-common`. Thank you, @nisargthakkar

nisargthakkar:
> I remember there was another util file `SchemaUtils`. What about that one?

This one is okay to live in `venice-common` since VPJ, DVC and Servers share this

PR #1211, opened 2024-10-03
[controller] Add ability to configure cloud config for venice-controllers and storage clusters, and remove Azure config

kvargha:

1. Removed Azure config, since it's not being used currently.
2. Added ability to set cloud config generically. So, if a user is on any public/private cloud they can set the appropriate configs. This is also necessary to enable controllers to automatically join a cluster.

## How was this PR tested?

Added new tests


internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
```diff
@@ -320,9 +320,29 @@ private ConfigKeys() {
   public static final String ALLOW_CLUSTER_WIPE = "allow.cluster.wipe";
 
   /**
-   * Whether the controller is in Azure fabric. Default is false.
+   * Whether the controller is in a cloud environment. Default is false.
    */
-  public static final String CONTROLLER_IN_AZURE_FABRIC = "controller.in.azure.fabric";
+  public static final String CONTROLLER_CLOUD_ENABLED = "controller.cloud.enabled";
```
nisargthakkar, line 325: I think we should keep two different flags to act as the feature flags for controller cluster and storage cluster. We can roll both out separately in that case

services/venice-controller/src/main/java/com/linkedin/venice/controller/ZkHelixAdminClient.java
```diff
@@ -328,4 +325,47 @@ public void close() {
...
+              + Arrays.toString(CloudProvider.values()));
+    }
+
+    String controllerCloudId = config.getControllerCloudId();
+    List<String> controllerCloudInfoSources = config.getControllerCloudInfoSources();
+    String controllerCloudInfoProcessorName = config.getControllerCloudInfoProcessorName();
+    CloudConfig.Builder cloudConfigBuilder =
+        new CloudConfig.Builder().setCloudEnabled(true).setCloudProvider(cloudProvider).setCloudID(controllerCloudId);
+
+    if (!cloudProvider.equals(CloudProvider.AZURE)) {
```
nisargthakkar, line 354: We should not have logic specific to AZURE, and other cloud providers. This will become a maintenance overhead. We can let Helix do the validations.

We can instead rely on the presence/absence of these configs and decide if we should set them or skip them. Helix docs say only  `CLOUD_ENABLED` and `CLOUD_PROVIDER ` are mandatory. We can apply the above checks for the rest.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerClusterConfig.java
```diff
@@ -894,7 +904,13 @@ public VeniceControllerClusterConfig(VeniceProperties props) {
...
     this.allowClusterWipe = props.getBoolean(ALLOW_CLUSTER_WIPE, false);
     this.concurrentInitRoutinesEnabled = props.getBoolean(CONCURRENT_INIT_ROUTINES_ENABLED, false);
-    this.controllerInAzureFabric = props.getBoolean(CONTROLLER_IN_AZURE_FABRIC, false);
+    this.controllerClusterHelixCloudEnabled = props.getBoolean(CONTROLLER_CLUSTER_HELIX_CLOUD_ENABLED, false);
+    this.controllerStorageClusterHelixCloudEnabled =
+        props.getBoolean(CONTROLLER_STORAGE_CLUSTER_HELIX_CLOUD_ENABLED, false);
+    this.controllerHelixCloudProvider = props.getString(CONTROLLER_HELIX_CLOUD_PROVIDER, "");
+    this.controllerHelixCloudId = props.getString(CONTROLLER_HELIX_CLOUD_ID, "");
+    this.controllerHelixCloudInfoProcessorName = props.getString(CONTROLLER_HELIX_CLOUD_INFO_PROCESSOR_NAME, "");
+    this.controllerHelixCloudInfoSources = props.getList(CONTROLLER_HELIX_CLOUD_INFO_SOURCES, Collections.emptyList());
```
nisargthakkar, line 913: We need to handle the empty string issue here as well, right?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -516,7 +516,7 @@ public VeniceHelixAdmin(
       throw new VeniceException("Failed to connect to ZK within " + ZkClient.DEFAULT_CONNECTION_TIMEOUT + " ms!");
     }
     this.admin = new ZKHelixAdmin(zkClientForHelixAdmin);
-    this.helixAdminClient = new ZkHelixAdminClient(multiClusterConfigs, metricsRepository);
+    this.helixAdminClient = new ZkHelixAdminClient(commonConfig, multiClusterConfigs, metricsRepository);
```
nisargthakkar, line 519: Can we make `ZkHelixAdminClient` extract common config from `multiClusterConfigs.getCommonConfig()` instead?

services/venice-controller/src/main/java/com/linkedin/venice/controller/ZkHelixAdminClient.java
```diff
@@ -328,4 +325,35 @@ public void close() {
   public void addInstanceTag(String clusterName, String instanceName, String tag) {
     helixAdmin.addInstanceTag(clusterName, instanceName, tag);
   }
+
+  public void setCloudConfig(VeniceControllerClusterConfig config, boolean isControllerCluster) {
```
nisargthakkar, line 329: I like how you've abstracted both of these. But instead of passing `isControllerCluster` can we check in the caller if the cloud config should be set and only then call `setCloudConfig`?

So, that'd be like these two:
```java
if (commonConfig.isControllerClusterHelixCloudEnabled) {
    setCloudConfig(commonConfig);
}
```

```java
if (config.isControllerStorageClusterHelixCloudEnabled) {
    setCloudConfig(config);
}
```

Passing a boolean and handling it two different ways based on knowledge of when this boolean is set kind of breaks abstraction principles

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerClusterConfig.java
```diff
@@ -894,7 +904,13 @@ public VeniceControllerClusterConfig(VeniceProperties props) {
     this.emergencySourceRegion = props.getString(EMERGENCY_SOURCE_REGION, "");
     this.allowClusterWipe = props.getBoolean(ALLOW_CLUSTER_WIPE, false);
     this.concurrentInitRoutinesEnabled = props.getBoolean(CONCURRENT_INIT_ROUTINES_ENABLED, false);
-    this.controllerInAzureFabric = props.getBoolean(CONTROLLER_IN_AZURE_FABRIC, false);
+    this.controllerClusterHelixCloudEnabled = props.getBoolean(CONTROLLER_CLUSTER_HELIX_CLOUD_ENABLED, false);
+    this.controllerStorageClusterHelixCloudEnabled =
+        props.getBoolean(CONTROLLER_STORAGE_CLUSTER_HELIX_CLOUD_ENABLED, false);
+    this.controllerHelixCloudProvider = props.getString(CONTROLLER_HELIX_CLOUD_PROVIDER, "");
```
nisargthakkar, line 910: Let's make this into a `CloudProvider` object one time instead of doing `.toUpperCase` and `.valueOf` every time we read this value?

services/venice-controller/src/main/java/com/linkedin/venice/controller/ZkHelixAdminClient.java
```diff
@@ -328,4 +325,35 @@ public void close() {
...
+              + Arrays.toString(CloudProvider.values()));
+    }
+
+    String controllerCloudId = config.getControllerHelixCloudId();
+    List<String> controllerCloudInfoSources = config.getControllerHelixCloudInfoSources();
+    String controllerCloudInfoProcessorName = config.getControllerHelixCloudInfoProcessorName();
+    CloudConfig.Builder cloudConfigBuilder =
+        new CloudConfig.Builder().setCloudEnabled(true).setCloudProvider(cloudProvider).setCloudID(controllerCloudId);
+
+    controllerCloudInfoSources.forEach(cloudConfigBuilder::addCloudInfoSource);
```
nisargthakkar, line 352: We don't need to do it this way. We can instead do `builder.setCloudInfoSources(controllerCloudInfoSources)`. But we need to have an list emptiness check

services/venice-controller/src/main/java/com/linkedin/venice/controller/ZkHelixAdminClient.java
```diff
@@ -328,4 +325,35 @@ public void close() {
...
+      throw new VeniceException(
+          "Invalid Helix cloud provider: " + controllerCloudProvider + ". Must be one of: "
+              + Arrays.toString(CloudProvider.values()));
+    }
+
+    String controllerCloudId = config.getControllerHelixCloudId();
+    List<String> controllerCloudInfoSources = config.getControllerHelixCloudInfoSources();
+    String controllerCloudInfoProcessorName = config.getControllerHelixCloudInfoProcessorName();
+    CloudConfig.Builder cloudConfigBuilder =
+        new CloudConfig.Builder().setCloudEnabled(true).setCloudProvider(cloudProvider).setCloudID(controllerCloudId);
```
nisargthakkar, line 350: `CLOUD_ID` is an optional field. Let's not set it if the configs don't specify a value for it

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerClusterConfig.java
kvargha: Yes. Good catch.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerClusterConfig.java
kvargha: That would require moving the exception to `VeniceControllerClusterConfig`. Is that ok?

If not, I can definitely move the toUpperCase there.

nisargthakkar: Yes, that's okay. It is better that way actually so that we fail on host startup instead of waiting to fail till we need to use it

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerClusterConfig.java
```diff
@@ -894,7 +905,33 @@ public VeniceControllerClusterConfig(VeniceProperties props) {
     this.emergencySourceRegion = props.getString(EMERGENCY_SOURCE_REGION, "");
     this.allowClusterWipe = props.getBoolean(ALLOW_CLUSTER_WIPE, false);
     this.concurrentInitRoutinesEnabled = props.getBoolean(CONCURRENT_INIT_ROUTINES_ENABLED, false);
-    this.controllerInAzureFabric = props.getBoolean(CONTROLLER_IN_AZURE_FABRIC, false);
+    this.controllerClusterHelixCloudEnabled = props.getBoolean(CONTROLLER_CLUSTER_HELIX_CLOUD_ENABLED, false);
+    this.controllerStorageClusterHelixCloudEnabled =
+        props.getBoolean(CONTROLLER_STORAGE_CLUSTER_HELIX_CLOUD_ENABLED, false);
+
+    if (controllerClusterHelixCloudEnabled || controllerStorageClusterHelixCloudEnabled) {
+      String controllerCloudProvider = props.getString(CONTROLLER_HELIX_CLOUD_PROVIDER, "").toUpperCase();
```
nisargthakkar, line 913: Since we need this field to be set if `controllerClusterHelixCloudEnabled` or `controllerStorageClusterHelixCloudEnabled`, can we remove the default value of `""` ? That way we ensure that this is always provided

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerClusterConfig.java
```diff
@@ -894,7 +905,33 @@ public VeniceControllerClusterConfig(VeniceProperties props) {
...
+    }
+
+    this.controllerHelixCloudId = props.getString(CONTROLLER_HELIX_CLOUD_ID, "");
+    this.controllerHelixCloudInfoProcessorName = props.getString(CONTROLLER_HELIX_CLOUD_INFO_PROCESSOR_NAME, "");
+
+    if (props.getString(CONTROLLER_HELIX_CLOUD_INFO_SOURCES, "").isEmpty()) {
+      this.controllerHelixCloudInfoSources = Collections.emptyList();
+    } else {
+      this.controllerHelixCloudInfoSources =
+          props.getList(CONTROLLER_HELIX_CLOUD_INFO_SOURCES, Collections.emptyList());
```
nisargthakkar, line 932: Do we need to provide a default value here?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerClusterConfig.java
```diff
@@ -1533,8 +1570,28 @@ public boolean isConcurrentInitRoutinesEnabled() {
...
+    return controllerHelixCloudId;
+  }
+
+  public List<String> getControllerHelixCloudInfoSources() {
+    return controllerHelixCloudInfoSources;
+  }
+
+  public String getControllerHelixCloudInfoProcessorName() {
+    return controllerHelixCloudInfoProcessorName;
   }
```
nisargthakkar, line 1595: Can we remove the `Controller` name from the method and variable names?

services/venice-controller/src/main/java/com/linkedin/venice/controller/ZkHelixAdminClient.java
```diff
@@ -328,4 +326,25 @@ public void close() {
...
+
+    if (!controllerCloudInfoSources.isEmpty()) {
+      cloudConfigBuilder.setCloudInfoSources(controllerCloudInfoSources);
+    }
+
+    if (!controllerCloudInfoProcessorName.isEmpty()) {
+      cloudConfigBuilder.setCloudInfoProcessorName(controllerCloudInfoProcessorName);
+    }
+    helixAdmin.addCloudConfig(controllerClusterName, cloudConfigBuilder.build());
+  }
```
nisargthakkar, line 349: Same here, let's remove the `Controller` from naming here

nisargthakkar (approved):
LGTM. Thanks

PR #1213, opened 2024-10-03
[server] Increased Wait After Unsubscribe During State Transitions

KaiSernLim:
### Problem

The problem is that `StoreIngestionTask` does not always wait for all inflight messages to be processed before transitioning the leader-follower state in the `PartitionConsumptionState`.

[`waitAfterUnsubscribe()`](https://github.com/linkedin/venice/blob/48b155c58b0a08a71577ffc970fd184fb0c5982c/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/SharedKafkaConsumer.java#L204) waits up to 10 seconds for the consumers next [`poll()`](https://github.com/linkedin/venice/blob/48b155c58b0a08a71577ffc970fd184fb0c5982c/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/SharedKafkaConsumer.java#L247) (which would indicate that the inflight messages from the last `poll()` were processed). This can lead to state mismatches such as from the [leader-follower transition](https://github.com/linkedin/venice/blob/48b155c58b0a08a71577ffc970fd184fb0c5982c/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java#L2177) and [follower-leader transition](https://github.com/linkedin/venice/blob/48b155c58b0a08a71577ffc970fd184fb0c5982c/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java#L2571). The `10s` timeout has been hit 150K times in the past month.

### Partial Solution

Several possible solutions were discussed but they all could be complicated. As an immediate action, we can increase the timeout value so that the consumer will more frequently safely unsubscribe instead of timing out. This should only be during partition state transitions because those are the times when not waiting for all inflight messages to be processed could result in a state mismatch. According to the values of the metric `consumer_records_producing_to_write_buffer_latency`, increasing the timeout to `1800s` should cover almost 100% of all cases.

### Changes

* During state transition `unsubscribe()` calls from `StoreIngestionTask`, increased the timeout wait in `SharedKafkaConsumer#waitAfterUnsubscribe()` from `1800s` / `30m` based on the maximum value of the `consumer_records_producing_to_write_buffer_latency` metric.
* Added log message for when `SharedKafkaConsumer#waitAfterUnsubscribe()` takes longer than 15 seconds to gather some data about how long it would take.

## How was this PR tested?

GHCI


clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaConsumerService.java
```diff
@@ -68,6 +68,15 @@
  * @see AggKafkaConsumerService which wraps one instance of this class per Kafka cluster.
  */
 public abstract class KafkaConsumerService extends AbstractKafkaConsumerService {
+  /**
+   * Waiting for the next poll() after unsubscribing, indicating that all previous inflight messages were processed
+   */
+  public static final int DEFAULT_WAIT_AFTER_UNSUBSCRIBE_TIMEOUT_MS = (int) TimeUnit.MINUTES.toMillis(5);
```
ZacAttack, line 74: ech.  I guess this is ok, but I think in most parts of the code we'd just write 300000 with a comment of // 5 minutes

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/KafkaConsumerServiceStats.java
```diff
@@ -102,6 +103,7 @@ public KafkaConsumerServiceStats(
     detectedNoRunningIngestionTopicPartitionNumSensor =
         registerSensor("detected_no_running_ingestion_topic_partition_num", new Total());
     delegateSubscribeLatencySensor = registerSensor("delegate_subscribe_latency", new Avg(), new Max());
+    waitAfterUnsubscribeLatencySensor = registerSensor("wait_after_unsubscribe_latency", new Avg(), new Max());
```
ZacAttack, line 106: nit/request: Can we drop a comment somewhere in here that describes that this is measuring?  The name alone is a little cryptic

ZacAttack: Also... What does this tells us?  Why is it useful?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/SharedKafkaConsumer.java
```diff
@@ -219,11 +239,12 @@ protected void waitAfterUnsubscribe(long currentPollTimes, Set<PubSubTopicPartit
       LOGGER.info("Wait for poll request in `unsubscribe` function got interrupted.");
       Thread.currentThread().interrupt();
     }
+    stats.recordTotalWaitAfterUnsubscribeLatency((int) (time.getMilliseconds() - startTimeMs));
```
ZacAttack, line 242: This will overflow right? time.getMilliseconds if it's returning the system epoch time in millis will overflow the integer.  I think we should just use long's here

ZacAttack: Oh, the -, I guess that wouldn't overflow.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/KafkaConsumerServiceStats.java
ZacAttack: also, and I know most/all the other stats here don't do this, but It's a pattern we're hoping to correct going forward, let's include the time unit in the name of the metric.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/KafkaConsumerServiceStats.java
```diff
@@ -155,6 +157,10 @@ public void recordDelegateSubscribeLatency(double value) {
     delegateSubscribeLatencySensor.record(value);
   }
 
+  public void recordWaitAfterUnsubscribeLatency(int value) {
+    waitAfterUnsubscribeLatencySensor.record(value);
```
ZacAttack, line 161: I think we're leaking more casts because we're doing an int parameter here on this interface.  Why not cast to int in here and remove the casts from other parts of the code?

ZacAttack: actually, why cast at all? record takes a long.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaConsumerService.java
KaiSernLim: I don't know why you're so afraid of casts, hehe

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/KafkaConsumerServiceStats.java
KaiSernLim: You bring up a good point: why is this metric needed? It turns out that there is an existing metric that tells us about the slowdown on the drainer: `consumer_records_producing_to_write_buffer_latency`. Thus, this metric is unnecessary and has been removed.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/KafkaConsumerServiceStats.java
KaiSernLim: Reverted back to using `long` everywhere. The `record()` function seems to only take `double` values, so I'm just going to keep it as `long` everywhere instead of casting to `int` at the end?

KaiSernLim:
Closing to consider approach in #1251 instead.

gaojieliu (commented):
From my understanding, we don't need to change the constructor of `KafkaConsumerService` or `SharedKafkaConsumer`, but just add one param: `waitingTimeMS` to `unsubscribe` function call, and when the `SIT` calls `unsubscribe`, it knows when to use a longer timeout and when to use a shorter timeout.
By default, it should use a short timeout, and only for `leader` <-> `follower` transition, `SIT` needs to pass a long timeout to `unsubscribe` call.

Let me know if I have any misunderstanding.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaConsumerService.java
```diff
@@ -243,15 +243,15 @@ public void unsubscribeAll(PubSubTopic versionTopic) {
    * Stop specific subscription associated with the given version topic.
    */
   @Override
-  public void unSubscribe(PubSubTopic versionTopic, PubSubTopicPartition pubSubTopicPartition) {
-    PubSubConsumerAdapter consumer = getConsumerAssignedToVersionTopicPartition(versionTopic, pubSubTopicPartition);
+  public void unSubscribe(PubSubTopic versionTopic, PubSubTopicPartition pubSubTopicPartition, long timeoutMs) {
+    SharedKafkaConsumer consumer = getConsumerAssignedToVersionTopicPartition(versionTopic, pubSubTopicPartition);
```
KaiSernLim, line 247: This change suggested by @haoxu07 is really neat. `PubSubConsumerAdapter` is under `venice-common` and cannot access the default value static variable because it's in `clients/da-vinci-client`. Thus, it cannot easily create an overloaded version of `unSubscribe()` referencing the default wait.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -3405,7 +3405,12 @@ public void consumerUnSubscribe(PubSubTopic topic, PartitionConsumptionState par
     Instant startTime = Instant.now();
```
gaojieliu, line 3405: Nit:
Can we rename it to be `consumerUnSubscribeForStateTransition` to make the intention clear?

I hope other people won't use it accidentally.

gaojieliu (approved):
Sorry for the late review..
The code change itself looks good and just left a minor comment.

gaojieliu (approved):

PR #1214, opened 2024-10-03
[common][controller] Introduce new VeniceView config: MaterializedView

xunyin8:
## [common][controller] Introduce new VeniceView config: MaterializedView
1. Added new VeniceView implementation MaterializedView. This view will be used later in controller, server ingestion and DVC to create, maintain and materialize the re-partition view. Later we will also add projection and etc.
2. Modified corresponding static methods to ensure they function as expected for both ChangeCaputureView and RePartitionView.
3. Added logic in update view methods in controller to put view name into the view parameters for MaterializedView since one store version can have multiple materialized view.

## How was this PR tested?
New and existing unit tests. Will add corresponding integration tests once the config is actually being used (in the following PRs)

### DIFF
--- a/internal/venice-common/src/main/java/com/linkedin/venice/views/MaterializedView.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/views/MaterializedView.java
@@ -74,14 +74,8 @@ public class MaterializedView extends VeniceView {
       if (viewConfig.getViewClassName().equals(MaterializedView.class.getCanonicalName())) {
         String configPartitioner =
             viewConfig.getViewParameters().get(ViewParameterKeys.MATERIALIZED_VIEW_PARTITIONER.name());
-        if (configPartitioner == null) {
-          configPartitioner = store.getPartitionerConfig().getPartitionerClass();
-        }
-        String configPartitionCountString =
-            viewConfig.getViewParameters().get(ViewParameterKeys.MATERIALIZED_VIEW_PARTITION_COUNT.name());
-        int configPartitionCount = configPartitionCountString == null
-            ? store.getPartitionCount()
-            : Integer.parseInt(configPartitionCountString);
+        int configPartitionCount = Integer
+            .parseInt(viewConfig.getViewParameters().get(ViewParameterKeys.MATERIALIZED_VIEW_PARTITION_COUNT.name()));
         if (configPartitionCount == viewPartitionCount && configPartitioner.equals(viewPartitioner)) {
           throw new VeniceException(
               "A view with identical view configs already exist, view name: " + viewConfigEntries.getKey());
--- a/internal/venice-common/src/main/java/com/linkedin/venice/views/MaterializedView.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/views/MaterializedView.java
@@ -10,48 +10,46 @@ import java.util.Map;
 import java.util.Properties;
 
 
-public class RePartitionView extends VeniceView {
-  public static final String RE_PARTITION_TOPIC_SUFFIX = "_rp";
+public class MaterializedView extends VeniceView {
+  public static final String MATERIALIZED_VIEW_TOPIC_SUFFIX = "_mv";
   private static final String MISSING_PARAMETER_MESSAGE = "%s is required for re-partition view!";
 
-  public RePartitionView(Properties props, Store store, Map<String, String> viewParameters) {
+  public MaterializedView(Properties props, Store store, Map<String, String> viewParameters) {
     super(props, store, viewParameters);
   }
 
   @Override
   public Map<String, VeniceProperties> getTopicNamesAndConfigsForVersion(int version) {
     VeniceProperties properties = new VeniceProperties(props);
-    String viewName = viewParameters.get(ViewParameterKeys.RE_PARTITION_VIEW_NAME.name());
+    String viewName = viewParameters.get(ViewParameterKeys.MATERIALIZED_VIEW_NAME.name());
     return Collections.singletonMap(
         Version.composeKafkaTopic(store.getName(), version) + VIEW_TOPIC_SEPARATOR + viewName
-            + RE_PARTITION_TOPIC_SUFFIX,
+            + MATERIALIZED_VIEW_TOPIC_SUFFIX,
         properties);
   }
 
   /**
-   * {@link ViewParameterKeys#RE_PARTITION_VIEW_PARTITION_COUNT} is required to configure a new re-partition view.
-   * {@link ViewParameterKeys#RE_PARTITION_VIEW_PARTITIONER} is optional. The re-partition view will use the store level
+   * {@link ViewParameterKeys#MATERIALIZED_VIEW_PARTITION_COUNT} is required to configure a new re-partition view.
+   * {@link ViewParameterKeys#MATERIALIZED_VIEW_PARTITIONER} is optional. The re-partition view will use the store level
    * partitioner config if it's not specified in the view parameters.
-   * {@link ViewParameterKeys#RE_PARTITION_VIEW_PARTITIONER_PARAMS} is optional.
+   * {@link ViewParameterKeys#MATERIALIZED_VIEW_PARTITIONER_PARAMS} is optional.
    */
   @Override
   public void validateConfigs() {
     super.validateConfigs();
-    String viewName = viewParameters.get(ViewParameterKeys.RE_PARTITION_VIEW_NAME.name());
+    String viewName = viewParameters.get(ViewParameterKeys.MATERIALIZED_VIEW_NAME.name());
     if (viewName == null) {
       throw new VeniceException(
-          String.format(MISSING_PARAMETER_MESSAGE, ViewParameterKeys.RE_PARTITION_VIEW_NAME.name()));
+          String.format(MISSING_PARAMETER_MESSAGE, ViewParameterKeys.MATERIALIZED_VIEW_NAME.name()));
     }
-    String partitionCountString = viewParameters.get(ViewParameterKeys.RE_PARTITION_VIEW_PARTITION_COUNT.name());
-    if (partitionCountString == null) {
-      throw new VeniceException(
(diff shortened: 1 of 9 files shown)


internal/venice-common/src/main/java/com/linkedin/venice/views/MaterializedView.java
```diff
@@ -0,0 +1,57 @@
...
+public class MaterializedView extends VeniceView {
+  public static final String MATERIALIZED_VIEW_TOPIC_SUFFIX = "_mv";
+  private static final String MISSING_PARAMETER_MESSAGE = "%s is required for re-partition view!";
+
+  public MaterializedView(Properties props, Store store, Map<String, String> viewParameters) {
+    super(props, store, viewParameters);
+  }
+
+  @Override
+  public Map<String, VeniceProperties> getTopicNamesAndConfigsForVersion(int version) {
```
gaojieliu, line 22: I understand this is an existing API, but the returned object seems to be a `Pair` rather a map as it would only contain one entry, right?

internal/venice-common/src/main/java/com/linkedin/venice/views/MaterializedView.java
```diff
@@ -0,0 +1,57 @@
...
+  /**
+   * {@link ViewParameterKeys#MATERIALIZED_VIEW_PARTITION_COUNT} is required to configure a new re-partition view.
+   * {@link ViewParameterKeys#MATERIALIZED_VIEW_PARTITIONER} is optional. The re-partition view will use the store level
+   * partitioner config if it's not specified in the view parameters.
+   * {@link ViewParameterKeys#MATERIALIZED_VIEW_PARTITIONER_PARAMS} is optional.
+   */
+  @Override
+  public void validateConfigs() {
+    super.validateConfigs();
+    String viewName = viewParameters.get(ViewParameterKeys.MATERIALIZED_VIEW_NAME.name());
```
gaojieliu, line 40: Shall we check whether the current store has a view with a same name or not?

internal/venice-common/src/main/java/com/linkedin/venice/views/MaterializedView.java
```diff
@@ -0,0 +1,57 @@
...
+    if (viewName == null) {
+      throw new VeniceException(
+          String.format(MISSING_PARAMETER_MESSAGE, ViewParameterKeys.MATERIALIZED_VIEW_NAME.name()));
+    }
+    String partitionCountString = viewParameters.get(ViewParameterKeys.MATERIALIZED_VIEW_PARTITION_COUNT.name());
+    boolean samePartitionCount =
+        partitionCountString == null || store.getPartitionCount() == Integer.parseInt(partitionCountString);
+    // A re-partition view with the exact same partitioner and partition count makes no sense
+    String partitioner = viewParameters.get(ViewParameterKeys.MATERIALIZED_VIEW_PARTITIONER.name());
+    boolean samePartitioner =
```
gaojieliu, line 50: Shall we check whether the store already has a view with the same partition count/partitioner or not to avoid duplication?

internal/venice-common/src/main/java/com/linkedin/venice/views/VeniceView.java
```diff
@@ -99,7 +101,7 @@ public static String parseStoreFromViewTopic(String topicName) {
   // TODO: see above TODO for isViewTopic function, same applies here
   public static int parseVersionFromViewTopic(String topicName) {
     int versionStartIndex = Version.getLastIndexOfVersionSeparator(topicName) + Version.VERSION_SEPARATOR.length();
```
gaojieliu, line 103: Shall we check whether this topic is a view topic or not before parsing?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
```diff
@@ -2780,18 +2785,32 @@ && getVeniceHelixAdmin().isHybrid(setStore.getHybridStoreConfig()) && setStore.g
...
+    return validatedConfigs;
   }
 
-  private void validateStoreViewConfig(Store store, ViewConfig viewConfig) {
+  private ViewConfig validateAndDecorateStoreViewConfig(Store store, ViewConfig viewConfig, String viewName) {
     // TODO: Pass a proper properties object here. Today this isn't used in this context
+    if (viewConfig.getViewClassName().equals(MaterializedView.class.getCanonicalName())) {
+      if (viewName.contains(VERSION_SEPARATOR)) {
+        throw new VeniceException(
+            String.format("Re-partition View name cannot contain version separator: %s", VERSION_SEPARATOR));
```
gaojieliu, line 2804: `Re-partition` -> `Materialized`?

internal/venice-common/src/main/java/com/linkedin/venice/views/MaterializedView.java
```diff
@@ -0,0 +1,57 @@
...
+import com.linkedin.venice.meta.ViewParameterKeys;
+import com.linkedin.venice.utils.VeniceProperties;
+import java.util.Collections;
+import java.util.Map;
+import java.util.Properties;
+
+
+public class MaterializedView extends VeniceView {
+  public static final String MATERIALIZED_VIEW_TOPIC_SUFFIX = "_mv";
+  private static final String MISSING_PARAMETER_MESSAGE = "%s is required for re-partition view!";
```
gaojieliu, line 15: `re-partition` => `materialized`?

internal/venice-common/src/main/java/com/linkedin/venice/views/MaterializedView.java
```diff
@@ -0,0 +1,57 @@
...
+    String viewName = viewParameters.get(ViewParameterKeys.MATERIALIZED_VIEW_NAME.name());
+    if (viewName == null) {
+      throw new VeniceException(
+          String.format(MISSING_PARAMETER_MESSAGE, ViewParameterKeys.MATERIALIZED_VIEW_NAME.name()));
+    }
+    String partitionCountString = viewParameters.get(ViewParameterKeys.MATERIALIZED_VIEW_PARTITION_COUNT.name());
+    boolean samePartitionCount =
+        partitionCountString == null || store.getPartitionCount() == Integer.parseInt(partitionCountString);
+    // A re-partition view with the exact same partitioner and partition count makes no sense
+    String partitioner = viewParameters.get(ViewParameterKeys.MATERIALIZED_VIEW_PARTITIONER.name());
```
gaojieliu, line 49: Shall we check whether the partitioner class is present or not?

internal/venice-common/src/main/java/com/linkedin/venice/views/MaterializedView.java
xunyin8: Correct, I'm guessing that the original intention is that some view could have multiple topics?

internal/venice-common/src/main/java/com/linkedin/venice/views/MaterializedView.java
xunyin8: Good call, let me add that check.

internal/venice-common/src/main/java/com/linkedin/venice/views/MaterializedView.java
gaojieliu: Really?
@ZacAttack Do you know such use case?

internal/venice-common/src/main/java/com/linkedin/venice/views/MaterializedView.java
xunyin8: hmm the idea is that if it's not specified then it will use the default partitioner of the store. This is for users that would only like to have a different partition count but don't want to mingle with the existing partitioner.

gaojieliu: If the customer specifies a different partitioner class, how do we know whether this class is valid or not?

internal/venice-common/src/main/java/com/linkedin/venice/views/MaterializedView.java
gaojieliu: To speed up the review, and we can ignore this comment for now as it is not a big issue and we can refactor it later if needed.

internal/venice-common/src/main/java/com/linkedin/venice/views/MaterializedView.java
xunyin8: I see what you mean now.

internal/venice-common/src/main/java/com/linkedin/venice/views/MaterializedView.java
```diff
@@ -0,0 +1,88 @@
...
+      throw new VeniceException(
+          "A materialized view with the same partitioner and partition count as the original store is not allowed!");
+    }
+    // Check if there is already a materialized view with identical configs
+    for (Map.Entry<String, ViewConfig> viewConfigEntries: store.getViewConfigs().entrySet()) {
+      ViewConfig viewConfig = viewConfigEntries.getValue();
+      if (viewConfig.getViewClassName().equals(MaterializedView.class.getCanonicalName())) {
+        String configPartitioner =
+            viewConfig.getViewParameters().get(ViewParameterKeys.MATERIALIZED_VIEW_PARTITIONER.name());
+        if (configPartitioner == null) {
```
gaojieliu, line 73: In theory, the configPartitioner should never be null as if it is absent at the view config registration time, we will backfill it by the store-level config, right?
We should always have a valid partitioner in the view config as the store-level partitioner can change in the future.

gaojieliu (commented):
Left one minor comment and the code change looks good overall.

internal/venice-common/src/main/java/com/linkedin/venice/views/MaterializedView.java
xunyin8: Good call, let me add that filling action.

gaojieliu (approved):
Thanks!

PR #1216, opened 2024-10-04
[changelog] Centralize pubsubconsumer assignment to synchronized collection

ZacAttack:
## [changelog] Centralize pubsubconsumer assignment to synchronized collection

getAssignment returns a new HashSet, but it's backed by the raw collection, making it non threadsafe.  Tweaks here to make sure we don't hit any weird behavior or concurrent modification exceptions by serializing on the assignment.

Resolves #XXX


clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumerImpl.java
```diff
@@ -930,10 +940,15 @@ private boolean filterRecordByVersionSwapHighWatermarks(
     return false;
   }
 
+  protected Set<PubSubTopicPartition> getTopicAssignment() {
+    return Collections.synchronizedSet(pubSubConsumer.getAssignment());
```
majisourav99, line 944: This would just return a synchronized set to the caller. IIUC we need to add sync on `pubSubConsumer` in places like switchToNewTopic to prevent concurrent exception.

ZacAttack: The exception thrown from earlier is because the assignment collection returned isn't a copy.  This change addresses that particular problem.

I'm not a big fan of adding synchronized blocks unless theres a strong logical need to.  My scan through of uses of pubsubconsumer seem to indicate that most of the calls are already protected by synchronized calls, and where it's not I don't see an obvious concurrency issue.  Are there calls in particular you are worried about?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumerImpl.java
```diff
@@ -219,7 +220,7 @@ public CompletableFuture<Void> internalSubscribe(Set<Integer> partitions, PubSub
       }
 
       synchronized (pubSubConsumer) {
```
majisourav99, line 222: `getTopicAssignment` now locks on pubSubConsumer, we don need this synchronized anymore?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumerImpl.java
```diff
@@ -219,7 +220,7 @@ public CompletableFuture<Void> internalSubscribe(Set<Integer> partitions, PubSub
...
-        Set<PubSubTopicPartition> topicPartitionSet = new HashSet<>(pubSubConsumer.getAssignment());
+        Set<PubSubTopicPartition> topicPartitionSet = getTopicAssignment();
```
majisourav99, line 223: We need lock on `topicPartitionSet` here?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumerImpl.java
lusong64: Per javadoc, even the set has Synchronized keyword on it, it still need be put a synchronzied keyword around the traverse. Is it what you want Zac? would a ConcurrentSkipListSet or concurrencyHashMap do a better job? 

Per the javadoc, even the set has Synchronized keyword on it, it still need be put a synchronzied keyword around the traverse. Is it what you want Zac? would a ConcurrentSkipListSet or concurrencyHashMap do a better job?

ZacAttack: Which is what the PR is already doing.

But more aside to the problem, even just calling getAssignment() on the pubsubconsumer can barf, so the underlying implementation of pubsubconsumer is entirey non thread safe.  So the PR has just been updated to synchronize all access instead.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumerImpl.java
ZacAttack: I don't think so.  We're already synchronized on pubSubConsumer and the synchronizedSet to boot won't throw the exception.  I think we're safe.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumerImpl.java
ZacAttack: We do if only because we later in this synchronized block change the assignment via subscription.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceAfterImageConsumerImpl.java
```diff
@@ -191,24 +189,22 @@ public void run() {
...
-                    + " on VersionSwapDetectionThread: " + this.getId() + "will retry...",
-                e);
-          }
+        }
+        try {
+          LOGGER.info(
+              "New Version detected!  Seeking consumer to version: " + currentVersion + " in consumer: "
+                  + changelogClientConfig.getConsumerName());
+          seekToEndOfPush(partitions).get();
+        } catch (InterruptedException | ExecutionException e) {
```
majisourav99, line 201: can we catch all Exception here?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumerImpl.java
majisourav99: The previous synchronized is for preventing the internal new HashSet<> concurrent exception issue, but now its not needed as you already added the check inside `getTopicAssignment`. But for the synchronizedSet, we need to do a locks on the returned object. https://docs.oracle.com/javase/8/docs/api/java/util/Collections.html#synchronizedSet-java.util.Set-

ZacAttack: The above is speaking to the behavior of iteration on this set if the contents change during the iteration.  We're syncrhonizing on the pubsubconsumer.  If we synchronize on the pubsub consumer, I don't know how this collection can change.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceAfterImageConsumerImpl.java
ZacAttack: Yeah ok

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceAfterImageConsumerImpl.java
```diff
@@ -34,7 +36,8 @@ public class VeniceAfterImageConsumerImpl<K, V> extends VeniceChangelogConsumerI
   // TODO: We shouldn't use this in the long run. Once the EOP position is queryable from venice and version
   // swap is produced to VT, then we should remove this as it's no longer needed.
   final private Lazy<VeniceChangelogConsumerImpl<K, V>> internalSeekConsumer;
-  private Thread versionSwapDetectionThread;
+  private final ScheduledExecutorService versionSwapExecutorService = Executors.newSingleThreadScheduledExecutor();
+  boolean versionSwapThreadScheduled = false;
```
majisourav99, line 40: Should we make this AtomicBoolean?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceAfterImageConsumerImpl.java
majisourav99: Also a bigger try-catch.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumerImpl.java
majisourav99: My point was we dont need to synchronize on the pubsub consumer given `getTopicAssignment()` is doing it now already.

ZacAttack: Right, but the rest of the code block does need the synchronization because we're changing the assignment.  I guess we could shrink it if we want?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceAfterImageConsumerImpl.java
ZacAttack: sure

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceAfterImageConsumerImpl.java
ZacAttack: ok

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumerImpl.java
lusong64: I went back to double check the places that call getTopicAssignment. And almost all places used synchronized except for one, line 953 of VeniceChangelogConsumerImpl.java, which simply traverses the set. So some thread's change on the thread may not be seen by whoever runs on line 953. Are we good with it?

ZacAttack: Yeah, this context is pretty safe.  Making the copy is probably cheaper.  However, we have to use synchronization because of the implementation of the KafkaPubSubConsumer.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceAfterImageConsumerImpl.java
```diff
@@ -150,65 +163,50 @@ protected CompletableFuture<Void> internalSeek(
...
+              partitions.add(partitionSubscription.getPartitionNumber());
+            }
           }
+
+          LOGGER.info(
+              "New Version detected!  Seeking consumer to version: " + currentVersion + " in consumer: "
+                  + changelogClientConfig.getConsumerName());
+          seekToEndOfPush(partitions).get();
         }
+      } catch (InterruptedException | ExecutionException e) {
```
majisourav99, line 204: can we catch all exception here?

ZacAttack: Sure!

majisourav99 (approved):
LGTM! Thanks Zac

PR #1217, opened 2024-10-04
[controller] Refresh OfflinePushStatus after change listener subscription during loadAllPushes() call

sixpluszero:
## [controller] Refresh OfflinePushStatus after change listener subscription during loadAllPushes() call
The issue:
We observed such race condition in controller deployment/restart. When it restarts, it will call loadAllPushes() which load all OfflinePushStatus snapshot for all topics from ZK, it will then iterate the list and try to subscribe to the changes. If there is any change in between, it will be missed and push job can hang forever.

This PR adds a refresh offline pushes call after listener subscription. It is using the same store level lock, so it should not cause any race condition or deadlock with change listener callback.

## How was this PR tested?
Add an unit test

### DIFF
--- a/services/venice-controller/src/main/java/com/linkedin/venice/pushmonitor/AbstractPushMonitor.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/pushmonitor/AbstractPushMonitor.java
@@ -151,7 +151,7 @@ public abstract class AbstractPushMonitor
       LOGGER.info("Load all pushes started for cluster {}'s {}", clusterName, getClass().getSimpleName());
       for (OfflinePushStatus offlinePushStatus: offlinePushStatusList) {
         try {
-          topicToPushMap.put(offlinePushStatus.getKafkaTopic(), offlinePushStatus);
+          // topicToPushMap.put(offlinePushStatus.getKafkaTopic(), offlinePushStatus);
           routingDataRepository.subscribeRoutingDataChange(offlinePushStatus.getKafkaTopic(), this);
           getOfflinePushAccessor().subscribePartitionStatusChange(offlinePushStatus, this);
           /**
@@ -893,12 +893,15 @@ public abstract class AbstractPushMonitor
       LOGGER.warn("Resource is remaining in the ideal state. Ignore the deletion in the external view.");
       return;
     }
-    OfflinePushStatus pushStatus;
-    pushStatus = getOfflinePush(kafkaTopic);
-    if (pushStatus != null && pushStatus.getCurrentStatus().equals(ExecutionStatus.STARTED)) {
-      String statusDetails = "Helix resource for Topic:" + kafkaTopic + " is deleted, stopping the running push.";
-      LOGGER.info(statusDetails);
-      handleTerminalOfflinePushUpdate(pushStatus, new ExecutionStatusWithDetails(ERROR, statusDetails));
+    String storeName = Version.parseStoreFromKafkaTopicName(kafkaTopic);
+    try (AutoCloseableLock ignore = clusterLockManager.createStoreWriteLock(storeName)) {
+      OfflinePushStatus pushStatus;
+      pushStatus = getOfflinePush(kafkaTopic);
+      if (pushStatus != null && pushStatus.getCurrentStatus().equals(ExecutionStatus.STARTED)) {
+        String statusDetails = "Helix resource for Topic:" + kafkaTopic + " is deleted, stopping the running push.";
+        LOGGER.info(statusDetails);
+        handleTerminalOfflinePushUpdate(pushStatus, new ExecutionStatusWithDetails(ERROR, statusDetails));
+      }
     }
   }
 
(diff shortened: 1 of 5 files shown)


services/venice-controller/src/main/java/com/linkedin/venice/pushmonitor/AbstractPushMonitor.java
```diff
@@ -149,30 +149,16 @@ private void loadAllPushes(List<OfflinePushStatus> offlinePushStatusList) {
...
     try (AutoCloseableLock ignore = clusterLockManager.createClusterWriteLock()) {
       LOGGER.info("Load all pushes started for cluster {}'s {}", clusterName, getClass().getSimpleName());
-      // Subscribe to changes first
-      List<OfflinePushStatus> refreshedOfflinePushStatusList = new ArrayList<>();
       for (OfflinePushStatus offlinePushStatus: offlinePushStatusList) {
         try {
+          topicToPushMap.put(offlinePushStatus.getKafkaTopic(), offlinePushStatus);
           routingDataRepository.subscribeRoutingDataChange(offlinePushStatus.getKafkaTopic(), this);
-
+          getOfflinePushAccessor().subscribePartitionStatusChange(offlinePushStatus, this);
```
lluwm, line 156: I am trying to understand where the race comes from. The `change` mentioned in the description, was it come from the `onPartitionStatusChange` code path? In that function, we seem to need to acquire a cluster level read lock, so supposedly, the change would be blocked as this code block is already holding a cluster write lock, isn't it? or the change is from some other code path that I misunderstood?

services/venice-controller/src/main/java/com/linkedin/venice/pushmonitor/AbstractPushMonitor.java
```diff
@@ -771,7 +769,9 @@ protected RoutingDataRepository getRoutingDataRepository() {
   @Override
   public void onPartitionStatusChange(String topic, ReadOnlyPartitionStatus partitionStatus) {
     String storeName = Version.parseStoreFromKafkaTopicName(topic);
+    LOGGER.info("DEBUGGING 123");
```
lluwm, line 772: remove the debugging log and there are several place in the later parts.

services/venice-controller/src/main/java/com/linkedin/venice/pushmonitor/AbstractPushMonitor.java
sixpluszero: Yeah, so it is not like this -
In the original code, we load OfflinePushStatus from ZK first for everyone in a loop and then we subscribe to change one by one, apparently, if there is any update to the ZNODE in between, they should get ignored.

lluwm: Got it. Thanks for the explanation!

services/venice-controller/src/main/java/com/linkedin/venice/pushmonitor/AbstractPushMonitor.java
```diff
@@ -321,6 +307,18 @@ protected OfflinePushStatus getOfflinePush(String topic) {
     return topicToPushMap.get(topic);
   }
 
+  protected void updateOfflinePush(String topic) {
+    String store = Version.parseStoreFromKafkaTopicName(topic);
+    try (AutoCloseableLock ignored = clusterLockManager.createStoreWriteLock(store)) {
+      OfflinePushStatus offlinePushStatus = getOfflinePushAccessor().getOfflinePushStatusAndItsPartitionStatuses(topic);
+      topicToPushMap.put(topic, offlinePushStatus);
+      LOGGER.info(
```
lluwm, line 315: nit, the log line can be moved out of the lock scope.

services/venice-controller/src/main/java/com/linkedin/venice/pushmonitor/AbstractPushMonitor.java
```diff
@@ -149,30 +149,16 @@ private void loadAllPushes(List<OfflinePushStatus> offlinePushStatusList) {
     pushStatusCollector.start();
     try (AutoCloseableLock ignore = clusterLockManager.createClusterWriteLock()) {
       LOGGER.info("Load all pushes started for cluster {}'s {}", clusterName, getClass().getSimpleName());
-      // Subscribe to changes first
-      List<OfflinePushStatus> refreshedOfflinePushStatusList = new ArrayList<>();
       for (OfflinePushStatus offlinePushStatus: offlinePushStatusList) {
         try {
+          topicToPushMap.put(offlinePushStatus.getKafkaTopic(), offlinePushStatus);
```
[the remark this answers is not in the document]
sixpluszero: Yes, so I think about this - if you check listener callback, it will first read the OfflinePushStatus object from the local map, if you don't put it first (even if it might going to be a bit stale), it will result in change got dropped. So this is necessary.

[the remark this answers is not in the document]
lluwm: thanks for the explain, we can keep them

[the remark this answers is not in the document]
sixpluszero: Synced offline, we decided to keep it for now as safeguard, while we add store write lock for another EV listener callback method

lluwm (approved):
LGTM and thanks for the fix @sixpluszero!

PR #1218, opened 2024-10-04
[server][dvc] Retry/Skip hosts that fail to connect or transfer files during blob transfer

jingy-li:
## [server][dvc] Retry/Skip hosts that fail to connect or transfer files during blob transfer
There is an issue in the peer discovery process during blob transfers where only the first host provided by discoverPeers is used. If the initial host fails to establish a connection, the system does not attempt to retry or connect to other available hosts.

This problem arises because, within the `NettyP2PTransferManager`, the inputStream obtained from `nettyClient.get` is returned even when it is associated with a failed CompletableFuture due to a connection error. NettyP2PBlobTransferManager do not explicitly verify whether the inputStream has failed before return it. Consequently, the inputStream from the first host is returned immediately, causing the for-loop that iterates through discoverPeers to exit prematurely.

---

This PR fix: 
1. Error handling: 
    - [Fatal case] if there is NO peers found for the requested blob, skip bootstrap via blob transfer.
    - [Fatal case] if ALL founded peers failed to connect or has no snapshot, skip bootstrap via blob transfer.
    - [Fatal case] if any unexpected exception occurs during the file/metadata transfer, skip bootstrap via blob transfer for saving time, do cleanup.
    - [Retry case] if one host has connect error,  it will retry up to max-retry times. Automatically switches to the next available host after reaching the maximum retry limit, ensuring continued attempts to complete the file transfer.
    - [Skip host case] If one host return 404 snapshot not found, it will skip and move to next host.

2. Relocates the completion check from the DefaultIngestionBackend to the NettyP2PBlobTransferManager. This adjustment allows for more clearer error handling and retry processes logic.
3. Minor: simplifies the server blob finder logic
4. Minor: shuffling discoverPeers list to ensure the hosts are randomly picked. 

## How was this PR tested?
Unit Test: Verify the retry mechanism and the ability to skip a faulty host and subsequently transfer files to the local host.

Integration Test: Ensure that if a server's snapshot does not exist, the mechanism moves to the next available host. If no hosts are available, the process should default to using Kafka for ingestion.


internal/venice-common/src/main/java/com/linkedin/venice/blobtransfer/ServerBlobFinder.java
```diff
@@ -27,17 +26,12 @@ public BlobPeersDiscoveryResponse discoverBlobPeers(String storeName, int versio
...
-          .getAllPartitions()) {
-        if (partition.getId() == partitionId) {
-          for (Instance instances: partition.getReadyToServeInstances()) {
-            hostNames.add(instances.getHost());
-          }
-        }
-      }
+      List<Instance> instances = customizedViewRepository.getReadyToServeInstances(currentVersionResource, partitionId);
+
+      List<String> hostNames = instances.stream().map(Instance::getHost).collect(Collectors.toList());
```
adamxchen, line 33: can we further streamline the codes to make it one liner?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/NettyP2PBlobTransferManager.java
```diff
@@ -51,19 +53,42 @@ public CompletionStage<InputStream> get(String storeName, int version, int parti
...
+      int retryCount = 0;
+      while (retryCount < MAX_RETRIES_FOR_BLOB_TRANSFER_PER_HOST) {
+        try {
+          // instanceName comes as a format of <hostName>_<applicationPort>
+          String chosenHost = peer.split("_")[0];
+          LOGGER.info("Attempt {} to connect to host: {}", retryCount + 1, chosenHost);
+          CompletionStage<InputStream> inputStreamStage = nettyClient.get(chosenHost, storeName, version, partition);
+          CompletableFuture<InputStream> inputStreamFuture = inputStreamStage.toCompletableFuture();
+
+          InputStream inputStream = inputStreamFuture.get(30, TimeUnit.MINUTES);
```
adamxchen, line 68: If we fail 3 times, doesn't it mean potentially we has wait up to 90 mins and makes no progress and we are doing this for every peer?? That would be unacceptable for use cases that needs fast bootstrap. Ideally, we should only try to connect and retrieve blobs up to one (or two) peer(s) before giving up the blob transfer and using kafka ingestion.
How about we split the `nettyClient.get` into two parts? The first part is for connecting and the second is to send the GET request to request blobs?

So we can have something like
```
CompletableFuture <Channel> connectionFuture = nettyClient.connect(chosenHost);
// if it connects, then
   CompletableFuture <InputStream> fileTransferFuture = nettyClient.get(storeName, version, partition); // attach handlers and send the GET request

// if it fails, then try next peer until a valid peer is found.
```
What do you think?

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/BlobP2PTransferAmongServersTest.java
```diff
@@ -90,28 +101,87 @@ public void testBlobP2PTransferAmongServers() throws Exception {
...
+      }
+    }
+
+    TestUtils.waitForNonDeterministicAssertion(2, TimeUnit.MINUTES, () -> {
+      cluster.stopAndRestartVeniceServer(server1.getPort());
+      Assert.assertTrue(server1.isRunning());
+    });
+
+    // wait for server 1 to ingest
+    Thread.sleep(120000);
```
adamxchen, line 170: hmm, this is not favorable. Can we try another way, e.g. check with router or controller to see if server 1 has finished the ingestion?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/NettyP2PBlobTransferManager.java
jingy-li: Yes, it's possible that after 90 minutes the transfer isn't completed and we have to move on to the next hosts. However, this would be a corner case where every host has a good connection, but the file transfer isn’t finished within 30 minutes. 

As for giving up after a few hosts, I have some doubts. In perf tests, there are 10+ hosts in the discoverPeers list, but we only need one specific host which has snapshot. Therefore, deciding to abandon the process based on the number of hosts may not be appropriate. Instead, let's establish a target timeframe. For instance, if the blob transfer isn't completed in 5 hours, we should give up and proceed to Kafka ingestion.

It may not be necessary to separate `nettyClient.get`, as we want to be able to retry or skip both connection issues and file transfer problems.

eldernewborn:
Have we considered picking random hosts, rather than first available ? 
picking the first can create un-even and at times excessive load on the hosts that happen to show up as first on the list for the rest of the fleet.

jingy-li:
> Have we considered picking random hosts, rather than first available ? picking the first can create un-even and at times excessive load on the hosts that happen to show up as first on the list for the rest of the fleet.

Yes, thanks for this good call! Add shuffling list step to ensure hosts are randomly picked.

internal/venice-common/src/main/java/com/linkedin/venice/blobtransfer/BlobPeersDiscoveryResponse.java
```diff
@@ -31,6 +32,9 @@ public void setDiscoveryResult(List<String> hostNames) {
   }
 
   public List<String> getDiscoveryResult() {
+    if (this.hostNameList != null && !this.hostNameList.isEmpty()) {
+      Collections.shuffle(this.hostNameList);
```
adamxchen, line 36: Every call to `getDiscoveryResult` would return a different list. The method name doesn't imply this. How about changing the method name so the caller would know?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/NettyP2PBlobTransferManager.java
```diff
@@ -39,31 +48,124 @@ public void start() throws Exception {
...
-        return inputStream;
-      } catch (Exception e) {
-        LOGGER.warn("Failed to connect to peer: {}", peer, e);
+      int retryCount = 0;
+      while (retryCount < MAX_RETRIES_FOR_BLOB_TRANSFER_PER_HOST) {
+        try {
+          // instanceName comes as a format of <hostName>_<applicationPort>
+          String chosenHost = peer.split("_")[0];
+          LOGGER.info("Attempt {} to connect to host: {}", retryCount + 1, chosenHost);
+          CompletionStage<InputStream> inputStreamStage = nettyClient.get(chosenHost, storeName, version, partition);
```
adamxchen, line 98: this can be done in on line. I'm also thinking whether we should change the interface return type. Codes are getting verbose unnecessarily. If needed, we can use `CompletionStage` in the future..

clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/DefaultIngestionBackend.java
```diff
@@ -110,32 +108,24 @@ CompletionStage<Void> bootstrapFromBlobs(Store store, int versionNumber, int par
     String storeName = store.getName();
     String baseDir = serverConfig.getRocksDBPath();
     try {
-      CompletableFuture<InputStream> p2pFuture =
-          blobTransferManager.get(storeName, versionNumber, partitionId).toCompletableFuture();
-      LOGGER.info(
-          "Bootstrapping from blobs for store {}, version {}, partition {}",
+      blobTransferManager.get(storeName, versionNumber, partitionId).toCompletableFuture();
```
sixpluszero, line 111: I think it is good to move all the peer discovery logic into the method, but there is gotcha here. It looks like you make this method a blocking call, and if you check the upstream caller method - VersionBackend.subscribe(), it is iterating over the list of partitions as previously it assumes 
        backend.getIngestionBackend().startConsumption(config, partition);
is a fast action, but now we make the blob transfer for multiple partitions sequential. 
I think we need to make them run in parallel, otherwise we are not utilizing the HW of DVC host.

sixpluszero (commented):
Thanks for making this change and adding a lot of tests, it is good practice to add some unit tests for critical behavior coverage. 
I left a few comments, most of them are simple, but the sequential vs parallel blob transfer issue I think it is needed to be tackled.

internal/venice-common/src/main/java/com/linkedin/venice/blobtransfer/BlobPeersDiscoveryResponse.java
adamxchen: I thought about that too.. But that'd mean we are adding the duplicated logic (tho very minimal) to server finder and dvc finder. That said, it's minor and i'm okay with that

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/NettyP2PBlobTransferManager.java
```diff
@@ -39,31 +48,124 @@ public void start() throws Exception {
...
+          LOGGER.info(
+              "Successfully fetched blob from peer {} for store {} partition {} version {} in {} seconds",
+              peer,
+              storeName,
+              partition,
+              version,
+              Duration.between(startTime, Instant.now()).getSeconds());
+          return CompletableFuture.completedFuture(inputStream);
+        } catch (Exception e) {
+          if (e.getCause() instanceof VenicePeersCannotConnectException) {
```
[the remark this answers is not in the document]
jingy-li: The reason is, the error is throw exception when use `inputStreamFuture.get(MAX_TIMEOUT_FOR_BLOB_TRANSFER_IN_MIN, TimeUnit.MINUTES);` the exception is actually `Execution exception`, not  `VeniceBlobTransferFileNotFoundException` or `VenicePeersCannotConnectException`. 

Because both connect error (at NettyFileTransferClient.get) or transfer error (at P2PFileTransferClientHandler.channelRead0), all exceptions are catch and then do ` inputStreamFuture.toCompletableFuture().completeExceptionally(cause);`

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/NettyP2PBlobTransferManager.java
```diff
@@ -39,31 +48,124 @@ public void start() throws Exception {
...
+            // solution: retry connecting to the peer again up to MAX_RETRIES_FOR_BLOB_TRANSFER_PER_HOST times
+            LOGGER.warn(
+                "Get error when connect to peer: {} for store {} partition {}, retrying {}/{}",
+                peer,
+                storeName,
+                partition,
+                retryCount + 1,
+                MAX_RETRIES_FOR_BLOB_TRANSFER_PER_HOST,
+                e);
+            retryCount++;
```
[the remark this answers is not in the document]
jingy-li: It because for single host connect, this is reasonable that there could be connect issue, so we can retry connect, but for other issue, [file transfer error, file not existed error], we can directly move on.

internal/venice-common/src/main/java/com/linkedin/venice/blobtransfer/BlobPeersDiscoveryResponse.java
jingy-li: Moved it to peer finder logic (server finder and dvc finder).

clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/DefaultIngestionBackend.java
sixpluszero: I see, that's fine if we want to fix it separately (but it might impact your batch only store benchmark though, unless you want to resize the store to only single partition)

sixpluszero (approved):
Looks good to me! Thank you for the fixes!

adamxchen (approved):

PR #1219, opened 2024-10-04
[controller] Enable backend to do dual reads on DVC incremental push status

huangminchn:
## Summary
Controller will first try to read from version level key for DVC incremental push status; if version level key exists and the overall status is END_OF_INCREMENTAL_PUSH, controller would still check the partition level key and combine the result from both keys, in case the DVC is partially upgraded.

Other changes:
Add a few more logs for each job status polling request to help us further confirm that the version level keys are working well, besides the signal of successful push jobs.

## How was this PR tested?
New tests in progress


services/venice-controller/src/main/java/com/linkedin/venice/pushmonitor/PushMonitorUtils.java
```diff
@@ -170,6 +170,12 @@ public static ExecutionStatusWithDetails getDaVinciPushStatusAndDetails(
       }
       String statusDetail = statusDetailStringBuilder.toString();
       if (allInstancesCompleted) {
+        LOGGER.info(
```
gaojieliu, line 173: Just want to call it out here is that during the feature rollout, the push job might fail as follows:
1. All the nodes are emitting partition-level status.
2. During the deployment, a few nodes will stop emitting the partition-level status, but only emit version-level status.
3. This logic will check both version-level report and partition-level report, even all the nodes finish the ingestion, the partition-level status aggregation will fail as it will detect a list of offline nodes (partition-level report stops because of the deployment in the deployed nodes).

Ideally, we should ignore the nodes , which are emitting version-level status in partition-level status aggregation.

Does ^ make sense?
cc @sixpluszero

huangminchn: @gaojieliu Good catch! Updated the diff to address this edge case.

gaojieliu (commented):
Code change looks good.
It is hard to cover all the edge cases, such as the application would roll forward and rollback, which means the nodes can reporting version-level status and then report partition-level after the rollback, but it is fine as DVC push status is still a best effort and it is not necessary to make the logic very complex to cover all the edge cases.

Can you add some tests for the code change?

huangminchn:
> Code change looks good. It is hard to cover all the edge cases, such as the application would roll forward and rollback, which means the nodes can reporting version-level status and then report partition-level after the rollback, but it is fine as DVC push status is still a best effort and it is not necessary to make the logic very complex to cover all the edge cases.
> 
> Can you add some tests for the code change?

Yeah, I thought about the rollback case too. To cover that, refactoring would be inevitable, which could be too complicated. We can ask for pausing inc pushes during rollback, or rerunning inc pushes after rollback completes.

Added unit test in the latest diff.

gaojieliu (approved):
Thanks!

PR #1220, opened 2024-10-05
[controller] fix topic priority determination in TopicCleanupService

arjun4084346:

The comparator used in TopicCleanupService to assign higher priority to RT topics over non RT topics `(s1, s2) -> s1.isRealTime() ? -1 : 0` is incorrect. It does not assign two RT topics (or two non RT topics) as equal priority, which violates the fundamental contract of sorting algorithms.
In some cases, this results in RT topic being assigned lower priority than a non RT topic. This PR fixes the comparator and uses both the topics in determining the relative priority.

Resolves #XXX

## How was this PR tested?

added a unit test in PubSubTopicPartitionInfoTest, which fails without the fix.

### DIFF
--- a/internal/venice-common/src/test/java/com/linkedin/venice/pubsub/PubSubTopicPartitionInfoTest.java
+++ b/internal/venice-common/src/test/java/com/linkedin/venice/pubsub/PubSubTopicPartitionInfoTest.java
@@ -3,7 +3,12 @@ package com.linkedin.venice.pubsub;
 import static org.testng.Assert.assertEquals;
 import static org.testng.Assert.assertFalse;
 
+import com.linkedin.venice.controller.kafka.TopicCleanupService;
+import com.linkedin.venice.meta.Version;
 import com.linkedin.venice.pubsub.api.PubSubTopic;
+import java.util.PriorityQueue;
+import java.util.Random;
+import org.testng.Assert;
 import org.testng.annotations.Test;
 
 
@@ -11,6 +16,8 @@ import org.testng.annotations.Test;
  * Unit tests for {@link PubSubTopicPartitionInfo}.
  */
 public class PubSubTopicPartitionInfoTest {
+  Random rand = new Random();
+
   @Test
   public void testTopic() {
     PubSubTopic pubSubTopic = new PubSubTopicImpl("testTopic");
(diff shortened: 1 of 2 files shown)


internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestHybridStoreDeletion.java
```diff
@@ -124,7 +124,7 @@ public void testHybridStoreRTDeletionWhileIngesting() {
 
       createStoresAndVersions(storeNames, streamingRewindSeconds, streamingMessageLag);
 
-      // Wait until the rt topic of the first store is fully deleted.
+      // Wait until the rt topic of the first store is fully created.
```
arjun4084346, line 127: this file's changes are unrelated to the fix.

internal/venice-common/src/test/java/com/linkedin/venice/pubsub/PubSubTopicPartitionInfoTest.java
```diff
@@ -36,4 +40,31 @@ public void testToString() {
...
+    PubSubTopic topic3 = new PubSubTopicImpl("topic3" + Version.REAL_TIME_TOPIC_SUFFIX);
+    PubSubTopic topic4 = new PubSubTopicImpl("topic4");
+    PubSubTopic topic5 = new PubSubTopicImpl("topic5");
+    PubSubTopic topic6 = new PubSubTopicImpl("topic6" + Version.REAL_TIME_TOPIC_SUFFIX);
+    allTopics.add(topic1);
+    allTopics.add(topic2);
+    allTopics.add(topic3);
+    allTopics.add(topic4);
+    allTopics.add(topic5);
+    allTopics.add(topic6);
```
lluwm, line 59: Is it possible to add some `randomness` into this testing? For example, let's say, in a for loop, we generate real time and version topics based on probability, and count their appearances respectively. In the verification phase, we poll and assert them.

lluwm (approved):
LGTM! Thanks @arjun4084346 (congrats on your first Venice PR)

KaiSernLim (approved):

PR #1226, opened 2024-10-08
[controller] Make PS3 for incremental push status reads config cluster-specific; query PS3 only if it is enabled for user store

sushantmane:
## Make inc-push status read from ps3 per cluster config and query push status store only if it's enabled for use store

- Enable incremental push status reads from PS3 based on cluster
  configuration.  
- Query the push status store only if it’s enabled for the specific user store;
  otherwise, fallback to Zookeeper for status reads.  
- Ensure removal of removeFromSupposedlyOngoingIncrementalPushVersions status from
  PS3 when PS3 is enabled, regardless of read configuration.

### DIFF
--- a/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
@@ -6071,23 +6071,22 @@ public class VeniceHelixAdmin implements Admin, StoreCleaner {
       Store store) {
     HelixCustomizedViewOfflinePushRepository cvRepo =
         getHelixVeniceClusterResources(clusterName).getCustomizedViewRepository();
-
-    if (multiClusterConfigs.getControllerConfig(clusterName).usePushStatusStoreForIncrementalPush()
-        && store.isDaVinciPushStatusStoreEnabled()) {
-      return monitor.getIncrementalPushStatusFromPushStatusStore(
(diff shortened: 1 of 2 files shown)


services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -6069,17 +6067,28 @@ private ExecutionStatusWithDetails getIncrementalPushStatus(
...
+          cvRepo,
+          getPushStatusStoreReader());
     }
-    return monitor.getIncrementalPushStatusFromPushStatusStore(
-        kafkaTopic,
-        incrementalPushVersion,
-        cvRepo,
-        getPushStatusStoreReader());
+
+    if (multiClusterConfigs.getControllerConfig(clusterName).usePushStatusStoreForIncrementalPush()) {
```
sixpluszero, line 6084: nit: we can merge that into above check, which to me looks more straightforward. but it is just a suggestion.

sixpluszero (dismissed):
Thanks for the quick fix!

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
sushantmane: Done

sixpluszero (approved):

PR #1228, opened 2024-10-09
[controller] Stop writing writing ready-to-serve instances info in meta system store

sushantmane:
##  Stop writing writing ready-to-serve instances info in meta system store

When materializing the meta store, we produce a snapshot of
readyToServeInstances for all user store versions. If a version is in an ERROR
state, its Helix resources may not exist. Attempting to check these resources
can throw exceptions, potentially causing the addVersion process to fail. 

The info about RTS instances in meta store was required in the past but it is no  
longer used by any clients so it's safe to stop writing this into meta system store.

### DIFF
--- a/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
@@ -7964,6 +7964,9 @@ public class VeniceHelixAdmin implements Admin, StoreCleaner {
       return;
     }
     for (Version version: versions) {
+      if (version.getStatus() == ERROR) {
+        continue;
+      }
       int versionNumber = version.getNumber();
       String topic = Version.composeKafkaTopic(regularStoreName, versionNumber);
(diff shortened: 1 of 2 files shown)


services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -7958,41 +7958,6 @@ void setUpMetaStoreAndMayProduceSnapshot(String clusterName, String regularStore
         "Wrote value schemas to meta system store for venice store: {} in cluster: {}",
         regularStoreName,
         clusterName);
-    // replica status for all the available versions
```
sixpluszero, line 7961: Hmm, just to confirm, FC don't need this information anymore?

sushantmane: That's right. FC does not need  meta stores anymore

sixpluszero (approved):

sushantmane:
Thanks for the review, @sixpluszero!

PR #1229, opened 2024-10-09
[server] Add ability to configure Helix InstanceOperation to UNKNOWN

kvargha:

For a participant to auto-register with Helix without causing a rebalance every time a new participant joins the cluster (i.e. during deployment), we need to set the instance operation to UNKNOWN. Then these participants would be ENABLED in a batch, so it only rebalances once.

## How was this PR tested?

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/config/VeniceServerConfig.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/config/VeniceServerConfig.java
@@ -1104,7 +1104,7 @@ public class VeniceServerConfig extends VeniceClusterConfig {
     return nettyWorkerThreadCount;
   }
 
-  public boolean getHelixJoinAsUnknown() {
+  public boolean isHelixJoinAsUnknownEnabled() {
     return helixJoinAsUnknown;
   }
 
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/config/VeniceServerConfig.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/config/VeniceServerConfig.java
@@ -84,6 +84,7 @@ import static com.linkedin.venice.ConfigKeys.SERVER_ENABLE_LIVE_CONFIG_BASED_KAF
 import static com.linkedin.venice.ConfigKeys.SERVER_ENABLE_PARALLEL_BATCH_GET;
 import static com.linkedin.venice.ConfigKeys.SERVER_FORKED_PROCESS_JVM_ARGUMENT_LIST;
 import static com.linkedin.venice.ConfigKeys.SERVER_GLOBAL_RT_DIV_ENABLED;
+import static com.linkedin.venice.ConfigKeys.SERVER_HELIX_JOIN_AS_UNKNOWN;
 import static com.linkedin.venice.ConfigKeys.SERVER_HTTP2_HEADER_TABLE_SIZE;
 import static com.linkedin.venice.ConfigKeys.SERVER_HTTP2_INBOUND_ENABLED;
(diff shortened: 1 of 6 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/HelixParticipationService.java
```diff
@@ -158,13 +161,33 @@ private ThreadPoolExecutor initHelixStateTransitionThreadPool(int size, String t
     return helixStateTransitionThreadPool;
   }
 
+  public HelixManagerProperty buildHelixManagerProperty(VeniceServerConfig config) {
+    InstanceConstants.InstanceOperation instanceOperation;
+    if (config.getHelixJoinAsUnknown()) {
+      instanceOperation = InstanceConstants.InstanceOperation.UNKNOWN;
```
majisourav99, line 167: can we start as disabled?

lluwm: From the following javadoc, when a node joins in a UNKOWN state, seems that it drops all its current assigned replicas, it sounds a good fit for a new host, but for an existing going through a deployment, is this a desired behavior or there is something else?
    
    /**
     * Behavior: Node will have all of its replicas dropped immediately and will be removed from the RoutingTableProvider.
     * Final State: Node will not hold replicas, be considered for assignment, or be populated in the RoutingTableProvider.
     */
    UNKNOWN

kvargha: Will follow up with Helix

clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/HelixParticipationService.java
```diff
@@ -158,13 +161,33 @@ private ThreadPoolExecutor initHelixStateTransitionThreadPool(int size, String t
...
+    } else {
+      instanceOperation = InstanceConstants.InstanceOperation.ENABLE;
+    }
```
nisargthakkar, line 170: When not specified explicitly, can we skip setting any value and keep the code same as current?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/HelixParticipationService.java
kvargha: After talking with Helix, if an InstanceConfig already exists for a participant this change won't override it. So this UNKNOWN will only be reflected on new participants.

lluwm: Do you think it's appropriate to add that as a comment in the code to clarify the `UNKNOWN` behavior or it's a bit too LI internal?

kvargha: I think it would be a good idea to add a comment here. I'll try to make it as general as possible.

lluwm (approved):
I think it looks good to me. Thanks @kvargha !

clients/da-vinci-client/src/main/java/com/linkedin/davinci/config/VeniceServerConfig.java
```diff
@@ -1101,6 +1104,10 @@ public int getNettyWorkerThreadCount() {
     return nettyWorkerThreadCount;
   }
 
+  public boolean getHelixJoinAsUnknown() {
```
majisourav99, line 1107: extreme nit: boolean config API are generally called `isXYXEnabled`

majisourav99 (approved):
Thanks @kvargha !

nisargthakkar (approved):
LGTM. Thanks

PR #1230, opened 2024-10-09
[changelog] Add message compaction as an opt in feature to changelog consumer

ZacAttack:
## [changelog] Add message compaction as an opt in feature to changelog consumer

This introduces a configuration which will compact down data returned from poll to only contain the latest records for a given key.  This also maintains the order of results returned.

Resolves #XXX


clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumerImpl.java
```diff
@@ -620,6 +621,21 @@ protected Collection<PubSubMessage<K, ChangeEvent<V>, VeniceChangeCoordinate>> i
...
     }
+    if (changelogClientConfig.shouldCompactMessages()) {
+      Map<K, PubSubMessage<K, ChangeEvent<V>, VeniceChangeCoordinate>> tempMap = new LinkedHashMap<>();
+      // The behavior of LinkedHashMap is such that it maintains the order of insertion, but for values which are
+      // replaced,
+      // it's put in at the position of the first insertion. This isn't quite what we want, we want to keep only
+      // a single key (just as a map would), but we want to keep the position of the last insertion as well. So in order
+      // to do that, we remove the entry before inserting it.
+      for (PubSubMessage<K, ChangeEvent<V>, VeniceChangeCoordinate> message: pubSubMessages) {
+        if (tempMap.containsKey(message.getKey())) {
```
arjun4084346, line 632: In your test case, `K` is `String`, which is comparable as expected.
But in this file, key is `byte[]`. byte[] arrays are compared based on reference equality rather than content equality.

KaiSernLim (approved):

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumerImpl.java
ZacAttack: That's an interesting point. The type of message.getKey will be parameterized by the type given in the client...  So it would imply that that type needs to be something that implements equals(), which probably won't be universally true.  I know in java we have Arrays.equals() for this sort of thing, but then we'd have to peek under the covers a bit.....  I'm not totally sure how to overcome this

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumerImpl.java
```diff
@@ -620,6 +621,21 @@ protected Collection<PubSubMessage<K, ChangeEvent<V>, VeniceChangeCoordinate>> i
...
       changeCaptureStats.recordRecordsConsumed(pubSubMessages.size());
     }
+    if (changelogClientConfig.shouldCompactMessages()) {
+      Map<K, PubSubMessage<K, ChangeEvent<V>, VeniceChangeCoordinate>> tempMap = new LinkedHashMap<>();
+      // The behavior of LinkedHashMap is such that it maintains the order of insertion, but for values which are
+      // replaced,
+      // it's put in at the position of the first insertion. This isn't quite what we want, we want to keep only
+      // a single key (just as a map would), but we want to keep the position of the last insertion as well. So in order
+      // to do that, we remove the entry before inserting it.
+      for (PubSubMessage<K, ChangeEvent<V>, VeniceChangeCoordinate> message: pubSubMessages) {
```
xunyin8, line 631: IIUC `pubSubMessages` could contain control messages if `includeControlMessage == true`. Could this break anything for CC client? I know heartbeat keys are equal but compacting heartbeat should is fine. What about other control messages?

ZacAttack: I wasn't able to think of a bad scenario.  Fortunately, the include control messages feature isn't really getting used by any customers yet, it's more used for internal interfaces right now.  I think we're good here for now.

PR #1233, opened 2024-10-10
[test] Make sure meta store ready before DVC memory limiter test

majisourav99:
##  Make sure meta store ready before DVC memory limiter test

Trying to deflake `testDaVinciMemoryLimitShouldFailLargeDataPush` 


internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/DaVinciClientMemoryLimitTest.java
```diff
@@ -257,6 +260,7 @@ public void testDaVinciMemoryLimitShouldFailLargeDataPush(
     }
   }
 
+  @Test(enabled = false)
```
sushantmane, line 263: Just to be sure can run this test 50 times and see if the fix is good

majisourav99: I ran the test without the fix, 5 iteration passed without any issue. IMO there is some resource limitation/constraint on GHE host which steadily repro the time. I dont think just running multiple times would capture that

sushantmane (approved):

PR #1234, opened 2024-10-10
[controller] do not delay RT topic deletion if the store is no longer hybrid

arjun4084346:

When TopicCleanupService tries to delete RT topics, it does not have to wait for all the corresponding VT topics to get deleted if all the versions of the store has changed from hybrid to batch. If all the versions are batch, it can be safely assumed that no VT topic is consuming from that RT topic.

Resolves #XXX

## How was this PR tested?

added a unit test in TestTopicCleanupService


services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/TopicCleanupService.java
sushantmane, line 1: We want to check 'version' level configs and not just store level config

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/TopicCleanupService.java
```diff
@@ -253,14 +253,24 @@ void cleanupVeniceTopics() {
...
           for (Map.Entry<String, Map<String, Integer>> mapEntry: multiDataCenterStoreToVersionTopicCount.entrySet()) {
-            if (mapEntry.getValue().containsKey(topic.getStoreName())) {
-              canDelete = false;
-              LOGGER.info(
-                  "Topic deletion for topic: {} is delayed due to {} version topics found in datacenter {}",
-                  topic.getName(),
-                  mapEntry.getValue().get(topic.getStoreName()),
-                  mapEntry.getKey());
-              break;
+            String storeName = topic.getStoreName();
```
sushantmane, line 256: This will only check local region's state. We need to check hybrid versions in other colos as well. See if you can leverage https://github.com/linkedin/venice/blob/14d05d439b404df2a6d7957f229be941899f7cb3/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java#L3526

sushantmane (commented):
Can you also add E2E test that verifies RT topic marked for deletion is 
1. it is not deleted when there are hybrid versions 
2. deleted once all hybrid versions are removed

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -3523,17 +3522,26 @@ private boolean hasFatalDataValidationError(PushMonitor pushMonitor, String topi
...
+  public boolean canDeleteRTTopic(String clusterName, String storeName, Store store) {
     // Perform RT cleanup checks for batch only store that used to be hybrid. Check the local versions first
     // to see if any version is still using RT and then also check other fabrics before deleting the RT. Since
     // we perform this check everytime when a store version is deleted we can afford to do best effort
     // approach if some fabrics are unavailable or out of sync (temporarily).
-    boolean canDeleteRT = !Version.containsHybridVersion(store.getVersions());
+    return !Version.containsHybridVersion(store.getVersions()) && canDeleteRTTopic(clusterName, storeName);
+  }
+
+  public boolean canDeleteRTTopic(String clusterName, String storeName) {
```
lluwm, line 3542: Can we rename the function somehow to reflect the fact that  we are querying the child controllers for their opinions?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -3542,24 +3550,29 @@ private void safeDeleteRTTopic(String clusterName, String storeName, Store store
...
+    return true;
+  }
+
+  private void deleteRTTopic(String clusterName, String storeName) {
+    Map<String, ControllerClient> controllerClientMap = getControllerClientMap(clusterName);
+    String rtTopicToDelete = Version.composeRealTimeTopic(storeName);
+    truncateKafkaTopic(rtTopicToDelete);
+    for (ControllerClient controllerClient: controllerClientMap.values()) {
+      controllerClient.deleteKafkaTopic(rtTopicToDelete);
+    }
```
lluwm, line 3568: these lines are duplicated later. Can we extract them into a small helper function?

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/TopicCleanupService.java
```diff
@@ -135,10 +122,26 @@ public TopicCleanupService(
     } else {
       this.sourceOfTruthPubSubAdminAdapter = null;
     }
+
+    if (!this.admin.isParent()) {
```
lluwm, line 126: is keyword `this` here necessary? My understanding for our code convention is that `this` is used when there is ambiguity or cases it has to be used.

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/TopicCleanupService.java
```diff
@@ -235,49 +238,46 @@ Admin getAdmin() {
    * If version topic deletion takes more than certain time it refreshes the entire topic list and start deleting from RT topics again.
     */
   void cleanupVeniceTopics() {
+    // todo - instead of adding all topics into the same queue we must simply use separate queues for RT and nonRT
```
lluwm, line 241: nite, todo -> "TODO" so that IDE can highlight it.

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/TopicCleanupService.java
lluwm: What's main reason to move the child controller api calls into constructor?

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/TopicCleanupService.java
arjun4084346: in my IDE, small case todo also highlights

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/TopicCleanupService.java
arjun4084346: the reason is that the value of `localDatacenter` does not change so we do not need to re-calculate it everytime topics are deleted. cc @xunyin8 correct me if I missed anything

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/TopicCleanupService.java
```diff
@@ -235,49 +238,48 @@ Admin getAdmin() {
...
+            topic.getName(),
+            e.toString());
       } catch (VeniceException e) {
         LOGGER.warn("Caught exception when trying to delete topic: {} - {}", topic.getName(), e.toString());
         topicCleanupServiceStats.recordTopicDeletionError();
         // No op, will try again in the next cleanup cycle.
       }
-
+      topicCleanupServiceStats.recordTopicDeleted();
+      getTopicManager().ensureTopicIsDeletedAndBlockWithRetry(topic);
```
lluwm, line 278: is this right to move it out of the try-catch block and into the common path. This means that when exception happens, topic deleted will still be reported?

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/controller/TestParentControllerWithMultiDataCenter.java
```diff
@@ -93,6 +96,91 @@ public void cleanUp() {
...
+     todo - RT topic is not deleted in parent colo yet.
+     we need to probably modify TopicCleanupServiceForParentController and check if all the child colos have deleted
+     RT topic, then it can also be deleted from parent colo. I believe we can extend it to non RT topics as well and
+     then can get rid of `storeToCountdownForDeletion`
+    
+    TestUtils.waitForNonDeterministicAssertion(60, TimeUnit.SECONDS, true,  true, () -> {
+      Assert.assertFalse(parentTopicManager.containsTopic(rtPubSubTopic));
+      }
+     );
+    */
```
lluwm, line 181: What's your thought on the comment here, do we want to add this functionality in this change or a separate one?

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/TopicCleanupService.java
```diff
@@ -235,40 +239,48 @@ Admin getAdmin() {
...
+          clusterDiscovered = admin.discoverCluster(storeName).getFirst();
+        } catch (VeniceNoStoreException e) {
+          LOGGER.warn("Store {} is not found", storeName);
+          getTopicManager().ensureTopicIsDeletedAndBlockWithRetry(topic);
+          topicCleanupServiceStats.recordTopicDeleted();
+          continue;
+        }
+        Store store = admin.getStore(clusterDiscovered, storeName);
+        if (topic.isRealTime() && (this.isRTTopicDeletionBlocked
+            || !admin.isRTTopicDeletionPermittedByAllControllers(clusterDiscovered, store))) {
```
lluwm, line 264: In our old implementation, if a topic cannot be deleted (canDelete = false), it will continue and bypass the code from L278 - L289, but in the new version, the bypassing logic is missing. is this on purpose?

lluwm (approved):
Thanks @arjun4084346 for this fix. LGTM!

PR #1235, opened 2024-10-11
[changelog][dvc] Disable TC internal retry in ThinClientMetaStoreBasedRepository

xunyin8:
## [changelog][dvc] Disable TC internal retry in ThinClientMetaStoreBasedRepository

1. The existing TC internal retry implemented in RetriableStoreClient is vulnerable to a dead lock issue due to deserialization thread performing and waiting on the retry. Once the thread pool is exhausted all deserialization threads will be waiting forever because there won't be any threads left to handle the transport response. We will fix this internal retry in another commit.

2. Disable the TC internal retry in ThinClientMetaStoreBasedRepository because we recently added external retry with RetryUtils.executeWithMaxAttempt and timeout. Remove hardcoded retry configs from NativeMetadataRepository since it's no longer needed and it was leaking implementation details.

3. Added staleness metric in NativeMetadataRepository that should help us detect any issue causing the local metadata to be stale (including the dead lock issue mentioned above) for both DVC and change log use cases.

## How was this PR tested?
New and existing tests

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/repository/NativeMetadataRepository.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/repository/NativeMetadataRepository.java
@@ -5,6 +5,7 @@ import static com.linkedin.venice.system.store.MetaStoreWriter.KEY_STRING_SCHEMA
 import static com.linkedin.venice.system.store.MetaStoreWriter.KEY_STRING_STORE_NAME;
 import static java.lang.Thread.currentThread;
 
+import com.linkedin.davinci.stats.NativeMetadataRepositoryStats;
 import com.linkedin.venice.client.exceptions.ServiceDiscoveryException;
 import com.linkedin.venice.client.store.ClientConfig;
 import com.linkedin.venice.common.VeniceSystemStoreType;
@@ -30,6 +31,7 @@ import com.linkedin.venice.systemstore.schemas.StoreMetaKey;
 import com.linkedin.venice.systemstore.schemas.StoreMetaValue;
 import com.linkedin.venice.utils.VeniceProperties;
 import com.linkedin.venice.utils.concurrent.VeniceConcurrentHashMap;
+import java.time.Clock;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
@@ -57,9 +59,6 @@ import org.apache.logging.log4j.Logger;
  */
 public abstract class NativeMetadataRepository
     implements SubscriptionBasedReadOnlyStoreRepository, ReadOnlySchemaRepository, ClusterInfoProvider {
-  protected static final int THIN_CLIENT_RETRY_COUNT = 3;
-  protected static final long THIN_CLIENT_RETRY_BACKOFF_MS = 10000;
-
   private static final long DEFAULT_REFRESH_INTERVAL_IN_SECONDS = 60;
   private static final Logger LOGGER = LogManager.getLogger(NativeMetadataRepository.class);
 
@@ -78,13 +77,23 @@ public abstract class NativeMetadataRepository
 
   private final long refreshIntervalInSeconds;
 
+  private final NativeMetadataRepositoryStats nativeMetadataRepositoryStats;
+
+  private final Clock clock;
   private AtomicBoolean started = new AtomicBoolean(false);
 
   protected NativeMetadataRepository(ClientConfig clientConfig, VeniceProperties backendConfig) {
+    this(clientConfig, backendConfig, Clock.systemUTC());
+  }
+
+  protected NativeMetadataRepository(ClientConfig clientConfig, VeniceProperties backendConfig, Clock clock) {
     refreshIntervalInSeconds = backendConfig.getLong(
         CLIENT_SYSTEM_STORE_REPOSITORY_REFRESH_INTERVAL_SECONDS,
         NativeMetadataRepository.DEFAULT_REFRESH_INTERVAL_IN_SECONDS);
     this.clientConfig = clientConfig;
+    this.nativeMetadataRepositoryStats =
+        new NativeMetadataRepositoryStats(clientConfig.getMetricsRepository(), "native_metadata_repository", clock);
+    this.clock = clock;
   }
 
   public synchronized void start() {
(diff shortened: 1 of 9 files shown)


ZacAttack (approved):
Looks good to me, but I think some of your the tests are belly aching about something.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/NativeMetadataRepositoryStats.java
```diff
@@ -0,0 +1,42 @@
...
+    this.storeMetadataStalenessSensor = registerSensor(new AsyncGauge((ignored1, ignored2) -> {
+      if (this.metadataCacheTimestampMapInMs.isEmpty()) {
+        return Double.NaN;
+      } else {
+        Iterator<Long> iterator = metadataCacheTimestampMapInMs.values().iterator();
+        long oldest = iterator.next();
+        while (iterator.hasNext()) {
+          oldest = Math.min(oldest, iterator.next());
+        }
+        return clock.millis() - oldest;
```
mynameborat, line 30: nit: Personally would simplify into 
```
metdataCacheTimestampmapsInMs.values().stream().min(Long::compareTo)
  .map(oldest -> clock.millis() - oldest)
  .orElse(Double.Nan);
```

clients/da-vinci-client/src/main/java/com/linkedin/davinci/repository/ThinClientMetaStoreBasedRepository.java
```diff
@@ -112,10 +112,9 @@ private AvroSpecificStoreClient<StoreMetaKey, StoreMetaValue> getAvroClientForMe
       ClientConfig<StoreMetaValue> clonedClientConfig = ClientConfig.cloneConfig(clientConfig)
           .setStoreName(VeniceSystemStoreType.META_STORE.getSystemStoreName(storeName))
           .setSpecificValueClass(StoreMetaValue.class)
-          .setRetryOnAllErrors(true)
-          .setRetryCount(THIN_CLIENT_RETRY_COUNT)
-          .setForceClusterDiscoveryAtStartTime(true)
-          .setRetryBackOffInMs(THIN_CLIENT_RETRY_BACKOFF_MS);
+          .setRetryOnRouterError(false)
+          .setRetryOnAllErrors(false)
+          .setForceClusterDiscoveryAtStartTime(true);
```
mynameborat, line 117: can you add a git issue or JIRA here to follow up on re-enabling the retries?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/repository/NativeMetadataRepository.java
```diff
@@ -78,13 +77,23 @@ public abstract class NativeMetadataRepository
 
   private final long refreshIntervalInSeconds;
 
+  private final NativeMetadataRepositoryStats nativeMetadataRepositoryStats;
```
mynameborat, line 80: Thank you for adding metrics :)

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/NativeMetadataRepositoryStats.java
```diff
@@ -0,0 +1,42 @@
...
+      }
+    }, "store_metadata_staleness_high_watermark_ms"));
```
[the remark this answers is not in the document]
xunyin8: Good call

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/NativeMetadataRepositoryStats.java
```diff
@@ -0,0 +1,42 @@
...
+
+public class NativeMetadataRepositoryStats extends AbstractVeniceStats {
+  private final Sensor storeMetadataStalenessSensor;
+  private final Map<String, Long> metadataCacheTimestampMapInMs = new VeniceConcurrentHashMap<>();
+  private final Clock clock;
+
+  public NativeMetadataRepositoryStats(MetricsRepository metricsRepository, String name, Clock clock) {
+    super(metricsRepository, name);
+    this.clock = clock;
+    this.storeMetadataStalenessSensor = registerSensor(new AsyncGauge((ignored1, ignored2) -> {
```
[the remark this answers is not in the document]
xunyin8: If we use the `now` from the measurable, it might be more difficult to mock the clock during testing. Let me double check.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/repository/NativeMetadataRepository.java
```diff
@@ -172,6 +181,7 @@ public Store refreshOneStore(String storeName) {
       if (newStore != null && !storeConfig.isDeleting()) {
```
majisourav99, line 181: why do we aggresively remove the store in the `else` part here? A metadata fetch call fail would not return null also.

majisourav99: Its not related to this PR, seems like we reconstruct a branch new store on every refresh,  even though we just to look at the store properties only.

xunyin8: Right, the store deletion path is very fragile (another reason to move away from meta system store). In reality it's very possible that the system store is already deleted together with the user store before the NativeMetadataRepository even got a chance to refresh. Meaning the TC will throw exception even earlier on `getAndSetStoreConfigFromSystemStore`. Right now the only case this `else` block can be executed is when the `newStore` object is not null and when `storeConfig.isDeleting()` is true. This is because we will throw exception if the fetch is actually null in `getStoreMetaValue`:
```
    if (value == null) {
      throw new MissingKeyInStoreMetadataException(key.toString(), StoreMetaValue.class.getSimpleName());
    }
```
This ensures we won't accidentally remove store when encountering a null value from TC.

majisourav99 (approved):
LGTM! Thanks Xun

mynameborat (approved):

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/NativeMetadataRepositoryStats.java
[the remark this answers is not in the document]
mynameborat: For closure, can you update the thread and the decision of pursuing the change vs not.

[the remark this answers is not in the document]
xunyin8: Will keep the current behavior. NativeMetadataRepositoryStats is not latency sensitive or invoked very frequently therefore getting `now` through a local clock for easier testing is acceptable IMO.

PR #1239, opened 2024-10-14
[router] More efficient handling of batch get key count limit breach

FelixGV:
- In VeniceMultiKeyPath::initialize, we check the key count sooner, which is possible thanks to an earlier refactoring to the Avro deserialization code which now returns a List rather than an Iterable.

- In VenicePathParser::parseResourceUri, when the batch get key count is exceeded, we mark it as a 413, and then treat it as an "expected exception" in RouterExceptionAndTrackingUtils, which suppresses the stacktrace from the logs.

- In RouterExceptionAndTrackingUtils, we further improve the handling of "expected exceptions" so that not only we avoid logging the stacktrace, but we also skip filling it in in the first place. Previously, the stacktrace would be filled in and then suppressed, and filling in stacktraces can be expensive. This change should help not only batch get key count limit breaches but also other forms of quota breaches, store not found, etc.

Miscellaneous:

- Tweaked the logging of the build which had a bookkeeping error in the time taken to run each test class.

- Added a null check in D2ClientUtils::shutdownClient since it caused some integration tests to fail during their tear down.

## How was this PR tested?
GHCI.

Clients will receive the 413 status code now instead of 400 if they breach the batch get key count limit.

### DIFF
--- a/internal/venice-client-common/src/main/java/com/linkedin/venice/D2/D2ClientUtils.java
+++ b/internal/venice-client-common/src/main/java/com/linkedin/venice/D2/D2ClientUtils.java
@@ -62,6 +62,9 @@ public class D2ClientUtils {
   }
 
   public static void shutdownClient(D2Client client, long timeoutInMs) {
+    if (client == null) {
+      return;
+    }
     long startTime = System.currentTimeMillis();
     CompletableFuture<Void> future = new CompletableFuture<>();
 
--- a/services/venice-router/src/main/java/com/linkedin/venice/router/api/RouterExceptionAndTrackingUtils.java
+++ b/services/venice-router/src/main/java/com/linkedin/venice/router/api/RouterExceptionAndTrackingUtils.java
@@ -37,9 +37,6 @@ public class RouterExceptionAndTrackingUtils {
 
   private static final Logger LOGGER = LogManager.getLogger(RouterExceptionAndTrackingUtils.class);
 
-  /** We do not fill in the stacktrace at all for "expected exceptions" (quota, etc) */
-  private static final boolean DO_NOT_FILL_IN_STACKTRACE = false;
-
   private static final RedundantExceptionFilter EXCEPTION_FILTER =
       RedundantExceptionFilter.getRedundantExceptionFilter();
 
@@ -63,7 +60,8 @@ public class RouterExceptionAndTrackingUtils {
             false,
             null,
             true,
-            DO_NOT_FILL_IN_STACKTRACE)
+            /** We do not fill in the stacktrace at all for "expected exceptions" (quota, etc) */
+            false)
         : new RouterException(HttpResponseStatus.class, responseStatus, responseStatus.code(), msg, false);
     String name = storeName.isPresent() ? storeName.get() : "";
     if (!EXCEPTION_FILTER.isRedundantException(name, String.valueOf(e.code()))) {
@@ -119,7 +117,7 @@ public class RouterExceptionAndTrackingUtils {
     String name = storeName.isPresent() ? storeName.get() : "";
     VeniceException e = isExpected(responseStatus, failureType)
         // Do not dump stack-trace for Quota exceed exception as it might blow up memory on high load
-        ? new VeniceException(msg, DO_NOT_FILL_IN_STACKTRACE)
+        ? new VeniceException(msg, false)
         : new VeniceException(msg);
 
     if (!EXCEPTION_FILTER.isRedundantException(name, e)) {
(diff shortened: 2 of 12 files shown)


services/venice-router/src/main/java/com/linkedin/venice/router/api/path/VeniceMultiKeyPath.java
```diff
@@ -107,25 +108,7 @@ public void initialize(
...
-      routerKey.setPartitionId(partitionId);
-      K routerRequestKey = createRouterRequestKey(key, keyIdx, partitionId);
-      this.routerKeyMap.put(routerKey, routerRequestKey);
-      ++keyIdx;
-    }
-    setPartitionKeys(this.routerKeyMap.keySet());
-
-    int keyCount = getPartitionKeys().size();
+    int keyCount = keys.size();
     if (keyCount > maxKeyCount) {
```
majisourav99, line 112: Curious, how does the request get pass the client key count check ?

services/venice-router/src/main/java/com/linkedin/venice/router/api/RouterExceptionAndTrackingUtils.java
```diff
@@ -30,12 +33,13 @@ public enum FailureType {
...
   }
 
-  private static final StackTraceElement[] emptyStackTrace = new StackTraceElement[0];
-
   private static RouterStats<AggRouterHttpRequestStats> ROUTER_STATS;
 
   private static final Logger LOGGER = LogManager.getLogger(RouterExceptionAndTrackingUtils.class);
 
+  /** We do not fill in the stacktrace at all for "expected exceptions" (quota, etc) */
+  private static final boolean DO_NOT_FILL_IN_STACKTRACE = false;
```
majisourav99, line 41: Kinda do not like the final `false` boolean.  How about create a new VeniceExceptionNoStackTrace() and call that instead of passing a final false boolean?

internal/venice-client-common/src/main/java/com/linkedin/venice/D2/D2ClientUtils.java
```diff
@@ -62,6 +62,9 @@ public static void shutdownClient(D2Client client) {
   }
 
   public static void shutdownClient(D2Client client, long timeoutInMs) {
+    if (client == null) {
```
majisourav99, line 65: why do we call shutdown on a null client?

services/venice-router/src/main/java/com/linkedin/venice/router/api/path/VeniceMultiKeyPath.java
FelixGV: I think that check is only implemented in (recent versions of) the FC and not in the TC? Also, the check is based on some kind of periodically refreshed metadata that could be out of aync with the backend?

services/venice-router/src/main/java/com/linkedin/venice/router/api/RouterExceptionAndTrackingUtils.java
FelixGV: Hmm... that could work in some cases but seems less flexible? Java does not have multiple inheritance, so if a subclass of `VeniceException` wants to achieve this behavior, they would still need the constructor param...?

internal/venice-client-common/src/main/java/com/linkedin/venice/D2/D2ClientUtils.java
FelixGV: It happens in cases where the setup of a test has multiple steps, some early step fails and a latter step of spinning up a d2 client gets short-circuited, and then the cleanup function of the test still runs, but passes null into this function.

services/venice-router/src/main/java/com/linkedin/venice/router/api/RouterExceptionAndTrackingUtils.java
FelixGV: But I can get rid of the final variable and just pass false directly at the call site, if you prefer...?

majisourav99: Yeah that would work too.

services/venice-router/src/main/java/com/linkedin/venice/router/api/path/VeniceMultiKeyPath.java
majisourav99: Right, I see the check only in FC but not in TC. Should we add the check in TC as well ? it will save lots of resources for router and client as well for large requests.

FelixGV: At first glance I would rather not... it seems to be "not very thin" 😅 ... but we can consider it.

services/venice-router/src/main/java/com/linkedin/venice/router/api/RouterExceptionAndTrackingUtils.java
FelixGV: Done.

majisourav99 (approved):
LGTM. Thanks Felix for perf improvement!

PR #1240, opened 2024-10-15
[server][dvc] support hybrid store in blob transfer with recreating snapshots

jingy-li:
## [server][dvc] support hybrid store in blob transfer with recreating snapshots

Blob transfer currently does not support hybrid stores because snapshot generation occurs only after Kafka ingestion is complete in batch store, which can result in outdated snapshots for hybrid stores.

To resolve this issue, this PR will:

1. Regenerate a new snapshot if snapshot is staled when a client sends a GET request.
2. Ensure that the transferred offset record accurately reflects the one immediately preceding the snapshot's recreation. If the snapshot manager generates the snapshot first and uses the current offset record for the client, the most recent/largest offset record may be excluded from the snapshot.
3. Throttle concurrent users. The blob snapshot manager will maintain a concurrentSnapshotUsers map to limit the number of hosts that can initiate a snapshot simultaneously for each topic and partition. If too many hosts attempt to request a snapshot for the same topic and partition, the server will respond with a 404 error. The maximum number of allowed concurrent users will be controlled by config.
4. Manage snapshot timestamps by implementing a config-controlled snapshot retention time to ensure the freshness of hybrid snapshots.

## How was this PR tested?

unit test.
[WIP] integration test

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/BlobSnapshotManager.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/BlobSnapshotManager.java
@@ -119,10 +119,9 @@ public class BlobSnapshotManager {
 
     // check if the concurrent user count exceeds the limit
     checkIfConcurrentUserExceedsLimit(topicName, partitionId);
-    concurrentSnapshotUsers.putIfAbsent(topicName, new VeniceConcurrentHashMap<>());
-    concurrentSnapshotUsers.get(topicName).putIfAbsent(partitionId, new AtomicLong(0));
-
-    concurrentSnapshotUsers.get(topicName).get(partitionId).incrementAndGet();
+    concurrentSnapshotUsers.computeIfAbsent(topicName, k -> new VeniceConcurrentHashMap<>())
+        .computeIfAbsent(partitionId, k -> new AtomicLong(0))
+        .incrementAndGet();
 
     boolean isHybrid = isStoreHybrid(payload.getStoreName());
     if (!isHybrid) {
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/BlobSnapshotManager.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/BlobSnapshotManager.java
@@ -1,15 +1,17 @@
 package com.linkedin.davinci.blobtransfer;
 
-import com.google.common.annotations.VisibleForTesting;
+import com.linkedin.davinci.storage.StorageEngineRepository;
+import com.linkedin.davinci.store.AbstractStorageEngine;
+import com.linkedin.davinci.store.AbstractStoragePartition;
 import com.linkedin.venice.exceptions.VeniceException;
 import com.linkedin.venice.meta.ReadOnlyStoreRepository;
 import com.linkedin.venice.meta.Store;
-import com.linkedin.venice.store.rocksdb.RocksDBUtils;
 import com.linkedin.venice.utils.concurrent.VeniceConcurrentHashMap;
 import com.linkedin.venice.utils.locks.AutoCloseableLock;
 import java.io.File;
 import java.io.IOException;
 import java.util.Map;
+import java.util.Objects;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
@@ -28,11 +30,11 @@ import org.rocksdb.RocksDBException;
  */
 
 public class BlobSnapshotManager {
+  // A map to keep track of the number of hosts using a snapshot for a particular topic and partition
   private final Map<String, Map<Integer, AtomicLong>> concurrentSnapshotUsers;
-  private Map<String, Map<Integer, Long>> snapShotTimestamps;
-  private final long snapshotRetentionTime;
-  private final String basePath;
+
   private final ReadOnlyStoreRepository readOnlyStoreRepository;
+  private final StorageEngineRepository storageEngineRepository;
   private final Lock lock = new ReentrantLock();
   private static final Logger LOGGER = LogManager.getLogger(BlobSnapshotManager.class);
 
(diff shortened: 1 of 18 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/server/P2PFileTransferServerHandler.java
```diff
@@ -172,6 +198,51 @@ public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) throws E
...
   }
 
+  /**
+   * Cleanup the snapshot directory if the snapshot is new generated for hybrid mode
+   * Decrease the user count for the snapshot to allow other hosts can use it.
+   * @param newSnapshotGenerated whether the snapshot is generated for hybrid mode
+   * @param snapshotDir the snapshot directory
+   * @param blobTransferRequest the blob transfer request
+   */
+  private void cleanupAndRemoveConcurrentUserRestriction(
```
adamxchen, line 208: I haven't checked the method correctness in terms of the implementation but I think it's better to have this implemented in the blobSnapshotManager so it has some central locks to verify a single partition won't generate snapshots too frequently.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/BlobSnapshotManager.java
```diff
@@ -28,60 +30,65 @@
...
-        if (concurrentSnapshotUsers.get(topicName).get(partitionId) == null
-            || concurrentSnapshotUsers.get(topicName).get(partitionId).get() == 0) {
-          updateHybridSnapshot(rocksDB, topicName, partitionId);
-          snapShotTimestamps.get(topicName).put(partitionId, System.currentTimeMillis());
-        }
+      if (concurrentSnapshotUsers.get(topicName).get(partitionId) == null
+          || concurrentSnapshotUsers.get(topicName).get(partitionId).get() == 0) {
+        createSnapshot(topicName, partitionId);
+      } else {
+        // the snapshot is being used, cannot generate a new snapshot at the same time
```
adamxchen, line 75: why not just re-use the same snapshot?

jingy-li: In hybrid stores, we do not allow multiple hosts to share the same snapshot. When the second GET request is received, the system checks the map and recognizes that one host is already utilizing that snapshot. However, the offset record will send to the second client (obtained before checking the snapshot's existence) may not correspond to the snapshot.

jingy-li: Sync offline, we will (1) restrict the number of concurrent users, (2) allow the use of only non-stale snapshot within the set concurrency limit, and (3) monitor and record snapshot offsets and timestamps, updating them when generating a new snapshot to replace any stale ones.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/server/P2PFileTransferServerHandler.java
```diff
@@ -101,9 +107,24 @@ protected void channelRead0(ChannelHandlerContext ctx, FullHttpRequest httpReque
...
+
     try {
       blobTransferRequest = parseBlobTransferPayload(URI.create(httpRequest.uri()));
       snapshotDir = new File(blobTransferRequest.getSnapshotDir());
+      isHybridStore = blobSnapshotManager.isStoreHybrid(blobTransferRequest.getStoreName());
+
+      // recreate snapshot for hybrid store
+      if (isHybridStore) {
+        try {
+          BlobTransferPartitionMetadata metadataForHybrid = prepareMetadata(blobTransferRequest, ctx);
```
adamxchen, line 120: nitpick on the code style.
We don't need to pass `ctx` to every downstream methods. We can just throw exceptions and catch it in this main method. Once caught, we can `setupResponseAndFlush` accordingly. It can make the codes more clean. As we are adding more if-else conditions, this method will become clumsy and a bit hard to read

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/BlobSnapshotManager.java
```diff
@@ -22,66 +25,142 @@
...
+
     this.concurrentSnapshotUsers = new VeniceConcurrentHashMap<>();
+    this.snapshotTimestamps = new VeniceConcurrentHashMap<>();
+    this.snapshotMetadataRecords = new VeniceConcurrentHashMap<>();
   }
 
   /**
-   * Checks if the snapshot is stale, if it is and no one is using it, it updates the snapshot,
-   * otherwise it increases the count of people using the snapshot
+   * Recreate a snapshot for a hybrid store and return true if the snapshot is successfully created
```
adamxchen, line 69: the method doesn't return true/false anymore

adamxchen (commented):
got some initial comments

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/server/P2PFileTransferServerHandler.java
```diff
@@ -101,9 +107,24 @@ protected void channelRead0(ChannelHandlerContext ctx, FullHttpRequest httpReque
...
+          blobSnapshotManager.recreateSnapshotForHybrid(blobTransferRequest, metadataForHybrid);
```
adamxchen, line 121: per our offline discussion, we want to do a bit of refactoring so `blobSnapshotManager` would handle metadata creation, lifecycle management, snapshot creation etc., so server handler would just call `blobSnapshotManager` to obtain all of these to make codes more readable and easy to maintain.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/server/P2PFileTransferServerHandler.java
```diff
@@ -145,6 +170,8 @@ protected void channelRead0(ChannelHandlerContext ctx, FullHttpRequest httpReque
         LOGGER.error("Failed to send all files for {}", blobTransferRequest.getFullResourceName(), future.cause());
       }
     });
+
+    blobSnapshotManager.removeConcurrentUserRestriction(isHybridStore, blobTransferRequest);
```
adamxchen, line 174: this method should also be a part of `exceptionCaught` or `channelInActive`(which i think it's a better place) to decrease the concurrent user restriction, otherwise if there's some exceptions happen, the count never get decreased.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/BlobSnapshotManager.java
```diff
@@ -22,66 +25,142 @@
...
-    if (rocksDB == null || topicName == null) {
-      throw new IllegalArgumentException("RocksDB instance and topicName cannot be null");
-    }
-    if (!isStoreHybrid(topicName)) {
-      LOGGER.warn("Store {} is not hybrid, skipping snapshot update", topicName);
-      return;
-    }
-    String fullPathForPartitionDBSnapshot = RocksDBUtils.composeSnapshotDir(this.basePath, topicName, partitionId);
-    snapShotTimestamps.putIfAbsent(topicName, new VeniceConcurrentHashMap<>());
+  public void recreateSnapshotForHybrid(
```
adamxchen, line 74: If we make this method also work for batch, I am thinking we can do something like this
```
BlobTransferPartitionMetadata  createSnapshot(BlobTransferPayload payload) {
  if batch:
       generate the metadata
       return it
  else:
     // hybrid
     check if we should throttle the request due to many concurrent users
     check whether we have snapshot & metadata available within the staleness threshold
     if yes:
        get the metadata from map
        return metadata
    else:
       generate the metadata
       generate the snapshot
       return metadata
}
```
You can further streamline the logic if needed

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/BlobSnapshotManager.java
```diff
@@ -22,66 +25,142 @@
...
+      return true;
+    }
+    return System.currentTimeMillis()
+        - snapshotTimestamps.get(topicName).get(partitionId) > SNAPSHOT_RETENTION_TIME_IN_MILLIS;
+  }
+
+  /**
+   * Decrease the count of hosts using the snapshot
+   */
   public void decreaseConcurrentUserCount(String topicName, int partitionId) {
```
adamxchen, line 164: I'm thinking whether we need the lock in
```
    try (AutoCloseableLock ignored = AutoCloseableLock.of(lock)) {
      long result = concurrentUsers.decrementAndGet();
      if (result < 0) {
        throw new VeniceException("Concurrent user count cannot be negative");
      }
```

AtomicLong looks sufficient to me, no?

clients/da-vinci-client/src/test/java/com/linkedin/davinci/blobtransfer/TestP2PFileTransferServerHandler.java
```diff
@@ -277,4 +267,37 @@ public void testTransferMultipleFiles() throws IOException {
     Assert.assertEquals(endOfTransfer.headers().get(BLOB_TRANSFER_STATUS), BLOB_TRANSFER_COMPLETED);
     // end of STATUS response
   }
+
+  @Test
+  public void testWhenMetadataCreateError() throws IOException {
```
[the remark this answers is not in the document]
jingy-li: Yes, it is for testing when the metadata get from storageMetadataService with exception, response to client should show errors. Added in comments.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/BlobSnapshotManager.java
```diff
@@ -24,19 +25,31 @@
...
+  public final static int MAX_CONCURRENT_USERS = 5;
+
+  // A map to keep track of the number of hosts using a snapshot for a particular topic and partition, use to restrict
+  // concurrent user count
+  // Example: <topicName, <partitionId, concurrentUsersAmount>>
+  private Map<String, Map<Integer, AtomicLong>> concurrentSnapshotUsers;
+  // A map to keep track of the snapshot timestamps for a particular topic and partition, use to track snapshot
+  // staleness
+  // Example: <topicName, <partitionId, timestamp>>
+  private Map<String, Map<Integer, Long>> snapshotTimestamps;
```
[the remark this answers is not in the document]
jingy-li: Let's separate them (timestamps and metadata records), as the timestamp is called separately when checking the staleness.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/BlobSnapshotManager.java
```diff
@@ -184,16 +269,94 @@ public static void createSnapshotForBatch(RocksDB rocksDB, String fullPathForPar
...
+  }
+
+  /**
+   * Prepare the metadata for a blob transfer request
+   * @param blobTransferRequest the blob transfer request
+   * @return the metadata for the blob transfer request
+   */
+  public BlobTransferPartitionMetadata prepareMetadata(BlobTransferPayload blobTransferRequest) {
+    // prepare metadata
+    BlobTransferPartitionMetadata metadata = null;
```
adamxchen, line 344: looks like we don't need this line.
We can just do 
```
return  new BlobTransferPartitionMetadata(
        blobTransferRequest.getTopicName(),
        blobTransferRequest.getPartition(),
        offsetRecordByte,
        storeVersionStateByte);
```

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/BlobSnapshotManager.java
```diff
@@ -22,66 +32,183 @@
...
+        storageMetadataService,
+        DEFAULT_MAX_CONCURRENT_USERS,
+        DEFAULT_SNAPSHOT_RETENTION_TIME_IN_MIN);
+  }
+
+  /**
+   * Get the transfer metadata for a particular payload
+   * 1.  the store is not hybrid, it will prepare the metadata and return it
+   * 2.  the store is hybrid,
+   *   2. 1. throttle the request if many concurrent users
```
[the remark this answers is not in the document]
jingy-li: Good point! yes both hybrid and batch should check the concurrent user amount.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/BlobSnapshotManager.java
```diff
@@ -103,33 +103,35 @@ public BlobSnapshotManager(
...
    * @param payload the blob transfer payload
    * @return the need transfer metadata to client
    */
   public BlobTransferPartitionMetadata getTransferMetadata(BlobTransferPayload payload) throws VeniceException {
+    String topicName = payload.getTopicName();
+    int partitionId = payload.getPartition();
+
+    // check if the concurrent user count exceeds the limit
+    checkIfConcurrentUserExceedsLimit(topicName, partitionId);
+    concurrentSnapshotUsers.putIfAbsent(topicName, new VeniceConcurrentHashMap<>());
```
adamxchen, line 122: nit: you can probably do
```
concurrentSnapshotUsers
    .computeIfAbsent(topicName, k -> new VeniceConcurrentHashMap<>())
    .computeIfAbsent(partitionId, k -> new AtomicLong(0))
    .incrementAndGet();
```
 which is more compact

adamxchen (approved):
Left one nit but overall looks good to me! Thanks for the great work

adamxchen (approved):
🚀

adamxchen (approved):

PR #1242, opened 2024-10-16
[controller] Set min RF on controllers and storage clusters to RF - 1 with a floor of 1

kvargha:

Currently, the min replication factor (RF) for controllers is equal to replication factor. During a deployment, the number of instances could go below what is set to RF, which would cause a rebalance failure when using Waged.

To prevent this, we will be setting min RF = max(RF - 1, 1). This will be applied to the storage clusters too.

## How was this PR tested?

CI

### DIFF
--- a/services/venice-controller/src/main/java/com/linkedin/venice/controller/ZkHelixAdminClient.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/controller/ZkHelixAdminClient.java
@@ -164,7 +164,7 @@ public class ZkHelixAdminClient implements HelixAdminClient {
           AutoRebalanceStrategy.class.getName());
       VeniceControllerClusterConfig config = multiClusterConfigs.getControllerConfig(clusterName);
       IdealState idealState = helixAdmin.getResourceIdealState(controllerClusterName, clusterName);
-      idealState.setMinActiveReplicas(controllerClusterReplicaCount - 1);
+      idealState.setMinActiveReplicas(Math.max(controllerClusterReplicaCount - 1, 1));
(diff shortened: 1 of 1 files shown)


services/venice-controller/src/main/java/com/linkedin/venice/controller/ZkHelixAdminClient.java
```diff
@@ -164,7 +164,7 @@ public void addVeniceStorageClusterToControllerCluster(String clusterName) {
           AutoRebalanceStrategy.class.getName());
       VeniceControllerClusterConfig config = multiClusterConfigs.getControllerConfig(clusterName);
       IdealState idealState = helixAdmin.getResourceIdealState(controllerClusterName, clusterName);
-      idealState.setMinActiveReplicas(controllerClusterReplicaCount);
+      idealState.setMinActiveReplicas(controllerClusterReplicaCount - 1);
```
nisargthakkar, line 167: Can we do `Math.max(controllerClusterReplicaCount - 1,1)` instead?

kvargha: Do we want this for the storage cluster too?

nisargthakkar (approved):

services/venice-controller/src/main/java/com/linkedin/venice/controller/ZkHelixAdminClient.java
nisargthakkar: We should have it there as well

nisargthakkar (approved):

PR #1243, opened 2024-10-16
[common][dvc][server] evolve partition state schema for separate incr…

adamxchen:
…emetnal push topic

## Summary

In this PR, ~~deprecate `leaderOffset` field in the `Partitionstate` since upstreamOffsetMap's been used for a few years already.~~ (removed since it's breaking lots of integ tests so need to do it in a separate effort)

Also, in order to support separate incremental push topic, we need to evolve the schema such that we can persist both RT and separate incremental push related metadata .

Two new fields are added:
```
{
      "name": "separateLeaderTopic",
      "type": ["null", "string"],
      "doc": "The separate leader topic that leader is consuming from; It's used when leader has to ingest from two upstream topics, for example, real time topic and separate incremental push topic. ",
      "default": null
    }, {
      "name": "separateTopicUpstreamOffsetMap",
      "type": {
        "type": "map",
        "values": "long",
        "avro.java.string": "String"
      },
      "doc": "A map of upstream pubsub bootstrap server url -> the last pubsub offset consumed for the separate leader topic.",
      "default": {}
    }, 
```

## How was this PR tested?

existing unit test and integ test


adamxchen:
decided to not evolve and use existing fields

PR #1245, opened 2024-10-17
[server] Incremental push status should not update replica current status

sixpluszero:
## [server] Incremental push status should not update replica current status
Offline push job polling checks replica's current status to determine if replicas are ready, if status is not terminal, it will also check if it is ready to start buffer rewind.
Incremental push will also update the replica's current status, so it is possible that it will flip the current status from terminal (COMPLETED) -> non terminal (SOIP)
If there are concurrent / consecutive incremental push, it could potentially result in offline push job polling could not complete forever (especially when there are other factors - slow nodes and such). 

This PR changes the behavior so that SOIP/EOIP won't affect replica current status polling.
Also, it should stop logging tons of "not ready to start buffer replay" due to certain partition status flipped to non-terminal status and which is not EOIP

## How was this PR tested?

Add new integration test to validate the behavior and adjust unit test.

### DIFF
--- a/internal/venice-common/src/main/java/com/linkedin/venice/pushmonitor/ReplicaStatus.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/pushmonitor/ReplicaStatus.java
@@ -47,14 +47,16 @@ public class ReplicaStatus {
     this.statusHistory = enableStatusHistory ? new LinkedList<>() : null;
   }
 
-  public void updateStatus(ExecutionStatus newStatus) {
-    currentStatus = newStatus;
+  public void updateStatus(ExecutionStatus newStatus, String incrementalPushVersion) {
+    setIncrementalPushVersion(incrementalPushVersion);
+    if (!isIncrementalPushStatus(newStatus)) {
+      setCurrentStatus(newStatus);
+    }
     addHistoricStatus(newStatus);
   }
 
-  public void updateStatus(ExecutionStatus newStatus, String incrementalPushVersion) {
-    setIncrementalPushVersion(incrementalPushVersion);
-    updateStatus(newStatus);
+  void updateStatus(ExecutionStatus newStatus) {
+    updateStatus(newStatus, "");
   }
 
   public String getInstanceId() {
--- a/internal/venice-common/src/main/java/com/linkedin/venice/pushmonitor/ReplicaStatus.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/pushmonitor/ReplicaStatus.java
@@ -51,7 +51,7 @@ public class ReplicaStatus {
     if (!isIncrementalPushStatus(newStatus)) {
       setCurrentStatus(newStatus);
     }
-    addHistoricStatus(newStatus);
+    addHistoricStatus(newStatus, incrementalPushVersion);
   }
 
   void updateStatus(ExecutionStatus newStatus) {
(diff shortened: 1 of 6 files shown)


internal/venice-common/src/main/java/com/linkedin/venice/pushmonitor/ReplicaStatus.java
```diff
@@ -47,14 +47,16 @@ public ReplicaStatus(String instanceId, boolean enableStatusHistory) {
     this.statusHistory = enableStatusHistory ? new LinkedList<>() : null;
   }
 
-  public void updateStatus(ExecutionStatus newStatus) {
-    currentStatus = newStatus;
+  public void updateStatus(ExecutionStatus newStatus, String incrementalPushVersion) {
+    setIncrementalPushVersion(incrementalPushVersion);
```
gaojieliu, line 51: Do we need the replica-level incremental push version field?
I think this should be only required in the snapshot history, right?

internal/venice-common/src/main/java/com/linkedin/venice/pushmonitor/ReplicaStatus.java
```diff
@@ -47,14 +47,16 @@ public ReplicaStatus(String instanceId, boolean enableStatusHistory) {
...
+    if (!isIncrementalPushStatus(newStatus)) {
+      setCurrentStatus(newStatus);
+    }
     addHistoricStatus(newStatus);
```
gaojieliu, line 55: I assume we should pass `incrementalPushVersion` to this function, and right now, it is using the `incrementalPushVersion` updated by this function, but it seems very error prone, and passing it via param is more robust.

internal/venice-common/src/main/java/com/linkedin/venice/pushmonitor/ReplicaStatus.java
sixpluszero: Good question, let me check the usage

sixpluszero: No, I think this is misleading...Venice has been using this field to record different kinds of errors and different kinds of messages. Personally I am against this mixed usage, but I think the cleanup (or renaming) can happen in a separate PR as it is all over the places..
Anyway, for inc push, we are not using this field to determine job status

sushantmane (dismissed):
LGTM. Thanks, @sixpluszero!

internal/venice-common/src/main/java/com/linkedin/venice/pushmonitor/ReplicaStatus.java
```diff
@@ -120,8 +121,8 @@ private void addHistoricStatus(ExecutionStatus status) {
     removeOldStatuses();
 
     StatusSnapshot snapshot = new StatusSnapshot(status, LocalDateTime.now().toString());
-    if (!StringUtils.isEmpty(incrementalPushVersion)) {
-      snapshot.setIncrementalPushVersion(incrementalPushVersion);
+    if (isIncrementalPushStatus(status)) {
```
gaojieliu, line 124: This can't be right..
So the status can be an incremental push version?
I was thinking we should pass `incrementalPushVersion` as a separate param to function: `addHistoricStatus`.

sixpluszero: Yeah I was wrong......updated

gaojieliu (approved):

PR #1252, opened 2024-10-18
[server][common] Fixed bug in AA/WC parallel processing support

gaojieliu:
This PR fixed the following issues:
1. AASIT should pass non-null `KeyLevelLocksManager` to `IngestionBatchProcessor`, otherwise, race condition will happen.
2. Fixed the locking order in `IngestionBatchProcessor` to avoid deadlock.
3. Updated `SparseConcurrentList#computeIfAbsent` to skip adjust list size if the computed result is `null`.

## How was this PR tested?
CI

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
@@ -153,11 +153,10 @@ public class ActiveActiveStoreIngestionTask extends LeaderFollowerStoreIngestion
     this.remoteIngestionRepairService = builder.getRemoteIngestionRepairService();
     this.ingestionBatchProcessorLazy = Lazy.of(() -> {
       if (!serverConfig.isAAWCWorkloadParallelProcessingEnabled()) {
-        LOGGER
-            .info("AA/WC workload parallel processing enabled is false for store version: {}", getKafkaVersionTopic());
+        LOGGER.info("AA/WC workload parallel processing is disabled for store version: {}", getKafkaVersionTopic());
         return null;
       }
-      LOGGER.info("AA/WC workload parallel processing enabled is true for store version: {}", getKafkaVersionTopic());
+      LOGGER.info("AA/WC workload parallel processing is enabled for store version: {}", getKafkaVersionTopic());
       return new IngestionBatchProcessor(
           kafkaVersionTopic,
           parallelProcessingThreadPool,
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
@@ -153,14 +153,15 @@ public class ActiveActiveStoreIngestionTask extends LeaderFollowerStoreIngestion
     this.remoteIngestionRepairService = builder.getRemoteIngestionRepairService();
     this.ingestionBatchProcessorLazy = Lazy.of(() -> {
       if (!serverConfig.isAAWCWorkloadParallelProcessingEnabled()) {
-        LOGGER.info("AA/WC workload parallel processing enabled is false");
+        LOGGER
+            .info("AA/WC workload parallel processing enabled is false for store version: {}", getKafkaVersionTopic());
         return null;
       }
-      LOGGER.info("AA/WC workload parallel processing enabled is true");
+      LOGGER.info("AA/WC workload parallel processing enabled is true for store version: {}", getKafkaVersionTopic());
       return new IngestionBatchProcessor(
           kafkaVersionTopic,
           parallelProcessingThreadPool,
-          null,
+          keyLevelLocksManager.get(),
           this::processActiveActiveMessage,
           isWriteComputationEnabled,
           isActiveActiveReplicationEnabled(),
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/IngestionBatchProcessor.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/IngestionBatchProcessor.java
@@ -15,9 +15,9 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.TreeMap;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutorService;
-import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.locks.ReentrantLock;
 
 
@@ -81,39 +81,32 @@ public class IngestionBatchProcessor {
    * When {@link #lockManager} is not null, this function will try to lock all the keys
    * (except Control Messages) passed by the params.
    */
-  public List<ReentrantLock> lockKeys(List<PubSubMessage<KafkaKey, KafkaMessageEnvelope, Long>> records) {
+  public Map<ByteArrayKey, ReentrantLock> lockKeys(List<PubSubMessage<KafkaKey, KafkaMessageEnvelope, Long>> records) {
     if (lockManager != null) {
-      List<ReentrantLock> locks = new ArrayList<>(records.size());
-
       /**
-       * Collect all the keys and sort them to avoid deadlock.
+       * Need to use a {@link TreeMap} to make sure the locking will be executed in a deterministic order, otherwise
+       * deadlock can happen.
+       * Considering there could be multiple consumers, which are executing this function concurrently, and if they
+       * are trying to lock the same set of keys with different orders, deadlock can happen.
        */
-      List<byte[]> keys = new ArrayList<>(records.size());
-
+      Map<ByteArrayKey, ReentrantLock> keyLockMap =
+          new TreeMap<>((o1, o2) -> ByteUtils.compare(o1.getContent(), o2.getContent()));
       records.forEach(r -> {
         if (!r.getKey().isControlMessage()) {
-          keys.add(r.getKey().getKey());
+          keyLockMap.computeIfAbsent(ByteArrayKey.wrap(r.getKey().getKey()), k -> lockManager.acquireLockByKey(k));
         }
       });
-      Collections.sort(keys, (o1, o2) -> ByteUtils.compare(o1, o2));
-
-      keys.forEach(key -> {
-        ReentrantLock lock = lockManager.acquireLockByKey(ByteArrayKey.wrap(key));
-        locks.add(lock);
-        lock.lock();
-      });
-      return locks;
+      keyLockMap.forEach((k, v) -> v.lock());
+      return keyLockMap;
     }
-    return Collections.emptyList();
+    return Collections.emptyMap();
   }
 
-  public void unlockKeys(List<PubSubMessage<KafkaKey, KafkaMessageEnvelope, Long>> records, List<ReentrantLock> locks) {
+  public void unlockKeys(Map<ByteArrayKey, ReentrantLock> keyLockMap) {
     if (lockManager != null) {
-      locks.forEach(lock -> lock.unlock());
-      records.forEach(r -> {
-        if (!r.getKey().isControlMessage()) {
-          lockManager.releaseLock(ByteArrayKey.wrap(r.getKey().getKey()));
-        }
+      keyLockMap.forEach((key, lock) -> {
+        lock.unlock();
+        lockManager.releaseLock(key);
       });
     }
   }
@@ -140,23 +133,9 @@ public class IngestionBatchProcessor {
     if (records.isEmpty()) {
       return Collections.emptyList();
     }
-    AtomicBoolean isAllMessagesFromRTTopic = new AtomicBoolean(true);
+    boolean isAllMessagesFromRTTopic = true;
     List<PubSubMessageProcessedResultWrapper<KafkaKey, KafkaMessageEnvelope, Long>> resultList =
         new ArrayList<>(records.size());
-    records.forEach(r -> {
-      resultList.add(new PubSubMessageProcessedResultWrapper<>(r));
-      if (!r.getTopicPartition().getPubSubTopic().isRealTime()) {
-        isAllMessagesFromRTTopic.set(false);
-      }
-    });
-    if (!isWriteComputationEnabled && !isActiveActiveReplicationEnabled) {
-      return resultList;
-    }
-    // Only handle records from the real-time topic
-    if (!isAllMessagesFromRTTopic.get()) {
-      return resultList;
-    }
-
     /**
      * We would like to process the messages belonging to the same key sequentially to avoid race conditions.
      */
@@ -164,13 +143,24 @@ public class IngestionBatchProcessor {
     Map<ByteArrayKey, List<PubSubMessageProcessedResultWrapper<KafkaKey, KafkaMessageEnvelope, Long>>> keyGroupMap =
         new HashMap<>(records.size());
 
-    for (PubSubMessageProcessedResultWrapper<KafkaKey, KafkaMessageEnvelope, Long> r: resultList) {
-      PubSubMessage<KafkaKey, KafkaMessageEnvelope, Long> message = r.getMessage();
-      if (!message.getKey().isControlMessage()) {
+    for (PubSubMessage<KafkaKey, KafkaMessageEnvelope, Long> message: records) {
+      if (!message.getTopicPartition().getPubSubTopic().isRealTime()) {
+        isAllMessagesFromRTTopic = false;
+      }
+      PubSubMessageProcessedResultWrapper resultWrapper = new PubSubMessageProcessedResultWrapper<>(message);
+      resultList.add(resultWrapper);
+      if (!message.getKey().isControlMessage() && isAllMessagesFromRTTopic) {
         ByteArrayKey byteArrayKey = ByteArrayKey.wrap(message.getKey().getKey());
-        keyGroupMap.computeIfAbsent(byteArrayKey, (ignored) -> new ArrayList<>()).add(r);
+        keyGroupMap.computeIfAbsent(byteArrayKey, (ignored) -> new ArrayList<>()).add(resultWrapper);
         totalNumOfRecords++;
       }
+    }
+    if (!isWriteComputationEnabled && !isActiveActiveReplicationEnabled) {
+      return resultList;
+    }
+    // Only handle records from the real-time topic
+    if (!isAllMessagesFromRTTopic) {
+      return resultList;
     }
     aggVersionedIngestionStats
         .recordBatchProcessingRequest(storeName, version, totalNumOfRecords, System.currentTimeMillis());
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/IngestionBatchProcessor.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/IngestionBatchProcessor.java
@@ -28,7 +28,7 @@ import java.util.concurrent.locks.ReentrantLock;
  * resources to speed up the leader ingestion.
  */
 public class IngestionBatchProcessor {
-  public static final TreeMap EMPTY_TREE_MAP = new TreeMap();
+  private static final TreeMap EMPTY_TREE_MAP = new TreeMap();
 
   interface ProcessingFunction {
     PubSubMessageProcessedResult apply(
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/IngestionBatchProcessor.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/IngestionBatchProcessor.java
@@ -8,13 +8,13 @@ import com.linkedin.venice.kafka.protocol.KafkaMessageEnvelope;
 import com.linkedin.venice.message.KafkaKey;
 import com.linkedin.venice.meta.Version;
 import com.linkedin.venice.pubsub.api.PubSubMessage;
-import com.linkedin.venice.utils.ByteUtils;
 import com.linkedin.venice.utils.LatencyUtils;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.NavigableMap;
 import java.util.TreeMap;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutorService;
@@ -83,7 +83,7 @@ public class IngestionBatchProcessor {
    * When {@link #lockManager} is not null, this function will try to lock all the keys
    * (except Control Messages) passed by the params.
    */
-  public TreeMap<ByteArrayKey, ReentrantLock> lockKeys(
+  public NavigableMap<ByteArrayKey, ReentrantLock> lockKeys(
       List<PubSubMessage<KafkaKey, KafkaMessageEnvelope, Long>> records) {
     if (lockManager != null) {
       /**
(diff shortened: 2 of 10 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/IngestionBatchProcessor.java
```diff
@@ -71,20 +72,36 @@ public IngestionBatchProcessor(
...
    * (except Control Messages) passed by the params.
    */
   public List<ReentrantLock> lockKeys(List<PubSubMessage<KafkaKey, KafkaMessageEnvelope, Long>> records) {
     if (lockManager != null) {
       List<ReentrantLock> locks = new ArrayList<>(records.size());
+
+      /**
+       * Collect all the keys and sort them to avoid deadlock.
+       */
+      List<byte[]> keys = new ArrayList<>(records.size());
```
FelixGV, line 91: Should it be a Set instead? Do you want to dedupe the keys in the batch? What if the same key is there twice and you attempt to acquire the lock for it twice? Can this scenario be unit tested?

If deduping is desired, using a Set is not the only way to achieve it of course. It could also be done as part of the latter iteration, after the list is sorted...

internal/venice-client-common/src/main/java/com/linkedin/venice/utils/SparseConcurrentList.java
```diff
@@ -249,6 +249,12 @@ public E computeIfAbsent(int index, IntFunction<? extends E> mappingFunction) {
         element = get(index);
         if (element == null) {
           element = mappingFunction.apply(index);
+          /**
+           * Don't update the list if the computed result is `null`.
```
FelixGV, line 253: Good catch...

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/IngestionBatchProcessor.java
```diff
@@ -71,20 +72,36 @@ public IngestionBatchProcessor(
...
+
       records.forEach(r -> {
         if (!r.getKey().isControlMessage()) {
-          ReentrantLock lock = lockManager.acquireLockByKey(ByteArrayKey.wrap(r.getKey().getKey()));
-          locks.add(lock);
-          lock.lock();
+          keys.add(r.getKey().getKey());
         }
       });
+      Collections.sort(keys, (o1, o2) -> ByteUtils.compare(o1, o2));
```
FelixGV, line 98: Why does the sorting matter?

FelixGV: Oh I think I understand... because of A/A, two batches with different orders of the same two keys would deadlock?

FelixGV: This seems quite subtle and should be documented with a comment, IMHO...

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/IngestionBatchProcessor.java
KaiSernLim: I think that the current use-case dedupes the messages in the function call immediately before, so it's not "currently" an issue but could be if these locks are used elsewhere ([link](https://github.com/linkedin/venice/blob/073e61d672734234b0c434f266c541cd2b69a388/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java#L1147))?

gaojieliu: That is a good question.
I think both List and Set can work as the same lock can be `locked`/`unlocked` multiple times.

I will add a unit test for repetitive keys in the same batch.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/IngestionBatchProcessor.java
gaojieliu: Sure, will add the javadoc to include the details.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/IngestionBatchProcessor.java
FelixGV: If the deduping is elsewhere then no point in doing it again here...

That being said, I would encourage looking at the algorithm end-to-end, across all of the steps, and trying to see how we can minimize the collection traversals and instantiation and copying. Maybe there are ways to collapse multiple steps into one and do deduping and sorting at the same time (e.g. with a `TreeMap`, perhaps). I'm not prescribing to do that specifically, but just to think about it, since ingestion is sensitive to efficiency.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/IngestionBatchProcessor.java
```diff
@@ -71,32 +72,41 @@ public IngestionBatchProcessor(
...
     if (lockManager != null) {
-      List<ReentrantLock> locks = new ArrayList<>(records.size());
+      /**
+       * Need to use a {@link TreeMap} to make sure the locking will be executed in a deterministic order, otherwise
+       * deadlock can happen.
+       * Considering there could be multiple consumers, which are executing this function concurrently, and if they
+       * are trying to lock the same set of keys with different orders, deadlock can happen.
+       */
+      Map<ByteArrayKey, ReentrantLock> keyLockMap =
+          new TreeMap<>((o1, o2) -> ByteUtils.compare(o1.getContent(), o2.getContent()));
```
majisourav99, line 93: can we compare with the `ByteArrayKey::hashCode()` should be faster than comparing byte arrays.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/IngestionBatchProcessor.java
```diff
@@ -81,39 +81,32 @@ KeyLevelLocksManager getLockManager() {
...
+  public void unlockKeys(Map<ByteArrayKey, ReentrantLock> keyLockMap) {
     if (lockManager != null) {
-      locks.forEach(lock -> lock.unlock());
-      records.forEach(r -> {
-        if (!r.getKey().isControlMessage()) {
-          lockManager.releaseLock(ByteArrayKey.wrap(r.getKey().getKey()));
-        }
+      keyLockMap.forEach((key, lock) -> {
+        lock.unlock();
+        lockManager.releaseLock(key);
```
FelixGV, line 109: This is pre-existing code, not something you are changing in this PR, but I wonder... why did we design the `KeyLevelLocksManager` in this way? IIUC, the `lock.unlock()` unlocks (duh) and then the `lockManager.releaseLock(key)` decreases the ref count and potentially returns the lock to the pool... Is there ever a situation where we would wish to do one of these operations but not the other? At first glance, it seems to me that the only correct way to use this API is to do both of these calls. So why aren't they fused into a single function? IOW, why do I need to handle the `ReentrantLock` object at all? Why not make the unlock a side-effect of the `lockManager.releaseLock(key)` call?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/IngestionBatchProcessor.java
```diff
@@ -123,38 +133,35 @@ public List<PubSubMessageProcessedResultWrapper<KafkaKey, KafkaMessageEnvelope,
...
      */
     int totalNumOfRecords = 0;
     Map<ByteArrayKey, List<PubSubMessageProcessedResultWrapper<KafkaKey, KafkaMessageEnvelope, Long>>> keyGroupMap =
         new HashMap<>(records.size());
 
-    for (PubSubMessageProcessedResultWrapper<KafkaKey, KafkaMessageEnvelope, Long> r: resultList) {
-      PubSubMessage<KafkaKey, KafkaMessageEnvelope, Long> message = r.getMessage();
-      if (!message.getKey().isControlMessage()) {
+    for (PubSubMessage<KafkaKey, KafkaMessageEnvelope, Long> message: records) {
+      if (!message.getTopicPartition().getPubSubTopic().isRealTime()) {
```
majisourav99, line 147: should we check for not control message here?

FelixGV (commented):
Sorry, I am doing mostly white line commenting in this round of review. I should have reviewed the first round of implementation of this functionality, but didn't get a chance, so I'm coming to the party a bit late. I just want to provide the following feedback:

The part of `produceToStoreBufferServiceOrKafkaInBatch` which splits the records coming from Kafka into "mini batches" seems not ideal to me, and I feel like we should get rid of it. To be clear, I don't think the current code cannot work, but just that it is more complex than it should be. I will list below some issues/concerns that I see with it, and all of these can be fixed individually, but I still think getting rid of that part of the algorithm could be overall better:

1. The batch size comes from `serverConfig.getAAWCWorkloadParallelProcessingThreadPoolSize()`, which does not seem to make sense. The default value of that config is 8, which is likely too small. I think the intent was probably to have a separate config for the batch size.
2. If the Kafka batch is smaller than the configured batch size, then we still instantiate an `ongoingBatch` of the configured size, which is wasteful in terms of memory allocation (easy to fix, but is it needed?).
3. I expect that the batch size coming from Kafka should not be too large to handle, and so there is no point in splitting it. If anything, I would think there may be cases where the batch from Kafka is _too small_ and that the optimal thing may be to accumulate records for some time and then create a _mega batch_ (the opposite of a mini batch!).
4. One of the benefits we hope to get from this functionality is to process together the updates to the same key, so that we do fewer RocksDB reads and writes in those cases. This effect is reduced by the mini batches.

So my recommendation would be to simplify this code by deleting the mini batching altogether, and just pass the Kafka batch directly into the next step. No need to implement the mega batching yet, we can make that call later.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/IngestionBatchProcessor.java
```diff
@@ -71,32 +72,41 @@ public IngestionBatchProcessor(
...
 
-  public void unlockKeys(List<PubSubMessage<KafkaKey, KafkaMessageEnvelope, Long>> records, List<ReentrantLock> locks) {
+  public void unlockKeys(Map<ByteArrayKey, ReentrantLock> keyLockMap) {
     if (lockManager != null) {
-      locks.forEach(lock -> lock.unlock());
-      records.forEach(r -> {
-        if (!r.getKey().isControlMessage()) {
-          lockManager.releaseLock(ByteArrayKey.wrap(r.getKey().getKey()));
-        }
+      keyLockMap.forEach((key, lock) -> {
```
lluwm, line 107: Nit, i don't think anything could go wrong, but if we release the locks in the reverse order of their acquisition, it might be easy to reason about in the future.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/IngestionBatchProcessor.java
gaojieliu: Yeah, this is reasonable.
@huangminchn Can you chime in here regarding the above question?

gaojieliu: Fusing two calls into one should work and we do have cases to call `lock/unlock` multiple times in the same thread, so maintaining a counter will be beneficial.
When this feature is enabled, consumer will lock a batch of keys, and to be backward compatible, the existing function will lock the same key again:
https://github.com/linkedin/venice/blob/main/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java#L225

As you mentioned, we can create a wrapper of lock to handle the reference counting logic.
So far, I don't see a case, which would call one of the functions only.

gaojieliu:
> Sorry, I am doing mostly white line commenting in this round of review. I should have reviewed the first round of implementation of this functionality, but didn't get a chance, so I'm coming to the party a bit late. I just want to provide the following feedback:
> 
> The part of `produceToStoreBufferServiceOrKafkaInBatch` which splits the records coming from Kafka into "mini batches" seems not ideal to me, and I feel like we should get rid of it. To be clear, I don't think the current code cannot work, but just that it is more complex than it should be. I will list below some issues/concerns that I see with it, and all of these can be fixed individually, but I still think getting rid of that part of the algorithm could be overall better:
> 
> 1. The batch size comes from `serverConfig.getAAWCWorkloadParallelProcessingThreadPoolSize()`, which does not seem to make sense. The default value of that config is 8, which is likely too small. I think the intent was probably to have a separate config for the batch size.
> 2. If the Kafka batch is smaller than the configured batch size, then we still instantiate an `ongoingBatch` of the configured size, which is wasteful in terms of memory allocation (easy to fix, but is it needed?).
> 3. I expect that the batch size coming from Kafka should not be too large to handle, and so there is no point in splitting it. If anything, I would think there may be cases where the batch from Kafka is _too small_ and that the optimal thing may be to accumulate records for some time and then create a _mega batch_ (the opposite of a mini batch!).
> 4. One of the benefits we hope to get from this functionality is to process together the updates to the same key, so that we do fewer RocksDB reads and writes in those cases. This effect is reduced by the mini batches.
> 
> So my recommendation would be to simplify this code by deleting the mini batching altogether, and just pass the Kafka batch directly into the next step. No need to implement the mega batching yet, we can make that call later.

The biggest reason of mini-batching is to make the key-level lock usage deterministic...
If we always pass what we received to `IngestionBatchProcessor`, the `KeyLevelLockManager` will need to consider the max possible batch size, which is `100` IIRC.
4) is not implemented yet, so the smaller batch size in theory shouldn't affect performance much.
When working on 4), we need to use a separate config to control the batch size, so that we can tune the memory usage/performance benefits gained by de-duping. 

For 2), today's interface is `Iterable`, so we can't get the size without iterating it and to reduce the unnecessary copy, we need to change the interface to use something like `List`.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -1256,7 +1257,7 @@ protected void produceToStoreBufferServiceOrKafkaInBatch(
      * Process records batch by batch.
```
FelixGV, line 1257: >> I'm going to start a proper thread otherwise tracking replies in the top-level "Conversation" tab is tedious... pasting the previous discussion here:
>> 
>> Sorry, I am doing mostly white line commenting in this round of review. I should have reviewed the first round of implementation of this functionality, but didn't get a chance, so I'm coming to the party a bit late. I just want to provide the following feedback:
>> 
>> The part of produceToStoreBufferServiceOrKafkaInBatch which splits the records coming from Kafka into "mini batches" seems not ideal to me, and I feel like we should get rid of it. To be clear, I don't think the current code cannot work, but just that it is more complex than it should be. I will list below some issues/concerns that I see with it, and all of these can be fixed individually, but I still think getting rid of that part of the algorithm could be overall better:
>> 
>> 1. The batch size comes from serverConfig.getAAWCWorkloadParallelProcessingThreadPoolSize(), which does not seem to make sense. The default value of that config is 8, which is likely too small. I think the intent was probably to have a separate config for the batch size.
>> 2. If the Kafka batch is smaller than the configured batch size, then we still instantiate an ongoingBatch of the configured size, which is wasteful in terms of memory allocation (easy to fix, but is it needed?).
>> 3. I expect that the batch size coming from Kafka should not be too large to handle, and so there is no point in splitting it. If anything, I would think there may be cases where the batch from Kafka is too small and that the optimal thing may be to accumulate records for some time and then create a mega batch (the opposite of a mini batch!).
>> 4. One of the benefits we hope to get from this functionality is to process together the updates to the same key, so that we do fewer RocksDB reads and writes in those cases. This effect is reduced by the mini batches.
>> 
>> So my recommendation would be to simplify this code by deleting the mini batching altogether, and just pass the Kafka batch directly into the next step. No need to implement the mega batching yet, we can make that call later.
> 
> The biggest reason of mini-batching is to make the key-level lock usage deterministic...
> 
> If we always pass what we received to IngestionBatchProcessor, the KeyLevelLockManager will need to consider the max possible batch size, which is 100 IIRC.

Here, do you mean that the `KeyLevelLocksManager.maxPoolSize` must be smaller than the processed batch size, otherwise the function to lock all keys of the batch would stall?

> 4) is not implemented yet, so the smaller batch size in theory shouldn't affect performance much.
> When working on 4), we need to use a separate config to control the batch size, so that we can tune the memory usage/performance benefits gained by de-duping.
> 
> For 2), today's interface is Iterable, so we can't get the size without iterating it and to reduce the unnecessary copy, we need to change the interface to use something like List.

That is easy, since the upstream function is `StorePartitionDataReceiver::write(List<PubSubMessage<KafkaKey, KafkaMessageEnvelope, Long>> consumedData)`, which calls `SIT::produceToStoreBufferServiceOrKafka` (at which point the type gets constrained to `Iterable`. It could be made into a `List` all the way through.

Coming back to point 1... I've read some more of the code and I think I finally understand what's going on... the batch size is set so that `IngestionBatchProcessor::process` submits roughly 1 key per thread to the `batchProcessingThreadPool`... But is this going to be efficient? I wonder if the granularity is too fine... We will submit a lot of tasks into this thread pool. It might be better to submit fewer tasks, each of which contains a bigger collection of events to process.

Anyway, we can keep iterating on this new functionality and optimizing it later. No need to get everything perfect from the get go... but it seems to me that there might be many such opportunities. Lots of copying and processing that might be collapsible into fewer steps, and then also taking advantage of writes to the same key.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/IngestionBatchProcessor.java
gaojieliu: Hash collision can become a disaster, which might result in deadlock.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
```diff
@@ -154,14 +154,15 @@ public ActiveActiveStoreIngestionTask(
     this.remoteIngestionRepairService = builder.getRemoteIngestionRepairService();
     this.ingestionBatchProcessorLazy = Lazy.of(() -> {
       if (!serverConfig.isAAWCWorkloadParallelProcessingEnabled()) {
-        LOGGER.info("AA/WC workload parallel processing enabled is false");
+        LOGGER
+            .info("AA/WC workload parallel processing enabled is false for store version: {}", getKafkaVersionTopic());
```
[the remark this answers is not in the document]
gaojieliu: Sure

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/IngestionBatchProcessor.java
gaojieliu: Emm.. I don't think there is a need and we don't have a use case where the CM is from VT, but actual messages are from RT, right?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/IngestionBatchProcessor.java
majisourav99: Not sure where the current ByteArrayKey hashCode is being used, but current hash generation looks too simple, does not use SHA or MD5. It might use hash collision more frequently.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
FelixGV: Talked offline and we agreed to keep this PR minimal to fix the bug ASAP, and to continue exploring other refactorings later.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/utils/ByteArrayKey.java
```diff
@@ -31,6 +31,10 @@ public boolean equals(Object o) {
     return Arrays.equals(content, that.content);
   }
 
+  public byte[] getContent() {
+    return this.content;
+  }
```
FelixGV, line 36: This seems like a dangerous change... the `content` field is intended to be immutable, but there is no way to ensure this besides encapsulation and visibility rules, but we are now violating that with this getter... Since the goal is to use it in a comparator, can we instead make this class implement `Comparable`? It seems that it would achieve the same goal while preserving encapsulation.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/IngestionBatchProcessor.java
```diff
@@ -99,12 +102,12 @@ public Map<ByteArrayKey, ReentrantLock> lockKeys(List<PubSubMessage<KafkaKey, Ka
       keyLockMap.forEach((k, v) -> v.lock());
       return keyLockMap;
     }
-    return Collections.emptyMap();
+    return EMPTY_TREE_MAP;
```
FelixGV: Instead of this, can you please make the `lockKeys` function return a `NavigableMap` and return `Collections.emptyNavigableMap()` here? The issue with the current code is that the statically-defined `EMPTY_TREE_MAP` is modifiable, which is not safe... (of course, the current code never modifies it, but it would be more defensive to make it unmodifiable). `unlockKeys` and whatever else function gets passed a `TreeMap` would need to be switched to `NavigableMap` as well.

FelixGV (commented):
Just a couple of very minor comments, but I think this is soon good to go...! Thanks for the fix!

clients/da-vinci-client/src/main/java/com/linkedin/davinci/utils/ByteArrayKey.java
gaojieliu: Fixed.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/IngestionBatchProcessor.java
gaojieliu: Fixed.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/IngestionBatchProcessor.java
```diff
@@ -92,8 +92,7 @@ public TreeMap<ByteArrayKey, ReentrantLock> lockKeys(
        * Considering there could be multiple consumers, which are executing this function concurrently, and if they
        * are trying to lock the same set of keys with different orders, deadlock can happen.
        */
-      TreeMap<ByteArrayKey, ReentrantLock> keyLockMap =
-          new TreeMap<>((o1, o2) -> ByteUtils.compare(o1.getContent(), o2.getContent()));
+      TreeMap<ByteArrayKey, ReentrantLock> keyLockMap = new TreeMap<>();
```
FelixGV, line 95: Nice, looks even cleaner this way.

FelixGV (approved):
Thank you Gaojie! 🚀 🚢

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/IngestionBatchProcessor.java
gaojieliu: Will optimize this in a separated PR.

PR #1253, opened 2024-10-21
[router] Added more debugging details to compression inconsistency issue

FelixGV:
The router expects to get the same compression scheme from all server responses when performing a batch get, but it seems that this is not always the case. This change will log the response headers in case this happens, in the hope that it gives more clues about what went wrong.

## How was this PR tested?
GHCI.


sixpluszero (approved):

FelixGV:
Thanks for the review @sixpluszero!

PR #1255, opened 2024-10-22
[server] fix flaky test in blob transfer integration test

jingy-li:
## [server] fix flaky test in blob transfer integration test
Fix the flaky integration test in blob transfer. The offset records of server1 and server2 could differ because the ingestion speed may vary.

## How was this PR tested?
Integration test.


internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/BlobP2PTransferAmongServersTest.java
adamxchen, line 1: Can you plz also add a test-level timeout just in case?

adamxchen (approved):

PR #1256, opened 2024-10-23
[server] Fixed storage node read quota usage ratio spikes

xunyin8:
## [server] Fixed storage node read quota usage ratio spikes

1. We observed unusually high quota usage ratio spikes (7x) without any rejected requests. This is mathematically
impossible if the spikes were caused by an actual KPS spike since our default sampling window is 30s and token
bucket capacity multiplier is only 5x. This is possible because we currently can have a mismatch between KPS
and node responsibility of different versions (current, backup and future). i.e. Depending on the timing of things we
could be calculating ratio using the KPS of current + backup and divide that by the node responsibility of a future
version which has its replicas partially assigned, resulting in a huge spike in usage ratio.

2. To solve this we will monitor the KPS and QPS based on corresponding versions. Version swap is not atomic when we
have many routers and fast clients updating their current version metadata separately. Therefore, it's expected for
some short periods of time to receive traffic for both current and backup versions. We will track the requested quota
stats separately and use the current version stats to calculate the usage ratio. We can also have alerts on the
backup version requested stats since it's expected to drop to 0 if all the fast clients and routers are updating
their metadata correctly.

3. Quota rejection will remain to be enforced at the version level but emit stats at the store level. We can convert
it to versioned as well but currently don't see the needs to do so.

## How was this PR tested?
New unit and existing integration tests


sushantmane:
Can you add some info about what's the fix for the problems found?

xunyin8:
> Can you add some info about what's the fix for the problems found?

Sure.
For #1 the fix is to reset the KPS metric that's sampled at a 30s interval. This way we don't get in the situation where the node responsibility is updated and most recent KPS is low but the KPS used as the dividend remains high while the divisor could be a lot smaller resulting in a spike in usage ratio.

For #2 we are only updating the store level stats node responsibility if the change is on the serving/current version. Otherwise we keep it track in a separate map and make the update later when we check for current version change in `handleStoreChanged`

services/venice-server/src/main/java/com/linkedin/venice/listener/ReadQuotaEnforcementHandler.java
```diff
@@ -328,7 +330,10 @@ private void updateQuota(PartitionAssignment partitionAssignment) {
           clock);
 
       if (rateLimiter != v) {
-        stats.setNodeQuotaResponsibility(storeName, (long) Math.ceil(quotaInRcu * thisNodeQuotaResponsibility));
+        storeVersionNodeResponsibility.put(topic, (long) Math.ceil(quotaInRcu * thisNodeQuotaResponsibility));
```
majisourav99, line 333: Can we return early from `updateQuota` for non-current versions?

xunyin8: I think we could since we are also calling `onCustomizedViewChange` for every version in `handleStoreChanged`. Do you see any other potential holes where we can result in a current version change without having the appropriate token bucket updated and ready?

xunyin8: On a second thought I think it's better to update as often as we can in case of potential race when local store is updated with new current version and with new request coming in but `handleStoreChanged` hasn't finished yet and we cannot find the corresponding new token bucket. The update cost of some in-memory data structure should be negligible especially when partition assignment don't change that frequently.

services/venice-server/src/main/java/com/linkedin/venice/stats/ServerQuotaUsageStats.java
```diff
@@ -6,38 +6,44 @@
...
   private final Sensor rejectedQPS; // rejected query per second
   private final Sensor rejectedKPS; // rejected key per second
   private final Sensor allowedUnintentionallyKPS; // allowed KPS unintentionally due to error or insufficient info
-  private final Sensor usageRatioSensor; // requested qps divided by nodes quota responsibility
+  private final Sensor usageRatioSensor; // requested kps divided by nodes quota responsibility
+  private final Sensor nodeResponsibilitySensor; // kps assigned to the node based on store quota and replica assignment
+  private static final String REQUESTED_KPS_SENSOR_NAME = "quota_request_key_count";
 
-  private long nodeQpsResponsibility = 0;
+  private final AtomicLong nodeKpsResponsibility = new AtomicLong(0);
```
gaojieliu, line 27: I have some question about the calculation of request KPS/quota and usage.
Here is my assumption:
1. Request KPS will cover all the KPS for the given store, which can from the current version (mostly) and backup version(during swap time as fast-client version swap is not atomic and has delays and the switch can happen at different times in different nodes).
2. KPS quota will be bounded to a specific version.
3. Quota is the total kps (can from mixed versions) / total quota for a specific version.

If the above assumption is correct, it is possible that the quota usage during the version swap can still be weird and if some fast client nodes stick to the backup version for longer period unexpectedly, the ratio can be hard to reason.
Ideally, we should align the version between these two metrics, so that the quota usage ratio will be more accurate.
Here is an idea:
1. Maintain a separate KPS metric for backup and current version.
2. Quota will be calculated based on the current version.
3. Calculate the quota usage ratio by the current version KPS / current version quota.
4. We can emit the KPS metric for backup version, and we can setup alert against this metric if this value doesn't go down quickly (which indicates some fast-client nodes are not flipping and it will be great if we can log the problematic nodes as well, for example, if the backup version KPS doesn't go down to 0 within 3 mins, we will log some sample request node hostnames periodically for debugging purpose).
WDYT?

xunyin8: That's a good observation, I do think the recommendation is worth pursing since it will also give us some additional visibility on top of a more accurate usage ratio. Let me make the changfe.

services/venice-server/src/main/java/com/linkedin/venice/listener/ReadQuotaEnforcementHandler.java
```diff
@@ -312,6 +313,8 @@ private void updateQuota(PartitionAssignment partitionAssignment) {
           "Routing data changed on quota enforcement handler with 0 replicas assigned to this node, removing quota for resource: {}",
           topic);
       storeVersionRateLimiters.remove(topic);
+      stats.getStoreStats(Version.parseStoreFromKafkaTopicName(topic))
+          .removeVersion(Version.parseVersionFromKafkaTopicName(topic));
```
gaojieliu, line 317: Nit,
Can you parse store-name/version from the topic name once and reuse them in the whole function?

services/venice-server/src/main/java/com/linkedin/venice/stats/ServerReadQuotaUsageStats.java
```diff
@@ -0,0 +1,182 @@
...
+    rejectedQPS = registerSensor("quota_rejected_request", new Rate());
+    rejectedKPS = registerSensor("quota_rejected_key_count", new Rate());
+    allowedUnintentionallyKPS = registerSensor("quota_unintentionally_allowed_key_count", new Count());
+    usageRatioSensor =
+        registerSensor(new AsyncGauge((ignored, ignored2) -> getReadQuotaUsageRatio(), "quota_requested_usage_ratio"));
+    this.time = time;
+    metricConfig = new MetricConfig(); // use default configs
+  }
+
+  public void setCurrentVersion(int version) {
```
gaojieliu, line 67: Can this method be invoked by multiple threads concurrently?
If yes, there is a race condition as compare/swap are not atomic here.
So if the method can only be invoked very infrequently, we can make it `synchronized`.

gaojieliu: Or use `AtomicInteger#compareAndSet` to use the returned `boolean` to decide whether to set `backupVersion` or not.

xunyin8: Good call, I don't think it can be invoked by multiple threads concurrently as of now but just to be future-proof let me make it thread safe.

gaojieliu (dismissed):
Code change LGTM now, but you do need to fix the CI failure.

services/venice-server/src/main/java/com/linkedin/venice/listener/ReadQuotaEnforcementHandler.java
```diff
@@ -465,6 +472,8 @@ private void removeTopics(Set<String> topicsToRemove) {
     for (String topic: topicsToRemove) {
       customizedViewRepository.unSubscribeRoutingDataChange(topic, this);
       storeVersionRateLimiters.remove(topic);
+      stats.getStoreStats(Version.parseStoreFromKafkaTopicName(topic))
```
majisourav99, line 475: Can we add a check for existing of the storeStats, otherwise it will unnecessarily add a start just to be removed later.

xunyin8: Sure

gaojieliu (approved):

services/venice-server/src/main/java/com/linkedin/venice/listener/ReadQuotaEnforcementHandler.java
```diff
@@ -465,6 +473,10 @@ private void removeTopics(Set<String> topicsToRemove) {
     for (String topic: topicsToRemove) {
       customizedViewRepository.unSubscribeRoutingDataChange(topic, this);
       storeVersionRateLimiters.remove(topic);
+      ServerReadQuotaUsageStats storeStats = stats.getStoreStats(Version.parseStoreFromKafkaTopicName(topic));
+      if (storeStats != null) {
```
majisourav99, line 477: `getStoreStats` will always return a non-null (does `computeIfAbsent`). You will need a new API to check contains.

services/venice-server/src/main/java/com/linkedin/venice/listener/ReadQuotaEnforcementHandler.java
```diff
@@ -379,6 +384,8 @@ public void onPartitionStatusChange(String topic, ReadOnlyPartitionStatus partit
   @Override
   public void onRoutingDataDeleted(String kafkaTopic) {
     storeVersionRateLimiters.remove(kafkaTopic);
+    stats.getStoreStats(Version.parseStoreFromKafkaTopicName(kafkaTopic))
```
majisourav99, line 387: can you please apply the same change to check for null  here?

services/venice-server/src/main/java/com/linkedin/venice/listener/ReadQuotaEnforcementHandler.java
xunyin8: Ops, good catch. Done.

majisourav99 (approved):
Thanks Xun, LGTM, hopefully tests will all pass :)

PR #1258, opened 2024-10-24
[compat][server][controller] Introduced options to speed up KafkaProducer in Server for nearline workload

gaojieliu:
During benchmarking, we noticed that the producing is slow and sometimes it can take 10+ms for producing one single record for some medium size record: 20-30KB.
There are two issues we discovered during the benchmarking:
1. Kafka producer compression is taking a lot of time. Even we disable KafkaProducer compression, Kafka broker will compress it according to the target compression setup in the broker.
2. Kafka Producer is not scalable and when there are multiple threads invoking the same producer, there are a lot of contentions.

These optimizations have overheads:
1. By disabling compression, the Kafka workload will increase and considering the optimization is only for nearline workload, hope the workload increase is not too much. Also there is a store-level control.
2. More Producers in Venice Server. More memory consumption as each producer has its own buffer, and more producer threads.

Ideally, we would like to enable these optimizations in some critical stores, which have heavy AA/WC nearline workload to improve the E2E write latency.

New Server Config:
server.nearline.workload.producer.throughput.optimization.enabled: default true

This setting can override all the store-level config to disable the optimization entirely.

Two new store-level config:
NearlineProducerCountPerWriter: default 1
NearlineProducerCompressionEnabled: default true

After changing these store-level config, if we want to apply it to the current ongoing versions, we need to bounce Venice Servers, and for the future pushes, it will apply automatically.

We can use admin-tool to update these store-level configs.

This PR also changes the admin operation protocol, so we need to deploy child controllers first and then parent controller.

## How was this PR tested?

CI


clients/venice-admin-tool/src/main/java/com/linkedin/venice/Arg.java
```diff
@@ -275,7 +275,15 @@ public enum Arg {
...
   DEBUG("debug", "d", false, "Print debugging messages for execute-data-recovery"),
-  BLOB_TRANSFER_ENABLED("blob-transfer-enabled", "bt", true, "Flag to indicate if the blob transfer is allowed or not");
+  BLOB_TRANSFER_ENABLED("blob-transfer-enabled", "bt", true, "Flag to indicate if the blob transfer is allowed or not"),
+  NEARLINE_PRODUCER_COMPRESSION_ENABLED(
+      "nearline-producer-compression-enabled", "npce", true,
+      "Flag to control whether KafkaProducer will use compression or not for nearline workload"
+  ),
+  NEARLINE_PRODUCER_COUNT_PER_WRITER(
+      "nearline-producer-count-per-writer", "npcpw", true,
+      "How many producer will be used to write nearline workload in Server"
```
ZacAttack, line 285: nit: producer**s**

ZacAttack (dismissed):
This looks good to me, but let's think through if there are any ordering issues when it comes to setting the writer on the pcs, I'm concerned swapping the writer might open up some edge cases.  I guess before we swap we need to be certain the writer has flushed everything?  And then maybe actually we should consider shutting down the old one at some point?  I don't think we'd use it anymore once all the assigned partitions have passed the EOP.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -2296,6 +2392,19 @@ protected DelegateConsumerRecordResult delegateConsumerRecord(
             controlMessage,
             consumerRecord.getPubSubMessageHeaders(),
             partitionConsumptionState);
+        /**
+         * For the local consumption, the batch data won't be produce to the local VT again, so we will switch
+         * to real-time writer upon EOP here and for the remote consumption of VT, the switch will be handled
+         * in the following section as it needs to flush the messages and then switch.
```
ZacAttack, line 2398: Do we need to call flush in the below?

gaojieliu: In theory, we don't need to flush if the leader doesn't `produceToLocalKafka` and I am moving this section into the following `if` scope.

gaojieliu:
> This looks good to me, but let's think through if there are any ordering issues when it comes to setting the writer on the pcs, I'm concerned swapping the writer might open up some edge cases. I guess before we swap we need to be certain the writer has flushed everything? And then maybe actually we should consider shutting down the old one at some point? I don't think we'd use it anymore once all the assigned partitions have passed the EOP.

The flush will be invoked before the switch and we do have logic to shutdown the old one after all partitions finish the batch consumption, and please check this function: `checkWhetherToCloseUnusedVeniceWriter` in `LFSIT`.

ZacAttack (approved):

gaojieliu:
@ZacAttack 
Thanks for the rapid review!

PR #1259, opened 2024-10-25
[controller] Client side foundations and refactor to support gRPC transport for controller APIs

mynameborat:
- Introduces new controller transport client for gRPC
- Introduce interfaces and models for writing new controller APIs / migrating existing ones to use gRPC transport
- Demonstrate the foundations and patterns with `createStore` API

## How was this PR tested?
Unit tested

### DIFF
--- a/internal/venice-common/src/main/java/com/linkedin/venice/controller/converters/ConverterUtil.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/controller/converters/ConverterUtil.java
@@ -1,0 +1,57 @@
+package com.linkedin.venice.controller.converters;
+
+import static com.linkedin.venice.controllerapi.ControllerApiConstants.*;
+
+import com.linkedin.venice.controller.requests.ControllerHttpRequest;
+import com.linkedin.venice.controller.requests.CreateStoreRequest;
+import com.linkedin.venice.controllerapi.NewStoreResponse;
+import com.linkedin.venice.controllerapi.QueryParams;
+import com.linkedin.venice.protocols.CreateStoreGrpcRequest;
+import com.linkedin.venice.protocols.CreateStoreGrpcResponse;
+import java.util.Optional;
+
+
+/**
+ * A utility class to hold all the conversion logic from {@link com.linkedin.venice.controller.requests.ControllerRequest}
+ * to transport specific request and transport specific responses back to {@link com.linkedin.venice.controllerapi.ControllerResponse}.
+ * The utility methods are registered with {@link GrpcConvertersRegistry} and {@link HttpConvertersRegistry} for the
+ * runtime to perform appropriate conversions. Refer to the documentation of the above converter registries for more
+ * information.
+ */
+public final class ConverterUtil {
+
+  // store related converters
+  public static CreateStoreGrpcRequest convertCreateStoreToGrpcRequest(CreateStoreRequest request) {
+    CreateStoreGrpcRequest.Builder builder = CreateStoreGrpcRequest.newBuilder()
+        .setStoreName(request.getStoreName())
+        .setKeySchema(request.getKeySchema())
+        .setValueSchema(request.getValueSchema())
+        .setIsSystemStore(request.isSystemStore());
+
+    Optional.ofNullable(request.getOwner()).ifPresent(builder::setOwner);
+    Optional.ofNullable(request.getAccessPermissions()).ifPresent(builder::setAccessPermission);
+
+    return builder.build();
+  }
+
+  public static ControllerHttpRequest convertCreateStoreToHttpRequest(CreateStoreRequest request) {
+    QueryParams params = new QueryParams();
+
+    params.add(NAME, request.getStoreName())
+        .add(OWNER, request.getOwner())
+        .add(KEY_SCHEMA, request.getKeySchema())
+        .add(VALUE_SCHEMA, request.getValueSchema())
+        .add(IS_SYSTEM_STORE, request.isSystemStore())
+        .add(CLUSTER, request.getClusterName());
+
+    Optional.ofNullable(request.getAccessPermissions()).ifPresent(perm -> params.add(ACCESS_PERMISSION, perm));
+
+    return ControllerHttpRequest.newBuilder().setParam(params).build();
+  }
+
+  public static NewStoreResponse convertCreateStoreGrpcResponse(CreateStoreGrpcResponse grpcResponse) {
+    NewStoreResponse response = new NewStoreResponse();
+    response.setOwner(grpcResponse.getOwner());
+    return response;
+  }
+}
--- a/internal/venice-common/src/main/java/com/linkedin/venice/controller/converters/HttpConvertersRegistry.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/controller/converters/HttpConvertersRegistry.java
@@ -1,0 +1,42 @@
+package com.linkedin.venice.controller.converters;
+
+import com.linkedin.venice.controller.requests.ControllerHttpRequest;
+import com.linkedin.venice.controller.requests.ControllerRequest;
+import com.linkedin.venice.controller.requests.CreateStoreRequest;
+import java.util.HashMap;
+import java.util.Map;
+
+
+/**
+ * A registry to hold all the converters for {@link ControllerRequest} to Http specific request.
+ * New addition of converters need to be registered below under the
+ * {@link #initialize()}. e.g.,
+ * <pre>
+ *   // Register sample request converter
+ *   registerRequestConverter(NewApiRequest.class, ConverterUtil::convertNewApiToHttpRequest)
+ * </pre>
+ */
+public class HttpConvertersRegistry {
+  private final Map<Class<?>, RequestConverter<? extends ControllerRequest, ControllerHttpRequest>> requestRegistry =
+      new HashMap<>();
+
+  public HttpConvertersRegistry() {
+    initialize();
+  }
+
+  public void initialize() {
+    registerRequestConverter(CreateStoreRequest.class, ConverterUtil::convertCreateStoreToHttpRequest);
+  }
+
+  private <T extends ControllerRequest> void registerRequestConverter(
+      Class<T> requestType,
+      RequestConverter<T, ControllerHttpRequest> converter) {
+    requestRegistry.computeIfAbsent(requestType, k -> converter);
+  }
+
+  @SuppressWarnings("unchecked")
+  public <T extends ControllerRequest> RequestConverter<T, ControllerHttpRequest> getRequestConverter(
+      Class<T> requestType) {
+    return (RequestConverter<T, ControllerHttpRequest>) requestRegistry.get(requestType);
+  }
+}
--- a/internal/venice-common/src/main/java/com/linkedin/venice/controller/converters/RequestConverter.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/controller/converters/RequestConverter.java
@@ -1,0 +1,15 @@
+package com.linkedin.venice.controller.converters;
+
+import com.linkedin.venice.controller.requests.ControllerRequest;
+
+
+/**
+ * A general interface to represent request converters that operates on {@Link T} to produce a transport specific
+ * request {@link D}
+ * @param <T> Source controller request type
+ * @param <D> Transport specific request type
+ */
+@FunctionalInterface
+public interface RequestConverter<T extends ControllerRequest, D> {
+  D convert(T request);
+}
--- a/internal/venice-common/src/main/java/com/linkedin/venice/controller/converters/ResponseConverter.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/controller/converters/ResponseConverter.java
@@ -1,0 +1,14 @@
+package com.linkedin.venice.controller.converters;
+
+import com.linkedin.venice.controllerapi.ControllerResponse;
+
+
+/**
+ * A general interface to represent response converters that operates on transport specific response {@Link T} to produce
+ * a controller specific responesn {@link D}
+ * @param <T> Source transport specific response type
+ * @param <D> Controller response type
+ */
+public interface ResponseConverter<T, D extends ControllerResponse> {
+  D convert(T response);
+}
--- a/internal/venice-common/src/main/java/com/linkedin/venice/controller/requests/ControllerHttpRequest.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/controller/requests/ControllerHttpRequest.java
@@ -1,0 +1,80 @@
+package com.linkedin.venice.controller.requests;
+
+import com.linkedin.venice.controllerapi.QueryParams;
+import com.linkedin.venice.utils.Time;
+
+
+/**
+ * A general class to represent Http specific transport request that embodies all the Http requests
+ * send to the controller as part of controller APIs. This is a container class defined to help with the
+ * refactoring of ControllerClient to become transport protocol agnostic.
+ */
+public class ControllerHttpRequest {
+  private static final int DEFAULT_REQUEST_TIMEOUT_MS = 600 * Time.MS_PER_SECOND;
+  private static final int DEFAULT_MAX_ATTEMPTS = 10;
+  private final QueryParams params;
+  private final byte[] data;
+
+  private final int timeoutMs;
+
+  private final int maxRetries;
+
+  public ControllerHttpRequest(QueryParams params, byte[] data, int timeoutMs, int maxRetries) {
+    this.params = params;
+    this.data = data;
+    this.timeoutMs = timeoutMs;
+    this.maxRetries = maxRetries;
+  }
+
+  public QueryParams getParams() {
+    return params;
+  }
+
+  public byte[] getData() {
+    return data;
+  }
+
+  public int getTimeoutMs() {
+    return timeoutMs;
+  }
+
+  public int getMaxRetries() {
+    return maxRetries;
+  }
+
+  public static Builder newBuilder() {
+    return new Builder();
+  }
+
+  public static class Builder {
+    private QueryParams params;
+    private byte[] data;
+    private int timeoutMs = DEFAULT_REQUEST_TIMEOUT_MS;
+
+    private int maxRetries = DEFAULT_MAX_ATTEMPTS;
+
+    public Builder setParam(QueryParams params) {
+      this.params = params;
+      return this;
+    }
+
+    public Builder setData(byte[] data) {
+      this.data = data;
+      return this;
+    }
+
+    public Builder setTimeoutMs(int timeoutMs) {
+      this.timeoutMs = timeoutMs;
+      return this;
+    }
+
+    public Builder setMaxRetries(int retries) {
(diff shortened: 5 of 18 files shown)


internal/venice-common/src/main/java/com/linkedin/venice/controller/converters/HttpConvertersRegistry.java
```diff
@@ -0,0 +1,42 @@
...
+ * New addition of converters need to be registered below under the
+ * {@link #initialize()}. e.g.,
+ * <pre>
+ *   // Register sample request converter
+ *   registerRequestConverter(NewApiRequest.class, ConverterUtil::convertNewApiToHttpRequest)
+ * </pre>
+ */
+public class HttpConvertersRegistry {
+  private final Map<Class<?>, RequestConverter<? extends ControllerRequest, ControllerHttpRequest>> requestRegistry =
+      new HashMap<>();
```
FelixGV, line 21: The `GrpcConvertersRegistry` has both request and response converters. Does this one likewise need to have both?

internal/venice-common/src/main/java/com/linkedin/venice/controller/converters/RequestConverter.java
```diff
@@ -0,0 +1,15 @@
+package com.linkedin.venice.controller.converters;
+
+import com.linkedin.venice.controller.requests.ControllerRequest;
+
+
+/**
+ * A general interface to represent request converters that operates on {@Link T} to produce a transport specific
+ * request {@link D}
+ * @param <T> Source controller request type
+ * @param <D> Transport specific request type
```
FelixGV, line 10: What do `T` and `D` stand for? Although it is fine to use single letters for the generic type names, I think it's important to try to make these names as self-explanatory as possible. For example, `I`/`O` which stand for Input/Output might be fine. It is also fine to spell out whole words instead of using single letters, e.g. `SOURCE`/`DESTINATION`, or `SRC`/`DEST`.

internal/venice-common/src/main/java/com/linkedin/venice/controller/converters/ResponseConverter.java
```diff
@@ -0,0 +1,14 @@
+package com.linkedin.venice.controller.converters;
+
+import com.linkedin.venice.controllerapi.ControllerResponse;
+
+
+/**
+ * A general interface to represent response converters that operates on transport specific response {@Link T} to produce
+ * a controller specific responesn {@link D}
+ * @param <T> Source transport specific response type
+ * @param <D> Controller response type
```
FelixGV, line 10: If you decide to change the generic type names in the above interface, then let's adopt the same convention here.

internal/venice-common/src/main/java/com/linkedin/venice/controller/requests/ControllerHttpRequest.java
```diff
@@ -0,0 +1,80 @@
...
+import com.linkedin.venice.controllerapi.QueryParams;
+import com.linkedin.venice.utils.Time;
+
+
+/**
+ * A general class to represent Http specific transport request that embodies all the Http requests
+ * send to the controller as part of controller APIs. This is a container class defined to help with the
+ * refactoring of ControllerClient to become transport protocol agnostic.
+ */
+public class ControllerHttpRequest {
```
FelixGV, line 12: What is the relationship between this class and `ControllerRequest`? Should this one extend the other?

internal/venice-common/src/main/java/com/linkedin/venice/controller/transport/ControllerGrpcTransportClient.java
```diff
@@ -0,0 +1,103 @@
...
+  public <REQ extends ControllerRequest, RES extends ControllerResponse> RES request(REQ request)
+      throws ExecutionException, TimeoutException {
+
+    GrpcRoute grpcRoute = RouteUtils.getGrpcRouteFor(request.getRoute());
+
+    try {
+      VeniceControllerGrpcServiceGrpc.VeniceControllerGrpcServiceBlockingStub stub = buildStub();
+
+      Method method = stub.getClass().getMethod(grpcRoute.getMethod(), grpcRoute.getRequestType());
+      Object responseObject = method.invoke(stub, convertToGrpcRequest(request));
```
FelixGV, line 57: Must we use reflection? Is here not a design pattern that could allow us to plug in the correct stuff without it? e.g. perhaps this "stub" class can return a `Function<REQ, RES>`, or similar?

internal/venice-common/src/main/proto/VeniceControllerGrpcService.proto
```diff
@@ -0,0 +1,85 @@
+syntax = 'proto3';
+package com.linkedin.venice.protocols;
```
FelixGV, line 2: Should we include `controller` in the package name? Either in the last or second to last position?

FelixGV (commented):
Left some comments, which may be a bit superficial in nature. I don't have a solid end-to-end view of the whole work, so maybe some of my suggestions don't make sense. Please take a look and LMK.

Thanks for this work! Looking forward to it landing!

internal/venice-common/src/main/java/com/linkedin/venice/controller/converters/RequestConverter.java
mynameborat: T -> The controller request type (e.g., `storeCreateRequest`, `jobStatusRequest`)
D -> controller transport specific request type (e.g., http, grpc etc)

I did use some of elaborative naming for generic types in few places in the code where I saw no ambiguity. I thought a bit and then wasn't fully convinced if `SRC` and `DEST` was going convey its intent; I even wondered if `SRCREQ` to `DESTREQ` would help.

Given this converter interface is not a general converter interface rather specific to the controller requests, I found the above choices too generic and hence resorted write the parameter java document to provide some clarity.  Do we have any convention on naming generic types within our code base?

For e.g., Can we have something like `ControllerRequestType` `TransportRequestType`?

internal/venice-common/src/main/java/com/linkedin/venice/controller/converters/ResponseConverter.java
mynameborat: Started a thread above and will reconcile this one once we land on something.

internal/venice-common/src/main/java/com/linkedin/venice/controller/requests/ControllerHttpRequest.java
mynameborat: `ControllerRequest` abstraction is meant to decouple the actual transport related information. `ControllerHttpRequest` however, describes how `ControllerRequest` will look like if http was chosen as the medium of transport.

Originally, I thought about having some coupling between `ControllerRequest`, `ControllerHttpRequest` and `GeneratedV3..` (grpc general request type)in the form of having `ControllerRequest` host few abstract methods like 

```
  public abstract ControllerRequest {
      ...
      abstract ControllerHttpRequest convertToHttpRequest();

      abstract <T extends GeneratedV3..> T convertToGrpcRequest();
 }
```

However, the above spills some of the transport details into the model and hence chose against it;

internal/venice-common/src/main/java/com/linkedin/venice/controller/transport/ControllerGrpcTransportClient.java
mynameborat: Good point. Of the 5 days in the hackathon, I spent considerable amount of time to make this generic instead of using reflection. Most of my attempts failed due to type erasure and the need to introduce type token to carry information in a way it could potentially work. 

That said, GRPC way of defining these methods are also sort of either too abstracted or too narrow; e.g., the stub generated with createStore method takes a concrete CreateStoreRequest so having type bounds on `REQ` and `RES` to work seemed a bit more tricky and hence I chose this one for now.

I will continue looking into this and see if `TypeToken` is the way to go and perhaps refactor this out if I land on something promising.

internal/venice-common/src/main/proto/VeniceControllerGrpcService.proto
mynameborat: should be fine; just need to coordinate the change w/ @sushantmane

internal/venice-common/src/main/java/com/linkedin/venice/controller/transport/ControllerGrpcTransportClient.java
FelixGV: Ok well, this is not a hot path so the performance downside of reflection is not a concern... the only concern is the maintainability aspect since we lost strong typing. I guess it could be fine to let this in as is, and keep iterating later.

internal/venice-common/src/main/proto/VeniceControllerGrpcService.proto
FelixGV: We are going to have client/server (or "data plane") protocols as well. It would seem cleaner to me to have a clear separation between control plane and data plane protocols.

internal/venice-common/src/main/java/com/linkedin/venice/controller/converters/RequestConverter.java
FelixGV: `T` usually stands for "Type", which is fine if there is just a single thing to maintain a generic "type" for. In this case there are too things, however, so `T` seems ambiguous. `D` might stand for "Destination" in this context, though I'm not sure?

internal/venice-common/src/main/java/com/linkedin/venice/controller/converters/HttpConvertersRegistry.java
FelixGV: I guess I just don't understand why we need it for one but not the other...?

mynameborat: that is because, we decided to use a separate response data model for gRPC and `ControllerResponse` POJO isn't suitable, unless, we loose the typing information on the gRPC path and have responses represent byte[] in which case we can reuse all of it; 

We decided as a team to follow gRPC convention and make the new request, response strongly typed.

internal/venice-common/src/main/java/com/linkedin/venice/controller/converters/RequestConverter.java
mynameborat: ah i see; `T` is your concern as it by default has the type connotation. Let me update `ControllerRequestType` and `TransportRequestType` if you have no objections. If not, I can fallback to `SRC` and `DEST` which i think is okay along with the java docs

internal/venice-common/src/main/java/com/linkedin/venice/controller/converters/ResponseConverter.java
mynameborat: Will follow the same guidelines as the above comment.

internal/venice-common/src/main/java/com/linkedin/venice/controller/converters/ConverterUtil.java
```diff
@@ -0,0 +1,57 @@
...
+ * A utility class to hold all the conversion logic from {@link com.linkedin.venice.controller.requests.ControllerRequest}
+ * to transport specific request and transport specific responses back to {@link com.linkedin.venice.controllerapi.ControllerResponse}.
+ * The utility methods are registered with {@link GrpcConvertersRegistry} and {@link HttpConvertersRegistry} for the
+ * runtime to perform appropriate conversions. Refer to the documentation of the above converter registries for more
+ * information.
+ */
+public final class ConverterUtil {
+
+  // store related converters
+  public static CreateStoreGrpcRequest convertCreateStoreToGrpcRequest(CreateStoreRequest request) {
```
FelixGV, line 24: Can this code be a function of `CreateStoreRequest`? Perhaps the parent class `ControllerRequest` can become `ControllerRequest<GRPC>` and define a `public abstract GRPC getGRPC()` API. Then `CreateStoreRequest extends ControllerRequest<CreateStoreGrpcRequest>`? Would that eliminate the need for the whole registry?

internal/venice-common/src/main/java/com/linkedin/venice/controller/converters/ConverterUtil.java
```diff
@@ -0,0 +1,57 @@
...
+        .setValueSchema(request.getValueSchema())
+        .setIsSystemStore(request.isSystemStore());
+
+    Optional.ofNullable(request.getOwner()).ifPresent(builder::setOwner);
+    Optional.ofNullable(request.getAccessPermissions()).ifPresent(builder::setAccessPermission);
+
+    return builder.build();
+  }
+
+  public static ControllerHttpRequest convertCreateStoreToHttpRequest(CreateStoreRequest request) {
```
FelixGV, line 37: And similarly, the function to convert into a `ControllerHttpRequest` could likewise live over there?

FelixGV (commented):
One more idea/suggestion about this...

internal/venice-common/src/main/java/com/linkedin/venice/controller/converters/ConverterUtil.java
```diff
@@ -0,0 +1,57 @@
...
+        .add(VALUE_SCHEMA, request.getValueSchema())
+        .add(IS_SYSTEM_STORE, request.isSystemStore())
+        .add(CLUSTER, request.getClusterName());
+
+    Optional.ofNullable(request.getAccessPermissions()).ifPresent(perm -> params.add(ACCESS_PERMISSION, perm));
+
+    return ControllerHttpRequest.newBuilder().setParam(params).build();
+  }
+
+  public static NewStoreResponse convertCreateStoreGrpcResponse(CreateStoreGrpcResponse grpcResponse) {
```
FelixGV, line 52: And similarly to the idea I mentioned in my last review... can this become a function of `NewStoreResponse`? We could have `ControllerResponse<GRPC, R extends ControllerResponse >`, with `NewStoreResponse extends ControllerResponse<CreateStoreGrpcResponse, NewStoreResponse>`? Although the API would be the reverse, e.g. `public R fromGRPC(GRPC) {}` or similar?

internal/venice-common/src/main/java/com/linkedin/venice/controller/converters/ConverterUtil.java
mynameborat: This goes back to my earlier comment - https://github.com/linkedin/venice/pull/1259#discussion_r1884316176

Controller models (responses and requests) taking dependencies on the underlying transport through methods like `toGrpc` and `fromGrpc` would make the consumers of these models exposed to some underlying details (although we could control the access modifier) this also potential brings in the dependencies of the underlying transport dependencies for consumers of models.

Hence the choice to invert the dependency and have these concrete transport implementations depend on the models of controller abstraction and the conversion residing outside these models.

internal/venice-common/src/main/java/com/linkedin/venice/controller/requests/ControllerHttpRequest.java
FelixGV: Hmm... but isn't it Protobuf's design goal to provide a one stop shop for models and their transport? And isn't that also what we do with Jackson (i.e. the model classes have transport-related annotations sprinkled over them). What is the interest in decoupling model from transport? (I understand the purpose of decoupling in the abstract, but I am asking specifically in this context, what does it buy us?)

internal/venice-common/src/main/java/com/linkedin/venice/controller/transport/ControllerGrpcTransportClient.java
sushantmane: Personally, I’m not a fan of using GrpcConvertersRegistry solely to avoid little boilerplate code. I think having some boilerplate is fine as long as it’s simple, easy to maintain, and easy to test.

sushantmane: I believe this could be really be done something like:

```java

// ControllerTransport Interface
public interface ControllerTransport {
    ClusterDiscoveryResponse discoverCluster(ClusterDiscoveryRequest storeName);
    LeaderControllerResponse getLeaderController(GetLeaderControllerRequest clusterName);
}

// ControllerGrpcTransport Implementation
public class ControllerGrpcTransport implements ControllerTransport {
    @Override
    public ClusterDiscoveryResponse discoverCluster(ClusterDiscoveryRequest request) {
        System.out.println("gRPC implementation for discoverCluster");
        // Implementation goes here
        return new ClusterDiscoveryResponse("gRPC: Cluster for store " + request.getStoreName());
    }

    @Override
    public LeaderControllerResponse getLeaderController(GetLeaderControllerRequest request) {
        System.out.println("gRPC implementation for getLeaderController");
        // Implementation goes here
        return new LeaderControllerResponse("gRPC: Leader for cluster " + request.getClusterName());
    }
}

// ControllerHttpTransport Implementation
public class ControllerHttpTransport implements ControllerTransport {
    @Override
    public ClusterDiscoveryResponse discoverCluster(ClusterDiscoveryRequest request) {
        System.out.println("HTTP implementation for discoverCluster");
        // Implementation goes here
        return new ClusterDiscoveryResponse("HTTP: Cluster for store " + request.getStoreName());
    }

    @Override
    public LeaderControllerResponse getLeaderController(GetLeaderControllerRequest request) {
        System.out.println("HTTP implementation for getLeaderController");
        // Implementation goes here
        return new LeaderControllerResponse("HTTP: Leader for cluster " + request.getClusterName());
    }
}
```

internal/venice-common/src/main/proto/VeniceControllerGrpcService.proto
sushantmane: Yeah this fine. I've already made this change in my PR: https://github.com/linkedin/venice/pull/1396. Could you please pickup updated proto files?

internal/venice-common/src/main/java/com/linkedin/venice/controller/transport/ControllerGrpcTransportClient.java
```diff
@@ -0,0 +1,103 @@
...
+    this.converterRegistry = new GrpcConvertersRegistry();
+  }
+
+  public <REQ extends ControllerRequest, RES extends ControllerResponse> RES request(REQ request)
+      throws ExecutionException, TimeoutException {
+
+    GrpcRoute grpcRoute = RouteUtils.getGrpcRouteFor(request.getRoute());
+
+    try {
+      VeniceControllerGrpcServiceGrpc.VeniceControllerGrpcServiceBlockingStub stub = buildStub();
```
sushantmane, line 54: We've merged several PRs introducing server-side changes. In the latest updates, we've split APIs into separate services (e.g., StoreGrpcService and ClusterAdminOpsGrpcService). Going forward, additional services will be introduced to ensure clear separation of responsibilities and avoid overloading a single service.

With these changes, the client-side can no longer rely on a single stub. Instead, service-specific stubs will need to be used for each API. However, we can still reuse the same channels for these different APIs.

sushantmane:
@mynameborat - When you get a chance, could you please rebase and update the PR to incorporate the latest server-side changes? Thanks!

PR #1261, opened 2024-10-25
[server] Add store-aware partition-wise shared consumer assignment strategy

sixpluszero:
## [server] Add store-aware partition-wise shared consumer assignment strategy

Add a new version of partition wise shared consumer assignment strategy.
We have been seeing subscriptions to the same topic / store be assigned to the same consumer, and for a particular store push's view (can be inc push / full push) it can be competing with each other and becomes the long-tail partition and slow down the overall progress. 
Assuming the store/topic itself does not have data-skew, then we should try to assign these subscriptions to different consumers as even as possible. Especially for RT topics, backup / current / future version will share the same input volume, so we should not treat them differently within the same pool, but we can further optimize that in different level (Pool assignment strategy) 
This PR adds the new strategy so when a new topic partition is looking for assignment, it will compute and sort all the consumer's load based on general load and the store-specific load. It will assign the new topic partition to the least loaded consumer based on the computed load.

## How was this PR tested?

Add a new unit test, will add integration test as well.

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreAwarePartitionWiseKafkaConsumerService.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreAwarePartitionWiseKafkaConsumerService.java
@@ -2,6 +2,7 @@ package com.linkedin.davinci.kafka.consumer;
 
 import com.linkedin.davinci.stats.AggKafkaConsumerServiceStats;
 import com.linkedin.venice.meta.ReadOnlyStoreRepository;
+import com.linkedin.venice.meta.Store;
 import com.linkedin.venice.meta.Version;
 import com.linkedin.venice.pubsub.PubSubConsumerAdapterFactory;
 import com.linkedin.venice.pubsub.api.PubSubConsumerAdapter;
@@ -20,16 +21,20 @@ import org.apache.logging.log4j.LogManager;
 
 
 /**
- * {@link StoreAwarePartitionWiseKafkaConsumerService} is used to allocate share consumer from consumer pool at partition granularity.
- * One shared consumer may have multiple topics, and each topic may have multiple consumers.
+ * {@link StoreAwarePartitionWiseKafkaConsumerService} is used to allocate share consumer from consumer pool at partition
+ * granularity. One shared consumer may have multiple topics, and each topic may have multiple consumers.
  * This is store-aware version of topic-wise shared consumer service. The topic partition assignment in this service has
- * a heuristic that we should distribute the all the subscriptions related to a same store as even as possible.
- * The load calculation for each consumer will be consumer assignment size + STORE_SUBSCRIPTION_LOAD * subscription count
- * that belongs to the same store, and we will pick the least loaded consumer for a new topic partition request.
+ * a heuristic that we should distribute the all the subscriptions related to a same store / version as even as possible.
+ * The load calculation for each consumer will be:
+ * Hybrid store: Consumer assignment size + STORE_SUBSCRIPTION_LOAD * subscription count for the same store;
+ * Batch only store: Consumer assignment size + STORE_SUBSCRIPTION_LOAD * subscription count for the same store version;
+ * and we will pick the least loaded consumer for a new topic partition request.
  */
 public class StoreAwarePartitionWiseKafkaConsumerService extends PartitionWiseKafkaConsumerService {
   private static final long DEFAULT_STORE_SUBSCRIPTION_LOAD = 100;
-  private final Map<String, Map<PubSubConsumerAdapter, Integer>> storeToConsumerMap = new VeniceConcurrentHashMap<>();
+  private final Map<String, Map<PubSubConsumerAdapter, Integer>> resourceIdentifierToConsumerMap =
+      new VeniceConcurrentHashMap<>();
+  private final ReadOnlyStoreRepository storeRepository;
 
   StoreAwarePartitionWiseKafkaConsumerService(
       final ConsumerPoolType poolType,
@@ -71,12 +76,14 @@ public class StoreAwarePartitionWiseKafkaConsumerService extends PartitionWiseKa
         isUnregisterMetricForDeletedStoreEnabled);
     this.LOGGER =
         LogManager.getLogger(StoreAwarePartitionWiseKafkaConsumerService.class + " [" + kafkaUrlForLogger + "]");
+    this.storeRepository = metadataRepository;
   }
 
   @Override
   protected synchronized SharedKafkaConsumer pickConsumerForPartition(
       PubSubTopic versionTopic,
       PubSubTopicPartition topicPartition) {
+    String resourceIdentifier = getResourceIdentifier(versionTopic);
     Map<Integer, Long> consumerToLoadMap = new VeniceConcurrentHashMap<>();
     for (int i = 0; i < consumerToConsumptionTask.size(); i++) {
       SharedKafkaConsumer consumer = consumerToConsumptionTask.getByIndex(i).getKey();
@@ -90,8 +97,9 @@ public class StoreAwarePartitionWiseKafkaConsumerService extends PartitionWiseKa
         continue;
       }
       long baseAssignmentCount = consumer.getAssignmentSize();
-      long storeSubscriptionCount = storeToConsumerMap.getOrDefault(versionTopic.getStoreName(), Collections.emptyMap())
-          .getOrDefault(consumer, 0);
+      long storeSubscriptionCount =
+          resourceIdentifierToConsumerMap.getOrDefault(resourceIdentifier, Collections.emptyMap())
+              .getOrDefault(consumer, 0);
       long overallLoad = storeSubscriptionCount * DEFAULT_STORE_SUBSCRIPTION_LOAD + baseAssignmentCount;
       consumerToLoadMap.put(i, overallLoad);
     }
@@ -112,10 +120,9 @@ public class StoreAwarePartitionWiseKafkaConsumerService extends PartitionWiseKa
     }
 
     SharedKafkaConsumer consumer = consumerToConsumptionTask.getByIndex(index).getKey();
-    // Update store to consumer load map.
-    String storeName = versionTopic.getStoreName();
+    // Update resource identifier to consumer load map.
     Map<PubSubConsumerAdapter, Integer> consumerMap =
-        storeToConsumerMap.computeIfAbsent(storeName, key -> new VeniceConcurrentHashMap<>());
+        resourceIdentifierToConsumerMap.computeIfAbsent(resourceIdentifier, key -> new VeniceConcurrentHashMap<>());
     int existCount = consumerMap.getOrDefault(consumer, 0);
     consumerMap.put(consumer, existCount + 1);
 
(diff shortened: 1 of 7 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreAwarePartitionWiseKafkaConsumerService.java
```diff
@@ -0,0 +1,200 @@
...
+        getResourceIdentifierToConsumerMap().remove(resourceIdentifier);
+      }
+    } else {
+      getResourceIdentifierToConsumerMap().get(resourceIdentifier).put(consumer, count - 1);
+    }
+  }
+
+  String getResourceIdentifier(PubSubTopic topic) {
+    String storeName = topic.getStoreName();
+    Store store = getStoreRepository().getStore(storeName);
```
haoxu07, line 173: We rely on zk `store` to figure out the resource identifier, as the zk `store` could change due to hybrid enabling or disabling, this will cause resource identifier change for the same version topic, as we need to maintain a map of `resourceIdentifierToConsumerMap` for subscribe or unsubscribe, this might cause some problem..

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreAwarePartitionWiseKafkaConsumerService.java
```diff
@@ -0,0 +1,200 @@
...
+            i,
+            topicPartition);
+        consumerToLoadMap.put(i, Long.MAX_VALUE);
+        continue;
+      }
+      long baseAssignmentCount = consumer.getAssignmentSize();
+      long storeSubscriptionCount =
+          getResourceIdentifierToConsumerMap().getOrDefault(resourceIdentifier, Collections.emptyMap())
+              .getOrDefault(consumer, 0);
+      long overallLoad = storeSubscriptionCount * DEFAULT_STORE_SUBSCRIPTION_LOAD + baseAssignmentCount;
```
haoxu07, line 103: Seems the resourceIdentifier only cares about the versionTopic passing here: . For leader and follower partitions, there might be such a case for when leader and follow sharing same pool: 

assume there are 2 consumers, subscription comes in this order (leader rt topic partition 1, follower vt partition 2, leader rt topic partition 3), so ideally leader rt topic partition will get consumer 1 and follower vt partition 2 get consumer 2, but leader rt topic partition 3 will still goes to consumer 1.  Consumer 1 serves 2 leader rt topic partition, but consumer 2 serves 0 leader rt topic partition. So there still some long-tail issue if we want to evenly distribute the leader partition.

haoxu07 (commented):
Thanks for preparing this change! Left some comments, we might need a integration test to verify this: we could leverage previous `testHybridWithPartitionWiseConsumer`'s change or ingestion context debugging funcationality to verify the consumer topic partition assingment.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreAwarePartitionWiseKafkaConsumerService.java
sixpluszero: Hmm, I am looking at version level hybrid config, I will only look at store level config when version is not found.

haoxu07: Oh，you are right. If store version level config, then that should not cause different identifier. But ideally version should exist when we subscribe a topic partition for that store version, can we throw IllegalStateException?

sixpluszero: Synced offline, we decided to keep as it is right now

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreAwarePartitionWiseKafkaConsumerService.java
```diff
@@ -0,0 +1,200 @@
...
+        sharedConsumerNonExistingTopicCleanupDelayMS,
+        topicExistenceChecker,
+        liveConfigBasedKafkaThrottlingEnabled,
+        pubSubDeserializer,
+        time,
+        stats,
+        isKafkaConsumerOffsetCollectionEnabled,
+        metadataRepository,
+        isUnregisterMetricForDeletedStoreEnabled);
+    this.LOGGER =
```
gaojieliu, line 77: This is not a good pattern as the constructor of the super class will construct a `LOGGER` and get overrided right away here.
We can pass a `Logger name` into the constructor, so that only the super class will be in charge of logger initialization, no here.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreAwarePartitionWiseKafkaConsumerService.java
```diff
@@ -0,0 +1,200 @@
...
+ * This is store-aware version of topic-wise shared consumer service. The topic partition assignment in this service has
+ * a heuristic that we should distribute the all the subscriptions related to a same store / version as even as possible.
+ * The load calculation for each consumer will be:
+ * Hybrid store: Consumer assignment size + STORE_SUBSCRIPTION_LOAD * subscription count for the same store;
+ * Batch only store: Consumer assignment size + STORE_SUBSCRIPTION_LOAD * subscription count for the same store version;
+ * and we will pick the least loaded consumer for a new topic partition request.
+ */
+public class StoreAwarePartitionWiseKafkaConsumerService extends PartitionWiseKafkaConsumerService {
+  private static final long DEFAULT_STORE_SUBSCRIPTION_LOAD = 100;
+  private final Map<String, Map<PubSubConsumerAdapter, Integer>> resourceIdentifierToConsumerMap =
```
gaojieliu, line 35: The naming of this field is not good enough to describe the purpose, so can you leave some javadoc?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreAwarePartitionWiseKafkaConsumerService.java
sixpluszero: Also synced offline, we can start with this, as we mostly care about heavy write workload pool's which is AA/WC pool, and for that LEADER has already been picked out.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreAwarePartitionWiseKafkaConsumerService.java
```diff
@@ -0,0 +1,200 @@
...
+  }
+
+  @Override
+  protected synchronized SharedKafkaConsumer pickConsumerForPartition(
+      PubSubTopic versionTopic,
+      PubSubTopicPartition topicPartition) {
+    String resourceIdentifier = getResourceIdentifier(versionTopic);
+    Map<Integer, Long> consumerToLoadMap = new VeniceConcurrentHashMap<>();
+    for (int i = 0; i < getConsumerToConsumptionTask().size(); i++) {
+      SharedKafkaConsumer consumer = getConsumerToConsumptionTask().getByIndex(i).getKey();
```
gaojieliu, line 89: Why not use some `iterative` pattern to loop the map?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreAwarePartitionWiseKafkaConsumerService.java
```diff
@@ -0,0 +1,200 @@
...
+      if (topicPartition.getPubSubTopic().isRealTime()
+          && alreadySubscribedRealtimeTopicPartition(consumer, topicPartition)) {
+        getLOGGER().info(
+            "Consumer ID: {} has already subscribed the same real time topic-partition: {}, will assign MAX load to avoid being picked.",
+            i,
+            topicPartition);
+        consumerToLoadMap.put(i, Long.MAX_VALUE);
```
gaojieliu, line 96: We should skip such consumer from the map, right?
In some extreme scenario, all the consumers are marked as `Long.MAX_VALUE`, the following logic will still assign, which is wrong.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreAwarePartitionWiseKafkaConsumerService.java
```diff
@@ -0,0 +1,200 @@
...
+        LogManager.getLogger(StoreAwarePartitionWiseKafkaConsumerService.class + " [" + kafkaUrlForLogger + "]");
+    this.storeRepository = metadataRepository;
+  }
+
+  @Override
+  protected synchronized SharedKafkaConsumer pickConsumerForPartition(
+      PubSubTopic versionTopic,
+      PubSubTopicPartition topicPartition) {
+    String resourceIdentifier = getResourceIdentifier(versionTopic);
+    Map<Integer, Long> consumerToLoadMap = new VeniceConcurrentHashMap<>();
```
gaojieliu, line 87: `Map<Integer, Integer>` should be more than enough, right?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreAwarePartitionWiseKafkaConsumerService.java
```diff
@@ -0,0 +1,200 @@
...
+    }
+    if (count == 1) {
+      getResourceIdentifierToConsumerMap().get(resourceIdentifier).remove(consumer);
+      if (getResourceIdentifierToConsumerMap().get(resourceIdentifier).isEmpty()) {
+        getResourceIdentifierToConsumerMap().remove(resourceIdentifier);
+      }
+    } else {
+      getResourceIdentifierToConsumerMap().get(resourceIdentifier).put(consumer, count - 1);
+    }
+  }
```
gaojieliu, line 169: I see a few problems here:
1. This function is not synchronized and it can be invoked concurrently and also `pickConsumerForPartition` would read the variables being updated here, so how can we guarantee correctness?
2. There are duplicate lookups here, and we should lookup once and continue to use it throughout the whole function.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreAwarePartitionWiseKafkaConsumerService.java
sixpluszero: Hmm, I did throw IlligalStateException in the below actually.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreAwarePartitionWiseKafkaConsumerService.java
sixpluszero: Good point, I missed unsubscribe path. I only looked at subscribe path and think that's good enough. Updated.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreAwarePartitionWiseKafkaConsumerService.java
sixpluszero: lol I think I get what you mean in the other comment. there is no need to create map and sort at all, just find the minimum load and corresponding consumer should be good

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreAwarePartitionWiseKafkaConsumerService.java
```diff
@@ -0,0 +1,196 @@
...
+    String resourceIdentifier = getResourceIdentifier(versionTopic);
+    int minLoad = Integer.MAX_VALUE;
+    SharedKafkaConsumer minLoadConsumer = null;
+    Map<Integer, Integer> consumerToLoadMap = new VeniceConcurrentHashMap<>();
+    for (SharedKafkaConsumer consumer: getConsumerToConsumptionTask().keySet()) {
+      int index = getConsumerToConsumptionTask().indexOf(consumer);
+      if (topicPartition.getPubSubTopic().isRealTime()
+          && alreadySubscribedRealtimeTopicPartition(consumer, topicPartition)) {
+        getLOGGER().info(
+            "Consumer id: {} has already subscribed the same real time topic-partition: {}, will assign MAX load to avoid being picked.",
```
gaojieliu, line 98: Nit,
```
will assign MAX load to avoid being picked
```
I guess we should mention this consumer will be `skipped`, right?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreAwarePartitionWiseKafkaConsumerService.java
```diff
@@ -0,0 +1,196 @@
...
+    }
+    if (minLoad == Integer.MAX_VALUE) {
+      throw new IllegalStateException(
+          "Unable to find least loaded consumer entry. Current load map: " + consumerToLoadMap);
+    }
+    // Update resource identifier to consumer load map.
+    Map<PubSubConsumerAdapter, Integer> consumerMap = getResourceIdentifierToConsumerMap()
+        .computeIfAbsent(resourceIdentifier, key -> new VeniceConcurrentHashMap<>());
+    int existCount = consumerMap.getOrDefault(minLoadConsumer, 0);
+    consumerMap.put(minLoadConsumer, existCount + 1);
```
gaojieliu, line 122: I am wondering what the main reason is to maintain this map considering the above logic will count it every time when picking up the consumer.
Can we remove this map completely?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreAwarePartitionWiseKafkaConsumerService.java
```diff
@@ -0,0 +1,196 @@
...
+        getConsumerToConsumptionTask().indexOf(minLoadConsumer),
+        minLoadConsumer.getAssignmentSize(),
+        minLoad,
+        topicPartition,
+        versionTopic);
+    return minLoadConsumer;
+  }
+
+  @Override
+  void handleUnsubscription(
```
gaojieliu, line 140: Unsubscription can happen concurrently with function: `pickConsumerForPartition`, that means the logic in `pickConsumerForPartition` won't be accurate, so is it expected?
Maybe it doesn't matter that much, but I was wondering whether we have considered such behavior or not.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreAwarePartitionWiseKafkaConsumerService.java
sixpluszero: Hmm, the above calculation reads the existing counting from the map. Are you suggesting, instead of maintaining a map, everytime we just iterate through the set of the assignment and calculate it ad-hoc?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreAwarePartitionWiseKafkaConsumerService.java
sixpluszero: Yes, I can update the doc to be more clear.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreAwarePartitionWiseKafkaConsumerService.java
sixpluszero: I am actually thinking that - if we simplify the problem to always use store name as identifier regardless of hybrid store or not, then we can remove this map, as each check will only extract store name from assignment. Otherwise, we will have to always refer to store repo for every assignment to compute the identifier, if we don't keep this map.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreAwarePartitionWiseKafkaConsumerService.java
gaojieliu: yeah, they are guarded per shared consumer instance.
I guess this would be fine as the partition assignment strategy is still a best effort.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreAwarePartitionWiseKafkaConsumerService.java
gaojieliu: Got it.
I am fine to pursue the simplified approach as we do have plan to unsubscribe batch-only store after completion and make it as the default setup.
Also batch-only store won't show up in AA/WC leader consumer pool, which is the main concern.

gaojieliu (approved):
Thanks!

PR #1262, opened 2024-10-25
[dvc][server] Support separate inc push topic with different pubsub entries

adamxchen:
## Background
In the effort of supporting separate inc push topic feature and persisting its state on disk, in addition, to minimize schema evolution and code changes, we decided to embed state of "separate inc push topic" into existing fields, e.g. `upstreamOffsetMap` of `PartitionState`. The `upstreamOffsetMap` currently has # of entries that correspond to the number of regions we've built. However the problem is the key of the map is the pubsub server url. Technically, in the same region, the pubsub server url is used for both RT and separate inc push topic so we needs to append a special suffix "_sep" as new keys to differentiate. This also implies a new cluster id, a new alias.

## Summary

As describe in the background, while we have new cluster id, new alias, and new pubsub server url for separate inc push topic, in some certain circumstances, it gotta be resolved to the `normal` form in order for pubsub subscription.
This PR makes sure these values will be resolved as needed in the consumer thread and drainer thread. 
I checked `kafka url`and `cluster id` usage in every downstream method and confirmed only these places are needed to resolve from new values to the old ones:
1. Need to obtain `KafkaConsumerService` and `TopicManager`
2. Heartbeat tracking
3. RegionHybridConsumptionStats tracking
~~4. Validation prior to producing to local pubsub~~

## How was this PR tested?

Pending new unit tests.


clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -2421,12 +2425,14 @@ protected DelegateConsumerRecordResult delegateConsumerRecord(
 
       // If we are here the message must be produced to local kafka or silently consumed.
       LeaderProducedRecordContext leaderProducedRecordContext;
-
+      // No need to resolve cluster id and kafka url because sep topics are real time topic and it's not VT
       validateRecordBeforeProducingToLocalKafka(consumerRecord, partitionConsumptionState, kafkaUrl, kafkaClusterId);
```
adamxchen, line 2429: One thing not clear to me yet is if we resolve the cluster id and url here, which are supposed to be no-op, a couple integ tests would fail with or without my change. So I'm sure by design, we are not supposed to use kafka url resolver here but I don't know why. 😕

sixpluszero: Although we can resolve later, could you point out which test will be failing for example?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StorePartitionDataReceiver.java
```diff
@@ -135,4 +135,9 @@ public String toString() {
     return this.getClass().getSimpleName() + "{" + "VT=" + storeIngestionTask.getVersionTopic() + ", topicPartition="
         + topicPartition + '}';
   }
+
+  // for testing purpose only
+  public int getKafkaClusterId() {
```
sixpluszero, line 140: nit: If it is for unit test then it can just be default modifier for least exposure.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/config/VeniceClusterConfig.java
```diff
@@ -93,6 +94,8 @@ public class VeniceClusterConfig {
   private final Map<String, SecurityProtocol> kafkaBootstrapUrlToSecurityProtocol;
   private final Optional<SSLConfig> sslConfig;
 
+  private static final String SEPARATE_TOPIC_SUFFIX = "_sep";
```
sixpluszero, line 97: nit: can reuse Utils's public static variable.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -3423,12 +3425,16 @@ public void consumerBatchUnsubscribe(Set<PubSubTopicPartition> topicPartitionSet
   public abstract void consumerUnSubscribeAllTopics(PartitionConsumptionState partitionConsumptionState);
 
   public void consumerSubscribe(PubSubTopicPartition pubSubTopicPartition, long startOffset, String kafkaURL) {
-    final boolean consumeRemotely = !Objects.equals(kafkaURL, localKafkaServer);
+    // TODO: pass special format of kafka url as input here when subscribe to the separate incremental push topic
+    String resolvedKafkaURL = kafkaClusterUrlResolver != null ? kafkaClusterUrlResolver.apply(kafkaURL) : kafkaURL;
```
sixpluszero, line 3429: Just to confirm for future work: this is the remaining work we will need to change in the next phase of the implementation right? Because as I see it is still subscribing to original context.

sixpluszero (approved):
Thank you Adam for adding this! I left 1 question and a few nits, but it is mostly ready to merge.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
adamxchen: Those are:
IntegTest_1220
```
1. TestPushJobWithEmergencySourceRegionSelection#testNativeReplicationForBatchPushWithEmergencySourceOverride[0](50, 2)
2. TestPushJobWithNativeReplication#testTargetedRegionPushJobFullConsumptionForBatchStore
3. TestPushJobWithSourceGridFabricSelection#testPushJobWithSourceGridFabricSelection[0](50, 2)

```
and
```
IntegTest_1050
1. TestActiveActiveReplicationForIncPush#testAAReplicationForIncrementalPushToRT[0]
```

clients/da-vinci-client/src/main/java/com/linkedin/davinci/config/VeniceClusterConfig.java
adamxchen: This is no longer used so I am going to remove it

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
adamxchen: That's right. When SIT needs to subscribe sep topic, it should pass the kafka topic that has "_sep" suffix to this method such that the corresponding `KafkaConsumerServic` and `TopicManager` will be used and the `StorePartitionDataReceiver` will also be spin up.

sixpluszero (approved):
LGTM, thank you!

PR #1263, opened 2024-10-26
[server][test] Race condition fix for re-subscription of real-time topic partitions.

haoxu07:

This improvement addresses a previously reverted https://github.com/linkedin/venice/pull/1192. The unit test in the prior fix triggered separate threads resubscribing to the same real-time topic partition, resulting in frequent resubscriptions. These frequent resubscriptions caused excessive metric emissions, leading to GC issues and causing other unit tests to fail in CI. To prevent metric emissions, we use a mocked `AggKafkaConsumerServiceStats` instead of passing null when instantiating `KafkaConsumerService` in the unit test, as `KafkaConsumerService` would otherwise create a real `AggKafkaConsumerServiceStats` when null is passed.

Previous https://github.com/linkedin/venice/pull/1192's commit message: 

> Recently we introduced re-subscription feature for different store versions. Leader and follower partitions from different version topics will experience re-subscription triggered by store version role change concurrently.
> 
> Problem:
> Two StoreIngestionTask threads for different store versions (store version 1 and store version 2) try to do re-subscription (unsub and sub) for the same real-time topic partition concurrently. During re-subscription triggered by one store version, operation of consumer.unSubscribe to remove assignment of the topic partition and operation of consumerToConsumptionTask.get(consumer).removeDataReceiver(pubSubTopicPartition) are sequentially executed. It is possible that StoreIngestionTask thread for store version 2 got the same ConsumptionTask, but DataReceiver inside ConsumptionTask is still from store version 1). As each ConsumptionTask will only allow one DataReceiver for one particular real-time topic partition, the DataReceiver from store version 2 will not be able to be added to this ConsumptionTask.
> 
> Solution:
> Using SharedKafkaConsumer level lock to protect KafkaConsumcerService for unSubscribe, SetDataReceiver, unSubscribeAll to guarantee there will one real-time partition from specific store version doing DataReceiver related assignment change.

## How was this PR tested?

internal CI

### DIFF
--- a/clients/da-vinci-client/src/test/java/com/linkedin/davinci/kafka/consumer/KafkaConsumerServiceDelegatorTest.java
+++ b/clients/da-vinci-client/src/test/java/com/linkedin/davinci/kafka/consumer/KafkaConsumerServiceDelegatorTest.java
@@ -11,8 +11,7 @@ import static org.mockito.Mockito.never;
 import static org.mockito.Mockito.reset;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
-import static org.testng.Assert.assertFalse;
-import static org.testng.Assert.assertTrue;
+import static org.testng.Assert.*;
 
 import com.linkedin.davinci.config.VeniceServerConfig;
 import com.linkedin.davinci.ingestion.consumption.ConsumedDataReceiver;
@@ -494,7 +493,7 @@ public class KafkaConsumerServiceDelegatorTest {
    * condition is encountered or the test times out (30 seconds).
    * TODO: removing the invocationCount after the PR is approved.
    */
-  @Test(invocationCount = 20)
+  @Test(invocationCount = 100)
   public void testKafkaConsumerServiceResubscriptionConcurrency() throws Exception {
     ApacheKafkaConsumerAdapter consumer1 = mock(ApacheKafkaConsumerAdapter.class);
     PubSubConsumerAdapterFactory factory = mock(PubSubConsumerAdapterFactory.class);
@@ -584,6 +583,7 @@ public class KafkaConsumerServiceDelegatorTest {
     for (Thread infiniteSubUnSubThread: infiniteSubUnSubThreads) {
       infiniteSubUnSubThread.interrupt();
       infiniteSubUnSubThread.join();
+      assertEquals(Thread.State.TERMINATED, infiniteSubUnSubThread.getState());
     }
     Assert.assertFalse(
         raceConditionFound,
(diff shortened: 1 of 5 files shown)


clients/da-vinci-client/src/test/java/com/linkedin/davinci/kafka/consumer/KafkaConsumerServiceDelegatorTest.java
```diff
@@ -458,4 +482,141 @@ private void verifyConsumerServiceStartConsumptionIntoDataReceiver(
...
+          partitionReplicaIngestionContext,
+          consumedDataReceiver,
+          countDownLatch);
+      Thread infiniteSubUnSubThread = new Thread(infiniteSubUnSub, "infiniteResubscribe: " + topicV1ForStoreName3);
+      infiniteSubUnSubThread.start();
+      infiniteSubUnSubThreads.add(infiniteSubUnSubThread);
+    }
+
+    long currentTime = System.currentTimeMillis();
+    Boolean raceConditionFound = countDownLatch.await(30, TimeUnit.SECONDS);
```
sushantmane, line 574: can we check thread state was not `NEW` or `TERMINATED` before and after `Boolean raceConditionFound = countDownLatch.await(30, TimeUnit.SECONDS);`?

clients/da-vinci-client/src/test/java/com/linkedin/davinci/kafka/consumer/KafkaConsumerServiceDelegatorTest.java
```diff
@@ -458,4 +483,149 @@ private void verifyConsumerServiceStartConsumptionIntoDataReceiver(
...
+          delegator,
+          partitionReplicaIngestionContext,
+          consumedDataReceiver,
+          countDownLatch);
+      Thread infiniteSubUnSubThread = new Thread(infiniteSubUnSub, "infiniteResubscribe: " + versionTopicForStoreName3);
+      infiniteSubUnSubThread.start();
+      infiniteSubUnSubThreads.add(infiniteSubUnSubThread);
+      // Wait for the thread to start.
+      waitForNonDeterministicAssertion(10, TimeUnit.SECONDS, () -> {
+        assertTrue(
```
sushantmane, line 574: let's add it after countDownLatch.await without waitForNonDeterministicAssertion

clients/da-vinci-client/src/test/java/com/linkedin/davinci/kafka/consumer/KafkaConsumerServiceDelegatorTest.java
sushantmane: I still don't see thread states being checked after `Boolean raceConditionFound = countDownLatch.await(30, TimeUnit.SECONDS)`. Not sure if it's intentional though

haoxu07: Oh, I see your point. You mean I need to check  timeout instead of doing `assertEquals(Thread.State.TERMINATED, infiniteSubUnSubThread.getState());` after killing.

sushantmane (approved):
LGTM, Thanks!

PR #1266, opened 2024-10-28
[server][dvc] Enable parallel transfers for each partition during blob transfer.

jingy-li:
## [server][dvc] Enable parallel transfers for each partition during blob transfer.

**Major Changes:**
Previously, blob transfers occurred sequentially for each partition. Even in parallel subscription, the NettyP2PBlobTransferManager would only return a future once the transfer was complete for error handling, resulting in a blocking process. This PR updates the process to be non-blocking.

**Minor Changes:**
(1) In the event of a connection issue, we will no longer attempt a retry; instead, we will proceed to the next host.
(2) For unknown or unexpected errors, instead of abandoning the blob transfer and executing Kafka ingestion, the process will now move on to the next host for the blob transfer.
(3) If files transfer exceeds the allowed time limit, the P2PFileTransferServerHandler will send a timeout response and will cease sending any files. 

## How was this PR tested?
unit test and integration test.


clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/NettyP2PBlobTransferManager.java
```diff
@@ -41,132 +57,143 @@ public NettyP2PBlobTransferManager(
...
+      resultFuture.completeExceptionally(new VenicePeersNotFoundException(errorMsg));
+      return resultFuture;
+    }
+
+    List<String> discoverPeers = response.getDiscoveryResult();
+    LOGGER
+        .info("Discovered peers {} for store {} version {} partition {}", discoverPeers, storeName, version, partition);
+
+    // 2: Process peers sequentially to fetch the blob
+    processPeersSequentially(discoverPeers, storeName, version, partition, resultFuture);
```
adamxchen, line 87: Why not just return the future from this method so we can eliminate the additional argument?

adamxchen (commented):
Some initial comments. I need to check some details in another round

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/NettyP2PBlobTransferManager.java
```diff
@@ -26,147 +25,179 @@
...
+                  String.format(
+                      FETCHED_BLOB_SUCCESS_MSG_FORMAT,
+                      chosenHost,
+                      storeName,
+                      partition,
+                      version,
+                      Duration.between(startTime, Instant.now()).getSeconds()));
+              resultFuture.complete(inputStream);
+            })
+            .exceptionally(ex -> {
```
sixpluszero, line 160: I am a bit lost in the future chaining, could you clarify about this: thenAccept returns a new future, but exceptionally  is applied on the new future, instead of the original future. Is my understanding correct?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/NettyP2PBlobTransferManager.java
```diff
@@ -26,147 +25,179 @@
  */
 public class NettyP2PBlobTransferManager implements P2PBlobTransferManager<Void> {
   private static final Logger LOGGER = LogManager.getLogger(NettyP2PBlobTransferManager.class);
-  protected static final int MAX_RETRIES_FOR_BLOB_TRANSFER_PER_HOST = 3;
-  protected static final int MAX_TIMEOUT_FOR_BLOB_TRANSFER_IN_MIN = 60;
+  // Log messages format
+  private static final String NO_PEERS_FOUND_ERROR_MSG_FORMAT =
```
sixpluszero, line 29: nit: Thank you for adding the log pattern formatting - since you already try to format the logging, could we use Utils.getReplicaId() API to log all replica logs? This will help us to build log dashboard to filter all ingestion related logs for a certain replica, when we want to have a clear timeline to analyze.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/NettyP2PBlobTransferManager.java
jingy-li: `thenAccept` is called when the CompletableFuture completes successfully (i.e., without throwing an exception), whereas `exceptionally` is invoked when it completes with an exception.

These represent two exclusive cases: one for success and one for exceptions. The `exceptionally` method is used to handle exceptions from the original future, not from the new future created by `thenAccept`.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/NettyP2PBlobTransferManager.java
adamxchen: +1

majisourav99: IMO having too many format specifiers could be confusing and leads to wrong order of types which can cause prod issues. For log4j reporting using `{}` is a safer choice. Only for exception messages we might need them, which again can be skipped using string concatenation.

jingy-li: Thank you @sixpluszero @adamxchen @majisourav99 for the suggestion. PR is updated with using `Utils.getReplicaId()` in log and using `{}` instead.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/NettyP2PBlobTransferManager.java
sixpluszero: Synced offline, you are correct. The thenApply won't be executed and no side effect is generated so exceptioanlly should apply on original completion stage or future.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/NettyP2PBlobTransferManager.java
```diff
@@ -26,147 +26,171 @@
...
-    if (response == null || response.isError()) {
-      throw new VenicePeersNotFoundException("Failed to obtain the peers for the requested blob");
-    }
+  private void processPeersSequentially(
+      List<String> peers,
+      String storeName,
+      int version,
+      int partition,
+      CompletableFuture<InputStream> resultFuture) {
+    String replicaId = Utils.getReplicaId(storeName + "_v" + version, partition);
```
sixpluszero, line 123: Let's use Version.composeKakaTopic() for concat store and version.

sixpluszero (dismissed):
Overall looks good to me, please address the one formatting issue and it should be good to go.

sixpluszero (approved):
LGTM, thank you!

clients/da-vinci-client/src/test/java/com/linkedin/davinci/blobtransfer/TestNettyP2PBlobTransferManager.java
```diff
@@ -55,6 +56,12 @@ public class TestNettyP2PBlobTransferManager {
   String TMP_PARTITION_DIR = "tmpPartitionDir";
   int TEST_VERSION = 1;
   int TEST_PARTITION = 0;
+  Path file1;
```
adamxchen, line 59: nit: you can make them fewer lines

adamxchen (approved):
Thank you a lot!

PR #1269, opened 2024-10-30
[test] Generate test CI report for E2E tests in local fork

majisourav99:
## Generate CI report for E2E tests.

TODO: generate report for unit tests and codecov


internal/venice-test-common/build.gradle
```diff
@@ -312,9 +312,6 @@ task generateGHCI() {
   appendToGHCI(paramFileContent, targetFilePath, otherTest, 120, integTestGradleArgs + "9999")
   jobs << otherTest
 
-  // define a job that depends others to manage the status check
-  appendToGHCI(paramFileContent, targetFilePath, "E2ETestsCompletion", 120, "clean", jobs)
-
```
nisargthakkar, line 317: Can we keep this, or change it somehow. It makes creating and maintaining branch protection rules much simpler

.github/rawWorkflows/gh-ci-parameterized-flow.txt
```diff
@@ -34,6 +34,19 @@
           find . -path "**/build/reports/*" -or -path "**/build/test-results/*" > artifacts.list
           rsync -R --files-from=artifacts.list . ${{ github.job }}-artifacts
           tar -zcvf ${{ github.job }}-jdk${{ matrix.jdk }}-logs.tar.gz ${{ github.job }}-artifacts
+      - name: Generate Test Reports
+        if: success() || failure()
```
FelixGV, line 38: What does this do? Is it the equivalent of `if (true || false)`? Or are there more states besides success and failure, for which we would want to avoid generating a report...?

internal/venice-test-common/build.gradle
majisourav99: Sure. I could create a separate generator file for completion.  Currently there is no point to upload artifacts for it anyways.

.github/rawWorkflows/gh-ci-parameterized-flow.txt
majisourav99: By default GitHub workflow will skip if the previous steps fail. By specifying both it means to run regardless of the outcome of the previous steps.

.github/rawWorkflows/gh-ci-completion-flow.txt
```diff
@@ -0,0 +1,15 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        jdk: [17]
+    runs-on: ubuntu-latest
+    needs: $Dependency
+    if: $Conditional
+    timeout-minutes: $TimeOut
+    steps:
+    - name: AllIsWell
```
nisargthakkar, line 11: Super nit: Can we call it something else? Like `NoOp`?

internal/venice-test-common/build.gradle
```diff
@@ -314,7 +317,7 @@ task generateGHCI() {
   jobs << otherTest
 
   // define a job that depends others to manage the status check
-  appendToGHCI(paramFileContent, targetFilePath, "E2ETestsCompletion", 120, "clean", jobs)
+  appendToGHCI(completionFileContent, targetFilePath, "E2ETestsCompletion", 20, "clean", jobs)
```
nisargthakkar, line 320: Also, can we replace `"clean"` with `null`?

.github/rawWorkflows/gh-ci-parameterized-flow.txt
```diff
@@ -4,6 +4,10 @@
       matrix:
         jdk: [17]
     runs-on: ubuntu-latest
+    permissions:
+     id-token: write
+     contents: read
+     checks: write
```
nisargthakkar, line 10: Do we still need this?

majisourav99: I believe I do, one run failed without these https://github.com/majisourav99/venice/actions/runs/11675167928

nisargthakkar (approved):

PR #1273, opened 2024-11-01
[controller] Enable configuring a customized health check URL for storage clusters

nisargthakkar:
## Enable configuring a customized health check URL for storage clusters

Helix has added a new functionality that allows configuring a customized health check endpoint that Helix will invoke to decide on the health of the cluster. This URL can be configured by setting the controller config `controller.helix.rest.customized.health.url`.

## How was this PR tested?

Added unit tests. GH CI for regression testing

### DIFF
--- a/internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
@@ -351,6 +351,12 @@ public class ConfigKeys {
   public static final String CONTROLLER_HELIX_CLOUD_INFO_PROCESSOR_NAME = "controller.helix.cloud.info.processor.name";
 
   /**
+   * Base URL for customized health checks triggered by Helix. Default is empty string.
+   */
+  public static final String CONTROLLER_HELIX_REST_CUSTOMIZED_HEALTH_URL =
+      "controller.helix.rest.customized.health.url";
+
+  /**
    * Whether to enable graveyard cleanup for batch-only store at cluster level. Default is false.
    */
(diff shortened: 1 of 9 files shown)


internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
```diff
@@ -350,6 +350,12 @@ private ConfigKeys() {
    */
   public static final String CONTROLLER_HELIX_CLOUD_INFO_PROCESSOR_NAME = "controller.helix.cloud.info.processor.name";
 
+  /**
+   * Base URL for customized health checks triggered by Helix. Default is empty string.
+   */
+  public static final String CONTROLLER_HELIX_REST_CUSTOMIZED_HEALTH_URL =
```
majisourav99, line 356: nit: can we add `check` after health ?

nisargthakkar: I followed the naming convention that Helix has specified: `CUSTOMIZED_HEALTH_URL`. This URL is an umbrella URL, and it doesn't get invoked. Helix will invoke `${CONTROLLER_HELIX_REST_CUSTOMIZED_HEALTH_URL}/aggregatedHealthStatus`. So, the healthcheck part is one API in this umbrella. I'm not sure if we should make this config specific to health check in that case.

What do you think?

majisourav99: I see, there could be more endpoints under this.

majisourav99 (approved):
LGTM

PR #1274, opened 2024-11-01
[controller] Make to_be_stopped_instances in aggregatedHealthStatus API optional

nisargthakkar:
## Make to_be_stopped_instances in aggregatedHealthStatus API optional

In Helix's aggregated stoppable check, the `to_be_stopped_instances` is optional and will be omitted in case there are no instances that have previously been approved. This commit makes our controller handle this.

## How was this PR tested?

Added unit test and integration tests. GHCI

### DIFF
--- a/internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/controller/server/TestAdminSparkServer.java
+++ b/internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/controller/server/TestAdminSparkServer.java
@@ -73,6 +73,7 @@ import org.apache.http.entity.ContentType;
 import org.apache.http.entity.StringEntity;
 import org.apache.http.impl.nio.client.CloseableHttpAsyncClient;
 import org.apache.http.message.BasicNameValuePair;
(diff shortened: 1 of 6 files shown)


internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/controller/server/TestAdminSparkServer.java
```diff
@@ -73,6 +73,7 @@
 import org.apache.http.entity.StringEntity;
 import org.apache.http.impl.nio.client.CloseableHttpAsyncClient;
 import org.apache.http.message.BasicNameValuePair;
+import org.apache.http.nio.client.HttpAsyncClient;
```
majisourav99, line 76: why this?

nisargthakkar: It's the base class for `CloseableHttpAsyncClient` and only used in a helper method. Let me know if you want me to remove it

nisargthakkar:
Thanks a lot, @majisourav99 !

PR #1275, opened 2024-11-02
[admin-tool][server] Add new admin-tool command to dump heartbeat status from server; Scan and log stale HB replica

sixpluszero:
## [admin-tool][server] Add new admin-tool command to dump heartbeat status from server; Scan and log stale HB replica
Previous PR #1260 is too complex, this PR only focus on HB related usability improvements

This PR adds two features related to heartbeat:
1. **Add a heartbeat scan thread to periodically run and log lagging resources** (every minute by default, this should be good enough not to spam logging). This can be further collected by other logging collecting system and we can easily detect on which host, what replica is lagging by how much.
2. **Add a command to dump heartbeat status from a host.** It has 2 optional filter: partition filter and lag filter. You can choose to see only specific topic / topic-partition or you can choose to only see resources that are lagging. This serves as the manual helper when (1) might be missing stuff.

## How was this PR tested?

Added new integration test


clients/venice-admin-tool/src/main/java/com/linkedin/venice/AdminTool.java
```diff
@@ -3074,6 +3077,43 @@ private static void dumpTopicPartitionIngestionContext(CommandLine cmd) throws E
     }
   }
 
+  static void dumpHostHeartbeatLag(
+      TransportClient transportClient,
+      String topicFilter,
+      String partitionFilter,
+      String lagFilter) throws Exception {
+    String topicName = topicFilter == null ? "" : topicFilter;
```
haoxu07, line 3085: How about we just use null to replace "" to indicate we did not explicitly provide string topic and partition int partition number. For example,` ./admin-tool --server-url xxxx  --kafka-topic topic_2 --lag-filter-enabled false` to show all partitions' lag on that host.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/storage/IngestionMetadataRetrieverDelegator.java
```diff
@@ -0,0 +1,68 @@
...
+  @Override
+  public TopicPartitionIngestionContextResponse getTopicPartitionIngestionContext(
+      String versionTopic,
+      String topicName,
+      int partitionNum) {
+    return kafkaStoreIngestionService.getTopicPartitionIngestionContext(versionTopic, topicName, partitionNum);
+  }
+
+  @Override
+  public TopicPartitionIngestionContextResponse getHeartbeatLag(
```
haoxu07, line 52: Maybe here we need to rename the Response class: `TopicPartitionIngestionContextResponse` a bit, it basically just wrap byte array for returning json string of server internal information.

haoxu07 (commented):
Overall LGTM! Thanks for preparing this! Just a few comments!

clients/venice-admin-tool/src/main/java/com/linkedin/venice/AdminTool.java
sixpluszero: But it needs to be passed into the request URI. Will null be converted into "null"? or actually null will throw exception in this case?

haoxu07 (approved):

PR #1278, opened 2024-11-04
[server] System store current version should also place latch when becoming from offline to standby

sixpluszero:
## [server] System store current version should also place latch when becoming from offline to standby
If not, system store can be come standby/leader prematurely for current version and serve stale data.

## How was this PR tested?

Add unit test


sushantmane (commented):
Changes look good. I was wondering if we should use smaller `bootstrapToOnlineTimeoutInHours` value for system stores

sixpluszero:
> Changes look good. I was wondering if we should use smaller `bootstrapToOnlineTimeoutInHours` value for system stores. The default is 24H which seems high for sys stores

It is 1h for system store though, as we expect it to have not so much data.

sushantmane (approved):
LGTM! Thanks, @sixpluszero!

PR #1279, opened 2024-11-04
[server] Fix stats handling handling for early exiting requests in servers

nisargthakkar:
## Fix stats handling handling for early exiting requests in servers

In servers, when handling stats, we refer to a cached value of status code. However, in case of early exiting responses like `FORBIDDEN`, `UNAUTHORIZED`, etc. we missed updating the cached value, and that triggered incorrect code execution leading to excessive logging.

## How was this PR tested?

Added unit tests

### DIFF
--- a/services/venice-server/src/main/java/com/linkedin/venice/listener/OutboundHttpWrapperHandler.java
+++ b/services/venice-server/src/main/java/com/linkedin/venice/listener/OutboundHttpWrapperHandler.java
@@ -52,6 +52,8 @@ public class OutboundHttpWrapperHandler extends ChannelOutboundHandlerAdapter {
     int responseRcu = 1;
     CompressionStrategy compressionStrategy = CompressionStrategy.NO_OP;
     boolean isStreamingResponse = false;
+
+    FullHttpResponse response = null;
     try {
       if (msg instanceof AbstractReadResponse) {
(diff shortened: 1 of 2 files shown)


services/venice-server/src/main/java/com/linkedin/venice/listener/OutboundHttpWrapperHandler.java
```diff
@@ -168,7 +171,12 @@ public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise)
       statsHandler.setResponseStatus(responseStatus);
     }
 
-    FullHttpResponse response = new DefaultFullHttpResponse(HTTP_1_1, responseStatus, body);
+    if (response != null) {
+      ctx.writeAndFlush(response);
```
sushantmane, line 175: Since just one branch returns early it might be good to have early return in that branch itself?

nisargthakkar: This change prevents the race condition too as it ensures that `statsHandler.setResponseStatus(responseStatus)` will be invoked before the handlers get invoked

sushantmane (approved):
LGTM. Thanks, @nisargthakkar!

PR #1281, opened 2024-11-04
[dvc][server][samza] Heap size estimation improvement

FelixGV:
[dvc][server][samza] Heap size estimation improvement

Introduced two new utilities to make our on-heap memory usage assessment more 
accurate, and easier to maintain as class hierarchies evolve:

- ClassSizeEstimator: Predicts (based on assumptions about how the JVM's
  memory layout works) the shallow size of a class. This includes the object
  header, all primitive fields, and all references to other objects (but it
  does not count these other objects, hence the shallowness). Reflection is
  used in this class.
- InstanceSizeEstimator: Predicts the size of instances of a limited number
  of classes. This is not a general-purpose utility, and it requires some
  manual effort to onboard a new class. Reflection is not used in this class.

The general design goals are the following:

- Reflection should only be used once per class per runtime, and the result
  of this logic should be stored in static constants.
- On the hot path, there should be no reflection, and we should leverage our
  knowledge of the Venice code base to determine which objects are meant to
  be counted or not. For example, singleton or otherwise shared instances
  should not be counted, since their amortized cost is negligible (besides
  the size of the pointer to refer to them).

The above utilities have been integrated in all classes that implement the
Measurable interface, and several new classes have been given this interface
as well. The Measurable::getSize function has been renamed getHeapSize, to
minimize the chance that it could clash with other function names, and to
make it extra clear what kind of size is meant.

Miscellaneous:

- Minor efficiency improvements to PubSubMessageHeaders and ApacheKafkaUtils
  so that empty headers (a common case) carry less overhead. Also made the
  PubSubMessageHeaders implement Iterable.
- Created a DefaultLeaderMetadata static class in VeniceWriter, so that a
  shared instance can be leveraged in cases where that object is always the
  same (e.g. when producing to the RT topic).
- BlobSnapshotManagerTest improvements:
  - Added timeouts to all tests.
  - Fixed a race condition in testMultipleThreads.

## How was this PR tested?
New unit tests.


gaojieliu:
Have you got a chance to take a look at `openjdk/jol` implementation?
https://github.com/openjdk/jol
It might carry some issues, but I guess we can analyze the strategy this lib is using to see whether we can borrow some or not.

internal/venice-common/src/main/java/com/linkedin/venice/memory/HeapSizeEstimator.java
```diff
@@ -0,0 +1,188 @@
...
+
+
+/**
+ * Utility class to help in implementing {@link Measurable#getSize()}. A couple of important points:
+ *
+ * 1. This utility class does not "measure" the heap size, but rather attempts to "predict" it, based on knowledge of
+ *    the internals of the JVM. If any of the assumptions are wrong, then of course the results will be inaccurate.
+ * 2. This utility class assumes we are using the HotSpot JVM.
+ */
+public class HeapSizeEstimator {
```
FelixGV, line 22: Starting a thread to respond to @gaojieliu's general comment:

>Have you got a chance to take a look at openjdk/jol implementation?
>https://github.com/openjdk/jol
>It might carry some issues, but I guess we can analyze the strategy this lib is using to see whether we can borrow some or not.

I posted this work on social media and received a lot of interesting examples from folks about other open source projects with the same or similar purpose. I'll maintain a list of these here:

1. https://github.com/openjdk/jol
2. https://github.com/ehcache/sizeof
3. https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/util/RamUsageEstimator.java

For now, this work looks pretty similar to (at least some parts of) the above, but I think there is still potential to have a good implementation as part of Venice. In particular, I want a solution in which there is no hot path reflection. For now, this tradeoff is not apparent because I have not yet included examples of how we can leverage this utility in the rest of our code, but I will add it soon when the basic functionality is stabilized...

internal/venice-common/src/main/java/com/linkedin/venice/memory/HeapSizeEstimator.java
```diff
@@ -0,0 +1,188 @@
...
+    }
+
+    int finalSize = roundUpToNearestAlignment(size);
+
+    return finalSize;
+  }
+
+  /** Deal with alignment by rounding up to the nearest alignment boundary. */
+  private static int roundUpToNearestAlignment(int size) {
+    int partialAlignmentWindowUsage = size % ALIGNMENT_SIZE;
```
alex-dubrouski, line 121: Alignment could be modified with command line.

internal/venice-common/src/main/java/com/linkedin/venice/memory/HeapSizeEstimator.java
```diff
@@ -0,0 +1,188 @@
...
+  }
+
+  private static int overheadOfFields(Class c) {
+    int size = 0;
+
+    /**
+     * {@link Class#getDeclaredFields()} returns the fields (of all visibility, from private to public and everything
+     * in between) of the class, but not of its parent class.
+     */
+    for (Field f: c.getDeclaredFields()) {
```
alex-dubrouski, line 134: Some of the fields are not visible. For example there are native references.

internal/venice-common/src/main/java/com/linkedin/venice/memory/HeapSizeEstimator.java
gaojieliu: The `reflection` mentioned here is mainly about performance, right?
Have you tried these alternatives and run it with JMH?
I tried `jol` and the initialization is slow, but after that, it is fairly fast.

internal/venice-common/src/main/java/com/linkedin/venice/memory/HeapSizeEstimator.java
FelixGV: I know that it can be configured at start time. But are you saying it can be modified at run time?

internal/venice-common/src/main/java/com/linkedin/venice/memory/HeapSizeEstimator.java
```diff
@@ -0,0 +1,188 @@
...
+    // TODO: Fix imprecision due to internal loss in parent class layouts
+    Class parentClass = c.getSuperclass();
+    while (parentClass != null) {
+      size += overheadOfFields(parentClass);
+      parentClass = parentClass.getSuperclass();
+    }
+
+    int finalSize = roundUpToNearestAlignment(size);
+
+    return finalSize;
```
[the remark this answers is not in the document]
FelixGV: Yes... for common classes like String it may be good to do so... although the efficiency of this utility is not a prime concern since it's intended to be called only once per class of interest per runtime.

internal/venice-common/src/main/java/com/linkedin/venice/memory/HeapSizeEstimator.java
FelixGV: Could you point me to a class (perhaps from the JDK) that has a native reference in it? I'd like to test that scenario and see what happens...

alex-dubrouski: Java.lang.Object :)

FelixGV: Then I'm definitely not taking it into account 😅 ... where can I learn more about this?

internal/venice-common/src/main/java/com/linkedin/venice/memory/HeapSizeEstimator.java
gaojieliu: I did a small test against `jol` and it seems it can work properly with Oracle OpenJDK8 release, but not MSFT OpenJDK11 release.

internal/venice-common/src/main/java/com/linkedin/venice/memory/HeapSizeEstimator.java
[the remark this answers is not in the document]
FelixGV: I added memoization in 9f1ef31.

internal/venice-common/src/main/java/com/linkedin/venice/memory/HeapSizeEstimator.java
```diff
@@ -0,0 +1,231 @@
...
+   * 1. For any null pointer, we can subtract from the base a value equal to the output of this function for the type of
+   *    the field which is null.
+   * 2. For any variable-sized object, then the actual size has to be taken into account.
+   * 3. If polymorphism comes into play, then the specific sizes of actual instances need to take into account the
+   *    concrete classes they are made of.
+   *
+   * @throws {@link StackOverflowError} It should be noted that this function is recursive in nature and so any class
+   *         which contains a loop in its class graph will cause a stack overflow.
+   */
+  public static <T> int getClassOverhead(Class<T> c) {
```
gaojieliu: So this function is mainly calculating the class overhead, not the deep size, right?
If the object contains a nested hashmap, this class will only measure the overhead of the class, and the user needs to manually add the class overhead on top of the `hashmap` cost, which needs to be calculated manually?

I think it will be useful to have a function to measure both class overhead and the actual data.

gaojieliu: The above requirement is mainly about measuring the deep size of the object.

If we do want to implement such feature, it will be great if we can have some filtering logic to ignore certain types of fields, which can be referring to some common references.

internal/venice-common/src/main/java/com/linkedin/venice/memory/HeapSizeEstimator.java
gaojieliu: @alex-dubrouski 
Do you know why MSFT-jdk11 won't work well with attached agent to measure the object size?
I tried to add this jvm arg, but it still doesn't work:
```
-Djdk.attach.allowAttachSelf=true
```

And both `jol` and `sizeof` are suffering from the same issue with JDK11 as they are using the similar technologies..

internal/venice-common/src/main/java/com/linkedin/venice/memory/HeapSizeEstimator.java
FelixGV: Yes, this function does not measure the deep size of an object, and that is also the reason why it takes a `Class` param and not an `Object`. Essentially, my idea is to hand-code the "variable" part of the measurement of the classes of interest, and the function in this utility is just to make it easier to have the "base cost" (the non-variable part). This utility is _not_ intended to be called in the hot path.

I will integrate the usage of this utility into the `MemoryBoundBlockingQueue` measurement code. I have not done so yet because I want to be extra sure the utility is as accurate as possible (which it may or may not be, as I am not sure if I have chased all of the bugs properly yet, but at least it should be pretty close...).

internal/venice-common/src/main/java/com/linkedin/venice/memory/HeapSizeEstimator.java
alex-dubrouski: It works, just shows warning that dynamic attach failed:
```
 [ ~/tools/jol ]$ . ~/.j17
 [ ~/tools/jol ]$ java -cp .:jol-cli.jar org.openjdk.jol.Main internals java.lang.String
# WARNING: Unable to get Instrumentation. Dynamic Attach failed. You may add this JAR as -javaagent manually, or supply -Djdk.attach.allowAttachSelf
# WARNING: Unable to attach Serviceability Agent. sun.jvm.hotspot.memory.Universe.getNarrowOopBase()
# Running 64-bit HotSpot VM.
# Using compressed oop with 3-bit shift.
# Using compressed klass with 3-bit shift.
# WARNING | Compressed references base/shifts are guessed by the experiment!
# WARNING | Therefore, computed addresses are just guesses, and ARE NOT RELIABLE.
# WARNING | Make sure to attach Serviceability Agent to get the reliable addresses.
# Objects are 8 bytes aligned.
# Field sizes by type: 4, 1, 1, 2, 2, 4, 4, 8, 8 [bytes]
# Array element sizes: 4, 1, 1, 2, 2, 4, 4, 8, 8 [bytes]

Instantiated the sample instance via default constructor.

java.lang.String object internals:
OFF  SZ      TYPE DESCRIPTION               VALUE
  0   8           (object header: mark)     0x0000000000000001 (non-biasable; age: 0)
  8   4           (object header: class)    0x0000eba0
 12   4       int String.hash               0
 16   1      byte String.coder              0
 17   1   boolean String.hashIsZero         false
 18   2           (alignment/padding gap)   
 20   4    byte[] String.value              []
Instance size: 24 bytes
Space losses: 2 bytes internal + 0 bytes external = 2 bytes total

```

internal/venice-common/src/main/java/com/linkedin/venice/memory/HeapSizeEstimator.java
alex-dubrouski: No, it can't be changed at runtime, but you defined static alignment which might not be accurate. That's an edge case of course, I was just saying.

internal/venice-common/src/main/java/com/linkedin/venice/memory/HeapSizeEstimator.java
alex-dubrouski: Alex Shipilev explained that in one of the videos (video is not in English...)

internal/venice-common/src/main/java/com/linkedin/venice/memory/HeapSizeEstimator.java
FelixGV: Oh I see... you mean that alignment can be overridden by a JVM option, in which case it wouldn't decide just based on whether it's a 32 bits or 64 bits VM?

Is it via `-XX:ObjectAlignmentInBytes`?

alex-dubrouski: Yes, it is set statically at bootstrap, but code does not take this into account. Feel free to ignore that, since this is very rare edge case.

internal/venice-common/src/main/java/com/linkedin/venice/memory/HeapSizeEstimator.java
FelixGV: What is this `String.hashIsZero` field? I don't see that when looking at the source code 🤔

gaojieliu: It turned out that msft-jdk11.0.8/11.0.13 has a bug, which triggered the crash while using 'jol' or 'sizeof' lib and JDK17 apparently fixed this issue.

alex-dubrouski: Well, Felix, that's why we should probably use JOL :D ;)

FelixGV: But where is the learning value in that, @alex-dubrouski :D ?

Besides, this memory estimation code needs to run in DVC, which is one of our client libs, meaning we cannot tightly control which JVM it's going to run in. Ideally, it should not depend on agents or finicky `Unsafe` code...

That being said, using JOL in unit test to validate the accuracy of this utility might be a good idea 🤔

FelixGV: BTW... you did not answer my question @alex-dubrouski 😅 but I found out that the `hashIsZero` was added 5 years ago: https://github.com/openjdk/jdk/commit/89a267ca46d6252e745fe8f2f1a9cb618af5f83f

According to [JDK-8221836](https://bugs.openjdk.org/browse/JDK-8221836), this change got released in Java 13, which is weird because I thought I was looking at the Java 17 source code in my IDE, but apparently not; it was probably hanging on to the source of an older Java version.

FelixGV: And indeed, I can get it to show me the proper source and see that it's there now.

alex-dubrouski: I tested on JDK17 so yeah, it is there. Plus it is a field not method.

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/consumer/AdminConsumptionTask.java
```diff
@@ -575,16 +579,22 @@ private void executeMessagesAndCollectResults() throws InterruptedException {
...
+                  problematicStores.put(storeName, errorInfo);
+                  LOGGER.warn(errorInfo.exception.getMessage());
+                }
               }
             } else {
-              errorInfo.exception = e;
-              errorInfo.offset = storeAdminOperationsMapWithOffset.get(storeName).peek().getOffset();
-              problematicStores.put(storeName, errorInfo);
+              AdminOperationWrapper nextOperation = storeAdminOperationsMapWithOffset.get(storeName).peek();
+              if (nextOperation != null) {
```
xunyin8, line 593: Since an exception was thrown, we probably still want to set the exception and propagate it for logging despite the admin operation is null and we cannot retrieve the offset.

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/consumer/AdminConsumptionTask.java
```diff
@@ -575,16 +579,22 @@ private void executeMessagesAndCollectResults() throws InterruptedException {
 
               if (lastSucceededId == newLastSucceededId && perStorePendingMessagesCount > 0) {
                 // only mark the store problematic if no progress is made and there are still message(s) in the queue.
-                errorInfo.exception = new VeniceException(
-                    "Could not finish processing admin message for store " + storeName + " in time");
-                errorInfo.offset = storeAdminOperationsMapWithOffset.get(storeName).peek().getOffset();
-                problematicStores.put(storeName, errorInfo);
-                LOGGER.warn(errorInfo.exception.getMessage());
+                AdminOperationWrapper nextOperation = storeAdminOperationsMapWithOffset.get(storeName).peek();
+                if (nextOperation != null) {
```
xunyin8, line 583: Similarly here, might be worth still reporting the issue just without the offset if we caught a `CancellationException` but the admin operation is somehow null.

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/consumer/AdminConsumptionTask.java
FelixGV: In https://github.com/linkedin/venice/pull/1281/commits/2cc911156125577d647b6a1921095505b300f473 I have made it so that the logging always happens.

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/consumer/AdminConsumptionTask.java
FelixGV: In https://github.com/linkedin/venice/pull/1281/commits/2cc911156125577d647b6a1921095505b300f473 I have made it so that the logging always happens.

internal/venice-common/src/main/java/com/linkedin/venice/memory/ClassSizeEstimator.java
```diff
@@ -0,0 +1,245 @@
...
+    classHierarchyFromSubclassToParent.add(c);
+    Class parentClass = c.getSuperclass();
+    while (parentClass != null) {
+      classHierarchyFromSubclassToParent.add(parentClass);
+      parentClass = parentClass.getSuperclass();
+    }
+
+    // Iterate from the end to the beginning, so we go from parent to sub
+    for (int i = classHierarchyFromSubclassToParent.size() - 1; i >= 0; i--) {
+      int classFieldsOverhead = overheadOfFields(classHierarchyFromSubclassToParent.get(i));
```
gaojieliu, line 126: Do we want to use the cache here? Some parent class might be calculated before.

internal/venice-common/src/main/java/com/linkedin/venice/memory/ClassSizeEstimator.java
```diff
@@ -0,0 +1,245 @@
...
+      }
+      Class fieldClass = f.getType();
+      if (fieldClass.isPrimitive()) {
+        size += KNOWN_SHALLOW_SIZES.get(fieldClass);
+      } else {
+        /**
+         * Only primitives are stored in-line within the object, while all non-primitives are stored elsewhere on the
+         * heap, with a pointer within the object to reference them.
+         */
+        size += POINTER_SIZE;
```
gaojieliu, line 217: What is the main reason that we don't want to calculate the class overhead recursively? Stack overflow? or the referenced objects can be shared across multiple instances?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreBufferService.java
```diff
@@ -491,9 +506,16 @@ public int hashCode() {
...
     }
+
+    @Override
+    protected int getBaseClassOverhead() {
+      return PARTIAL_CLASS_OVERHEAD;
+    }
   }
 
   private static class LeaderQueueNode extends QueueNode {
+    private static final int SHALLOW_CLASS_OVERHEAD = ClassSizeEstimator.getClassOverhead(QueueNode.class);
```
gaojieliu, line 517: Why not `LeaderQueueNode`?

gaojieliu (commented):

internal/venice-common/src/main/java/com/linkedin/venice/memory/InstanceSizeEstimator.java
```diff
@@ -0,0 +1,180 @@
...
+        return value -> getSize((Delete) value);
+      } else if (Update.class.isAssignableFrom(type)) {
+        return value -> getSize((Update) value);
+      } else if (ControlMessage.class.isAssignableFrom(type)) {
+        return value -> getSize((ControlMessage) value);
+      } else if (KafkaMessageEnvelope.class.isAssignableFrom(type)) {
+        return value -> getSize((KafkaMessageEnvelope) value);
+      } else if (ProducerMetadata.class.isAssignableFrom(type)) {
+        return value -> getSize((ProducerMetadata) value);
+      } else if (ByteBuffer.class.isAssignableFrom(type)) {
```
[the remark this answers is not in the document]
FelixGV: That's right. For now, that function is only meant to be used with `HeapByteBuffer` instances and will throw `IllegalArgumentException` if a `DirectBB` is passed. If a subclass of `HeapBB` is passed, there could be imprecision. I'm not aware of such case though. Are you?

internal/venice-common/src/main/java/com/linkedin/venice/memory/ClassSizeEstimator.java
FelixGV: I think that would be incorrect, because the cache contains the final size including object header and alignment. Here, we are looking for the "intermediate result" (the fields only), so we can add all these intermediate results together and account for alignment afterwards. We could have a separate cache for these intermediate results but I don't think it's worth it given that the intent is to only call this once per class per runtime, never on the hot path...

internal/venice-common/src/main/java/com/linkedin/venice/memory/ClassSizeEstimator.java
FelixGV: Yeah so... I changed that. I removed the recursiveness because when I made use of these utilities in all of the code paths where we do heap measurement, it turns out that I was always invoking the function with `shallow == true`. The only code calling `shallow == false` was the unit test code, which indicated to me that we didn't really need that.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreBufferService.java
FelixGV: Good catch! And that is one of the cases I didn't write a unit test for... I think I need to write one now to have a clean conscience 😅

internal/venice-common/src/main/java/com/linkedin/venice/memory/ClassSizeEstimator.java
```diff
@@ -0,0 +1,245 @@
...
+  static int roundUpToNearestAlignment(int size) {
+    return roundUpToNearest(size, ALIGNMENT_SIZE);
+  }
+
+  /** Deal with alignment by rounding up to the nearest boundary. */
+  private static int roundUpToNearest(int size, int intervalSize) {
+    int partialAlignmentWindowUsage = size % intervalSize;
+    int waste = partialAlignmentWindowUsage == 0 ? 0 : intervalSize - partialAlignmentWindowUsage;
+    int finalSize = size + waste;
+    return finalSize;
```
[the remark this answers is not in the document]
FelixGV: That's right, and that is the logic I'm using in `HeapSizeEstimatorTest::roundUpToNearestAlignment`, so they are fully equivalent AFAICT. We could pick either one for the main code. I thought this one was slightly more clear, so I picked it for the main code, but we could do the reverse. LMK if you have a preference.

internal/venice-common/src/main/java/com/linkedin/venice/memory/InstanceSizeEstimator.java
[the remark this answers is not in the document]
gaojieliu: So can we just use `HeapByteBuffer` here?

internal/venice-common/src/main/java/com/linkedin/venice/memory/ClassSizeEstimator.java
gaojieliu: Got it, makes sense.
Can you also convert these explanation into Javadoc?

internal/venice-common/src/main/java/com/linkedin/venice/memory/ClassSizeEstimator.java
gaojieliu: Got it.

internal/venice-common/src/main/java/com/linkedin/venice/memory/InstanceSizeEstimator.java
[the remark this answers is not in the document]
FelixGV: We cannot, actually, because that class is package-private! The JDK designers intended to present only `ByteBuffer` as the API.

internal/venice-common/src/main/java/com/linkedin/venice/memory/ClassSizeEstimator.java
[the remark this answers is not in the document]
FelixGV: I hesitated on this... I already need to invoke or copy a fair amount of the main code into the test code, which leaves me with a feeling that it is somewhat inadequate. It is very likely that `actual == expected` if both of these have been computed with the same code 😅 ... and I especially want to avoid the case where the main code gets a regression introduced, but it is undetected because the test code just calls the main code for both actual and expected paths...

And so, in this particular case, since as part of experimenting I ended up coding two valid and equivalent versions of the algorithm (at least AFAICT) I thought it may be good to let the test code have its own implementation, different from the main code.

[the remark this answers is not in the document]
gaojieliu: Sounds good.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreBufferService.java
FelixGV: I added a unit test for this and it does catch the issue, then I fixed it. Thanks again for catching this!

gaojieliu (approved):
Nice job!

FelixGV:
> Nice job!

Thanks a lot for the reviews!

PR #1282, opened 2024-11-05
[controller] log compaction

WhitneyDeng:
## Summary
- purpose: to get stores ready for compaction based on programmatically defined criteria
- added API endpoints to get compaction-ready stores

## How was this PR tested?
unit test: filter all stores on a cluster for compaction-ready stores


internal/venice-common/src/main/java/com/linkedin/venice/controllerapi/ControllerClient.java
```diff
@@ -1156,6 +1156,16 @@ public MultiStoreInfoResponse getClusterStores(String clusterName) {
     return request(ControllerRoute.GET_STORES_IN_CLUSTER, params, MultiStoreInfoResponse.class);
   }
 
+  /**
+   * This method gets a list of store names that are ready for compaction.
+   * @param clusterName, the name of the cluster to query for compaction eligible stores
+   * @return The list of store names that are ready for compaction.
+   */
+  public MultiStoreInfoResponse getStoresForCompaction(String clusterName) {
+    QueryParams params = newParams().add(CLUSTER, clusterName);
```
mynameborat, line 1165: `ControllerClient` is currently scoped to a `cluster` as part of creation and thus by default if you were to create a request with `newParams()` it already includes `CLUSTER` parameter that the controller was instantiated with.

However, if the intent of this API is to explicitly take a cluster, you will need to update this to just instantiate an empty `QueryParams` and set the `CLUSTER` parameter.

You can check out how `newParams()` is implemented below to realize it adds the `CLUSTER` parameters through common parameters method.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/controller/VeniceParentHelixAdminTest.java
```diff
@@ -14,6 +14,7 @@
 import static com.linkedin.venice.utils.TestUtils.assertCommand;
 import static com.linkedin.venice.utils.TestUtils.waitForNonDeterministicAssertion;
 import static com.linkedin.venice.utils.TestUtils.waitForNonDeterministicPushCompletion;
+import static org.mockito.Mockito.*;
```
mynameborat, line 17: nit: revert changes since your PR isn't adding anything new to this class

services/venice-controller/src/main/java/com/linkedin/venice/controller/Admin.java
```diff
@@ -935,6 +936,12 @@ default boolean isAdminTopicConsumptionEnabled(String clusterName) {
 
   Map<String, StoreDataAudit> getClusterStaleStores(String clusterName);
 
+  /**
+   * @param clusterName
+   * @return the list of stores ready for compaction
+   */
+  ArrayList<StoreInfo> getStoresForCompaction(String clusterName);
```
mynameborat, line 943: Replace `ArrayList<Store>` with `List<Store>` as this forces the API to return a concrete implementation of a `List` as opposed to any implementations of `List`; e.g., `LinkedList`, `Vector` etc. 

Generally, using interfaces as return parameters is favorable as it removes the dependencies from callers depending on concrete implementation methods as opposed to well defined publicly exposed methods.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
```diff
@@ -4801,6 +4801,81 @@ public Map<String, StoreDataAudit> getClusterStaleStores(String clusterName) {
...
+      // filter out
+      filterStoresForCompaction(storeInfoList, compactionReadyStores);
+    } catch (Exception e) {
+      throw new VeniceException("Something went wrong trying to fetch stores for compaction.", e);
+    }
+    return compactionReadyStores;
+  }
+
+  // package exclusive for testing
+  void filterStoresForCompaction(ArrayList<StoreInfo> storeInfoList, ArrayList<StoreInfo> compactionReadyStores) {
```
mynameborat, line 4835: Suggest that this returns the `compactionReadyStores` instead of mutating the input parameters.

It is a good practice to not mutate input parameters inside a function as it helps to reason about the changes to the variable. Keeping things immutable when possible makes it easy to reason about changes even when the code evolves.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
```diff
@@ -4801,6 +4801,81 @@ public Map<String, StoreDataAudit> getClusterStaleStores(String clusterName) {
...
+  // This function abstracts the criteria for a store to be ready for compaction
+  private boolean isCompactionReady(StoreInfo storeInfo) {
+    return isHybridStore(storeInfo) && isLastCompactionTimeOlderThanThresholdHours(24, storeInfo);
+    // TODO: code style: store compactionThresholdHours as final var/enum
+  }
+
+  // START isCompactionReady() helper methods: each method below encapsulates a log compaction readiness criterion
+  private boolean isHybridStore(StoreInfo storeInfo) {
+    return storeInfo.getHybridStoreConfig() != null;
+  }
```
mynameborat, line 4852: While I was the one to suggest modularization. Feel free to make the call on breaking methods; E.g., `storeInfo.getHybridStoreConfig() != null` is probably simple enough to wrap inside the `isCompactionReady` method. We can always break the methods if the logic identifying hybrid stores evolves.

It is more of a subjective comment. Feel free to leave this as is or merge it together.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
```diff
@@ -4801,6 +4801,81 @@ public Map<String, StoreDataAudit> getClusterStaleStores(String clusterName) {
...
+  void filterStoresForCompaction(ArrayList<StoreInfo> storeInfoList, ArrayList<StoreInfo> compactionReadyStores) {
+    for (StoreInfo storeInfo: storeInfoList) {
+      if (isCompactionReady(storeInfo)) {
+        compactionReadyStores.add(storeInfo);
+      }
+    }
+  }
+
+  // This function abstracts the criteria for a store to be ready for compaction
+  private boolean isCompactionReady(StoreInfo storeInfo) {
```
mynameborat, line 4844: Consider making this package private so that we can add tests to specifically check the core logic of what constitutes the readiness for a store to be compacted.

I see you have tested this transitively through the higher order methods. It might simpler to author and perhaps also add various scenarios this way.

services/venice-controller/src/main/java/com/linkedin/venice/controller/server/AdminSparkServer.java
```diff
@@ -1,108 +1,7 @@
...
-import static com.linkedin.venice.controllerapi.ControllerRoute.UPDATE_ADMIN_TOPIC_METADATA;
-import static com.linkedin.venice.controllerapi.ControllerRoute.UPDATE_CLUSTER_CONFIG;
-import static com.linkedin.venice.controllerapi.ControllerRoute.UPDATE_KAFKA_TOPIC_LOG_COMPACTION;
-import static com.linkedin.venice.controllerapi.ControllerRoute.UPDATE_KAFKA_TOPIC_MIN_IN_SYNC_REPLICA;
-import static com.linkedin.venice.controllerapi.ControllerRoute.UPDATE_KAFKA_TOPIC_RETENTION;
-import static com.linkedin.venice.controllerapi.ControllerRoute.UPDATE_STORAGE_PERSONA;
-import static com.linkedin.venice.controllerapi.ControllerRoute.UPDATE_STORE;
-import static com.linkedin.venice.controllerapi.ControllerRoute.UPLOAD_PUSH_JOB_STATUS;
-import static com.linkedin.venice.controllerapi.ControllerRoute.WIPE_CLUSTER;
+import static com.linkedin.venice.controllerapi.ControllerRoute.*;
```
mynameborat, line 4: Can you revert this change to keep the codestyle intact? Agree it is pain to handle the disparity in codestyles between the IDE and the codebase.

However, you should be able to revert this through intellij with a click.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
```diff
@@ -4801,6 +4801,81 @@ public Map<String, StoreDataAudit> getClusterStaleStores(String clusterName) {
...
+    return isHybridStore(storeInfo) && isLastCompactionTimeOlderThanThresholdHours(24, storeInfo);
```
mynameborat, line 4845: minor: extract 24 into a top level class constant

services/venice-controller/src/test/java/com/linkedin/venice/controller/TestVeniceParentHelixAdmin.java
```diff
@@ -2835,4 +2836,65 @@ public void testGetFinalReturnStatus() {
...
+    when(store1.getVersion(anyInt())).thenReturn(Optional.ofNullable(version1));
+    when(store2.getVersion(anyInt())).thenReturn(Optional.ofNullable(version2));
+    when(store3.getVersion(anyInt())).thenReturn(Optional.ofNullable(version3));
+
+    // Set createTime for Version mocks
+    long currentTime = System.currentTimeMillis();
+    long millisecondsPerHour = 60 * 60 * 1000;
+    when(version1.getCreatedTime()).thenReturn(currentTime - (25 * millisecondsPerHour)); // 25 hours ago
+    when(version2.getCreatedTime()).thenReturn(currentTime - (50 * millisecondsPerHour)); // 50 hours ago
+    when(version3.getCreatedTime()).thenReturn(currentTime - (23 * millisecondsPerHour)); // 23 hours ago
```
mynameborat, line 2881: This store is short-circuiting the check as it is a non-hybrid store.
Can you please add one more scenario to include a hybrid store that has had a version within the last 24 hours.

mynameborat (commented):
Good job on putting up the first PR quickly. Looks very good. I have put in some comments.

Let me know if you have questions.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/controller/VeniceParentHelixAdminTest.java
WhitneyDeng: will do.

what is nit?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
WhitneyDeng: Thank you for your suggestion. I had the same thought process and decided to modularise for readability.

services/venice-controller/src/main/java/com/linkedin/venice/controller/Admin.java
WhitneyDeng: done.

I'd like to further my understanding. We want the interface return type to be more generic to reduce dependency on  the return type of the concrete implementation, correct?

What does this mean for the specificity of the return type of the concrete implementation? i.e. should VeniceParentHelixAdmin::getStoresForCompaction() implementation return List or ArrayList?

services/venice-controller/src/test/java/com/linkedin/venice/controller/TestVeniceParentHelixAdmin.java
WhitneyDeng: version3 has a version 23 hours ago, which is <24 hours.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
WhitneyDeng: Yes I also agree and want to implement with this pattern. However, I need you help with some design decisions that I went through.

I made this choice because `getStoresForCompaction()` returns `compactionReadyStores` outside the `try` clause where `filterStoresForCompaction()` is invoked. Therefore, `compactionReadyStores` currently needs to be created outside the `try` clause.

What can we evolve to implement your suggested change?

internal/venice-common/src/main/java/com/linkedin/venice/controllerapi/ControllerClient.java
WhitneyDeng: Aha! I see that `newParams()` populate `cluster` downstream.

I wrote this code by referring to `getClusterStores()` and `getStoreLargestUsedVersion()` where the API takes `clusterName` but uses `newParams()`. Can you help me understand why/if `getClusterStores()` and getStoreLargestUsedVersion()` and `getStoresForCompaction()` should instantiate `QueryParams` differently? It seems to me they use `clusterName` similarly.

services/venice-controller/src/main/java/com/linkedin/venice/controller/Admin.java
mynameborat: >I'd like to further my understanding. We want the interface return type to be more generic to reduce dependency on the return type of the concrete implementation, correct?

Right. Ideally, you want the methods to return generic interfaces where possible unless you don't have one.

services/venice-controller/src/test/java/com/linkedin/venice/controller/TestVeniceParentHelixAdmin.java
mynameborat: True. However, version3 store is non-hybrid and hence the `shouldCompact` will already return false since its non-hybrid and in essence you are not validating the < 24 hour constraint explicitly.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
```diff
@@ -4801,6 +4802,78 @@ public Map<String, StoreDataAudit> getClusterStaleStores(String clusterName) {
...
+    // get the last compaction time
+    int currentVersionNumber = storeInfo.getCurrentVersion();
+    Optional<Version> currentVersion = storeInfo.getVersion(currentVersionNumber);
+    if (!currentVersion.isPresent()) {
+      LOGGER.warn("Couldn't find current version: {} from store: {}", currentVersionNumber, storeInfo.getName());
+      return false; // invalid store because no current version, this store is not eligible for compaction
+    }
+
+    // calculate hours since last compaction
+    long lastCompactionTime = currentVersion.get().getCreatedTime();
```
majisourav99, line 4868: Trying to understand the logic behind this, so we allow compaction where the current version is older than certain threshold ? or should we look at this field `getLatestVersionPromoteToCurrentTimestamp` which tell when the current version was promoted.

WhitneyDeng: Good question, Sourav. Bharath and I went through the same consideration. Here's why:

Since the purpose of repush is to compact the store, the time elapsed since version creation is the target metric. Hence, we look at creation rather than promotion timestamp. Does this make more sense?

services/venice-controller/src/main/java/com/linkedin/venice/controller/Admin.java
WhitneyDeng: I'm starting to understand. Could we walk through examples when we next sync to lock in my understanding?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
```diff
@@ -4801,6 +4801,23 @@ public Map<String, StoreDataAudit> getClusterStaleStores(String clusterName) {
...
 
+  /**
+   * This function iterates over all of Helix Parent Admin's child controllers,
+   * in order to obtain the stores and respective store information of each child controller,
+   * and then filter out the stores that are ready for compaction.
+   * @param clusterName, the name of the cluster to search for stores that are ready for compaction
+   * @return a list of StoreInfo objects of stores that are ready for compaction
+   */
+  @Override
+  public ArrayList<StoreInfo> getStoresForCompaction(String clusterName) {
```
mynameborat, line 4812: Same as above

mynameborat (commented):
I took another pass and put in comments. Looks like some parts are still work in progress. Can you ping me when you have them ready for review?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
WhitneyDeng: done

services/venice-controller/src/main/java/com/linkedin/venice/controller/logcompaction/CompactionManager.java
```diff
@@ -0,0 +1,102 @@
...
+      storeInfoList.addAll(response.getStoreInfoList());
+    }
+
+    // filter out
+    return filterStoresForCompaction(storeInfoList);
+  }
+
+  // public for testing
+  // TODO: make private or package protected
+  public ArrayList<StoreInfo> filterStoresForCompaction(ArrayList<StoreInfo> storeInfoList) {
```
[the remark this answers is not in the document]
WhitneyDeng: re access modifier strictness: done

re test: ongoing discussion

services/venice-controller/src/main/java/com/linkedin/venice/controller/logcompaction/CompactionManager.java
```diff
@@ -0,0 +1,102 @@
...
+      LOGGER.info(
+          "Repush job triggered for store: {} with job name: {} and job exec id: {}",
+          response.getStoreName(),
+          response.getJobName(),
+          response.getJobExecId());
+    } catch (Exception e) {
+      LOGGER.error("Failed to compact store: {}", storeName, e);
+    }
+  }
+  // END isCompactionReady() helper methods
```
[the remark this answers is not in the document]
WhitneyDeng: done

services/venice-controller/src/main/java/com/linkedin/venice/controller/logcompaction/CompactionManager.java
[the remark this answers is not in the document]
WhitneyDeng: done

services/venice-controller/src/main/java/com/linkedin/venice/controller/logcompaction/CompactionManager.java
```diff
@@ -0,0 +1,102 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+
+public class CompactionManager {
+  public static final int COMPACTION_THRESHOLD_HOURS = 24;
+  private static final Logger LOGGER = LogManager.getLogger(CompactionManager.class);
+
+  private static RepushOrchestrator repushOrchestrator;
+  // TODO: where to configure implementation of RepushOrchestrator to use?
```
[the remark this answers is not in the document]
WhitneyDeng: no longer static!

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceController.java
```diff
@@ -248,6 +252,7 @@ public void start() {
       secureAdminServer.start();
     }
     topicCleanupService.start();
+    logCompactionService.start();
```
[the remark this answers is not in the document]
WhitneyDeng: Can you help me understand what you mean by a feature flag? Any examples you can point to?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
WhitneyDeng: implemented

internal/venice-common/src/main/java/com/linkedin/venice/controllerapi/ControllerClient.java
WhitneyDeng: meaning we should remove `clusterName` param since VeniceController is already responsible for particular clusters?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
```diff
@@ -4801,6 +4801,23 @@ public Map<String, StoreDataAudit> getClusterStaleStores(String clusterName) {
...
+    throw new UnsupportedOperationException("This function is implemented in VeniceHelixAdmin.");
+  }
+
+  @Override
+  public void compactStore(String storeName) {
+    // TODO
```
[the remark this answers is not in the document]
WhitneyDeng: Can I confirm that `VPHA::compactStore()` does nothing since the compaction is done by child controllers?

Can I also confirm that `VHA::compactStore()` only calls `CompactionManager::compactStore()` and let `CompactionManager` handle the trigger response & logging?

WhitneyDeng:
closed because branch renamed to `whdeng/log-compaction-repush` to follow branch naming convention

WhitneyDeng:
reverted branch name to `log-compaction` for PR continuity.

internal/venice-common/src/main/java/com/linkedin/venice/controllerapi/ControllerClient.java
```diff
@@ -1156,6 +1156,21 @@ public MultiStoreInfoResponse getClusterStores(String clusterName) {
...
+    return request(ControllerRoute.GET_STORES_FOR_COMPACTION, params, MultiStoreInfoResponse.class);
+  }
+
+  public ControllerResponse triggerRepush(String storeName) {
+    QueryParams params = newParams().add(NAME, storeName);
+    return request(ControllerRoute.COMPACT_STORE, params, ControllerResponse.class);
+  }
```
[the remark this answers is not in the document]
WhitneyDeng: I think I follow but I'd like to make sure. Let's sync on this to get clarity.

internal/venice-common/src/main/java/com/linkedin/venice/controllerapi/ControllerClient.java
mynameborat: Nope. Just instantiate the `QueryParams` differently; i.e., 
```
new QueryParams()
   .setCluster(cluster)
   ....
```

instead of using `newParams()`

internal/venice-common/src/main/java/com/linkedin/venice/controllerapi/ControllerClient.java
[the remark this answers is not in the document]
mynameborat: Yes. I was suggesting the same and add a comment like
`TODO: Use byte[] to pass parameters instead of QueryParams as it is a post method so that you don't forget when you work later on this in the next PR`

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceController.java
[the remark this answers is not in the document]
mynameborat: Yes. that is right; just being able to turn on/turn off the service without code change and CI hurdles.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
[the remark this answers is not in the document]
mynameborat: Depends on what sort of thing happens within `CompactionManager::compactStore`. E.g., if you are invoking other APIs to get some other store metadata. My guess is it should happen on the child controller and also the leader child controller (potentially) to make sure it sees the most up-to-date store metadata e.g., version information (if at all it needs to invalidate some of the request parameters).

internal/venice-common/src/main/java/com/linkedin/venice/controllerapi/ControllerClient.java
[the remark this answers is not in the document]
WhitneyDeng: - comment added
- parked `ControllerClient::triggerRepush` for later PR

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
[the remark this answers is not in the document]
WhitneyDeng: Let's have `CompactionManager::compactStore` log the response status but have `VHA::compactStore` receive and handle errors. 

We can think of `CompactionManager` as an "irresponsible" class with low visibility, and all it does is execute and hands off any results back upstream.

services/venice-controller/src/main/java/com/linkedin/venice/controller/logcompaction/CompactionManager.java
```diff
@@ -0,0 +1,123 @@
...
+      return false; // invalid store because no current version, this store is not eligible for compaction
+    }
+
+    // calculate hours since last compaction
+    long lastCompactionTime = currentVersion.get().getCreatedTime();
+    long currentTime = System.currentTimeMillis();
+    long millisecondsSinceLastCompaction = currentTime - lastCompactionTime;
+    long hoursSinceLastCompaction = TimeUnit.MILLISECONDS.toHours(millisecondsSinceLastCompaction);
+
+    return hoursSinceLastCompaction > compactionThresholdHours;
```
lusong64, line 97: if you are a big fan of lambda or Optional, this can be re-written in 
```java
return storeInfo.getVersion(currentVersionNumber)
            .map(v -> {
                 // calculate hours since last compaction
                 long lastCompactionTime = v.get().getCreatedTime();
                 long currentTime = System.currentTimeMillis();
                 long millisecondsSinceLastCompaction = currentTime - lastCompactionTime;
                 long hoursSinceLastCompaction = 
                 TimeUnit.MILLISECONDS.toHours(millisecondsSinceLastCompaction);
  
                  return hoursSinceLastCompaction > compactionThresholdHours;
             ).orElse (false);
```

WhitneyDeng: elegant! implemented.

internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
```diff
@@ -235,6 +235,36 @@ private ConfigKeys() {
...
 
+  /**
+   * Class name of the implementation of interface {@link com.linkedin.venice.controller.repush.RepushOrchestrator} in {@link com.linkedin.venice.controller.logcompaction.CompactionManager}
+   */
+  public static final String REPUSH_ORCHESTRATOR_CLASS_NAME = "repush.orchestrator.class.name";
+
+  /**
+   * Prefix of configs to configure RepushOrchestrator
+   */
+  public static final String CONTROLLER_REPUSH_PREFIX = "controller.repush.";
```
mynameborat, line 246: Update `REPUSH_ORCHESTRATOR_CLASS_NAME` to also use the same prefix.
`s/controller.repush.repush.orchestrator.class.name/controller.repush.orchestrator.class.name`

internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
```diff
@@ -235,6 +235,18 @@ private ConfigKeys() {
...
+
+  /**
+   * Number of threads to use for scheduled log compaction
+   */
+  public static final String SCHEDULED_LOG_COMPACTION_THREAD_COUNT = "scheduled.log.compaction.thread.count";
+
+  /**
+   * Time between each scheduled log compaction
+   */
+  public static final String SCHEDULED_LOG_COMPACTION_INTERVAL_MS = "scheduled.log.compaction.interval.ms";
```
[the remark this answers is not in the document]
mynameborat: Ping on this comment.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestHybrid.java
```diff
@@ -1029,6 +1043,53 @@ public void testHybridWithZeroLagThreshold() throws Exception {
...
+      System.out.println("Log compaction job failed");
+      throw new RuntimeException(e);
+    }
+  }
+
+  public static class TestRepushOrchestratorImpl implements RepushOrchestrator {
+    @Override
+    public RepushJobResponse repush(String storeName) {
+      latch.countDown();
+      System.out.println("Repush job triggered for store: " + storeName);
```
mynameborat, line 1088: Please remove this or change this to log statement

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/logcompaction/TestRepushOrchestratorImpl.java
```diff
@@ -0,0 +1,18 @@
+package com.linkedin.venice.endToEnd.logcompaction;
+
+import com.linkedin.venice.controller.repush.RepushJobResponse;
+import com.linkedin.venice.controller.repush.RepushOrchestrator;
+
+
+/**
+ * This is a dummy no-op implementation of {@link RepushOrchestrator} for testing purposes in {@link TestLogCompactionService}.
+ */
+public class TestRepushOrchestratorImpl implements RepushOrchestrator {
```
mynameborat, line 10: Can this be removed? Looks like we already have one inside the test class.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceController.java
```diff
@@ -190,6 +192,11 @@ private void createServices() {
     unusedValueSchemaCleanupService = Optional.empty();
 
     Admin admin = controllerService.getVeniceHelixAdmin();
+    if (multiClusterConfigs.isLogCompactionEnabled()) {
+      logCompactionService = new LogCompactionService(admin, multiClusterConfigs);
+      LOGGER.info("LogCompactionService is initialised");
```
mynameborat, line 197: nit: logging statements doesn't have follow camelCasing. `s/LogCompactionService/Log compaction service is initialized`

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceController.java
```diff
@@ -312,6 +324,7 @@ public void stop() {
     storeBackupVersionCleanupService.ifPresent(Utils::closeQuietlyWithErrorLogged);
     disabledPartitionEnablerService.ifPresent(Utils::closeQuietlyWithErrorLogged);
     Utils.closeQuietlyWithErrorLogged(topicCleanupService);
+    Utils.closeQuietlyWithErrorLogged(logCompactionService);
```
mynameborat, line 327: Should this be closed only if the `isLogCompactionEnabled()`?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerClusterConfig.java
```diff
@@ -967,6 +988,14 @@ public VeniceControllerClusterConfig(VeniceProperties props) {
     this.serviceDiscoveryRegistrationRetryMS =
         props.getLong(SERVICE_DISCOVERY_REGISTRATION_RETRY_MS, 30L * Time.MS_PER_SECOND);
     this.pushJobUserErrorCheckpoints = parsePushJobUserErrorCheckpoints(props);
+    this.repushOrchestratorClassName = props.getString(REPUSH_ORCHESTRATOR_CLASS_NAME); // TODO LC: default value?
```
mynameborat, line 991: It doesn't have default value. Maybe throw an error if log compaction service is enabled but this configuration is null or missing?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -7467,6 +7487,49 @@ public Map<String, StoreDataAudit> getClusterStaleStores(String clusterName) {
...
+   * - serves as API endpoint to trigger scheduled & adhoc log compaction
+   * @param storeName
+   * @return
+   */
+  @Override
+  public RepushJobResponse compactStore(String storeName) {
+    try {
+      return compactionManager.compactStore(storeName);
+    } catch (Exception e) {
+      LOGGER.error("Error while compacting store: {}", storeName, e);
```
mynameborat, line 7521: log format statement missing additional {} given we pass two arguments.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -602,6 +607,21 @@ public VeniceHelixAdmin(
...
+      Class<? extends RepushOrchestrator> repushOrchestratorClass =
+          ReflectUtils.loadClass(multiClusterConfigs.getRepushOrchestratorClassName());
+      Class<VeniceProperties> venicePropertiesClass = ReflectUtils.loadClass(VeniceProperties.class.getName());
+      RepushOrchestrator repushOrchestrator = ReflectUtils.callConstructor(
+          repushOrchestratorClass,
+          new Class[] { venicePropertiesClass },
+          new Object[] { multiClusterConfigs.getRepushOrchestratorConfigs() });
+      compactionManager =
+          new CompactionManager(repushOrchestrator, multiClusterConfigs.getTimeSinceLastLogCompactionThresholdMS());
+      // TODO LC: pass VeniceProperties instead of multiClusterConfigs.<specific config>
```
mynameborat, line 622: Are you doing this in this PR? If not, can we drop the comment.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -7467,6 +7487,49 @@ public Map<String, StoreDataAudit> getClusterStaleStores(String clusterName) {
...
+    try {
+      return compactionManager.compactStore(storeName);
+    } catch (Exception e) {
+      LOGGER.error("Error while compacting store: {}", storeName, e);
+      throw e; // this method is the first common point for scheduled & adhoc log compaction, each has different error
+               // handling. Thus, the exception is rethrown after logging.
+    }
+  }
+
+  // TODO: for testing
```
mynameborat, line 7527: nit: remove `TODO`

services/venice-controller/src/main/java/com/linkedin/venice/controller/logcompaction/LogCompactionService.java
```diff
@@ -0,0 +1,118 @@
...
+    return true;
+  }
+
+  @Override
+  public void stopInner() throws Exception {
+    executor.shutdown();
+    try {
+      if (!executor.awaitTermination(SCHEDULED_EXECUTOR_TIMEOUT_S, TimeUnit.SECONDS)) {
+        executor.shutdownNow();
+        LOGGER.info("log compaction service shut down gracefully");
```
mynameborat, line 66: Update log statement to describe timeout to make it explicit.

services/venice-controller/src/main/java/com/linkedin/venice/controller/logcompaction/LogCompactionService.java
```diff
@@ -0,0 +1,118 @@
...
+    private final Set<String> clusters;
+    private final String triggerSource;
+
+    private LogCompactionTask(Set<String> clusters, String triggerSource) {
+      this.clusters = clusters;
+      this.triggerSource = triggerSource;
+    }
+
+    @Override
+    public void run() {
```
mynameborat, line 84: Can we fold this into run itself? Seems it doesn't do much in its current state.

services/venice-controller/src/main/java/com/linkedin/venice/controller/repush/RepushJobResponse.java
```diff
@@ -0,0 +1,36 @@
+package com.linkedin.venice.controller.repush;
+
+/**
+ * Data model of response from a repush job trigger request for a store
+ */
+public class RepushJobResponse {
+  private final String storeName;
+  private final String execId;
+  private final String execUrl;
+  private final String execDate;
```
mynameborat, line 10: what is this parameter for? It should be timestamp instead of date if at all we need one.

services/venice-controller/src/main/java/com/linkedin/venice/controller/repush/RepushJobResponse.java
```diff
@@ -0,0 +1,36 @@
+package com.linkedin.venice.controller.repush;
+
+/**
+ * Data model of response from a repush job trigger request for a store
+ */
+public class RepushJobResponse {
+  private final String storeName;
+  private final String execId;
+  private final String execUrl;
```
mynameborat, line 9: `s/execId/executionId` and same for the rest.

Also, how is `execUrl` used? Consider adding another API in `RepushOrchestrator` that provides status for a given `executionId`. The status API can provide additional information like,
* when it was launched
* what store it is for
* which concrete orchestrator platform it uses
* weblink 
etc

services/venice-controller/src/test/java/com/linkedin/venice/controller/logcompaction/TestLogCompactionService.java
```diff
@@ -0,0 +1,70 @@
...
+    logCompactionService.startInner();
+    if (latch.await(testDuration, TimeUnit.SECONDS)) {
+      logCompactionService.stopInner();
+    }
+    // ensures you have invoked the compact store once
+    verify(admin, atLeast(2)).getStoresForCompaction(any());
+    verify(admin, times(expectedCompactStoreInvocationCount)).compactStore(storeForCompaction);
+  }
+
+  // TODO: test LogCompactionTask::run() focus on edge cases
```
mynameborat, line 69: Are you adding one?

mynameborat (commented):
I think overall the PR looks fine. General comments
* Please remove any TODOs if you are not planning to address them in the PR or short-term
* Mostly minor comments on code style. 
* Update the status of the PR from draft.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestHybrid.java
```diff
@@ -136,6 +144,12 @@ public class TestHybrid {
   private static final Logger LOGGER = LogManager.getLogger(TestHybrid.class);
   public static final int STREAMING_RECORD_SIZE = 1024;
 
+  // Log compaction test constants
+  private static final int TEST_TIMEOUT = 999999; // ms
```
lusong64, line 148: Using Integer.MAX_VALUE ?

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestHybrid.java
```diff
@@ -1029,6 +1043,53 @@ public void testHybridWithZeroLagThreshold() throws Exception {
...
+    LOGGER.info("LogCompactionService test store created: {}", storeName);
+    Admin admin = sharedVenice.getLeaderVeniceController().getVeniceAdmin();
+    String clusterName = sharedVenice.getClusterName();
+
+    StoreInfo compactionReadyStore = sharedVenice.getControllerClient().getStore(storeName).getStore();
+    Assert.assertNotNull(compactionReadyStore.getHybridStoreConfig());
+    Assert.assertTrue(admin.getCompactionManager().isCompactionReady(compactionReadyStore));
+
+    // Wait for the latch to count down
+    latch = new CountDownLatch(1);
```
lusong64, line 1073: Why can't you put the initialization of the countDownLatch into its declaration? If you only need to use it in this method, you don't have to declare it as a static member as the scope of static is unnecessarily too large.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/logcompaction/TestLogCompactionService.java
```diff
@@ -0,0 +1,125 @@
...
+        new VeniceControllerCreateOptions.Builder(CLUSTER_NAME, zkServer, pubSubBrokerWrapper)
+            .extraProperties(extraProperties)
+            .regionName(STANDALONE_REGION_NAME)
+            .build());
+    TestUtils.waitForNonDeterministicCompletion(
+        5,
+        TimeUnit.SECONDS,
+        () -> childControllerWrapper.isLeaderController(CLUSTER_NAME));
+
+    latch = new CountDownLatch(1);
```
lusong64, line 71: Same comment as that I put into TestHybrid.java.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceController.java
lusong64: I would argue that the word initialised is used by 68.35 million British. So give it a pass?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
lusong64: The second argument is the exception and log4j expects it.

internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
```diff
@@ -235,6 +235,36 @@ private ConfigKeys() {
...
+
+  /**
+   * Time since last log compaction before a store is considered for log compaction
+   */
+  public static final String TIME_SINCE_LAST_LOG_COMPACTION_THRESHOLD_MS =
```
arjun4084346, line 265: usually the var and config name are same.
so it should have been
`public static final String TIME_SINCE_LAST_LOG_COMPACTION_THRESHOLD_MS = "time.since.last.log.compaction.threashold.ms`

arjun4084346: like you have named other vars above.
or you can change the var name to match the config name

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/logcompaction/TestLogCompactionService.java
```diff
@@ -0,0 +1,125 @@
...
+      throw new RuntimeException(e);
+    }
+
+    // // Verify that repush was called the expected number of timesCapture the argument passed to repush()
+    // ArgumentCaptor<String> argumentCaptor = ArgumentCaptor.forClass(String.class);
+    // verify(mockRepushOrchestrator, times(expectedRepushInvocationCount)).repush(argumentCaptor.capture());
+    //
+    // // Assert the value of the captured argument
+    // assertEquals(argumentCaptor.getValue(), "" /* TODO LC: store to be compacted */);
+    Utils.sleep(20000);
```
arjun4084346, line 109: Utils.sleep(20000L)
as the param is long :)

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceController.java
WhitneyDeng: I see your point. But referring to similar patterns like `secureAdminServer` where the service is conditionally started, the enabled flag is not checked upon close()ing.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceController.java
WhitneyDeng: camel casing reverted. british english stands!

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestHybrid.java
WhitneyDeng: all System.out.println changed to logs

internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
WhitneyDeng: `repush.orchestrator.class.name` changed to `controller.repush.orchestrator.class.name`

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/logcompaction/TestRepushOrchestratorImpl.java
WhitneyDeng: removed

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
WhitneyDeng: deleted

services/venice-controller/src/main/java/com/linkedin/venice/controller/logcompaction/LogCompactionService.java
WhitneyDeng: Do you mean describe how long shutdown took?

services/venice-controller/src/main/java/com/linkedin/venice/controller/logcompaction/LogCompactionService.java
WhitneyDeng: `errorHandlingWrapper()` content folded into `run()`

services/venice-controller/src/main/java/com/linkedin/venice/controller/repush/RepushJobResponse.java
WhitneyDeng: see logical_date (previously execution_date) of type String in postDagRun(): https://airflow.apache.org/docs/apache-airflow/stable/stable-rest-api-ref.html#operation/post_dag_run

services/venice-controller/src/main/java/com/linkedin/venice/controller/repush/RepushJobResponse.java
WhitneyDeng: execUrl is the link to the AirflowUI, referring to its usage in SmartAssistantHandler

`return String.format("%s/admin/airflow/graph?dag_id=%s&execution_date=%s&run_id=%s",
          airflowClient.getBasePath(), SMART_ASSISTANT_DAG_ID, result.getLogicalDate(),
          result.getDagRunId());
`
https://github.com/linkedin-multiproduct/nuage-venice/pull/696/files#diff-0dd3d578a191ea41c4cd025ddbe113157a3785de4648368e87e139a1047ac997

services/venice-controller/src/test/java/com/linkedin/venice/controller/logcompaction/TestLogCompactionService.java
WhitneyDeng: If any edge cases comes to mind, I welcome your input.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestHybrid.java
WhitneyDeng: Fair comment, but here's the catch: `latch.countDown()` happens in `TestRepushOrchestratorImpl.repush()` which is outside this `testHybridStoreLogCompaction()` function.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/logcompaction/TestLogCompactionService.java
WhitneyDeng: Same usage pattern as `latch` in `TestHybrid`

internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
WhitneyDeng: agreed. config string changed to `log.compaction.last.compaction.time.threshold.ms`
https://github.com/linkedin/venice/pull/1282/commits/1a18f0eb5b11124e6dfbf3f96c98309efea30e35

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/logcompaction/TestLogCompactionService.java
WhitneyDeng: Good shout! Thank you, Arjun. `L` appended. https://github.com/linkedin/venice/pull/1282/commits/654444af501cb1ad439d553b5b531fe6b19913f4

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestHybrid.java
lusong64: That sounds quite weird. Let me explain why: it is a test but your unit tests relies on an action that would be triggered in a method of a different class.

lusong64: The bigger problem is that in setup and this method you initialize the countDownlatch twice. if you have another test method that also inits it in the future, you literally created a race condition among test methods. For example, one test method does the countdown may not be the one that was assigned to the static variable before. And it would be really hard to troubleshoot. 

Is there anyway to keep countdown latch local?

lusong64 (changes requested):

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/logcompaction/TestLogCompactionService.java
WhitneyDeng: implemented. reduced `latch` scope: https://github.com/linkedin/venice/pull/1282/commits/f577cc771239ad543e1e409eb1fa74ade5be464c

services/venice-controller/src/main/java/com/linkedin/venice/controller/repush/RepushJobResponse.java
WhitneyDeng: deleted execDate. https://github.com/linkedin/venice/pull/1282/commits/589ba1bff244767a1f014dece586fab8883196fd

services/venice-controller/src/main/java/com/linkedin/venice/controller/logcompaction/LogCompactionService.java
mynameborat: it says shut down gracefully but in reality it is attempting to shut down forcefully because waiting for normal shutdown failed due to timeout.

Something along the lines of `Timed out waiting for executor to shutdown. Attempting to forcefully close or shut it down`

services/venice-controller/src/main/java/com/linkedin/venice/controller/Admin.java
```diff
@@ -932,6 +934,24 @@ default boolean isAdminTopicConsumptionEnabled(String clusterName) {
...
+   * @return the list of stores ready for compaction
+   */
+  List<StoreInfo> getStoresForCompaction(String clusterName);
+
+  /**
+   * triggers repush for storeName for log compaction of store topic implemented in
+   * {@link VeniceHelixAdmin#compactStore}
+   *
+   * @param storeName
+   * @return
```
kvargha, line 949: Can we add some explanation on the return JavaDoc?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
kvargha: Does the error need to be logged here, if we're throwing it on the next line?

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestHybrid.java
xunyin8: Can we add some reasonable timeout to the tests and rename this to something like LC_TIMEOUT? i.e.
```
private static final int TEST_TIMEOUT = 60000; 

@Test(timeOut = TEST_TIMEOUT)
```
This way when something does go wrong we don't drag the integration test for 999999ms (16minutes) or `Integer.MAX_VALUE` (24.8 days)

services/venice-controller/src/test/java/com/linkedin/venice/controller/logcompaction/TestCompactionManager.java
```diff
@@ -0,0 +1,77 @@
+package com.linkedin.venice.controller.logcompaction;
+
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyInt;
+import static org.mockito.Mockito.*;
+import static org.testng.Assert.*;
```
kvargha, line 6: Nit: Please try to remove wildcard imports

services/venice-controller/src/test/java/com/linkedin/venice/controller/logcompaction/TestLogCompactionService.java
```diff
@@ -0,0 +1,70 @@
+package com.linkedin.venice.controller.logcompaction;
+
+import static org.mockito.Mockito.*;
```
kvargha, line 3: Nit: Please try to remove wildcard imports

services/venice-controller/src/test/java/com/linkedin/venice/controller/logcompaction/TestLogCompactionService.java
```diff
@@ -0,0 +1,70 @@
...
+    when(multiClusterConfigs.getScheduledLogCompactionIntervalMS())
+        .thenReturn(TimeUnit.SECONDS.toMillis(scheduleInterval));
+
+    String storeForCompaction = "hello";
+    StoreInfo mockStoreInfo = mock(StoreInfo.class);
+    doReturn(storeForCompaction).when(mockStoreInfo).getName();
+
+    // compactStores() invoked for non-empty list returned
+    when(admin.getStoresForCompaction(anyString())).thenReturn(Collections.emptyList());
+    when(admin.getStoresForCompaction(anyString())).thenReturn(Collections.singletonList(mockStoreInfo));
```
kvargha, line 51: The first statement gets overriden by the second one

kvargha (changes requested):

internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
```diff
@@ -266,6 +266,56 @@ private ConfigKeys() {
...
+
+  /**
+   * Username for Airflow client authentication
+   */
+  public static final String REPUSH_AIRFLOW_USERNAME = "controller.repush.airflowUsername";
+
+  /**
+   * Password for Airflow client authentication
+   */
+  public static final String REPUSH_AIRFLOW_PASSWORD = "controller.repush.airflowPassword";
```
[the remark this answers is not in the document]
WhitneyDeng: point taken. I was referencing the config naming convention in the [SmartAssistant PR](https://github.com/linkedin-multiproduct/nuage-venice/pull/696/files#diff-4c8a67ebd805e693553081b5b4919f06522d66f24af7106c4705ef7c4db69fa1). 

camel case converted to lowercase 
`venice`: https://github.com/linkedin/venice/pull/1282/commits/aa10e2941317e686f9ecdb655633d748dd804efd 
`venice-backend`: https://github.com/linkedin-multiproduct/venice-backend/pull/2061/commits/bf85f0ca6d3e1f5690e492abe6136aee570a319a

internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
```diff
@@ -266,6 +266,56 @@ private ConfigKeys() {
...
 
+  /**
+   * Class name of the implementation of interface {@link com.linkedin.venice.controller.repush.RepushOrchestrator} in {@link com.linkedin.venice.controller.logcompaction.CompactionManager}
+   */
+  public static final String REPUSH_ORCHESTRATOR_CLASS_NAME = "controller.repush.orchestrator.class.name";
+
+  /**
+   * Airflow Url for repush DAG
+   */
+  public static final String REPUSH_AIRFLOW_URL = "controller.repush.airflow.url";
```
sixpluszero, line 277: We should not expose Airflow concept in OSS venice though. Technically it can be any service to instrument this repush

sushantmane: I agree with this approach. We can define a dedicated namespace for compaction service configs and let the extraction happen within that service. For example, using a pattern like controller.repush.<stdConstant>.config.name, and then filtering out controller.repush.stdConstant settings within the service.

services/venice-controller/src/main/java/com/linkedin/venice/controller/logcompaction/LogCompactionService.java
```diff
@@ -0,0 +1,114 @@
...
+                triggerSource,
+                clusterName,
+                response.getName(),
+                response.getExecutionUrl());
+          } catch (Exception e) {
+            LOGGER.error(
+                "Error checking if store is ready for log compaction for cluster: {} store: {}",
+                clusterName,
+                storeInfo.getName(),
+                e);
```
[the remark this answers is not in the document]
WhitneyDeng: Will have follow up metrics/observability PR. Have you any additional concerns?

internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
WhitneyDeng: good shout. all Airflow mentions removed:
https://github.com/linkedin/venice/pull/1282/commits/00c2273d08d6e67872450c57b981e6d657d2e528
https://github.com/linkedin/venice/pull/1282/commits/8997015ace8dc4870f47c99beb561d3063cb8c1f
https://github.com/linkedin/venice/pull/1282/commits/2d6ecba736acf4055de2c137c6c1a3f42baaaa66

services/venice-controller/src/test/java/com/linkedin/venice/controller/logcompaction/TestLogCompactionService.java
WhitneyDeng: good catch. changed to single class imports https://github.com/linkedin/venice/pull/1282/commits/5e5fa20e180b2004d32573f38e5a1b089946d668

services/venice-controller/src/test/java/com/linkedin/venice/controller/logcompaction/TestCompactionManager.java
WhitneyDeng: eagle eyed. changed to single class imports https://github.com/linkedin/venice/pull/1282/commits/6ce15e367921ead943fad8d287351fc15fa991e7

services/venice-controller/src/main/java/com/linkedin/venice/controller/logcompaction/LogCompactionService.java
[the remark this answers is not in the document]
WhitneyDeng: @bharathkk What's the level of alerting do you think this warrants?

services/venice-controller/src/main/java/com/linkedin/venice/controller/logcompaction/LogCompactionService.java
WhitneyDeng: done. https://github.com/linkedin/venice/pull/1282/commits/b360da8bf3a5445a89cf16211996d69ae8f92d9e

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
WhitneyDeng: `TODO` removed. comment kept for documentation https://github.com/linkedin/venice/pull/1282/commits/33c054e6920abd26eb6a867ad1e9aff7860ec2e8

services/venice-controller/src/main/java/com/linkedin/venice/controller/logcompaction/LogCompactionService.java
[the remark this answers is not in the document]
mynameborat: I think this version of PR doesn't fully address (rather doesn't take it up at all) metrics and alerting. @kvargha We plan to follow it up with adding metrics extensively across the feature flow at which point this suggestion would be more appropriate.

Let us know if its okay to defer it to that point.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceController.java
mynameborat: Probably an oversight instead of a pattern.
It is good to just perform the clean up only if you perform startup process. Otherwise you are either looking at error logs that are misleading

WhitneyDeng: good call. shutdown conditioned on `isLogCompactionEnabled()` https://github.com/linkedin/venice/pull/1282/commits/5551f4c5adb50ea46daabfea3e264d381bd4fc83

services/venice-controller/src/test/java/com/linkedin/venice/controller/logcompaction/TestLogCompactionService.java
WhitneyDeng: redundant stub test class deleted https://github.com/linkedin/venice/pull/1282/commits/8b499451222ec74a9beee827ceb49d18e46a38ee

services/venice-controller/src/main/java/com/linkedin/venice/controller/repush/RepushOrchestratorConfig.java
```diff
@@ -0,0 +1,40 @@
...
+  private final String airflowUsername;
+  private final String airflowPassword;
+  private final String dagId;
+
+  // TODO: exception here if any of the configs are empty
+  public RepushOrchestratorConfig(VeniceProperties repushClusterConfigs) {
+    this.airflowUrl = repushClusterConfigs.getString(AIRFLOW_URL_CFG_KEY);
+    this.airflowUsername = repushClusterConfigs.getString(AIRFLOW_USERNAME_CFG_KEY);
+    this.airflowPassword = repushClusterConfigs.getString(AIRFLOW_PASSWORD_CFG_KEY);
+    this.dagId = repushClusterConfigs.getString(DAG_ID_CFG_KEY);
```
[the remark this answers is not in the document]
WhitneyDeng: Airflow references deleted https://github.com/linkedin/venice/pull/1282#discussion_r1956604272

services/venice-controller/src/main/java/com/linkedin/venice/controller/logcompaction/LogCompactionService.java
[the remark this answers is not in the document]
kvargha: Sounds good

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestHybrid.java
WhitneyDeng: fair comments.

lowered timeout duration: https://github.com/linkedin/venice/pull/1282/commits/c44671f5d9d80393f305d3f875719b4db8812cb0
renamed constant to log compaction specific: https://github.com/linkedin/venice/pull/1282/commits/fa37404d127613cd146f782f93e48dc2171acb0d

services/venice-controller/src/main/java/com/linkedin/venice/controller/Admin.java
WhitneyDeng: good shout. added https://github.com/linkedin/venice/pull/1282/commits/c0047362027747880dec460bd2fbcf4531fd0d94

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/logcompaction/TestLogCompactionService.java
```diff
@@ -0,0 +1,131 @@
...
+      throw new RuntimeException(e);
+    }
+
+    // // Verify that repush was called the expected number of timesCapture the argument passed to repush()
+    // ArgumentCaptor<String> argumentCaptor = ArgumentCaptor.forClass(String.class);
+    // verify(mockRepushOrchestrator, times(expectedRepushInvocationCount)).repush(argumentCaptor.capture());
+    //
+    // // Assert the value of the captured argument
+    // assertEquals(argumentCaptor.getValue(), "" /* TODO LC: store to be compacted */);
+    Utils.sleep(20000L);
```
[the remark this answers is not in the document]
WhitneyDeng: class deleted https://github.com/linkedin/venice/pull/1282/commits/8b499451222ec74a9beee827ceb49d18e46a38ee

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
WhitneyDeng: `VeniceHelixAdmin::compactStore()` is where the adhoc/API & scheduled/LogCompactionService log compaction paths intersect. The thinking is to log the error at the common point then throw the error back to the respective paths for specific handling (e.g. scheduled might want to retry. adhoc wants to send error response & leave recourse up to user)

kvargha: Makes sense to me

kvargha (approved):

lusong64 (approved):
Whitney had an offline conversation about the latch thing and we resolved it.

internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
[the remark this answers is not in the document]
WhitneyDeng: removed scheduled prefix https://github.com/linkedin/venice/pull/1282/commits/470b47c1e3fdc7c1b7c69f694dfa476ca0dc3771

services/venice-controller/src/main/java/com/linkedin/venice/controller/server/StoresRoutes.java
```diff
@@ -937,6 +907,37 @@ public void internalHandle(Request request, MultiStoreInfoResponse veniceRespons
...
+        veniceResponse.setCluster(cluster);
+      }
+    };
+  }
+
+  /**
+   * @see Admin#compactStore(String)
+   */
+  public Route compactStore(Admin admin) {
+    return new VeniceRouteHandler<ControllerResponse>(ControllerResponse.class) {
```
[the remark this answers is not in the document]
WhitneyDeng: returns RepushJobResponse which has execution id and store name. https://github.com/linkedin/venice/pull/1282/commits/0a618c3f12659fbec5ef42f8c13098df5b9ab071

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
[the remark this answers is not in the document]
WhitneyDeng: VPHA::compactStore() implemented: https://github.com/linkedin/venice/pull/1282/commits/98171a5f2332e6d5337f701c1d115c90ab6c20aa#diff-36114654b45af0adf03521759f0301f267c4b453321c47a4d1c1c4726abddca2

services/venice-controller/src/main/java/com/linkedin/venice/controller/repush/RepushJobResponse.java
WhitneyDeng: deleted execution url from response object https://github.com/linkedin/venice/pull/1282/commits/f2951ac2668ba5f2e6897717b4eabdb1a8127f2e

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -602,6 +606,13 @@ public VeniceHelixAdmin(
     dataRecoveryManager =
         new DataRecoveryManager(this, icProvider, pubSubTopicRepository, participantStoreClientsManager);
 
+    // TODO extends interchangeable with implements?
+    Class<? extends RepushOrchestrator> repushOrchestratorClass =
+        ReflectUtils.loadClass(multiClusterConfigs.getRepushOrchestratorClassName());
+    RepushOrchestrator repushOrchestrator =
+        ReflectUtils.callConstructor(repushOrchestratorClass, new Class[0], new Object[0]);
+    compactionManager = new CompactionManager(repushOrchestrator);
```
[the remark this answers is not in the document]
WhitneyDeng: Follow up PR. VeniceProperties + ReflectionsUtils for now. discussed offline.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerClusterConfig.java
WhitneyDeng: discussed offline: if service enabled and this config missing, bring down controller.

services/venice-controller/src/main/java/com/linkedin/venice/controller/server/StoresRoutes.java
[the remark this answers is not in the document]
WhitneyDeng: discussed offline: follow up PR for metrics/observability

PR #1288, opened 2024-11-07
[server] Augment completion report with previous ready to serve state

ZacAttack:
## [server] Augment completion report with previous ready to serve state

This change persists into the offset record weather or not this partition originally marked itself as ready to serve.  This is then used on restart to determine if a node should report completion or not when coming online.

The intention is that when a node is lagged under normal conditions (things like heavy load or some other bad condition), we can still restart the node or do deployments and have the node come up and serve traffic.  This is done under the notion that it's generally preferable to be online and serving and stale as opposed to caught up/catching up, but unavailable.

This will not impact buffer replay on a version push because a replay won't pass a ready to serve check.

**ALSO**

Did some test code cleanup in StoreIngestionTask which had an excessive amount of function overloading.  Switched it over to use a config object instead and migrated the code over.

Resolves #XXX


clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -2031,7 +2031,8 @@ private void checkConsumptionStateWhenStart(
                 offsetLag,
                 previousOffsetLag,
                 offsetLagThreshold);
-            if (offsetLag < previousOffsetLag + offsetLagDeltaRelaxFactor * offsetLagThreshold) {
+            if (newPartitionConsumptionState.getReadyToServeInOffsetRecord()
```
sixpluszero, line 2034: This means we might see deployment stuck / slowdown in next round of deploymenmt, due to the fact that certain resources were falsely marked as RTS with lag right? Although I think it is actually a good catch and it should only affect one time.

sixpluszero (approved):

PR #44, opened 2022-10-07
[server][dvc][fc][vpj][doc] Lots of tidying up, mainly in DVC & server

FelixGV:
Stability improvements:

1. The cached state in MainIngestionStorageMetadataService could get out of sync with the one in AbstractStoragePartition. To resolve this, the MainIngestionStorageMetadataService now takes in a storeVersionStateSyncer at construction-time, thus giving it a constrained API for mutating the internal state of the correct objects when necessary. In turn, this means we can retrieve the StoreVersionState either out of the StorageMetadataService (which requires an extra lookup) or out of AbstractStoragePartition and both are in-sync with one another.

2. Replaced the StorageMetadataService interface's put function with computeStoreVersionState, which provides first-class support for read-modify-write operations against the StoreVersionState, with a promise of proper synchronization. There are 3 code paths doing RMW operations against the StoreVersionState, so they now have proper synchronization. In practice, the race condition seems unlikely (perhaps impossible) between the current code paths, so this is defensive coding more than anything else.

3. Fixed a mem leak in the topicStorageEngineReferenceMap inside the DefaultIngestionBackend. It is now cleaned up from the close of VersionBackend.

4. Fixed a NPE in the ParticipantStoreConsumptionTask, and in the StorageUtilizationManager.

5. In SIT::produceToStoreBufferServiceOrKafka, processStartOfPush is now invoked as soon as we get a SOP, so that the StoreVersionState can be primed early, rather than later, in processControlMessage, which is called after passing through the StoreBufferService. This enables SIT::deserializeValue to perform its defensive check on compressed topics as well (which was previously unsupported). Also eliminated the logging of the value in case deserialization fails.

Object allocation reduction:

1. StoreVersionState, PartitionConsumptionState and the future inside StoreBufferService.QueueNode are not wrapped in Optionals anymore.

2. StoreBufferService.QueueNode is now specialized into 2 subclasses, each having different properties, resulting in 1 fewer null per ingested msg.

3. The compactingPartitions in SIT::produceToStoreBufferServiceOrKafka now only gets allocated if the related functionality is on.

Performance improvement:

1. ChunkingUtils and all related adapters now take in a specific compressor instance rather than a compressor factory, which means the responsibility of looking up the compressor is moved out of the adapters and into the calling code. This used to result in 1 map lookup for NO_OP and GZIP compression strategies, and 2 map lookups for ZSTD. The new structure amortizes the lookups better:
[...]

PR #1290, opened 2024-11-07
[server][dvc] Open store before bootstrapping from blob transfer.

jingy-li:
## Summary
Open the store before bootstrapping from blob transfer, otherwise the offset record and StorageVersionState can not update successfully. 

## How was this PR tested?
Run at 2 dvc hosts. 
1. The first host successfully bootstraps from Kafka, because there is no any available hosts, and successfully generate snapshot.
Screenshot 2024-11-07 at 11 11 46 AM
2. The second host successfully bootstraps from blob transfer, and without snapshot, because snapshot will generate when it is ingested from Kafka. 
Screenshot 2024-11-07 at 11 16 41 AM


clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/DefaultIngestionBackend.java
```diff
@@ -79,6 +80,7 @@ public void startConsumption(VeniceStoreVersionConfig storeConfig, int partition
     if (!storeAndVersion.getFirst().isBlobTransferEnabled() || blobTransferManager == null) {
       runnable.run();
     } else {
+      storageService.openStore(storeConfig, svsSupplier);
```
gaojieliu, line 83: @FelixGV 
I saw you modified this interface to pass `svsSupplier` in the past:
https://github.com/linkedin/venice/pull/44

For the usage in this class, it is a little confusing:
```
Supplier<StoreVersionState> svsSupplier = () -> storageMetadataService.getStoreVersionState(storeVersion);
    syncStoreVersionConfig(storeAndVersion.getFirst(), storeConfig);

    Runnable runnable = () -> {
      AbstractStorageEngine storageEngine =
          storageService.openStoreForNewPartition(storeConfig, partition, svsSupplier);
```

So it first looks up SVS state from the metadata partition and then pass it to `StorageService#openStoreForNewPartition`, I wondered whether it is necessary to use this pattern as internally if `versionStateCache` is `null` in the get path, it will try to do the lookup internally.

My main question is that whether this param: `svsSupplier` is necessary or not in `openStore` and `openStoreForNewPartition`.

gaojieliu (approved):
Code change itself looks good and I left an open-question to @FelixGV .

clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/DefaultIngestionBackend.java
gaojieliu: I guess the user can pass a `null` supplier for such usage, right?

sixpluszero (dismissed):

sixpluszero (approved):

PR #1291, opened 2024-11-08
[server] Fix NPE for ServerReadQuotaUsageStats when store has no version or traffic

xunyin8:
## [server] Fix NPE for ServerReadQuotaUsageStats when store has no version or traffic

1. Prevent NPE in ServerReadQuotaUsageStats's async gauges when a store has no version or traffic.
2. Set backup version using version list in handle store change since the old behavior will not set the correct backup version after server restart.

## How was this PR tested?
New unit test

### DIFF
--- a/services/venice-server/src/main/java/com/linkedin/venice/listener/ReadQuotaEnforcementHandler.java
+++ b/services/venice-server/src/main/java/com/linkedin/venice/listener/ReadQuotaEnforcementHandler.java
@@ -469,8 +469,12 @@ public class ReadQuotaEnforcementHandler extends SimpleChannelInboundHandler<Rou
       }
     }
     removeTopics(toBeRemovedTopics);
-    stats.setCurrentVersion(store.getName(), currentVersion);
-    stats.setBackupVersion(store.getName(), backupVersion);
+    if (currentVersion > 0) {
+      stats.setCurrentVersion(store.getName(), currentVersion);
+    }
+    if (backupVersion > 0) {
+      stats.setBackupVersion(store.getName(), backupVersion);
+    }
   }
 
   private Set<String> getStoreTopics(String storeName) {
--- a/services/venice-server/src/main/java/com/linkedin/venice/listener/ReadQuotaEnforcementHandler.java
+++ b/services/venice-server/src/main/java/com/linkedin/venice/listener/ReadQuotaEnforcementHandler.java
@@ -413,9 +413,12 @@ public class ReadQuotaEnforcementHandler extends SimpleChannelInboundHandler<Rou
 
     List<String> topics =
         store.getVersions().stream().map((version) -> version.kafkaTopicName()).collect(Collectors.toList());
+    int currentVersion = store.getCurrentVersion();
+    int backupVersion = 0;
     for (String topic: topics) {
       toBeRemovedTopics.remove(topic);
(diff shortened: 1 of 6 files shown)


services/venice-server/src/main/java/com/linkedin/venice/listener/ReadQuotaEnforcementHandler.java
```diff
@@ -427,8 +430,12 @@ public void handleStoreChanged(Store store) {
          *
          */
         this.onCustomizedViewChange(customizedViewRepository.getPartitionAssignments(topic));
+        if (versionNumber != currentVersion && versionNumber > backupVersion
```
huangminchn, line 433: A safer condition would be "versionNumber < currentVersion" instead of "versionNumber != currentVersion"; WDYT?

services/venice-server/src/test/java/com/linkedin/venice/stats/ServerReadQuotaUsageStatsTest.java
```diff
@@ -21,4 +21,33 @@ public void testGetReadQuotaUsageRatio() {
...
+
+  @Test
+  public void testGetReadQuotaMetricsWithNoVersionOrRecordings() {
+    MetricsRepository metricsRepository = new MetricsRepository();
+    String storeName = "test-store";
+    int currentVersion = 3;
+    int backupVersion = 2;
+    ServerReadQuotaUsageStats stats = new ServerReadQuotaUsageStats(metricsRepository, storeName);
+    // Stats shouldn't fail if the store don't have any versions yet
+    Assert.assertEquals(stats.getVersionedRequestedQPS(backupVersion), 0d);
```
huangminchn, line 33: Thanks for adding the test case!

huangminchn (commented):
Thanks Xun! Left a minor comment.

services/venice-server/src/main/java/com/linkedin/venice/listener/ReadQuotaEnforcementHandler.java
xunyin8: If we rollback the current version then it is valid for the new "backup" version to be greater than the current version right?

huangminchn: IC! I am not sure if we would keep such "backup" version in the system. It's not a big concern, let me resolve this comment, but it would be great if we can double check.

huangminchn (approved):
Thanks for the quick turnaround, Xun!

services/venice-server/src/main/java/com/linkedin/venice/listener/ReadQuotaEnforcementHandler.java
```diff
@@ -413,9 +413,12 @@ public void handleStoreChanged(Store store) {
 
     List<String> topics =
         store.getVersions().stream().map((version) -> version.kafkaTopicName()).collect(Collectors.toList());
+    int currentVersion = store.getCurrentVersion();
```
majisourav99, line 416: Can check if the currentVersion is valid ie != VERSION_ID_UNSET

majisourav99 (approved):
LGTM! Thanks for the quick fix

PR #1293, opened 2024-11-08
[compat] [controller] add rt topic name in store config

arjun4084346:
## Add a field `realTimeTopicName` in store and store-version hybrid config

In future, we want to provide users/admin ability to create a new real time topic, probably with some different configurations, e.g. partition count. This new real time topic would co-exist with the old real time topic for some time, so we want to give it a different name.
This PR will make all the components check the real time topic's name, if it exists, in Store/Store-Version hybrid config. If the config does not have any name, it would be assumed, that there is no new topic right now and we can use the old methodology "<store_name>_rt" to determine the real time topic name.

Verson.composeRealTimeTopic method is moved to Utils.
Utils.getRealTimeTopicName are the core methods of this PR.

When store is not hybrid, current version is also not hybrid, and two older versions are using two different rt topics.
This method will return one of those RT topic, because both the RT topics are "active" and will be having data mirroring tasks copying data to/from each other

Note that, in future, when creating new RT topic, I plan to add condition that there should not exists more than one rt topics, because it will increase data replication processes exponentially. So at one point, we will allow only two RT topics

Resolves #XXX

## How was this PR tested?

added integration tests.


internal/venice-common/src/main/java/com/linkedin/venice/meta/Version.java
```diff
@@ -36,6 +37,14 @@ public interface Version extends Comparable<Version>, DataModelBackedStructure<S
    */
   String VENICE_RE_PUSH_PUSH_ID_PREFIX = "venice_re_push_";
 
+  static String getRealTimeTopicNameIfEmpty(String realTimeTopicName, String storeName) {
+    if (Strings.isBlank(realTimeTopicName)) {
+      return composeRealTimeTopic(storeName);
+    } else {
+      return realTimeTopicName;
+    }
```
lluwm, line 45: Nit, this function can actually be simplified to a one line statement if we use ternary operands (?:). I don't know if it worthy to define a new function with 2 parameters.

internal/venice-common/src/main/java/com/linkedin/venice/meta/VersionImpl.java
```diff
@@ -31,23 +31,40 @@ public class VersionImpl implements Version {
    * Currently used in tests only.
    */
   @Deprecated
-  public VersionImpl(String storeName, int number) {
+  public VersionImpl(String storeName, int number, String realTimeTopicName) {
```
lluwm, line 34: Question: shall we keep the old interface `VersionImpl(String storeName, int number)` and add a new one for explicit setting the realtime topic name? It could reduce lots of changes to the existing testing etc. what do you think? is there any downsides of keeping the old one?

arjun4084346: Yes, it was a pain to update all the references.
The only reason I added it in all the existing constructors because it would be a required field in Version and forgetting to explicitly set it will leads to discrepancy and bugs. @FelixGV do you have any thoughts?

clients/da-vinci-client/src/test/java/com/linkedin/davinci/kafka/consumer/KafkaConsumerServiceDelegatorTest.java
```diff
@@ -544,7 +544,7 @@ public void testKafkaConsumerServiceResubscriptionConcurrency() throws Exception
     KafkaConsumerServiceDelegator delegator =
         new KafkaConsumerServiceDelegator(mockConfig, consumerServiceBuilder, isAAWCStoreFunc);
     PubSubTopicPartition realTimeTopicPartition =
-        new PubSubTopicPartitionImpl(TOPIC_REPOSITORY.getTopic(Version.composeRealTimeTopic(storeName)), 0);
+        new PubSubTopicPartitionImpl(TOPIC_REPOSITORY.getTopic(storeName + Version.REAL_TIME_TOPIC_SUFFIX), 0);
```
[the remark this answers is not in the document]
arjun4084346: Yea, I actually wanted to reduce the use of that method; and eventually remove it. Let me mark it @Deprecated 
Another reason for not using it, is that it does not need to be Version.composeRealTimeTopic(storeName) in tests like this, it can just be any random string.

internal/venice-common/src/main/java/com/linkedin/venice/meta/VersionImpl.java
FelixGV: IMHO it is fine to add this param to all constructors since, as Arjun said, if we leave a constructor without this param and let it fill in the new field with today's behavior (i.e. `Version::composeRealTimeTopic`), then we'll probably end up with bugs...

Most likely, in the end state, we need to completely delete `Version::composeRealTimeTopic`, so making use of it in one of the default constructors here goes against that...

lluwm: I see. It makes sense to me as we will delete its usage eventually.

internal/venice-common/src/main/java/com/linkedin/venice/meta/Version.java
arjun4084346: I also personally do not like one line methods, but two things made me do this, a) it is used at many places, b) it helps in emphasizing on the fact that RT topic name is strongly related to store name.

PR #1294, opened 2024-11-09
[router] Replace unhealthy/tardy requests metrics from percentile to regular bucket

majisourav99:
## Replace unhealthy/tardy requests metrics from percentile to regular bucket

We do not need to detailed percentile metrics of the negative events in route.

## How was this PR tested?

GHCI


gaojieliu (approved):
Thanks!

PR #1298, opened 2024-11-12
[controller] Fixed a bug that logger is trying to access a removed map entry

huangminchn:
## Summary
The top store admin operation entry is already removed before the logger trying to access the top entry. This PR fixes this bug by storing the offset of top entry before hand.

## How was this PR tested?
Github CI


debug.txt
xunyin8, line 1: Accidentally included?

xunyin8 (commented):
LGTM, one minor comment

debug.txt
huangminchn: Ah, yeah, good catch!

KaiSernLim (approved):

xunyin8 (approved):

PR #1299, opened 2024-11-13
Check all stages pass in unit test

majisourav99:
## Check all stages pass in unit test

Even if we have unit test failure, we are able to merge the PR as the last state gets skipped. This PR tries to explicit fail the final completion stage if there are failures in the previous stage.
## How was this PR tested?

simulated GHCI test failure
image


sushantmane (approved):
LGTM! Thanks, @majisourav99!

PR #1300, opened 2024-11-13
[beam] Introduce VeniceCDCIO connector for Apache beam applications

suketkar:
[fast-client] Introduce VeniceCDCIO connector for Apache beam applications

VeniceCDCIO connector allows Apache beam applications to integrate Venice CDC in beam DAG.  

## How was this PR tested?
Functional/Performance tested with internal LinkedIn large scale applications.

No

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -122,6 +122,8 @@ ext.libraries = [
     restliCommon: "com.linkedin.pegasus:restli-common:${pegasusVersion}",
     rocksdbjni: 'org.rocksdb:rocksdbjni:8.8.1',
     samzaApi: 'org.apache.samza:samza-api:1.5.1',
+    beamSdk: 'org.apache.beam:beam-sdks-java-core:2.60.0',
+    beamExtensionAvro: 'org.apache.beam:beam-sdks-java-extensions-avro:2.60.0',
     slf4j: 'org.slf4j:slf4j:1.7.36',
     slf4jApi: 'org.slf4j:slf4j-api:1.7.36',
     slf4jSimple: 'org.slf4j:slf4j-simple:1.7.36',
--- a/integrations/venice-beam/build.gradle
+++ b/integrations/venice-beam/build.gradle
@@ -10,9 +10,7 @@ dependencies {
   implementation project(':clients:da-vinci-client')
   implementation project(':clients:venice-thin-client')
   implementation project(':clients:venice-client')
-
-  implementation libraries.samzaApi
-
+  
   implementation libraries.log4j2api
   implementation libraries.log4j2core
 
--- a/integrations/venice-beam/build.gradle
+++ b/integrations/venice-beam/build.gradle
@@ -1,0 +1,26 @@
+plugins {
+  id 'io.github.lhotari.gradle-nar-plugin'
+}
+
+dependencies {
+  implementation(project(':internal:venice-common')) {
+    exclude module: 'kafka_2.10'
+    exclude group: 'org.scala-lang'
+  }
+  implementation project(':clients:da-vinci-client')
+  implementation project(':clients:venice-thin-client')
+  implementation project(':clients:venice-client')
+
+  implementation libraries.samzaApi
+
+  implementation libraries.log4j2api
+  implementation libraries.log4j2core
+
+  implementation libraries.beamSdk
+  implementation libraries.beamExtensionAvro
+}
+
+ext {
+  // tested in integration test
+  jacocoCoverageThreshold = 0.00
+}
--- a/integrations/venice-beam/build.gradle
+++ b/integrations/venice-beam/build.gradle
@@ -10,7 +10,7 @@ dependencies {
   implementation project(':clients:da-vinci-client')
   implementation project(':clients:venice-thin-client')
   implementation project(':clients:venice-client')
-  
+
   implementation libraries.log4j2api
   implementation libraries.log4j2core
 
@@ -19,6 +19,6 @@ dependencies {
 }
 
 ext {
-  // tested in integration test
+  // to be tested in integration test
   jacocoCoverageThreshold = 0.00
 }
--- a/integrations/venice-beam/build.gradle
+++ b/integrations/venice-beam/build.gradle
@@ -1,7 +1,3 @@
-plugins {
-  id 'io.github.lhotari.gradle-nar-plugin'
-}
-
 dependencies {
   implementation(project(':internal:venice-common')) {
     exclude module: 'kafka_2.10'
--- a/integrations/venice-beam/src/main/java/com/linkedin/venice/beam/consumer/VeniceChangelogConsumerIO.java
+++ b/integrations/venice-beam/src/main/java/com/linkedin/venice/beam/consumer/VeniceChangelogConsumerIO.java
@@ -1,4 +1,4 @@
-package com.linkedin.venice.cdc;
+package com.linkedin.venice.beam.consumer;
 
 import com.google.common.collect.Iterators;
 import com.google.common.collect.PeekingIterator;
@@ -57,10 +57,10 @@ import org.joda.time.Instant;
  * Beam Connector for Venice Change data capture IO. Uses {@link VeniceChangelogConsumer} underneath
  * to pull messages from configured venice store.
  */
-public final class VeniceCDCIO {
-  private static final Logger LOG = LogManager.getLogger(VeniceCDCIO.class);
+public final class VeniceChangelogConsumerIO {
+  private static final Logger LOG = LogManager.getLogger(VeniceChangelogConsumerIO.class);
 
-  private VeniceCDCIO() {
+  private VeniceChangelogConsumerIO() {
   }
 
   public static <K, V> Read<K, V> read() {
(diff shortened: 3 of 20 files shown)


clients/venice-client/src/main/java/com/linkedin/venice/cdc/VeniceCDCIO.java
```diff
@@ -0,0 +1,607 @@
...
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.SynchronousQueue;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.function.BiFunction;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+import javax.annotation.Nullable;
```
FelixGV, line 32: We have been using this annotation flavor elsewhere in the Venice code base, and it is of course fine to keep using it. But I see that this PR introduces similar annotations from (at least?) two other sources: jetbrains and checker framework.

I will admit that I am not well-versed in the nuances between these various flavors of annotations, though I expect that it is likely reasonable to use just one kind of annotation? Unless there are compelling reasons to adopt a new one, then I would suggest sticking to the javax flavor we have already been using, and not introducing the extra jetbrains and checker framework dependencies.

That being said, I'm happy to get educated about the merits of other annotation dependencies if some significant benefit exists.

FelixGV (commented):
Thanks a lot for this contribution! I have been looking forward to it!

I left just a couple of superficial comments about new dependencies.

One more: this commit should probably not be tagged with `[fast-client]`. I am not sure what the right tag would be, perhaps a new one is needed given that this is, essentially, a new kind of client. We'll get back to you on that.

Thanks again for taking the time to contribute this!

build.gradle
```diff
@@ -122,6 +123,9 @@ ext.libraries = [
     restliCommon: "com.linkedin.pegasus:restli-common:${pegasusVersion}",
     rocksdbjni: 'org.rocksdb:rocksdbjni:8.8.1',
     samzaApi: 'org.apache.samza:samza-api:1.5.1',
+    samzaCore:'org.apache.samza:samza-core_2.12:1.5.1',
```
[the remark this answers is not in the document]
suketkar: Good catch! Yes it not required, think that got introduced as part of resolving some other dependency but is extraneous. Removed.

clients/venice-client/src/main/java/com/linkedin/venice/cdc/VeniceCDCIO.java
suketkar: Checker framework has @NonNull annotation vs @NotNull in jetbrains. Believe the former is used for static analysis vs the later during runtime. @NotNull is required here by the code and recommended by the analyzer as well. Happy to switch if a viable alternative to @NotNull exists.

suketkar:
> Thanks a lot for this contribution! I have been looking forward to it!
> 
> I left just a couple of superficial comments about new dependencies.
> 
> One more: this commit should probably not be tagged with `[fast-client]`. I am not sure what the right tag would be, perhaps a new one is needed given that this is, essentially, a new kind of client. We'll get back to you on that.
> 
> Thanks again for taking the time to contribute this!

Thanks. Let me know what tag is recommended, will switch to the same.

FelixGV:
Regarding where to put this code, my preference would be to create a new `venice-beam` module for it. I have published PR #1323 to propose having a new directory for all of the integrations to external systems. I will let you know if/when this PR is merged, and how to proceed accordingly... The commit prefix tag should be simply: `[beam]`

clients/venice-client/src/main/java/com/linkedin/venice/cdc/VeniceCDCIO.java
suketkar: Thanks for the suggestion. Fixed.

integrations/venice-beam/build.gradle
```diff
@@ -0,0 +1,24 @@
...
+  implementation libraries.log4j2api
+  implementation libraries.log4j2core
+
+  implementation libraries.beamSdk
+  implementation libraries.beamExtensionAvro
+}
+
+ext {
+  // to be tested in integration test
+  jacocoCoverageThreshold = 0.00
```
FelixGV, line 23: This is not great but for the initial commit we could let it go... I assume there are no tests in the internal project where this is coming from? Or are there any we may be able to contribute here as well?

In any case, it would be good to bump it up in the future.

settings.gradle
```diff
@@ -71,6 +71,7 @@ include 'internal:alpini:router:alpini-router-impl'
 // 3rd-party system integration modules
 include 'integrations:venice-pulsar'
 include 'integrations:venice-samza'
+include 'integrations:venice-beam'
```
FelixGV, line 74: Thanks for adding the new module here. Just a minor nitpick: can we put it at the top of the list, so it's sorted alphabetically?

Also, if you would like, you could likewise add it to the `navigating_project.md` file along with a one-line description (e.g. "`venice-beam`, which implements the Beam `Read` API, enabling a Beam job to consume the Venice changelog.", or something to that effect...). This will make it show up in the docs at venicedb.org :D ...

FelixGV (commented):
Thanks a lot @suketkar! I think this is almost ready to merge! Just a few more comments, probably last round... thanks for hanging in there!

integrations/venice-beam/src/main/java/com/linkedin/venice/cdc/VeniceCDCIO.java
```diff
@@ -0,0 +1,607 @@
+package com.linkedin.venice.cdc;
```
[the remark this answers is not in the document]
ZacAttack: changelog would at least align with the current naming in the code.  Let's keep it consistent with that for the sake of future code greppers.

FelixGV:
One more thing... looks like `spotbugs` is [complaining](https://github.com/linkedin/venice/actions/runs/11926786639/job/33273745995?pr=1300) about certain things... you can download the report here: https://github.com/linkedin/venice/actions/runs/11926786639/artifacts/2214185373

Or run it on your local:

```
./gradlew :integrations:venice-beam:spotbugsMain
```

Let's fix these issues, or if these are not legit issues, then let's add exclusions to `gradle/spotbugs/exclude.xml`

integrations/venice-beam/src/main/java/com/linkedin/venice/cdc/VeniceChangelogConsumerProvider.java
```diff
@@ -0,0 +1,15 @@
+package com.linkedin.venice.cdc;
+
+import com.linkedin.davinci.consumer.VeniceChangelogConsumer;
+import java.io.Serializable;
+import java.lang.reflect.InvocationTargetException;
+
+
+/** Interface to provide a configured {@link VeniceChangelogConsumer} instance. */
+public interface VeniceChangelogConsumerProvider<K, V> extends Serializable {
```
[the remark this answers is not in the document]
suketkar: Yes this is required. KafkaIO follows the same pattern https://github.com/apache/beam/blob/master/sdks/java/io/kafka/src/main/java/org/apache/beam/sdk/io/kafka/KafkaIO.java

integrations/venice-beam/build.gradle
suketkar: The tests are internal application specific and unfortunately no generic tests that can be ported currently. We relied on testing in dark environment given the dependency on Venice and Beam. Yes agree on tests should be added in future.

suketkar:
> [complaining](https://github.com/linkedin/venice/actions/runs/11926786639/job/33273745995?pr=1300)

sketkar-mn1:venice sketkar$ ./gradlew :integrations:venice-beam:spotbugsMain

> Configure project :
Setting up default git config
Setting core.hooksPath = gradle/githooks
Setting blame.ignoreRevsFile = .git-blame-ignore-revs
Setting branch.autoSetupMerge = true
Setting branch.autoSetupRebase = always
Setting pull.rebase = true
JDK version 11 is valid.

Deprecated Gradle features were used in this build, making it incompatible with Gradle 8.0.

You can use '--warning-mode all' to show the individual deprecation warnings and determine if they come from your own scripts or plugins.

See https://docs.gradle.org/7.3/userguide/command_line_interface.html#sec:command_line_warnings

BUILD SUCCESSFUL in 6s
50 actionable tasks: 1 executed, 49 up-to-date

FelixGV (dismissed):
LGTM! Thanks a lot Sunny!

integrations/venice-beam/src/main/java/com/linkedin/venice/beam/consumer/VeniceChangelogConsumerIO.java
```diff
@@ -0,0 +1,609 @@
...
+
+    public RemoveMetadata withoutMetadata() {
+      return new RemoveMetadata();
+    }
+
+    public CurrentValueTransform withOnlyCurrentValue(Coder returnTypeCoder) {
+      return new CurrentValueTransform(returnTypeCoder);
+    }
+
+    class RemoveMetadata extends PTransform<PBegin, PCollection<KV<K, ChangeEvent<V>>>> {
```
FelixGV, line 248: Can you make this inner class static?

integrations/venice-beam/src/main/java/com/linkedin/venice/beam/consumer/VeniceChangelogConsumerIO.java
```diff
@@ -0,0 +1,609 @@
...
+                  @Override
+                  public KV<K, ChangeEvent<V>> apply(PubSubMessage<K, ChangeEvent<V>, VeniceChangeCoordinate> message) {
+                    return KV.of(message.getKey(), message.getValue());
+                  }
+                }))
+            .setCoder(VeniceMessageCoder.of());
+      }
+    }
+
+    private class CurrentValueTransform extends PTransform<PBegin, PCollection<KV<K, GenericRecord>>> {
```
FelixGV, line 266: Can you make this inner class static?

FelixGV (commented):
The latest static analysis run also failed... the report can be downloaded here:

> Artifact download URL: https://github.com/linkedin/venice/actions/runs/11947409055/artifacts/2221106015

(IDK why it succeeds locally but fails on the CI. That is weird...)

In short, it is all about `Serializable` issues... but it does not look that important, so I think we could add an exclusion for `SE_INNER_CLASS` applying to `com.linkedin.venice.beam.consumer.VeniceChangelogConsumerIO` and `com.linkedin.venice.beam.consumer.PubSubMessageCoder` in the `gradle/spotbugs/exclude.xml` file.

FelixGV (approved):
Thanks a lot Sunny! It's great to receive your first contribution to the Venice project!

PR #1301, opened 2024-11-13
[server] Avoid race condition between pick consumer call and actual consumer subscribe API

sixpluszero:
## [server] Avoid race condition between pick consumer call and actual consumer subscribe API
Current implementation does not cover the race condition case between pick consumer API call and actual consumer subscribe call. So the assignment size can be inaccurate during calculation. To address this issue, this PR adds maps to keep track of the basic load and store level load for each consumer, and update it during pickConsumerForPartition() API

## How was this PR tested?

Adjust unit test to more coverage.


gaojieliu (approved):
This code change is fine overall, but it is not fully accurate as unsubscription can happen   concurrently with `pickConsumerForPartition` call.
Since `pickConsumerForPartition` function is tracking these maps syncly, it can guarantee at least the even distribution for the same resource, which is potentially good enough.

If ^ is a correct understanding, can you convert it into javadoc to avoid confusion.

PR #1302, opened 2024-11-13
[controller] Change task list member variable back to local variable in AdminConsumptionTask

huangminchn:
## Summary
The current implementation is error prone for multi-thread
 access to function "executeMessagesAndCollectResults",
 unless we add a synchronized keyword to "executeMessagesAndCollectResults".

## How was this PR tested?
Github CI


majisourav99 (approved):
Thanks Min!

PR #1305, opened 2024-11-14
[server][metrics] Fix sawtooth behavior on follower lag monitoring

ZacAttack:
## [server][metrics] Fix sawtooth behavior on follower lag monitoring

This tweaks the behavior of the follower metric to retain the highest value of the last received timestamp.  Prior to this, it was strictly reporting the last received timestamp.  The result of this was that when one data center was lagged, the follower metric would sawtooth between healthy and unhealthy signals.  This tweak makes it so the most recent timestamp will be retained and reported for the metric for all received data.

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/ingestion/heartbeat/HeartbeatMonitoringService.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/ingestion/heartbeat/HeartbeatMonitoringService.java
@@ -12,8 +12,6 @@ import io.tehuti.metrics.MetricsRepository;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
-import org.apache.commons.lang3.tuple.MutablePair;
-import org.apache.commons.lang3.tuple.Pair;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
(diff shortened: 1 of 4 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/ingestion/heartbeat/HeartbeatTimeStampEntry.java
```diff
@@ -0,0 +1,24 @@
+package com.linkedin.davinci.stats.ingestion.heartbeat;
+
+public class HeartbeatTimeStampEntry {
```
sixpluszero, line 3: Nice, we get rid of Pair

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/ingestion/heartbeat/HeartbeatMonitoringService.java
```diff
@@ -274,13 +273,24 @@ private void recordHeartbeat(
...
       heartbeatTimestamps.computeIfPresent(store, (storeKey, perVersionMap) -> {
         perVersionMap.computeIfPresent(version, (versionKey, perPartitionMap) -> {
           perPartitionMap.computeIfPresent(partition, (partitionKey, perRegionMap) -> {
-            perRegionMap.put(region, new MutablePair<>(timestamp, isReadyToServe));
+            // If we are retaining only the highest timestamp for a given heartbeat, if the current held heartbeat
+            // is of a higher value AND was an entry was consumed (not a place holder value by the process) then
+            // we will No-Op in favor of retaining that higher timestamp. This behavior is specific to follower
+            // nodes because the intent of this metric is to only show the lag of the follower relative to the leader
+            if (retainHighestTimeStamp && perRegionMap.get(region) != null
+                && perRegionMap.get(region).timestamp > timestamp && !perRegionMap.get(region).consumedFromUpstream) {
```
sixpluszero, line 288: !perRegionMap.get(region).consumedFromUpstream
Should this be true?

sixpluszero (commented):
Thank you for putting up the PR and get rid of the pair.
I left one question about the skip logic.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/ingestion/heartbeat/HeartbeatMonitoringService.java
ZacAttack: oh that's right, it should be true, good catch.

sixpluszero (approved):
Thanks for the fix!

PR #1306, opened 2024-11-15
[router] remove store level metrics for non-streaming multiget

xunyin8:
## [router] remove store level metrics for non-streaming multiget

The store level metrics in AggRouterHttpRequestStats are registered lazily. Add a flag in AggRouterHttpRequestStats to disable store level metric emission for non-streaming multiget request type. This is cleaner and the less intrusive than adding checks and handling elsewhere because AggRouterHttpRequestStats and the triggers to corresponding record functions are shared across different request types (SINGLE_GET, MULTI_GET, MULTI_GET_STREAMING, COMPUTE, etc...).

We will keep the total stats for non-streaming multiget to give us visibility in case any legacy clients start to make non-streaming multiget requests.

Fixed a bug where recordNoAvailableReplicaAbortedRetryRequest was calling the wrong method for the store metric recording.

## How was this PR tested?
Existing and new tests

### DIFF
--- a/services/venice-router/src/main/java/com/linkedin/venice/router/stats/AggRouterHttpRequestStats.java
+++ b/services/venice-router/src/main/java/com/linkedin/venice/router/stats/AggRouterHttpRequestStats.java
@@ -9,11 +9,13 @@ import com.linkedin.venice.utils.concurrent.VeniceConcurrentHashMap;
 import io.netty.handler.codec.http.HttpResponseStatus;
 import io.tehuti.metrics.MetricsRepository;
 import java.util.Map;
+import java.util.function.Consumer;
 import java.util.function.Function;
 
 
 public class AggRouterHttpRequestStats extends AbstractVeniceAggStoreStats<RouterHttpRequestStats> {
   private final Map<String, ScatterGatherStats> scatterGatherStatsMap = new VeniceConcurrentHashMap<>();
+  private final boolean isStoreStatsEnabled;
 
   public AggRouterHttpRequestStats(
       String clusterName,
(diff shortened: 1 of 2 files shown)


services/venice-router/src/main/java/com/linkedin/venice/router/stats/AggRouterHttpRequestStats.java
```diff
@@ -29,14 +32,18 @@ public AggRouterHttpRequestStats(
       ReadOnlyStoreRepository metadataRepository,
       boolean isUnregisterMetricForDeletedStoreEnabled) {
     super(metricsRepository, metadataRepository, isUnregisterMetricForDeletedStoreEnabled);
+    // Disable store level non-streaming multi get stats reporting because it's no longer used in clients. We still
+    // report to the total stats for visibility of potential old clients.
+    isStoreStatsEnabled = !RequestType.MULTI_GET.equals(requestType);
```
m-nagarajan, line 37: Should we just remove MULTI_GET? Do we have any use cases for this?

services/venice-router/src/main/java/com/linkedin/venice/router/stats/AggRouterHttpRequestStats.java
```diff
@@ -46,35 +53,45 @@ public AggRouterHttpRequestStats(
   }
 
   public ScatterGatherStats getScatterGatherStatsForStore(String storeName) {
-    return scatterGatherStatsMap.computeIfAbsent(storeName, k -> new ScatterGatherStats());
+    if (isStoreStatsEnabled) {
+      return scatterGatherStatsMap.computeIfAbsent(storeName, k -> new ScatterGatherStats());
+    } else {
+      return totalScatterGatherStats;
```
m-nagarajan, line 59: Will this result in dual recording of the total stats?
if we are checking `isStoreStatsEnabled` and not recording the store stats as seen in `recordStoreStats()`, should we clean this to not return anything?

services/venice-router/src/main/java/com/linkedin/venice/router/stats/AggRouterHttpRequestStats.java
xunyin8: Let's remove it when we actually remove the code to support it too.

services/venice-router/src/main/java/com/linkedin/venice/router/stats/AggRouterHttpRequestStats.java
xunyin8: Good catch, removed and rebased.

services/venice-router/src/main/java/com/linkedin/venice/router/stats/AggRouterHttpRequestStats.java
m-nagarajan: LGTM

m-nagarajan (dismissed):
Thanks @xunyin8

m-nagarajan (approved):

PR #1308, opened 2024-11-15
[server] Fix resubscription race condition caused by consumer object level lock.

haoxu07:

This PR addresses a race condition caused by the `wait()` and `notifyAll()` mechanism between the `StoreIngestionTask` thread's call to `SharedKafkaConsumer#unSubscribeAction` and the `ConsumptionTask` thread's call to `SharedKafkaConsumer`#poll. This mechanism ensures that the next poll only occurs after the unsubscription has completed. The wait() inside `SharedKafkaConsumer#unSubscribeAction` releases the object-level lock on the `SharedKafkaConsumer`.

However, the changes introduced in [PR #1263](https://github.com/linkedin/venice/pull/1263) added a consumer intrinsic lock in `KafkaConsumerService` to protect unsubscription operations (inside `unSubscribe`, `unSubscribeAll`, and `batchUnsubscribe`) and data receiver settings (inside `assignConsumerFor`). These changes required the entire lifecycle of unsubscription to be synchronized under the lock, as multiple `StoreIngestionTask` threads could perform re-subscription (unsubscribe followed by resubscribe) concurrently. The `wait()` in `SharedKafkaConsumer#unSubscribeAction` violated this locking assumption, leading to race conditions.

Since the intrinsic `SharedKafkaConsumer` lock cannot be used for both unsubscription/data receiver settings and for coordination between unsubscribe and poll, the solution is to introduce a dedicated lock for each consumer. This ensures proper synchronization while resolving the race condition.

## How was this PR tested?

CI

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaConsumerService.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaConsumerService.java
@@ -25,6 +25,7 @@ import com.linkedin.venice.utils.Time;
 import com.linkedin.venice.utils.Utils;
 import com.linkedin.venice.utils.VeniceProperties;
 import com.linkedin.venice.utils.concurrent.VeniceConcurrentHashMap;
+import com.linkedin.venice.utils.locks.AutoCloseableLock;
 import io.tehuti.metrics.MetricsRepository;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -35,6 +36,7 @@ import java.util.Set;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.IntConsumer;
 import java.util.function.LongSupplier;
 import java.util.function.Supplier;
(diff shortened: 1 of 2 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaConsumerService.java
```diff
@@ -76,6 +78,9 @@ public abstract class KafkaConsumerService extends AbstractKafkaConsumerService
   protected final Map<PubSubTopic, Map<PubSubTopicPartition, SharedKafkaConsumer>> versionTopicToTopicPartitionToConsumer =
       new VeniceConcurrentHashMap<>();
 
+  // This per consumer lock is for protecting the partition unsubscription and data receiver setting operations.
```
lluwm, line 81: Nit point, maybe one sentence to briefly remind future readers that using the `SharedKafkaConsumer's` intrinsic lock could cause the race and redirect them to the PR's description for details.

lluwm (dismissed):
Thanks for catching this race and LGTM! Feel free to ignore my nit comments.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaConsumerService.java
haoxu07: Add one sentence as suggested

lluwm (approved):

PR #1309, opened 2024-11-15
[controller] Log request body for certain requests

nisargthakkar:
## Log request body for certain Controller requests

Certain Controller APIs carry body params that are useful to be logged for auditing. Today, we do not log any request body as there might be sensitive content in there. Added support to log the request body.

### DIFF
--- a/internal/venice-common/src/main/java/com/linkedin/venice/controllerapi/ControllerRoute.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/controllerapi/ControllerRoute.java
@@ -86,6 +86,7 @@ import static com.linkedin.venice.controllerapi.ControllerApiConstants.TOPIC_COM
 import static com.linkedin.venice.controllerapi.ControllerApiConstants.TO_BE_STOPPED_INSTANCES;
 import static com.linkedin.venice.controllerapi.ControllerApiConstants.UPSTREAM_OFFSET;
 import static com.linkedin.venice.controllerapi.ControllerApiConstants.VALUE_SCHEMA;
+import static com.linkedin.venice.controllerapi.ControllerApiConstants.VALUE_SCHEMA_IDS;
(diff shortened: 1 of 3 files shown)


internal/venice-common/src/main/java/com/linkedin/venice/controllerapi/ControllerRoute.java
```diff
@@ -120,45 +124,83 @@ public enum ControllerRoute {
...
   ClUSTER_HEALTH_INSTANCES(
-      "/cluster_health_instances", HttpMethod.GET, Collections.emptyList(), ENABLE_DISABLED_REPLICAS
+      "/cluster_health_instances", HttpMethod.GET, Collections.emptyList(),
+      Collections.singletonList(ENABLE_DISABLED_REPLICAS)
   ), LIST_REPLICAS("/list_replicas", HttpMethod.GET, Arrays.asList(NAME, VERSION)),
   NODE_REPLICAS("/storage_node_replicas", HttpMethod.GET, Collections.singletonList(STORAGE_NODE_ID)),
   NODE_REMOVABLE(
-      "/node_removable", HttpMethod.GET, Collections.singletonList(STORAGE_NODE_ID), INSTANCE_VIEW,
-      TO_BE_STOPPED_INSTANCES
+      "/node_removable", HttpMethod.GET, Collections.singletonList(STORAGE_NODE_ID),
```
majisourav99, line 202: This can also not have sensitive body ?

nisargthakkar: This endpoint doesn't have have any body, right?

nisargthakkar: I realize that `bodyHasSensitiveContent` might not be the right variable to decide if body should be logged or not. I wanted to not make a very specific config for audit logging. But I'm open to suggestions if you have better ones.

PR #1310, opened 2024-11-15
[server][dvc] Drop partitions asynchronously

kvargha:

When a storage node is transitioning from `LEADER -> STANDBY -> OFFLINE -> DROPPED`, a race condition can occur. Specifically, the `DROPPED` state transition may be executed synchronously before the other state transitions are processed (`LEADER -> STANDBY -> OFFLINE` are executed asynchronously). This results in the store partition being deleted prematurely. Consequently, when the `LEADER -> STANDBY` message is eventually processed, it triggers a `PersistenceFailureException` since the storage partition no longer exists.

The solution for this is to drop the store partition asynchronously by adding a `DROP_PARTITION` message to the `consumerActionsQueue` if the ingestion task is still running. In the case that it's not running (this can happen if it was killed), the store partition will be dropped synchronously. Additionally, if the ingestion task is killed after adding a `DROP_PARTITION` message, we will process this message in `StoreIngestionTask::internalClose`.

## How was this PR tested?


clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaStoreIngestionService.java
```diff
@@ -920,6 +925,37 @@ public void stopConsumptionAndWait(
...
+  public void dropStoragePartitionGracefully(VeniceStoreVersionConfig veniceStore, int partitionId) {
+    final String topic = veniceStore.getStoreVersionName();
+
+    if (isPartitionConsuming(topic, partitionId)) {
+      throw new VeniceException("Tried to drop storage partition that is still consuming");
+    }
+
+    try (AutoCloseableLock ignore = topicLockManager.getLockForResource(topic)) {
+      StoreIngestionTask ingestionTask = topicNameToIngestionTaskMap.get(topic);
+      if (ingestionTask != null && ingestionTask.isRunning()) {
```
lluwm, line 943: I am thinking of a race condition that after we add `DROP_PARTITION` actions to the queue, then SIT terminates due to some exceptions (as we see several cases today) before executing all the remaining actions from the queue and it could probably cause some partition leaks. If this race is possible, we probably need to add some logic in the SIT to make sure that all `DROP_PARTITION` actions have to be executed before it can terminate itself.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
```diff
@@ -252,6 +252,7 @@ public DaVinciBackend(
           .map(objectCacheConfig -> new ObjectCacheBackend(clientConfig, objectCacheConfig, schemaRepository));
 
       ingestionService = new KafkaStoreIngestionService(
+          storageService,
           storageService.getStorageEngineRepository(),
```
eldernewborn, line 256: do we need both ? if we pass the storage service, shouldnt the repo be extracted internally ?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/isolated/IsolatedIngestionServer.java
```diff
@@ -695,6 +695,7 @@ private void initializeIsolatedIngestionServer() {
 
     // Create KafkaStoreIngestionService
     storeIngestionService = new KafkaStoreIngestionService(
+        storageService,
         storageService.getStorageEngineRepository(),
```
eldernewborn, line 699: ditto

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaStoreIngestionService.java
```diff
@@ -920,6 +925,37 @@ public void stopConsumptionAndWait(
...
+   * Drops the corresponding Venice Partition gracefully.
+   * This should only be called after {@link #stopConsumptionAndWait} has been called
+   * @param veniceStore Venice Store for the partition.
+   * @param partitionId Venice partition's id.
+   */
+  public void dropStoragePartitionGracefully(VeniceStoreVersionConfig veniceStore, int partitionId) {
+    final String topic = veniceStore.getStoreVersionName();
+
+    if (isPartitionConsuming(topic, partitionId)) {
+      throw new VeniceException("Tried to drop storage partition that is still consuming");
```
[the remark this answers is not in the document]
kvargha: `STANDBY->OFFLINE` issues an `UNSUBSCRIBE` message, and so will `stopConsumptionAndWait`.

By the time `SIT` processes the `DROP_PARTITION` message, it should have been already unsubscribed. 

I think it's safe to remove this check. What do you think?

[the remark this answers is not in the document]
lluwm: I agree that it should be safe to remove the check completely here and we already have a warning message in `stopConsumptionAndWait` when it times out.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -1744,12 +1790,15 @@ private void internalClose(boolean doFlush) {
         if (opType == ConsumerActionType.RESET_OFFSET) {
           LOGGER.info("Cleanup Reset OffSet. Replica: {}", replica);
           storageMetadataService.clearOffset(topic, partition);
+        } else if (opType == DROP_PARTITION) {
+          PubSubTopicPartition topicPartition = message.getTopicPartition();
```
lluwm, line 1794: Nit, I don't really know if we need to add a log line here (follow patterns for `RESET_OFFSET`) to indicate the deletion actually happened in the `internalClose`, given that there are already logs in `dropPartitionSynchronously`. Just a heads up, I guess.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/server/VeniceServerTest.java
```diff
@@ -338,4 +340,73 @@ public void testStartServerWithSystemSchemaInitialization() {
       });
     }
   }
+
+  @Test
```
lluwm, line 344: Let's add a timeout for the integration test

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/server/VeniceServerTest.java
```diff
@@ -338,4 +340,73 @@ public void testStartServerWithSystemSchemaInitialization() {
...
+      Assert.assertEquals(repository.getAllLocalStorageEngines().size(), 1);
+
+      TestUtils.waitForNonDeterministicAssertion(10, TimeUnit.SECONDS, () -> {
+        // Partitions should have been dropped asynchronously due to rebalancing
+        Assert.assertTrue(storageService.getStorageEngine(storeVersionName).getPartitionIds().size() < 3);
+      });
+    }
+  }
+
+  @Test
```
lluwm, line 380: Ditto.

lluwm (approved):
Thanks for the fix @kvargha ! The change looks good to me.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
kvargha: I think it makes sense to maintain consistency.

lluwm (approved):

PR #1311, opened 2024-11-15
[server][vpj][controller] Add support to ingest from separate RT topic in A/A SIT when feature is enabled

sixpluszero:
## [server][vpj][controller] Add support to ingest from separate RT topic in A/A SIT when feature is enabled
This PR continue the work from #1262 and #1172 to add server side ingestion functionality for separate realtime topic.

1. The scope of this PR only covers A/A SIT, and does not support in L/F SIT. This can be left as future work if we identify it is  useful in other use cases other than incremental push.
2. This PR adjusts an assumption in previous PR: We will only write incremental push to separate RT topic when specified. This is to make sure we will have a smooth transition period, as there can be mixed separated RT config specification for current / backup version, and VPJ directly writing incremental push to the separate RT topic will result in data loss in one version. For this, we add a new config in requestTopic API in VPJ and user can override it in VPJ config. 
4. For subscription, we will require adding new config in server kafka cluster mapping. This PR added some utils methods to check / resolve special URL/region naming and topic resolution to make sure we have correct subscription and offset tracking. For TopicSwitch I decided not to touch it at all, and we will only complement the separate RT offset calculation when needed. 
5. For consumer assignment, this PR assigned separate RT into 2nd level of priority pool. For AAWC leader strategy, it is assigned to the default pool; For current version prioritization strategy, this PR adds a current_sep_rt pool, it is assigned to that pool; For non_current version, it is on non AAWC leader pool. Personally I think it is fine as beginning, we might want to adjust to allocate separate pool if there is performance issue.
6. For metrics: There is no HB metrics as leader does not emit HB into it (should we?? maybe we should as we might still need stuck pipeline detection, but without alert). For lag/rate metrics, AASIT stats implementation natively offer the metrics, and we will mute separate RT region metric registration if the SIT does not enable separate topic.
8. Did minor code clean up in AASIT as I think there are some code duplication, but tried to keep it as minimal as the PR itself is large.
9. Adjusted a watermark filter implementation in RmdUtils used in CDC (Although it is somewhat weird and open for comment to make sure I understand the original intention)
## How was this PR tested?

Added new integration test; Added some unit tests for test coverage; Adjusted some old integration test / unit test behaviors.

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
@@ -1555,7 +1555,7 @@ public abstract class StoreIngestionTask implements Runnable, Closeable {
           forceUnSubscribedCount++;
         }
       }
-      if (topicPartitionsToUnsubscribe.size() != 0) {
+      if (!topicPartitionsToUnsubscribe.isEmpty()) {
         consumerBatchUnsubscribe(topicPartitionsToUnsubscribe);
       }
     }
@@ -3526,11 +3526,6 @@ public abstract class StoreIngestionTask implements Runnable, Closeable {
   void consumerSubscribe(PubSubTopicPartition pubSubTopicPartition, long startOffset, String kafkaURL) {
     // TODO: pass special format of kafka url as input here when subscribe to the separate incremental push topic
     String resolvedKafkaURL = kafkaClusterUrlResolver != null ? kafkaClusterUrlResolver.apply(kafkaURL) : kafkaURL;
-    LOGGER.info(
-        "DEBUGGING resolved kafka URL: {}, original URL: {}, kafka props: {}",
-        resolvedKafkaURL,
-        kafkaURL,
-        kafkaProps);
     final boolean consumeRemotely = !Objects.equals(resolvedKafkaURL, localKafkaServer);
     if (!Objects.equals(resolvedKafkaURL, kafkaURL) && !isSeparatedRealtimeTopicEnabled
         && pubSubTopicPartition.getPubSubTopic().isRealTime()) {
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
@@ -3524,10 +3524,9 @@ public abstract class StoreIngestionTask implements Runnable, Closeable {
   }
 
   void consumerSubscribe(PubSubTopicPartition pubSubTopicPartition, long startOffset, String kafkaURL) {
-    // TODO: pass special format of kafka url as input here when subscribe to the separate incremental push topic
     String resolvedKafkaURL = kafkaClusterUrlResolver != null ? kafkaClusterUrlResolver.apply(kafkaURL) : kafkaURL;
     final boolean consumeRemotely = !Objects.equals(resolvedKafkaURL, localKafkaServer);
-    if (!Objects.equals(resolvedKafkaURL, kafkaURL) && !isSeparatedRealtimeTopicEnabled
+    if (!Objects.equals(resolvedKafkaURL, kafkaURL) && !isSeparatedRealtimeTopicEnabled()
         && pubSubTopicPartition.getPubSubTopic().isRealTime()) {
       return;
     }
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
@@ -4497,17 +4497,16 @@ public abstract class StoreIngestionTask implements Runnable, Closeable {
       PartitionConsumptionState partitionConsumptionState,
       String kafkaURL) {
     PubSubTopicPartition pubSubTopicPartition = partitionConsumptionState.getSourceTopicPartition(topic);
-    if (kafkaClusterUrlResolver != null && !kafkaURL.equals(kafkaClusterUrlResolver.apply(kafkaURL))
-        && topic.isRealTime()) {
+    if (topic.isRealTime() && getKafkaClusterUrlResolver() != null
+        && !kafkaURL.equals(getKafkaClusterUrlResolver().apply(kafkaURL))) {
       PubSubTopic separateRealTimeTopic =
-          pubSubTopicRepository.getTopic(Version.composeSeparateRealTimeTopic(storeName));
+          getPubSubTopicRepository().getTopic(Version.composeSeparateRealTimeTopic(topic.getStoreName()));
       pubSubTopicPartition =
           new PubSubTopicPartitionImpl(separateRealTimeTopic, partitionConsumptionState.getPartition());
     }
     LOGGER.info("Resolved topic-partition: {} from kafkaURL: {}", pubSubTopicPartition, kafkaURL);
     return pubSubTopicPartition;
   }
-
 
   PubSubTopicRepository getPubSubTopicRepository() {
     return pubSubTopicRepository;
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
@@ -287,6 +287,8 @@ public abstract class StoreIngestionTask implements Runnable, Closeable {
 
   protected final boolean isWriteComputationEnabled;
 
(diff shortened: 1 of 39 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -4381,6 +4411,34 @@ public boolean hasAllPartitionReportedCompleted() {
...
+      PubSubTopic separateRealTimeTopic =
+          getPubSubTopicRepository().getTopic(Version.composeSeparateRealTimeTopic(topic.getStoreName()));
+      pubSubTopicPartition =
+          new PubSubTopicPartitionImpl(separateRealTimeTopic, partitionConsumptionState.getPartition());
+    }
+    LOGGER.info("Resolved topic-partition: {} from kafkaURL: {}", pubSubTopicPartition, kafkaURL);
+    return pubSubTopicPartition;
+  }
+
+  PubSubTopicRepository getPubSubTopicRepository() {
```
haoxu07, line 4434: Seems this test-only PubSubTopicRepository is not necessary, PubSubTopicImpl could handle equal check by comparing String name

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/IngestionStatsReporter.java
```diff
@@ -344,6 +345,11 @@ protected void registerConditionalStats() {
           .getServerConfig()
           .getKafkaClusterIdToAliasMap()
           .int2ObjectEntrySet()) {
+        // We will only register sensor for SIT with separate RT topic enabled to avoid unnecessary metrics.
+        if (getStats().getIngestionTask().isSeparatedRealtimeTopicEnabled()
+            && Utils.isSeparateTopicRegion(entry.getValue())) {
+          continue;
```
haoxu07, line 351: Seems here it will skip and not regiester topic when separate topic is enabled.

haoxu07 (commented):
Thanks for preparing this! I am still in the middle of reading the SIT changes, just left a few comments.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
sixpluszero: Hmm, actually this is used in main code as well. PubSubTopicImpl constructor is package private by design so it is not accessible from SIT class.

internal/venice-client-common/src/main/java/com/linkedin/venice/schema/rmd/RmdUtils.java
```diff
@@ -87,13 +87,13 @@ public static long sumOffsetVector(Object offsetVector) {
    * @return                True if the advancedOffset vector has grown beyond the baseOffset
    */
   static public boolean hasOffsetAdvanced(@NotNull List<Long> baseOffset, @NotNull List<Long> advancedOffset) {
-    if (baseOffset.size() > advancedOffset.size()) {
```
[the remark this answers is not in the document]
sixpluszero: This is only used in CDC code. I think base offset is the offset map, which contains all the regions, including separate RT region (they will be 0 if we are not writing to it). advancedOffset is from RMD vector, which will only be updated when we consumed a message from specific region.
In this case, we will have 3 advanced offset vs 6 base offset, even though 3 of them are all 0, so the previous short-circuit logic no longer applies.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -3489,9 +3506,26 @@ public void consumerBatchUnsubscribe(Set<PubSubTopicPartition> topicPartitionSet
...
 
-  public void consumerSubscribe(PubSubTopicPartition pubSubTopicPartition, long startOffset, String kafkaURL) {
-    // TODO: pass special format of kafka url as input here when subscribe to the separate incremental push topic
+  /**
+   * This method will try to resolve actual topic-partition from input Kafka URL and subscribe to the resolved
+   * topic-partition.
+   */
+  public void consumerSubscribe(
+      PubSubTopic pubSubTopic,
+      PartitionConsumptionState partitionConsumptionState,
```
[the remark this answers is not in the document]
sixpluszero: I think either way should be ok. For now, all the places calling this method has PCS available, so I prefer for now let's keep using PCS in the method signature, otherwise it will result in a bunch of changes in files, which can potentially be done in a separate PR.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -3489,9 +3506,26 @@ public void consumerBatchUnsubscribe(Set<PubSubTopicPartition> topicPartitionSet
...
+
+  void consumerSubscribe(PubSubTopicPartition pubSubTopicPartition, long startOffset, String kafkaURL) {
     String resolvedKafkaURL = kafkaClusterUrlResolver != null ? kafkaClusterUrlResolver.apply(kafkaURL) : kafkaURL;
+    if (!Objects.equals(resolvedKafkaURL, kafkaURL) && !isSeparatedRealtimeTopicEnabled()
+        && pubSubTopicPartition.getPubSubTopic().isRealTime()) {
+      return;
+    }
     final boolean consumeRemotely = !Objects.equals(resolvedKafkaURL, localKafkaServer);
     // TODO: Move remote KafkaConsumerService creating operations into the aggKafkaConsumerService.
     aggKafkaConsumerService
```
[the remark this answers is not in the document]
sixpluszero: This is a good catch. I think we should not expose this to AggKafkaConsumerService, so I made the logic tweak inside StorageUtilizationManager. It seems like this is the only place where these pause/resume is used. 
For reset, it seems like it will always on VT??? (This is what I got from code reading, I did not pay much attention to this API before)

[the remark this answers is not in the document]
haoxu07: I was thinking if we could limit the scope of resolve topics for hybrid leader inside PartitionConsumptionState to let it figure out if we need to return 1 or 2 topics, then SIT consumer dealing functions (e.g consumerSubscribe) only handle the topics PCS returned.  The reason is there are a lot of interaction between SIT and AggkafkaConsumerService. AggKafkaConsumerService could only deal with one topic partition a time, that means SIT function or its related class should take a lot of consideration about ingestion from 1 or 2 topics.

internal/venice-client-common/src/main/java/com/linkedin/venice/schema/rmd/RmdUtils.java
```diff
@@ -87,13 +87,13 @@ public static long sumOffsetVector(Object offsetVector) {
    * @return                True if the advancedOffset vector has grown beyond the baseOffset
    */
   static public boolean hasOffsetAdvanced(@NotNull List<Long> baseOffset, @NotNull List<Long> advancedOffset) {
```
haoxu07, line 89: Not related to this code line here, but seems we did not take care of RMD offset vector for separate topic. As the offset vector's element is for each region, then we probably messed together normal real-time upstream offset and separate real-time upstream offset. This will cause trouble for Deterministic Conflict Resolution Validation, as it reads differet region's VT to detect violation by comparing real-time upstream offset. We may need a way to resolve this.

sixpluszero: Hmm, I am actually a bit confused by this. As I check the code in updateReplicationCheckpointVector method, which is used when doing all the AA operation, it will update the offset vector by kafkaClusterId, this ID is coming all the way from the StorePartitionDataReceiver. 
In the AggKafkaConsumerService#subscribeConsumerFor, when we are wiring the data receiver, we use the original KafkaURL and ID, which is unique for sep topic.

haoxu07: For exmaple, if we have 3 regions for AA ingestion, without separate real-time topic, the offset vector will have 3 elements at most. If separate real-time topic is enabled, we will have 3 * 2 = 6 elements for offset vector at most?

sixpluszero: Yes, I updated the integration test to verify this. Please check TestSeparateRealtimeTopicIngestion#testIncrementalPushPartialUpdate. If I set source.grid.fabric to dc-1, it will have 4 value. If I don't set it (default to dc-0) it will be 3 value.

haoxu07 (dismissed):
Thanks a lot for preparing this! This feature is complicated, as we have config to turn on and off safely, we can merge current change and evolve with it.

haoxu07 (approved):

sixpluszero:
Thanks @haoxu07 for detailed reviews!

PR #1312, opened 2024-11-15
[server] SN read quota versioned stats not initialized after restart

xunyin8:
## [server] SN read quota versioned stats not initialized after restart

The currentVersion and backupVersion of ServerReadQuotaUsageStats are not set after server restart because handleStoreChanged is invoked for all stores when the store repo undergoing refresh before we initialize and register store change listener in ReadQuotaEnforcementHandler (part of the ListenerService). As a result metrics that depend on current and backup versions will not show up properly until store is updated.

The fix is to during initialization of ReadQuotaEnforcementHandler we will invoke handleStoreChanged for all stores after we register store change listener.

The bug is actually reproducible in existing integration test. However, it was not caught because the test was broken/misconfigured...

## How was this PR tested?


majisourav99 (approved):
LGTM! Thanks for the quick fix.

PR #1317, opened 2024-11-18
[server][dvc] Implement File Checksum Verification and Offset Lag Check for Bootstrap Method Selection

jingy-li:
## Summary

This PR fix:
1. File checksum verification: The PR adds a file checksum to the transferred header and performs a checksum comparison when the file is received.
2. Offset lag check: When blob transfer is enabled, it checks the current offset lag. If the lag is within the preset threshold (blobTransferDisabledOffsetLagThreshold), it skips all blob transfer related operations and directly attempts to bootstrap from Kafka to save time.
3. Race condition during snapshot recreation while being transferred, For example:
    - At 8:29 AM, host0 creates a snapshot, and the snapshot retention time is set to 30 minutes.
    - At 9:00 AM, host1 requests the snapshot from host0, and the snapshot is being transferred since it’s not stale.
    - At 9:01 AM, host2 requests the snapshot from host0, and at this point, the snapshot is considered stale. Host0 then attempts to regenerate a new snapshot and sends it to host2. However, at the same time, the old snapshot is still being sent to host1, which leads host1 to detect a file size change and ultimately abandon the blob transfer.
    - To resolve this, the blob snapshot manager only allows snapshot recreation if no concurrent users are using the snapshot. If there are concurrent users and old snapshot is staled, the later request is rejected, and host2 should request the snapshot from another host.
4. Add config: 
    - BLOB_TRANSFER_DISABLED_OFFSET_LAG_THRESHOLD to control the max allowed offset lag to not use blob transfer but use Kafka instead, even if the blob transfer is enable. 
    - BLOB_TRANSFER_MAX_TIMEOUT_IN_MIN to control the max allowed file transfer time, if exceed, the blob transfer will fail.

## How was this PR tested?
- Add unit test for file checksum.
- Run E2E testing in test store for checking the bootstrap lag.


clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/BlobTransferUtils.java
```diff
@@ -1,12 +1,17 @@
...
+import java.io.InputStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
 
 
 public class BlobTransferUtils {
   public static final String BLOB_TRANSFER_STATUS = "X-Blob-Transfer-Status";
   public static final String BLOB_TRANSFER_COMPLETED = "Completed";
   public static final String BLOB_TRANSFER_TYPE = "X-Blob-Transfer-Type";
+  public static final long BLOB_TRANSFER_DISABLED_OFFSET_LAG_THRESHOLD = 1000L;
```
sixpluszero, line 14: 1000 is too small as we usually can do a few k records ingestion within a minute for single topic-partition; Or at least let's make it configurable as mentioned below.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/DefaultIngestionBackend.java
```diff
@@ -115,6 +124,51 @@ CompletionStage<Void> bootstrapFromBlobs(Store store, int versionNumber, int par
...
+    String topicName = Version.composeKafkaTopic(store, versionNumber);
+    OffsetRecord offsetRecord = storageMetadataService.getLastOffset(topicName, partition);
+    LOGGER.info(
+        "Store {} partition {} topic offset is {}, offset lag is {}",
+        store,
+        partition,
+        offsetRecord.getLocalVersionTopicOffset(),
+        offsetRecord.getOffsetLag());
+
+    if (offsetRecord == null || (offsetRecord.getOffsetLag() == 0 && offsetRecord.getLocalVersionTopicOffset() == -1)) {
```
sixpluszero, line 156: If OffsetRecord is false, the above log will throw NPE then

clients/da-vinci-client/src/main/java/com/linkedin/davinci/ingestion/DefaultIngestionBackend.java
```diff
@@ -115,6 +124,51 @@ CompletionStage<Void> bootstrapFromBlobs(Store store, int versionNumber, int par
...
+      return true;
+    }
+
+    if (offsetRecord.getOffsetLag() < blobTransferDisabledOffsetLagThreshold) {
```
sixpluszero, line 160: I am a bit hesitant between HB lag vs offset lag, though right now I think maybe offset lag can also work for DVC as they all ingest from VT instead of RT. HB can be more accurate about how stale we are for a given replica, but we can't get the information unless we make the change to persist it in the OffsetRecord. 

Maybe it is good for now, and let's make this configurable so we can tune it easily in client app.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/BlobSnapshotManager.java
```diff
@@ -204,6 +214,15 @@ private boolean isSnapshotStale(String topicName, int partitionId) {
         - snapshotTimestamps.get(topicName).get(partitionId) > snapshotRetentionTimeInMillis;
   }
 
+  /**
+   * Increase the count of hosts using the snapshot
+   */
+  public void increaseConcurrentUserCount(String topicName, int partitionId) {
```
sixpluszero, line 220: nit: don't need to expose as _public_, we can use default modifier for unit test.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/BlobTransferUtils.java
```diff
@@ -24,4 +29,20 @@ public static boolean isMetadataMessage(HttpResponse msg) {
...
+  /**
+   * Generate MD5 checksum for a file
+   * @param filePath the path to the file
+   * @return a hex string
+   * @throws IOException if an I/O error occurs
+   */
+  public static String generateFileChecksum(Path filePath) throws IOException {
+    String md5Hex;
+    try (InputStream inputStream = Files.newInputStream(filePath)) {
+      md5Hex = org.apache.commons.codec.digest.DigestUtils.md5Hex(inputStream);
```
sixpluszero, line 42: nit: let's follow the import pattern to properly import this method.

sixpluszero (commented):
Thanks for the change! I left a few questions in the PR, but overall the change looks good.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/BlobSnapshotManager.java
```diff
@@ -150,6 +149,17 @@ public BlobTransferPartitionMetadata getTransferMetadata(BlobTransferPayload pay
    * @param blobTransferRequest the blob transfer request
    */
   private void recreateSnapshotAndMetadata(BlobTransferPayload blobTransferRequest) {
+    // Only create a new snapshot if there is no concurrent user.
+    // Otherwise, the snapshot is still in use and being transferred, and should not be recreated.
+    if (getConcurrentSnapshotUsers(blobTransferRequest.getTopicName(), blobTransferRequest.getPartition()) != 0L) {
```
sixpluszero, line 154: I am wondering if we also make stale snapshot threshold configurable? I think for now it is 30mins by default, but for certain heavy hybrid use case we might want to lower the threshold down on the app side.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/BlobTransferUtils.java
jingy-li: Changed the default lag to 100000L and make this value configurable.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/BlobSnapshotManager.java
jingy-li: snapshot retention time is configurable with default value 60, `snapshotRetentionTimeInMin = serverProperties.getInt(BLOB_TRANSFER_SNAPSHOT_RETENTION_TIME_IN_MIN, 60);`

sixpluszero (dismissed):

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/BlobSnapshotManager.java
```diff
@@ -117,12 +117,10 @@ public BlobTransferPartitionMetadata getTransferMetadata(BlobTransferPayload pay
 
     // check if the concurrent user count exceeds the limit
     checkIfConcurrentUserExceedsLimit(topicName, partitionId);
-    concurrentSnapshotUsers.computeIfAbsent(topicName, k -> new VeniceConcurrentHashMap<>())
-        .computeIfAbsent(partitionId, k -> new AtomicLong(0))
-        .incrementAndGet();
 
     boolean isHybrid = isStoreHybrid(payload.getStoreName());
```
gaojieliu, line 121: BTW,
Can this function be invoked concurrently?
If yes, we may need to make it `synchronized` to avoid the race condition.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/BlobSnapshotManager.java
```diff
@@ -150,6 +149,17 @@ public BlobTransferPartitionMetadata getTransferMetadata(BlobTransferPayload pay
    * @param blobTransferRequest the blob transfer request
    */
   private void recreateSnapshotAndMetadata(BlobTransferPayload blobTransferRequest) {
+    // Only create a new snapshot if there is no concurrent user.
```
gaojieliu, line 152: Nit:
`concurrent` -> `active`

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/BlobSnapshotManager.java
```diff
@@ -204,6 +214,15 @@ private boolean isSnapshotStale(String topicName, int partitionId) {
         - snapshotTimestamps.get(topicName).get(partitionId) > snapshotRetentionTimeInMillis;
   }
 
+  /**
+   * Increase the count of hosts using the snapshot
+   */
+  void increaseConcurrentUserCount(String topicName, int partitionId) {
```
gaojieliu, line 220: In function: `getConcurrentSnapshotUsers`:
```
 if (!concurrentSnapshotUsers.containsKey(topicName)
        || !concurrentSnapshotUsers.get(topicName).containsKey(partitionId)) {
      return 0;
    }
    return concurrentSnapshotUsers.get(topicName).get(partitionId).get();
```

We can change it to the following to avoid the double map lookups:
```
Map<Integer, AtomicLong> partitionUsageMap = concurrentSnapshotUsers.get(topicName);
if (partitionUsageMap == null) {
  return 0;
}
AtomicLong usage = partitionUsageMap.get(partitionId);
if (usage == null) {
  return 0;
}
return usage.get();
} 
```
In theory `int` should be more than enough to track the usage, right?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/server/P2PFileTransferServerHandler.java
```diff
@@ -201,10 +206,13 @@ private void sendFile(File file, ChannelHandlerContext ctx) throws IOException {
     ChannelFuture sendFileFuture;
     ChannelFuture lastContentFuture;
     long length = raf.length();
+    String fileChecksum = BlobTransferUtils.generateFileChecksum(file.toPath());
+
     HttpResponse response = new DefaultHttpResponse(HttpVersion.HTTP_1_1, HttpResponseStatus.OK);
     response.headers().set(HttpHeaderNames.CONTENT_LENGTH, length);
     response.headers().set(HttpHeaderNames.CONTENT_TYPE, APPLICATION_OCTET_STREAM);
     response.headers().set(HttpHeaderNames.CONTENT_DISPOSITION, "attachment; filename=\"" + file.getName() + "\"");
+    response.headers().set(HttpHeaderNames.CONTENT_MD5, "attachment; checksum=\"" + fileChecksum + "\"");
```
gaojieliu, line 215: Is it necessary to use this value prefix: `attachment; checksum=`? Can we just store the checksum as the full value?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/config/VeniceServerConfig.java
```diff
@@ -578,8 +582,11 @@ public VeniceServerConfig(VeniceProperties serverProperties, Map<String, Map<Str
         serverProperties.getInt(MAX_LEADER_FOLLOWER_STATE_TRANSITION_THREAD_NUMBER, 20);
 
     blobTransferManagerEnabled = serverProperties.getBoolean(BLOB_TRANSFER_MANAGER_ENABLED, false);
-    snapshotRetentionTimeInMin = serverProperties.getInt(BLOB_TRANSFER_SNAPSHOT_RETENTION_TIME_IN_MIN, 30);
+    snapshotRetentionTimeInMin = serverProperties.getInt(BLOB_TRANSFER_SNAPSHOT_RETENTION_TIME_IN_MIN, 60);
     maxConcurrentSnapshotUser = serverProperties.getInt(BLOB_TRANSFER_MAX_CONCURRENT_SNAPSHOT_USER, 5);
+    blobTransferMaxTimeoutInMin = serverProperties.getInt(BLOB_TRANSFER_MAX_TIMEOUT_IN_MIN, 60);
+    blobTransferDisabledOffsetLagThreshold =
```
gaojieliu, line 588: This one is hard to config properly as the different stores would configure different kinds of ready-to-serve threshold.
As the VT would contain the heart-beat message, so can we configure a time-based threshold instead?

sixpluszero: My concern is, today we don't persist the HB we have consumed up to in OffsetRecord, so if we don't subscribe and ingestion we don't know the HB lag we are at... We can persist that in OffsetRecord in theory, but that would be another change and protocol evolution involved..

gaojieliu: Can we consume VT backward  from the persisted checkpoint to figure out the latest HB timestamp?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/BlobSnapshotManager.java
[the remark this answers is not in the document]
jingy-li: Exactly, thanks for call out. change it to check the hybridStoreConfig present.

jingy-li: Yes, it could be called when there are several requests at the same time.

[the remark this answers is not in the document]
jingy-li: Exactly, thanks for call out, change it to check the hybridStoreConfig present or not.

gaojieliu (commented):
Looks good overall, and left one minor comment.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/BlobSnapshotManager.java
gaojieliu: So shall we make it thread-safe to avoid creating multiple snapshots?

jingy-li: Sure, added synchronized.

gaojieliu (approved):
Thanks!

PR #1318, opened 2024-11-19
[server] ConsumptionTask Refactor

KaiSernLim:
## Summary

ConsumerRefactorx2

`StoreIngestionTask.java` is getting too large and should be refactored.

There are three threads: `ConsumptionTask`, `StoreBufferDrainer`, and `StoreIngestionTask` that all share the methods in `StoreIngestionTask.java`, and it is difficult to know which part of the ingestion pipeline a `StoreIngestionTask` method is part of, just by looking at it.

1. Moved all `ConsumptionTask`-only functions and member variables out of `StoreIngestionTask` into a class such as `StorePartitionDataReceiver`.
2. I also plan on renaming `StorePartitionDataReceiver` to `ConsumedPartitionReceiver` to better indicate that it's part of the `ConsumptionTask` thread, but I'll leave it out of the diff to help with the review. Feel free to provide additional suggestions.

## How was this PR tested?

GHCI


clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -4381,6 +3979,101 @@ public boolean hasAllPartitionReportedCompleted() {
     return true;
   }
 
+  public StorageUtilizationManager getStorageUtilizationManager() {
+    return storageUtilizationManager;
+  }
+
+  public AbstractStoreBufferService getStoreBufferService() {
+    return storeBufferService;
+  }
```
FelixGV, line 3988: Do we need a getter for the whole `SBS`? Or would it be better to expose just a `putConsumerRecord` function? (as that seems like the only way we're using the gotten `SBS`, but maybe I missed something?)

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -3600,28 +2911,15 @@ protected void processVersionSwapMessage(
...
-  }
-
   protected Lazy<VeniceWriter<byte[], byte[], byte[]>> getVeniceWriter(
       PartitionConsumptionState partitionConsumptionState) {
     return partitionConsumptionState.getVeniceWriterLazyRef();
   }
 
+  protected void setRealTimeVeniceWriterRef(PartitionConsumptionState partitionConsumptionState) {
+    partitionConsumptionState.setVeniceWriterLazyRef(veniceWriterForRealTime);
+  }
```
FelixGV, line 2921: Do we need functions for this? Why not just call `partitionConsumptionState.getVeniceWriterLazyRef()` directly wherever we need to get the VW? And same for the setter?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StorePartitionDataReceiver.java
```diff
@@ -75,12 +123,1462 @@ public void write(List<PubSubMessage<KafkaKey, KafkaMessageEnvelope, Long>> cons
...
+      hostLevelIngestionStats.recordStorageQuotaUsed(storageUtilizationManager.getDiskQuotaUsage());
+    }
+  }
+
+  public Iterable<PubSubMessage<KafkaKey, KafkaMessageEnvelope, Long>> validateAndFilterOutDuplicateMessagesFromLeaderTopic(
+      Iterable<PubSubMessage<KafkaKey, KafkaMessageEnvelope, Long>> records,
+      String kafkaUrl,
+      PubSubTopicPartition topicPartition) {
+    PartitionConsumptionState partitionConsumptionState =
+        storeIngestionTask.getPartitionConsumptionState(topicPartition.getPartitionNumber());
```
FelixGV, line 322: I think it'd be great if the `PCS` became a class property of the `StorePartitionDataReceiver`, rather than be held in a map inside the `SIT`. I suspect this is easier said than done, and no need to do it just yet... A large refactoring like this is ideally kept free of functional changes, and that one may be too risky. Better do it as a smaller follow up change. But just wanted to share this crazy dream 😂

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -4381,6 +3979,101 @@ public boolean hasAllPartitionReportedCompleted() {
...
+  protected PubSubMessageProcessedResult processActiveActiveMessage(
+      PubSubMessage<KafkaKey, KafkaMessageEnvelope, Long> consumerRecord,
+      PartitionConsumptionState partitionConsumptionState,
+      int partition,
+      String kafkaUrl,
+      int kafkaClusterId,
+      long beforeProcessingRecordTimestampNs,
+      long beforeProcessingBatchRecordsTimestampMs) {
+    throw new VeniceException("processActiveActiveMessage() should only be called in active active mode");
+  }
```
FelixGV, line 4050: Do you view these functions as needing to be here because it's a work in progress in the migration of the code from `SIT`/`LFSIT` towards `StoreDataPartitionReceiver` (but eventually these functions will also be moved)? Or because this is the meaningful boundary where the handoff between the two classes should take place?

FelixGV (commented):
Overall, I'm supportive of reducing the size of the `SIT` classes and subclasses.

I left a few minor comments and questions. Please take a look and LMK what you think.

sixpluszero:
I have a general comment - from my understanding, SIT thread also interacting with the SIT functions and PCS states. Even though I don't think this part of change is scoped into current PR, but the diagram created above seems to not including SIT thread.  Do you mind adding a minor update to include that as well? This can be added to future diagram to talk about how the overall refactor works before vs after (btw I like all your diagrams, as they are clear to understand, thank you!)

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StorePartitionDataReceiver.java
```diff
@@ -75,12 +123,1462 @@ public void write(List<PubSubMessage<KafkaKey, KafkaMessageEnvelope, Long>> cons
...
+   *    availableSchemaIds by {@link StoreIngestionTask#waitUntilValueSchemaAvailable}.
+   * 2. Ingestion isolation is enabled, in which case ingestion happens on forked process instead of this main process.
+   */
+  private void deserializeValue(
+      int schemaId,
+      ByteBuffer value,
+      PubSubMessage<KafkaKey, KafkaMessageEnvelope, Long> record) throws IOException {
+    final SparseConcurrentList<Object> availableSchemaIds = storeIngestionTask.getAvailableSchemaIds();
+    final SparseConcurrentList<Object> deserializedSchemaIds = storeIngestionTask.getDeserializedSchemaIds();
+    if (schemaId < 0 || deserializedSchemaIds.get(schemaId) != null || availableSchemaIds.get(schemaId) == null) {
```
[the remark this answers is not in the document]
KaiSernLim: That's a fair point. I'll leave the `deserializeValue()` function in SIT, then.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
KaiSernLim: That's a good idea.

I wanted to change as little as possible when initially opening the PR, and then the improvements like this can come with review.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
KaiSernLim: I think it's because the writer on the PCS is not the real time writer, and it's being swapped out for `veniceWriterForRealTime`, which is a member of `LeaderFollowerStoreIngestionTask`.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StorePartitionDataReceiver.java
KaiSernLim: Ooo, that's an interesting idea. I've never thought of that before. I'd definitely be interested in attempting something like that out someday.

The way I saw it was that the `PCS` map is a "global" variable that is shared among the `ConsumptionTask`, `Drainer`, and `SIT` threads.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
KaiSernLim: Yes, they're WIP artifacts that won't be present with all the commits. I'm adding all the commits now into one PR, so it won't just be a Part 1 PR anymore.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
```diff
@@ -193,59 +153,9 @@ public static int getKeyLevelLockMaxPoolSizeBasedOnServerConfig(VeniceServerConf
         * serverConfig.getKafkaClusterIdToUrlMap().size() * multiplier + 1;
   }
 
-  @Override
-  protected DelegateConsumerRecordResult delegateConsumerRecord(
```
sixpluszero, line 197: The change is too big and is a bit confusing with this override method. I think you moved these all into StorePartitionDataReceiver class, but how do you do the override there then?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StorePartitionDataReceiver.java
```diff
@@ -75,12 +155,2310 @@ public void write(List<PubSubMessage<KafkaKey, KafkaMessageEnvelope, Long>> cons
...
+
+  public long getWriteTimestampFromKME(KafkaMessageEnvelope kme) {
+    if (kme.producerMetadata.logicalTimestamp >= 0) {
+      return kme.producerMetadata.logicalTimestamp;
+    } else {
+      return kme.producerMetadata.messageTimestamp;
+    }
+  }
+
+  private void validatePostOperationResultsAndRecord(
```
sixpluszero, line 2147: Nit: Is it possible that we move some util methods out of this class and make the file size smaller?

sixpluszero (commented):
I finally took a pass over this gigantic change. Overall it looks good, left a comment on whether we can further split the data receiver class out by moving util method out. But overall the structure is ok. And need a batch of rebase again...

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StorePartitionDataReceiver.java
sixpluszero: Or do you think it is wise to instead let the DataReceiver class sub-class by AA or not?

PR #1320, opened 2024-11-19
[controller] Log request and response body for aggregated stoppable check

nisargthakkar:
## Log request and response body for aggregated stoppable check

Currently, controller logs audit logs for API requests and responses. These audit logs however, do not log the request body.
In the aggregated stoppable checks API endpoint, the request and response are JSON objects. This PR logs these objects to get visibility into these.

## How was this PR tested?

Added unit tests

### DIFF
--- a/internal/venice-common/src/test/java/com/linkedin/venice/controllerapi/TestStoppableNodeStatusResponse.java
+++ b/internal/venice-common/src/test/java/com/linkedin/venice/controllerapi/TestStoppableNodeStatusResponse.java
@@ -1,41 +1,21 @@
 package com.linkedin.venice.controllerapi;
 
 import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertTrue;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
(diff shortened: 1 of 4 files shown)


internal/venice-common/src/test/java/com/linkedin/venice/controllerapi/TestStoppableNodeStatusResponse.java
```diff
@@ -0,0 +1,70 @@
...
+import com.linkedin.venice.utils.ObjectMapperFactory;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import org.testng.annotations.Test;
+
+
+public class TestStoppableNodeStatusResponse {
```
majisourav99, line 16: Do we need this test class ? The code path being tested is executed in other places

nisargthakkar: This is to test the JSON serde of `StoppableNodeStatusResponse` class. The coverage check fails without this as `StoppableNodeStatusResponse` is in `venice-common`.

Let me modify the class a bit to remove the inner `StoppableNodeStatusResponseForTest` class

majisourav99 (approved):
LGTM

PR #1325, opened 2024-11-19
[vpj][controller] Do not optimize backup version deletion for TTL repush

majisourav99:
## Do not optimize backup version deletion for TTL repush

Today Venice deletes the repush source version as it is exactly same as repushed version immediately after the push.  But for TTL repush since it differs from the source version this assumption is not correct. This PR prevents early deletion of  TTL repush version and treats it as regular push.
Also it creates a new pushID for ttl repush which would help tracking regular repush vs ttl repush.

## How was this PR tested?

GHCI


ZacAttack (dismissed):
This looks ok.  Was a bug opened for this?

sixpluszero (approved):
LGTM, thank you!

PR #1328, opened 2024-11-20
[common][client] Fix metadata schema fetcher cache to use map instead of list

ZacAttack:
## [common][client] Fix metadata schema fetcher cache to use map instead of list

It seems the current schema cache implementation is prone to bugs if there are gaps in the cache or if the number of elements goes to double digits.  This fixes it by using a map instead.

Resolves #XXX


gaojieliu (dismissed):
Thanks

gaojieliu (dismissed):

gaojieliu (approved):

PR #1329, opened 2024-11-20
[controller] Add push job duration in the push job details log

huangminchn:
## Summary
Provide extra data points to determine if a push job is killed by timeout setting prematurely.

## How was this PR tested?
Github CI

### DIFF
--- a/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
@@ -1255,10 +1255,9 @@ public class VeniceHelixAdmin implements Admin, StoreCleaner {
             pushJobStatusStats.recordBatchPushSuccessSensor();
           }
         }
-        // Append job duration in ms and seconds to log message
-        double jobDurationInMinutes = pushJobDetailsValue.getJobDurationInMs() / 60000.0
-        logMessage.append(", duration: ")
-            .append(String.format("%.1f mins", jobDurationInMinutes);
+        // Append job duration in minutes to log message
+        double jobDurationInMinutes = pushJobDetailsValue.getJobDurationInMs() / 60000.0;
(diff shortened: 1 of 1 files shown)


services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -1255,6 +1255,12 @@ static void emitPushJobStatusMetrics(
             pushJobStatusStats.recordBatchPushSuccessSensor();
           }
         }
+        // Append job duration in ms and seconds to log message
+        logMessage.append(", duration: ")
```
KaiSernLim, line 1259: @m-nagarajan I believe the regex for the push job details dashboard will need to be modified, since it ends in `$` and this extends the log message.

KaiSernLim (dismissed):

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -1255,6 +1255,12 @@ static void emitPushJobStatusMetrics(
...
+            .append(pushJobDetailsValue.getJobDurationInMs())
+            .append(" ms (")
+            .append(TimeUnit.MILLISECONDS.toSeconds(pushJobDetailsValue.getJobDurationInMs()))
+            .append(" s)");
```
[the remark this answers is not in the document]
huangminchn: Sounds good!

KaiSernLim (approved):

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
m-nagarajan: We can modify the regex in the dashboard once this is shipped. All good from me 👌

m-nagarajan: I want to change this to have per colo information anyways, so it's going to break regardless.

PR #1331, opened 2024-11-20
[server][dvc] Increase graceful shutdown time for Da Vinci to reduce timeout exception

sixpluszero:
## [server][dvc] Increase graceful shutdown time for Da Vinci to reduce timeout exception
For Da Vinci, when we close / retire a version, it will call shutdownIngestionTask API, which will try to unsubscribe and sync offset for graceful shutdown. Today the timeout is 10 seconds and it could timeout for topic with many partition assignment.
This will result in two kinds of exceptions:
1. When it is doing graceful shutdown and syncing offset, it may fail after the wait timeout as it will also remove storage engine
2. It may also failed to persist certain lingering data in drainer after timeout. 

This PR tries to do minimal changes to reduce the noises in application that runs DVC by increasing the total timeout for per SIT shutdown wait. 

## How was this PR tested?

N/A

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
@@ -135,6 +135,7 @@ import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.ConcurrentSkipListSet;
+import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.PriorityBlockingQueue;
@@ -342,6 +343,8 @@ public abstract class StoreIngestionTask implements Runnable, Closeable {
   protected final boolean batchReportIncPushStatusEnabled;
 
   protected final ExecutorService parallelProcessingThreadPool;
+
+  protected final CountDownLatch gracefulShutdownLatch = new CountDownLatch(1);
 
   public StoreIngestionTask(
       StoreIngestionTaskFactory.Builder builder,
@@ -1605,6 +1608,7 @@ public abstract class StoreIngestionTask implements Runnable, Closeable {
          * to avoid infinite wait in case there is some regression.
          */
         CompletableFuture.allOf(shutdownFutures.toArray(new CompletableFuture[0])).get(60, SECONDS);
+        getGracefulShutdownLatch().countDown();
       }
     } catch (VeniceIngestionTaskKilledException e) {
       LOGGER.info("{} has been killed.", ingestionTaskName);
@@ -3916,14 +3920,20 @@ public abstract class StoreIngestionTask implements Runnable, Closeable {
     long startTimeInMs = System.currentTimeMillis();
     close();
     try {
-      wait(waitTime);
+      if (!getGracefulShutdownLatch().await(waitTime, SECONDS)) {
+        LOGGER.info(
+            "Unable to shutdown ingestion task of topic: {} gracefully in {}ms",
+            kafkaVersionTopic,
+            SECONDS.toMillis(waitTime));
+      } else {
+        LOGGER.info(
+            "Ingestion task of topic: {} is shutdown in {}ms",
+            kafkaVersionTopic,
+            LatencyUtils.getElapsedTimeFromMsToMs(startTimeInMs));
+      }
     } catch (Exception e) {
       LOGGER.error("Caught exception while waiting for ingestion task of topic: {} shutdown.", kafkaVersionTopic);
     }
-    LOGGER.info(
-        "Ingestion task of topic: {} is shutdown in {}ms",
-        kafkaVersionTopic,
-        LatencyUtils.getElapsedTimeFromMsToMs(startTimeInMs));
   }
 
   /**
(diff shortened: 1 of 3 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -3916,14 +3921,20 @@ public synchronized void shutdown(int waitTime) {
     long startTimeInMs = System.currentTimeMillis();
     close();
     try {
-      wait(waitTime);
+      if (!getGracefulShutdownLatch().await(waitTime, SECONDS)) {
```
lusong64, line 3924: The method is protected by the keyword synchronized, meaning that the thread that executes the method already acquired the intrinsic lock of the SIT object. When the countdownLatch.await is called, the thread itself will go into wait status, and it may or may not forfeit the lock of SIT object acquired through "synchronized" before it goes into wait. It would be really bad for any thread going into sleep with a lock with it. 

Would you please verify that the lock of SIT is released? If not, we may want to change this method.

sixpluszero: Thanks for the review! I think it is a good call, and I think to address the concern, I can actually remove the existing synchronized keyword here.
The actual part that requires lock protection is the close() action, but close() will also acquire SIT lock via synchronized keyword, so this outer method does not need to hold it.

lusong64: I wrote a small piece of code to check whether what I said is right. The result shows that when one thread tries to wait for an object (a countdownLatch), it will bring the lock without releasing it with itself. Here is the code and output:

```java
import java.util.concurrent.CountDownLatch;
import java.util.concurrent.ExecutorService;
import java.util.concurrent.Executors;


public class Main {
  private final static Object OBJECT = new Object();
  private final static ExecutorService SERVICE = Executors.newFixedThreadPool(3);
  public static void main(String[] args) {
    CountDownLatch theLatch = new CountDownLatch(1);
    System.out.println("Start to the lock test.");
    SERVICE.execute(new Runnable() {
      @Override
      // The one that accquires the lock
      public void run() {
        String header = "Thread A: ";
        System.out.println(header + "Start to acquire the lock.");
        synchronized (OBJECT) {
          System.out.println(header + "get the lock and wait for the latch");
          try {
            theLatch.await();
          } catch (InterruptedException e) {
            throw new RuntimeException(e);
          }
          System.out.println(header + "Get the latch and releasing the lock");
        }

      }
    });

    SERVICE.execute(new Runnable() {
      @Override
      // The one that acquires the lock after sleeping for some time.
      public void run() {
        String header = "Thread B: ";
        long sleepTimeInSeconds = 2;
        System.out.println(header + "Sleep for " + sleepTimeInSeconds + " seconds.");
        try {
          Thread.sleep(sleepTimeInSeconds * 1000);
        } catch (InterruptedException e) {
          throw new RuntimeException(e);
        }
        System.out.println(header + "Start to acquire the lock.");
        synchronized (OBJECT) {
          System.out.println(header + "Get the lock and releasing the lock");
        }
      }
    });

    SERVICE.execute(new Runnable() {
      @Override
      // The one whose only job is to sleep and release a countdownLatch.
      public void run() {
        String header = "Thread C: ";
        long sleepTimeInSeconds = 4;
        System.out.println(header + "Sleep for " + sleepTimeInSeconds + " seconds.");
        try {
          Thread.sleep(sleepTimeInSeconds * 1000);
        } catch (InterruptedException e) {
          throw new RuntimeException(e);
        }
        System.out.println(header + "Start to release the countDownLatch.");
        theLatch.countDown();
        System.out.println(header + " Counting down finished.");
      }
    });

  }
}
``` 

Here is the output that shows ThreadB never got the lock (OBJECT) until thread A woke up: 

```shell
Start to the lock test.
Thread A: Start to acquire the lock.
Thread A: get the lock and wait for the latch
Thread B: Sleep for 2 seconds.
Thread C: Sleep for 4 seconds.
Thread B: Start to acquire the lock.
Thread C: Start to release the countDownLatch.
Thread C:  Counting down finished.
Thread A: Get the latch and releasing the lock
Thread B: Get the lock and releasing the lock

```

lusong64: Sorry Jialin. I did not realize that you already address my concerns. However, it was fun to run an experiment to make sure the suggestion is correct. Thanks!

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -3912,18 +3917,24 @@ public synchronized void close() {
...
    * @param waitTime Maximum wait time for the shutdown operation.
    */
-  public synchronized void shutdown(int waitTime) {
+  public void shutdownAndWait(int waitTime) {
     long startTimeInMs = System.currentTimeMillis();
     close();
     try {
-      wait(waitTime);
+      if (!getGracefulShutdownLatch().await(waitTime, SECONDS)) {
+        LOGGER.info(
```
gaojieliu, line 3925: This should be `WARN`, right?

sixpluszero: Sure, WARN makes more sense. Let me tweak the level.

gaojieliu (approved):

PR #1332, opened 2024-11-20
[dvc] Flag enablement for automatic subscription of partitions

### LINKED ISSUE #650: [BUG] Reconcile on disk state with assigned partitions for given resource and delete unused partitions

kristyelee:
## [dvc] Flag enablement for automatic subscription of partitions

In this PR, to make sure the correct partitions are subscribed/added to the subscription set in processes of Da Vinci Client, a code change is introduced to include a flag (`DA_VINCI_SUBSCRIBE_ON_DISK_PARTITIONS_AUTOMATICALLY`) for enabling/confirming automatic subscription of on-disk partitions. This flag is added to the config map in the VeniceConfigLoader. Flag value is checked during bootstrapping in DaVinciBackend in case automatic subscription is not wanted, so that subscription occurs correctly.
Applying this issue to Da Vinci Client which has information over partitions in storage. 

DA_VINCI_SUBSCRIBE_ON_DISK_PARTITIONS_AUTOMATICALLY is by default set to true, where in DaVinciBackend the subscription by default occurs. 

This PR is an extension of the PR https://github.com/linkedin/venice/pull/1196 that checks which partitions should be kept [in `StorageService`] and applies the check in `VeniceServer`.

Resolves #650 

## How was this PR tested?

A corresponding test will be written. 
EDIT: A corresponding integration test has been written. Example testing includes the case where the config key is set to false to prevent automatic subscription. 


clients/da-vinci-client/src/main/java/com/linkedin/davinci/StoreBackend.java
```diff
@@ -154,7 +154,13 @@ synchronized CompletableFuture<Void> subscribe(
       // Recreate store config that was potentially deleted by unsubscribe.
       config.store();
     }
+
+    ComplementSet<Integer> unassignedPartitionSet = ComplementSet.newSet(subscription);
```
kristyelee, line 158: Upon bootstrapping, it seems that there is a setting to throw a VeniceException when a version is present. 
Bootstrapping is invoked in DaVinciBackend. 
Additionally, subscription of partitions is invoked in Da Vinci Client where set of partitions can be passed into the subscribe function directly.

kvargha: What can happen is that when you call `unsubscribe` [subscription](https://github.com/linkedin/venice/blob/ce23e3895159478b5305714644f9bbeb85c1251e/clients/da-vinci-client/src/main/java/com/linkedin/davinci/StoreBackend.java#L198-L211) can be empty, causing it to delete the current version.

Also, I don't think `subscribe` is the right place to perform the validation. If a user is subscribing in a for loop, not everything would have been added to subscription yet on the first n-1 invocations. It would be better to do this validation somewhere we can guarantee all subscriptions have been performed. `DaVinciBackend::bootstrap` might be a good place.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
```diff
@@ -474,6 +475,8 @@ private synchronized void bootstrap() {
       AbstractStorageEngine storageEngine = storageService.getStorageEngine(versionTopic);
       aggVersionedStorageEngineStats.setStorageEngine(versionTopic, storageEngine);
       StoreBackend storeBackend = getStoreOrThrow(storeName);
+      ComplementSet<Integer> persistedPartitionIds = ComplementSet.newSet(storageEngine.getPersistedPartitionIds());
+      storeBackend.unsubscribe(persistedPartitionIds);
```
kvargha, line 479: This would unsubscribe every partition on disk, which isn't what we want to happen.

We want to unsubscribe persisted partitions that don't exist in `storeBackend::subscription`

clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
```diff
@@ -466,16 +468,22 @@ private synchronized void bootstrap() {
...
+        List<Integer> partitions = storeNameToPartitionListMap.get(storeName);
+        String versionTopic = version.kafkaTopicName();
+        LOGGER.info("Bootstrapping partitions {} for {}", partitions, versionTopic);
+        AbstractStorageEngine storageEngine = storageService.getStorageEngine(versionTopic);
+        aggVersionedStorageEngineStats.setStorageEngine(versionTopic, storageEngine);
+        StoreBackend storeBackend = getStoreOrThrow(storeName);
+        ComplementSet<Integer> subscription = ComplementSet.newSet(storeBackend.getSubscription());
+        ComplementSet<Integer> unassignedPartitionSet = ComplementSet.newSet(storageEngine.getPersistedPartitionIds());
+        unassignedPartitionSet.removeAll(subscription);
+        storeBackend.unsubscribe(unassignedPartitionSet);
```
kvargha, line 483: With Jialin's new approach, we won't need to unsubscribe here.

We'll just keep the state on disk, but not automatically subscribe.

Although we pay the cost of extra disk space, at least we save in CPU/Memory.

internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
```diff
@@ -97,6 +97,9 @@ private ConfigKeys() {
   public static final String DA_VINCI_CURRENT_VERSION_BOOTSTRAPPING_QUOTA_BYTES_PER_SECOND =
       "da.vinci.current.version.bootstrapping.quota.bytes.per.second";
 
+  // Prevent automatic subscription of partitions while bootstrapping.
+  public static final String DA_VINCI_BOOTSTRAP_SUBSCRIPTION_DISABLED = "da.vinci.bootstrap.subscription.disabled";
```
kvargha, line 101: Can we rename this to `da.vinci.subscribe.on.disk.partitions.automatically`?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
kristyelee: [This resolves the [comment](https://github.com/linkedin/venice/pull/1332#discussion_r1870376282) above over subscription.]

kristyelee: Right now, a ConfigKey is used to enable automatic subscription of on disk partitions.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/DaVinciClientTest.java
```diff
@@ -1283,6 +1283,94 @@ public void testBootstrap(DaVinciConfig daVinciConfig) throws Exception {
     }
   }
 
+  @Test(timeOut = TEST_TIMEOUT, dataProvider = "dv-client-config-provider", dataProviderClass = DataProviderUtils.class)
+  public void testBootstrapSubscription(DaVinciConfig clientConfig) throws Exception {
```
kvargha, line 1287: Since `DA_VINCI_SUBSCRIBE_ON_DISK_PARTITIONS_AUTOMATICALLY` is already true by default, we should instead write a test where it's set to false.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/AvroGenericDaVinciClient.java
```diff
@@ -710,6 +711,7 @@ private VeniceConfigLoader buildVeniceConfig() {
         .put(CLUSTER_NAME, clusterName)
         .put(ZOOKEEPER_ADDRESS, zkAddress)
         .put(KAFKA_BOOTSTRAP_SERVERS, kafkaBootstrapServers)
+        .put(DA_VINCI_SUBSCRIBE_ON_DISK_PARTITIONS_AUTOMATICALLY, true)
```
ZacAttack, line 714: Maybe this should default to false since it's a change in behavior?  @sixpluszero what do you think?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
```diff
@@ -466,16 +468,18 @@ private synchronized void bootstrap() {
...
-    storeNameToBootstrapVersionMap.forEach((storeName, version) -> {
-      List<Integer> partitions = storeNameToPartitionListMap.get(storeName);
-      String versionTopic = version.kafkaTopicName();
-      LOGGER.info("Bootstrapping partitions {} for {}", partitions, versionTopic);
-      AbstractStorageEngine storageEngine = storageService.getStorageEngine(versionTopic);
-      aggVersionedStorageEngineStats.setStorageEngine(versionTopic, storageEngine);
-      StoreBackend storeBackend = getStoreOrThrow(storeName);
-      storeBackend.subscribe(ComplementSet.newSet(partitions), Optional.of(version));
-    });
+    if (configLoader.getCombinedProperties().getBoolean(DA_VINCI_SUBSCRIBE_ON_DISK_PARTITIONS_AUTOMATICALLY, true)) {
```
ZacAttack, line 471: This looks like a good start, but do we need to do clean up?  @sixpluszero @kvargha ?

ZacAttack: Maybe we add a method to the interface which allows users to clean up unsubscribed state?

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/DaVinciClientTest.java
```diff
@@ -83,15 +86,7 @@
...
-import com.linkedin.venice.utils.DataProviderUtils;
-import com.linkedin.venice.utils.ForkedJavaProcess;
-import com.linkedin.venice.utils.IntegrationTestPushUtils;
-import com.linkedin.venice.utils.Pair;
-import com.linkedin.venice.utils.PropertyBuilder;
-import com.linkedin.venice.utils.TestUtils;
-import com.linkedin.venice.utils.TestWriteUtils;
-import com.linkedin.venice.utils.Utils;
-import com.linkedin.venice.utils.VeniceProperties;
+import com.linkedin.venice.utils.*;
```
ZacAttack, line 89: nit turn off * imports

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/DaVinciClientTest.java
```diff
@@ -1283,6 +1278,88 @@ public void testBootstrap(DaVinciConfig daVinciConfig) throws Exception {
...
+        metricsRepository,
+        backendConfig)) {
+      DaVinciClient<Integer, Object> client1 = factory.getAndStartGenericAvroClient(storeName1, daVinciConfig);
+
+      // Test non-existent key access
+      client1.subscribeAll().get();
+      assertNull(client1.get(KEY_COUNT + 1).get());
+
+      // Test single-get access
+      Map<Integer, Integer> keyValueMap = new HashMap<>();
```
ZacAttack, line 1310: Let's repeat this logic on the reinstantiation but make sure it always returns null/empty since we don't want to serve data we're not subscribed to.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
kvargha: Gaojie proposed that we keep the state on disk to play it on the safe side, as with a new version it will automatically get cleaned up.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/DaVinciClientTest.java
```diff
@@ -1283,6 +1287,97 @@ public void testBootstrap(DaVinciConfig daVinciConfig) throws Exception {
...
+
+        Integer expectedValue = cluster.createVersion(storeName1, KEY_COUNT);
+        TestUtils.waitForNonDeterministicAssertion(TEST_TIMEOUT, TimeUnit.MILLISECONDS, () -> {
+          for (int k = 0; k < KEY_COUNT; ++k) {
+            Object readValue = client1.get(k).get();
+            assertEquals(readValue, expectedValue);
+          }
+        });
+      }
+    }
```
kvargha, line 1351: I don't think we need this for this test.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/DaVinciClientTest.java
```diff
@@ -1283,6 +1287,97 @@ public void testBootstrap(DaVinciConfig daVinciConfig) throws Exception {
...
+      }
+
+      client1.subscribe(partitions);
+      client1.unsubscribeAll();
+
+      DaVinciBackend daVinciBackend = AvroGenericDaVinciClient.getBackend();
+      if (daVinciBackend != null) {
+        StoreBackend storeBackend = daVinciBackend.getStoreOrThrow(storeName1);
+        ComplementSet<Integer> subscription = storeBackend.getSubscription();
+        assertTrue(subscription.isEmpty());
```
kvargha, line 1376: What we can do here instead is we can stick with the default partition count of 3 instead of modifying it, and in the new client we only subscribe to 2 partitions and validate that `subscription` only has 2 in the `ComplementSet`.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/AvroGenericDaVinciClient.java
sixpluszero: Just to clarify, this is the current behavior, we should mark it as true.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
```diff
@@ -665,6 +669,10 @@ protected void handleStoreChanged(StoreBackend storeBackend) {
     storeBackend.trySubscribeDaVinciFutureVersion();
   }
 
+  protected void executeBootstrapping() {
+    bootstrap();
+  }
+
```
kvargha, line 675: Instead of doing this, can you change the access modifier on `bootstrap()` to `protected`?

clients/da-vinci-client/src/test/java/com/linkedin/davinci/DaVinciBackendTest.java
```diff
@@ -139,4 +167,101 @@ public void testBootstrappingAwareCompletableFuture()
...
+
+    // DA_VINCI_SUBSCRIBE_ON_DISK_PARTITIONS_AUTOMATICALLY == false
+    when(mockCombinedProperties.getBoolean(anyString(), anyBoolean())).thenReturn(false);
+    doCallRealMethod().when(backend).executeBootstrapping();
+    backend.executeBootstrapping();
+
+    ComplementSet<Integer> subscription = mockStoreBackend.getSubscription();
+    assertTrue(subscription.contains(0));
+    assertTrue(subscription.contains(1));
+    assertFalse(subscription.contains(2));
```
kvargha, line 265: Can you add another section to this test where `DA_VINCI_SUBSCRIBE_ON_DISK_PARTITIONS_AUTOMATICALLY` == `true` and all of the partitions are in `subscription`?

kristyelee: This has been added.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java
kristyelee: Reverted access modifier of `bootstrap()` back to `private`. Used reflection of method in unit testing for correct outcome.

sixpluszero: Hmm, I have a different opinion on this. Using reflection is generally not recommend in unit test. It is considered less readable and maintainable for other developer and it breaks encapsulation of private modifier. 
We can use default modifier (just nothing) or protected to make it accessible in unit test.

clients/da-vinci-client/src/test/java/com/linkedin/davinci/DaVinciBackendTest.java
```diff
@@ -139,4 +168,114 @@ public void testBootstrappingAwareCompletableFuture()
...
+    ingestionListenerField.setAccessible(true);
+    ingestionListenerField.set(backend, ingestionListener);
+    KafkaStoreIngestionService storeIngestionService = mock(KafkaStoreIngestionService.class);
+    Field storeIngestionServiceField = DefaultIngestionBackend.class.getDeclaredField("storeIngestionService");
+    storeIngestionServiceField.setAccessible(true);
+    storeIngestionServiceField.set(ingestionBackend, storeIngestionService);
+    Field ingestionServiceField = DaVinciBackend.class.getDeclaredField("ingestionService");
+    ingestionServiceField.setAccessible(true);
+    ingestionServiceField.set(backend, storeIngestionService);
+    doNothing().when(ingestionBackend).addIngestionNotifier(any());
```
kvargha, line 258: Since `ingestionBackend` gets defined inside `bootstrap`, can we try removing some of the unnecessary mocks?

clients/da-vinci-client/src/test/java/com/linkedin/davinci/DaVinciBackendTest.java
```diff
@@ -139,4 +163,103 @@ public void testBootstrappingAwareCompletableFuture()
...
 
+  @Test
+  public void testBootstrappingSubscription()
+      throws IllegalAccessException, NoSuchFieldException, NoSuchMethodException, InvocationTargetException {
+    DaVinciBackend backend = mock(DaVinciBackend.class);
+    StorageService mockStorageService = mock(StorageService.class);
+
+    Field storageServiceField = DaVinciBackend.class.getDeclaredField("storageService");
+    storageServiceField.setAccessible(true);
+    storageServiceField.set(backend, mockStorageService);
```
kvargha, line 174: We don't need this as storageService is now being mocked by line 187.

sixpluszero (approved):
LGTM, thank you!

PR #1333, opened 2024-11-21
[controller] Fix node removable check with locked instances

nisargthakkar:
## Fix node removable check with locked instances

Currently, there is a bug where even if an instance is non-removable, if it doesn't share any partition with some locked instance, it returns a removable result for the instance.

This commit also depracates the cluster view in the `node_removable` check and removes the `instance_view` API param.

## How was this PR tested?

Added extra checks in unit tests

    - Before this change, invoking `/node_removable` endpoint on controllers would check if the cluster view is healthy after removal of the specified instance. However, this is not desirable because it would block the removal of the instance even if it doesn't cause any change to the cluster's healthiness. This mode has been deprecated in favor of the `instance_view` mode that was previously available by specifying an API param. Specifying the API param is now a no-op, so, it is still a backwards compatible change

### DIFF
--- a/services/venice-controller/src/main/java/com/linkedin/venice/controller/InstanceStatusDecider.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/controller/InstanceStatusDecider.java
@@ -32,27 +32,18 @@ public class InstanceStatusDecider {
   }
 
   static NodeRemovableResult isRemovable(HelixVeniceClusterResources resources, String clusterName, String instanceId) {
-    return isRemovable(resources, clusterName, instanceId, Collections.emptyList(), false);
+    return isRemovable(resources, clusterName, instanceId, Collections.emptyList());
   }
 
   private static PartitionAssignment removeLockedResources(
       PartitionAssignment partitionAssignment,
       List<String> lockedNodes,
-      String resourceName,
-      boolean isInstanceView) {
+      String resourceName) {
     PartitionAssignment assignment = partitionAssignment;
     for (String lockedNodeId: lockedNodes) {
-      assignment = getPartitionAssignmentAfterRemoving(lockedNodeId, assignment, resourceName, isInstanceView);
+      assignment = getPartitionAssignmentAfterRemoving(lockedNodeId, assignment, resourceName, false);
     }
     return assignment;
-  }
-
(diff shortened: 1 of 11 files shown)


services/venice-controller/src/main/java/com/linkedin/venice/controller/InstanceStatusDecider.java
```diff
@@ -32,29 +32,20 @@ static List<Replica> getReplicasForInstance(HelixVeniceClusterResources resource
...
   private static PartitionAssignment removeLockedResources(
       PartitionAssignment partitionAssignment,
       List<String> lockedNodes,
-      String resourceName,
-      boolean isInstanceView) {
+      String resourceName) {
     PartitionAssignment assignment = partitionAssignment;
     for (String lockedNodeId: lockedNodes) {
-      assignment = getPartitionAssignmentAfterRemoving(lockedNodeId, assignment, resourceName, isInstanceView);
+      assignment = getPartitionAssignmentAfterRemoving(lockedNodeId, assignment, resourceName, false);
```
majisourav99, line 44: Can we add some comment here why we are passing instanceView false for locked nodes.

nisargthakkar: Sure. Will add one

services/venice-controller/src/main/java/com/linkedin/venice/controller/InstanceStatusDecider.java
```diff
@@ -241,14 +225,18 @@ private static PartitionAssignment getPartitionAssignmentAfterRemoving(
       String instanceId,
       PartitionAssignment partitionAssignment,
       String resourceName,
-      boolean isInstanceView) {
+      boolean lockedInstance) {
     PartitionAssignment partitionAssignmentAfterRemoving =
         new PartitionAssignment(resourceName, partitionAssignment.getExpectedNumberOfPartitions());
     for (Partition partition: partitionAssignment.getAllPartitions()) {
-      if (isInstanceView) {
+      if (!lockedInstance) {
```
majisourav99, line 232: Thanks renaming the variable, more informative

majisourav99 (approved):
LGTM

PR #1334, opened 2024-11-21
Create separate path for CI reports

majisourav99:
## Create separate path for CI reports

Show CI report for different modules in different links.

### DIFF
--- a/.github/workflows/UnitTests-core.yml
+++ b/.github/workflows/UnitTests-core.yml
@@ -39,7 +39,7 @@ jobs:
           git fetch upstream
       - name: Setup Gradle
         uses: gradle/actions/setup-gradle@v4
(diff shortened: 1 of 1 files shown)


.github/workflows/UnitTests-core.yml
```diff
@@ -52,9 +52,11 @@ jobs:
       - name: Publish Test Report
         env:
           NODE_OPTIONS: "--max_old_space_size=4096"
-        uses: mikepenz/action-junit-report@v3
```
sushantmane, line 55: IMO this has to be right after 
```
Run Unit Tests & Code Coverage
```
If we want to run it on `failure` only

sushantmane:
Instead of check can we always use annotations always?

PR #1335, opened 2024-11-21
[server] swap map for list fields when examining idealstate for disk cleanup

ZacAttack:
## [server] swap map for list fields when examining idealstate for disk cleanup

In a helix ideal state, helix maintains information in listfields and mapfields.  The mapfields inform which messages needs to be sent based on the current LIVEINSTANCES, and will therefore be reflective of what nodes are available to receive messages.  The listfields however will contain the assignment, and will only change if a rebalance is calculated.  For this clean up code, we rely on entries in the listfield based on which state transitions we're anticipating to receive.

Resolves #XXX

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/storage/StorageService.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/storage/StorageService.java
@@ -395,7 +395,8 @@ public class StorageService extends AbstractVeniceService {
             continue;
           }
           String partitionDbName = storeName + "_" + partitionId;
-          if (!listFields.containsKey(partitionDbName) || !listFields.get(partitionDbName).contains(instanceHostName)) {
+          List<String> hostNames = listFields.get(partitionDbName);
+          if (hostNames != null || !listFields.get(partitionDbName).contains(instanceHostName)) {
             LOGGER.info(
                 "the following partition is not assigned to the current host {} and is being dropped from storage engine {}: {}",
                 instanceHostName,
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/storage/StorageService.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/storage/StorageService.java
@@ -387,7 +387,13 @@ public class StorageService extends AbstractVeniceService {
           new PropertyKey.Builder(configLoader.getVeniceClusterConfig().getClusterName());
       SafeHelixDataAccessor helixDataAccessor = manager.getHelixDataAccessor();
       IdealState idealState = helixDataAccessor.getProperty(propertyKeyBuilder.idealStates(storeName));
-
+      /**
+       * In a helix ideal state, helix maintains information in listfields and mapfields.  The mapfields inform which
+       * messages needs to be sent based on the current LIVEINSTANCES, and will therefore be reflective of what
+       * nodes are available to receive messages.  The listfields however will contain the assignment, and will only
+       * change if a rebalance is calculated.  For this clean up code, we rely on entries in the listfield based
+       * on which state transitions we're anticipating to receive.
+       */
       if (idealState != null) {
         Map<String, List<String>> listFields = idealState.getRecord().getListFields();
         for (Integer partitionId: storageEnginePartitionIds) {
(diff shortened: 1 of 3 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/storage/StorageService.java
```diff
@@ -389,14 +389,13 @@ public synchronized void checkWhetherStoragePartitionsShouldBeKeptOrNot(SafeHeli
       IdealState idealState = helixDataAccessor.getProperty(propertyKeyBuilder.idealStates(storeName));
 
       if (idealState != null) {
-        Map<String, Map<String, String>> mapFields = idealState.getRecord().getMapFields();
+        Map<String, List<String>> mapFields = idealState.getRecord().getListFields();
```
huangminchn: Nitpick: shall we change the variable name to "listFields"?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/storage/StorageService.java
```diff
@@ -389,14 +389,13 @@ public synchronized void checkWhetherStoragePartitionsShouldBeKeptOrNot(SafeHeli
...
-        Map<String, Map<String, String>> mapFields = idealState.getRecord().getMapFields();
+        Map<String, List<String>> listFields = idealState.getRecord().getListFields();
         for (Integer partitionId: storageEnginePartitionIds) {
           if (storageEngine.isMetadataPartition(partitionId)) {
             continue;
           }
           String partitionDbName = storeName + "_" + partitionId;
-          if (!mapFields.containsKey(partitionDbName)
-              || !mapFields.get(partitionDbName).containsKey(instanceHostName)) {
+          if (!listFields.containsKey(partitionDbName) || !listFields.get(partitionDbName).contains(instanceHostName)) {
```
FelixGV, line 398: Nitpick, can we avoid `containsKey`? e.g.,

```
List<String> hostNames = listFields.get(partitionDbName);
if (hostNames == null || !hostNames.contains(instanceHostName)) {
```

Looks cleaner IMHO.

FelixGV (commented):
Can we flesh out the commit message a bit further? Is this some Helix API behavior we're relying on? Why does it matter concretely for Venice? (I think I can guess why today, but I'm pretty sure I'll have forgotten by the time I read this commit message 6 months from now)

kvargha:
> Can we flesh out the commit message a bit further? Is this some Helix API behavior we're relying on? Why does it matter concretely for Venice? (I think I can guess why today, but I'm pretty sure I'll have forgotten by the time I read this commit message 6 months from now)

Maybe this should be a comment in the code too, such as describing why we're using listFields instead of mapFields.

kvargha (dismissed):

clients/da-vinci-client/src/main/java/com/linkedin/davinci/storage/StorageService.java
```diff
@@ -387,16 +387,22 @@ public synchronized void checkWhetherStoragePartitionsShouldBeKeptOrNot(SafeHeli
...
+        Map<String, List<String>> listFields = idealState.getRecord().getListFields();
         for (Integer partitionId: storageEnginePartitionIds) {
           if (storageEngine.isMetadataPartition(partitionId)) {
             continue;
           }
           String partitionDbName = storeName + "_" + partitionId;
-          if (!mapFields.containsKey(partitionDbName)
-              || !mapFields.get(partitionDbName).containsKey(instanceHostName)) {
+          List<String> hostNames = listFields.get(partitionDbName);
+          if (hostNames != null || !listFields.get(partitionDbName).contains(instanceHostName)) {
```
majisourav99, line 405: minor efficiency: can we convert the list to set ?

ZacAttack: I'm not keen to convert anything helix gives me, that sounds risky.

FelixGV: In any case, traversing the list in order to populate the set and then doing a single lookup in it is probably not significantly more efficient (if at all) than getting the list as is and traversing it to compare each element to the one we're looking for?

FelixGV, line 405: `!containsKey` should become `== null`, right? Otherwise you're staring down the barrel of a NPE?

Also, can you use `hostNames` rather than looking it up again?

FelixGV (changes requested):

clients/da-vinci-client/src/main/java/com/linkedin/davinci/storage/StorageService.java
ZacAttack: whoops

FelixGV (approved):
Thanks!

PR #1338, opened 2024-11-22
[controller] Ensure real-time topic partition count matches hybrid version partition count

sushantmane:
##  Ensure real-time topic partition count matches hybrid version partition count

This fix addresses an issue where the real-time topic partition count did not align 
with the hybrid version partition count, which caused ingestion failures in hybrid 
stores.

### Problem Description

The issue occurs in the following scenario:
1. Create a store with 1 partition.
2. Perform a batch push, creating a batch version with 1 partition.
3. Update the store to 3 partitions and convert it to a hybrid store.
4. Start real-time writes using push type `STREAM`.
5. Perform a full push to create a hybrid version with 3 partitions. This push fails 
   because, after the topic switch, real-time consumers cannot find partitions 2 and 3 
   due to the real-time topic having only 1 partition.

### Root Cause
- In step 4, if the real-time topic did not exist, it gets created with a partition 
  count derived from the largest existing version (batch version with 1 partition), 
  which led to a mismatch.

### Solution

- Stream and incremental push types are now disallowed if there is no online hybrid 
  version.
- If an online hybrid version exists, the real-time topic partition count is now 
  validated to match the hybrid version partition count.
- The `requestTopicForPushing` method no longer creates a real-time topic if it does 
  not already exist, except during new hybrid version creation.
- For new hybrid versions, the real-time topic is created at version creation time 
  instead of during a topic switch. The logic for real-time topic creation during a 
  topic switch is retained for backward compatibility and will be removed once the 
  release with these changes is fully rolled out.

### Miscellaneous Changes

- Refactored `CreateVersion::requestTopicForPushing` to improve unit testability.
- Added similar validation checks for incremental push job types.
- Stopped creating real-time topics for regional system stores like meta and PS3 in 
  the parent region.
- Improved logging around topic switches to make it easier to debug and search.

## How was this PR tested?

UTs and E2E in CI

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
@@ -846,6 +846,7 @@ public class LeaderFollowerStoreIngestionTask extends StoreIngestionTask {
   private boolean isLocalVersionTopicPartitionFullyConsumed(PartitionConsumptionState pcs) {
     long localVTOff = pcs.getLatestProcessedLocalVersionTopicOffset();
     long localVTEndOffset = getTopicPartitionEndOffSet(localKafkaServer, versionTopic, pcs.getPartition());
+
     if (localVTEndOffset == StatsErrorCode.LAG_MEASUREMENT_FAILURE.code) {
       return false;
     }
--- a/internal/venice-common/src/main/java/com/linkedin/venice/controllerapi/RequestTopicForPushRequest.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/controllerapi/RequestTopicForPushRequest.java
@@ -49,15 +49,6 @@ public class RequestTopicForPushRequest {
     this.pushJobId = pushJobId;
   }
 
-  public static PushType extractPushType(String pushTypeString) {
-    try {
-      return PushType.valueOf(pushTypeString);
-    } catch (IllegalArgumentException e) {
-      throw new IllegalArgumentException(
-          pushTypeString + " is an invalid push type. Valid push types are: " + Arrays.toString(PushType.values()));
-    }
-  }
-
   public String getClusterName() {
     return clusterName;
   }
--- a/internal/venice-common/src/main/java/com/linkedin/venice/controllerapi/RequestTopicForPushRequest.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/controllerapi/RequestTopicForPushRequest.java
@@ -1,0 +1,187 @@
+package com.linkedin.venice.controllerapi;
+
+import com.linkedin.venice.meta.Version.PushType;
+import java.security.cert.X509Certificate;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+
+
+public class RequestTopicForPushRequest {
+  private final String clusterName;
+  private final String storeName;
+  private final PushType pushType;
+  private final String pushJobId;
+
+  private boolean sendStartOfPush = false;
+  private boolean sorted = false; // an inefficient but safe default
+  private boolean isWriteComputeEnabled = false;
+  private boolean separateRealTimeTopicEnabled = false;
+  private long rewindTimeInSecondsOverride = -1L;
+  private boolean deferVersionSwap = false;
+  private String targetedRegions = null;
+  private int repushSourceVersion = -1;
+  private Set<String> partitioners = Collections.emptySet();
+  private String compressionDictionary = null;
+  private X509Certificate certificateInRequest = null;
+  private String sourceGridFabric = null;
+  private String emergencySourceRegion = null;
+
+  public RequestTopicForPushRequest(String clusterName, String storeName, PushType pushType, String pushJobId) {
+    if (clusterName == null || clusterName.isEmpty()) {
+      throw new IllegalArgumentException("clusterName is required");
+    }
+    if (storeName == null || storeName.isEmpty()) {
+      throw new IllegalArgumentException("storeName is required");
+    }
+    if (pushType == null) {
+      throw new IllegalArgumentException("pushType is required");
+    }
+
+    if (pushJobId == null || pushJobId.isEmpty()) {
+      throw new IllegalArgumentException("pushJobId is required");
+    }
+
+    this.clusterName = clusterName;
+    this.storeName = storeName;
+    this.pushType = pushType;
+    this.pushJobId = pushJobId;
+  }
+
+  public static PushType extractPushType(String pushTypeString) {
+    try {
+      return PushType.valueOf(pushTypeString);
+    } catch (IllegalArgumentException e) {
+      throw new IllegalArgumentException(
+          pushTypeString + " is an invalid push type. Valid push types are: " + Arrays.toString(PushType.values()));
+    }
+  }
+
+  public String getClusterName() {
+    return clusterName;
+  }
+
+  public String getStoreName() {
+    return storeName;
+  }
+
+  public PushType getPushType() {
+    return pushType;
+  }
+
+  public String getPushJobId() {
+    return pushJobId;
+  }
+
+  public boolean isSendStartOfPush() {
+    return sendStartOfPush;
+  }
+
+  public boolean isSorted() {
+    return sorted;
+  }
+
+  public boolean isWriteComputeEnabled() {
+    return isWriteComputeEnabled;
+  }
+
+  public String getSourceGridFabric() {
+    return sourceGridFabric;
+  }
+
+  public long getRewindTimeInSecondsOverride() {
+    return rewindTimeInSecondsOverride;
+  }
+
+  public boolean isDeferVersionSwap() {
+    return deferVersionSwap;
+  }
+
+  public String getTargetedRegions() {
+    return targetedRegions;
+  }
+
+  public int getRepushSourceVersion() {
+    return repushSourceVersion;
+  }
+
+  public Set<String> getPartitioners() {
+    return partitioners;
+  }
+
+  public String getCompressionDictionary() {
+    return compressionDictionary;
+  }
+
+  public X509Certificate getCertificateInRequest() {
+    return certificateInRequest;
+  }
+
+  public String getEmergencySourceRegion() {
+    return emergencySourceRegion;
+  }
+
+  public void setSendStartOfPush(boolean sendStartOfPush) {
+    this.sendStartOfPush = sendStartOfPush;
+  }
+
+  public void setSorted(boolean sorted) {
+    this.sorted = sorted;
+  }
+
+  public void setWriteComputeEnabled(boolean writeComputeEnabled) {
+    isWriteComputeEnabled = writeComputeEnabled;
+  }
+
+  public void setSourceGridFabric(String sourceGridFabric) {
+    this.sourceGridFabric = sourceGridFabric;
+  }
+
+  public void setRewindTimeInSecondsOverride(long rewindTimeInSecondsOverride) {
+    this.rewindTimeInSecondsOverride = rewindTimeInSecondsOverride;
+  }
+
+  public void setDeferVersionSwap(boolean deferVersionSwap) {
+    this.deferVersionSwap = deferVersionSwap;
+  }
+
+  public void setTargetedRegions(String targetedRegions) {
+    this.targetedRegions = targetedRegions;
+  }
+
+  public void setRepushSourceVersion(int repushSourceVersion) {
+    this.repushSourceVersion = repushSourceVersion;
+  }
+
+  public void setPartitioners(String commaSeparatedPartitioners) {
+    if (commaSeparatedPartitioners == null || commaSeparatedPartitioners.isEmpty()) {
+      return;
+    }
+    setPartitioners(new HashSet<>(Arrays.asList(commaSeparatedPartitioners.split(","))));
+  }
+
+  public void setPartitioners(Set<String> partitioners) {
+    this.partitioners = partitioners;
+  }
+
+  public void setCompressionDictionary(String compressionDictionary) {
+    this.compressionDictionary = compressionDictionary;
+  }
+
+  public void setCertificateInRequest(X509Certificate certificateInRequest) {
+    this.certificateInRequest = certificateInRequest;
+  }
+
+  public void setEmergencySourceRegion(String emergencySourceRegion) {
+    this.emergencySourceRegion = emergencySourceRegion;
+  }
+
+  public boolean isSeparateRealTimeTopicEnabled() {
+    return separateRealTimeTopicEnabled;
+  }
+
+  public void setSeparateRealTimeTopicEnabled(boolean separateRealTimeTopicEnabled) {
+    this.separateRealTimeTopicEnabled = separateRealTimeTopicEnabled;
+  }
+}
--- a/internal/venice-common/src/main/java/com/linkedin/venice/pubsub/PubSubConstants.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/pubsub/PubSubConstants.java
@@ -65,7 +65,7 @@ public class PubSubConstants {
    * Default value of sleep interval for polling topic deletion status from ZK.
    */
   public static final int PUBSUB_TOPIC_DELETION_STATUS_POLL_INTERVAL_MS_DEFAULT_VALUE = 2 * Time.MS_PER_SECOND;
-  public static final long UNKNOWN_LATEST_OFFSET = -12345;
+  public static final long UNKNOWN_LATEST_OFFSET = -404L;
 
   private static final Duration PUBSUB_OFFSET_API_TIMEOUT_DURATION_DEFAULT_VALUE_DEFAULT = Duration.ofMinutes(1);
   private static Duration PUBSUB_OFFSET_API_TIMEOUT_DURATION_DEFAULT_VALUE =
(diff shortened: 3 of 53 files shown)


sushantmane:
Moving to draft as I would like to move topic creation in addVersion and see how that works out

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -834,6 +834,7 @@ private boolean canSwitchToLeaderTopic(PartitionConsumptionState pcs) {
   private boolean isLocalVersionTopicPartitionFullyConsumed(PartitionConsumptionState pcs) {
     long localVTOff = pcs.getLatestProcessedLocalVersionTopicOffset();
     long localVTEndOffset = getTopicPartitionEndOffSet(localKafkaServer, versionTopic, pcs.getPartition());
+
```
sushantmane, line 837: I suspect there is a bug here in this function.  I need to think through

internal/venice-common/src/main/java/com/linkedin/venice/controllerapi/RequestTopicForPushRequest.java
```diff
@@ -0,0 +1,178 @@
...
+      throw new IllegalArgumentException("pushJobId is required");
+    }
+
+    this.clusterName = clusterName;
+    this.storeName = storeName;
+    this.pushType = pushType;
+    this.pushJobId = pushJobId;
+  }
+
+  public static PushType extractPushType(String pushTypeString) {
```
lusong64, line 51: In the PushType enum itself, we already have an valueOf (int), would it be better to add this method there to have all the valueOf in one place? 

And this method is simply catch an illegalArgumentException and throw a new illegalArgumentException with a more informative error message. Catching exception is expensive. So maybe inside the PushType enum, you can pre-build a static map of pushType string, if the inputString does not match any, you can throw an illegalArgumentException. 

Just my two-cents.

services/venice-controller/src/main/java/com/linkedin/venice/controller/server/CreateVersion.java
```diff
@@ -71,6 +74,356 @@ public CreateVersion(
...
+
+    request.setCompressionDictionary(httpRequest.queryParamOrDefault(COMPRESSION_DICTIONARY, null));
+
+    // Retrieve certificate from request if ACL is enabled
+    request.setCertificateInRequest(isAclEnabled ? getCertificate(httpRequest) : null);
+  }
+
+  private static void verifyPartitioner(PartitionerConfig storePartitionerConfig, Set<String> partitionersFromRequest) {
+    // If partitioners are provided, check if the store partitioner is in the list
+    if (partitionersFromRequest != null && !partitionersFromRequest.isEmpty()
```
lusong64, line 122: is the isEmpty test here necessary? If a set is empty, your third test contains(xXX) will return false.

services/venice-controller/src/main/java/com/linkedin/venice/controller/server/CreateVersion.java
```diff
@@ -71,6 +74,356 @@ public CreateVersion(
...
+        Optional.ofNullable(request.getCertificateInRequest()),
+        request.getRewindTimeInSecondsOverride(),
+        Optional.ofNullable(request.getEmergencySourceRegion()),
+        request.isDeferVersionSwap(),
+        request.getTargetedRegions(),
+        request.getRepushSourceVersion());
+
+    // If Version partition count different from calculated partition count use the version count as store count
+    // may have been updated later.
+    if (version.getPartitionCount() != response.getPartitions()) {
```
lusong64, line 280: should here we log the partition count change so that we won't have to scratch our head in future troubleshoot?

arjun4084346:
I agree with the solution "STREAM push type is now disallowed if there is no online hybrid version." However, I think, instead of returning error, we should create a new version (that would be of right partition count) and then let the push start.
If we do not auto create the version, we can at least tell the user to create a version, if we find store's partition count different from the VT's partition count.

If we look carefully, changing partition count without creating a version (even for batch stores) is similar to changing partition for a hybrid store; even for the batch stores, because batch stores can be changed to hybrid

I am skeptical about "The requestTopicForPushing method no longer creates a real-time topic if it does not already exist."; I believe in most of the cases, this is a benign operation and we should not prevent creation of the RT.

sushantmane:
> I agree with the solution "STREAM push type is now disallowed if there is no online hybrid version." However, I think, instead of returning error, we should create a new version (that would be of right partition count) and then let the push start. If we do not auto create the version, we can at least tell the user to create a version, if we find store's partition count different from the VT's partition count.
> 
> If we look carefully, changing partition count without creating a version (even for batch stores) is similar to changing partition for a hybrid store; even for the batch stores, because batch stores can be changed to hybrid
> 
> I am skeptical about "The requestTopicForPushing method no longer creates a real-time topic if it does not already exist."; I believe in most of the cases, this is a benign operation and we should not prevent creation of the RT.


> we should create a new version (that would be of right partition count)  
I think this may not be the safe in some cases (user wants to create a new version but keep old data). 


>> If we look carefully, changing partition count without creating a version (even for batch stores) is similar to changing partition for a hybrid store; even for the batch stores, because batch stores can be changed to hybrid
I'm don't really follow the rest of the comment. Could you please elaborate? Thanks

internal/venice-common/src/main/java/com/linkedin/venice/controllerapi/RequestTopicForPushRequest.java
sushantmane: I agree with you. I've updated code to use pre-built map.

services/venice-controller/src/main/java/com/linkedin/venice/controller/server/CreateVersion.java
sushantmane: Yes, the `partitionersFromRequest.isEmpty()` check is necessary. The intent of the method is to verify if the optional partitioner list provided by the user contains the partitioner from the store’s configuration. If we do not explicitly check for isEmpty, and the user provides an empty set, we would inadvertently throw an exception. This is not the expected behavior because an empty set implies that no specific partitioners were requested, and we should simply fall back to the store’s partitioner without raising an error.

services/venice-controller/src/main/java/com/linkedin/venice/controller/server/CreateVersion.java
sushantmane: Removed this as it is confusing and prone to errors. New change will rely on partition count of reference version only

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/PubSubConstants.java
```diff
@@ -65,6 +65,7 @@ public class PubSubConstants {
    * Default value of sleep interval for polling topic deletion status from ZK.
    */
   public static final int PUBSUB_TOPIC_DELETION_STATUS_POLL_INTERVAL_MS_DEFAULT_VALUE = 2 * Time.MS_PER_SECOND;
+  public static final long UNKNOWN_LATEST_OFFSET = -12345;
```
arjun4084346, line 68: is there any significance for this number or any negative number would work?

arjun4084346: *nit long should end in `L`

internal/venice-common/src/main/java/com/linkedin/venice/utils/Utils.java
```diff
@@ -309,6 +309,12 @@ public static long parseLongFromString(String value, String fieldName) {
    * any string that are not equal to 'true', We validate the string by our own.
    */
   public static boolean parseBooleanFromString(String value, String fieldName) {
+    if (value == null) {
```
arjun4084346, line 312: I am confused, why does this method require `fieldName` ???

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerClusterConfig.java
```diff
@@ -826,7 +826,7 @@ public VeniceControllerClusterConfig(VeniceProperties props) {
     this.parentControllerMaxErroredTopicNumToKeep = props.getInt(PARENT_CONTROLLER_MAX_ERRORED_TOPIC_NUM_TO_KEEP, 0);
 
     this.pushJobStatusStoreClusterName = props.getString(PUSH_JOB_STATUS_STORE_CLUSTER_NAME, "");
-    this.participantMessageStoreEnabled = props.getBoolean(PARTICIPANT_MESSAGE_STORE_ENABLED, false);
+    this.participantMessageStoreEnabled = props.getBoolean(PARTICIPANT_MESSAGE_STORE_ENABLED, true);
```
arjun4084346, line 829: Why this change?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -3005,6 +2971,172 @@ private Pair<Boolean, Version> addVersion(
...
+   *     </ul>
+   *   </li>
+   * </ul>
+   *
+   * @param store the store being evaluated
+   * @param version the version being evaluated
+   * @return {@code true} if real-time topics should be created; {@code false} otherwise
+   */
+  boolean isRealTimeTopicRequired(Store store, Version version) {
+    if (!store.isHybrid() || !version.isHybrid()) {
```
arjun4084346, line 2995: no problem with this, but can the newly created `version`'s "hybridness" be different from that of its `store` ?

sushantmane: Yes that's possible if store was updated right after version was created

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerClusterConfig.java
sushantmane: We have enabled `PARTICIPANT_MESSAGE_STORE_ENABLED` everywhere in prod. However, it appears we overlooked enabling it in CI. As a result, I noticed some workarounds being used, such as manually creating RT for PARTICIPANT_MESSAGE_STORE_ENABLED to make things work. This change aims to address and fix this inconsistency.

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/PubSubConstants.java
sushantmane: > is there any significance for this number or any negative number would work?
This is sentinel value. And yes any negative number can work

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -3257,45 +3424,180 @@ public Optional<Schema> getReplicationMetadataSchema(
...
+          pushJobId);
+
+      int partitionCount = hybridVersion.getPartitionCount();
+      if (isParent()) {
+        /**
+         * Create RT topic here in parent region. For child regions, RT should always be created as part of
+         * {@link RealTimeTopicSwitcher#ensurePreconditions(PubSubTopic, PubSubTopic, Store, Optional)}
+         */
+        // getRealTimeTopic(clusterName, storeName, partitionCount);
+        System.out.println("getRealTimeTopic");
```
sixpluszero, line 3583: is this debug log?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
```diff
@@ -1730,14 +1717,19 @@ public String getSeparateRealTimeTopic(String clusterName, String storeName) {
...
    */
   @Override
-  public Version getIncrementalPushVersion(String clusterName, String storeName) {
-    Version incrementalPushVersion = getVeniceHelixAdmin().getIncrementalPushVersion(clusterName, storeName);
+  public Version getIncrementalPushVersion(String clusterName, String storeName, String pushJobId) {
+    Version incrementalPushVersion = getVeniceHelixAdmin().getIncrementalPushVersion(clusterName, storeName, pushJobId);
     String incrementalPushTopic = incrementalPushVersion.kafkaTopicName();
     ExecutionStatus status = getOffLinePushStatus(clusterName, incrementalPushTopic).getExecutionStatus();
 
     return getIncrementalPushVersion(incrementalPushVersion, status);
```
sixpluszero, line 1725: Did not read this code path before, but I am confused by the method implementation, as it seems like it requires the retrieved hybrid version to be terminal status, which I don't think is really needed?

services/venice-controller/src/main/java/com/linkedin/venice/controller/Admin.java
```diff
@@ -289,26 +288,9 @@ Version incrementVersionIdempotent(
...
 
-  default String getRealTimeTopic(String clusterName, String storeName) {
-    Store store = getStore(clusterName, storeName);
-    if (store == null) {
-      throw new VeniceNoStoreException(storeName, clusterName);
-    }
-    return getRealTimeTopic(clusterName, store);
-  }
-
-  String getSeparateRealTimeTopic(String clusterName, String storeName);
```
sixpluszero, line 302: It seems like these two APIs are removed. I am a bit lost here, do we move the logic to a certain method that ensures RT is created? Could you point me to that part of the code?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerClusterConfig.java
sushantmane: We've some tests that assume the existence of just one store: regular user store. They perform some operation on that store and then look at the metrics. But if we enable  `PARTICIPANT_MESSAGE_STORE` then the metrics get polluted and the tests either becomes  flaky or fail completely.  In such test I've disable this store with this config and hence we can't delete it.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -3005,6 +2971,172 @@ private Pair<Boolean, Version> addVersion(
...
+
+    // Child controllers always create real-time topics for hybrid stores in their region
+    if (!isParent()) {
+      return true;
+    }
+
+    // Parent controllers create real-time topics in the parent region only under certain conditions
+    return !store.isActiveActiveReplicationEnabled()
+        && (store.getHybridStoreConfig().getDataReplicationPolicy() == DataReplicationPolicy.AGGREGATE
+            || store.isIncrementalPushEnabled());
```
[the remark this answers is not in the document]
sushantmane: We need this for push job details store IMO. Once we migrate that store to AA we can remove this condition

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
sushantmane: yeah. Deleting

services/venice-controller/src/main/java/com/linkedin/venice/controller/Admin.java
sushantmane: We moved RT topic creation logic in `addVersion`. This way we won't have to worry about mismatch between partition count in version topic and RT topic. This also helps us to simplify the logic for RT repartition project 

Code is in `VHA:: createOrUpdateRealTimeTopics`

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -3257,45 +3424,180 @@ public Optional<Schema> getReplicationMetadataSchema(
...
                 + " because the topic: " + rtTopic + " is either absent or being truncated");
       }
-      return version;
+
+      if (hybridVersion.isSeparateRealTimeTopicEnabled()) {
+        PubSubTopic separateRtTopic = pubSubTopicRepository.getTopic(Version.composeSeparateRealTimeTopic(storeName));
+        if (!getTopicManager().containsTopicAndAllPartitionsAreOnline(separateRtTopic)
+            || isTopicTruncated(separateRtTopic.getName())) {
+          LOGGER.error(
+              "Incremental push: {} cannot be started for store: {} in cluster: {} because the topic: {} is either absent or being truncated",
```
sixpluszero, line 3502: Just to clarify that by implementation, inc push is only pushed to one of the above topics based on VPJ config. But I think more checking is always good. nit: can consider wrapping the lines into a small check method and reuse for RT and sepRT.

services/venice-controller/src/main/java/com/linkedin/venice/controller/Admin.java
sixpluszero: Thanks for offline explanation, got it now.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
sushantmane: 've updated check to required version to not be in either `ERROR` or `KILLED` status

internal/venice-common/src/test/java/com/linkedin/venice/controllerapi/RequestTopicForPushRequestTest.java
```diff
@@ -0,0 +1,109 @@
+package com.linkedin.venice.controllerapi;
+
+import static com.linkedin.venice.meta.Version.PushType.BATCH;
+import static org.mockito.Mockito.*;
```
sixpluszero, line 4: nit: * import

sixpluszero (approved):
LGTM. Thank you for the changes!

internal/venice-common/src/test/java/com/linkedin/venice/controllerapi/RequestTopicForPushRequestTest.java
sushantmane: Thanks for catching it. Fixed this and several other places.

sixpluszero (approved):

PR #1339, opened 2024-11-22
[server] [config] Configurable Increased Max Wait After Unsubscribe

KaiSernLim:
## Summary

1. Made configurable the 30 minute maximum wait (introduced in #1213) for the next consumer poll, after unsubscribing from a topic, during a state transition.
2. Improved the log message when reaching the default 10 second maximum wait or if it waits longer than 15 seconds during a state transition.

## How was this PR tested?

GHCI. No change in functionality.


lluwm (approved):
LGTM!

PR #1340, opened 2024-11-22
[compat][common][controller] Add store and version configs for target region swap and wait time

misyel:

Context: Currently with target region push, we push to one colo first then push to the remaining colos. This means that a push job takes 2x the time because we have to wait for the push to the first target colo to succeed first before pushing to the remaining colos concurrently. For phase 2 of target region push, we would like to push to all regions concurrently, but only swap to the new version in the target colo. The target colo will stay on the new version for a specified waitTime and if no errors occur by the end of the waitTiime, we will swap to the new version on the remaining colos. With this approach, push time will not be doubled and it will be the normal push time + waitTime + DVC ingestion time (if enabled)

Change: This pr adds in three store and version configs: `targetSwapRegion`, `targetSwapRegionWaitTime`, and `isDavinciHeartbeatReported` to enable target colo push with delayed version swap

## How was this PR tested?

Existing ci tests


clients/venice-admin-tool/src/main/java/com/linkedin/venice/Arg.java
```diff
@@ -285,7 +285,15 @@ public enum Arg {
       "How many producers will be used to write nearline workload in Server"
   ), INSTANCES("instances", "in", true, "Input list of helix ids of nodes to check if they can removed or not"),
   TO_BE_STOPPED_NODES("to-be-stopped-nodes", "tbsn", true, "List of helix ids of nodes assumed to be stopped"),
-  LAG_FILTER_ENABLED("lag-filter-enabled", "lfe", true, "Enable heartbeat lag filter for a heartbeat request");
+  LAG_FILTER_ENABLED("lag-filter-enabled", "lfe", true, "Enable heartbeat lag filter for a heartbeat request"),
+  TARGET_SWAP_REGION("target-region-swap", "trs", true, "Region to start target colo push with deferred swap in"),
```
majisourav99, line 289: nit: region to swap current version during target colo push.

clients/venice-admin-tool/src/main/java/com/linkedin/venice/Arg.java
```diff
@@ -285,7 +285,15 @@ public enum Arg {
...
+  TARGET_SWAP_REGION_WAIT_TIME(
+      "target-region-swap-wait-time", "trswt", true,
+      "How long to wait before swapping to the new version in a target colo push"
+  ),
+  DAVINCI_HEARTBEAT_REPORTED(
+      "dvc-heartbeat-reported", "dvchb", true, "Flag to indicate whether DVC is enabled and healthy"
```
majisourav99, line 295: dont think we going to check `healthy` using this config. It will indicate if there are any dvc instances are bootstrapping and sending hb

majisourav99 (commented):
Add a `[compat]` to the PR title to indicate there has been a protocol update and we need special ordering of deployment.

majisourav99 (approved):
LGTM. Thanks Michelle

PR #1342, opened 2024-11-23
[server] Validate PubSub address against cluster map in AASIT::consumerSubscribe

sushantmane:
## Validate PubSub address against cluster map in AASIT::consumerSubscribe
Ensure the PubSub URL is present in the PubSub cluster map before  
subscribing to a topic. This prevents issues caused by attempting to  
subscribe to unknown PubSub URLs during message consumption.  

Example case:
If the cluster map contains entries like `[a: 0, b: 1]` and the TopicSwitch (TS) message references address c, the subscription check ensures c is present in the cluster map. If `c` is not found, the subscription fails, and ingestion errors out. This prevents consuming any records from address `c`, and hence no records with a` -1` upstream offset are produced to VT.

## How was this PR tested?

CI, UT, and local runs


xunyin8 (approved):

sushantmane:
Thank you, @xunyin8!

PR #1350, opened 2024-11-26
Increase heap size to 8G for CI reporting

majisourav99:
## Increase heap size to 8G for CI reporting

Resolves   FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory

## How was this PR tested?

GHCI

### DIFF
--- a/.github/workflows/UnitTests-core.yml
+++ b/.github/workflows/UnitTests-core.yml
@@ -51,7 +51,7 @@ jobs:
           tar -zcvf ${{ inputs.artifact_suffix }}-jdk${{ matrix.jdk }}-logs.tar.gz ${{ inputs.artifact_suffix }}-artifacts
       - name: Publish Test Report
         env:
-          NODE_OPTIONS: "--max_old_space_size=4096"
+          NODE_OPTIONS: "--max_old_space_size=8192"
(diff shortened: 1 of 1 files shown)


.github/workflows/UnitTests-core.yml
```diff
@@ -51,7 +51,7 @@ jobs:
           tar -zcvf ${{ inputs.artifact_suffix }}-jdk${{ matrix.jdk }}-logs.tar.gz ${{ inputs.artifact_suffix }}-artifacts
       - name: Publish Test Report
         env:
-          NODE_OPTIONS: "--max_old_space_size=4096"
+          NODE_OPTIONS: "--max_old_space_size=8192"
```
FelixGV, line 54: Node? So it's a nodejs thing? I have to update the "other languages: section of my recent docs update then 😅

FelixGV (approved):
Thanks Sourav!

.github/workflows/UnitTests-core.yml
majisourav99: yup, these reporters uses node.js. GH can infer that no? currently it shows

Java 99.6%
Other 0.4%

PR #1351, opened 2024-11-27
[router] Fix invalid tehuti sensor name passed in for latencies

m-nagarajan:
## Summary
If an invalid `TehutiMetricNameEnum` that was not registered with `MetricEntityState` is passed in, it will be skipped during record() and will be silently ignored. Found this issue for `healthy_request_latency` and this PR fixes this issue and adds logging with redundant logging filter to log such issues. Not throwing an exception as we will be moving a lot of the existing metrics to Otel and tracking this log will help find and fix the issue during the migration and not affect the rollout process.

## How was this PR tested?
unit test


FelixGV (commented):
I see only a logging change, and no mention of `healthy_request_latency`. Is this expected?

FelixGV (approved):

m-nagarajan:
> I see only a logging change, and no mention of `healthy_request_latency`. Is this expected?
I missed to stage that change. Thanks for catching.

PR #1352, opened 2024-11-27
[server][dvc] Add Blob Transfer Related Metrics

jingy-li:
## [server][dvc] Add Blob Transfer Related Metrics

This PR introduces 5 metrics to track blob transfer activity as a client side:

As a client side:
- **blob_transfer_total_num_responses**: The total number of responses received.
- **blob_transfer_successful_num_responses**: The number of successful responses received, which also indicates how many partitions were successfully bootstrapped via blob transfer.
- **blob_transfer_failed_num_responses**: The number of failed responses, which also indicates how many partitions failed to bootstrap via blob transfer.
- **blob_transfer_file_receive_throughput**: The file receive rate, measured in MB/sec.
- **blob_transfer_time**: The total file transfer time per partition, measured in seconds.

## How was this PR tested?
Testing with private build at testing hosts.
 


clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/BlobTransferUtils.java
```diff
@@ -46,4 +60,55 @@ public static String generateFileChecksum(Path filePath) throws IOException {
...
+   */
+  private static double calculateThroughputInMBPerSec(File partitionDir, double transferTimeInSec) throws IOException {
+    if (!partitionDir.exists() || !partitionDir.isDirectory()) {
+      throw new IllegalArgumentException(
+          "Partition directory does not exist or is not a directory: " + partitionDir.getAbsolutePath());
+    }
+    // Calculate total size of all files in the directory
+    long totalSizeInBytes = getTotalSizeOfFiles(partitionDir);
+    // Convert bytes to MB
+    double totalSizeInMB = totalSizeInBytes / (1024.0 * 1024.0);
```
sixpluszero, line 75: For file size calculation, it should be by 1000 instead of 1024

sixpluszero (approved):
LGTM, thank you!

PR #1355, opened 2024-12-02
Revert "[controller] do not delay RT topic deletion if the store is n…

arjun4084346:
…o longer hybrid (#1234)"

This reverts commit af40771ae446b5bf4e855a4d6e21d395fe6d2355.

## revert changes to optimize TopicCleanupService 

Reverting the changes https://github.com/linkedin/venice/pull/1234 because in this change, the host that TCS runs is trying to get `Store` for which it is not the leader controller.

Resolves #XXX


huangminchn (approved):
Thanks Arjun!

PR #1358, opened 2024-12-02
[controller] do not delay RT topic deletion if the store is no longer hybrid

arjun4084346:
## Avoid `getStore` from the controller which is not the leader of cluster this store is a part of 

When TopicCleanupService tries to delete RT topics, it does not have to wait for all the corresponding VT topics to get deleted if all the versions of the store has changed from hybrid to batch. If all the versions are batch, it can be safely assumed that no VT topic is consuming from that RT topic.
This was already done in https://github.com/linkedin/venice/pull/1234 but that commit was reverted in https://github.com/linkedin/venice/pull/1355
It was reverted because TCS was trying to get `Store` for which it was not the leader controller. Only leader controller caches and stores the information for all the stores in the cluster it is leader of. 

So this time, `isRTTopicDeletionPermittedByAllControllers` will not take `Store` object, it will use the ControllerClientMap to get Store information for all the stores. 

Resolves #XXX

### DIFF
--- a/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
@@ -3517,7 +3517,7 @@ public class VeniceHelixAdmin implements Admin, StoreCleaner {
       PubSubTopic rtTopic = pubSubTopicRepository.getTopic(Version.composeRealTimeTopic(storeName));
       if (!store.isHybrid() && getTopicManager().containsTopic(rtTopic)) {
         store = resources.getStoreMetadataRepository().getStore(storeName);
-        safeDeleteRTTopic(clusterName, storeName, store);
+        safeDeleteRTTopic(clusterName, store);
       }
     }
   }
@@ -3532,17 +3532,29 @@ public class VeniceHelixAdmin implements Admin, StoreCleaner {
     }
   }
 
-  private void safeDeleteRTTopic(String clusterName, String storeName, Store store) {
+  private void safeDeleteRTTopic(String clusterName, Store store) {
+    if (isRTTopicDeletionPermittedByAllControllers(clusterName, store)) {
+      deleteRTTopic(clusterName, store.getName());
+    }
+  }
+
+  public boolean isRTTopicDeletionPermittedByAllControllers(String clusterName, Store store) {
     // Perform RT cleanup checks for batch only store that used to be hybrid. Check the local versions first
     // to see if any version is still using RT and then also check other fabrics before deleting the RT. Since
     // we perform this check everytime when a store version is deleted we can afford to do best effort
     // approach if some fabrics are unavailable or out of sync (temporarily).
-    boolean canDeleteRT = !Version.containsHybridVersion(store.getVersions());
+    String storeName = store.getName();
+    String rtTopicName = Version.composeRealTimeTopic(storeName);
+    if (Version.containsHybridVersion(store.getVersions())) {
+      LOGGER.warn(
+          "Topic {} cannot be deleted yet because the store {} has at least one hybrid version",
+          rtTopicName,
+          storeName);
+      return false;
+    }
+
     Map<String, ControllerClient> controllerClientMap = getControllerClientMap(clusterName);
     for (Map.Entry<String, ControllerClient> controllerClientEntry: controllerClientMap.entrySet()) {
-      if (!canDeleteRT) {
-        return;
-      }
       StoreResponse storeResponse = controllerClientEntry.getValue().getStore(storeName);
       if (storeResponse.isError()) {
         LOGGER.warn(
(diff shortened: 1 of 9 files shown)


services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -3551,24 +3554,37 @@ private void safeDeleteRTTopic(String clusterName, String storeName, Store store
...
+  }
+
+  private void deleteRealTtimeTopic(String clusterName, String storeName) {
+    Map<String, ControllerClient> controllerClientMap = getControllerClientMap(clusterName);
+    String rtTopicToDelete = Version.composeRealTimeTopic(storeName);
+    deleteRTTopicFromAllFabrics(rtTopicToDelete, controllerClientMap);
+    // Check if there is incremental push topic exist. If yes, delete it and send out to let other controller to
+    // delete it.
+    String incrementalPushRTTopicToDelete = Version.composeSeparateRealTimeTopic(storeName);
+    if (getTopicManager().containsTopic(pubSubTopicRepository.getTopic(incrementalPushRTTopicToDelete))) {
```
xunyin8, line 3577: Can we separate or extract the deletion of incremental push RT from RT? The triggers in this private method is a bit hard to follow and could be error prone (e.g. RT could be gone but incremental push RT is still around due to previous deletion failures).
1. The check for the existence of RT or need for RT deletion is outside of this private method.
2. The check for incremental push RT is only invoked in this private method.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -3551,24 +3554,37 @@ private void safeDeleteRTTopic(String clusterName, String storeName, Store store
...
+      deleteRTTopicFromAllFabrics(incrementalPushRTTopicToDelete, controllerClientMap);
+    }
+  }
+
+  private void deleteRTTopicFromAllFabrics(String topic, Map<String, ControllerClient> controllerClientMap) {
```
xunyin8, line 3582: Is the goal here just trying to expedite the RT topic deletion in all fabrics?
The trigger for `deleteRTTopicFromAllFabrics` is only checking the local state. i.e. `!store.isHybrid() && getTopicManager().containsTopic(rtTopic)`.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -3551,24 +3554,37 @@ private void safeDeleteRTTopic(String clusterName, String storeName, Store store
...
+    truncateKafkaTopic(topic);
```
xunyin8, line 3583: Did you verify if `controllerClientMap` excludes the local fabric? i.e. we might be making unnecessary duplicate calls to `truncateKafkaTopic`.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
arjun4084346: The goal is just to refactor this part and extract out method `isRTTopicDeletionPermittedByAllControllers` , so that `TopicCleanupService` can also call this method.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
arjun4084346: I don't think local fabric is excluded from `controllerClientMap`. removing the extra call

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
arjun4084346: refactored.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
arjun4084346: However, it was being done like that on left side too. I basically do not intend to change the code logic of this part, I am only trying to refactor this part to extract out `isRTTopicDeletionPermittedByAllControllers`

arjun4084346: I put it back because a test `testHybridStoreToBatchOnly` was failing in `Assert.assertTrue(veniceAdmin.isTopicTruncated(rtTopic))`. It was not truncating the topic because `multiClusterConfigs.getControllerConfig(clusterName).getChildDataCenterControllerUrlMap()` was empty

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -3532,17 +3531,27 @@ private boolean hasFatalDataValidationError(PushMonitor pushMonitor, String topi
...
-    // we perform this check everytime when a store version is deleted we can afford to do best effort
+  private void safeDeleteRTTopic(String clusterName, String storeName) {
+    if (isRTTopicDeletionPermittedByAllControllers(clusterName, storeName)) {
+      String rtTopicToDelete = Version.composeRealTimeTopic(storeName);
+      deleteRTTopicFromAllFabrics(rtTopicToDelete, clusterName);
+    }
+    // Check if there is incremental push topic exist. If yes, delete it and send out to let other controller to
+    // delete it.
+    String incrementalPushRTTopicToDelete = Version.composeSeparateRealTimeTopic(storeName);
+    if (getTopicManager().containsTopic(pubSubTopicRepository.getTopic(incrementalPushRTTopicToDelete))) {
```
xunyin8, line 3542: The semantic of this refactor is actually different now. I think what you actually want here is save the result of `isRTTopicDeletionPermittedByAllControllers` to a boolean like `rtDeletionPermitted` and then here you want to have both condition to be true before deleting, i.e.
```
if (rtDeletionPermitted && getTopicManager().containsTopic(pubSubTopicRepository.getTopic(incrementalPushRTTopicToDelete)))
```
Otherwise it's not a safe delete

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
xunyin8: See new comment.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
xunyin8: Ok, just wanted to ask because it's possible for each fabric to receive multiple requests to truncate topic. It should be harmless with the way we implemented the truncate API.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
arjun4084346: got it

xunyin8 (approved):

PR #1359, opened 2024-12-02
[da-vinci][server] Added flag to enable RocksDB blob file for blocked based format

gaojieliu:
This PR adds options to enable BlobDB feature in RocksDB to reduce write amplification for large value stores.
Currently, these options are only valid for block-based format.

New configs:
rocksdb.blob.files.enabled : default false
rocksdb.min.blob.size.in.bytes : default 4KB
rocksdb.blob.file.size.in.bytes : default 256MB
rocksdb.blob.garbage.collection.age.cutoff : default 0.25
rocksdb.blob.garbage.collection.force.threshold : default 0.8
rocksdb.blob.file.starting.level: default 0

This feature is not perfect, as the main goal is to reduce the write amplification, but space amplification will increase and we can tune the above params to find the right balance.

## How was this PR tested?
CI

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBStoragePartition.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBStoragePartition.java
@@ -412,6 +412,7 @@ public class RocksDBStoragePartition extends AbstractStoragePartition {
         options.setBlobFileSize(rocksDBServerConfig.getBlobFileSizeInBytes());
         options.setBlobGarbageCollectionAgeCutoff(rocksDBServerConfig.getBlobGarbageCollectionAgeCutOff());
         options.setBlobGarbageCollectionForceThreshold(rocksDBServerConfig.getBlobGarbageCollectionForceThreshold());
+        options.setBlobFileStartingLevel(rocksDBServerConfig.getBlobFileStartingLevel());
       }
     }
 
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBStoragePartition.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBStoragePartition.java
@@ -401,6 +401,18 @@ public class RocksDBStoragePartition extends AbstractStoragePartition {
       tableConfig.setCacheIndexAndFilterBlocks(rocksDBServerConfig.isRocksDBSetCacheIndexAndFilterBlocks());
       tableConfig.setFormatVersion(rocksDBServerConfig.getBlockBaseFormatVersion());
       options.setTableFormatConfig(tableConfig);
+
+      /**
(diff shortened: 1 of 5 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBStoragePartition.java
```diff
@@ -401,6 +401,18 @@ protected Options getStoreOptions(StoragePartitionConfig storagePartitionConfig,
       tableConfig.setCacheIndexAndFilterBlocks(rocksDBServerConfig.isRocksDBSetCacheIndexAndFilterBlocks());
       tableConfig.setFormatVersion(rocksDBServerConfig.getBlockBaseFormatVersion());
       options.setTableFormatConfig(tableConfig);
+
+      /**
+       * Only enable blob files for block-based format.
+       */
+      if (rocksDBServerConfig.isBlobFilesEnabled()) {
+        options.setEnableBlobFiles(true);
+        options.setEnableBlobGarbageCollection(true);
```
FelixGV, line 410: Is this GC useful for batch-only stores? Should it be enabled only post-EOP instead? If the overhead is null or negligible on a batch-only store then no problem in leaving it always on...

gaojieliu: RocksDB SST file writer doesn't respect `blob file` option as all the input data are ordered, so there is no compaction.
That means this option doesn't hurt for batch-only stores.

I am planning to add some metrics and tests to validate this new mode.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestHybrid.java
```diff
@@ -1716,6 +1717,8 @@ private static VeniceClusterWrapper setUpCluster(boolean enablePartitionWiseShar
           KafkaConsumerService.ConsumerAssignmentStrategy.PARTITION_WISE_SHARED_CONSUMER_ASSIGNMENT_STRATEGY.name());
     }
     cluster.addVeniceServer(new Properties(), serverProperties);
+    // Enable blob files in one server
+    serverProperties.setProperty(ROCKSDB_BLOB_FILES_ENABLED, "true");
```
majisourav99, line 1721: May be we could add a JMH benchmark test for blob-file mode.

majisourav99 (approved):
Thanks, LGTM.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestHybrid.java
gaojieliu: It is hard to test it properly as it is mainly about write amplification.
I would prefer to validate it in a perf test env.

majisourav99 (approved):

PR #1360, opened 2024-12-03
[controller] Fix hybrid store migration failure during ongoing batch push jobs

sushantmane:
## Fix hybrid store migration failure during ongoing batch push jobs

This PR fixes an issue where hybrid store migration fails if a batch push job is ongoing
and a non-truncated version topic exists in the parent region.

The problem stemmed from a safeguard intended to prevent hybrid-to-batch conversion during
ongoing push jobs (added to address targeted region push issues). The safeguard failed to
account for store migration scenarios where hybrid configurations are added during the
updateStore operation.

This misclassification of the migration attempt as a batch-to-hybrid store conversion
caused unintended operations, including a call to VPHA::getTopicForCurrentPushJob. This
method attempts to issue a killOfflinePushJob when the version config is missing in the
store’s configuration. The kill job, in turn, tries to update the version status to
KILLED, which does not exist, which leads to a version not found exception from
AbstractStore::updateVersionStatus.

Ref: https://github.com/linkedin/venice/pull/1143  

### DIFF
--- a/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
@@ -2301,7 +2301,9 @@ public class VeniceParentHelixAdmin implements Admin {
       setStore.storeName = storeName;
       setStore.owner = owner.map(addToUpdatedConfigList(updatedConfigsList, OWNER)).orElseGet(currStore::getOwner);
 
-      if (!currStore.isHybrid() && (hybridRewindSeconds.isPresent() || hybridOffsetLagThreshold.isPresent())) {
+      boolean isUpdateForStoreMigration = storeMigration.isPresent() && storeMigration.get();
+      if (!isUpdateForStoreMigration && !currStore.isHybrid()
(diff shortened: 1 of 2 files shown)


services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
```diff
@@ -2301,7 +2301,9 @@ public void updateStore(String clusterName, String storeName, UpdateStoreQueryPa
       setStore.storeName = storeName;
       setStore.owner = owner.map(addToUpdatedConfigList(updatedConfigsList, OWNER)).orElseGet(currStore::getOwner);
 
-      if (!currStore.isHybrid() && (hybridRewindSeconds.isPresent() || hybridOffsetLagThreshold.isPresent())) {
+      boolean isUpdateForStoreMigration = storeMigration.isPresent() && storeMigration.get();
```
majisourav99, line 2304: nit: can we use `boolean isUpdateForStoreMigration = storeMigration.orElse(false);`

sushantmane: Done

majisourav99 (approved):
LGTM. Thanks for the fix!

sushantmane:
Thank you, @smaji for the quick review!

PR #1364, opened 2024-12-04
[vpj][admin-tool] Remove subscription from consumer after dictionary download in DictionaryUtils::readDictionaryFromKafka

sushantmane:
## Remove subscription from consumer after dictionary download in DictionaryUtils::readDictionaryFromKafka

This change ensures that the subscription is cleaned up from the PubSub consumer once the
dictionary is downloaded in DictionaryUtils::readDictionaryFromKafka. Without this
cleanup, subsequent attempts to resubscribe to the same topic partition are ignored which
can have side effects.

This issue particularly impacts the admin tool, as it uses the same consumer for
dictionary downloads and then tries to subscribe to the same partition for dumping
messages. The resubscription request is rejected because the consumer already has an
active subscription from the dictionary fetch operation. Consequently, the dumper cannot
seek to the provided offset if the partition being dumped is 0 (the same partition used
during dictionary download).

By removing the subscription after the dictionary is fetched, this fix ensures that
subsequent operations, such as seek and subscribe for message dumping, work as expected.

## How was this PR tested?

UT and CI


nisargthakkar (approved):

sushantmane:
Thanks, @nisargthakkar!

PR #1366, opened 2024-12-04
[admin-tool] Add support for dumping messages within a specified time range

sushantmane:
## [admin-tool] Add support for dumping messages within a specified time range
This commit adds support to the admin tool for dumping messages from a PubSub topic within a specified time range, using start and end date-time in the format "yyyy-MM-dd HH:mm:ss" (PST). Previously, the admin tool could not consume messages based on time ranges. This code change allows targeted message dumping to streamline troubleshooting.

## How was this PR tested?

CI and PROD 

### DIFF
--- a/clients/venice-admin-tool/src/main/java/com/linkedin/venice/Arg.java
+++ b/clients/venice-admin-tool/src/main/java/com/linkedin/venice/Arg.java
@@ -118,7 +118,10 @@ public enum Arg {
   VENICE_CLIENT_SSL_CONFIG_FILE(
       "venice-client-ssl-config-file", "vcsc", true, "Configuration file for querying key in Venice client through SSL."
   ), STARTING_OFFSET("starting_offset", "so", true, "Starting offset when dumping admin messages, inclusive"),
-  MESSAGE_COUNT("message_count", "mc", true, "Max message count when dumping admin messages"),
+  STARTING_TIMESTAMP(
+      "starting_timestamp", "sts", true,
+      "Dump messages from a topic with a timestamp (in long epoch format) greater than the specified value."
+  ), MESSAGE_COUNT("message_count", "mc", true, "Max message count when dumping admin messages"),
   PARENT_DIRECTORY(
       "parent_output_directory", "pod", true,
       "A directory where output can be dumped to.  If dumping a kafka topic, the output will be dumped under this directory."
--- a/clients/venice-admin-tool/src/main/java/com/linkedin/venice/Arg.java
+++ b/clients/venice-admin-tool/src/main/java/com/linkedin/venice/Arg.java
@@ -118,10 +118,7 @@ public enum Arg {
   VENICE_CLIENT_SSL_CONFIG_FILE(
       "venice-client-ssl-config-file", "vcsc", true, "Configuration file for querying key in Venice client through SSL."
   ), STARTING_OFFSET("starting_offset", "so", true, "Starting offset when dumping admin messages, inclusive"),
-  STARTING_TIMESTAMP(
-      "starting_timestamp", "sts", true,
-      "Dump messages from a topic with a timestamp (in long epoch format) greater than the specified value."
-  ), MESSAGE_COUNT("message_count", "mc", true, "Max message count when dumping admin messages"),
+  MESSAGE_COUNT("message_count", "mc", true, "Max message count when dumping admin messages"),
   PARENT_DIRECTORY(
       "parent_output_directory", "pod", true,
       "A directory where output can be dumped to.  If dumping a kafka topic, the output will be dumped under this directory."
--- a/clients/venice-admin-tool/src/main/java/com/linkedin/venice/KafkaTopicDumper.java
+++ b/clients/venice-admin-tool/src/main/java/com/linkedin/venice/KafkaTopicDumper.java
@@ -36,6 +36,7 @@ import com.linkedin.venice.pubsub.PubSubTopicRepository;
 import com.linkedin.venice.pubsub.api.PubSubConsumerAdapter;
 import com.linkedin.venice.pubsub.api.PubSubMessage;
 import com.linkedin.venice.pubsub.api.PubSubTopicPartition;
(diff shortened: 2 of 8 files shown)


clients/venice-admin-tool/src/main/java/com/linkedin/venice/Arg.java
```diff
@@ -118,7 +118,10 @@ public enum Arg {
   VENICE_CLIENT_SSL_CONFIG_FILE(
       "venice-client-ssl-config-file", "vcsc", true, "Configuration file for querying key in Venice client through SSL."
   ), STARTING_OFFSET("starting_offset", "so", true, "Starting offset when dumping admin messages, inclusive"),
-  MESSAGE_COUNT("message_count", "mc", true, "Max message count when dumping admin messages"),
+  STARTING_TIMESTAMP(
```
nisargthakkar, line 121: Can we use `START_DATE` arg?
```  START_DATE("start-date", "sd", true, "Start date in PST. Example: 2020-10-10 10:10:10"),```

Comment (sushantmane) [reply clients/venice-admin-tool/src/main/java/com/linkedin/venice/Arg.java:121]: Updated to use `START_DATE` and `END_DATE` with topic dump

File: clients/venice-admin-tool/src/main/java/com/linkedin/venice/KafkaTopicDumper.java [hunk 2]
```diff
@@ -234,66 +226,200 @@ public KafkaTopicDumper(
...
+            Utils.iterateOnMapOfLists(records);
+        boolean hasProcessedRecords = false;
+        while (recordIterator.hasNext() && processedMessageCount < messageCount) {
+          PubSubMessage<KafkaKey, KafkaMessageEnvelope, Long> record = recordIterator.next();
+          processedMessageCount++;
+          lastProcessedRecord = record;
+          hasProcessedRecords = true;
+          processRecord(record);
+          // Exit early if endOffset is reached
+          if (record.getOffset() >= endOffset) {
```
nisargthakkar, line 390: Can we move this block before `processRecord`, and just do `record.getOffset() > endOffset`? That way we're stricter about what we dump

clients/venice-admin-tool/src/main/java/com/linkedin/venice/KafkaTopicDumper.java
```diff
@@ -234,66 +226,200 @@ public KafkaTopicDumper(
...
+            LOGGER.info("Reached endOffset. Total messages processed: {}", processedMessageCount);
+            return processedMessageCount;
```
nisargthakkar, line 392: Can we just break out of the loop here?

clients/venice-admin-tool/src/main/java/com/linkedin/venice/KafkaTopicDumper.java
```diff
@@ -234,66 +226,200 @@ public KafkaTopicDumper(
...
+   *                          If set to -1, the method uses the provided {@code startingOffset}.
+   * @return the calculated starting offset for the partition
+   * @throws PubSubClientException if no offset is found for the provided timestamp
+   */
+  static long calculateStartingOffset(
+      PubSubConsumerAdapter consumer,
+      PubSubTopicPartition partition,
+      long startingOffset,
+      long startingTimestamp) {
+    if (startingTimestamp != -1) {
```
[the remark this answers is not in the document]
sushantmane: I intentionally kept this behavior because sometimes the startDate can overshoot the endOffset, and in such cases, the fallback would be the provided startingOffset. I personally prefer this approach, but if you think it might cause confusion, I can update it.

clients/venice-admin-tool/src/main/java/com/linkedin/venice/KafkaTopicDumper.java
sushantmane: We can but since this is nested loop we'll have to repeat the condition in both inner and outer loop conditions

clients/venice-admin-tool/src/main/java/com/linkedin/venice/KafkaTopicDumper.java
sushantmane: > That way we're stricter about what we dump
I don't follow this part fully. Is current logic not enforcing this strictly?

clients/venice-admin-tool/src/main/java/com/linkedin/venice/KafkaTopicDumper.java
nisargthakkar: Oh, didn't realize it's a nested loop

clients/venice-admin-tool/src/main/java/com/linkedin/venice/KafkaTopicDumper.java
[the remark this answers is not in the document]
nisargthakkar: Yeah, it's confusing

clients/venice-admin-tool/src/main/java/com/linkedin/venice/KafkaTopicDumper.java
nisargthakkar: There is an off by one error.

Say there are these offsets: 1, 2, 5, 6

Now if I put `endOffset` as 4 - the logic will also output offset 5 because it will process 1 and 2 and go to the next record which skips to offset 5, then it'll process offset 5, and then detect `record.getOffset() >= endOffset`

sushantmane: Good catch. I forgot about gaps in the offsets. Thanks

clients/venice-admin-tool/src/main/java/com/linkedin/venice/KafkaTopicDumper.java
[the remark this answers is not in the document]
sushantmane: Done

sushantmane:
Thanks, @nisargthakkar!

PR #1367, opened 2024-12-04
[router] Reduced allocations in router read path

FelixGV:
A lot of multi-key specific state was moved out of VenicePath and into its subclasses, which makes single get queries require less heap space. This includes streaming (chunkedResponse), HAR (helixGroupId and requestId), longTailRetryThresholdMs and responseHeaders.

In addition, all kinds of VenicePaths are also getting their heap size reduced by leveraging shared instances where possible. For example:

- The resourceName, storeName and versionNumber properties are replaced by a single reference to a shared instance of StoreVersionName, which is a new class obtained from a NameRepository (same idea as the PubSubTopicRepository).

- The smartLongTailRetryEnabled, smartLongTailRetryAbortThresholdMs and longTailRetryThresholdMs properties are replaced by a single pointer to a RouterRetryConfig object, which is a simple facade wrapping the VeniceRouterConfig.

- The responseDecompressor is now coming from a map of shared instances in the VenicePathParser.

Miscellaneous:

- Various refactorings enable the VenicePath constructors to have fewer params and to make more of the properties final.

- Made as many properties as possible final in the VeniceRouterConfig.

- Minor tweaks to the main README page.

## How was this PR tested?

CI.


services/venice-router/src/main/java/com/linkedin/venice/router/api/RouterExceptionAndTrackingUtils.java
```diff
@@ -23,14 +23,25 @@
...
+    SMART_RETRY_ABORTED_BY_DELAY_CONSTRAINT(AggRouterHttpRequestStats::recordDelayConstraintAbortedRetryRequest),
+    SMART_RETRY_ABORTED_BY_MAX_RETRY_ROUTE_LIMIT(AggRouterHttpRequestStats::recordRetryRouteLimitAbortedRetryRequest),
+    RESOURCE_NOT_FOUND(null),
+    RETRY_ABORTED_BY_NO_AVAILABLE_REPLICA(AggRouterHttpRequestStats::recordNoAvailableReplicaAbortedRetryRequest);
+
+    final boolean reportUnhealthy;
+    final BiConsumer<AggRouterHttpRequestStats, String> statsRecorder;
+
+    FailureType(BiConsumer<AggRouterHttpRequestStats, String> statsRecorder) {
+      this.reportUnhealthy = statsRecorder == null;
```
lusong64, line 40: So the entire reportUnhealthy is used to recorded whether we have a non-null BiConsumer? Would it possible to just use 

```java
this.reportUnhealthy = Optional.ofNullable(statsRecorder)
    .orElse((stats, store) -> {});
```

FelixGV: The `reportUnhealthy` is not only used within the constructor, it is also used inside the of `metricTracking` function to control some behavior.

Regarding the use of `Optional`, I personally do not find that it improves readability, so not a big fan... but that can be a matter of opinion 🤷

lusong64: I see.

FelixGV:
Carved out a portion of this big-ish PR into this other fairly trivial PR: #1405.

FelixGV:
Rebased after merging https://github.com/linkedin/venice/pull/1405 to reduce the PR size, but also added a flaky test fix which still increases it a bit more 😅 ... oh well 🤷

FelixGV:
Will be able to further reduce the size of this PR after rebasing on top of #1413.

FelixGV:
Rebased

services/venice-router/src/main/java/com/linkedin/venice/router/api/VenicePartitionFinder.java
```diff
@@ -106,15 +110,17 @@ private VenicePartitioner searchPartitioner(String storeName, int versionNum) {
...
-        Map<Integer, VenicePartitioner> versionByPartitionerMap = storeByVersionByPartitionerMap.get(storeName);
-        for (Integer candidateVersion: versionByPartitionerMap.keySet()) {
+      Map<Integer, VenicePartitioner> versionByPartitionerMap = storeByVersionByPartitionerMap.get(storeName);
+      if (versionByPartitionerMap != null) {
+        Iterator<Integer> candidateVersionIterator = versionByPartitionerMap.keySet().iterator();
+        Integer candidateVersion;
+        while (candidateVersionIterator.hasNext()) {
+          candidateVersion = candidateVersionIterator.next();
           if (!upToDateVersionsSet.contains(candidateVersion)) {
-            versionByPartitionerMap.remove(candidateVersion);
```
ZacAttack, line 117: Do we not still need to prune this?

ZacAttack: er wait, nevermind I think?  The candidateVersionIterator is based on this.  So I guess removing it from the iterator has the same effect?

services/venice-router/src/main/java/com/linkedin/venice/router/api/path/VeniceSingleGetPath.java
```diff
@@ -32,18 +35,17 @@ public class VeniceSingleGetPath extends VenicePath {
       Integer.toString(ReadAvroProtocolDefinition.SINGLE_GET_ROUTER_REQUEST_V1.getProtocolVersion());
 
   private final RouterKey routerKey;
-  private final String partition;
 
   public VeniceSingleGetPath(
-      String storeName,
-      int versionNumber,
-      String resourceName,
+      StoreVersionName storeVersionName,
```
ZacAttack, line 40: This is a very satisfying compaction of arguments.

storeName, versionNumber, resourceName -> storeVersionName

services/venice-router/src/main/java/com/linkedin/venice/router/api/path/VenicePath.java
```diff
@@ -54,69 +47,35 @@ public abstract class VenicePath implements ResourcePath<RouterKey> {
...
-    if (requestId < 0) {
-      requestId = REQUEST_ID_GENERATOR.getAndIncrement();
-    }
-    return requestId;
+    return -1;
   }
 
   public int getHelixGroupId() {
-    return helixGroupId;
+    return -1;
```
ZacAttack, line 66: This is kind of neat.  So we're reducing allocations by removing the member variables on the super class?  At the expense of having the child classes implement these?

I'm wondering about some of these though, the requestId one, do child classes need their own version of the requestId member?  Could we just rely on the super and a default function for getRequestId?  Or is there some cost saving that I'm missing?

ZacAttack: Or... maybe we don't actually use request id?

services/venice-router/src/main/java/com/linkedin/venice/router/api/VenicePartitionFinder.java
FelixGV: That's right, and from my understanding it is also safer to remove from the iterator...

services/venice-router/src/main/java/com/linkedin/venice/router/api/path/VeniceSingleGetPath.java
FelixGV: One of those cases where doing the efficient thing is also cleaner (more brief, strongly-typed) 😁

services/venice-router/src/main/java/com/linkedin/venice/router/api/path/VenicePath.java
FelixGV: The request ID is only uses for HAR requests I believe.

services/venice-router/src/main/java/com/linkedin/venice/router/api/VenicePathParser.java
```diff
@@ -294,26 +319,6 @@ public VenicePath parseResourceUri(String uri, HTTP_REQUEST request) throws Rout
         }
       }
 
-      boolean decompressOnClient = routerConfig.isDecompressOnClient();
```
ZacAttack, line 297: How do we dodge needing this?  Is it because this decision is delegated to the responseDecompressor?

services/venice-router/src/main/java/com/linkedin/venice/router/api/VeniceResponseDecompressor.java
```diff
@@ -42,30 +39,20 @@ public class VeniceResponseDecompressor {
...
   private final RouterStats<AggRouterHttpRequestStats> routerStats;
-  private final String storeName;
-  private final int version;
-  private final String kafkaTopic;
+  private final StoreVersionName storeVersionName;
   private final CompressorFactory compressorFactory;
 
   public VeniceResponseDecompressor(
-      boolean decompressOnClient,
+      CompressionStrategy clientCompression,
```
ZacAttack, line 46: Do we even need this?  clientCompression is a private final and I don't see it used outside this classes constructor.

ZacAttack (approved):
Looks good, dropped a few questions here and there, but I didn't see anything untoward.  Thanks for the clean up!

services/venice-router/src/main/java/com/linkedin/venice/router/api/path/VenicePath.java
ZacAttack: I realize the comment line is misplaced.  I mean a few lines up on the getRequestId() function

services/venice-router/src/main/java/com/linkedin/venice/router/api/VenicePathParser.java
FelixGV: This code is moved to a separate function. See line 401.

services/venice-router/src/main/java/com/linkedin/venice/router/api/path/VenicePath.java
FelixGV: Yeah I thought so. Both the Helix group ID and request ID are used for HAR, IIUC.

FelixGV:
Thanks for the review!

PR #1368, opened 2024-12-04
[changelog] Add resiliency to version swap in change capture client

ZacAttack:
## [changelog] Add resiliency to version swap in change capture client

This improves resiliency in a layered approach.  It:

- Makes the metadata update a change listener on the store repository 
- Makes the original detection thread a last ditch effort should something go wrong when reacting to that call back 
- Makes the unknown topic or partition exception a signal to try and force a refresh of metadata and fix the subscription
- Actually uses the correct executor task method so that the refresh is properly scheduled (as opposed to the original one which only did it once).

Resolves #XXX

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceAfterImageConsumerImpl.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceAfterImageConsumerImpl.java
@@ -7,9 +7,6 @@ import com.linkedin.venice.kafka.protocol.ControlMessage;
 import com.linkedin.venice.kafka.protocol.KafkaMessageEnvelope;
 import com.linkedin.venice.kafka.protocol.enums.ControlMessageType;
 import com.linkedin.venice.message.KafkaKey;
-import com.linkedin.venice.meta.Store;
-import com.linkedin.venice.meta.StoreDataChangedListener;
-import com.linkedin.venice.meta.Version;
 import com.linkedin.venice.pubsub.adapter.kafka.ApacheKafkaOffsetPosition;
 import com.linkedin.venice.pubsub.api.PubSubConsumerAdapter;
 import com.linkedin.venice.pubsub.api.PubSubMessage;
@@ -42,7 +39,7 @@ public class VeniceAfterImageConsumerImpl<K, V> extends VeniceChangelogConsumerI
   final private Lazy<VeniceChangelogConsumerImpl<K, V>> internalSeekConsumer;
   private final ScheduledExecutorService versionSwapExecutorService = Executors.newSingleThreadScheduledExecutor();
   AtomicBoolean versionSwapThreadScheduled = new AtomicBoolean(false);
-  private final VersionSwapDataChangeListener versionSwapListener = new VersionSwapDataChangeListener();
+  private final VersionSwapDataChangeListener<K, V> versionSwapListener;
 
   public VeniceAfterImageConsumerImpl(ChangelogClientConfig changelogClientConfig, PubSubConsumerAdapter consumer) {
     this(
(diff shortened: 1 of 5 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceAfterImageConsumerImpl.java
```diff
@@ -166,47 +181,58 @@ protected CompletableFuture<Void> internalSeek(
...
+    public void handleStoreChanged(Store store) {
+      synchronized (this) {
+        Set<Integer> partitions = new HashSet<>();
+        try {
+          // Check the current version of the server
+          int currentVersion = storeRepository.getStore(storeName).getCurrentVersion();
+
+          // Check the current ingested version
+          Set<PubSubTopicPartition> subscriptions = getTopicAssignment();
+          synchronized (subscriptions) {
```
xunyin8, line 195: Can you explain me the scenario when we need to synchronized on `subscriptions` when we have synchronized on this? I'm worried this may cause some deadlock.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceAfterImageConsumerImpl.java
```diff
@@ -166,47 +181,58 @@ protected CompletableFuture<Void> internalSeek(
...
             }
-          }
 
-          // Seek to end of push
-          if (currentVersion != maxVersion) {
-            // get current subscriptions and seek to endOfPush
-            for (PubSubTopicPartition partitionSubscription: subscriptions) {
-              partitions.add(partitionSubscription.getPartitionNumber());
+            // Seek to end of push
+            if (currentVersion != maxVersion) {
```
xunyin8, line 208: Will this work when we rollback current version?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceAfterImageConsumerImpl.java
```diff
@@ -166,47 +181,58 @@ protected CompletableFuture<Void> internalSeek(
...
-        synchronized (subscriptions) {
-          if (subscriptions.isEmpty()) {
-            return;
-          }
-          int maxVersion = -1;
-          for (PubSubTopicPartition topicPartition: subscriptions) {
-            int version = Version.parseVersionFromVersionTopicName(topicPartition.getPubSubTopic().getName());
-            if (version >= maxVersion) {
-              maxVersion = version;
+    public void handleStoreChanged(Store store) {
```
xunyin8, line 186: Maybe leave a comment or something that indicate it's expected for this `handleStoreChanged` callback to have a null value for store?

ZacAttack: good idea

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceAfterImageConsumerImpl.java
ZacAttack: it's kind of a long story, but getTopicAssignment returns a synchronized set that wraps the assignment list returned by the pubsub consumer.  The pubsub consumer assignment listing is not thread safe.  It does like an assignments.values() wrapped to a hashmap, but that values is backed by the actual map that's in the pubsub consumer.  We wrap it to the synchronized set to prevent thrown exceptions for concurrent access, though, rereading the implementation of synchronized set, it occurs to me that the synchronized blocks here actually don't really do anything since getAssignments returns a new instance every time.  So it's harmless, but it's also not helpful.  Might just prefer to remove it.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceAfterImageConsumerImpl.java
ZacAttack: lol nope.

Actually on reading this code it's kind of whack.  It basically says that so long as there is any partition that is consuming on the correction version, leave the rest alone, otherwise, migrate all partitions.  So a lot of this should change. (all of this code was copy pasted from the original version swap code)

xunyin8 (dismissed):

xunyin8 (approved):

PR #1369, opened 2024-12-06
[client-common][router] minor cleanup and some tests for otel

m-nagarajan:
## Summary
some minor cleanups and tests including
1. Not initializing `VeniceOpenTelemetryMetricsRepository` when otel is not configured
2. if `setMetricEntities` is not configured in VeniceMetricsConfig and if exponential bucket histogram is enabled (which is enabled by default), builder will fail fast.
3. Disabling OTel exporting to endpoint in routers for integration tests
4. Removing the explicit names passed in `RouterMetricEntity` and use the enum name itself to be less error prone
5. some tests around these changes

## How was this PR tested?
GH CI

### DIFF
--- a/internal/venice-client-common/src/main/java/com/linkedin/venice/utils/metrics/MetricsRepositoryUtils.java
+++ b/internal/venice-client-common/src/main/java/com/linkedin/venice/utils/metrics/MetricsRepositoryUtils.java
@@ -2,6 +2,7 @@ package com.linkedin.venice.utils.metrics;
 
 import com.linkedin.venice.stats.VeniceMetricsConfig;
 import com.linkedin.venice.stats.VeniceMetricsRepository;
+import com.linkedin.venice.stats.VeniceOpenTelemetryMetricNamingFormat;
 import io.tehuti.metrics.MetricConfig;
 import io.tehuti.metrics.MetricsRepository;
 import io.tehuti.metrics.stats.AsyncGauge;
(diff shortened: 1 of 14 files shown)


internal/venice-client-common/src/main/java/com/linkedin/venice/utils/metrics/MetricsRepositoryUtils.java
```diff
@@ -45,9 +61,17 @@ public static MetricsRepository createSingleThreadedMetricsRepository(
 
   public static MetricsRepository createSingleThreadedVeniceMetricsRepository(
       long maxMetricsMeasurementTimeoutMs,
-      long initialMetricsMeasurementTimeoutMs) {
+      long initialMetricsMeasurementTimeoutMs,
+      boolean isOtelEnabled,
+      VeniceOpenTelemetryMetricNamingFormat otelFormat) {
+    Collection<MetricEntity> metricEntities = new ArrayList<>();
+    metricEntities
+        .add(new MetricEntity("test_metric", MetricType.HISTOGRAM, MetricUnit.MILLISECOND, "Test description"));
```
FelixGV, line 69: That does not seem right within the `main` src. Is it leftover of some debugging? Should it be cleaned up? If there is a reason to keep it, then please add a comment explaining why.

services/venice-router/src/main/java/com/linkedin/venice/router/stats/RouterHttpRequestStats.java
```diff
@@ -129,9 +129,10 @@ public class RouterHttpRequestStats extends AbstractVeniceHttpStats {
   private final Sensor errorRetryAttemptTriggeredByPendingRequestCheckSensor;
 
   private final String systemStoreName;
-  private final Attributes commonMetricDimensions;
-  private final boolean emitOpenTelemetryMetrics;
-  private final VeniceOpenTelemetryMetricNamingFormat openTelemetryMetricFormat;
+  private Attributes commonMetricDimensions = null;
+  private boolean emitOpenTelemetryMetrics = false;
+  private VeniceOpenTelemetryMetricNamingFormat openTelemetryMetricFormat =
+      VeniceOpenTelemetryMetricNamingFormat.getDefaultFormat();
```
FelixGV, line 135: I would prefer that we keep these field final. Is there a reason to change this?

FelixGV (commented):
Just a couple of minor comments, but LGTM overall. Thanks!

services/venice-router/src/main/java/com/linkedin/venice/router/stats/RouterHttpRequestStats.java
m-nagarajan: I modified this to cut a couple of lines in the code, but having these class fields as final makes better sense. Changed it back.

internal/venice-client-common/src/main/java/com/linkedin/venice/utils/metrics/MetricsRepositoryUtils.java
m-nagarajan: Good catch, moved this code away from here.

m-nagarajan (commented):
Thanks @FelixGV for the review. Fixed the review comments.

FelixGV (dismissed):

FelixGV:
Thanks @m-nagarajan !

sushantmane (approved):

PR #1372, opened 2024-12-06
[server] Block OFFLINE->DROPPED ST to wait for drop partition action

xunyin8:
## [server] Block OFFLINE->DROPPED ST to wait for drop partition action

Drop partition is performed asynchronously through consumer action in SIT. This behavior allowed a race to occur when we have OFFLINE->DROPPED followed by OFFLINE->STANDBY for the same partition on the same host. This is because we initialize the partition synchronously in OFFLINE->STANDBY and there is a chance the previously queued drop partition consumer action is executed after OFFLINE->STANDBY and that will cause ingestion to fail.

The fix is to block and wait on the drop partition consumer action future in the OFFLINE->DROPPED state transition. This way when the host is picked again for OFFLINE->STANDBY for the same partition, it can do so with a clean state without any lingering async actions.

## How was this PR tested?
Existing and new unit tests

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/AbstractPartitionStateModel.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/AbstractPartitionStateModel.java
@@ -46,6 +46,15 @@ public abstract class AbstractPartitionStateModel extends StateModel {
   private static final int RETRY_COUNT = 5;
   private static final int RETRY_DURATION_MS = 1000;
   private static final int WAIT_PARTITION_ACCESSOR_TIME_OUT_MS = (int) TimeUnit.MINUTES.toMillis(5);
+  // The drop partition consumer action may take longer than expected and eventually timeout if:
+  // 1. The corresponding ingestion task is stuck and unable to process consumer actions.
+  // 2. The corresponding ingestion task died due to errors.
+  // We want to wait for drop partition because consumer actions are executed asynchronously w.r.t. Helix state
+  // transitions. Currently, we open a partition synchronously during OFFLINE->STANDBY. This can be problematic and have
+  // race conditions if we don't wait on the drop partition during X->DROPPED. This is because Helix can send the
+  // following state transitions right after each other for the same partition to the same node:
+  // X->DROPPED, OFFLINE->STANDBY
+  protected static final long WAIT_DROP_PARTITION_TIME_OUT_MS = TimeUnit.MINUTES.toMillis(60);
 
   private final IngestionBackend ingestionBackend;
   private final ReadOnlyStoreRepository storeRepository;
@@ -156,7 +165,8 @@ public abstract class AbstractPartitionStateModel extends StateModel {
   public void onBecomeDroppedFromError(Message message, NotificationContext context) {
     executeStateTransition(message, context, () -> {
       try {
-        removePartitionFromStoreGracefully();
+        CompletableFuture<Void> dropPartitionFuture = removePartitionFromStoreGracefully();
+        dropPartitionFuture.get(WAIT_DROP_PARTITION_TIME_OUT_MS, TimeUnit.MILLISECONDS);
       } catch (Exception e) {
         // Catch exception here to ensure state transition can complete ana avoid error->dropped->error->... loop
         logger.error("Encountered exception during the transition from ERROR to DROPPED.", e);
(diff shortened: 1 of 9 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/AbstractPartitionStateModel.java
```diff
@@ -46,6 +46,15 @@ public abstract class AbstractPartitionStateModel extends StateModel {
...
   private static final int WAIT_PARTITION_ACCESSOR_TIME_OUT_MS = (int) TimeUnit.MINUTES.toMillis(5);
+  // The drop partition consumer action may take longer than expected and eventually timeout if:
+  // 1. The corresponding ingestion task is stuck and unable to process consumer actions.
+  // 2. The corresponding ingestion task died due to errors.
+  // We want to wait for drop partition because consumer actions are executed asynchronously w.r.t. Helix state
+  // transitions. Currently, we open a partition synchronously during OFFLINE->STANDBY. This can be problematic and have
+  // race conditions if we don't wait on the drop partition during X->DROPPED. This is because Helix can send the
+  // following state transitions right after each other for the same partition to the same node:
+  // X->DROPPED, OFFLINE->STANDBY
+  protected static final long WAIT_DROP_PARTITION_TIME_OUT_MS = TimeUnit.MINUTES.toMillis(60);
```
kvargha, line 57: Is 60 minutes sufficient?

clients/da-vinci-client/src/test/java/com/linkedin/davinci/helix/VeniceLeaderFollowerStateModelTest.java
```diff
@@ -129,4 +130,25 @@ public void testWhenBecomeOfflineFromStandbyWithVersionDeletion() {
...
+    when(mockStore.getCurrentVersion()).thenReturn(2);
+    testStateModel.onBecomeDroppedFromOffline(mockMessage, mockContext);
+    verify(mockParticipantStateTransitionStats, times(1)).incrementThreadBlockedOnOfflineToDroppedTransitionCount();
+    verify(mockParticipantStateTransitionStats, times(1)).decrementThreadBlockedOnOfflineToDroppedTransitionCount();
+
+    // if the resource is the current serving version, state transition thread will be blocked. And it will be blocked
+    // again for drop partition action if it's done asynchronously via SIT
+    when(mockStore.getCurrentVersion()).thenReturn(1);
+    testStateModel.onBecomeDroppedFromOffline(mockMessage, mockContext);
+    verify(mockParticipantStateTransitionStats, times(3)).incrementThreadBlockedOnOfflineToDroppedTransitionCount();
```
kvargha, line 151: Why does this get invoked 3 times?

I can see once [here](https://github.com/linkedin/venice/blob/626568143325c3564eaf1362a1fd3b16e7a73df4/clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/LeaderFollowerPartitionStateModel.java#L191-L194), and another when `waitForDropPartition` is true. Where does the third execution happen?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/AbstractPartitionStateModel.java
xunyin8: Normally it shouldn't take that long right? IMO if the SIT is stuck or slow we will have other alerts/signal firing before 60 minutes.

kvargha: Makes sense to me.

clients/da-vinci-client/src/test/java/com/linkedin/davinci/helix/VeniceLeaderFollowerStateModelTest.java
kvargha: Gotcha.

Nit: Can we reset this mock to make it clear that second invocation of `testStateModel.onBecomeDroppedFromOffline` only calls `incrementThreadBlockedOnOfflineToDroppedTransitionCount` twice?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/helix/LeaderFollowerPartitionStateModel.java
```diff
@@ -195,12 +195,27 @@ public void onBecomeDroppedFromOffline(Message message, NotificationContext cont
...
-      removePartitionFromStoreGracefully();
+      CompletableFuture<Void> dropPartitionFuture = removePartitionFromStoreGracefully();
+      boolean waitForDropPartition = !dropPartitionFuture.isDone();
+      try {
+        if (waitForDropPartition) {
+          this.threadPoolStats.incrementThreadBlockedOnOfflineToDroppedTransitionCount();
+        }
+        dropPartitionFuture.get(WAIT_DROP_PARTITION_TIME_OUT_MS, TimeUnit.MILLISECONDS);
+      } catch (InterruptedException e) {
+        throw new VeniceException("Got interrupted while waiting for drop partition future to complete", e);
```
lluwm, line 211: When time-out happened, we will throw exception here, but not in the ERROR->DROPPED state transition above. Do we want to make them consistent or they are totally different and we need separate ways of handling them?

xunyin8: They are different, we explicitly don't want to throw exception from ERROR->DROPPED to prevent us from getting into an infinite loop when we have bad internal state. i.e. ERROR->DROPPED->ERROR->......

lluwm: SG. Can we log the error messages for debugging purpose, in case the timeout happens?

kvargha: Which function are you referring to for ERROR->DROPPED? I see `onBecomeOfflineFromError` but it's a No-op. @lluwm

lluwm: No, it's the `AbstractPartitionStateModel.onBecomeDroppedFromError` that's also modified in this PR.

kvargha (dismissed):

lluwm (approved):
Thank you @xunyin8 for fixing this and LGTM!

PR #1373, opened 2024-12-06
Cancel previous workflow run on new push

majisourav99:
## Cancel previous workflow check-run on new push

Often times developers push new commits to the same PR which kicks off new check runs, while the previous runs still running which causes longer wait times for the new workflow to start. This change will cancel the previous workflow runs on every new push leading to quicker check runs results for the new change.

## How was this PR tested?

GHCI

### DIFF
--- a/.github/rawWorkflows/gh-ci-parameterized-flow.txt
+++ b/.github/rawWorkflows/gh-ci-parameterized-flow.txt
@@ -11,6 +11,9 @@
     needs: $Dependency
     if: $Conditional
     timeout-minutes: $TimeOut
+    concurrency:
+     group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
+     cancel-in-progress: true
     steps:
       - uses: actions/checkout@v4
         with:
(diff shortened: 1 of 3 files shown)


.github/rawWorkflows/gh-ci-parameterized-flow.txt
```diff
@@ -11,6 +12,9 @@
     needs: $Dependency
     if: $Conditional
     timeout-minutes: $TimeOut
+    concurrency:
+     group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}-jdk${{ matrix.jdk }}-$FlowName
+     cancel-in-progress: true
```
nisargthakkar, line 17: For the primary repo, can we only cancel ongoing runs of the current pull request when it gets updated. For non-PRs, we should let them run

nisargthakkar: This is what I was thinking of 
```
cancel-in-progress:  ${{ github.repository != 'linkedin/venice' || github.event_name == 'pull_request' }}
```

nisargthakkar (approved):

PR #1375, opened 2024-12-06
[vpj][controller] Add deferred swap for target region push

misyel:

Currently, when target region push is enabled, it only ingests the data & swap to the new version in if the current region matches the target region defined by the target region push list or the native replication source fabric. The ingestion for the remaining regions are kicked off after the first region is able to swap to the new version.

The pr adds in a deferred version swap option to the existing target region push. In this version, data is ingested in all regions, but it will only swap to the new version in the target region. The behavior to swap the version out in the remaining regions will be added in later.

## How was this PR tested?

unit & integration tests

This will be enabled if `targeted.region.push.with.deferred.swap` is passed in in the job settings

### DIFF
--- a/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
+++ b/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
@@ -76,6 +76,7 @@ import static com.linkedin.venice.vpj.VenicePushJobConstants.SUPPRESS_END_OF_PUS
 import static com.linkedin.venice.vpj.VenicePushJobConstants.SYSTEM_SCHEMA_READER_ENABLED;
 import static com.linkedin.venice.vpj.VenicePushJobConstants.TARGETED_REGION_PUSH_ENABLED;
 import static com.linkedin.venice.vpj.VenicePushJobConstants.TARGETED_REGION_PUSH_LIST;
+import static com.linkedin.venice.vpj.VenicePushJobConstants.TARGETED_REGION_PUSH_WITH_DEFERRED_SWAP;
 import static com.linkedin.venice.vpj.VenicePushJobConstants.TEMP_DIR_PREFIX;
 import static com.linkedin.venice.vpj.VenicePushJobConstants.UNCREATED_VERSION_NUMBER;
 import static com.linkedin.venice.vpj.VenicePushJobConstants.USE_MAPPER_TO_BUILD_DICTIONARY;
@@ -502,6 +503,13 @@ public class VenicePushJob implements AutoCloseable {
       pushJobSettingToReturn.dataWriterComputeJobClass = objectClass;
     }
 
+    // If target region push with deferred version swap is enabled, enable deferVersionSwap and
+    // isTargetedRegionPushEnabled
+    if (props.getBoolean(TARGETED_REGION_PUSH_WITH_DEFERRED_SWAP, false)) {
+      pushJobSettingToReturn.deferVersionSwap = true;
+      pushJobSettingToReturn.isTargetedRegionPushEnabled = true;
+    }
+
     return pushJobSettingToReturn;
   }
 
@@ -851,7 +859,8 @@ public class VenicePushJob implements AutoCloseable {
       sendPushJobDetailsToController();
 
       // only kick off the validation and post-validation flow when everything has to be done in a single VPJ
-      if (!pushJobSetting.isTargetedRegionPushEnabled) {
+      if (!pushJobSetting.isTargetedRegionPushEnabled
+          || (pushJobSetting.deferVersionSwap && pushJobSetting.isTargetedRegionPushEnabled)) {
         return;
       }
 
@@ -2075,7 +2084,8 @@ public class VenicePushJob implements AutoCloseable {
     jobSetting.storeStorageQuota = storeResponse.getStore().getStorageQuotaInByte();
 
     // Do not enable for deferred swap or hybrid store
-    if (pushJobSetting.deferVersionSwap || storeResponse.getStore().getHybridStoreConfig() != null) {
+    boolean isDeferredSwap = pushJobSetting.deferVersionSwap && !pushJobSetting.isTargetedRegionPushEnabled;
+    if (isDeferredSwap || storeResponse.getStore().getHybridStoreConfig() != null) {
       LOGGER.warn(
           "target region is not available for {} as it hybrid or deferred version swap enabled.",
           jobSetting.storeName);
--- a/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
+++ b/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
@@ -400,6 +400,15 @@ public class VenicePushJob implements AutoCloseable {
     if (pushJobSettingToReturn.isIncrementalPush && pushJobSettingToReturn.isTargetedRegionPushEnabled) {
       throw new VeniceException("Incremental push is not supported while using targeted region push mode");
     }
+
+    // If target region push with deferred version swap is enabled, enable deferVersionSwap and
+    // isTargetedRegionPushEnabled
+    if (props.getBoolean(TARGETED_REGION_PUSH_WITH_DEFERRED_SWAP, false)) {
+      pushJobSettingToReturn.deferVersionSwap = true;
+      pushJobSettingToReturn.isTargetedRegionPushEnabled = true;
+      pushJobSettingToReturn.isTargetRegionPushWithDeferredSwapEnabled = true;
+    }
+
     if (props.containsKey(TARGETED_REGION_PUSH_LIST)) {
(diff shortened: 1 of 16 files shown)


internal/venice-common/src/main/java/com/linkedin/venice/meta/Store.java
```diff
@@ -341,4 +341,6 @@ static boolean isSystemStore(String storeName) {
   void setIsDavinciHeartbeatReported(boolean isReported);
 
   boolean getIsDavinciHeartbeatReported();
+
+  void updateVersionForDaVinciHeartbeat(int versionNumber, boolean reported);
```
majisourav99, line 345: We may not need this API, we can just the get the version and set the config.

misyel: The existing `getVersion()` function returns a read only version so I created this new function to get and update the config similar to `updateVersionStatus`

clients/venice-push-job/src/main/java/com/linkedin/venice/vpj/VenicePushJobConstants.java
```diff
@@ -324,6 +324,14 @@ private VenicePushJobConstants() {
    */
   public static final String TARGETED_REGION_PUSH_LIST = "targeted.region.push.list";
 
+  /**
+   * Config to enable target region push with deferred version swap
+   * In this mode, the VPJ will push data to all regions and only switch to the new version in a single region.
+   * The single region is decided by the store config in {@link StoreInfo#getNativeReplicationSourceFabric()}}.
```
majisourav99, line 330: We should have an override for this, either we can reuse the `"targeted.region.push.list"` or create a new one.

misyel: Yeah agreed, we're reusing `targeted.region.push.list` to override which regions we will switch to the new version on. Updated the comment to reflect that it can be overridden with `targeted.region.push.list`

services/venice-controller/src/main/java/com/linkedin/venice/pushmonitor/AbstractPushMonitor.java
```diff
@@ -1120,15 +1123,37 @@ private void updateStoreVersionStatus(String storeName, int versionNumber, Versi
...
+           * 2.target region push is enabled and the current region does NOT match the target region (it will be swapped
+           *  after targetSwapRegionWaitTime passes)
+           */
+          Set<String> targetRegions = RegionUtils.parseRegionsFilterList(version.getTargetSwapRegion());
+          boolean isValidTargetRegion = targetRegions.contains(regionName);
+          boolean isTargetRegionPushWithDeferredSwap = version.isVersionSwapDeferred() && isValidTargetRegion;
+          boolean isNormalPush = !version.isVersionSwapDeferred();
+          boolean isDeferredSwap = version.isVersionSwapDeferred() && targetRegions.isEmpty();
+          if (isTargetRegionPushWithDeferredSwap || isNormalPush) {
+            int previousVersion = store.getCurrentVersion();
```
majisourav99, line 1143: can you add a log here.

services/venice-controller/src/main/java/com/linkedin/venice/pushmonitor/AbstractPushMonitor.java
```diff
@@ -1120,15 +1123,37 @@ private void updateStoreVersionStatus(String storeName, int versionNumber, Versi
...
+           * 2.target region push w/ deferred swap is enabled and the current region matches the target region
+           *
+           * Do not switch to the new version now if:
+           * 1.deferred version swap is enabled (it will be manually swapped at a later date)
+           * 2.target region push is enabled and the current region does NOT match the target region (it will be swapped
+           *  after targetSwapRegionWaitTime passes)
+           */
+          Set<String> targetRegions = RegionUtils.parseRegionsFilterList(version.getTargetSwapRegion());
+          boolean isValidTargetRegion = targetRegions.contains(regionName);
+          boolean isTargetRegionPushWithDeferredSwap = version.isVersionSwapDeferred() && isValidTargetRegion;
```
majisourav99, line 1139: maybe combine the 2 boolean into one.
`boolean performVersionSwap = !version.isVersionSwapDeferred() ||  isValidTargetRegion;`

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/DaVinciClientTest.java
```diff
@@ -1547,6 +1548,39 @@ public void testDVCSnapshotGeneration(boolean useDiskStorage) throws Exception {
...
+      writeSimpleAvroFileWithIntToStringSchema(inputDirectory);
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    }
+    Properties vpjProperties = defaultVPJProps(cluster, inputDirectoryPath, storeName);
+    runVPJ(vpjProperties, 2, cluster);
+
+    try (ControllerClient controllerClient = cluster.getControllerClient()) {
+      StoreInfo store = controllerClient.getStore(storeName).getStore();
+      TestUtils.waitForNonDeterministicAssertion(
```
majisourav99, line 1573: could be part of same  
`TestUtils.waitForNonDeterministicAssertion(5, TimeUnit.SECONDS, () -> {
Assert.assertEquals(store.getIsDavinciHeartbeatReported(), true));
Assert.assertEquals(store.getVersion(2).getDVCHB(), true));
 }`

services/venice-controller/src/main/java/com/linkedin/venice/pushmonitor/AbstractPushMonitor.java
```diff
@@ -1120,15 +1123,45 @@ private void updateStoreVersionStatus(String storeName, int versionNumber, Versi
...
             LOGGER.info(
                 "Version swap is deferred for store {} on version {}. Skipping version swap.",
                 store.getName(),
                 versionNumber);
           } else {
-            int previousVersion = store.getCurrentVersion();
-            store.setCurrentVersion(versionNumber);
-            realTimeTopicSwitcher.transmitVersionSwapMessage(store, previousVersion, versionNumber);
+            LOGGER.info(
+                "Version swap is deferred for store {} on version {} in region {} due to target region push w/ deferred swap",
```
majisourav99, line 1161: may be log why its deferred, eg deferred as target region {} is different from the region {} running this push.

services/venice-controller/src/main/java/com/linkedin/venice/pushmonitor/AbstractPushMonitor.java
```diff
@@ -1120,15 +1123,45 @@ private void updateStoreVersionStatus(String storeName, int versionNumber, Versi
...
+           * 2.target region push is enabled and the current region does NOT match the target region (it will be swapped
+           *  after targetSwapRegionWaitTime passes)
+           */
+          Set<String> targetRegions = RegionUtils.parseRegionsFilterList(version.getTargetSwapRegion());
+          boolean isTargetRegionPushWithDeferredSwap =
+              version.isVersionSwapDeferred() && targetRegions.contains(regionName);
+          boolean isNormalPush = !version.isVersionSwapDeferred();
+          boolean isDeferredSwap = version.isVersionSwapDeferred() && targetRegions.isEmpty();
+          if (isTargetRegionPushWithDeferredSwap || isNormalPush) {
+            LOGGER.debug(
```
majisourav99, line 1143: you plan to keep the debug log ? may be make it regular info log with message like
Setting current version {} to store {} during " + isRegularPush ? "normal push" : " target region deferred push".

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
```diff
@@ -2075,14 +2085,17 @@ private void validateStoreSettingAndPopulate(ControllerClient controllerClient,
...
+        pushJobSetting.deferVersionSwap && !pushJobSetting.isTargetRegionPushWithDeferredSwapEnabled;
+    if (isDeferredSwap || storeResponse.getStore().getHybridStoreConfig() != null) {
       LOGGER.warn(
           "target region is not available for {} as it hybrid or deferred version swap enabled.",
           jobSetting.storeName);
       jobSetting.isTargetedRegionPushEnabled = false;
     }
 
-    if (jobSetting.isTargetedRegionPushEnabled && jobSetting.targetedRegions == null) {
+    if ((jobSetting.isTargetedRegionPushEnabled || jobSetting.isTargetRegionPushWithDeferredSwapEnabled)
```
majisourav99, line 2097: Also can we validate we do not allow both `isTargetedRegionPushEnabled` and `isTargetRegionPushWithDeferredSwapEnabled` to be enabled ?

majisourav99: Also add some unit tests to not allow combinations of these configs

majisourav99 (approved):
LGTM! Thanks Michelle!

PR #1377, opened 2024-12-09
[router][server] Reduce the granularity of latency histogram percentiles and disable key value profiling for single gets

nisargthakkar:
## Reduce the granularity of latency histogram percentiles and disable key value profiling for single gets

Currently, we record fine-grained percentiles for key and value of all single get requests. This bloats up the metrics significantly. This commit restricts key and value profiling only for cases where this config is enabled.

In addition to this, removed P77 and P90 metrics from latency metrics. We still have p50, p95, p99 and p99.9.

## How was this PR tested?

GH CI

### DIFF
--- a/services/venice-server/src/main/java/com/linkedin/venice/stats/ServerHttpRequestStats.java
+++ b/services/venice-server/src/main/java/com/linkedin/venice/stats/ServerHttpRequestStats.java
@@ -295,8 +295,8 @@ public class ServerHttpRequestStats extends AbstractVeniceHttpStats {
         () -> totalStats.earlyTerminatedEarlyRequestCountSensor,
         new OccurrenceRate());
 
-    if (isKeyValueProfilingEnabled || requestType == RequestType.SINGLE_GET) {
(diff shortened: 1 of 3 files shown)


m-nagarajan (approved):
LGTM, thanks @nisargthakkar

services/venice-server/src/main/java/com/linkedin/venice/stats/ServerHttpRequestStats.java
```diff
@@ -295,8 +295,8 @@ public ServerHttpRequestStats(
         () -> totalStats.earlyTerminatedEarlyRequestCountSensor,
         new OccurrenceRate());
 
-    if (isKeyValueProfilingEnabled || requestType == RequestType.SINGLE_GET) {
-      // size profiling is only expensive for requests with lots of keys, but we keep it always on for single gets...
+    if (isKeyValueProfilingEnabled) {
+      // size profiling is expensive
```
FelixGV, line 299: This is a regression 😅 ...

PR #1378, opened 2024-12-10
[server] Fix NPE when key value size profiling is not enabled

sushantmane:
## Fix NPE when key value size profiling is not enabled

This commit addresses a NullPointerException in `ServerHttpRequestStats` when attempting to record key or value sizes while the corresponding profiling sensors (`requestKeySizeSensor` or `requestValueSizeSensor`) are null. 

## How was this PR tested?

UT

### DIFF
--- a/services/venice-server/src/main/java/com/linkedin/venice/stats/ServerHttpRequestStats.java
+++ b/services/venice-server/src/main/java/com/linkedin/venice/stats/ServerHttpRequestStats.java
@@ -445,11 +445,15 @@ public class ServerHttpRequestStats extends AbstractVeniceHttpStats {
   }
 
   public void recordKeySizeInByte(int keySize) {
-    requestKeySizeSensor.record(keySize);
+    if (requestKeySizeSensor != null) {
+      requestKeySizeSensor.record(keySize);
+    }
   }
 
   public void recordValueSizeInByte(int valueSize) {
-    requestValueSizeSensor.record(valueSize);
+    if (requestValueSizeSensor != null) {
+      requestValueSizeSensor.record(valueSize);
+    }
   }
 
   public void recordMisroutedStoreVersionRequest() {


kvargha (approved):

nisargthakkar (approved):

services/venice-server/src/main/java/com/linkedin/venice/stats/ServerHttpRequestStats.java
```diff
@@ -445,11 +445,15 @@ public void recordEarlyTerminatedEarlyRequest() {
   }
 
   public void recordKeySizeInByte(int keySize) {
-    requestKeySizeSensor.record(keySize);
+    if (requestKeySizeSensor != null) {
+      requestKeySizeSensor.record(keySize);
+    }
```
FelixGV, line 450: Hmm... but it's weird though... why is this sensor null?

The server config should control both whether this sensor is null, and also whether this function gets called or not 🤔

FelixGV (commented):
Thanks!

services/venice-server/src/main/java/com/linkedin/venice/stats/ServerHttpRequestStats.java
FelixGV: What I mean is:

This function is only called from `MultiGetResponseStatsWithSizeProfiling` (and the equivalent compute-related class). And `MultiGetResponseStatsWithSizeProfiling` can only get constructed when `serverConfig.isKeyValueProfilingEnabled()` is true...

And `AggServerHttpRequestStats` also gets passed the same config, which (after a few indirections) should determine whether that sensor gets created or not.

So how can these be mismatched?

sushantmane: The issue is that sensor creation is controlled by a configuration flag (`isKeyValueProfilingEnabled`), but the reporting logic is not. Hence reporting just blindly tries to update sensor even when `isKeyValueProfilingEnabled == false`

sushantmane: I think this is coming from `SingleGetResponseStats::recordMetrics` --> `ResponseStatsUtil::recordValueSizeInByte`

sushantmane:
```
```

sushantmane: This is regression due to: https://github.com/linkedin/venice/pull/1377

nisargthakkar: Sorry about that. Any idea why this wasn't caught in any of the tests?

nisargthakkar (approved):

services/venice-server/src/main/java/com/linkedin/venice/stats/ServerHttpRequestStats.java
FelixGV: Oh yeah, right. I see it now: https://github.com/linkedin/venice/pull/1377#discussion_r1878842022

sushantmane: What's the motivation to disable KV size profiling for single gets?

FelixGV: Metric bloat maybe? IDK. @nisargthakkar ?

sushantmane: >Any idea why this wasn't caught in any of the tests?

Probably we don't have E2E a test which verifies all metrics for single-get request type?

sushantmane:
Closing this PR since @nisargthakkar updating KV size profiling behavior in https://github.com/linkedin/venice/pull/1381

PR #1381, opened 2024-12-10
[server] Fix stats handling for SINGLE_GET requests

nisargthakkar:
## Fix stats handling for `SINGLE_GET` requests

In a previous commit (#1377), when removing some metrics, we ended up removing some key metrics about key and value size values for single_get requests as well. That change also led to NPEs as the caller expected the K/V stats to always be recorded for Single-get requests. This commit adds `Avg` and `Max` stats by default for key and value sizes in single get requests.

## How was this PR tested?

Added unit tests

### DIFF
--- a/services/venice-server/src/main/java/com/linkedin/venice/stats/ServerHttpRequestStats.java
+++ b/services/venice-server/src/main/java/com/linkedin/venice/stats/ServerHttpRequestStats.java
@@ -295,23 +295,33 @@ public class ServerHttpRequestStats extends AbstractVeniceHttpStats {
         () -> totalStats.earlyTerminatedEarlyRequestCountSensor,
         new OccurrenceRate());
 
-    if (isKeyValueProfilingEnabled) {
-      // size profiling is expensive
+    if (isKeyValueProfilingEnabled || requestType == RequestType.SINGLE_GET) {
+      final MeasurableStat[] valueSizeStats;
+      final MeasurableStat[] keySizeStats;
       String requestValueSizeSensorName = "request_value_size";
+      String requestKeySizeSensorName = "request_key_size";
+      if (isKeyValueProfilingEnabled) {
+        valueSizeStats = TehutiUtils
+            .getFineGrainedPercentileStatWithAvgAndMax(getName(), getFullMetricName(requestValueSizeSensorName));
+        keySizeStats = TehutiUtils
+            .getFineGrainedPercentileStatWithAvgAndMax(getName(), getFullMetricName(requestKeySizeSensorName));
+      } else {
+        valueSizeStats = new MeasurableStat[] { new Avg(), new Max() };
+        keySizeStats = new MeasurableStat[] { new Avg(), new Max() };
+      }
+
       requestValueSizeSensor = registerPerStoreAndTotal(
           requestValueSizeSensorName,
           totalStats,
           () -> totalStats.requestValueSizeSensor,
-          TehutiUtils
-              .getFineGrainedPercentileStatWithAvgAndMax(getName(), getFullMetricName(requestValueSizeSensorName)));
-      String requestKeySizeSensorName = "request_key_size";
+          valueSizeStats);
       requestKeySizeSensor = registerPerStoreAndTotal(
           requestKeySizeSensorName,
           totalStats,
           () -> totalStats.requestKeySizeSensor,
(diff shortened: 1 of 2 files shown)


services/venice-server/src/main/java/com/linkedin/venice/stats/ServerHttpRequestStats.java
```diff
@@ -295,23 +295,35 @@ public ServerHttpRequestStats(
...
       String requestValueSizeSensorName = "request_value_size";
+      String requestKeySizeSensorName = "request_key_size";
+      if (isKeyValueProfilingEnabled) {
+        valueSizeStats = TehutiUtils
+            .getFineGrainedPercentileStatWithAvgAndMax(getName(), getFullMetricName(requestValueSizeSensorName));
+        keySizeStats = TehutiUtils
+            .getFineGrainedPercentileStatWithAvgAndMax(getName(), getFullMetricName(requestKeySizeSensorName));
+      } else {
+        valueSizeStats =
+            TehutiUtils.getPercentileStatWithAvgAndMax(getName(), getFullMetricName(requestValueSizeSensorName));
```
sushantmane, line 310: Can we just use min, max, avg for sizes?

nisargthakkar: Sorry, didn't see your comment and ended up force pushing

services/venice-server/src/main/java/com/linkedin/venice/stats/ServerHttpRequestStats.java
```diff
@@ -295,23 +295,33 @@ public ServerHttpRequestStats(
         () -> totalStats.earlyTerminatedEarlyRequestCountSensor,
         new OccurrenceRate());
 
-    if (isKeyValueProfilingEnabled) {
-      // size profiling is expensive
+    if (isKeyValueProfilingEnabled || requestType == RequestType.SINGLE_GET) {
```
sushantmane, line 298: Should we introduce a separate flag specifically for tracking percentiles?

With the current change, percentiles for single-gets won’t be tracked unless isKeyValueProfilingEnabled is enabled. Additionally, enabling isKeyValueProfilingEnabled automatically tracks key-value sizes for all request types. It might be worth considering separating these two functionalities with separate config options?.

sushantmane: OTHOH this might be fine if, by default, we want to track only average and maximum sizes for single-gets. If more detailed data is needed, such as percentiles, users can enable key-value profiling explicitly.

nisargthakkar: I don't think we need to split the two functionalities. We can add it later if we need it

services/venice-server/src/test/java/com/linkedin/venice/stats/AggServerHttpRequestStatsTest.java
```diff
@@ -13,10 +13,11 @@
 
 
 public class AggServerHttpRequestStatsTest {
-  protected MetricsRepository metricsRepository;
-  protected MockTehutiReporter reporter;
-  protected AggServerHttpRequestStats singleGetStats;
-  protected AggServerHttpRequestStats batchGetStats;
+  private MetricsRepository metricsRepository, metricsRepositoryForKVProfiling;
+  private MockTehutiReporter reporter, reporterForKVProfiling;
```
sushantmane, line 17: Can we  follow style consistent with other variable declarations in this class? (one declaration per line)

services/venice-server/src/main/java/com/linkedin/venice/stats/ServerHttpRequestStats.java
sushantmane: Sounds good.

services/venice-server/src/test/java/com/linkedin/venice/stats/AggServerHttpRequestStatsTest.java
nisargthakkar: Sure

sushantmane (approved):
LGTM! Thanks, @nisargthakkar!

PR #1383, opened 2024-12-10
[server] Make cleaning up partitions on startup configurable. Off by default

kvargha:

We need to test this feature thoroughly before rolling it out in prod.

## How was this PR tested?

CI


sushantmane (approved):
LGTM. Thanks, @kvargha!

sushantmane:
For hotfix select hotfix branch as base instead of linkedin:main

PR #1384, opened 2024-12-11
[test] Fix TestChangelogConsumer and address related issues

sushantmane:
## Fix TestChangelogConsumer and address related issues

- Fixed regression in `TestChangelogConsumer` caused by PR #1328, where it attempted to
register incompatible schemas.  
- Removed the creation of a pub-sub broker in the test as it was incorrectly relying on  
the ZK state of child controllers' Kafka broker.  
- Miscellaneous: Ensured proper cleanup by closing the cluster setup during teardown in
`TestDumpIngestionContext`.

## How was this PR tested?

CI

### DIFF
--- a/internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/consumer/TestChangelogConsumer.java
+++ b/internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/consumer/TestChangelogConsumer.java
@@ -6,7 +6,6 @@ import static com.linkedin.venice.ConfigKeys.CLUSTER_NAME;
 import static com.linkedin.venice.ConfigKeys.KAFKA_BOOTSTRAP_SERVERS;
 import static com.linkedin.venice.ConfigKeys.KAFKA_LINGER_MS;
 import static com.linkedin.venice.ConfigKeys.SERVER_AA_WC_WORKLOAD_PARALLEL_PROCESSING_ENABLED;
-import static com.linkedin.venice.ConfigKeys.ZOOKEEPER_ADDRESS;
 import static com.linkedin.venice.integration.utils.VeniceClusterWrapperConstants.DEFAULT_PARENT_DATA_CENTER_REGION_NAME;
 import static com.linkedin.venice.integration.utils.VeniceControllerWrapper.D2_SERVICE_NAME;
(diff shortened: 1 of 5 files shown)


internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/consumer/TestChangelogConsumer.java
```diff
@@ -687,89 +674,82 @@ public void testSpecificRecordBootstrappingVeniceChangelogConsumer() throws Exce
...
-        Assert.assertEquals(polledChangeEventsList.size(), 2);
-      });
-      Assert.assertTrue(
-          polledChangeEventsMap.get(Integer.toString(10000)).getValue().getCurrentValue() instanceof SpecificRecord);
+    PubSubBrokerWrapper localKafka = multiRegionMultiClusterWrapper.getChildRegions().get(0).getPubSubBrokerWrapper();
+    Properties consumerProperties = new Properties();
+    String localKafkaUrl = localKafka.getAddress();
+    consumerProperties.put(KAFKA_BOOTSTRAP_SERVERS, localKafkaUrl);
+    consumerProperties.put(CLUSTER_NAME, clusterName);
+    consumerProperties.put(ZOOKEEPER_ADDRESS, localZkServer.getAddress());
```
nisargthakkar, line 683: Can you see if we can remove this? No client should depend on ZK directly

nisargthakkar: There's one more place down in the next test that also has the same

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/consumer/TestChangelogConsumer.java
nisargthakkar, line 1: Looks like CC is not happy with the missing Zk address. Seems like a future incident to me, TBH. But we don't need to solve it in this PR

sushantmane: Maybe. I'm not gonna fix it in this PR.  

cc: @ZacAttack @sixpluszero

nisargthakkar (approved):

PR #1257, opened 2024-10-23
[compat][server][client][test] Global RT DIV improvement (part 2): Chunking support for DIV message

lluwm:
This change mainly focuses on adding chunking support for DIV messages when they are produced to Kafka topics, as the size of DIV message can be large. We leverage today's chunking mechanism for regular records and extend it to support DIV with the following modifications:

1. All the DIV messages are of type CONTROL_MESSAGE_DIV in its KafkaKey and their corresponding KafkaMessageEnvelope uses Put as the payload.
2. Inside the Put payload, the actual message is stored in the putValue field and the schemaId can have the following cases:
   - If the DIV message is non-chunked, the schemaId is set to GLOBAL_DIV_STATE.
   - If the DIV message is chunk message, the schemaId is set to CHUNK.
   - If the DIV message is a chunk manifest message, the schemaId is set to CHUNKED_VALUE_MANIFEST.
3. ChunkAssembler is adapted, on the receiver side, to buffer, assemble, and deserialize DIV messages (chunked/non-chunked).

## How was this PR tested?
- new integration test.
- passed CI

## Does this PR introduce any user-facing changes?

- [x] No. You can skip the rest of this section.
- [ ] Yes. Make sure to explain your proposed changes and call out the behavior change.

PR #1385, opened 2024-12-11
[server] [client] [test] Global RT DIV: Chunking Support

KaiSernLim:
## Summary

Continuation from #1257. Schema changes in #1523.

This PR mainly focuses on adding chunking support for DIV messages when they are produced to Kafka topics, as the size of the DIV message can surpass the ~1MB Kafka message limit. The existing chunking mechanism is reused, including the `CHUNK` and `CHUNKED_VALUE_MANIFEST` values in the message's `schemaId`:

Every DIV message has `GLOBAL_RT_DIV` for the header byte in its `KafkaKey`. The corresponding `KafkaMessageEnvelope` has a `Put` payload utilizing the `putValue` field containing the `GlobalRtDiv` data, and which has the following `schemaId`:
1. If the DIV message is non-chunked, the KME `schemaId` is set to the current protocol version of `GLOBAL_RT_DIV`.
2. If the DIV message is chunk message, the KME `schemaId` is set to `CHUNK`.
3. If the DIV message is a chunk manifest message, the KME `schemaId` is set to `CHUNKED_VALUE_MANIFEST`. The `schemaId` of the `ChunkedValueManifest` will be the current protocol version of `GLOBAL_RT_DIV`.

GlobalRtDivMessages@2x

DivChunking

## Changes
1. Added a new `MessageType` called `GlobalRtDiv`, which reuses the `Put` message type format and objects. When the Venice server encounters a message with `KafkaKey` containing the `GlobalRtDiv` header byte, it will know to process this message differently from a regular `Put`.
    * The only indication of this new `GlobalRtDiv` message type is the header byte in `KafkaKey`. Otherwise, it's identical to a regular `Put`.
    * `KafkaMessageEnvelope.avsc` will not be updated to avoid the unnecessary risk of incompatible avro formats when upgrading the cluster.
    * The risk in not creating a dedicated `GlobalRtDiv` message type in KME is that the `GlobalRtDiv` objects will be processed as user records and stored in the storage engine, which seems to be much less scary than a cluster upgrade issue.
2. `GlobalRtDiv` messages should not be processed if they originate from remote VT and RT, because those are invalid scenarios. These two conditions are checked.

### Minor Changes
1. Renamed `resetUpstreamOffsetMap()` to `mergeUpstreamOffsets()` in `OffsetRecord`.
2. Updated `toString()` in `KafkaKey`, which incorrectly assumed all messages would be `ControlMessage`, `Put`, or `Delete`. This misses `Update` messages and the new `GlobalRtDiv` message that is being added.
3. Added various helper methods (`buildPutPayload()` and `buildManifestPayload()`) in `VeniceWriter` for creating the `Put` payloads and when chunking is involved.

## Testing
1. Unit Tests
    1. `testGlobalRtDivChunking()` in `VeniceWriterUnitTest`
    2. `testShouldProcessRecordForGlobalRtDivMessage()` in `StoreIngestionTaskTest`
    3. `testProcessGlobalRtDivMessage()` in `StoreIngestionTaskTest`
2. Integration Test
    1. `testChunkedDiv()` in `TestGlobalRtDiv` 

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
@@ -1922,6 +1922,20 @@ public class LeaderFollowerStoreIngestionTask extends StoreIngestionTask {
               }
             }
           }
+
+          /**
+           * Leader is consuming from remote version topic, and if it read div control messages in the remote version
+           * topic, it should ignore them and not process or apply them to its own div state.
+           */
+          if (record.getKey().isDivControlMessage()) {
+            String msg = String.format(
+                "Leader for replica: %s received a div control message in remote version topic. Skipping the message.",
+                partitionConsumptionState.getReplicaId());
+            if (!REDUNDANT_LOGGING_FILTER.isRedundantException(msg)) {
+              LOGGER.info(msg);
+            }
+            return false;
+          }
         }
 
         if (!Utils.resolveLeaderTopicFromPubSubTopic(pubSubTopicRepository, record.getTopicPartition().getPubSubTopic())
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
@@ -1500,7 +1500,7 @@ public class LeaderFollowerStoreIngestionTask extends StoreIngestionTask {
         upstreamTopic = versionTopic;
       }
       if (upstreamTopic.isRealTime()) {
-        offsetRecord.mergeUpstreamOffsets(partitionConsumptionState.getLatestProcessedUpstreamRTOffsetMap());
+        offsetRecord.updateUpstreamOffsets(partitionConsumptionState.getLatestProcessedUpstreamRTOffsetMap());
       } else {
         offsetRecord.setCheckpointUpstreamVersionTopicOffset(
             partitionConsumptionState.getLatestProcessedUpstreamVersionTopicOffset());
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
@@ -7,6 +7,8 @@ import static com.linkedin.davinci.kafka.consumer.LeaderFollowerStateType.IN_TRA
 import static com.linkedin.davinci.kafka.consumer.LeaderFollowerStateType.LEADER;
 import static com.linkedin.davinci.kafka.consumer.LeaderFollowerStateType.PAUSE_TRANSITION_FROM_STANDBY_TO_LEADER;
 import static com.linkedin.davinci.kafka.consumer.LeaderFollowerStateType.STANDBY;
+import static com.linkedin.davinci.validation.PartitionTracker.TopicType.REALTIME_TOPIC_TYPE;
+import static com.linkedin.davinci.validation.PartitionTracker.TopicType.VERSION_TOPIC_TYPE;
 import static com.linkedin.venice.kafka.protocol.enums.ControlMessageType.END_OF_PUSH;
 import static com.linkedin.venice.kafka.protocol.enums.ControlMessageType.START_OF_SEGMENT;
 import static com.linkedin.venice.kafka.protocol.enums.MessageType.UPDATE;
@@ -35,6 +37,7 @@ import com.linkedin.davinci.store.view.MaterializedViewWriter;
 import com.linkedin.davinci.store.view.VeniceViewWriter;
 import com.linkedin.davinci.validation.KafkaDataIntegrityValidator;
 import com.linkedin.davinci.validation.PartitionTracker;
+import com.linkedin.davinci.validation.PartitionTracker.TopicType;
 import com.linkedin.venice.common.VeniceSystemStoreUtils;
 import com.linkedin.venice.compression.CompressionStrategy;
 import com.linkedin.venice.exceptions.VeniceException;
@@ -1497,7 +1500,7 @@ public class LeaderFollowerStoreIngestionTask extends StoreIngestionTask {
         upstreamTopic = versionTopic;
       }
       if (upstreamTopic.isRealTime()) {
-        offsetRecord.resetUpstreamOffsetMap(partitionConsumptionState.getLatestProcessedUpstreamRTOffsetMap());
+        offsetRecord.mergeUpstreamOffsets(partitionConsumptionState.getLatestProcessedUpstreamRTOffsetMap());
       } else {
         offsetRecord.setCheckpointUpstreamVersionTopicOffset(
             partitionConsumptionState.getLatestProcessedUpstreamVersionTopicOffset());
@@ -1637,7 +1640,6 @@ public class LeaderFollowerStoreIngestionTask extends StoreIngestionTask {
               int actualSchemaId = ByteUtils.readInt(actualValue, 0);
               Put put = (Put) envelope.payloadUnion;
               if (actualSchemaId == put.schemaId) {
-
                 if (put.putValue.equals(
                     ByteBuffer.wrap(
                         actualValue,
@@ -1922,18 +1924,8 @@ public class LeaderFollowerStoreIngestionTask extends StoreIngestionTask {
               }
             }
           }
-
-          /**
-           * Leader is consuming from remote version topic, and if it read div control messages in the remote version
-           * topic, it should ignore them and not process or apply them to its own div state.
-           */
-          if (record.getKey().isDivControlMessage()) {
-            String msg = String.format(
-                "Leader for replica: %s received a div control message in remote version topic. Skipping the message.",
-                partitionConsumptionState.getReplicaId());
-            if (!REDUNDANT_LOGGING_FILTER.isRedundantException(msg)) {
-              LOGGER.info(msg);
-            }
+          // Global RT DIV messages should be completely ignored when leader is consuming from remote version topic
+          if (record.getKey().isGlobalRtDiv()) {
             return false;
           }
         }
@@ -2306,18 +2298,17 @@ public class LeaderFollowerStoreIngestionTask extends StoreIngestionTask {
       Iterable<PubSubMessage<KafkaKey, KafkaMessageEnvelope, Long>> records,
       String kafkaUrl,
       PubSubTopicPartition topicPartition) {
-    PartitionConsumptionState partitionConsumptionState =
-        partitionConsumptionStateMap.get(topicPartition.getPartitionNumber());
-    if (partitionConsumptionState == null) {
+    final PartitionConsumptionState pcs = partitionConsumptionStateMap.get(topicPartition.getPartitionNumber());
+    if (pcs == null) {
       // The partition is likely unsubscribed, will skip these messages.
       LOGGER.warn(
-          "No partition consumption state for store version: {}, partition:{}, will filter out all the messages",
+          "No partition consumption state for store version: {}, partition: {}, will filter out all the messages",
           kafkaVersionTopic,
           topicPartition.getPartitionNumber());
       return Collections.emptyList();
     }
-    boolean isEndOfPushReceived = partitionConsumptionState.isEndOfPushReceived();
-    if (!shouldProduceToVersionTopic(partitionConsumptionState)) {
+    boolean isEndOfPushReceived = pcs.isEndOfPushReceived();
+    if (!shouldProduceToVersionTopic(pcs)) {
       return records;
     }
     /**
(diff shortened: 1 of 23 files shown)


internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
```diff
@@ -981,55 +981,58 @@ public CompletableFuture<PubSubProduceResult> put(
...
+      byte[] serializedValue) {
+    final Supplier<String> reportSizeGenerator = () -> getSizeReport(serializedKey.length, serializedValue.length, 0);
+    // TODO: This needs to be implemented later to support Global RT DIV.
     final CompletableFuture<Void> completableFuture = new CompletableFuture<>();
     PubSubProducerCallback callback = new ErrorPropagationCallback(new CompletableFutureCallback(completableFuture));
-    BiConsumer<KeyProvider, Put> sendMessageFunction = (keyProvider, putPayload) -> sendMessage(
+    BiConsumer<KeyProvider, Object> sendMessageFunction = (keyProvider, putPayload) -> sendMessage(
         keyProvider,
-        MessageType.PUT,
+        MessageType.GLOBAL_RT_DIV,
```
lluwm, line 1035: Is this correct? This is the KME type, and it should be `PUT` as you described in the description. Please also verify other places you set the KME type.

internal/venice-common/src/test/java/com/linkedin/venice/writer/VeniceWriterUnitTest.java
```diff
@@ -687,41 +690,34 @@ public void testGlobalDivChunking() {
...
+          ControlMessage controlMessage = ((ControlMessage) kme.getPayloadUnion());
+          assertEquals(ControlMessageType.START_OF_SEGMENT.getValue(), controlMessage.getControlMessageType());
         } else {
-          Assert.assertTrue(kme.messageType == MessageType.PUT.getValue());
           Put put = (Put) kme.payloadUnion;
-          Assert.assertTrue(
-              put.getSchemaId() == AvroProtocolDefinition.CHUNK.getCurrentProtocolVersion()
-                  || put.getSchemaId() == AvroProtocolDefinition.CHUNKED_VALUE_MANIFEST.getCurrentProtocolVersion()
-                  || put.getSchemaId() == AvroProtocolDefinition.GLOBAL_DIV_STATE.getCurrentProtocolVersion());
+          assertEquals(kme.messageType, MessageType.GLOBAL_RT_DIV.getValue());
```
lluwm, line 719: Ditto. The KME message type is `PUT` in the PR's description.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -1479,7 +1482,7 @@ protected void updateOffsetMetadataInOffsetRecord(PartitionConsumptionState part
         upstreamTopic = versionTopic;
       }
       if (upstreamTopic.isRealTime()) {
-        offsetRecord.resetUpstreamOffsetMap(partitionConsumptionState.getLatestProcessedUpstreamRTOffsetMap());
+        offsetRecord.mergeUpstreamOffsets(partitionConsumptionState.getLatestProcessedUpstreamRTOffsetMap());
```
sixpluszero, line 1485: Why do you do this change? Personally I think this makes less explicit about the Map we are tracking

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
KaiSernLim: Good catch. Fixed.

internal/venice-common/src/test/java/com/linkedin/venice/writer/VeniceWriterUnitTest.java
KaiSernLim: Fixed.

internal/venice-common/src/main/java/com/linkedin/venice/kafka/protocol/enums/MessageType.java
```diff
@@ -21,7 +21,8 @@
  */
 public enum MessageType implements VeniceEnumValue {
   PUT(0, Constants.PUT_KEY_HEADER_BYTE), DELETE(1, Constants.PUT_KEY_HEADER_BYTE),
-  CONTROL_MESSAGE(2, Constants.CONTROL_MESSAGE_KEY_HEADER_BYTE), UPDATE(3, Constants.UPDATE_KEY_HEADER_BYTE);
+  CONTROL_MESSAGE(2, Constants.CONTROL_MESSAGE_KEY_HEADER_BYTE), UPDATE(3, Constants.UPDATE_KEY_HEADER_BYTE),
+  GLOBAL_RT_DIV(4, Constants.GLOBAL_RT_DIV_KEY_HEADER_BYTE);
```
sixpluszero, line 25: I wonder if it is ok to not add it into KME.  I am thinking for example, Da Vinci and CDC, if the code is not upgraded, how can it be handling the global rt div? Or should we completely mute them from processing RT div message?
Will this controlled by store config or just server config?

KaiSernLim: It's controlled by [server config](https://github.com/linkedin/venice/pull/1179/files#diff-f1babd8c1a4cf090b308fdba3add6bf82d484d969b126a35b1b406976056ed62R511).

sixpluszero: I mean, with this code change, we will eventually enable the feature to produce RT DIV snapshot to VT topic. This will be processed by follower, which also applicable to DVC. I know DVC won't be leader, but will the existing code handle unknown new msg type gracefully?

lluwm: You're right. The reader of the msg type (to be able to ignore it) has to be deployed before enabling this feature. i don't know if there is other way to walk around it, feel free to suggest alternatives.

KaiSernLim: I didn't look very deep, but I think that the existing code will throw an exception when encountering an unknown message type.

Handling an unknown message type will be implemented in a future PR. Since we control the feature flag, we can choose not to enable the feature flag until all server instances support the new message type.

gaojieliu: We need to deploy the changes to all VT consumers before enabling this feature, which includes:
1. Server.
2. DaVinci.
3. ETL.
4. CDC.

gaojieliu: And I think it is better to treat the new KME type as a `Control Message`, and I would like to learn why if we can't do that.

internal/venice-common/src/main/resources/avro/GlobalRtDivState/v1/GlobalRtDivState.avsc
```diff
@@ -0,0 +1,85 @@
...
+      ],
+      "default": null
+    },
+    {
+      "name": "producerStates",
+      "type": {
+        "type": "map",
+        "doc": "A map that maps producer GUID -> producer state for realtime data.",
+        "values": {
+          "name": "ProducerPartitionState",
```
gaojieliu, line 22: Is it possible to reference the existing definition in `PartitionState.avsc`?

gaojieliu (commented):
I would like to call out another item we didn't discuss in the review meeting:
1. Cleanup the Global RT DIV messages from VT.

This new type of message can be large and with chunking support, we are leaking the data chunks as chunk id is unique.
I was thinking some strategy to let follower to send out a Kafka delete or Kafka message with empty value for the previous key after consuming a new `Global RT DIV message`, and if we don't do the cleanup, the size of Kafka topic might grow a lot depending on the  sending frequency.
Maybe we don't need to implement such cleanup in the MVP, but I think eventually, we need some way to clean them up from the version topics.

internal/venice-common/src/main/resources/avro/GlobalRtDivState/v1/GlobalRtDivState.avsc
gaojieliu: @sushantmane 
I recalled you did some experiment regarding referring existing schema definition, and how does that go?

sushantmane: At compile time, everything worked fine, but at runtime, I ran into schema not found exceptions. I didn’t look into it further and chose to stick with the existing pattern.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -1904,6 +1906,10 @@ protected boolean shouldProcessRecord(PubSubMessage<KafkaKey, KafkaMessageEnvelo
               }
             }
           }
+          // Global RT DIV messages should be completely ignored when leader is consuming from remote version topic
+          if (record.getKey().isGlobalRtDiv()) {
```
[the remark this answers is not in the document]
KaiSernLim: We don't think so, and this is just a sanity check. I'll update the comment and error message.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -467,6 +472,8 @@ public StoreIngestionTask(
         new IngestionNotificationDispatcher(notifiers, kafkaVersionTopic, isCurrentVersion);
     this.missingSOPCheckExecutor.execute(() -> waitForStateVersion(kafkaVersionTopic));
     this.chunkAssembler = new ChunkAssembler(storeName);
+    this.divChunkAssembler =
+        builder.getDivChunkAssembler() != null ? builder.getDivChunkAssembler() : new ChunkAssembler(storeName);
```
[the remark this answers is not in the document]
KaiSernLim: I also really wanted to reuse the instance from the builder, but there were some concerns about a nonzero chance of key collision with the records from record transformer.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -2397,6 +2432,12 @@ protected boolean shouldProcessRecord(PubSubMessage<KafkaKey, KafkaMessageEnvelo
       return false;
     }
 
+    // Global RT DIV messages should only be in version topics, not realtime topics. Skip it and log a warning.
+    if (record.getKey().isGlobalRtDiv() && record.getTopic().isRealTime()) {
```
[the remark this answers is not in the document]
KaiSernLim: There shouldn't ever be a scenario in which that happens, and this clause is a sanity check for that.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTaskFactory.java
```diff
@@ -336,5 +338,13 @@ public Builder setAAWCWorkLoadProcessingThreadPool(ExecutorService executorServi
     public ExecutorService getAAWCWorkLoadProcessingThreadPool() {
       return this.aaWCWorkLoadProcessingThreadPool;
     }
+
+    public Builder setDivChunkAssembler(ChunkAssembler divChunkAssembler) {
```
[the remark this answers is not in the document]
KaiSernLim: That's a good point. This setter is mainly for testing purposes, so perhaps it would be better to move the setter onto the SIT class itself as a test method.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -3786,7 +3827,7 @@ private int processKafkaDataMessage(
   private void waitReadyToProcessRecord(PubSubMessage<KafkaKey, KafkaMessageEnvelope, Long> record)
       throws InterruptedException {
     KafkaMessageEnvelope kafkaValue = record.getValue();
-    if (record.getKey().isControlMessage() || kafkaValue == null) {
+    if (record.getKey().isControlMessage() || record.getKey().isGlobalRtDiv() || kafkaValue == null) {
```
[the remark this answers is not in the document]
KaiSernLim: Discussed offline.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -1189,6 +1202,28 @@ private int handleSingleMessage(
...
+    Object assembledObject = divChunkAssembler.bufferAndAssembleRecord(
+        record.getTopicPartition(),
+        put.getSchemaId(),
+        key.getKey(),
+        put.getPutValue(),
+        record.getOffset(),
+        put.getSchemaId(),
+        new NoopCompressor(),
+        (valueBytes) -> GLOBAL_RT_DIV_STATE.getSerializer()
+            .deserialize(ByteUtils.extractByteArray(valueBytes), GLOBAL_RT_DIV_STATE_SCHEMA_ID));
```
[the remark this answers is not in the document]
KaiSernLim: `ChunkAssembler` has been removed.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -1189,6 +1202,28 @@ private int handleSingleMessage(
     return record.getPayloadSize();
   }
 
+  void processGlobalRtDivMessage(PubSubMessage<KafkaKey, KafkaMessageEnvelope, Long> record) {
+    KafkaKey key = record.getKey();
+    KafkaMessageEnvelope value = record.getValue();
+    Put put = (Put) value.getPayloadUnion();
+
+    Object assembledObject = divChunkAssembler.bufferAndAssembleRecord(
```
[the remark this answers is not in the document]
KaiSernLim: Discussed offline. Will use the metadata storage partition in the storage engine rather than in-memory storage engine.

clients/da-vinci-client/src/test/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTaskTest.java
```diff
@@ -366,6 +367,7 @@ public static Object[][] sortedInputAndAAConfigProvider() {
   private AggVersionedDIVStats mockVersionedDIVStats;
   private AggVersionedIngestionStats mockVersionedStorageIngestionStats;
   private StoreIngestionTask storeIngestionTaskUnderTest;
+  private ChunkAssembler divChunkAssembler;
```
lluwm, line 370: not needed?

lluwm (commented):
Looks good overall. Left a few minor comments. `ChunkAssembler` has been mentioned in the description part, can we remove it if it's not part of this change anymore?

clients/da-vinci-client/src/test/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTaskTest.java
```diff
@@ -551,6 +553,7 @@ public void methodSetUp() throws Exception {
     mockRemoteKafkaConsumer = mock(PubSubConsumerAdapter.class);
     kafkaUrlToRecordsThrottler = new HashMap<>();
     kafkaClusterBasedRecordThrottler = new KafkaClusterBasedRecordThrottler(kafkaUrlToRecordsThrottler);
+    divChunkAssembler = mock(ChunkAssembler.class);
```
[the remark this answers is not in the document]
KaiSernLim: removed.

internal/venice-common/src/main/java/com/linkedin/venice/kafka/validation/Segment.java
```diff
@@ -291,6 +291,7 @@ public synchronized boolean addToCheckSum(KafkaKey key, KafkaMessageEnvelope mes
                     + controlMessage.getControlMessageType());
         }
       case PUT:
+      case GLOBAL_RT_DIV:
```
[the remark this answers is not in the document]
KaiSernLim: added
```
        // TODO: revisit to see if the GLOBAL_RT_DIV message is needed as part of the checksum
```

internal/venice-common/src/main/resources/avro/GlobalRtDivState/v1/GlobalRtDivState.avsc
```diff
@@ -0,0 +1,85 @@
+{
```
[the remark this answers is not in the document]
KaiSernLim: this should automatically go away when the other pr gets merged first.

lluwm (approved):
LGTM!

PR #1387, opened 2024-12-12
[server] Reset current version replica to ERROR on ingestion error

majisourav99:
## Reset current version replica to ERROR on ingestion error

For hybrid stores whose current version is ingesting the real time data from a real time topic, if there is any exception/error thrown during ingestion they would log error message, sit still and stop ingestion, resulting in stale replicas. 
The mitigation mechanism is for oncall to manually restart every impacted cluster, which is very labor-intensive. 
This PR tries to resolve that by marking such replicas to ERROR state, which later in a controller task to reset such error replicas, will try to initiate new state transition of those replicas which can possibly mitigate such stale replica issue.

## How was this PR tested?

GHCI

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
@@ -1453,18 +1453,20 @@ public abstract class StoreIngestionTask implements Runnable, Closeable {
               partitionException);
           // No need to reset again, clearing out the exception.
           partitionIngestionExceptionList.set(exceptionPartition, null);
-        } else if (!partitionConsumptionState.isCompletionReported()) {
-          reportError(partitionException.getMessage(), exceptionPartition, partitionException);
         } else {
-          LOGGER.error(
-              "Ignoring exception for replica: {} since it is already online. The replica will continue serving reads, but the data may be stale as it is not actively ingesting data. Please engage the Venice DEV team immediately.",
-              Utils.getReplicaId(kafkaVersionTopic, exceptionPartition),
-              partitionException);
-        }
-        // Unsubscribe the partition to avoid more damages.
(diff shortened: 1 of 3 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -1439,10 +1439,23 @@ private void processIngestionException() {
             // DaVinci is always a follower.
             subscribePartition(pubSubTopicPartition, false);
           }
+        } else if (isCurrentVersion.getAsBoolean() && resetErrorReplicaEnabled && !isDaVinciClient) {
```
xunyin8, line 1442: Is this the right place to mark replicas ERROR? IIUC we may not call `processIngestionException` if `lastStoreIngestionException` is set.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -1790,6 +1803,15 @@ private void reportError(
     } else {
       ingestionNotificationDispatcher.reportError(pcsList, message, consumerEx);
     }
+    // Set the replica state to ERROR so that the controller can attempt to reset the partition.
+    if (!isDaVinciClient && resetErrorReplicaEnabled) {
```
xunyin8, line 1807: Based on the title of this PR I'm assuming that for now we don't want to perform reset for non-current version. If that's the case then I don't think this will work for the following reasons:
1. Most caller of this method (e.g. in `processConsumerActions`) will call this only when the SIT version hasn't report completed yet. Meaning it's a future/rebalancing version and not a current serving version and reset is actually undesirable.
2. This method is also called from ST timeout in `StateModelIngestionProgressNotifier`. Again, not a currently serving version and we probably shouldn't reset it.

xunyin8 (approved):

PR #1388, opened 2024-12-12
[server] AAWC delete operation should auto-convert value-level RMD record

sixpluszero:
## [server] AAWC delete operation should auto-convert value-level RMD record
When we enabled partial update as store config, existing SIT will pick up this config after a server restart and becomes AAWC SIT and requires to work on field level TS.
We solved the similar issue in PUT operation in #1081 , where we will make no assumption on existing RMD record and optionally convert it from value-level to field-level. This PR completes the fix by adding the same logic to DELETE operation.

## How was this PR tested?
Add a new unit test and new integration test.


ZacAttack (approved):
synced a little bit on slack, I think this looks good.  Thanks!

PR #1394, opened 2024-12-14
[router] Alpini dead code removal

FelixGV:
Besides dead code removal, also moved a few functions from Aplini's CollectionUtil to Venice's CollectionUtils.

## How was this PR tested?

CI.


internal/alpini/netty4/alpini-netty4-base/src/main/java/com/linkedin/alpini/netty4/misc/NettyUtils.java
```diff
@@ -32,143 +16,8 @@ private NettyUtils() {
...
-    return mode().socketChannel();
-  }
-
-  public static Class<? extends ServerSocketChannel> serverSocketChannel() {
-    return mode().serverSocketChannel();
-  }
-
-  public static Class<? extends DatagramChannel> datagramChannel() {
-    return mode().datagramChannel();
+    return in.readBytes(length);
```
nisargthakkar, line 20: 😂

internal/venice-client-common/src/main/java/com/linkedin/venice/utils/CollectionUtils.java
```diff
@@ -112,4 +120,92 @@ public static Map<String, String> convertCharSequenceMapToStringMap(Map<CharSequ
   public static <K, V> Map<K, V> substituteEmptyMap(Map<K, V> map) {
     return map == null || map.isEmpty() ? Collections.emptyMap() : map;
   }
+
+  public static <K, V> MapBuilder<K, V> mapBuilder() {
```
nisargthakkar, line 124: This is a great addition. Thanks

nisargthakkar (dismissed):
LGTM

nisargthakkar:
I'm a bit confused why the Alpini checks didn't run. Maybe we should make them run on every PR instead of the path based filtering?

nisargthakkar:
They did run and now they seem to be quite quick. We should definitely enable them for every build

nisargthakkar:
The functional tests are looking sketchy. Looks like no actual test ran. We should take a look at what's up with that

FelixGV:
I was thinking of enabling them on every run, yeah... will have to check what happened to the functional ones.

nisargthakkar (approved):

PR #1395, opened 2024-12-14
[test] Speed up TestControllerClient#testControllerClientWithInvalidUrls

nisargthakkar:
## Speed up `TestControllerClient#testControllerClientWithInvalidUrls`

This test is validating that as long as at least one of the hosts in the Controller urls is healthy, the `ControllerClient` won't see a `ConnectException`. However, the `ControllerClient` will randomize the controller order in every request, it is not deterministic that a single run will ever attempt to communicate with the unavailable controller. Our tests setup to retry failing tests multiple times, and if any of them passes, it assumes the test is good. So, regressions in this logic could get missed by our team. To prevent this, the test runs the test for 100 iterations.

However, this leads to a case where the test is running for 13 minutes on CI hosts. To speed up the test, we do two things:
1. Reduce the number of iterations to 50
2. Parallelize the loop to concurrently attempt multiple times

## How was this PR tested?

This is a test-only change


sushantmane (approved):
LGTM! Thanks a lot, @nisargthakkar!

PR #1244, opened 2024-10-17
[WIP] [controller][client][admin-tool] Add gRPC support in venice-controllers

sushantmane:
## Add gRPC support in controller
Following APIs are now supported with gRPC
```protobuf
  rpc createStore(CreateStoreGrpcRequest) returns (CreateStoreGrpcResponse) {}
  rpc updateAclForStore(UpdateAclForStoreGrpcRequest) returns (UpdateAclForStoreGrpcResponse);
  rpc getAclForStore(GetAclForStoreGrpcRequest) returns (GetAclForStoreGrpcResponse);
  rpc deleteAclForStore(DeleteAclForStoreGrpcRequest) returns (DeleteAclForStoreGrpcResponse);
  rpc checkResourceCleanupForStoreCreation(CheckResourceCleanupForStoreCreationGrpcRequest) returns (CheckResourceCleanupForStoreCreationGrpcResponse) {}
  rpc getAdminCommandExecutionStatus(AdminCommandExecutionStatusGrpcRequest) returns (AdminCommandExecutionStatusGrpcResponse) {}
  rpc getLastSuccessfulAdminCommandExecutionId(LastSuccessfulAdminCommandExecutionGrpcRequest) returns (LastSuccessfulAdminCommandExecutionGrpcResponse) {}
  rpc getAdminTopicMetadata(AdminTopicMetadataGrpcRequest) returns (AdminTopicMetadataGrpcResponse) {}
  rpc updateAdminTopicMetadata(UpdateAdminTopicMetadataGrpcRequest) returns (UpdateAdminTopicMetadataGrpcResponse) {}
  rpc discoverClusterForStore(DiscoverClusterGrpcRequest) returns (DiscoverClusterGrpcResponse) {}
  rpc listBootstrappingVersions(ListBootstrappingVersionsGrpcRequest) returns (ListBootstrappingVersionsGrpcResponse) {}
  rpc getLeaderController(LeaderControllerGrpcRequest) returns (LeaderControllerGrpcResponse);
  rpc emptyPush(EmptyPushGrpcRequest) returns (EmptyPushGrpcResponse) {}
  rpc listStores(ListStoresGrpcRequest) returns (ListStoresGrpcResponse) {}
  rpc getStore(GetStoreGrpcRequest) returns (GetStoreGrpcResponse) {}
```

## How was this PR tested?
CI

## Does this PR introduce any user-facing changes?

- [x] No. You can skip the rest of this section.
[...]

PR #1396, opened 2024-12-16
[controller] Add support for gRPC in venice-controller - Part I

sushantmane:
## [controller] Add initial gRPC support for Venice Controller - Part I

This commit introduces foundational server-side code to support gRPC-based endpoints in the 
Venice Controller. To keep the change set manageable, only 3 out of the 15 endpoints from 
#1244 are included in this change.

Subsequent PRs will implement the remaining endpoints to facilitate incremental progress.

Key changes:
- Added server-side infrastructure for gRPC endpoints.
- Implemented 3 initial endpoints from the planned set of 15.

Future work:
- Add remaining gRPC endpoints in follow-up commits.

## How was this PR tested?

UTs and E2E tests


internal/venice-common/src/main/java/com/linkedin/venice/grpc/VeniceGrpcServer.java
```diff
@@ -74,27 +71,45 @@ private void initServerCredentials() {
...
+  public boolean isRunning() {
+    return !server.isShutdown();
   }
 
   public boolean isTerminated() {
     return server.isTerminated();
   }
 
+  private boolean isSecure() {
+    return !(credentials instanceof InsecureServerCredentials);
```
FelixGV, line 98: Note that `null instanceof Anything` is always false. Right now, it doesn't seem like `credentials` can be null, but if they could be (e.g., due to some regression), then `isSecure` would return true (which would probably just result in a NPE sometime later, but in any event, it could be confusing?).

IDK what types can `credentials` take so IDK if a positive `instanceof` check is possible, or else maybe having a dedicated boolean flag for this, set in the same code path which sets the `credentials` prop. Or we can just leave it like this if you think it's fine. I don't have a strong opinion either way.

internal/venice-common/src/main/proto/VeniceEchoService.proto
```diff
@@ -0,0 +1,29 @@
...
+import "VeniceStore.proto";
+
+option java_multiple_files = true;
+
+// Define a custom option for method types
+extend google.protobuf.MethodOptions {
+  string methodType = 50001; // Assign a unique field number
+}
+
+service VeniceEchoService {
```
FelixGV, line 17: What is this for? Just some kind of hello world? Or a health check?

internal/venice-common/src/main/proto/VeniceStore.proto
```diff
@@ -0,0 +1,172 @@
...
+  repeated VersionGrpc versions = 50;
+  optional HybridStoreConfigGrpc hybridStoreConfig = 51;
+  CompressionStrategyGrpc compressionStrategy = 52;
+  BackupStrategyGrpc backupStrategy = 53;
+  optional ETLStoreConfigGrpc etlStoreConfig = 54;
+  PartitionerConfigGrpc partitionerConfig = 55;
+  map<string, ViewConfigGrpc> viewConfigs = 56;
+}
+
+message VersionGrpc {
```
FelixGV, line 143: I have a data modeling thought... it's not strictly gRPC-related, and also applies to the JSON stuff we shove into Zookeeper, but since we're building this one from scratch, we may have the option to reconsider some past decisions and do better (while leaving the JSON/ZK stuff unchanged -- at least for now)...

So the thing is that we have a lot of properties duplicated across store-version configs and store configs. And when that is the case, we implicitly understand it to mean that changing this config in the store config will affect future versions, and that we cannot change it in an existing version as those are immutable.

I wonder if we could codify this within the data model itself, and by the same token eliminate some duplication as well. How? By eliminating all of these props from the store config, and instead having a `futureVersion` record as one of the fields of the store config, which would house all of these configs.

IDK if the above idea is clear, please LMK either way. I'd like to know WDYT.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceController.java
```diff
@@ -132,11 +151,26 @@ public VeniceController(VeniceControllerContext ctx) {
...
+    this.pubSubTopicRepository = new PubSubTopicRepository();
+    this.controllerService = createControllerService();
+    this.adminServer = createAdminServer(false);
+    this.secureAdminServer = sslEnabled ? createAdminServer(true) : null;
+    this.topicCleanupService = createTopicCleanupService();
+    this.storeBackupVersionCleanupService = createStoreBackupVersionCleanupService();
+    this.disabledPartitionEnablerService = createDisabledPartitionEnablerService();
+    this.unusedValueSchemaCleanupService = createUnusedValueSchemaCleanupService();
+    this.storeGraveyardCleanupService = createStoreGraveyardCleanupService();
+    this.systemStoreRepairService = createSystemStoreRepairService();
```
FelixGV, line 163: I like this kind of change. It should allow us to mark all/most of these properties final. Could you do that?

services/venice-controller/src/main/java/com/linkedin/venice/controller/server/ClusterDiscovery.java
```diff
@@ -6,24 +6,21 @@
 import com.linkedin.venice.HttpConstants;
 import com.linkedin.venice.controller.Admin;
 import com.linkedin.venice.controllerapi.D2ServiceDiscoveryResponse;
-import com.linkedin.venice.utils.Pair;
+import com.linkedin.venice.controllerapi.request.ClusterDiscoveryRequest;
```
FelixGV, line 9: Thanks, good riddance!

services/venice-controller/src/main/java/com/linkedin/venice/controller/server/ControllerRoutes.java
```diff
@@ -64,13 +67,7 @@ public Route getLeaderController(Admin admin) {
...
         AdminSparkServer.validateParams(request, LEADER_CONTROLLER.getParams(), admin);
         String cluster = request.queryParams(CLUSTER);
-        responseObject.setCluster(cluster);
-        Instance leaderController = admin.getLeaderController(cluster);
-        responseObject.setUrl(leaderController.getUrl(isSslEnabled()));
-        if (leaderController.getPort() != leaderController.getSslPort()) {
-          // Controller is SSL Enabled
-          responseObject.setSecureUrl(leaderController.getUrl(true));
-        }
+        requestHandler.getLeaderController(new ControllerRequest(cluster), responseObject);
```
FelixGV, line 70: So this function will populate the `responseObject`? Can we not call it a getter if that is the case? e.g., `populateLeaderController`, perhaps?

services/venice-controller/src/main/java/com/linkedin/venice/controller/server/VeniceControllerGrpcServiceImpl.java
```diff
@@ -0,0 +1,183 @@
...
+      NewStoreResponse response = new NewStoreResponse();
+      requestHandler.createStore(request, response);
+
+      // Convert the internal response object to the gRPC response object and send the gRPC response
+      responseObserver.onNext(
+          CreateStoreGrpcResponse.newBuilder()
+              .setClusterStoreInfo(getClusterStoreGrpcInfo(response))
+              .setOwner(response.getOwner())
+              .build());
+      responseObserver.onCompleted();
```
FelixGV, line 162: I wonder if instead of having an intermediate POJO representation in the form of `CreateNewStoreRequest` and `NewStoreResponse`, could we use the gRPC generated classes, `CreateStoreGrpcRequest` and `CreateStoreGrpcResponse `, directly as our POJOs?

I think the intermediate representations can make sense if in the long-term we plan to _not_ migrate fully to gRPC, and/or support additional protocols (i.e., besides gRPC and REST, some third one).

But since the long-term direction is to go all in on gRPC, I think it would be fine to use the gRPC classes as our basic POJOs, and do this kind of marshalling only in the REST bridging code (as a temporary stop gap, during the migration). Then, when we're done with the migration we could delete the REST code and the remainder would be very lean, no marshalling into separate objects, etc.

In a nutshell, this means:

1. `requestHandler.createStore(request, response);` would take the gRPC generated classes as param types.
2. This class here should be greatly simplified.
3. The REST class should be similar in size as in this PR, but transferring state into the gRPC POJOs instead of our own POJOs.

WDYT?

FelixGV (commented):
Finished my first pass on this PR... left a few questions and comments. Thanks for the hard work on this! Exciting stuff!

internal/venice-common/src/main/java/com/linkedin/venice/grpc/VeniceGrpcServer.java
sushantmane: Here `credentials` can not be null as always set it to some value in the constructor of `VeniceGrpcServer`.  But I agree this could be confusing hence I'v updated the code to use dedicated boolean flag.

internal/venice-common/src/main/proto/VeniceEchoService.proto
sushantmane: Ya, this is hello world kind of service that is only used for testing of `VeniceGrpcServer` class

services/venice-controller/src/main/java/com/linkedin/venice/controller/server/VeniceControllerGrpcServiceImpl.java
sushantmane: We discussed this during the AR, and Nisarg also suggested the same approach as you. However, my personal preference is to avoid mixing gRPC-generated classes throughout the codebase for two reasons:
	1.	If we decide to continue supporting HTTP/REST in the future, we would need to maintain HTTP :left_right_arrow: gRPC conversion logic
	2.	Using gRPC-generated classes directly in handlers would make unit testing more challenging since most of these generated classes are final

internal/venice-common/src/main/proto/VeniceStore.proto
sushantmane: I like this idea a lot. I can update the PR to reflect this

internal/venice-common/src/main/proto/VeniceEchoService.proto
sushantmane: Replaced this with controller service. `VeniceEchoService` will be removed since we no longer need it.

internal/venice-common/src/main/proto/VeniceStore.proto
sushantmane: I'll be adopting this in my next PR as I've removed this proto def as it is not used yet.

services/venice-controller/src/main/java/com/linkedin/venice/controller/server/VeniceControllerGrpcServiceImpl.java
sushantmane: Updated to use grpc request/response instead of POJOs. Please take another look. Thanks.

FelixGV (approved):

sushantmane:
Thanks, @FelixGV!

PR #1398, opened 2024-12-16
[changelog] Clean up consumer state on changelog client

ZacAttack:
## [chanelog] Clean up consumer state on changelog client
And other protections like subscribing to an empty set of partitions.

Resolves #XXX


KaiSernLim (approved):
⛴️

PR #1401, opened 2024-12-17
[CI] Remove checks for test reports to avoid CI failures

sushantmane:
## Remove checks for test reports to avoid CI failures
- Replaced test reporter checks with annotation-only execution.
- Failed and flaky tests will now appear under the “Summary” tab and in the “Files Changed” tab (at the end for tests in unchanged files).
- Removed Gradle summary reporting.
 
 
image

 
image

### DIFF
--- a/.github/workflows/UnitTests-core.yml
+++ b/.github/workflows/UnitTests-core.yml
@@ -34,6 +34,7 @@ jobs:
           java-version: ${{ matrix.jdk }}
           distribution: 'temurin'
           cache: 'gradle'
+          add-job-summary: never
       - shell: bash
         run: |
           git remote set-head origin --auto
--- a/.github/workflows/UnitTests-core.yml
+++ b/.github/workflows/UnitTests-core.yml
@@ -42,6 +42,8 @@ jobs:
           git fetch upstream
       - name: Setup Gradle
         uses: gradle/actions/setup-gradle@v4
+        with:
+          add-job-summary: never
       - name: Run Unit Tests with Code Coverage
         run: ./gradlew  -x :internal:venice-avro-compatibility-test:test  ${{ inputs.arg }}
       - name: Package Build Artifacts
--- a/.github/workflows/UnitTests-core.yml
+++ b/.github/workflows/UnitTests-core.yml
@@ -58,7 +58,7 @@ jobs:
         if: always()
         with:
          check_name: ${{ inputs.artifact_suffix }}-jdk${{ matrix.jdk }} Report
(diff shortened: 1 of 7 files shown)


.github/workflows/UnitTests-core.yml
```diff
@@ -39,6 +41,8 @@ jobs:
           git fetch upstream
       - name: Setup Gradle
         uses: gradle/actions/setup-gradle@v4
+        with:
```
sixpluszero, line 44: What does this mean?

.github/workflows/VeniceCI-CompatibilityTests.yml
```diff
@@ -30,6 +30,8 @@ jobs:
           git fetch upstream
       - name: Setup Gradle
         uses: gradle/actions/setup-gradle@v4
+        with:
```
sixpluszero, line 33: Also here

.github/workflows/VeniceCI-E2ETests.yml
```diff
@@ -64,14 +68,28 @@ jobs:
...
+          path: '**/TEST-*.xml'
+          fail-on-error: 'false'
+          max-annotations: '10'
+          list-tests: 'all'
+          list-suites: 'all'
+          reporter: java-junit
+      - name: Publish Test Report
+        continue-on-error: true
+        env:
+          NODE_OPTIONS: "--max_old_space_size=8192"
```
sixpluszero, line 82: Why do we need 8G on publish test report though

.github/workflows/UnitTests-core.yml
sushantmane: By default gradle adds summary to the `Summary` page. We don't really use this info so I disabled it to avoid clutter on `Summary` page.

.github/workflows/VeniceCI-CompatibilityTests.yml
sushantmane: Same as the above

.github/workflows/VeniceCI-E2ETests.yml
sushantmane: This ensures the plugin functions correctly by increasing the Node.js heap size. Previously, report generation would fail due to running out of memory, so we set NODE_OPTIONS: "--max_old_space_size=8192" to resolve the issue.

sixpluszero (approved):

sushantmane:
Thanks, @sixpluszero!

PR #1404, opened 2024-12-18
[controller] add a new config that, when set, will update real time topic during partition count update

arjun4084346:
## update real time topic name, if needed during partition count update, when config `controller.enable.hybrid.store.partition.count.update` is set

With this PR, during partition count update, we will set the real time topic name if the partition count update is requested.
This is a part of the bigger project to achieve re-partitioning in real time topic and because there is a lot of work remaining, we will control this behavior through a config `controller.enable.hybrid.store.partition.count.update`; the default value of this config will be `false`

Resolves #XXX


services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -2780,7 +2781,8 @@ private Pair<Boolean, Version> addVersion(
                   && store.getHybridStoreConfig().getDataReplicationPolicy() == DataReplicationPolicy.AGGREGATE)
                   || store.isIncrementalPushEnabled())) {
                 // Create rt topic in parent colo if the store is aggregate mode hybrid store
-                PubSubTopic realTimeTopic = pubSubTopicRepository.getTopic(Utils.getRealTimeTopicName(store));
+                PubSubTopic realTimeTopic =
+                    pubSubTopicRepository.getTopic(Utils.getRealTimeTopicNameFromStoreConfig(store));
```
arjun4084346, line 2785: this change will create the correct rt topic

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -4322,6 +4336,20 @@ void preCheckStorePartitionCountUpdate(String clusterName, Store store, int newP
...
+  private void updateRealTimeTopicName(Store store) {
+    String oldRealTimeTopicName = Utils.getRealTimeTopicName(store);
+    String newRealTimeTopicName;
+    PubSubTopic newRealTimeTopic;
+
+    do {
+      newRealTimeTopicName = Utils.createNewRealTimeTopicName(oldRealTimeTopicName);
+      newRealTimeTopic = getPubSubTopicRepository().getTopic(newRealTimeTopicName);
+      oldRealTimeTopicName = newRealTimeTopicName;
+    } while (getTopicManager().containsTopic(newRealTimeTopic));
```
sushantmane, line 4348: Who's expected to create this new real time topic?

internal/venice-common/src/main/java/com/linkedin/venice/utils/Utils.java
```diff
@@ -632,6 +644,29 @@ private static String getRealTimeTopicNameIfEmpty(String realTimeTopicName, Stri
     return Strings.isBlank(realTimeTopicName) ? composeRealTimeTopic(storeName) : realTimeTopicName;
   }
 
+  public static String createNewRealTimeTopicName(String oldRealTimeTopicName) {
```
sushantmane, line 647: Can we define a standard format for topic names. This seems frail  approach to me. If we want to use versioning for RT topics then shouldn't we store those in store config? Also how would we know the partition count of RTs without querying the PubSub?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
arjun4084346: RealTimeTopicSwitcher I believe. This piece is just storing the name that is to be used the next time rt topic is created, whomsoever.

internal/venice-common/src/main/java/com/linkedin/venice/utils/Utils.java
arjun4084346: This is quite deterministic format, do you have a better way?
I am not really trying to support versioning of RT in the sense that at one time only one rt should exist, unless a replicator is running, in which case it should be maximum 2 topics at a time.
Yes, querying the pubsub topic is the way to know that, is it not a good way? We would not require to make a lot of such queries.

internal/venice-common/src/test/java/com/linkedin/venice/utils/UtilsTest.java
```diff
@@ -349,6 +350,59 @@ void testGetRealTimeTopicNameWithNonHybridVersion() {
...
+    // Define behavior
+    when(store.getHybridStoreConfig()).thenReturn(hybridStoreConfig);
+    when(store.getName()).thenReturn("test-store");
+    when(hybridStoreConfig.getRealTimeTopicName()).thenReturn("real-time-topic");
+
+    // Test
+    String result = Utils.getRealTimeTopicNameFromStoreConfig(store);
+
+    // Validate
+    assertEquals("real-time-topic", result);
```
sushantmane, line 368: in asserts `expectation` comes as second arg and we should follow the convention otherwise it's confusing to test failures. 
```
assertEquals(String actual, String expected)
```

Let's fix all the places to follow this pattern

internal/venice-common/src/main/java/com/linkedin/venice/utils/Utils.java
sushantmane: > This is quite deterministic format, do you have a better way?

If we take `storeName_v1` as a store name and `storeName_v11_v55_rt` as the old topic, the new expected RT topic name for this store should be `storeName_v11_v56_rt`, correct? However, if I understand correctly, this method would return `storeName_v11_v55_v2_rt`. Is that the intended behavior? It doesn’t seem right to me.


Maybe you want something like this:
```java
public static String createNewRealTimeTopicName(String oldRealTimeTopicName) {
    // Validate input
    if (oldRealTimeTopicName == null || !oldRealTimeTopicName.endsWith(Version.REAL_TIME_TOPIC_SUFFIX)) {
        throw new IllegalArgumentException("Invalid old name format");
    }

    // Find the base name by removing the suffix
    int suffixLength = Version.REAL_TIME_TOPIC_SUFFIX.length();
    String base = oldRealTimeTopicName.substring(0, oldRealTimeTopicName.length() - suffixLength);

    // Locate the last version separator "_v" in the base
    int versionSeparatorIndex = base.lastIndexOf("_v");
    if (versionSeparatorIndex > -1 && versionSeparatorIndex < base.length() - 2) {
        // Extract and increment the version
        String versionStr = base.substring(versionSeparatorIndex + 2);
        int version = Integer.parseInt(versionStr) + 1;
        base = base.substring(0, versionSeparatorIndex) + "_v" + version;
    } else {
        // Start with version 2 if no valid version is present
        base = base + "_v2";
    }

    return base + Version.REAL_TIME_TOPIC_SUFFIX;
}
```

sushantmane: > Yes, querying the pubsub topic is the way to know that, is it not a good way? We would not require to make a lot of such queries.


I should we should have our own mapping of partition count to rt name somewhere. It could be either via convention or some mapping somewhere in the store config

sushantmane (commented):
Left few more comments but I need to do one more pass

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
sushantmane: So are we going to block update store command until topic name is created?

sushantmane (changes requested):

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
arjun4084346: yes. is `containsTopic` that expensive? I am expecting only one call of this.

internal/venice-common/src/main/java/com/linkedin/venice/utils/Utils.java
```diff
@@ -545,6 +545,18 @@ public static String composeRealTimeTopic(String storeName) {
...
 
+  public static String getRealTimeTopicNameFromStoreConfig(Store store) {
+    HybridStoreConfig hybridStoreConfig = store.getHybridStoreConfig();
+    String storeName = store.getName();
+
+    if (hybridStoreConfig != null) {
+      String realTimeTopicName = hybridStoreConfig.getRealTimeTopicName();
+      return getRealTimeTopicNameIfEmpty(realTimeTopicName, storeName);
+    } else {
+      return composeRealTimeTopic(storeName);
```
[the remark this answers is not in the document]
arjun4084346: you mean *is not hybrid, right. Ideally we should not return RT topic name, but legacy code returns RT topic name, so I just let it return the name.

internal/venice-common/src/main/java/com/linkedin/venice/utils/Utils.java
```diff
@@ -632,6 +644,29 @@ private static String getRealTimeTopicNameIfEmpty(String realTimeTopicName, Stri
     return Strings.isBlank(realTimeTopicName) ? composeRealTimeTopic(storeName) : realTimeTopicName;
```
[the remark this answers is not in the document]
arjun4084346: ah, i didnt mean to use log4j's , i meant guava's. but because venice does not use guava, let me use commons-lang, okay?

internal/venice-common/src/main/java/com/linkedin/venice/utils/Utils.java
arjun4084346: As of now, I see little usage of partition count. Maybe we can add it later if needed?

internal/venice-common/src/main/java/com/linkedin/venice/utils/Utils.java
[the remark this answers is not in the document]
arjun4084346: Ideally we should not, but that is the current behavior... whenever `composeRealTimeTopic` returns the name no matter rt exist or not, or the store is hybrid or not

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
sushantmane: 1. In general, we shouldn't block any op that is done in admin channel. 
2. What if `RealTimeTopicSwitcher` never creates new RT topic say due to ingestion error of the hybrid version?
3. Why not create topic when updating partition count?

sushantmane: >yes. is containsTopic that expensive? I am expecting only one call of this.
btw, this is waiting just for creating new name, not new topic

`containsTopic` is not that expensive but spinning wait part of this is expensive IMO.


>btw, this is waiting just for creating new name, not new topic
What's the need to use `getTopicManager().containsTopic(newRealTimeTopic)`?  This does not sound right to me

internal/venice-common/src/main/java/com/linkedin/venice/utils/Utils.java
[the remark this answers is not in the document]
sushantmane: Let's just add it to our `Utils` class?

internal/venice-common/src/main/java/com/linkedin/venice/utils/Utils.java
```diff
@@ -545,6 +544,18 @@ public static String composeRealTimeTopic(String storeName) {
     return storeName + Version.REAL_TIME_TOPIC_SUFFIX;
   }
 
+  public static String getRealTimeTopicNameFromStoreConfig(Store store) {
+    HybridStoreConfig hybridStoreConfig = store.getHybridStoreConfig();
+    String storeName = store.getName();
+
+    if (hybridStoreConfig != null) {
```
sushantmane, line 551: Can we rewrite this as
 ```java
 if (hybridStoreConfig != null && StringUtils.isNotBlank(hybridStoreConfig.getRealTimeTopicName())) { 
   return hybridStoreConfig.getRealTimeTopicName();
 } 
 return composeRealTimeTopic(storeName);
 
 And then we can remove `getRealTimeTopicNameIfEmpty` if it's not used anywhere else?

internal/venice-common/src/test/java/com/linkedin/venice/utils/UtilsTest.java
sushantmane, line 1: I still see a pattern where the order of `actual` and `expected` result is reversed in testng assertions. Can we please fix it to make debugging easier? 

```java
  /**
   * Asserts that two objects are equal. If they are not,
   * an AssertionError is thrown.
   * @param actual the actual value
   * @param expected the expected value
   */
  public static void assertEquals(Object actual, Object expected);
  ```

File: services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerMultiClusterConfig.java [hunk 8]
```diff
@@ -286,4 +286,8 @@ public long getServiceDiscoveryRegistrationRetryMS() {
   public List<String> getControllerInstanceTagList() {
     return getCommonConfig().getControllerInstanceTagList();
   }
+
+  public boolean getUpdateRealTimeTopic() {
```
sushantmane, line 290: Can we make it per cluster config to make rollout more granular?

internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
```diff
@@ -2356,4 +2356,6 @@ private ConfigKeys() {
 
   public static final String SERVER_DELETE_UNASSIGNED_PARTITIONS_ON_STARTUP =
       "server.delete.unassigned.partitions.on.startup";
+
+  public static final String UPDATE_REAL_TIME_TOPIC = "controller.enable.hybrid.store.partition.count.update";
```
sushantmane, line 2360: can we fix the var name as well to reflect config better?

internal/venice-common/src/main/java/com/linkedin/venice/utils/Utils.java
arjun4084346: it is being used in multiple places...

sushantmane: I see then it's fine to keep it

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerMultiClusterConfig.java
arjun4084346: it is defined in `[VeniceControllerClusterConfig.java` `https://github.com/linkedin/venice/pull/1404/files#diff-cea201d965375cda81510a0338bc355594295fd736cfff25579b9d9b97218c6a` 
what else to be done to make it per cluster config?

internal/venice-common/src/test/java/com/linkedin/venice/utils/UtilsTest.java
arjun4084346: i found two more such instances and fixed them

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerMultiClusterConfig.java
sushantmane: For per cluster we can't use `getCommonConfig().getUpdateRealTimeTopic();`; we need to get configs for that cluster first and then lookup  this field

e.g. `VeniceControllerClusterConfig clusterConfig = multiClusterConfigs.getControllerConfig(clusterName);`

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerClusterConfig.java
```diff
@@ -1767,6 +1770,10 @@ public int getDanglingTopicOccurrenceThresholdForCleanup() {
     return danglingTopicOccurrenceThresholdForCleanup;
   }
 
+  public boolean getUpdateRealTimeTopic() {
```
sushantmane, line 1773: Can we update the name of this method to `isHybridStorePartitionCountUpdateEnabled()`

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerMultiClusterConfig.java
arjun4084346: done

sushantmane:
Could you please update commit message and title to reflect the new changes? Thanks

internal/venice-common/src/test/java/com/linkedin/venice/utils/UtilsTest.java
```diff
@@ -393,4 +447,57 @@ public void testGetLeaderTopicFromPubSubTopic() {
...
+  }
+
+  @Test
+  void testValidOldNameStartingNewVersion() {
+    String oldName = "storeName_rt";
+    String expectedNewName = "storeName_v2_rt";
+
+    String result = Utils.createNewRealTimeTopicName(oldName);
+
+    assertEquals(expectedNewName, result);
```
sushantmane, line 478: Reverse order here as well?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerClusterConfig.java
```diff
@@ -1767,6 +1770,10 @@ public int getDanglingTopicOccurrenceThresholdForCleanup() {
     return danglingTopicOccurrenceThresholdForCleanup;
   }
 
+  public boolean isHybridStorePartitionCountUpdateEnabled() {
+    return updateRealTimeTopic;
```
sushantmane, line 1774: Same comment again: let's keep var names and method name consistent please. Thanks

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerMultiClusterConfig.java
```diff
@@ -286,4 +286,8 @@ public long getServiceDiscoveryRegistrationRetryMS() {
   public List<String> getControllerInstanceTagList() {
     return getCommonConfig().getControllerInstanceTagList();
   }
+
+  public boolean getUpdateRealTimeTopic(String clusterName) {
```
sushantmane, line 290: Do we need this method?  If yes let's update it's name to be consistent with the new name: isHybridStorePartitionCountUpdateEnabled

internal/venice-common/src/test/java/com/linkedin/venice/utils/UtilsTest.java
sushantmane: There is one more place as mentioned below. Let's fix it. Thanks

sushantmane (approved):
Thanks, @arjun4084346 !

PR #1405, opened 2024-12-18
[router] Made most VeniceRouterConfig fields final

FelixGV:
There are only two exceptions, related to throttling, which are allowed to be mutated post-init.

Miscellaneous:

- Deleted a dead config: router.compute.fast.avro.enabled
- Made VeniceProperties::getMap return an unmodifiable map.

N.B.: This is a carve out of #1367 to make that other PR smaller. It is the most trivial part of that PR.

## How was this PR tested?
New unit tests.

### DIFF
--- a/internal/venice-client-common/src/main/java/com/linkedin/venice/utils/VeniceProperties.java
+++ b/internal/venice-client-common/src/main/java/com/linkedin/venice/utils/VeniceProperties.java
@@ -427,26 +427,22 @@ public class VeniceProperties implements Serializable {
     if (!containsKey(key)) {
       throw new UndefinedPropertyException(key);
     }
-    Map<String, String> map = new HashMap<>();
     List<String> keyValuePairs = this.getList(key);
+    Map<String, String> map = new HashMap<>(keyValuePairs.size());
 
-    for (String pair: keyValuePairs) {
-      Pair<String, String> keyValuePair = splitAnEntryToKeyValuePair(pair);
-      map.put(keyValuePair.getFirst(), keyValuePair.getSecond());
+    for (String keyValuePair: keyValuePairs) {
+      // One entry could have multiple ":". For example, "<ID>:<Kafka URL>:<port>". In this case, we split the String by
+      // its first ":" so that we get key=<ID> and value=<Kafka URL>:<port>
+      int indexOfFirstColon = keyValuePair.indexOf(':');
+      if (indexOfFirstColon == -1) {
+        throw new VeniceException(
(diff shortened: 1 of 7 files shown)


internal/venice-client-common/src/main/java/com/linkedin/venice/utils/VeniceProperties.java
```diff
@@ -427,26 +427,22 @@ public Map<String, String> getMap(String key) {
...
-    return map;
-  }
-
-  private Pair<String, String> splitAnEntryToKeyValuePair(String entry) {
-    if (!entry.contains(":")) {
-      throw new VeniceException("Invalid config. Expect each entry to contain at least one \":\". Got: " + entry);
+    Map<String, String> map = new HashMap<>(keyValuePairs.size());
+
+    for (String keyValuePair: keyValuePairs) {
+      // One entry could have multiple ":". For example, "<ID>:<Kafka URL>:<port>". In this case, we split the String by
```
majisourav99, line 434: can we use `keyValuePair.split(":", 2)` to perform this ?

FelixGV: Yes, we could write it like:

```
      String[] keyValueArray = keyValuePair.split(":", 2);
      if (keyValueArray.length != 2) {
        throw new VeniceException(
            "Invalid config. Expect each entry to contain at least one \":\". Got: " + keyValuePair);
      }
      String mapKey = keyValueArray[0];
      String mapValue = keyValueArray[1];
      map.put(mapKey, mapValue);
```

Is it more clear? I am not sure...

majisourav99: I was expecting using `split` will make it more compact. But does not look like it :(

majisourav99 (approved):
LGTM. Thanks Felix

internal/venice-client-common/src/main/java/com/linkedin/venice/utils/VeniceProperties.java
FelixGV: Well it could be more compact if we inline the array access like so:

```
map.put(keyValueArray[0], keyValueArray[1]);
```

Although our line length limit would also allow inlining the previous function calls as well...

```
map.put(keyValuePair.substring(0, indexOfFirstColon), keyValuePair.substring(indexOfFirstColon + 1));
```

But either way, there still needs to be a check for whether the colon is completely missing...

FelixGV:
Thanks Sourav!

PR #1409, opened 2024-12-19
[controller][server] Remove SIT ready-to-serve check for A/A and non-AGG store during L/F transition

sixpluszero:
## [server] Remove SIT ready-to-serve check for A/A and non-AGG store during L/F transition

Ready-to-serve check happens in two threads today:
(1) Drainer: which is reasonable
(2) SIT thread: there are two sub-cases:
- During L/F transition: It is only for empty RT topic, and today we have HB for non-agg and A/A already, so it only applies to the lingering AGG system store. In this PR, it is checked against store data replication type, if it is AGG we will still have it. We will completely remove it once AGG mode is removed.
- Before subscribe, it has a pre-check for short circuit to complete, and this is fine, as it is before the subscribe.

Also, I noticed that there are a bunch of integration tests which **enables incremental push on NON-AA store**. This is totally invalid setup and after removing the additional RTS check, these tests start to fail/very flaky. 
To make sure we don't do this in test and prod, this PR added a new check in update store command in parent controller, which will fail loudly if the new update request is to set incremental push to be true on a non-A/A store. 
I have another thinking about enabling all the configs that are enabled in production to be enabled in the test suite, but I'd like to do it in another separate PR, so it is easier for reviewers. 

Beyonds that, fixed a flaky unit test that can throw NPE.
## How was this PR tested?

Add a sinlge unit test.

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/config/VeniceClusterConfig.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/config/VeniceClusterConfig.java
@@ -96,7 +96,6 @@ public class VeniceClusterConfig {
 
   public VeniceClusterConfig(VeniceProperties clusterProps, Map<String, Map<String, String>> kafkaClusterMap)
       throws ConfigurationException {
-    LOGGER.info("Input Kafka Cluster Map: {}", kafkaClusterMap);
     this.clusterName = clusterProps.getString(CLUSTER_NAME);
     this.zookeeperAddress = clusterProps.getString(ZOOKEEPER_ADDRESS);
 
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/config/VeniceClusterConfig.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/config/VeniceClusterConfig.java
@@ -96,6 +96,7 @@ public class VeniceClusterConfig {
 
   public VeniceClusterConfig(VeniceProperties clusterProps, Map<String, Map<String, String>> kafkaClusterMap)
       throws ConfigurationException {
+    LOGGER.info("Input Kafka Cluster Map: {}", kafkaClusterMap);
     this.clusterName = clusterProps.getString(CLUSTER_NAME);
     this.zookeeperAddress = clusterProps.getString(ZOOKEEPER_ADDRESS);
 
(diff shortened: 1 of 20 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -4566,7 +4568,22 @@ void setVersionRole(PartitionReplicaIngestionContext.VersionRole versionRole) {
...
+  boolean isDaVinciClient() {
     return isDaVinciClient;
   }
+
+  boolean isHybridAggregateMode() {
+    return hybridStoreConfig.isPresent()
+        && hybridStoreConfig.get().getDataReplicationPolicy().equals(DataReplicationPolicy.AGGREGATE);
+  }
+
+  ReadyToServeCheck getReadyToServerChecker() {
```
KaiSernLim, line 4580:
```suggestion
  ReadyToServeCheck getReadyToServeChecker() {
```
You must've been typing `'server'` too much.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
```diff
@@ -2768,6 +2768,18 @@ && getVeniceHelixAdmin().isHybrid(setStore.getHybridStoreConfig()) && setStore.g
...
 
+      /**
+       * Pre-flight check for incremental push config update. We only allow incremental push config to be turned on
+       * when store is A/A. Otherwise, we should fail store update.
+       */
+      if (setStore.hybridStoreConfig != null && setStore.incrementalPushEnabled
+          && !setStore.activeActiveReplicationEnabled) {
+        throw new VeniceHttpException(
+            HttpStatus.SC_BAD_REQUEST,
+            "Hybrid store config invalid. Cannot have incremental push enabled while A/A not enabled",
```
KaiSernLim, line 2779: Some failing tests will need to be updated.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/config/VeniceClusterConfig.java
```diff
@@ -96,6 +96,7 @@ public class VeniceClusterConfig {
 
   public VeniceClusterConfig(VeniceProperties clusterProps, Map<String, Map<String, String>> kafkaClusterMap)
       throws ConfigurationException {
+    LOGGER.info("Input Kafka Cluster Map: {}", kafkaClusterMap);
```
KaiSernLim, line 99: Was this a debug log?

KaiSernLim (approved):

clients/da-vinci-client/src/main/java/com/linkedin/davinci/config/VeniceClusterConfig.java
sixpluszero: Yes, I can remove it.

KaiSernLim (approved):

PR #1411, opened 2024-12-20
[build] Add checker framework for nullness type-checking

sushantmane:
## Add checker framework for nullness type-checking
We've been running into a lot of NPEs lately, so we decided to bring in the Checker Framework for nullness type-checking to help catch and prevent these issues. The checker is off by default and will be turned on gradually for each module. To enable it, pass the `-PskipCheckerFramework=false` flag when running Gradle.

Example:
```bash
./gradlew -PskipCheckerFramework=false compileJava
```

For more info about checker framework, see [Checker Framework](https://checkerframework.org)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -17,6 +17,7 @@ plugins {
   id 'me.champeau.jmh' version '0.6.7' apply false
   id 'io.github.lhotari.gradle-nar-plugin' version '0.5.1' apply false
   id 'com.google.protobuf' version '0.9.3' apply false
+  id 'org.checkerframework' version '0.6.47' // Checker Framework pluggable type-checking
 }
 
 apply from: "$rootDir/gradle/helper/git.gradle"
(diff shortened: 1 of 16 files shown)


FelixGV (approved):
Thanks a lot Sushant!

There is a unit test failure, though it looks like a hang. I have seen that before recently, so I wonder if we have a regression of some kind. We need to investigate the test logs to find out which test is causing the hang / time out.

But it seems clear this PR is not the culprit, so let's just re-run the task for now...

sushantmane:
> Thanks a lot Sushant!
> 
> 
> 
> There is a unit test failure, though it looks like a hang. I have seen that before recently, so I wonder if we have a regression of some kind. We need to investigate the test logs to find out which test is causing the hang / time out.
> 
> 
> 
> But it seems clear this PR is not the culprit, so let's just re-run the task for now...

Yes, I raised this issue with the committer when it first occurred. I also spent some time investigating but couldn’t pinpoint the actual problem.

PR #1414, opened 2024-12-20
[router] Improve HAR routing in Router

gaojieliu:
Update Venice Router HAR least-loaded algo to take group latency into account and here
are the reasons:
a. When the total qps is relatively low, the pending request count based load rebalancing
      algo doesn't work well as most of the time, the pending count is 0. One example: let us
      say, the avg latency of one group is 10ms, and if the group qps is 10, that means very likely
      only 10-20% of the time, there are some pending requests and even another group latency is
      much lower: 1-2ms, the faster group won't be selected, and the latency-based LB algo will kick
      in when the above case happens to prefer the faster group if the pending count is equal among all
      the groups.
b. Completely getting rid of pending request count based LB will result in another issue as the relatively
      slower group will get almost zero request.
c. A combination of pending request and latency will select the faster groups when all groups are busy or idle,
      and it will still send some amount of requests to the slower groups in case it is too idle and it will help
      bring back the slower group into the rotation when it is recovered from the slowness.

## How was this PR tested?
CI

### DIFF
--- a/services/venice-router/src/main/java/com/linkedin/venice/router/api/routing/helix/HelixGroupLeastLoadedStrategy.java
+++ b/services/venice-router/src/main/java/com/linkedin/venice/router/api/routing/helix/HelixGroupLeastLoadedStrategy.java
@@ -2,6 +2,7 @@ package com.linkedin.venice.router.api.routing.helix;
 
 import com.linkedin.alpini.base.concurrency.TimeoutProcessor;
 import com.linkedin.venice.exceptions.VeniceException;
+import com.linkedin.venice.router.stats.HelixGroupStats;
 import com.linkedin.venice.utils.Pair;
 import java.util.HashMap;
 import java.util.Map;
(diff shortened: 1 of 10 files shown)


services/venice-router/src/main/java/com/linkedin/venice/router/api/routing/helix/HelixGroupLeastLoadedStrategy.java
```diff
@@ -60,10 +61,21 @@ public int selectGroup(long requestId, int groupCount) {
       }
       for (int i = 0; i < groupCount; ++i) {
         int currentGroup = (i + startGroupId) % groupCount;
-        long currentGroupCounter = counters[currentGroup];
+        int currentGroupCounter = counters[currentGroup];
```
majisourav99, line 64: instead of synchronized can we use `CopyOnWriteArrayList` for counters array?

gaojieliu: Counter will be updated per request, so it means the update frequency is fairly high, and I think `CopyOnWriteArrayList` is more suitable for less-update use cases and the logic being executed here is very fast, `synchronized` won't introduce much overhead.

services/venice-router/src/main/java/com/linkedin/venice/router/api/routing/helix/HelixGroupLeastLoadedStrategy.java
```diff
@@ -60,10 +61,21 @@ public int selectGroup(long requestId, int groupCount) {
...
           smallestCounter = currentGroupCounter;
           leastLoadedGroup = currentGroup;
+          lowestAvgLatency = helixGroupStats.getGroupResponseWaitingTimeAvg(currentGroup);
+        } else if (currentGroupCounter == smallestCounter) {
+          double currentGroupAvgLatency = helixGroupStats.getGroupResponseWaitingTimeAvg(currentGroup);
+          /**
+           * Here we don't check whether {@link #currentGroupAvgLatency} is less than 0 or not, as when the group is not
+           * being used at all, the average latency will be -1.0.
+           */
+          if (currentGroupAvgLatency < lowestAvgLatency) {
```
majisourav99, line 75: I was wondering if we should do `<=` check to round-robin the load across groups?

gaojieliu: Latency is very subtle as it is `double` type, and the difference between different group won't be exactly same, so `<=` check won't make any difference.

gaojieliu: If the pending request counter is same among several different zones and latency is truly same, `startGroupId` will do the RR load balancing.

majisourav99 (approved):
LGTM  Thanks Gaojie!

PR #1415, opened 2025-01-02
[server] Use thread safe map for versioned stats in SN read quota

xunyin8:
## [server] Use thread safe map for versioned stats in SN read quota

`Int2ObjectOpenHashMap` is not thread safe and using it for `ServerReadQuotaUsageStats` could result in unexpected behaviors during race conditions since the map is accessed and modified by multiple triggers and threads:

1. Read traffic
2. CV/routing change events
3. Version creation and deletion events

Added a test and confirmed that it will hang/timeout with `Int2ObjectOpenHashMap` but succeed with a thread safe map like `VeniceConcurrentHashMap`

Also minor defensive code change in `getReadQuotaUsageRatio` to avoid NPE since the map could change in between check and access.

## How was this PR tested?
New unit test and existing integration tests

### DIFF
--- a/services/venice-server/src/main/java/com/linkedin/venice/stats/ServerReadQuotaUsageStats.java
+++ b/services/venice-server/src/main/java/com/linkedin/venice/stats/ServerReadQuotaUsageStats.java
@@ -2,13 +2,13 @@ package com.linkedin.venice.stats;
 
 import com.linkedin.venice.utils.SystemTime;
 import com.linkedin.venice.utils.Time;
+import com.linkedin.venice.utils.concurrent.VeniceConcurrentHashMap;
 import io.tehuti.metrics.MetricConfig;
 import io.tehuti.metrics.MetricsRepository;
 import io.tehuti.metrics.Sensor;
 import io.tehuti.metrics.stats.AsyncGauge;
 import io.tehuti.metrics.stats.Count;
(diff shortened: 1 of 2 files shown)


KaiSernLim (dismissed):

services/venice-server/src/main/java/com/linkedin/venice/stats/ServerReadQuotaUsageStats.java
```diff
@@ -31,7 +30,7 @@ public class ServerReadQuotaUsageStats extends AbstractVeniceStats {
   private final Sensor rejectedKPS; // rejected key per second
   private final Sensor allowedUnintentionallyKPS; // allowed KPS unintentionally due to error or insufficient info
   private final Sensor usageRatioSensor; // requested kps divided by nodes quota responsibility
-  private final Int2ObjectMap<ServerReadQuotaVersionedStats> versionedStats = new Int2ObjectOpenHashMap<>();
+  private final ConcurrentHashMap<Integer, ServerReadQuotaVersionedStats> versionedStats = new ConcurrentHashMap<>();
```
nisargthakkar, line 33: Let's use `VeniceConcurrentHashMap` because of issues with CHM in JDK8 (the javadoc in `VeniceConcurrentHashMap` has the details of the issue)

nisargthakkar (approved):

PR #1417, opened 2025-01-03
[Server] Fixed the connection tracking

gaojieliu:
The previous connection tracking is not working properly because of two reasons:
1. `ServerConnectionStatsHandler` is being executed in the sub-channel, which is measuring the QPS rather than the new connection rate.
2. The way to extract the identity from cert is not right, so router connectin count is always 0.

To fix the above issues, this PR:
1. Move `ServerConnectionStatsHandler` up to be executed in the parent channel.
2. Use the correct identity parser to extract the application name from the cert, so that it can differentiate Router connections from Fast Client.
3. In function: `channelActive`, we can't extract `SslHandler` as SSL hasn't kicked in yet when the connection is established, so this PR introduces an async way to track new connections util the SSL handshake is fully done.

This PR also adds a new metric: `connection_request` to track the new connection rate in case the above async tracking would miss some connections, such as connection doesn't have a valid cert.

## How was this PR tested?
CI

### DIFF
--- a/services/venice-server/src/main/java/com/linkedin/venice/listener/ServerConnectionStatsHandler.java
+++ b/services/venice-server/src/main/java/com/linkedin/venice/listener/ServerConnectionStatsHandler.java
@@ -1,74 +1,147 @@
 package com.linkedin.venice.listener;
 
+import com.linkedin.venice.authorization.IdentityParser;
 import com.linkedin.venice.stats.ServerConnectionStats;
-import com.linkedin.venice.utils.SslUtils;
+import com.linkedin.venice.utils.DaemonThreadFactory;
+import io.netty.channel.ChannelHandler;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.ChannelInboundHandlerAdapter;
 import io.netty.handler.ssl.SslHandler;
 import io.netty.util.Attribute;
 import io.netty.util.AttributeKey;
+import java.security.cert.Certificate;
 import java.security.cert.X509Certificate;
-import javax.net.ssl.SSLPeerUnverifiedException;
+import java.util.Comparator;
+import java.util.HashSet;
+import java.util.Set;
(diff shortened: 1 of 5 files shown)


services/venice-server/src/main/java/com/linkedin/venice/listener/ServerConnectionStatsHandler.java
```diff
@@ -1,74 +1,147 @@
...
+          LOGGER.error("Failed to extract principal name from ssl handler in function: channelInactive");
+        } else if (principalName.contains(routerPrincipalName)) {
+          serverConnectionStats.decrementRouterConnectionCount();
+        } else {
+          serverConnectionStats.decrementClientConnectionCount();
+        }
+      }
     }
+
+    newConnections.remove(ctx);
```
xunyin8, line 115: The `connectionScanner` is scheduled to run every second. In theory there is a chance we might miss processing some connections before it gets removed right? I'm also curious what happens to bad requests? e.g. ACL is bad or we decided to throw error. Do we close the connection right the way?

xunyin8 (commented):
Thanks Gaojie! Change looks good just have some questions.

services/venice-server/src/main/java/com/linkedin/venice/listener/ServerConnectionStatsHandler.java
gaojieliu: That is correct and if the connection can't be setup properly, such as client cert is bad, it won't be tracked.
For ACL errors, I don't think the connection will be closed and Server would just respond with an error message/code.

One enhancement we can do is to add a new metric to track the connection setup rate regardless of the sources, as in `channelActive` function, it doesn't have access to the client cert right away.

xunyin8 (approved):

xunyin8 (approved):

PR #1420, opened 2025-01-07
[test] Introduced NameRepository, StoreName and StoreVersionName

FelixGV:
The new StoreName and StoreVersionName are meant to be used as map keys for these types of entities, to provide a more strongly-typed option than Strings. This is a similar idea to the PutSubTopic and its repository.

There is also an efficiency benefit since the instances of these types must come from the NameRepository, which gives out shared (cached) instances. This in turn reduces garbage generation, and facilitates equality checks which, in the common case, can resolve via identity comparisons (though deep comparison also works correctly, in case the caching limit is reached).

In this commit, the new classes are only used in a unit test, and they will be integrated later.

N.B.: This is yet another carve out from #1367, in an attempt to make that other PR easier to digest... will rebase the other one after this one gets merged.

## How was this PR tested?

New unit test.


internal/venice-common/src/main/java/com/linkedin/venice/meta/NameRepository.java
```diff
@@ -0,0 +1,39 @@
+package com.linkedin.venice.meta;
+
+import com.github.benmanes.caffeine.cache.Caffeine;
+import com.github.benmanes.caffeine.cache.LoadingCache;
+
+
+/**
+ * A repository of {@link StoreName} and {@link StoreVersionName}, which are intended to be used as shared instances.
+ */
+public class NameRepository {
```
sushantmane, line 10: I wish we had used singleton for `PubSubTopicRepository` to avoid creating and passing it through layers of abstractions. I wonder if we can use singleton here. I know you had some reservations about singletons in the past but I forgot the reasoning.

sushantmane: P.S IMHO this is really good candidate for singleton 😄

majisourav99: The caller of this class can maintain to use it as a singleton class, making it singleton by design might block future unforeseen usages.

internal/venice-common/src/main/java/com/linkedin/venice/meta/StoreName.java
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  @Override
+  public int hashCode() {
+    return this.name.hashCode();
+  }
+
+  @Override
+  public String toString() {
+    return this.getClass().getSimpleName() + "(" + this.name + ")";
```
sushantmane, line 60: Can we just return `name`? This will reduce clutter in the logs when we use `StoreName` object directly

internal/venice-common/src/main/java/com/linkedin/venice/meta/StoreName.java
```diff
@@ -0,0 +1,62 @@
...
+public class StoreName {
+  private final String name;
+
+  /**
+   * Not intended to be called directly! Instead, use:
+   *
+   * {@link NameRepository#getStoreName(String)}
+   */
+  StoreName(String name) {
+    if (!Store.isValidStoreName(name)) {
```
sushantmane, line 31: Should we move `isValidStoreName` from `Store` to `StoreName` class?

internal/venice-common/src/main/java/com/linkedin/venice/meta/StoreVersionName.java
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  @Override
+  public int hashCode() {
+    return this.name.hashCode();
+  }
+
+  @Override
+  public String toString() {
+    return this.getClass().getSimpleName() + "(" + this.name + ")";
```
sushantmane, line 80: Same here: can we just return store-version string?

internal/venice-common/src/main/java/com/linkedin/venice/meta/NameRepository.java
sushantmane: >The caller of this class can maintain to use it as a singleton class, making it singleton by design might block future unforeseen usages.


While these are well-intentioned arguments, based on my experience with PubSubTopicRepository, neither of these concerns holds up in practice.

FelixGV: This is intended to be a singleton. That being said, there are multiple implementation details for achieving the singleton pattern, e.g.:

1. Make the constructor private, and have a static function returning a reference to a static property.
2. Use an enum (syntactically different, but ultimately equivalent to 1).
3. Let it be constructed, but choose to construct it only once per process, and pass around the handle to the one instance. (This is the approach I've chosen here, and also for `PubSubTopicRepository`.)

The reasons why I personally prefer the 3rd approach are:

1. Mutable static state is (IMHO) an anti-pattern.
2. Configurability is sketchy with the static approach. We need to somehow set the static state related to the configuration prior to the singleton getting initialized, or else provide a hook for the singleton to get re-initialized. And then if the code which performs the configuration does not run first, we get weird effects like one code path using the singleton (which causes the lazy init to kick in) which a latter code path needing to pass in the config either fails or triggers a re-init (both of which are weird in different ways).
3. The static/enum approaches force having a single reference in the whole JVM (well... maybe in weird multi-classloader scenarios that is not true, but let's ignore that). That is fine for production code, but IMHO not good for test code. If we want test scenarios to be as faithful as possible, then each independent component should have its own singleton, even though it may happen to be spun up within the same JVM.

For stateless singletons, I think the static/enum approaches are fine, but for stateful, I prefer approach 3.

LMK if that makes sense...

internal/venice-common/src/main/java/com/linkedin/venice/meta/StoreName.java
FelixGV: Hmm... is it a big clutter to wrap this into the class name? To me it seems more clear this way... For example, it's not uncommon that we have users who create a store called, e.g., `My_Store_v2`, which means that their store-version names will be `My_Store_v2_v2`, `My_Store_v2_v3`, etc. In that scenario, if you saw a store-version name you could probably guess that it's a store-version, but if you saw a store name, you might not know that it's a store, and mistakenly assume that it's a store-version...?

internal/venice-common/src/main/java/com/linkedin/venice/meta/StoreName.java
FelixGV: Good idea, I've done that in my latest commit.

internal/venice-common/src/main/java/com/linkedin/venice/meta/NameRepository.java
sushantmane: I agree with all the points in the context of a stateful scenario where the individual entities themselves are mutable. However, in this case, neither StoreName nor StoreVersionName are mutable, so I don’t believe this qualifies as stateful in that regard.


I see this as analogous to a string interning approach, and therefore, I prefer using either option 1 or 2.

internal/venice-common/src/main/java/com/linkedin/venice/meta/StoreName.java
sushantmane: I see. Personally, I've not found this causing much confusion during troubleshooting though.

internal/venice-common/src/main/java/com/linkedin/venice/meta/NameRepository.java
FelixGV: What about the configurability concern? Should we just make this un-configurable to side-step those issues?

sushantmane: Why is configurability a concern here? Why not use eager initialization?

FelixGV: Let's say the Venice router config contains a property for the max number of entries in this repository's cache (or any other setting). Then the router would need to somehow pass that config via a static function prior to the init taking place. Or another pattern in these situations is to make the static singleton configurable via environment variables. Both of these approaches seem a bit sketchy to me. Whereas with non-static state it's very clear and clean (IMHO).

Separately from the above, mocking the behavior of static stuff is also messy.

I guess I would like to flip the question: why is static better? Because needing to wire in the shared instance is too annoying? Or is there some other reason?

sushantmane: Thanks for the clarification. I noticed that all the configurations were hard-coded, so I didn’t initially consider this particular issue. However, I now understand that it’s a valid concern if the plan is to make it configurable.


>why is static better? Because needing to wire in the shared instance is too annoying? Or is there some other reason

I find passing it through multiple layers a bit inconvenient, though YMMV. Additionally, since the primary goal is to minimize allocations, having one global repository would achieve that objective more cleanly, IMO.

That said, I want to clarify that not using a singleton is entirely fine. My suggestion is simply based on my experience with PubSubTopicRepository.

sushantmane (approved):
Thanks, @FelixGV !

internal/venice-common/src/main/java/com/linkedin/venice/meta/NameRepository.java
FelixGV: Ok, thanks Sushant. I'd like to keep it this way for now, but if later on we decide that the configurability is unimportant, then I think it should be easy to change it to be static...

PR #1421, opened 2025-01-08
[controller] Add deferred version swap service to parent controller

misyel:

Add in a new background service, `DeferredVersionSwapService`, to the parent controller to perform version swap for the remaining regions after the specified wait time has passed in target region push

## How was this PR tested?

Unit tests for the new class

### DIFF
--- a/services/venice-controller/src/main/java/com/linkedin/venice/controller/DeferredVersionSwapService.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/controller/DeferredVersionSwapService.java
@@ -1,7 +1,10 @@
 package com.linkedin.venice.controller;
 
+import com.linkedin.venice.controllerapi.ControllerClient;
+import com.linkedin.venice.controllerapi.StoreResponse;
 import com.linkedin.venice.meta.Store;
 import com.linkedin.venice.meta.Version;
+import com.linkedin.venice.meta.VersionStatus;
 import com.linkedin.venice.pushmonitor.ExecutionStatus;
 import com.linkedin.venice.service.AbstractVeniceService;
 import com.linkedin.venice.utils.RegionUtils;
@@ -10,6 +13,7 @@ import com.linkedin.venice.utils.Time;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
@@ -71,12 +75,15 @@ public class DeferredVersionSwapService extends AbstractVeniceService {
     return remainingRegions;
   }
 
-  private int getTargetVersionFromTargetRegion(Map<String, Integer> coloToVersions, Set<String> targetRegions) {
-    int targetVersion = 0;
-    for (String targetRegion: targetRegions) {
-      targetVersion = coloToVersions.get(targetRegion);
-    }
-    return targetVersion;
+  private String getTargetRegion(Set<String> targetRegions) {
+    return targetRegions.iterator().next();
+  }
+
+  private StoreResponse getStoreForRegion(String clusterName, Set<String> targetRegions, String storeName) {
+    Map<String, ControllerClient> controllerClientMap =
+        veniceParentHelixAdmin.getVeniceHelixAdmin().getControllerClientMap(clusterName);
+    ControllerClient targetRegionControllerClient = controllerClientMap.get(getTargetRegion(targetRegions));
+    return targetRegionControllerClient.getStore(storeName);
   }
 
   private class DeferredVersionSwapTask implements Runnable {
@@ -86,6 +93,11 @@ public class DeferredVersionSwapService extends AbstractVeniceService {
         for (String cluster: allClusters) {
           List<Store> stores = veniceParentHelixAdmin.getAllStores(cluster);
           for (Store store: stores) {
+            int parentVersion = store.getLargestUsedVersionNumber();
+            if (store.getVersion(parentVersion).getStatus() == VersionStatus.ONLINE) {
+              continue;
+            }
+
             if (StringUtils.isEmpty(store.getTargetSwapRegion())) {
               continue;
             }
@@ -97,9 +109,18 @@ public class DeferredVersionSwapService extends AbstractVeniceService {
 
             // Do not perform version swap for davinci stores
             // TODO remove this check once DVC delayed ingestion is completed
-            int targetVersion = getTargetVersionFromTargetRegion(coloToVersions, targetRegions);
-            Version version = store.getVersion(targetVersion);
-            if (version.getIsDavinciHeartbeatReported()) {
+            int targetVersion = coloToVersions.get(getTargetRegion(targetRegions));
+            StoreResponse targetRegionStore = getStoreForRegion(cluster, targetRegions, store.getName());
+            Optional<Version> version = targetRegionStore.getStore().getVersion(targetVersion);
+            if (!version.isPresent()) {
+              LOGGER.info(
+                  "Unable to find version {} for store: {} in regions: {}",
+                  targetVersion,
+                  store.getName(),
+                  store.getTargetSwapRegion());
+            }
+
+            if (version.get().getIsDavinciHeartbeatReported()) {
               LOGGER.info(
                   "Skipping version swap for store: {} on version: {} as it is davinci",
                   store.getName(),
@@ -147,7 +168,7 @@ public class DeferredVersionSwapService extends AbstractVeniceService {
               continue;
             }
 
-            // TODO add check for postStoreVersionSwap() once it is implemented
+            // TODO add call for postStoreVersionSwap() once it is implemented
 
             String remainingRegionsString = String.join(",\\s*", remainingRegions);
             LOGGER.info(
--- a/services/venice-controller/src/main/java/com/linkedin/venice/controller/DeferredVersionSwapService.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/controller/DeferredVersionSwapService.java
@@ -135,9 +135,12 @@ public class DeferredVersionSwapService extends AbstractVeniceService {
                 continue;
               }
 
+              // If we have a cached push completion for this store, check that the waitTime has elapsed before
+              // proceeding further
               String storeName = store.getName();
+              String kafkaTopicName = Version.composeKafkaTopic(storeName, targetVersionNum);
               Set<String> targetRegions = RegionUtils.parseRegionsFilterList(store.getTargetSwapRegion());
-              Map<String, Long> storePushCompletionTimes = storePushCompletionTimeCache.getIfPresent(storeName);
+              Map<String, Long> storePushCompletionTimes = storePushCompletionTimeCache.getIfPresent(kafkaTopicName);
               if (storePushCompletionTimes != null) {
                 if (!didWaitTimeElapseInTargetRegions(
                     storePushCompletionTimes,
@@ -185,7 +188,6 @@ public class DeferredVersionSwapService extends AbstractVeniceService {
               }
 
               // Check that push is completed in target regions
-              String kafkaTopicName = Version.composeKafkaTopic(storeName, targetVersionNum);
               Admin.OfflinePushStatusInfo pushStatusInfo =
                   veniceParentHelixAdmin.getOffLinePushStatus(cluster, kafkaTopicName);
               boolean didPushCompleteInTargetRegions = true;
@@ -206,16 +208,19 @@ public class DeferredVersionSwapService extends AbstractVeniceService {
               }
 
               // Check that push is complete in non target regions
-              int numNonTargetRegionsCompleted = remainingRegions.size();
+              int numNonTargetRegionsFailed = 0;
+              Set<String> nonTargetRegionsCompleted = new HashSet<>();
               for (String remainingRegion: remainingRegions) {
                 String executionStatus = pushStatusInfo.getExtraInfo().get(remainingRegion);
-                if (!executionStatus.equals(ExecutionStatus.COMPLETED.toString())) {
-                  numNonTargetRegionsCompleted -= 1;
-                  LOGGER.info(
-                      "Push is not complete for store: {} on version: {} in a non target region: {}",
+                if (executionStatus.equals(ExecutionStatus.ERROR.toString())) {
+                  numNonTargetRegionsFailed += 1;
+                  LOGGER.warn(
+                      "Push has error status store: {} on version: {} in a non target region: {}",
                       storeName,
                       targetVersionNum,
                       remainingRegion);
+                } else if (executionStatus.equals(ExecutionStatus.COMPLETED.toString())) {
+                  nonTargetRegionsCompleted.add(remainingRegion);
                 }
               }
 
@@ -225,13 +230,23 @@ public class DeferredVersionSwapService extends AbstractVeniceService {
                   veniceParentHelixAdmin.getVeniceHelixAdmin().getHelixVeniceClusterResources(cluster);
               ReadWriteStoreRepository repository = resources.getStoreMetadataRepository();
               Store parentStore = repository.getStore(storeName);
-              if (numNonTargetRegionsCompleted < remainingRegions.size() / 2) {
+              if (numNonTargetRegionsFailed > remainingRegions.size() / 2) {
                 LOGGER.warn(
-                    "Skipping version swap for store: {} on version: {} as majority of non target regions are not complete",
+                    "Skipping version swap for store: {} on version: {} as majority of non target regions have failed",
                     storeName,
                     targetVersionNum);
                 parentStore.updateVersionStatus(targetVersionNum, ERROR);
                 repository.updateStore(parentStore);
+                continue;
+              }
+
+              // If the majority of the remaining regions have not completed their push yet, do not perform versions
+              // swap yet
+              if (nonTargetRegionsCompleted.size() < remainingRegions.size() / 2) {
+                LOGGER.info(
+                    "Skipping version swap for store: {} on version: {} as majority of non target regions have not completed their push",
+                    storeName,
+                    targetVersionNum);
                 continue;
               }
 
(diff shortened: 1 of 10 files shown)


services/venice-controller/src/main/java/com/linkedin/venice/controller/DeferredVersionSwapService.java
```diff
@@ -0,0 +1,147 @@
...
+
+  public DeferredVersionSwapService(Admin veniceHelixAdmin, VeniceControllerMultiClusterConfig multiClusterConfig) {
+    this(veniceHelixAdmin, multiClusterConfig, new SystemTime());
+  }
+
+  protected DeferredVersionSwapService(Admin admin, VeniceControllerMultiClusterConfig multiClusterConfig, Time time) {
+    this.admin = admin;
+    this.time = time;
+    this.allClusters = multiClusterConfig.getClusters();
+    this.deferredVersionSwapThread = new Thread(new DeferredVersionSwapTask(), "DeferredVersionSwapTask");
```
majisourav99, line 43: can we use `newSingleThreadScheduledExecutor` instead and `scheduleAtFixedRate` the `Runnable` in startInner ?

services/venice-controller/src/main/java/com/linkedin/venice/controller/DeferredVersionSwapService.java
```diff
@@ -0,0 +1,147 @@
...
+            if (StringUtils.isEmpty(store.getTargetSwapRegion())) {
+              continue;
+            }
+
+            Map<String, Integer> coloToVersions = admin.getCurrentVersionsForMultiColos(cluster, store.getName());
+            Set<String> targetRegions = RegionUtils.parseRegionsFilterList(store.getTargetSwapRegion());
+            Set<String> remainingRegions = getRegionsForVersionSwap(coloToVersions, targetRegions);
+
+            // Do not perform version swap for davinci stores
+            int targetVersion = store.getCurrentVersion();
```
majisourav99, line 89: We are going to run this in parent controller right ? in parent controller current version for a store is not properly set. Get the version from target region of the store

services/venice-controller/src/main/java/com/linkedin/venice/controller/DeferredVersionSwapService.java
```diff
@@ -0,0 +1,147 @@
...
+
+/**
+ * This service is in charge of swapping to a new version after a specified wait time in the remaining regions of a target region push if enabled.
+ * The wait time is specified through a store/version level config (target_swap_region_wait_time) and the default wait time is 60m.
+ */
+public class DeferredVersionSwapService extends AbstractVeniceService {
+  private final AtomicBoolean stop = new AtomicBoolean(false);
+  private final Set<String> allClusters;
+  private final VeniceControllerMultiClusterConfig veniceControllerMultiClusterConfig;
+  private final Admin admin;
```
majisourav99, line 29: can we pull in `VeniceParentHelixAdmin` instead of admin

services/venice-controller/src/main/java/com/linkedin/venice/controller/DeferredVersionSwapService.java
```diff
@@ -0,0 +1,147 @@
...
+            for (String targetRegion: targetRegions) {
+              String executionStatus = pushStatusInfo.getExtraInfo().get(targetRegion);
+              if (executionStatus != ExecutionStatus.COMPLETED.toString()) {
+                didPushCompleteInTargetRegions = false;
+              }
+            }
+
+            if (!didPushCompleteInTargetRegions) {
+              LOGGER.info(
+                  "Skipping version swap for store: {} on version: {} as push is not complete",
```
majisourav99, line 112: nit: add info on which region it did not complete

services/venice-controller/src/main/java/com/linkedin/venice/controller/DeferredVersionSwapService.java
```diff
@@ -0,0 +1,147 @@
...
+                  store.getName(),
+                  targetVersion);
+              continue;
+            }
+
+            // Check that waitTime has elapsed in target regions
+            boolean didWaitTimeElapseInTargetRegions = true;
+            for (String targetRegion: targetRegions) {
+              Long completionTime = pushStatusInfo.getExtraInfoUpdateTimestamp().get(targetRegion);
```
majisourav99, line 121: no need to box to `Long`, just use `long`

services/venice-controller/src/main/java/com/linkedin/venice/controller/DeferredVersionSwapService.java
```diff
@@ -0,0 +1,147 @@
...
+              LOGGER.info(
+                  "Skipping version swap for store: {} on version: {} as wait time: {} has not passed",
+                  store.getName(),
+                  targetVersion,
+                  store.getTargetSwapRegionWaitTime());
+              continue;
+            }
+
+            // Issue roll forward for remaining regions
+            for (String region: remainingRegions) {
```
majisourav99, line 138: I believe `rollForwardToFutureVersion` can take the remainingRegions as region filter parameter.  ie
`rollForwardToFutureVersion(cluster, storeName, remainingRegions)` should work.

services/venice-controller/src/main/java/com/linkedin/venice/controller/DeferredVersionSwapService.java
```diff
@@ -0,0 +1,147 @@
...
+
+            if (!didWaitTimeElapseInTargetRegions) {
+              LOGGER.info(
+                  "Skipping version swap for store: {} on version: {} as wait time: {} has not passed",
+                  store.getName(),
+                  targetVersion,
+                  store.getTargetSwapRegionWaitTime());
+              continue;
+            }
+
```
majisourav99, line 136: Also can you add the placeholder `postStoreVersionSwap` call here ? Right now it will just return `PROCEED` signal, later we will add the implementation in venice-backend to do actual validation of the version swap.

services/venice-controller/src/main/java/com/linkedin/venice/controller/DeferredVersionSwapService.java
```diff
@@ -0,0 +1,160 @@
...
+            if (StringUtils.isEmpty(store.getTargetSwapRegion())) {
+              continue;
+            }
+
+            Map<String, Integer> coloToVersions =
+                veniceParentHelixAdmin.getCurrentVersionsForMultiColos(cluster, store.getName());
+            Set<String> targetRegions = RegionUtils.parseRegionsFilterList(store.getTargetSwapRegion());
+            Set<String> remainingRegions = getRegionsForVersionSwap(coloToVersions, targetRegions);
+
+            // Do not perform version swap for davinci stores
```
majisourav99, line 98: Keep a TODO comment here as later we need to remove this check once DVC delayed ingestion work is done.

services/venice-controller/src/main/java/com/linkedin/venice/controller/DeferredVersionSwapService.java
misyel: Left a todo comment to add a call for `postStoreVersionSwap` since there is no implementation yet. Let me know if you meant something else by placeholder

services/venice-controller/src/main/java/com/linkedin/venice/controller/DeferredVersionSwapService.java
```diff
@@ -0,0 +1,184 @@
...
+
+  private class DeferredVersionSwapTask implements Runnable {
+    @Override
+    public void run() {
+      while (!stop.get()) {
+        for (String cluster: allClusters) {
+          List<Store> stores = veniceParentHelixAdmin.getAllStores(cluster);
+          for (Store store: stores) {
+            int parentVersion = store.getLargestUsedVersionNumber();
+            if (store.getVersion(parentVersion).getStatus() == VersionStatus.ONLINE) {
```
majisourav99, line 97: I believe we should only check STARTED in the loop

misyel: Do we need to account for PUSHED too?

services/venice-controller/src/main/java/com/linkedin/venice/controller/DeferredVersionSwapService.java
majisourav99: Thats also ok.

majisourav99 (commented):
Can you add an e2e integration test for it.

services/venice-controller/src/main/java/com/linkedin/venice/controller/DeferredVersionSwapService.java
```diff
@@ -0,0 +1,193 @@
...
+            Version targetVersion = store.getVersion(targetVersionNum);
+            if (targetVersion == null) {
+              continue;
+            }
+
+            if (targetVersion.getStatus() != VersionStatus.STARTED) {
+              continue;
+            }
+
+            if (StringUtils.isEmpty(store.getTargetSwapRegion())) {
```
majisourav99, line 103: minor: move this as first check

services/venice-controller/src/main/java/com/linkedin/venice/controller/DeferredVersionSwapService.java
```diff
@@ -0,0 +1,193 @@
...
+    Set<String> remainingRegions = new HashSet<>(candidateRegions.keySet());
+    remainingRegions.removeAll(targetRegions);
+    return remainingRegions;
+  }
+
+  private String getTargetRegion(Set<String> targetRegions) {
+    return targetRegions.iterator().next();
+  }
+
+  private StoreResponse getStoreForRegion(String clusterName, Set<String> targetRegions, String storeName) {
```
majisourav99, line 74: Should it return a Map of store responses?

services/venice-controller/src/main/java/com/linkedin/venice/controller/DeferredVersionSwapService.java
```diff
@@ -0,0 +1,193 @@
...
+
+            // Check that push is completed in target regions
+            String kafkaTopicName = Version.composeKafkaTopic(store.getName(), targetVersionNum);
+            Admin.OfflinePushStatusInfo pushStatusInfo =
+                veniceParentHelixAdmin.getOffLinePushStatus(cluster, kafkaTopicName);
+            boolean didPushCompleteInTargetRegions = true;
+            for (String targetRegion: targetRegions) {
+              String executionStatus = pushStatusInfo.getExtraInfo().get(targetRegion);
+              if (!executionStatus.equals(ExecutionStatus.COMPLETED.toString())) {
+                didPushCompleteInTargetRegions = false;
```
majisourav99, line 147: you can just continue here without needing this variable.

majisourav99: actually if you have multiple target regions then its ok. Are we planning to support that?

misyel: Yeah, I think we should support multiple regions as well. Currently, the user can input a comma separated list of regions to perform the target region push in & if we choose to not support it right now, we should go back and add a check to prevent the user from doing that

services/venice-controller/src/main/java/com/linkedin/venice/controller/DeferredVersionSwapService.java
misyel: It only returns one store response because we only need to get the store for one target region to check the dvc heartbeat status

services/venice-controller/src/test/java/com/linkedin/venice/controller/TestDeferredVersionSwapService.java
```diff
@@ -0,0 +1,228 @@
+package com.linkedin.venice.controller;
+
+import static org.mockito.Mockito.*;
```
majisourav99, line 3: nit: can we remove wildcard imports?

services/venice-controller/src/main/java/com/linkedin/venice/controller/DeferredVersionSwapService.java
```diff
@@ -0,0 +1,192 @@
...
+        veniceParentHelixAdmin.getVeniceHelixAdmin().getControllerClientMap(clusterName);
+    ControllerClient targetRegionControllerClient = controllerClientMap.get(getTargetRegion(targetRegions));
+    return targetRegionControllerClient.getStore(storeName);
+  }
+
+  private class DeferredVersionSwapTask implements Runnable {
+    @Override
+    public void run() {
+      while (!stop.get()) {
+        for (String cluster: allClusters) {
```
majisourav99, line 85: put a big try-catch block around this logic and add metric in the catch block to detect failed version swaps

services/venice-controller/src/main/java/com/linkedin/venice/controller/DeferredVersionSwapService.java
majisourav99: in that case can you change the function signature to take in only a region string instead of set of strings

services/venice-controller/src/main/java/com/linkedin/venice/controller/DeferredVersionSwapService.java
```diff
@@ -0,0 +1,201 @@
...
+              if (targetVersion == null) {
+                continue;
+              }
+
+              if (targetVersion.getStatus() != VersionStatus.STARTED) {
+                continue;
+              }
+
+              Map<String, Integer> coloToVersions =
+                  veniceParentHelixAdmin.getCurrentVersionsForMultiColos(cluster, store.getName());
```
majisourav99, line 112: We should get the lastest version status instead of current version status

misyel: Looking at the code again, it's ok for us to keep this as is. We only use the value from this to infer the remaining regions to perform the version swap in (`getRegionsForVersionSwap`). We use `getOffLinePushStatus` to verify whether the push is completed before checking if the wait time has elapsed

services/venice-controller/src/main/java/com/linkedin/venice/controller/DeferredVersionSwapService.java
```diff
@@ -0,0 +1,237 @@
...
+                continue;
+              }
+
+              int targetVersionNum = store.getLargestUsedVersionNumber();
+              Version targetVersion = store.getVersion(targetVersionNum);
+              if (targetVersion == null) {
+                continue;
+              }
+
+              if (targetVersion.getStatus() != VersionStatus.PUSHED) {
```
majisourav99, line 110: Can you add some comments here explaining why we need to check for PUSHED status and what it means

services/venice-controller/src/main/java/com/linkedin/venice/controller/DeferredVersionSwapService.java
```diff
@@ -0,0 +1,237 @@
...
+
+              // TODO add call for postStoreVersionSwap() once it is implemented
+
+              String remainingRegionsString = String.join(",\\s*", remainingRegions);
+              LOGGER
+                  .info("Issuing roll forward message for store: {} in regions: {}", storeName, remainingRegionsString);
+              veniceParentHelixAdmin.rollForwardToFutureVersion(cluster, storeName, remainingRegionsString);
+
+              // Update parent status to ONLINE
+              LOGGER.info(
```
majisourav99, line 215: nit: can you move the log after its updated ?

services/venice-controller/src/main/java/com/linkedin/venice/controller/DeferredVersionSwapService.java
```diff
@@ -0,0 +1,281 @@
...
+
+              if (!didPushCompleteInTargetRegions) {
+                continue;
+              }
+
+              // Check that push is complete in non target regions
+              int numNonTargetRegionsCompleted = remainingRegions.size();
+              for (String remainingRegion: remainingRegions) {
+                String executionStatus = pushStatusInfo.getExtraInfo().get(remainingRegion);
+                if (!executionStatus.equals(ExecutionStatus.COMPLETED.toString())) {
```
majisourav99, line 212: can you explicitly check for error status, they could be in other non-terminal states,

services/venice-controller/src/main/java/com/linkedin/venice/controller/DeferredVersionSwapService.java
```diff
@@ -0,0 +1,281 @@
...
+                    storeName,
+                    targetVersionNum,
+                    store.getTargetSwapRegionWaitTime());
+                storePushCompletionTimeCache.put(storeName, pushStatusInfo.getExtraInfoUpdateTimestamp());
+                continue;
+              }
+
+              // TODO add call for postStoreVersionSwap() once it is implemented
+
+              String remainingRegionsString = String.join(",\\s*", remainingRegions);
```
majisourav99, line 256: We need to filter out failed regions from the roll-forward command, we do not want to switch the version in those regions.

services/venice-controller/src/main/java/com/linkedin/venice/controller/DeferredVersionSwapService.java
```diff
@@ -0,0 +1,281 @@
...
+                  targetRegions,
+                  store.getTargetSwapRegionWaitTime());
+
+              if (!didWaitTimeElapseInTargetRegions) {
+                LOGGER.info(
+                    "Skipping version swap for store: {} on version: {} as wait time: {} has not passed",
+                    storeName,
+                    targetVersionNum,
+                    store.getTargetSwapRegionWaitTime());
+                storePushCompletionTimeCache.put(storeName, pushStatusInfo.getExtraInfoUpdateTimestamp());
```
majisourav99, line 250: can we store store_version instead of just store name as a key?

misyel: updated to use the kafka topic name to account for the store name & version number

services/venice-controller/src/main/java/com/linkedin/venice/controller/DeferredVersionSwapService.java
```diff
@@ -0,0 +1,305 @@
...
+                  nonTargetRegionsCompleted.add(remainingRegion);
+                }
+              }
+
+              // If the majority of the remaining regions have failed their push jobs, mark the version status as ERROR
+              // so that we don't check this store again for this version
+              HelixVeniceClusterResources resources =
+                  veniceParentHelixAdmin.getVeniceHelixAdmin().getHelixVeniceClusterResources(cluster);
+              ReadWriteStoreRepository repository = resources.getStoreMetadataRepository();
+              Store parentStore = repository.getStore(storeName);
```
majisourav99, line 239: we already have the parentStore as 'store' ?

majisourav99 (approved):
LGTM! Thanks a lot @misyel  for iterating over this PR

PR #1422, opened 2025-01-09
[server] Enhance the producer timestamp fetch logic to include both data and heartbeat messages.

haoxu07:

Update the topic metadata producer timestamp fetch logic to include heartbeat messages. Previously, the getProducerTimestampOfLastDataMessage function only considered data messages when retrieving the producer timestamp from the most recent messages. However, for real-time topics without data ingestion, the last messages could be heartbeat control messages, causing endless searching during the ready-to-serve check due to the continuous arrival of new heartbeat messages.


sushantmane (approved):
LGTM. Thanks, @haoxu07!

PR #1423, opened 2025-01-09
[doc] Update guide for quick start single dc

minhmo1620:
[doc] Update the guide for quickstart for single dc

- [x] Update the guide to mention installing docker-compose before step 2.
- [x] Update the docker compose command since the current one is not runnable.
- [x] Update the example since incremental data push updates for range 90-110, not 50-150.
- [x] Add print out data sample for single dc.

## How was this PR tested?
doc only, no test is required.


FelixGV (commented):
Thanks for contributing to docs health!!

FelixGV (approved):

PR #1424, opened 2025-01-09
[test] Test to validate fully compatiblity of RocksDB BlobDB integration

gaojieliu:
The test case validates the BlobDB in the following way:
1. Disable BlobDB and insert a bunch of entries and make sure no blob files are generated.
2. Enable BlobDB and insert more entries and validate there are some blob files are generated and validate all the entries added in 1&2.
3. Disable BlobDB again and insert more entries and make sure the the total number of blob files doesn't change.
4. Validate all the entries inserted in 1&2&3.

## How was this PR tested?
CI


FelixGV (approved):
Thanks Gaojie!

PR #1426, opened 2025-01-10
[server] Fix IndexOutOfBoundsException in memory usage in SBS

majisourav99:
## Fix IndexOutOfBoundsException in memory usage in SBS

Since the `drainerList` is allocated during startInner, initial metric calls to `StoreBufferServiceStats` would return `IndexOutOfBoundsException` in the code. This PR fixes that.

## How was this PR tested?

GHCI

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreBufferService.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreBufferService.java
@@ -65,6 +65,8 @@ public class StoreBufferService extends AbstractStoreBufferService {
 
   private final boolean isSorted;
 
+  private boolean isStarted = false;
+
   public StoreBufferService(
       int drainerNum,
       long bufferCapacityPerDrainer,
@@ -308,6 +310,7 @@ public class StoreBufferService extends AbstractStoreBufferService {
       drainerList.add(drainer);
     }
     this.executorService.shutdown();
+    isStarted = true;
     return true;
   }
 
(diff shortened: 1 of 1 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreBufferService.java
```diff
@@ -319,6 +322,7 @@ public void stopInner() throws Exception {
       this.executorService.shutdownNow();
       this.executorService.awaitTermination(10, TimeUnit.SECONDS);
     }
+    isStarted = false;
```
sushantmane, line 325: Should we move this to the beginning of stopInner?

majisourav99: Sure

sushantmane (approved):

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreBufferService.java
```diff
@@ -65,6 +65,8 @@ public class StoreBufferService extends AbstractStoreBufferService {
 
   private final boolean isSorted;
 
+  private boolean isStarted = false;
```
FelixGV, line 68: If there are multiple threads involved, then let's make this field volatile...

majisourav99: looks like SBC creates all the drainer threads inside the `startInner` and it gets called once

FelixGV: Sure, `isStarted` doesn't get called only from `startInner`, right?

majisourav99: ah right! good catch

FelixGV (approved):
Thanks!

PR #1428, opened 2025-01-10
[server][dvc] add table format info in the blob transfer

jingy-li:
## [server][dvc] add table format info in the blob transfer

There is an edge case where multiple file formats may coexist within a single cluster. For instance, DVC1 might use a plain-table format while DVC2 uses a block-based format. If DVC2 requests data from DVC1 during bootstrap, it could end up with a plain-table store in DVC2, leading to read issues.

This PR addresses the issue by:

- Receiver side: Including the table format in the bootstrap request URI. An example URI: `[0]""/[1]"store"/[2]"version"/[3]"partition/[4]"table format"`.

- Sender side: Storing the current snapshot format in the snapshot manager. When a bootstrap request arrives, the snapshot manager compares the existing snapshot format with the requested table format. If they do not match, it returns a 404 error to reject the blob transfer request.

## How was this PR tested?
Integration test "testBlobTransferThrowExceptionIfTableFormatNotMatch".

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/server/P2PFileTransferServerHandler.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/server/P2PFileTransferServerHandler.java
@@ -3,6 +3,7 @@ package com.linkedin.davinci.blobtransfer.server;
 import static com.linkedin.davinci.blobtransfer.BlobTransferUtils.BLOB_TRANSFER_COMPLETED;
 import static com.linkedin.davinci.blobtransfer.BlobTransferUtils.BLOB_TRANSFER_STATUS;
 import static com.linkedin.davinci.blobtransfer.BlobTransferUtils.BLOB_TRANSFER_TYPE;
+import static com.linkedin.davinci.blobtransfer.BlobTransferUtils.BlobTransferTableFormat;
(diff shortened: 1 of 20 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/server/P2PFileTransferServerHandler.java
```diff
@@ -121,6 +122,16 @@ protected void channelRead0(ChannelHandlerContext ctx, FullHttpRequest httpReque
           setupResponseAndFlush(HttpResponseStatus.NOT_FOUND, errBody, false, ctx);
           return;
         }
+
+        // Check the snapshot table format
+        BlobTransferTableFormat currentSnapshotTableFormat = blobSnapshotManager.getBlobTransferTableFormat();
+        if (!blobTransferRequest.getRequestTableFormat().name().equals(currentSnapshotTableFormat.name())) {
```
lusong64, line 128: would using enum equal be more efficient than using String comparison?

jingy-li: Good call, updated in the next commit. Thank you.

sixpluszero (approved):

PR #1431, opened 2025-01-13
[server] Do not send heartbeat is VW is closed

majisourav99:
## Do not send heartbeat is VW is closed

Do not send heartbeat is VW is closed as it will spam the logs without adding a value.
## How was this PR tested?

GHCI


internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
```diff
@@ -1895,6 +1895,12 @@ public CompletableFuture<PubSubProduceResult> sendHeartbeat(
       boolean addLeaderCompleteState,
       LeaderCompleteState leaderCompleteState,
       long originTimeStampMs) {
+    if (isClosed) {
+      CompletableFuture<PubSubProduceResult> future = new CompletableFuture<>();
+      future.completedFuture(null);
+      logger.warn("VeniceWriter already closed for partition " + topicPartition.getTopicName());
```
sixpluszero, line 1901: Can you tweak the name to be real topic partition id, instead of only topic name?

sixpluszero (approved):

PR #1432, opened 2025-01-14
[test] Add timeouts for tests in TestTimeScheduledThreadPoolExecutor and TestResourceRegistry

sushantmane:
## [test] Add timeouts for tests in TestTimeScheduledThreadPoolExecutor and TestResourceRegistry

Resolves #XXX

## How was this PR tested?

WIP


FelixGV (approved):
Does it need such a high timeout? What is the longest run time you've seen where it succeeded?

sushantmane:
> Does it need such a high timeout? What is the longest run time you've seen where it succeeded?


Maybe not. I added it because I'm unsure about the origin of the problem. So far, I haven’t observed any particular tests taking a significant amount of time (I've tried several iterations, but no luck reproducing the issue). I suspect that some test might be leaking non-daemon threads, which prevents either Gradle or the JVM stop/close.

PR #1433, opened 2025-01-14
[server][da-vinci] Skip SSTFileWriter for hybrid store when RocksDB blob mode is on

gaojieliu:
We would like to skip {@link com.linkedin.davinci.store.rocksdb.RocksDBSstFileWriter} for hybrid stores when RocksDB blob mode is enabled and here are the reasons:
1. Hybrid stores will use the same amount of MemTables eventually even not in batch processing phase.
2. SSTFileWriter + RocksDB blob mode will introduce additional space overhead in the following way: a. SSTFileWriter doesn't support RocksDB blob mode, which means even with blob enabled, SSTFileWriter will still write both key and value into the same SST file regardless of value size. b. When RocksDB ingests the generated SST files, it will put them in the bottom level. c. After finishing the batch portion, RocksDB won't use SSTFileWriter anymore and in the regular mode, when RocksDB blob mode is enabled, RocksDB will store the value in blob files when the value size is larger than the configured threshold, and this also means the LSM tree built by the real-time writes will be much smaller as it contains keys and smaller values/value pointers. d. As LSM tree is small, it is not easy to trigger a compaction in the bottom level (the bottom - 1 level needs to keep enough data to trigger the bottom-level compaction), so the staled entries in the bottom level will remain for a long time. e. RocksDB blob config tuning won't affect the large bottom-level SST files.
3. If we disable SSTFileWriter for hybrid stores when RocksDB blob mode is enabled, all the writes will go through MemTable and key/value separation logic will apply all the time, and blob related configs will apply to all the writes.

This change is backward compatible, which means the batch ingestion for hybrid store won't break when we roll out/back RocksDB blob mode.
1. When SOP is processed before deploying RocksDB blob mode, `sorted` in `StoreVersionState` will be `true` and blob mode will be ignorede by SSTFileWriter as it doesn't support key/value separation today, so rolling back/forward blob mode doesn't change SSTFileWriter behavior.
2. When SOP is processed after deploying RocksDB blob mode, `sorted` in `StoreVersionState` will be `false` and SSTFileWriter will be skipped, so all the writes will go through MemTable and even we rollback RocksDB blob mode at this moment, the ingestion won't use SSTFileWriter as the `sorted` flag has been set as `false` already.

## How was this PR tested?
CI

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
@@ -2898,7 +2898,33 @@ public abstract class StoreIngestionTask implements Runnable, Closeable {
     /*
      * Notify the underlying store engine about starting batch push.
      */
-    beginBatchWrite(partition, startOfPush.sorted, partitionConsumptionState);
+    final boolean sorted;
+    if (serverConfig.getRocksDBServerConfig().isBlobFilesEnabled() && isHybridMode()) {
+      /**
+       * We would like to skip {@link com.linkedin.davinci.store.rocksdb.RocksDBSstFileWriter} for hybrid stores
+       * when RocksDB blob mode is enabled and here are the reasons:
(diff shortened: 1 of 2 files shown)


FelixGV (approved):
Thanks Gaojie!

PR #1435, opened 2025-01-14
[duckdb] Create first implementation of DuckDB x DVRT

kvargha:

Create first implementation of DuckDB x DVRT

## How was this PR tested?

New integration test

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/BlockingDaVinciRecordTransformer.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/BlockingDaVinciRecordTransformer.java
@@ -51,12 +51,12 @@ public class BlockingDaVinciRecordTransformer<K, V, O> extends DaVinciRecordTran
     this.recordTransformer.processDelete(key);
   }
 
-  public void onStartVersionIngestion() {
-    this.recordTransformer.onStartVersionIngestion();
+  public void onStartVersionIngestion(boolean isCurrentVersion) {
+    this.recordTransformer.onStartVersionIngestion(isCurrentVersion);
     startLatch.countDown();
   }
 
-  public void onEndVersionIngestion() {
-    this.recordTransformer.onEndVersionIngestion();
+  public void onEndVersionIngestion(int currentVersion) {
+    this.recordTransformer.onEndVersionIngestion(currentVersion);
   }
 }
--- a/integrations/venice-duckdb/src/main/java/com/linkedin/venice/sql/DuckDBDaVinciRecordTransformer.java
+++ b/integrations/venice-duckdb/src/main/java/com/linkedin/venice/sql/DuckDBDaVinciRecordTransformer.java
@@ -10,7 +10,7 @@ import java.sql.DriverManager;
 import java.sql.PreparedStatement;
 import java.sql.SQLException;
 import java.sql.Statement;
-import java.util.HashSet;
+import java.util.Collections;
 import java.util.Set;
 import java.util.function.BiConsumer;
 import org.apache.avro.Schema;
@@ -20,19 +20,21 @@ import org.apache.avro.generic.GenericRecord;
 
 public class DuckDBDaVinciRecordTransformer extends DaVinciRecordTransformer<String, GenericRecord, GenericRecord> {
   private static final String duckDBFilePath = "my_database.duckdb";
+  // ToDo: Don't hardcode the table name. Get it from the storeName
   private static final String baseVersionTableName = "my_table_v";
+  private static final Set<String> primaryKeys = Collections.singleton("firstName");
+  private static final String deleteStatementTemplate = "DELETE FROM %s WHERE %s = '%s';";
+  private static final String createViewStatementTemplate =
+      "CREATE OR REPLACE VIEW current_version AS SELECT * FROM %s;";
+  private static final String dropTableStatementTemplate =
+      "CREATE OR REPLACE VIEW current_version AS SELECT * FROM %s;";
   private final String duckDBUrl;
   private final String versionTableName;
-
-  // Can only support on primary key atm
-  private static final String primaryKey = "firstName";
-  private final Set<String> primaryKeys = new HashSet<>();
 
   public DuckDBDaVinciRecordTransformer(int storeVersion, boolean storeRecordsInDaVinci, String baseDir) {
     super(storeVersion, storeRecordsInDaVinci);
     versionTableName = baseVersionTableName + storeVersion;
     duckDBUrl = "jdbc:duckdb:" + baseDir + "/" + duckDBFilePath;
-    primaryKeys.add(primaryKey);
   }
 
   @Override
@@ -70,6 +72,7 @@ public class DuckDBDaVinciRecordTransformer extends DaVinciRecordTransformer<Str
     Schema valueSchema = value.get().getSchema();
     String upsertStatement = AvroToSQL.upsertStatement(versionTableName, valueSchema);
 
+    // ToDo: Instead of creating a connection on every call, have a long-term connection. Maybe a connection pool?
     try (Connection connection = DriverManager.getConnection(duckDBUrl)) {
       BiConsumer<GenericRecord, PreparedStatement> upsertProcessor = AvroToSQL.upsertProcessor(valueSchema);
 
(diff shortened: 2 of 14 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/BlockingDaVinciRecordTransformer.java
```diff
@@ -51,12 +51,12 @@ public void processDelete(Lazy<K> key) {
     this.recordTransformer.processDelete(key);
   }
 
-  public void onStartVersionIngestion() {
-    this.recordTransformer.onStartVersionIngestion();
+  public void onStartVersionIngestion(boolean isCurrentVersion) {
```
ZacAttack, line 54: What is the situation where this isn't the current version?

integrations/venice-duckdb/src/main/java/com/linkedin/venice/sql/DuckDBDaVinciRecordTransformer.java
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  @Override
+  public Schema getKeySchema() {
+    return Schema.create(Schema.Type.STRING);
+  }
+
+  // ToDo: Put real schema
+  @Override
+  public Schema getOutputValueSchema() {
```
ZacAttack, line 45: Does this need to be addressed before we commit?

kvargha: I forgot to remove the ToDo. Previously the schema in there was incorrect, but it's fine now.

integrations/venice-duckdb/src/main/java/com/linkedin/venice/sql/DuckDBDaVinciRecordTransformer.java
```diff
@@ -0,0 +1,130 @@
...
+  private final String duckDBUrl;
+  private final String versionTableName;
+
+  // Can only support on primary key atm
+  private static final String primaryKey = "firstName";
+  private final Set<String> primaryKeys = new HashSet<>();
+
+  public DuckDBDaVinciRecordTransformer(int storeVersion, boolean storeRecordsInDaVinci, String baseDir) {
+    super(storeVersion, storeRecordsInDaVinci);
+    versionTableName = baseVersionTableName + storeVersion;
```
FelixGV, line 33: Don't we get the store name passed down to us in the Transformer API? Hardcoding to "my_table" is fine in the prototype, but I think this will fail once we have multiple tables (unless we create separate DuckDB instances for each, but that is less ideal since it means the multiple tables could not be joined or otherwise queried together). Anyway, can be just a TODO for now.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/BlockingDaVinciRecordTransformer.java
kvargha: Let's say the current version is 1, and the future version is 2. The table for both versions are created, and the `current_version` view is created for v1. When `onEndVersionIngestion` is invoked for v1, then the view will be changed to v2.

FelixGV (commented):
Great to see this! I left a few minor comments. For most of them I suggest just leaving TODOs. Let's fix these issues iteratively after we merge this. Thanks a lot!

integrations/venice-duckdb/src/main/java/com/linkedin/venice/sql/DuckDBDaVinciRecordTransformer.java
```diff
@@ -0,0 +1,130 @@
...
+    duckDBUrl = "jdbc:duckdb:" + baseDir + "/" + duckDBFilePath;
+    primaryKeys.add(primaryKey);
```
ZacAttack, line 35: Do we need to do this?  Couldn't we just have a static set with the first name?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/BlockingDaVinciRecordTransformer.java
```diff
@@ -51,12 +51,12 @@ public void processDelete(Lazy<K> key) {
...
+    this.recordTransformer.onStartVersionIngestion(isCurrentVersion);
     startLatch.countDown();
   }
 
-  public void onEndVersionIngestion() {
-    this.recordTransformer.onEndVersionIngestion();
+  public void onEndVersionIngestion(int currentVersion) {
+    this.recordTransformer.onEndVersionIngestion(currentVersion);
   }
```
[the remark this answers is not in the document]
kvargha: This only gets invoked when a version is being retired. The intention behind this is when performing version swap, we know what table to switch to.

integrations/venice-duckdb/src/main/java/com/linkedin/venice/sql/DuckDBDaVinciRecordTransformer.java
```diff
@@ -0,0 +1,130 @@
...
+      String sqlDelete = String.format("DELETE FROM %s WHERE %s = '%s';", versionTableName, primaryKey, key.get());
+      stmt.execute(sqlDelete);
+    } catch (SQLException e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  @Override
+  public void onStartVersionIngestion(boolean isCurrentVersion) {
+    try (Connection connection = DriverManager.getConnection(duckDBUrl);
```
ZacAttack, line 97: nit that shoulnd't matter for MVP: I don't know much about how getConnection works.  It might be worth determining if it's worth caching or pooling these connections or if the DriverManager is smart enough to do that for us.

integrations/venice-duckdb/src/main/java/com/linkedin/venice/sql/DuckDBDaVinciRecordTransformer.java
```diff
@@ -0,0 +1,130 @@
...
+        Statement stmt = connection.createStatement()) {
+      String createTableStatement =
+          AvroToSQL.createTableStatement(versionTableName, getOutputValueSchema(), primaryKeys, FAIL);
+      stmt.execute(createTableStatement);
+
+      if (isCurrentVersion) {
+        String createViewStatement =
+            String.format("CREATE OR REPLACE VIEW current_version AS SELECT * FROM %s;", versionTableName);
```
ZacAttack, line 105: similar nit, let's static constant these strings.

ZacAttack: ALSO, take a look at using prepared statements for all of these (this is to prevent sql injection vulnerabilities).

integrations/venice-duckdb/src/main/java/com/linkedin/venice/sql/DuckDBDaVinciRecordTransformer.java
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  @Override
+  public void onEndVersionIngestion(int currentVersion) {
+    try (Connection connection = DriverManager.getConnection(duckDBUrl);
+        Statement stmt = connection.createStatement()) {
+      // Swap to current version
+      String currentVersionTableName = baseVersionTableName + currentVersion;
+      String createViewStatement =
+          String.format("CREATE OR REPLACE VIEW current_version AS SELECT * FROM %s;", currentVersionTableName);
```
ZacAttack, line 120: ditto, prepared statement/static final

integrations/venice-duckdb/src/main/java/com/linkedin/venice/sql/DuckDBDaVinciRecordTransformer.java
```diff
@@ -0,0 +1,130 @@
...
+      stmt.execute(createViewStatement);
+
+      // Drop DuckDB table for storeVersion as it's retired
+      String deleteTableStatement = String.format("DROP TABLE %s;", versionTableName);
```
ZacAttack, line 124: ditto for this one.

https://xkcd.com/327/

integrations/venice-duckdb/src/main/java/com/linkedin/venice/sql/DuckDBDaVinciRecordTransformer.java
kvargha: Yeah connection pooling is something I would like to implement, but for a first iteration this should be fine.

integrations/venice-duckdb/src/main/java/com/linkedin/venice/sql/DuckDBDaVinciRecordTransformer.java
FelixGV: I looked into this and prepared statements only work to inject values, IIUC, not to inject table names.

FelixGV (approved):
Thanks Koorous!

internal/venice-test-common/build.gradle
```diff
@@ -268,6 +266,9 @@ task integrationTests_9999(type: Test) {
         excludeTestsMatching pattern
       }
     }
+
+    // There's an issue where it passes locally but not on the CI
+    excludeTestsMatching 'com.linkedin.venice.endToEnd.DuckDBDaVinciRecordTransformerTest'
```
FelixGV, line 271: FYI: You can simply add `@Test(enabled = false)` to the test function.

FelixGV (approved):

ZacAttack (approved):

integrations/venice-duckdb/src/main/java/com/linkedin/venice/sql/DuckDBDaVinciRecordTransformer.java
FelixGV: We should have a unit test called `littleBobbyTables()`

PR #1436, opened 2025-01-14
[admin-tool][controller]Validate isMigrating flag during store deletion that happens as part of abortMigration

ranwang2024:
Validate isMigrating flag during store deletion that happens as part of abortMigration

When aborting a migration, store deletion should always validate the isMigrating flag. Deletion should only proceed if this flag is set to true. Currently, this validation happens only on the client side (admin tool) and only for source cluster configs, which is insufficient. 

To address this, the delete store command sent to the controller should include a flag indicating the need to check the isMigrating status in destination clusters.

This change adds validation to both the parent controller and the child controller, ensuring that the RT topic is not accidentally deleted when aborting a store migration.

## How was this PR tested?

This PR was tested by writing and including unit tests and integration tests within the PR itself. These tests were then executed successfully to ensure the changes work as intended.

If the `isMigrating` flag for a store is set to false during the abortion of migration in the Parent Controller, the deletion of the destination store will not be initiated. Instead, a delete response with an error message—'Store {storeName}'s migrating flag is false. It is not safe to delete a store assumed to be migrating without the migrating flag set to true.'—will be returned, accompanied by the ErrorType.INVALID_CONFIG. If this situation occurs in the Child Controller, the RT topic will not be deleted, and instead of throwing an exception, an error message will be logged.


clients/venice-admin-tool/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
sushantmane, line 1: Why do we need this change?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -1067,8 +1077,14 @@ private void deleteStore(
         storeRepository.updateStore(store);
       } catch (VeniceNoStoreException e) {
         // It's possible for a store to partially exist due to partial delete/creation failures.
-        LOGGER
-            .warn("Store object is missing for store: " + storeName + " will proceed with the rest of store deletion");
+        LOGGER.warn("Store object is missing for store: {} will proceed with the rest of store deletion", storeName);
+      }
+      if (abortStoreMigrating && store != null && !store.isMigrating()) {
+        LOGGER.error(
+            "Delete store: {} is triggered by --abort-migration command, however its migrating flag "
```
sushantmane, line 1084: Can we rephrase it something like this?

```java
LOGGER.warn("Deletion of store: {} in cluster: {} was issued as part of abort migration resource cleanup, but the store's isMigrating flag is false. Please ensure the store's isMigrating flag is set to true in the destination cluster before issuing the delete command to prevent accidental deletion of shared resources", storeName, clusterName);
```

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
```diff
@@ -959,8 +959,22 @@ private void setupResourceForBatchJobHeartbeatStore(String batchJobHeartbeatStor
...
+      boolean abortStoreMigrating,
       int largestUsedVersionNumber,
       boolean waitOnRTTopicDeletion) {
+    if (abortStoreMigrating) {
+      HelixVeniceClusterResources resources = getHelixVeniceClusterResources(clusterName);
+      try (AutoCloseableLock ignore = resources.getClusterLockManager().createStoreReadLock(storeName)) {
+        ReadWriteStoreRepository repository = resources.getStoreMetadataRepository();
+        Store store = repository.getStore(storeName);
+        if (store != null && !store.isMigrating()) {
+          throw new VeniceException(
```
sushantmane, line 971: Can we also add log here?

internal/venice-common/src/main/java/com/linkedin/venice/controllerapi/ControllerApiConstants.java
```diff
@@ -7,6 +7,7 @@ public class ControllerApiConstants {
   public static final String SOURCE_GRID_FABRIC = "source_grid_fabric";
   public static final String BATCH_JOB_HEARTBEAT_ENABLED = "batch_job_heartbeat_enabled";
 
+  public static final String ABORT_MIGRATION_ENABLED = "abort_migration_enabled";
```
sushantmane, line 10: Can we rename this to 
`public static final String IS_ABORT_MIGRATION_CLEANUP = "is_abort_migration_cleanup";`

services/venice-controller/src/test/java/com/linkedin/venice/controller/server/StoreRoutesTest.java
```diff
@@ -89,6 +93,39 @@ public void testRollForwardToFutureVersion() throws Exception {
...
+    String errMessage = "Store " + TEST_STORE_NAME + "'s migrating flag is false. Not safe to delete a store "
+        + "that is assumed to be migrating without the migrating flag setup as true.";
+    doThrow(new VeniceException(errMessage, INVALID_CONFIG)).when(mockAdmin)
+        .deleteStore(TEST_CLUSTER, TEST_STORE_NAME, true, Store.IGNORE_VERSION, false);
+    trackableControllerResponse = ObjectMapperFactory.getInstance()
+        .readValue(
+            deleteStoreRoute.handle(request, mock(Response.class)).toString(),
+            TrackableControllerResponse.class);
+    Assert.assertTrue(trackableControllerResponse.isError());
+    Assert.assertEquals(trackableControllerResponse.getErrorType(), INVALID_CONFIG);
```
sushantmane, line 124: We can add assertion for the message as well

services/venice-controller/src/main/java/com/linkedin/venice/controller/server/StoresRoutes.java
```diff
@@ -510,29 +511,39 @@ public Route deleteStore(Admin admin) {
...
+          // Only allow allowlist users to run this command
+          if (!checkIsAllowListUser(request, veniceResponse, () -> isAllowListUser(request))) {
+            return;
+          }
+          AdminSparkServer.validateParams(request, DELETE_STORE.getParams(), admin);
+          String clusterName = request.queryParams(CLUSTER);
+          String storeName = request.queryParams(NAME);
+          boolean abortMigratingStore =
+              Utils.parseBooleanFromString(request.queryParams(ABORT_MIGRATION_ENABLED), ABORT_MIGRATION_ENABLED);
+          System.out.println("rwang2 StoresRoute abortMigratingStore: " + abortMigratingStore);
```
sushantmane, line 524: remove sout?

services/venice-controller/src/main/java/com/linkedin/venice/controller/server/StoresRoutes.java
sushantmane, line 1: Can you remove printlns? Please use logger if you need to log messages

clients/venice-admin-tool/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
ranwang2024: In order to write unit tests for static methods using Mockito, we need to add the `mockito-inline` dependency. This is because the default mock maker used by Mockito, which is `SubclassByteBuddyMockMaker`, does not support the creation of static mocks. Without this dependency, attempting to mock static methods will result in the following exception:

```
org.mockito.exceptions.base.MockitoException: 
The used MockMaker SubclassByteBuddyMockMaker does not support the creation of static mocks.
```

Additionally, our `da-vinci-client` tests also include a file located at `clients/da-vinci-client/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker`. The org.mockito.plugins.MockMaker file specifies the mock maker to be used by Mockito. By adding the `mockito-inline` dependency, we ensure that Mockito uses the appropriate mock maker that supports static method mocking.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
ranwang2024: Sure, I have updated the logger warning message as suggested. However, I noticed that the parameter in the Venice Zookeeper store file is named `migrating` instead of `isMigrating`. Therefore, I changed the message to use "store's migrating flag" instead of "store's isMigrating flag" to avoid confusion. Please let me know if my understanding is incorrect.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
ranwang2024: Sure, added a logger warning message similar to the one in VeniceHelixAdmin.java

services/venice-controller/src/main/java/com/linkedin/venice/controller/server/StoresRoutes.java
ranwang2024: Sure, removed.

services/venice-controller/src/main/java/com/linkedin/venice/controller/Admin.java
```diff
@@ -180,7 +180,12 @@ void createStore(
   * Delete the entire store including both metadata and real user's data. Before deleting a store, we should disable
   * the store manually to ensure there is no reading/writing request hitting this tore.
   */
-  void deleteStore(String clusterName, String storeName, int largestUsedVersionNumber, boolean waitOnRTTopicDeletion);
+  void deleteStore(
+      String clusterName,
+      String storeName,
+      boolean abortMigratingStore,
```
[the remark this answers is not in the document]
ranwang2024: Sure, renamed.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -1042,17 +1042,27 @@ private void configureNewStore(Store newStore, VeniceControllerClusterConfig con
   public void deleteStore(
       String clusterName,
       String storeName,
+      boolean abortStoreMigrating,
```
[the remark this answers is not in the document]
ranwang2024: Yeah, renamed.

internal/venice-common/src/main/java/com/linkedin/venice/controllerapi/ControllerApiConstants.java
ranwang2024: Sure, updated.

services/venice-controller/src/main/java/com/linkedin/venice/controller/server/StoresRoutes.java
ranwang2024: Thank you for pointing this out, removed.

services/venice-controller/src/test/java/com/linkedin/venice/controller/server/StoreRoutesTest.java
ranwang2024: Sure, added an assertion for the errMessage

internal/venice-common/src/main/java/com/linkedin/venice/controllerapi/ControllerClient.java
```diff
@@ -602,7 +603,11 @@ public StoreMigrationResponse abortMigration(String storeName, String destCluste
   }
 
   public TrackableControllerResponse deleteStore(String storeName) {
-    QueryParams params = newParams().add(NAME, storeName);
+    return deleteStore(storeName, false);
+  }
+
+  public TrackableControllerResponse deleteStore(String storeName, boolean isAbortMigratingStore) {
```
sushantmane, line 609: Should we change `isAbortMigratingStore` to `isAbortMigrationCleanup`?

sushantmane (dismissed):
Overall LGTM. Thanks, @ranwang2024!

clients/venice-admin-tool/src/test/java/com/linkedin/venice/TestAdminTool.java
```diff
@@ -149,6 +157,88 @@ public void testIsClonedStoreOnline() {
...
+            .thenReturn(destControllerClient);
+
+        adminToolMockedStatic.when(() -> AdminTool.userGivesPermission("Do you still want to proceed"))
+            .thenReturn(false);
+
+        AdminTool.abortMigration("http://localhost:7036", storeName, srcCluster, dstCluster, false, new boolean[0]);
+        Mockito.verify(srcControllerClient, times(0)).discoverCluster(storeName);
+        Mockito.verify(srcControllerClient, times(0)).abortMigration(storeName, dstCluster); // verify dest client is
+                                                                                             // called with true flag
+                                                                                             // and store name as
```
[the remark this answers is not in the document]
ranwang2024: Good catch! Moved comments before verify statements.

internal/venice-common/src/main/java/com/linkedin/venice/controllerapi/ControllerClient.java
ranwang2024: Updated.

sushantmane (approved):
LGTM. Thanks!

PR #1438, opened 2025-01-15
[duckdb] Refactored AvroToSQL to make key schemas first-class

FelixGV:
This refactoring lets the calling code pass the two kinds of schemas, for keys and values, as separate params. Previously, the calling code would have had to assemble a new schema equivalent to the SQL row, which was unwieldy.

Also added support for projecting columns (unused for now).

## How was this PR tested?

Unit tests.

### DIFF
--- a/integrations/venice-duckdb/src/main/java/com/linkedin/venice/sql/AvroToSQL.java
+++ b/integrations/venice-duckdb/src/main/java/com/linkedin/venice/sql/AvroToSQL.java
@@ -1,21 +1,16 @@
 package com.linkedin.venice.sql;
 
-import com.linkedin.venice.utils.ByteUtils;
-import java.nio.ByteBuffer;
 import java.sql.JDBCType;
-import java.sql.PreparedStatement;
-import java.sql.SQLException;
 import java.util.Collections;
 import java.util.EnumMap;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
-import java.util.function.BiConsumer;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import org.apache.avro.Schema;
-import org.apache.avro.generic.GenericRecord;
 
 
 /**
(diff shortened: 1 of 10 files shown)


internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/DuckDBDaVinciRecordTransformerIntegrationTest.java
```diff
@@ -12,12 +12,8 @@
...
 import static com.linkedin.venice.utils.IntegrationTestPushUtils.createStoreForJob;
 import static com.linkedin.venice.utils.IntegrationTestPushUtils.defaultVPJProps;
-import static com.linkedin.venice.utils.TestWriteUtils.DEFAULT_USER_DATA_RECORD_COUNT;
-import static com.linkedin.venice.utils.TestWriteUtils.NAME_RECORD_V1_SCHEMA;
-import static com.linkedin.venice.utils.TestWriteUtils.getTempDataDirectory;
-import static com.linkedin.venice.utils.TestWriteUtils.writeSimpleAvroFileWithStringToNameRecordV1Schema;
-import static com.linkedin.venice.vpj.VenicePushJobConstants.DEFAULT_KEY_FIELD_PROP;
-import static com.linkedin.venice.vpj.VenicePushJobConstants.VENICE_STORE_NAME_PROP;
+import static com.linkedin.venice.utils.TestWriteUtils.*;
+import static com.linkedin.venice.vpj.VenicePushJobConstants.*;
```
kvargha, line 16: Nit: Fix wildcard imports

integrations/venice-duckdb/src/test/java/com/linkedin/venice/duckdb/DuckDBAvroToSQLTest.java
```diff
@@ -1,12 +1,14 @@
 package com.linkedin.venice.duckdb;
 
+import static com.linkedin.avroutil1.compatibility.AvroCompatibilityHelper.*;
```
kvargha, line 3: Nit: Fix wildcard imports

integrations/venice-duckdb/src/main/java/com/linkedin/venice/sql/AvroToSQL.java
```diff
@@ -64,23 +59,29 @@ private AvroToSQL() {
...
-      throw new IllegalArgumentException("Only Avro records can have a corresponding CREATE TABLE statement.");
-    }
+      @Nonnull Schema keySchema,
+      @Nonnull Schema valueSchema,
+      @Nonnull Set<String> columnsToProject,
+      @Nonnull UnsupportedTypeHandling unsupportedTypeHandling,
+      boolean primaryKey) {
+    Set<Schema.Field> allColumns = combineColumns(keySchema, valueSchema, columnsToProject);
     StringBuffer stringBuffer = new StringBuffer();
-    stringBuffer.append("CREATE TABLE IF NOT EXISTS " + cleanTableName(tableName) + "(");
```
kvargha, line 80: I think we need IF NOT EXISTS. One of the tests I had in mind is to verify that the data is persisted on restarts, and DVRT::onStartVersionIngestion will try to recreate the table on the second instantiation.

A stretch goal would be to figure out schema evolution.

FelixGV: Yeah so... schema evolution is the rub, here...

If the schema is the same before and after restart, then not recreating the table is right (and that can be achieved either by not running the statement at all, or via IF NOT EXIST).

But if the schema has changed, then I don't think we want to keep that table. We might want to ALTER it (but not sure, there are implications to that), or else DROP and re-CREATE (at which point, IF NOT EXIST wouldn't come into play).

So the way I see it, most likely, schema introspection is required anyway to do the right thing...

kvargha (approved):

integrations/venice-duckdb/src/main/java/com/linkedin/venice/sql/AvroToSQL.java
kvargha: Makes sense. We can deal with that at a later time

PR #1439, opened 2025-01-15
[controller] Add interceptor to enforce SSL/TLS and propagate client attributes in gRPC controller server

sushantmane:
## [controller] Add interceptor to enforce SSL/TLS and propagate client attributes in gRPC controller server
This commit introduces a gRPC interceptor that ensures all incoming connections use SSL/TLS 
and extracts client SSL attributes for downstream processing.

Key features:
- Validates the presence of an SSL session for each gRPC call.
- Extracts client certificate and remote address details.
- Injects extracted attributes into the gRPC context for downstream access.
- Rejects calls without SSL/TLS with an UNAUTHENTICATED status.


FelixGV (dismissed):
Thanks Sushant!

sushantmane:
This commit introduces a gRPC interceptor that ensures all incoming connections use SSL/TLS 
for secure gRPC server and extracts client SSL attributes for downstream processing.

Key features:
- Validates the presence of an SSL session for each gRPC call.
- Extracts client certificate and remote address details.
- Injects extracted attributes into the gRPC context for downstream access.
- Rejects calls without SSL/TLS with an UNAUTHENTICATED status.

FelixGV (approved):

PR #1443, opened 2025-01-16
[controller] Enforce ACL checks on gRPC createStore API

sushantmane:
## Enforce ACL checks on VeniceControllerGrpcServiceImpl::createStore API

- Use access controller manager to enforce ACL checks in gRPC createStore API.
- Add support for running E2E tests with a custom access controller.
- Introduce `GrpcControllerClientDetails` to encapsulate gRPC session details for 
  authorization checks.

### DIFF
--- a/internal/venice-common/src/main/java/com/linkedin/venice/controllerapi/transport/GrpcRequestResponseConverter.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/controllerapi/transport/GrpcRequestResponseConverter.java
@@ -33,7 +33,7 @@ public class GrpcRequestResponseConverter {
    *
    * @param code            The gRPC status code representing the error (e.g., {@link io.grpc.Status.Code}).
    * @param errorType       The specific controller error type represented by {@link ControllerGrpcErrorType}.
-   * @param e               The exception containing the error message.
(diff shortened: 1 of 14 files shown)


internal/venice-common/src/main/java/com/linkedin/venice/controllerapi/transport/GrpcRequestResponseConverter.java
```diff
@@ -85,6 +85,22 @@ public static void sendErrorResponse(
...
+      Code code,
+      ControllerGrpcErrorType errorType,
+      Exception exception,
+      String clusterName,
+      String storeName,
+      StreamObserver<?> responseObserver) {
+    sendErrorResponse(
+        code,
+        errorType,
+        exception != null ? exception.getMessage() : "",
```
majisourav99, line 98: can getMessage be null?

sushantmane: yeah it can be. Hence we explicitly check non-nullability of message in overridden method `sendErrorResponse`

majisourav99 (approved):
LGTM

sushantmane:
Thanks, @majisourav99!

PR #1444, opened 2025-01-16
[controller] AdminConsumptionTask fix around timed out or cancelled AdminExecutionTasks

xunyin8:
## [controller] AdminConsumptionTask fix around timed out or cancelled AdminExecutionTasks
This is based on Felix's commit with some additional changes to add an unit test for timed out or cancelled AdminExecutionTask to verify the fix. The original commit message is below:

The main bug fix in this commit is that in the AdminConsumptionTask's executeMessagesAndCollectResults function, tasks for many stores could be scheduled, but if some of them never began running, and the whole set of tasks timed out (after 30 minutes), then the tasks that didn't start at all would leave a lingering reference to the corresponding store in the storeToScheduledTask map, which is a class property. This in turn would cause the next execution of this function to skip the affected store, therefore never letting it run any task again. Since it is not actually needed to hang on this map at the class level, the fix is to change this to a local variable, which therefore does not carry any consequences across executions of the function.

In addition, several NPEs (or potential ones, anyway), are also fixed.

Finally, there are some cosmetic changes such as deleting unused code, and making some fields final.

## How was this PR tested?
Existing and new unit test


FelixGV (approved):
Thanks a lot Xun!

PR #1445, opened 2025-01-16
[duckdb] Cleanup and refactor DVRT interface to be more configurable

kvargha:

Previously, the user would have to define the key and output schema inside their DVRT implementation, which wasn't clean. Now, they're moved into the constructor. The user now only needs to provide the output schema, as the key schema will be automatically passed in by DVC.

I also added a constructor parameter for original output schema to support reader/writer schemas, but added a ToDo on how to implement the logic to pass in the correct schema.

Lastly, I make the DuckDB DVRT implementation more configurable, as it was previously more hardcoded.

This changes modifies the constructor of DVRT with the intention of making the implementation cleaner.


clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/DaVinciRecordTransformer.java
```diff
@@ -206,6 +205,33 @@ public final boolean getStoreRecordsInDaVinci() {
...
+  public final Schema getKeySchema() {
+    return keySchema;
+  }
+
+  /**
+   * Returns the schema for the output value used in {@link DaVinciClient}'s operations.
+   *
+   * @return a {@link Schema} corresponding to the type of {@link V}.
+   */
+  public final Schema getOriginalValueSchema() {
```
FelixGV, line 222: How about `getInputValueSchema`, or `getLatestInputValueSchema`, to contrast with `getOutputValueSchema`? "Original" can be confusing IMHO.

Let's also fix the first line of the JavaDoc where it says "output".

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -473,10 +473,20 @@ public StoreIngestionTask(
     this.chunkAssembler = new ChunkAssembler(storeName);
     this.cacheBackend = cacheBackend;
 
-    if (recordTransformerFunction != null) {
-      DaVinciRecordTransformer clientRecordTransformer = recordTransformerFunction.apply(versionNumber);
+    if (recordTransformerConfig != null && recordTransformerConfig.getRecordTransformerFunction() != null) {
+      Schema keySchema = schemaRepository.getKeySchema(storeName).getSchema();
+      // ToDo: Fetch latest value schema
```
FelixGV, line 478: The latest can be fetched via `schemaRepository.getSupersetOrLatestValueSchema(storeName).getSchema()`

clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/DaVinciRecordTransformerUtility.java
```diff
@@ -116,7 +122,7 @@ public final void onRecovery(
       for (iterator.seekToFirst(); iterator.isValid(); iterator.next()) {
         byte[] keyBytes = iterator.key();
         byte[] valueBytes = iterator.value();
-        Lazy<K> lazyKey = Lazy.of(() -> getKeyDeserializer().deserialize(ByteBuffer.wrap(keyBytes)));
+        Lazy<K> lazyKey = Lazy.of(() -> keyDeserializer.deserialize(ByteBuffer.wrap(keyBytes)));
```
FelixGV, line 125: Nitpick: There is no need to wrap into a `ByteBuffer`. There is also an API which takes in a `byte[]` directly. The `BB` can be useful when we need to use just a subset of the bytes, as is the case for the value, below.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/DaVinciRecordTransformerUtility.java
```diff
@@ -30,6 +30,12 @@ public class DaVinciRecordTransformerUtility<K, O> {
 
   public DaVinciRecordTransformerUtility(DaVinciRecordTransformer recordTransformer) {
     this.recordTransformer = recordTransformer;
+
+    Schema keySchema = recordTransformer.getKeySchema();
+    Schema outputValueSchema = recordTransformer.getOutputValueSchema();
+    this.keyDeserializer = new AvroGenericDeserializer<>(keySchema, keySchema);
+    this.outputValueDeserializer = new AvroGenericDeserializer<>(outputValueSchema, outputValueSchema);
+    this.outputValueSerializer = new AvroSerializer<>(outputValueSchema);
```
FelixGV, line 38: I think these properties can become final, now that they're set in the constructor rather than lazily, right?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/DaVinciRecordTransformerUtility.java
```diff
@@ -127,35 +133,11 @@ public final void onRecovery(
...
-    return outputValueDeserializer;
-  }
-
-  public AvroSerializer<O> getOutputValueSerializer() {
-    if (outputValueSerializer == null) {
-      Schema outputValueSchema = recordTransformer.getOutputValueSchema();
-      outputValueSerializer = new AvroSerializer<>(outputValueSchema);
-    }
-    return outputValueSerializer;
-  }
```
FelixGV, line 160: Nice. This lazy init code was not threadsafe anyway, so one less thing to go wrong...

clients/da-vinci-client/src/test/java/com/linkedin/davinci/config/DaVinciConfigTest.java
```diff
@@ -45,25 +40,36 @@ public void processPut(Lazy<Integer> key, Lazy<Integer> value) {
...
     assertFalse(config.isRecordTransformerEnabled());
+
     DaVinciRecordTransformerConfig recordTransformerConfig = new DaVinciRecordTransformerConfig(
-        (storeVersion) -> new TestRecordTransformer(storeVersion, true),
-        Integer.class,
+        (_storeVersion, _keySchema, _originalValueSchema, _outputValueSchema) -> new TestRecordTransformer(
+            _storeVersion,
+            _keySchema,
+            _originalValueSchema,
+            _outputValueSchema,
```
FelixGV, line 49: Nitpick: I would prefer that we don't use underscores since we don't use them in the rest of the code base. Moreover, in some coding styles (not in ours, however) the underscore at the beginning is used to denote a class property, which is not the case here, but could still confuse some readers...

There are several other occurrences of underscores in lambda params below. Can we please change all of them?

FelixGV (commented):
Thanks Koorous! This PR is a very good improvement! I left a few minor comments, but I think this is very close to merging! Looking forward to it!

FelixGV (dismissed):
Thanks a lot Koorous! LGTM! 🚀 🚢

clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/DaVinciRecordTransformer.java
```diff
@@ -40,33 +40,46 @@ public abstract class DaVinciRecordTransformer<K, V, O> {
    */
   private final boolean storeRecordsInDaVinci;
 
+  /**
+   * The key schema, which is immutable.
+   */
+  private final Schema keySchema;
```
ZacAttack, line 46: This makes sense, and it's probably what we have to deal with (as being able to mutate this runs into conflicting writer problems).  Though I'm trying to think through what kinds of use cases this might disqualify...

I suppose what this implies is that if you have or want state which is managed by DaVinci, then that means the key for that managed state cannot change.  A user still could conceivably do whatever they want with that key elsewhere, but that state would have to be managed by the user and be external to DVRT.  Not 'necessarily' bad, but it is a limitation that maybe we should make clear.

ZacAttack (approved):

FelixGV (dismissed):

ZacAttack (approved):

PR #1459, opened 2025-01-21
[duckdb] Fixed multi-store bug

FelixGV:
The view created by DuckDBDVRT was called current_version, so it could not work with multiple stores. Changed it so that the view inherits the store name instead.

Miscellaneous build changes:

- Fixed the duckdb exclusion rule in all-modules.
- Bumped up the duckdb snapshot dep to the current latest one.

## How was this PR tested?

New unit test.

### DIFF
--- a/integrations/venice-duckdb/src/test/java/com/linkedin/venice/duckdb/DuckDBDaVinciRecordTransformerTest.java
+++ b/integrations/venice-duckdb/src/test/java/com/linkedin/venice/duckdb/DuckDBDaVinciRecordTransformerTest.java
@@ -1,7 +1,6 @@
 package com.linkedin.venice.duckdb;
 
-import static com.linkedin.venice.utils.TestWriteUtils.NAME_RECORD_V1_SCHEMA;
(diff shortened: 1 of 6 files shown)


integrations/venice-duckdb/src/test/java/com/linkedin/venice/duckdb/DuckDBDaVinciRecordTransformerTest.java
```diff
@@ -1,7 +1,6 @@
 package com.linkedin.venice.duckdb;
 
-import static com.linkedin.venice.utils.TestWriteUtils.NAME_RECORD_V1_SCHEMA;
-import static com.linkedin.venice.utils.TestWriteUtils.SINGLE_FIELD_RECORD_SCHEMA;
+import static com.linkedin.venice.utils.TestWriteUtils.*;
```
kvargha, line 3: Nit: Fix wildcard imports.

kvargha (approved):

PR #1462, opened 2025-01-22
[server] Handle exceptions in HB reporting threads

majisourav99:
## Handle exceptions in HB reporting threads

Thread "Ingestion-Heartbeat-Reporter-Service-Thread" could exit due to unhandled exceptions, which leads to stopped HB threads. This PR handles such exceptions. 

## How was this PR tested?

GHCI

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/HeartbeatMonitoringServiceStats.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/HeartbeatMonitoringServiceStats.java
@@ -1,0 +1,22 @@
+package com.linkedin.davinci.stats;
+
+import com.linkedin.venice.stats.AbstractVeniceStats;
+import io.tehuti.metrics.MetricsRepository;
+import io.tehuti.metrics.Sensor;
+import io.tehuti.metrics.stats.Count;
+
+
+public class HeartbeatMonitoringServiceStats extends AbstractVeniceStats {
+  private static final String HB_SUFFIX = "-heartbeat-monitor-service";
+  private final Sensor heartbeatExceptionCountSensor;
+
+  public HeartbeatMonitoringServiceStats(MetricsRepository metricsRepository, String clusterName) {
+    super(metricsRepository, clusterName + HB_SUFFIX);
+    heartbeatExceptionCountSensor = registerSensorIfAbsent("heartbeat-monitor-service-exception-count", new Count());
+  }
+
+  public void recordHeartbeatExceptionCountSensor() {
+    heartbeatExceptionCountSensor.record();
+  }
+
+}
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/HeartbeatMonitoringServiceStats.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/HeartbeatMonitoringServiceStats.java
@@ -4,19 +4,31 @@ import com.linkedin.venice.stats.AbstractVeniceStats;
 import io.tehuti.metrics.MetricsRepository;
 import io.tehuti.metrics.Sensor;
 import io.tehuti.metrics.stats.Count;
+import io.tehuti.metrics.stats.OccurrenceRate;
 
 
 public class HeartbeatMonitoringServiceStats extends AbstractVeniceStats {
   private static final String HB_SUFFIX = "-heartbeat-monitor-service";
   private final Sensor heartbeatExceptionCountSensor;
+  private final Sensor heartbeatReporterSensor;
+  private final Sensor heartbeatLoggingSensor;
 
   public HeartbeatMonitoringServiceStats(MetricsRepository metricsRepository, String clusterName) {
     super(metricsRepository, clusterName + HB_SUFFIX);
     heartbeatExceptionCountSensor = registerSensorIfAbsent("heartbeat-monitor-service-exception-count", new Count());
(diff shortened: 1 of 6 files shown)


sixpluszero (approved):

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/ingestion/heartbeat/HeartbeatMonitoringService.java
```diff
@@ -435,12 +435,14 @@ private class HeartbeatReporterThread extends Thread {
...
     public void run() {
       while (!Thread.interrupted()) {
-        record();
         try {
+          record();
           TimeUnit.SECONDS.sleep(DEFAULT_REPORTER_THREAD_SLEEP_INTERVAL_SECONDS);
         } catch (InterruptedException e) {
           // We've received an interrupt which is to be expected, so we'll just leave the loop and log
           break;
+        } catch (Exception e) {
```
lluwm, line 444: Just one nit, how about potential errors. i.e. a Throwable but not an Exception, I guess we can let the thread to die, but at least we should log why it could happen.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/ingestion/heartbeat/HeartbeatMonitoringService.java
```diff
@@ -455,12 +457,14 @@ private class HeartbeatLagLoggingThread extends Thread {
...
     public void run() {
       while (!Thread.interrupted()) {
-        checkAndMaybeLogHeartbeatDelay();
         try {
+          checkAndMaybeLogHeartbeatDelay();
           TimeUnit.SECONDS.sleep(DEFAULT_LAG_LOGGING_THREAD_SLEEP_INTERVAL_SECONDS);
         } catch (InterruptedException e) {
           // We've received an interrupt which is to be expected, so we'll just leave the loop and log
           break;
+        } catch (Exception e) {
```
gaojieliu, line 466: Do we need to emit metrics to track this HB service as it is critical? 
If it keeps throwing exception in this function, it can result in a big issue in HB lag tracking.

majisourav99: So the symptom would be increasing lag, we can check logs for the following message would help in debugging in those cases?

ZacAttack: It's a good question and I wonder if we should do some tweaking in how this works.  Right now if the alert goes blank it can mean that the store was either deleted or the reporting is broken.

Right now the alert is structured as anything ><big number> triggers an alert.  I had heretofore resisted the idea of putting a negative number on this similar to our ingestion stats error codes as our alerting system didn't really handle this sort of thing well, but a lot has changed lately.  So maybe theres opportunity to publish into this stat some kind of negative number error code that will emit even if the thread dies?

majisourav99: You mean to say emit negative number into lag when the thread dies (case of Throwable), thats reasonable ask. But I think Gaojie asking about catching other types of exceptions which could be temporary in nature, do we need a stat for that or not.

gaojieliu: Yeah, I was thinking two layers of monitoring:
1. HB service itself.
2. HB lags.

If HB service is not running, do we have a way to emit a negative lag?

For #1, I was thinking some metrics to track the healthiness of the `HeartbeatMonitoringService` in a similar way as [`ParticipantStoreConsumptionTask` ](https://github.com/linkedin/venice/blob/main/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ParticipantStoreConsumptionTask.java#L64)

clients/da-vinci-client/src/main/java/com/linkedin/davinci/stats/HeartbeatMonitoringServiceStats.java
```diff
@@ -0,0 +1,22 @@
...
+import io.tehuti.metrics.stats.Count;
+
+
+public class HeartbeatMonitoringServiceStats extends AbstractVeniceStats {
+  private static final String HB_SUFFIX = "-heartbeat-monitor-service";
+  private final Sensor heartbeatExceptionCountSensor;
+
+  public HeartbeatMonitoringServiceStats(MetricsRepository metricsRepository, String clusterName) {
+    super(metricsRepository, clusterName + HB_SUFFIX);
+    heartbeatExceptionCountSensor = registerSensorIfAbsent("heartbeat-monitor-service-exception-count", new Count());
```
gaojieliu, line 15: Can you have a heart-beat metric as well?
If the thread is not running, the heart-beat metric will stop emitting any data, which can be a signal we can alert against.
One example:
https://github.com/linkedin/venice/blob/main/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ParticipantStoreConsumptionTask.java#L64

majisourav99: Sure, a heartbeat for heartbeat services :)

gaojieliu (approved):
Thanks

PR #1463, opened 2025-01-22
[vpj] Dual write push job detials report from VPJ to both parent and child region for push job system store replication mode migration

sixpluszero:
## [vpj] Dual write push job detials report from VPJ to both parent and child region for push job system store replication mode migration

Add a new controller to write to child fabric for push job details system store, so that when we migrate this particular system store to A/A mode, it will be able to receive updates.

## How was this PR tested?
N/A

### DIFF
--- a/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
+++ b/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
@@ -90,6 +90,7 @@ import static com.linkedin.venice.vpj.VenicePushJobConstants.VENICE_STORE_NAME_P
 
 import com.linkedin.avroutil1.compatibility.AvroCompatibilityHelper;
 import com.linkedin.venice.PushJobCheckpoints;
+import com.linkedin.venice.common.VeniceSystemStoreUtils;
 import com.linkedin.venice.compression.CompressionStrategy;
 import com.linkedin.venice.compression.ZstdWithDictCompressor;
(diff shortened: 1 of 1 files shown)


clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
```diff
@@ -1868,11 +1882,23 @@ private void sendPushJobDetailsToController() {
     updatePushJobDetailsWithConfigs();
     updatePushJobDetailsWithLivenessHeartbeatException();
 
+    /**
+     *  This logic is added to migrate push job details system store from Aggregate mode to Active/Active mode.
+     *  If it is multi-region, it will dual write push job details to parent RT and child region RT.
+     *  If it is single-region, it will only write push job details to child region RT.
```
nisargthakkar, line 1888: Why can't parent controller write it to child colo's RT instead? Why do we need to make a controller client to an arbitrary child colo?

nisargthakkar: There is also stuff in parent controller that emits metrics about push job success/failures that we'll need to handle

sixpluszero: Yeah you are right...I think I read the code wrong. It probably cannot work alone within VPJ changes. I will probably just change to dual write in parent controller.

PR #1464, opened 2025-01-22
[fast-client] Improved the failure detection

gaojieliu:
This PR is trying to improve the failure detection in Fast Client. In the past, Fast Client was relying on organic traffic to detect unhealthy nodes and once it detects some unhealthy nodes, it won't stop sending traffic to these nodes completely as it would try to use organic traffic to detect whether the unhealthy nodes are recovered or not, which could result in high latency issue.

This PR is trying to address two issues:
1. It is confirmed that httpclient5 lib doesn't provide an accurate request level timeout, so we will enforce a timeout in Fast Client by setting up a timer and the default timeout is 1s.
2. Failure detection logic will be as follows: 
<ol type="a">
  <li>When the organic traffic encounters timing out, the corresponding backend instances will be put into the suspicious instance set. </li>
  <li>The heartbeat runnable will start send heart-beat messages to these suspicious instances periodically, and if the heart-beat message fails, the suspcious instances will become unhealthy instances, and these instances will be blocked from the regular routing until heart-beat message succeeds later on. </li>
  <li>We are trying to limit the total amount of HB messages to the backend instances as much as possible.</li>
</ol>

 
4. `InstanceHealthMonitor` will be constructed as a singleton in the wrapper, so that Fast Client won't send out duplicate HB messages even there are multiple stores being used by the same application.

## How was this PR tested?
CI

### DIFF
--- a/clients/venice-client/src/main/java/com/linkedin/venice/fastclient/ClientConfig.java
+++ b/clients/venice-client/src/main/java/com/linkedin/venice/fastclient/ClientConfig.java
@@ -7,6 +7,8 @@ import com.linkedin.venice.client.exceptions.VeniceClientException;
 import com.linkedin.venice.client.store.AvroGenericStoreClient;
 import com.linkedin.venice.client.store.AvroSpecificStoreClient;
 import com.linkedin.venice.fastclient.meta.ClientRoutingStrategyType;
+import com.linkedin.venice.fastclient.meta.InstanceHealthMonitor;
+import com.linkedin.venice.fastclient.meta.InstanceHealthMonitorConfig;
 import com.linkedin.venice.fastclient.meta.StoreMetadataFetchMode;
 import com.linkedin.venice.fastclient.stats.ClusterStats;
 import com.linkedin.venice.fastclient.stats.FastClientStats;
@@ -20,7 +22,6 @@ import java.util.Collections;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.Executor;
-import java.util.concurrent.TimeUnit;
 import org.apache.avro.specific.SpecificRecord;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
@@ -43,15 +44,6 @@ public class ClientConfig<K, V, T extends SpecificRecord> {
   private final AvroGenericStoreClient<K, V> genericThinClient;
   private final AvroSpecificStoreClient<K, T> specificThinClient;
 
-  /**
-   * For Client Routing.
-   * Please check {@link com.linkedin.venice.fastclient.meta.InstanceHealthMonitor} to find more details.
-   */
-  private final long routingLeakedRequestCleanupThresholdMS;
-  private final long routingQuotaExceededRequestCounterResetDelayMS;
-  private final long routingErrorRequestCounterResetDelayMS;
-  private final long routingUnavailableRequestCounterResetDelayMS;
-  private final int routingPendingRequestCounterInstanceBlockThreshold;
   private final DaVinciClient<StoreMetaKey, StoreMetaValue> daVinciClientForMetaStore;
   /**
    * Config to enable/disable warm up connection to instances from fetched metadata.
@@ -94,6 +86,7 @@ public class ClientConfig<K, V, T extends SpecificRecord> {
 
   private boolean projectionFieldValidation;
   private Set<String> harClusters;
+  private final InstanceHealthMonitor instanceHealthMonitor;
 
   private final MetricsRepository metricsRepository;
 
@@ -109,11 +102,6 @@ public class ClientConfig<K, V, T extends SpecificRecord> {
       boolean dualReadEnabled,
       AvroGenericStoreClient<K, V> genericThinClient,
       AvroSpecificStoreClient<K, T> specificThinClient,
-      long routingLeakedRequestCleanupThresholdMS,
-      long routingQuotaExceededRequestCounterResetDelayMS,
-      long routingErrorRequestCounterResetDelayMS,
-      long routingUnavailableRequestCounterResetDelayMS,
-      int routingPendingRequestCounterInstanceBlockThreshold,
       DaVinciClient<StoreMetaKey, StoreMetaValue> daVinciClientForMetaStore,
       boolean isMetadataConnWarmupEnabled,
       long metadataRefreshIntervalInSeconds,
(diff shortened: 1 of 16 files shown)


clients/venice-client/src/main/java/com/linkedin/venice/fastclient/ClientConfig.java
```diff
@@ -263,6 +233,12 @@ private ClientConfig(
     this.projectionFieldValidation = projectionFieldValidation;
     this.longTailRetryBudgetEnforcementWindowInMs = longTailRetryBudgetEnforcementWindowInMs;
     this.harClusters = harClusters;
+    if (instanceHealthMonitor == null) {
+      this.instanceHealthMonitor =
+          new InstanceHealthMonitor(InstanceHealthMonitorConfig.builder().setClient(this.r2Client).build());
```
ZacAttack, line 238: The PR description says that instanceHealthmonitor should be a singleton, but i'm not sure I'm seeing where this is enforced?

gaojieliu: This will be enforced in the MP layer.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/fastclient/FastClientIndividualFeatureConfigurationTest.java
```diff
@@ -471,7 +515,13 @@ public void testMultiKeyFanoutStats() throws IOException, ExecutionException, In
         new ClientConfig.ClientConfigBuilder<>().setStoreName(storeName)
             .setR2Client(r2Client)
             .setLongTailRetryEnabledForBatchGet(true)
-            .setSpeculativeQueryEnabled(false);
+            .setSpeculativeQueryEnabled(false)
+            .setInstanceHealthMonitor(
+                new InstanceHealthMonitor(
+                    InstanceHealthMonitorConfig.builder()
+                        .setClient(r2Client)
+                        .setRoutingRequestDefaultTimeoutMS(10000)
```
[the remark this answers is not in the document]
gaojieliu: The default is 1s and several integration test breaches the limit as Venice Server is slow in integration test especially right after startup.
This PR has some unit tests to cover the new logic.

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/meta/InstanceHealthMonitor.java
```diff
@@ -26,33 +27,40 @@
...
  */
 public class InstanceHealthMonitor implements Closeable {
   private static final Logger LOGGER = LogManager.getLogger(InstanceHealthMonitor.class);
-  private final ClientConfig clientConfig;
+  private final InstanceHealthMonitorConfig config;
 
-  // Map/set of per store replica instances
   private final Map<String, Integer> pendingRequestCounterMap = new VeniceConcurrentHashMap<>();
   private final Set<String> unhealthyInstanceSet = new ConcurrentSkipListSet<>();
+  private final Set<String> suspiciousInstanceSet = new ConcurrentSkipListSet<>();
```
ZacAttack, line 48: Can you help me understand the functional difference between suspicious and unhealthy? (still reading the class, but from the javadoc I'm not sure it's apparant to me).

gaojieliu: For suspicious instances, fast-client can still send request to them, but the internal HB runnable will start sending HB request to these instances, and if the HB request fails, these suspicious instances will become unhealthy instances and fast-client will stop sending regular requests to them until HB requests succeed later on.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/fastclient/FastClientIndividualFeatureConfigurationTest.java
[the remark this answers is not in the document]
ZacAttack: How is that?  Also would it make sense though to do it in OS?  Or is there something that prevents that from being a good idea?

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/ClientConfig.java
ZacAttack: How is that?  Also would it make sense though to do it in OS?  Or is there something that prevents that from being a good idea?

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/fastclient/FastClientIndividualFeatureConfigurationTest.java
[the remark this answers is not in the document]
ZacAttack: Ok, so our supposition is that there is some likelihood that today this is triggering the retry at this threshhold?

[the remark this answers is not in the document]
gaojieliu: The requested timeout would trigger several integration test failures.

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/ClientConfig.java
gaojieliu: The idea is to have a global config (independent from client config) to control the behavior of `InstanceHealthMonitor`, which can be done easily in MP layer and it is not straightforward to do it in OS unless we hard-code all the configs for `InstanceHealthMonitorConfig` to constructor a singleton of `InstanceHealthMonitor`.

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/meta/InstanceHealthMonitor.java
```diff
@@ -106,74 +162,45 @@ public ChainedCompletableFuture<Integer, Integer> trackHealthBasedOnRequestToIns
     TimeoutProcessor.TimeoutFuture timeoutFuture = null;
     if (transportFuture != null) {
       timeoutFuture = timeoutProcessor.schedule(
-          /** Using a special http status to indicate the leaked request */
+          /** Using a special http status to indicate the timed out request */
           () -> {
             transportFuture.completeExceptionally(new VeniceClientHttpException("Request timed out", SC_GONE));
+            // TODO: maybe remove this log as the logging frequency can be high
+            LOGGER.warn(
```
xunyin8, line 169: We can use a `RedundantLogFilter` for this.

xunyin8 (dismissed):
Thanks Gaojie, LGTM just one minor logging comment.

xunyin8 (approved):
Thanks Gaojie!

clients/venice-client/src/main/java/com/linkedin/venice/fastclient/meta/InstanceHealthMonitor.java
gaojieliu: Fixed

PR #1466, opened 2025-01-22
[da-vinci][common] DVC consumer for materialized view (batch only)

xunyin8:
## [da-vinci][common] DVC consumer for materialized view (batch only)
Limiting the change to support batch only materialized view to keep the PR small. Hybrid DVC consumer support and related features such as heartbeat  will be added in a separate PR

1. New DaVinciClientFactory APIs for creating DVC for a given view.

2. Defined a new "storeName" rule for views to be used for metrics reporting and DVC client. See VeniceView.getViewStoreName for details. 

3. Introduced NativeMetadataRepositoryViewAdapter and HelixReadOnlyStoreViewConfigRepositoryAdapter to provide read-only interface to access various store metadata for both regular Venice stores and store views with the VeniceView.getViewStoreName.

There is some issue with chunking support on the read path. When chunking is enabled the view topic keys are doubly wrapped by `serializeNonChunkedKey`. This is because during NR pass-through mode the view writer is essentially trying to chunk the chunk. The tactical fix now is to unwrap the key with chunked suffix bytes appended and pass it to the view writer to be wrapped again and sent to the correct partition. This only works with non-large messages. I.e. chunking is enabled but nothing is actually getting chunked. Large messages will require a proper fix.

## How was this PR tested?
Unit tests and integration test


gaojieliu:
Hi @xunyin8 
I have a big question about how the adapting layer looks like and I read your PR and I think you are trying to plugin the adapting layer in a more spread fashion where sometimes we use the original store name and some other times, we use the combined name by concatenating  store name and view name.
There are two points I can think of:
1. The code would become less maintainable as we need to make a decision at different places to use either the original store name or the concatenated name.
2. In theory, this solution is not 100% safe as some store name can be identical to the concatenated name, and if the customer subscribes both, it will encounter weird issues.

Here is one idea in my mind:
1. Make sure the concatenated name unique in Venice platform, and it means we need to guarantee the uniqueness among store names and concatenated view names.
2. In DaVinci world, we can treat the concatenated view name as the store name, and we can build a new adapting layer in the common repositories, such as `StoreRepository`, `SchemaRepository` and they need to handle both view and store request. For example, for the passed param, it can detect whether it is a view or a store and if it is a view, it will find out the corresponding store name and take the corresponding actions.
With this way, it will be very safe to avoid conflict and we only need to change very few places to make it work everywhere and the idea is similar to `HelixReadOnlySchemaRepositoryAdapter`, which can handle both system stores and user stores in a transparent way.
3. Less error prone as the adapting logics are only in several important repositories, nowhere else (in theory).

If we build such capability, the change in DVRT or CDC will be fairly straightforward.

WDYT?

xunyin8:
> Member
> ###



> Hi @xunyin8 I have a big question about how the adapting layer looks like and I read your PR and I think you are trying to plugin the adapting layer in a more spread fashion where sometimes we use the original store name and some other times, we use the combined name by concatenating store name and view name. There are two points I can think of:
> 
> 1. The code would become less maintainable as we need to make a decision at different places to use either the original store name or the concatenated name.
> 2. In theory, this solution is not 100% safe as some store name can be identical to the concatenated name, and if the customer subscribes both, it will encounter weird issues.
> 
> Here is one idea in my mind:
> 
> 1. Make sure the concatenated name unique in Venice platform, and it means we need to guarantee the uniqueness among store names and concatenated view names.
> 2. In DaVinci world, we can treat the concatenated view name as the store name, and we can build a new adapting layer in the common repositories, such as `StoreRepository`, `SchemaRepository` and they need to handle both view and store request. For example, for the passed param, it can detect whether it is a view or a store and if it is a view, it will find out the corresponding store name and take the corresponding actions.
>    With this way, it will be very safe to avoid conflict and we only need to change very few places to make it work everywhere and the idea is similar to `HelixReadOnlySchemaRepositoryAdapter`, which can handle both system stores and user stores in a transparent way.
> 3. Less error prone as the adapting logics are only in several important repositories, nowhere else (in theory).
> 
> If we build such capability, the change in DVRT or CDC will be fairly straightforward.
> 
> WDYT?

This makes sense, I was hoping to be able to only keep the adapting layer in `StoreBackend` and `ReadOnlyVersion`. However, as you mentioned and probably noticed while reviewing the PR other components of DVC and ingestion path also requires differentiating a store view from a store hence the adapter keeps leaking. I think we can enforce 1 fairly easily during view creation and (2) could introduce more initial change but let me give it a try and it may be able to solve some of the known issue (e.g. chunking not working) cleanly since the suggested solution is more bullet proof.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/repository/NativeMetadataRepositoryViewAdapter.java
```diff
@@ -0,0 +1,223 @@
...
+    getSubscribedStoreViews(internalStoreName).add(storeName);
+  }
+
+  @Override
+  public void unsubscribe(String storeName) {
+    String internalStoreName = VeniceView.getStoreName(storeName);
+    Set<String> subscribedStoreViews = getSubscribedStoreViews(internalStoreName);
+    subscribedStoreViews.remove(storeName);
+    if (subscribedStoreViews.isEmpty()) {
+      nativeMetadataRepository.unsubscribe(internalStoreName);
```
gaojieliu, line 216: There is a race condition here..
I think we can use `compute` method of `VeniceConcurrentHasmMap` here to make it thread safe.

internal/venice-common/src/main/java/com/linkedin/venice/meta/ReadOnlyStoreView.java
```diff
@@ -0,0 +1,143 @@
...
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+
+public class ReadOnlyStoreView extends ReadOnlyStore {
+  private final Map<Integer, ReadOnlyVersion> viewVersionMap;
+
+  public ReadOnlyStoreView(Store delegate, Map<Integer, ReadOnlyVersion> viewVersionMap) {
+    super(delegate);
+    this.viewVersionMap = viewVersionMap;
```
gaojieliu, line 23: Can we move the logic from `NativeMetadataRepositoryViewAdapter#getStore` to this function?
Essentially `store` contains everything we need and we can do all the decoration work here.

internal/venice-common/src/main/java/com/linkedin/venice/meta/ReadOnlyStoreView.java
```diff
@@ -0,0 +1,143 @@
...
+import java.util.Optional;
+import java.util.Properties;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+
+public class ReadOnlyStoreView extends ReadOnlyStore {
+  private final Map<Integer, ReadOnlyVersion> viewVersionMap;
+
+  public ReadOnlyStoreView(Store delegate, Map<Integer, ReadOnlyVersion> viewVersionMap) {
```
gaojieliu, line 21: I think we need to pass the `viewName` here, so that we can override `getName` and `getStoreName` in `ReadOnlyMaterializedViewVersion`, right?

gaojieliu: One reason we need this IMO is that we would like to treat each view as an individual store from the interface POV.

internal/venice-common/src/main/java/com/linkedin/venice/views/VeniceView.java
```diff
@@ -106,7 +107,50 @@ public static int parseVersionFromViewTopic(String topicName) {
...
+    return viewStoreNameWithoutPrefix.substring(0, storeNameEndIndex);
+  }
+
+  public static String getViewNameFromViewStoreName(String viewStoreName) {
+    int viewNameStartIndex = viewStoreName.lastIndexOf(VIEW_NAME_SEPARATOR);
+    return viewStoreName.substring(viewNameStartIndex + 1);
+  }
+
+  public static boolean isViewStore(String storeName) {
+    return storeName.startsWith(VIEW_STORE_PREFIX);
```
gaojieliu, line 141: Can these util methods handle `non-view` params?

gaojieliu (commented):
Left some more comments about the adapting layer?

BTW, can you remind me of the place where we can guarantee the uniqueness of view names?
One way we can do is to make sure the store name doesn't contain the special separator being used to concatenate the store name and view name.

xunyin8:
> Left some more comments about the adapting layer?
> 
> BTW, can you remind me of the place where we can guarantee the uniqueness of view names? One way we can do is to make sure the store name doesn't contain the special separator being used to concatenate the store name and view name.

Currently we don't check uniqueness of view names across different stores and I think that's fine. The way we make sure view store names can be processed is with added check to ensure store names do not contain view separator:
See `VeniceHelixAdmin#checkStoreNameConflict`
```
    if (VeniceView.isViewStore(storeName)) {
      throw new VeniceException(
          "Store name: " + storeName + " clashes with the internal usage, please remove the prefix: "
              + VeniceView.VIEW_STORE_PREFIX);
    }
```

clients/da-vinci-client/src/main/java/com/linkedin/davinci/repository/NativeMetadataRepositoryViewAdapter.java
```diff
@@ -0,0 +1,223 @@
...
+  }
+
+  @Override
+  public Store refreshOneStore(String storeName) {
+    return nativeMetadataRepository.refreshOneStore(VeniceView.getStoreName(storeName));
+  }
+
+  @Override
+  public List<Store> getAllStores() {
+    return nativeMetadataRepository.getAllStores();
```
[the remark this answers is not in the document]
xunyin8: I'd recommend to leave it out until the day when we need it. Since it could flood the list and it requires work to compute the full list based on active versions.

internal/venice-common/src/main/java/com/linkedin/venice/meta/ReadOnlyStoreView.java
xunyin8: Yeah, we will need to do this if we want to also provide capabilities to register listeners for view stores and trigger change events for all view stores when the original store changes.

internal/venice-common/src/main/java/com/linkedin/venice/views/VeniceView.java
xunyin8: Good catch, and the answer is no. I will add `isViewStore` checks and throw exceptions to the ones that is expecting a view store name.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/repository/NativeMetadataRepositoryViewAdapter.java
```diff
@@ -0,0 +1,223 @@
...
+    Store store = getStore(storeName);
+    if (store != null) {
+      return store;
+    }
+    throw new VeniceNoStoreException(storeName);
+  }
+
+  @Override
+  public boolean hasStore(String storeName) {
+    return nativeMetadataRepository.hasStore(VeniceView.getStoreName(storeName));
```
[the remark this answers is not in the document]
xunyin8: Currently the usage of `hasStore` of `NativeMetadataRepository` is very limited. Only used internally for sanity checks before fetching schemas. Currently if a view is deprecated from version pushes the DVC behavior is that it will do nothing since in `trySwapDaVinciCurrentVersion` and `trySubscribeDaVinciFutureVersion` it will not detect any future version. We might want to change/define the behavior first.

internal/venice-common/src/main/java/com/linkedin/venice/helix/HelixReadOnlyStoreViewConfigRepositoryAdapter.java
```diff
@@ -0,0 +1,40 @@
...
+    storeConfigRepository.refresh();
+  }
+
+  @Override
+  public void clear() {
+    storeConfigRepository.clear();
+  }
+
+  @Override
+  public Optional<StoreConfig> getStoreConfig(String storeName) {
```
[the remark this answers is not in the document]
xunyin8: I think we don't need it now since everything in a StoreConfig related is tightly coupled to the original store. i.e. regardless which view are we talking about the corresponding store's cluster discovery and migration flags are always the same.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/view/MaterializedViewWriter.java
```diff
@@ -63,16 +65,32 @@ public CompletableFuture<PubSubProduceResult> processRecord(
...
+   */
   @Override
-  public CompletableFuture<PubSubProduceResult> processRecord(ByteBuffer newValue, byte[] key, int newValueSchemaId) {
+  public CompletableFuture<PubSubProduceResult> processRecord(
+      ByteBuffer newValue,
+      byte[] key,
+      int newValueSchemaId,
+      boolean isChunkedKey) {
+    byte[] viewTopicKey = key;
+    if (isChunkedKey) {
```
xunyin8, line 86: Alternatively we could add a new API in the `VeniceWriter` to perform pass through for k/v but regenerates DIV. This would be useful later if we do decide to take the approach of forwarding chunks based on pre-populated message footer for actual large value chunking support.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/repository/NativeMetadataRepositoryViewAdapter.java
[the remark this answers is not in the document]
gaojieliu: Sure.
Ideally, we shouldn't allow this happen as view deprecation should make sure no active users of the deprecated views.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/repository/NativeMetadataRepositoryViewAdapter.java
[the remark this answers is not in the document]
gaojieliu: Makes sense.
Can you leave some javadoc for this function?

internal/venice-common/src/main/java/com/linkedin/venice/helix/HelixReadOnlyStoreViewConfigRepositoryAdapter.java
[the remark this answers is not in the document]
gaojieliu: Sure.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/repository/NativeMetadataRepositoryViewAdapter.java
```diff
@@ -0,0 +1,234 @@
...
+  @Override
+  public void registerStoreDataChangedListener(StoreDataChangedListener listener) {
+    StoreDataChangedListener adapterListener = new StoreDataChangedListenerViewAdapter(listener, this);
+    nativeMetadataRepository.registerStoreDataChangedListener(adapterListener);
+    storeDataChangedAdapterMap.put(listener, adapterListener);
+  }
+
+  @Override
+  public void unregisterStoreDataChangedListener(StoreDataChangedListener listener) {
+    storeDataChangedAdapterMap.compute(listener, (dataChangeListener, dataChangedListenerAdapter) -> {
```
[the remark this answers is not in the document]
gaojieliu: And we can use `computeIfPresent` function.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/repository/NativeMetadataRepositoryViewAdapter.java
```diff
@@ -0,0 +1,234 @@
...
+      if (dataChangedListenerAdapter != null) {
+        nativeMetadataRepository.unregisterStoreDataChangedListener(dataChangedListenerAdapter);
+      }
+      return dataChangedListenerAdapter;
```
gaojieliu, line 187: Shall we return `null` here so that map will remove this entry?
I have never used `compute` method to remove the entry, so please run some quick verification.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/repository/NativeMetadataRepositoryViewAdapter.java
```diff
@@ -0,0 +1,234 @@
...
+  }
+
+  @Override
+  public void unsubscribe(String storeName) {
+    String internalStoreName = VeniceView.getStoreName(storeName);
+    subscribedStoreMap.compute(internalStoreName, (k, s) -> {
+      if (s != null) {
+        s.remove(storeName);
+        if (s.isEmpty()) {
+          nativeMetadataRepository.unsubscribe(internalStoreName);
```
gaojieliu, line 219: Shall we return `null` to remove the entry when the set becomes empty?
Again, please verify whether returning `null` can remove the entry from map or not.

internal/venice-common/src/main/java/com/linkedin/venice/serialization/KeyWithChunkingSuffixSerializer.java
```diff
@@ -53,6 +53,13 @@ public ByteBuffer serializeNonChunkedKey(ByteBuffer key) {
     return serialize(key, serializedNonChunkKeySuffix);
   }
 
+  public byte[] getKeyFromChunkedKey(byte[] keyBytesWithChunkSuffix) {
```
gaojieliu, line 56: The assumption here is that the chunked key is only for the whole value, and it doesn't handle the value chunks.
How can we guarantee the caller always invokes this method with chunked key for the whole value entry?

internal/venice-common/src/main/java/com/linkedin/venice/views/VeniceView.java
```diff
@@ -106,7 +107,63 @@ public static int parseVersionFromViewTopic(String topicName) {
...
-    int versionEndIndex = versionStartIndex + topicName.substring(versionStartIndex).indexOf(VIEW_TOPIC_SEPARATOR);
+    int versionEndIndex = versionStartIndex + topicName.substring(versionStartIndex).indexOf(VIEW_NAME_SEPARATOR);
     return Integer.parseInt(topicName.substring(versionStartIndex, versionEndIndex));
   }
+
+  /**
+   * @param topicName for the view topic
+   * @return the corresponding store name and view name for the given view topic in a single string.
+   */
+  public static String parseStoreAndViewFromViewTopic(String topicName) {
```
gaojieliu, line 118: Can you share some example of `viewTopic` in the javadoc to make it easier to understand the logic?

gaojieliu (commented):
Code change looks good overall and left some more comments.

internal/venice-common/src/main/java/com/linkedin/venice/serialization/KeyWithChunkingSuffixSerializer.java
xunyin8: We cannot, that's why it's not a proper fix like I mentioned in the comments. Just a tactical fix for now that will allow stores that require chunking enabled (read compute) but doesn't actually have any large values. Meanwhile, we will work on proper chunking support in another PR.

gaojieliu: Got it.

gaojieliu (approved):
Thanks!

PR #1467, opened 2025-01-23
[DVC] Add RequestBasedMetaRepository to enable metadata retrieval directly from server

pthirun:
## [DVC] Add RequestBasedMetaRepository to enable metadata retrieval directly from server

This PR introduces the use of the `/store_properties/{store_name}` endpoint to retrieve and cache store metadata for the DaVinci client. The endpoint was implemented in the Venice Server in this PR: https://github.com/linkedin/venice/pull/1374.

The `RequestBasedMetaRepository` leverages this endpoint to maintain a cache of `Store` and `SchemaData` objects for requested stores. The cache is refreshed by the `NativeMetadataRepository`, which this new class extends.

By default, this feature is disabled in the `ClientConfig` passed to the `NativeMetadataRepository`. The default configuration preserves the current behavior of using the `ThinClientMetaStoreBasedRepository`. The relevant configuration field is `ClientConfig.useRequestBasedMetaRepository`. For more details, see [NativeMetadataRepository.getInstance](https://github.com/linkedin/venice/compare/main...pthirun:feature/dvc-request-based-metadata?expand=1#diff-146847da5b273e88de69dbd96a0e36d42c466e41cd628e680a073fdf45ada46fR116).

## How was this PR tested?

Unit and Integration test included in PR.

> NOTE: behavior will not change by default, feature flag in ClientConfig must be turned on.


clients/da-vinci-client/src/test/java/com/linkedin/davinci/repository/NativeMetadataRepositoryTest.java
```diff
@@ -159,21 +180,71 @@ protected TestNMR(ClientConfig clientConfig, VeniceProperties backendConfig, Clo
...
+          // The value schemas might be too large to be stored in a single K/V.
+          StoreMetaKey individualValueSchemaKey =
+              MetaStoreDataType.STORE_VALUE_SCHEMA.getStoreMetaKey(new HashMap<String, String>() {
+                {
+                  put(KEY_STRING_STORE_NAME, storeName);
+                  put(KEY_STRING_SCHEMA_ID, entry.getKey().toString());
+                }
+              });
+          // Empty string is not a valid value schema therefore it's safe to throw exceptions if we also cannot find it
+          // in
```
xunyin8, line 236: NIT: interesting new line styling

clients/da-vinci-client/src/test/java/com/linkedin/davinci/repository/NativeMetadataRepositoryTest.java
```diff
@@ -159,21 +180,71 @@ protected TestNMR(ClientConfig clientConfig, VeniceProperties backendConfig, Clo
...
-    protected Store getStoreFromSystemStore(String storeName, String clusterName) {
+    protected Store fetchStoreFromRemote(String storeName, String clusterName) {
       Store store = mock(Store.class);
       when(store.getName()).thenReturn(storeName);
       when(store.getReadQuotaInCU()).thenReturn(1L);
       return store;
     }
 
     @Override
+    protected SchemaData getSchemaData(String storeName) {
```
xunyin8, line 198: With your refactor where we pushed a lot of the system store specific logic down to the `ThinClientMetaStoreBased` and the intent of TestNMR was to provide unit testing for common logic in `NativeMetadataRepository` I wonder if we need to provide such complicated and system store specific implementation of `getSchemaData`. Can just mock it with some in memory map for testing?

xunyin8 (commented):
Looks good overall, thanks for the refactoring. Just some minor comments and also looks like test coverage is not passing.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/repository/NativeMetadataRepository.java
```diff
@@ -128,8 +120,12 @@ public static NativeMetadataRepository getInstance(
     LOGGER.info(
         "Initializing {} with {}",
         NativeMetadataRepository.class.getSimpleName(),
-        ThinClientMetaStoreBasedRepository.class.getSimpleName());
-    return new ThinClientMetaStoreBasedRepository(clientConfig, backendConfig, icProvider);
+        RequestBasedMetaRepository.class.getSimpleName());
```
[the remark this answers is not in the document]
pthirun: Good catch, thanks!

clients/da-vinci-client/src/test/java/com/linkedin/davinci/repository/NativeMetadataRepositoryTest.java
pthirun: This is probably the linter, will shorten comment

clients/da-vinci-client/src/main/java/com/linkedin/davinci/repository/NativeMetadataRepository.java
```diff
@@ -171,20 +172,14 @@ public boolean hasStore(String storeName) {
...
         throw new VeniceException("StoreConfig is missing unexpectedly for store: " + storeName);
       }
-      Store newStore = getStoreFromSystemStore(storeName, storeConfig.getCluster());
-      // isDeleting check to detect deleted store is only supported by meta system store based implementation.
-      if (newStore != null && !storeConfig.isDeleting()) {
-        putStore(newStore);
-        getAndCacheSchemaDataFromSystemStore(storeName);
-        nativeMetadataRepositoryStats.updateCacheTimestamp(storeName, clock.millis());
-      } else {
-        removeStore(storeName);
```
xunyin8, line 186: Maybe put a comment on `NativeMetadataRepository` that it does not handle `StoreDataChangedListener#handleStoreDeleted` events? Since the current implementation will throw exceptions during metadata fetch and not really differentiating store deleted vs unable to retrieve metadata.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/MetaSystemStoreTest.java
```diff
@@ -287,6 +289,42 @@ public void testThinClientMetaStoreBasedRepository() throws InterruptedException
...
+      // Not providing a CLIENT_META_SYSTEM_STORE_VERSION_MAP, should use the default value of 1 for system store
+      // current version.
+      VeniceProperties backendConfig = new PropertyBuilder().put(CLIENT_USE_SYSTEM_STORE_REPOSITORY, true)
+          .put(CLIENT_SYSTEM_STORE_REPOSITORY_REFRESH_INTERVAL_SECONDS, 1)
+          .build();
+      nativeMetadataRepository = NativeMetadataRepository.getInstance(clientConfig, backendConfig);
+      nativeMetadataRepository.start();
+      // ThinClientMetaStoreBasedRepository implementation should be used since CLIENT_USE_META_SYSTEM_STORE_REPOSITORY
+      // is set to true without enabling other feature flags.
+      Assert.assertTrue(nativeMetadataRepository instanceof RequestBasedMetaRepository);
```
xunyin8, line 311: The comment doesn't match the assertion and also this is a weird place to put a  test for `RequestBasedMetaRepository` in a file called `MetaSystemStoreTest`. Can we add a new test file where we run a more comprehensive integration test with a DVC? i.e push some new versions or make some store config changes and make sure the DVC pick up those changes. You can see examples like the recently added `testBatchOnlyMaterializedViewDVCConsumer`. You probably don't need a `VeniceTwoLayerMultiRegionMultiClusterWrapper`, a single region will be sufficient.

xunyin8 (commented):
Thanks for the changes and unit tests, I think we are a bit lacking on the integration tests. See comments below.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/repository/NativeMetadataRepository.java
pthirun: Wrote a comment above the method stating it will throw the VeniceNoStoreException on either deleting or missing stores. Is this an accurate representation?

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestDaVinciRequestBasedMetaRepository.java
```diff
@@ -0,0 +1,266 @@
...
+    return pushJobAvroDataDirs.get(storeName);
+  }
+
+  private void runPushJob(String storeName, Schema schema) {
+
+    ControllerClient controllerClient;
+    File dataDir = getPushJobAvroFileDirectory(storeName);
+    String dataDirPath = "file:" + dataDir.getAbsolutePath();
+
+    if (!controllerClients.containsKey(storeName)) {
```
pthirun, line 243: Using the ContollerClients map to keep track of created stores. I think this is appropriate since the `IntegrationTestPushUtils.createStoreForJob` method returns a controllerClient, even though we only need one instance of the ControllerClient to create new value schemas.

Digging into the createStoreForJob method, it does look like the `D2ControllerClientFactory` caches the controllerClient by `(clusterName, d2ServiceName,d2ZkHost)`. So this should result in all of the clients in this map to point to the same client instance.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/repository/NativeMetadataRepository.java
```diff
@@ -70,7 +62,7 @@ public abstract class NativeMetadataRepository
   private final Map<String, StoreConfig> storeConfigMap = new VeniceConcurrentHashMap<>();
   // Local cache for key/value schemas. SchemaData supports one key schema per store only, which may need to be changed
   // for key schema evolvability.
-  private final Map<String, SchemaData> schemaMap = new VeniceConcurrentHashMap<>();
+  protected Map<String, SchemaData> schemaMap = new VeniceConcurrentHashMap<>();
```
pthirun, line 65: Removing the `final` here so we may initialize this map in unit test mocking.

xunyin8: Could you point me to the unit test you are talking about? I don't see why this is necessary to achieve unit test mocking.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/MetaSystemStoreTest.java
```diff
@@ -448,15 +451,28 @@ private void verifyRepository(NativeMetadataRepository nativeMetadataRepository,
...
-          parentControllerClient.createNewStore(storeName, "test_owner", INT_KEY_SCHEMA, valueSchema).isError());
+      NewStoreResponse resp =
+          parentControllerClient.createNewStore(storeName, "test_owner", INT_KEY_SCHEMA, valueSchema);
+      if (resp.isError()) {
+        System.out.println("Create new store failed: " + resp.getError());
+      }
+      assertFalse(resp.isError());
+      assertFalse(parentControllerClient.emptyPush(storeName, "test-push-job", 100).isError());
+      try {
+        Thread.sleep(10000);
```
xunyin8, line 473: We are already using `TestUtils.waitForNonDeterministicPushCompletion`. We should avoid using `Thread.sleep` and why do we even need it here?.

xunyin8 (commented):
Change looks good, just one minor concern about `Thread.sleep` usage in test.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/repository/NativeMetadataRepository.java
pthirun: [RequestBasedMetaRepositoryTest.java](https://github.com/linkedin/venice/blob/ea4f353c9665f0df21cbfb6cc12769d068d06df0/clients/da-vinci-client/src/test/java/com/linkedin/davinci/repository/RequestBasedMetaRepositoryTest.java#L167) and [ThinClientMetaStoreBasedRepositoryTest.java](https://github.com/linkedin/venice/blob/ea4f353c9665f0df21cbfb6cc12769d068d06df0/clients/da-vinci-client/src/test/java/com/linkedin/davinci/repository/ThinClientMetaStoreBasedRepositoryTest.java#L97). 
In the unit tests, I am partially mocking the client object by using Mockito's `thenCallRealMethod`.

pthirun: The mocked clients do not seem to run any initialization on private final class variables, so these maps need to be initialized by the test.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestDaVinciRequestBasedMetaRepository.java
xunyin8: Typically we share the cluster environment across tests because a cluster is expensive and the things we are testing for generally don't break the cluster environment. However, I'd recommend you stay away from sharing stores across different tests. Stores are cheaper to create and if certain test fail it might leave the store in a bad state where other tests will also fail. This will lead to scenarios where you see 10 tests fail but really only 1 test failed and caused the store to be in a bad state. The order of test execution might also have side effects on the result of the test (making it flaky). Can we refactor this class so we are not sharing stores across tests?

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/MetaSystemStoreTest.java
pthirun: The tests were not passing early on due to the stores not being ready fast enough, however removing it now and testing locally seems to work. I pushed changes to remove this sleep, waiting on CI to finish.

xunyin8 (approved):

PR #1468, opened 2025-01-23
Limit maximum number of lines changed in file

majisourav99:
## Limit maximum number of lines changed in file

During  reviewing large PR, reviewer often times miss out critical bugs which later can cause production issues. 
This PR limits the maximum number of lines changed in a PR to 500. It excludes test file changes.
To override the check, developer can add OVERRRIDE keyword in the title or description.

## How was this PR tested?

CI

### DIFF
--- a/.github/workflows/limit-pr-files-changed.yml
+++ b/.github/workflows/limit-pr-files-changed.yml
@@ -1,0 +1,70 @@
+name: Enforce Max Lines Changed Per File
+
+on:
+  pull_request:
+    paths:
+      - '**/*.java'  # Monitor changes to Java files only
+
+jobs:
+  enforce-lines-changed:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - name: Get PR Metadata
+        id: pr_details
+        uses: actions-ecosystem/action-get-PR@v2
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Check for Override Keyword
+        id: check_override
+        run: |
+          # Define the keyword for override
+          OVERRIDE_KEYWORD="override"
+          
+          # Check PR title and body for the keyword
+          if echo "${{ steps.pr_details.outputs.title }}${{ steps.pr_details.outputs.body }}" | grep -iq "$OVERRIDE_KEYWORD"; then
+            echo "Override keyword found. Skipping validation."
+            echo "override=true" >> $GITHUB_OUTPUT
+          else
+            echo "override=false" >> $GITHUB_OUTPUT
+          fi
+      - name: Calculate Lines Changed Per File
+        id: lines_changed_per_file
+        run: |
+          # Define the maximum allowed lines changed per file
+          MAX_LINES=500
+
+          # Get the diff of the PR and process each file
+          EXCEEDED=false
+          while IFS=$'\t' read -r added removed file; do
+          # Skip test files
+          if [[ "$file" == *Test.java ]]; then
+            echo "Skipping test file: $file"
+            continue
+          fi
+
+          # Calculate total lines changed for the file
+          TOTAL=$((added + removed))
+          echo "File: $file, Lines Changed: $TOTAL"
+
+          # Check if the total exceeds the max allowed
+          if [ "$TOTAL" -gt "$MAX_LINES" ]; then
+            echo "File $file exceeds the maximum allowed lines changed ($TOTAL > $MAX_LINES)"
+            EXCEEDED=true
+          fi
+          done < <(git diff --numstat origin/main | grep '\.java$')
+
+          # Fail if any file exceeded the limit
+          if [ "$EXCEEDED" = true ]; then
+            echo "Some files exceed the maximum allowed lines changed ($MAX_LINES)."
+            exit 1
+          fi
+
+      - name: Notify
+        if: failure()
+        run: |
+          echo "One or more files in the PR exceed the maximum allowed lines changed. Please review and adjust your changes to your files."
--- a/.github/workflows/limit-pr-files-changed.yml
+++ b/.github/workflows/limit-pr-files-changed.yml
@@ -42,7 +42,7 @@ jobs:
           EXCEEDED=false
           while IFS=$'\t' read -r added removed file; do
           # Skip test files
-          if [[ "$file" == *Test.java ]]; then
+          if [[ "${file,,}" == *test.java ]]; then
             echo "Skipping test file: $file"
             continue
           fi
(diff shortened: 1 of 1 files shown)


.github/workflows/limit-pr-files-changed.yml
```diff
@@ -0,0 +1,70 @@
+name: Enforce Max Lines Changed Per File
+
+on:
+  pull_request:
+    paths:
+      - '**/*.java'  # Monitor changes to Java files only
```
FelixGV, line 6: Can we also include `.avsc` and `.proto` files please?

Schema changes ideally deserve dedicated PRs anyway, but without making things too complicated at first, we could at least include them in the count?

.github/workflows/limit-pr-files-changed.yml
```diff
@@ -0,0 +1,70 @@
...
+        id: lines_changed_per_file
+        run: |
+          # Define the maximum allowed lines changed per file
+          MAX_LINES=500
+
+          # Get the diff of the PR and process each file
+          EXCEEDED=false
+          while IFS=$'\t' read -r added removed file; do
+          # Skip test files
+          if [[ "${file,,}" == *test.java ]]; then
```
FelixGV, line 45: Does this mean test class files must end with the suffix `Test`? That does not seem right, since we don't follow this convention consistently enough. Can we instead rely on the full path needing to contain `src/main` in order to be included?

majisourav99: Good idea! Let me update that  to check for full path, though developer should include "test" somewhere in the class name.

.github/workflows/limit-pr-files-changed.yml
majisourav99: But for new protocol updates, 'added' will count the entire file as added which would be very high most of the time. Do we ever do inplace avsc changes?

FelixGV: Well that is the point, protocol changes should be reviewed in isolation, prior to any code getting written 😁

.github/workflows/limit-pr-files-changed.yml
```diff
@@ -0,0 +1,70 @@
...
+          # Check if the total exceeds the max allowed
+          if [ "$TOTAL" -gt "$MAX_LINES" ]; then
+            echo "File $file exceeds the maximum allowed lines changed ($TOTAL > $MAX_LINES)"
+            EXCEEDED=true
+          fi
+          done < <(git diff --numstat origin/main | grep '\.java$')
+
+          # Fail if any file exceeded the limit
+          if [ "$EXCEEDED" = true ]; then
+            echo "Some files exceed the maximum allowed lines changed ($MAX_LINES)."
```
huangminchn, line 63: Can we change "Some files" to "Above files"?

.github/workflows/limit-pr-files-changed.yml
```diff
@@ -0,0 +1,70 @@
...
+          # Calculate total lines changed for the file
+          TOTAL=$((added + removed))
+          echo "File: $file, Lines Changed: $TOTAL"
+
+          # Check if the total exceeds the max allowed
+          if [ "$TOTAL" -gt "$MAX_LINES" ]; then
+            echo "File $file exceeds the maximum allowed lines changed ($TOTAL > $MAX_LINES)"
+            EXCEEDED=true
+          fi
+          done < <(git diff --numstat origin/main | grep '\.java$')
```
huangminchn, line 59: Just curious, how does "added" and "removed" get the number in the first loop?

.github/workflows/limit-pr-files-changed.yml
```diff
@@ -0,0 +1,70 @@
...
+            continue
+          fi
+
+          # Calculate total lines changed for the file
+          TOTAL=$((added + removed))
+          echo "File: $file, Lines Changed: $TOTAL"
+
+          # Check if the total exceeds the max allowed
+          if [ "$TOTAL" -gt "$MAX_LINES" ]; then
+            echo "File $file exceeds the maximum allowed lines changed ($TOTAL > $MAX_LINES)"
```
huangminchn, line 56: Something to explore a bit: can this error log be highlighted?

.github/workflows/limit-pr-files-changed.yml
```diff
@@ -0,0 +1,70 @@
...
+        id: pr_details
+        uses: actions-ecosystem/action-get-PR@v2
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Check for Override Keyword
+        id: check_override
+        run: |
+          # Define the keyword for override
+          OVERRIDE_KEYWORD="override"
```
huangminchn, line 26: Can we make the keyword a bit more specific please, like "DIFFSIZEOVERRIDE"?

huangminchn (commented):
Thanks a lot Sourav!

.github/workflows/limit-pr-files-changed.yml
majisourav99: The input to the while loop is from cli `done < <(git diff --numstat upstream/main | grep '\.java$')` which will look like
71      0       .github/workflows/limit-pr-files-changed.yml
The first number is line added, second number is lines deleted and third one is filename which are copied to added/removed varaible.

.github/workflows/limit-pr-files-changed.yml
majisourav99: Not sure how to do it. :(

huangminchn: It's alright; let's see how the current setup goes.

.github/workflows/limit-pr-files-changed.yml
```diff
@@ -0,0 +1,87 @@
...
+          # Calculate total lines changed for the file
+          TOTAL=$((added + removed))
+          echo "File: $file, Lines Changed: $TOTAL"
+
+          # Check if the total exceeds the max allowed
+          if [ "$TOTAL" -gt "$MAX_LINES" ]; then
+            echo "File $file exceeds the maximum allowed lines changed ($TOTAL > $MAX_LINES)"
+            EXCEEDED=true
+          fi
+          done < <(git diff --numstat upstream/main | grep -E '\.(java|avsc|proto)$')
```
huangminchn, line 70: Just double check, you changed the target branch from origin to upstream, is it intended?

huangminchn (approved):
Thanks a lot Sourav! Looks good overall.

.github/workflows/limit-pr-files-changed.yml
majisourav99: changed back to origin

huangminchn (approved):
Thanks!

PR #1469, opened 2025-01-23
[server] A few changes for separate RT ingestion

sixpluszero:
## [server] A few changes for separate RT ingestion
Fixed two issues found in tests:
(1) region level metrics should not be resolved to non-sep region, it will cause confusion as we can't tell whether it is coming from RT or sepRT.
(2) Add separate pool and throttler for sep RT in dedicated AAWC leader consumer pool mode. This is addition to original implementation in current version prioritization mode separate pool, as not all clusters have rolled out this feature.

## How was this PR tested?

Updated unit test.

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/config/VeniceServerConfig.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/config/VeniceServerConfig.java
@@ -84,6 +84,7 @@ import static com.linkedin.venice.ConfigKeys.SERVER_DB_READ_ONLY_FOR_BATCH_ONLY_
 import static com.linkedin.venice.ConfigKeys.SERVER_DEBUG_LOGGING_ENABLED;
 import static com.linkedin.venice.ConfigKeys.SERVER_DEDICATED_CONSUMER_POOL_FOR_AA_WC_LEADER_ENABLED;
 import static com.linkedin.venice.ConfigKeys.SERVER_DEDICATED_CONSUMER_POOL_SIZE_FOR_AA_WC_LEADER;
+import static com.linkedin.venice.ConfigKeys.SERVER_DEDICATED_CONSUMER_POOL_SIZE_FOR_SEP_RT_LEADER;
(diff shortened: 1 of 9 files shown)


gaojieliu (approved):
The code change itself looks good.
I have some question about rolling this feature out and rolling back.
Rolling back this feature seems to be a little risky if we don't have additional safeguard.
For example, there is a hybrid store is doing inc push + ttl repush, and if we enable separate in version N, and disable it in version N + 1, will tt repush for version N + 1 look at both RT topics?

sixpluszero:
> The code change itself looks good. I have some question about rolling this feature out and rolling back. Rolling back this feature seems to be a little risky if we don't have additional safeguard. For example, there is a hybrid store is doing inc push + ttl repush, and if we enable separate in version N, and disable it in version N + 1, will tt repush for version N + 1 look at both RT topics?

no it will not, as N+1 does not have this feature enabled. But if we are rolling back the feature, we should first switch the write back to normal RT first, then we do the repush, and then VT for version N contains existing writes, and N+1 will monitor new writes from RT

gaojieliu:
> > The code change itself looks good. I have some question about rolling this feature out and rolling back. Rolling back this feature seems to be a little risky if we don't have additional safeguard. For example, there is a hybrid store is doing inc push + ttl repush, and if we enable separate in version N, and disable it in version N + 1, will tt repush for version N + 1 look at both RT topics?
> 
> no it will not, as N+1 does not have this feature enabled. But if we are rolling back the feature, we should first switch the write back to normal RT first, then we do the repush, and then VT for version N contains existing writes, and N+1 will monitor new writes from RT

Got it, the rollback scenario is tricky and if we don't consider it carefully, it might result in data loss.

I wondered whether we should add some safeguard or not:
1. Version N has separate RT enabled.
2. Version N+1 doesn't.
3. When doing repush or ttl repush, make sure the source fabric fully drains the separate RT.

Also for the feature rollout, if we enable this feature in version N, backup version N - 1 doesn't have this feature enabled, so it won't consume the data in separate RT topic, which will result in a data gap.
We need to do a repush to get rid of the version without this feature enabled.

Let us write down these steps in a runbook since it is a little tricky.

PR #1470, opened 2025-01-23
[controller] Update parent version status to account for child version status

misyel:

Previously the parent version status is updated to `PUSHED` for a target region push if the version is not null and not already set to `PUSHED`. And in the next poll, it is set to `ONLINE` because the above conditions are no longer met. This, however, is inaccurate because it doesn't check if the child region is serving the new version already before changing the status to `ONLINE`. This change modifies the logic to check if all child regions are on the latest version before marking the parent version status as `ONLINE`

## How was this PR tested?

Updated integration tests

### DIFF
--- a/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
@@ -3661,9 +3661,9 @@ public class VeniceParentHelixAdmin implements Admin {
         Store parentStore = repository.getStore(storeName);
         Version version = parentStore.getVersion(versionNum);
         boolean isDeferredSwap = version != null && version.isVersionSwapDeferred();
-        if (!isDeferredSwap) {
-          // targetedRegions is non-empty for target region push of batch store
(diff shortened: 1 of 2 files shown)


services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
```diff
@@ -3683,14 +3683,27 @@ private OfflinePushStatusInfo getOffLineJobStatus(
...
+          /**
+           * Check if the latest version is online and is completed in all regions.
+           * If it is online in all regions, set parent version status to ONLINE
+           * Otherwise, set parent version status to PUSHED
+           */
+          Map<String, Integer> currentVersionsToColo = getCurrentVersionsForMultiColos(clusterName, storeName);
+          int latestVersionNum = parentStore.getLargestUsedVersionNumber();
+          boolean isLatestVersionOnlineInAllRegions = true;
+          for (Map.Entry<String, Integer> entry: currentVersionsToColo.entrySet()) {
+            if (currentVersionsToColo.get(entry.getKey()) != latestVersionNum) {
```
majisourav99, line 3696: This assumption is not correct, the versions in parent need not be same as in child.

misyel:
Moving this change to be target region push specific & to #1421

PR #1471, opened 2025-01-24
[dvc][compat] Add blob transfer support for DaVinciRecordTransformer

kvargha:
## [dvc][compat] Add blob transfer support for DaVinciRecordTransformer

This PR makes DVRT compatible with blob transfer. Here are the changes made:
1. Throw an exception when database checksum verification is enabled. This is because DVRT transforms the values, and when the transformed checksum is compared with the expected checksum the validation will fail. Due to the nature of this feature, it makes sense to not support it and checksum verification is off by default anyway for DVC.
2. Store the `classHash` in the `offsetRecord` so it can be transported during blob transfer. This is to ensure that the blob that was transferred is compatible with the current transformer logic.
3. Removed file based `classHash` persistence as it is now stored in the `offsetRecord`.
4. Fixed a bug in SIT where the `classHash` was being calculated based on the wrapper class not the user's class.

## How was this PR tested?

Add unit and integration tests.

DVRT is no longer compatible with database checksum verification, but this is fine since DVRT is not currently being used.


.gitignore
```diff
@@ -23,5 +23,3 @@ _site/
 Gemfile.lock
 .bundles_cache
 docs/vendor/
-clients/da-vinci-client/classHash*.txt
```
sixpluszero, line 26: Why do we remove this? classHash should not be included.

internal/venice-common/src/main/resources/avro/PartitionState/v14/PartitionState.avsc
```diff
@@ -186,6 +186,15 @@
...
       "default": {}
+    },
+    {
+      "name": "recordTransformerClassHash",
+      "doc": "An integer hash code used by the DaVinciRecordTransformer to detect changes to the user's class during bootstrapping.",
+      "type": [
+        "null",
+        "int"
+      ],
+      "default": null
```
jingy-li, line 197: We may need to copy this to a new file under v15.

internal/venice-common/src/main/resources/avro/PartitionState/v14/PartitionState.avsc
```diff
@@ -186,6 +186,15 @@
         "avro.java.string": "String"
       },
       "default": {}
+    },
```
sixpluszero, line 189: Can you add [compat] to the PR title so we know there is a protocol changes, even though it is Server/DVC only.

internal/venice-common/src/main/resources/avro/PartitionState/v14/PartitionState.avsc
```diff
@@ -186,6 +186,15 @@
...
+    {
```
sixpluszero, line 190: We should evolve the schema, instead of doing in place upgrade.

sixpluszero (commented):
I have not done the pass on the actual logic yet but I just leave a few general comments. We should consider fixing these protocol changes as well.

.gitignore
kvargha: This file will no longer get generated as the classHash is now stored in the offset record.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/DaVinciClientRecordTransformerTest.java
```diff
@@ -486,6 +487,96 @@ public void testUnchangedResultRecordTransformer(DaVinciConfig clientConfig) thr
...
+        backendConfig2)) {
+      DaVinciClient<Integer, Object> client2 = factory2.getAndStartGenericAvroClient(storeName, clientConfig);
+      client2.subscribeAll().get();
+
+      // Verify snapshots exists
+      for (int i = 0; i < 3; i++) {
+        String snapshotPath = RocksDBUtils.composeSnapshotDir(dvcPath1 + "/rocksdb", storeName + "_v1", i);
+        Assert.assertTrue(Files.exists(Paths.get(snapshotPath)));
+      }
+
```
jingy-li, line 566: Do we need to verify that the `RecordTransformerClassHash` info in the offset record is correct?

kvargha: It will be difficult to do that in an integration test as I don't see a good way to access the `storageEngine` to fetch the `offsetRecord`. But I did add an extremely trivial validation in a unit test.

jingy-li (approved):
Overall looks good! Thank you for the efforts! 

One concern is that in the blob transfer, the `partitionStateSerializer` is using the default one:  
`InternalAvroSpecificSerializer<PartitionState> partitionStateSerializer = AvroProtocolDefinition.PARTITION_STATE.getSerializer();`  

Please validate in E2E testing on the i353 hosts to ensure that the `recordTransformer.getClassHash()` value is correct.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -323,7 +323,7 @@ public abstract class StoreIngestionTask implements Runnable, Closeable {
   private final Schema recordTransformerInputValueSchema;
   private final AvroGenericDeserializer recordTransformerKeyDeserializer;
   private final SparseConcurrentList<AvroGenericDeserializer> recordTransformerDeserializersByPutSchemaId;
-  private DaVinciRecordTransformer recordTransformer;
+  private BlockingDaVinciRecordTransformer recordTransformer;
```
ZacAttack, line 326: Why do we need to downcast this?

kvargha: It's because I need to add a wrapper (`internalOnRecovery`) around the call to `onRecovery` to fix the bug where the `classHash` was being calculated based on the `BlockingDaVinciRecordTransformer` and not the user's implementation.

kvargha:
> Overall looks good! Thank you for the efforts!
> 
> One concern is that in the blob transfer, the `partitionStateSerializer` is using the default one: `InternalAvroSpecificSerializer<PartitionState> partitionStateSerializer = AvroProtocolDefinition.PARTITION_STATE.getSerializer();`

I'm not sure I understand this @jingy-li. DVRT is using the `partitionStateSerializer` that's created inside [DaVinciBackend](https://github.com/linkedin/venice/blob/5dfe33b8e653c64c8e7d42efabc5ec6e855753f0/clients/da-vinci-client/src/main/java/com/linkedin/davinci/DaVinciBackend.java#L152-L153) which gets propagated to `StoreIngestionTask` then to DVRT.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/DaVinciRecordTransformerUtility.java
```diff
@@ -80,45 +79,40 @@ public final ByteBuffer prependSchemaIdToHeader(ByteBuffer valueBytes, int schem
...
       Lazy<VeniceCompressor> compressor) {
-    // ToDo: Store class hash in RocksDB to support blob transfer
     int classHash = recordTransformer.getClassHash();
-    boolean transformerLogicChanged = hasTransformerLogicChanged(classHash);
+    Optional<OffsetRecord> optionalOffsetRecord = storageEngine.getPartitionOffset(partitionId);
+    OffsetRecord offsetRecord = optionalOffsetRecord.orElseGet(() -> new OffsetRecord(partitionStateSerializer));
+
+    boolean transformerLogicChanged = hasTransformerLogicChanged(classHash, offsetRecord);
 
     if (!recordTransformer.getStoreRecordsInDaVinci() || transformerLogicChanged) {
```
kvargha, line 105: If a user wants to redirect their records to an external storage, I don't think we should be bootstrapping from VT all of the time. Only when a change in transformer logic has been detected.

Instead of checking ` !recordTransformer.getStoreRecordsInDaVinci()`, we should instead add a boolean config something like `alwaysBootstrapFromBeginning`. A user piping their records to memory may use this, but it would probably be better for boostrapping time for them to store both in RocksDB and RAM.

What do you think @ZacAttack

kvargha: If we do go this route, I'm thinking of moving `storeRecordsInDaVinci` to `DaVinciRecordTransformerConfig` to clean up DVRT's constructor a bit.

.gitignore
kvargha: Ok we still need this. I switched to a different branch that didn't have this change in logic and still generated the file.  After switching back to this branch git is trying to add it.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/BlockingDaVinciRecordTransformer.java
```diff
@@ -58,6 +62,19 @@ public void onEndVersionIngestion(int currentVersion) {
...
+      AbstractStorageEngine storageEngine,
+      int partitionId,
+      InternalAvroSpecificSerializer<PartitionState> partitionStateSerializer,
+      Lazy<VeniceCompressor> compressor) {
+    // Using a wrapper around onRecovery because when calculating the class hash it grabs the name of the current class
+    // that is invoking it. If we directly invoke onRecovery from this class, the class hash will be calculated based
+    // on the contents of BlockingDaVinciRecordTransformer, not the user's implementation of DVRT.
+    // We also can't override onRecovery like the other methods because this method is final and the implementation
+    // should never be overriden.
+    this.recordTransformer.onRecovery(storageEngine, partitionId, partitionStateSerializer, compressor);
```
sixpluszero, line 75: hmm, I don't get it..wouldn't this getHash always be called by the transformer impl?

sixpluszero (commented):
Overall looks good, just minor nit and one question

clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/BlockingDaVinciRecordTransformer.java
kvargha: In `StoreIngestionTask`, the DVRT being called is a `BlockingDaVinciRecordTransformer`, which acts as a wrapper around the user's DVRT. When `BlockingDaVinciRecordTransformer::onRecovery` is called, it invokes the `DaVinciRecordTransformerUtility` created by `BlockingDaVinciRecordTransformer`. When `hasTransformerLogicChanged` is called, it uses the `BlockingDaVinciRecordTransformer` to calculate the classHash.

To ensure that the user's DVRT is used to calculate the `classHash`, we must invoke `onRecovery` on the user's implementation not on `BlockingDaVinciRecordTransformer`.

sixpluszero: Synced offline, and I think it is good.

ZacAttack (approved):

PR #1473, opened 2025-01-24
[duckdb] Created a CLI tool to interact with DuckDBDaVinciRecordTransformer

kvargha:
## [duckdb] Created a CLI tool to interact with DuckDBDaVinciRecordTransformer

This PR creates a CLI tool to interact with DuckDBDaVinciRecordTransformer, similar to the AdminTool.

## How was this PR tested?

I tested this locally pointed to a store that exists, but for the purpose of releasing this to OSS I generalized the variables.


integrations/venice-duckdb/src/main/java/com/linkedin/venice/duckdb/DuckDBDaVinciRecordTransformerExample.java
```diff
@@ -0,0 +1,78 @@
...
+
+    String outputDir = "/example/output/dir";
+    String storeName = "example_store_name";
+    Set<String> columnsToProject = Collections.emptySet();
+
+    String schema = IOUtils.toString(
+        Objects.requireNonNull(
+            DuckDBDaVinciRecordTransformerExample.class.getClassLoader().getResourceAsStream("ValueSchema.avsc")),
+        StandardCharsets.UTF_8);
+    Schema outputSchema = AvroCompatibilityHelper.parse(schema);
```
FelixGV, line 45: Can we get rid of this? We're not using it AFAIK. Maybe leaving it null will cause a NPE somewhere else, but we can probably guard against that somehow?

integrations/venice-duckdb/src/main/java/com/linkedin/venice/duckdb/DuckDBDaVinciRecordTransformerExample.java
```diff
@@ -0,0 +1,78 @@
...
+import org.apache.avro.generic.GenericRecord;
+import org.apache.commons.io.IOUtils;
+
+
+public class DuckDBDaVinciRecordTransformerExample {
+  public static void main(String[] args) throws Exception {
+    DaVinciConfig clientConfig = new DaVinciConfig();
+    String clusterDiscoveryD2ServiceName = "venice-discovery";
+    MetricsRepository metricsRepository = new MetricsRepository();
+    String zkHosts = "zk-example-url.com:12913";
```
FelixGV, line 32: Can we get this from the `args` rather than hardcoding? You can look at how the `AdminTool` leverages the `CommandLine` utility to handle CLI args parsing cleanly while minimizing boilerplate...

integrations/venice-duckdb/src/main/java/com/linkedin/venice/duckdb/DuckDBDaVinciRecordTransformerExample.java
```diff
@@ -0,0 +1,78 @@
...
+
+    VeniceProperties backendConfig =
+        new PropertyBuilder().put(DA_VINCI_CURRENT_VERSION_BOOTSTRAPPING_SPEEDUP_ENABLED, true).build();
+
+    String outputDir = "/example/output/dir";
+    String storeName = "example_store_name";
```
FelixGV, line 38: Should also be coming from `args`, ideally.

integrations/venice-duckdb/src/main/java/com/linkedin/venice/duckdb/DuckDBDaVinciRecordTransformerExample.java
```diff
@@ -0,0 +1,78 @@
...
+import java.util.Objects;
+import java.util.Set;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import org.apache.avro.Schema;
+import org.apache.avro.generic.GenericRecord;
+import org.apache.commons.io.IOUtils;
+
+
+public class DuckDBDaVinciRecordTransformerExample {
```
FelixGV, line 27: Can we call the class `DuckVinciTool`?

FelixGV (commented):
Thanks Koorous! Left some comments to try and make this more widely usable...

integrations/venice-duckdb/src/main/java/com/linkedin/venice/duckdb/DuckDBDaVinciRecordTransformerExample.java
kvargha: I can try to make some changes in the case where a user doesn't want to do any transformations so they don't need to pass this in. We can just rely on the value schema pulled from the schema repo.

integrations/venice-duckdb/src/main/java/com/linkedin/venice/duckdb/DuckVinciTool.java
```diff
@@ -0,0 +1,342 @@
...
+     * Gather all the options we have in "options"
+     */
+    Options options = new Options();
+    for (Arg arg: Arg.values()) {
+      createOpt(arg, arg.isParameterized(), arg.getHelpText(), options);
+    }
+
+    Options parameterOptionsForHelp = new Options();
+    for (Object obj: options.getOptions()) {
+      Option o = (Option) obj;
```
lusong64, line 190: should we use instanceOf to check whether the obj is an Option or we are pretty sure they are?

FelixGV (approved):
This is a super cool tool! Thanks a lot for putting it together!

Just one minor comment, but it could be addressed incrementally later on if you prefer, so you can decide either way.

PR #1475, opened 2025-01-27
[controller] Make Helix rebalance preferences and capacity keys configurable for the controller

kvargha:
## [controller] Make Helix rebalance preferences and capacity keys configurable for the controller

We are migrating from SEMI_AUTO to FULL_AUTO Waged. After thorough testing, we discovered a config that achieves overall even distribution for resource assignment.

1. Made the Helix rebalance preference configurable. This is helpful when using Waged, as we can configure it to priortize evenness.
2. Made instance capacity keys configurable, allowing Helix to consider the current top-state when calculating assignments.
3. To accommodate these configurations, I updated the logic of `updateClusterConfigs` to support more complex configurations beyond simple fields. This logic was also ported to `updateRESTConfigs`.
4. Enabled `PERSIST_BEST_POSSIBLE_ASSIGNMENT`, as this is the setting we use for our controller clusters.

## How was this PR tested?

Added unit and integration tests


services/venice-controller/src/main/java/com/linkedin/venice/controller/ZkHelixAdminClient.java
```diff
@@ -99,6 +100,32 @@ public void createVeniceControllerCluster() {
         // choose proper instance to hold the replica.
         clusterConfig.setTopologyAwareEnabled(false);
 
+        // We want to prioritize evenness over less movement when it comes to resource assignment, because the cost
+        // of rebalancing for the controller is cheap as it is stateless.
+        Map<ClusterConfig.GlobalRebalancePreferenceKey, Integer> globalRebalancePreference = new HashMap<>();
+        globalRebalancePreference.put(ClusterConfig.GlobalRebalancePreferenceKey.EVENNESS, 10);
+        globalRebalancePreference.put(ClusterConfig.GlobalRebalancePreferenceKey.LESS_MOVEMENT, 1);
+        // This should be turned off, so it doesn't overpower other constraint calculations
+        globalRebalancePreference.put(ClusterConfig.GlobalRebalancePreferenceKey.FORCE_BASELINE_CONVERGE, 0);
```
nisargthakkar, line 109: Let's make these configurable through configs. Is a value of -1 supported? If not, we can set that as a sentinel to skip setting that config

services/venice-controller/src/main/java/com/linkedin/venice/controller/ZkHelixAdminClient.java
```diff
@@ -99,6 +100,32 @@ public void createVeniceControllerCluster() {
...
+        clusterConfig.setGlobalRebalancePreference(globalRebalancePreference);
+
+        String resourceCapacityKey = "cluster_resource_weight";
```
nisargthakkar, line 112: Let's put this in a constants file

services/venice-controller/src/main/java/com/linkedin/venice/controller/ZkHelixAdminClient.java
```diff
@@ -99,6 +100,32 @@ public void createVeniceControllerCluster() {
...
+        List<String> instanceCapacityKeys = new ArrayList<>();
+        instanceCapacityKeys.add(resourceCapacityKey);
+        clusterConfig.setInstanceCapacityKeys(instanceCapacityKeys);
+
+        // This is how much capacity a participant can take. The Helix documentation recommends setting this to a high
+        // value to avoid rebalance failures. The primary goal of setting this is to enable a constraint that takes the
+        // current top-state distribution into account when rebalancing.
+        Map<String, Integer> defaultInstanceCapacityMap = new HashMap<>();
+        defaultInstanceCapacityMap.put(resourceCapacityKey, 10000);
```
nisargthakkar, line 121: Same for this, let's make these configurable through configs

services/venice-controller/src/main/java/com/linkedin/venice/controller/ZkHelixAdminClient.java
```diff
@@ -99,6 +100,32 @@ public void createVeniceControllerCluster() {
...
+        clusterConfig.setDefaultInstanceCapacityMap(defaultInstanceCapacityMap);
+
+        // This is how much weight each resource in a cluster has
+        Map<String, Integer> defaultPartitionWeightMap = new HashMap<>();
+        defaultPartitionWeightMap.put(resourceCapacityKey, 100);
```
nisargthakkar, line 126: Ditto

nisargthakkar:
Also, let's add some tests

services/venice-controller/src/main/java/com/linkedin/venice/controller/ZkHelixAdminClient.java
kvargha: Technically `EVENNESS` and `LESS_MOVEMENT` can be set to -1, but that will most likely mess up the ranking of the scores.

By default these values are set to 1, so I would recommend making the config values default to 1 too.

As for `FORCE_BASELINE_CONVERGE`, this is a boolean config represented in a binary format of 0 or 1, which is defaulted to 0. Based on my testing and looking over how the scores are ranked in the code, we should always have this turned off. We can also have the config default to 0 here, since that's what the Helix default is.

kvargha:
Working on the tests rn

internal/venice-common/src/main/java/com/linkedin/venice/ConfigConstants.java
```diff
@@ -20,6 +20,8 @@ public class ConfigConstants {
   public static final long DEFAULT_PUSH_STATUS_STORE_HEARTBEAT_EXPIRATION_TIME_IN_SECONDS =
       TimeUnit.MINUTES.toSeconds(10);
 
+  public static final String DEFAULT_HELIX_RESOURCE_CAPACITY_KEY = "cluster_resource_weight";
```
nisargthakkar, line 23: Can we put `CONTROLLER` in the name?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerClusterConfig.java
```diff
@@ -996,6 +1007,13 @@ public VeniceControllerClusterConfig(VeniceProperties props) {
     this.pushJobUserErrorCheckpoints = parsePushJobUserErrorCheckpoints(props);
     this.isHybridStorePartitionCountUpdateEnabled =
         props.getBoolean(ConfigKeys.CONTROLLER_ENABLE_HYBRID_STORE_PARTITION_COUNT_UPDATE, false);
+
+    this.helixRebalancePreferenceEvenness = props.getInt(CONTROLLER_HELIX_REBALANCE_PREFERENCE_EVENNESS, 1);
+    this.helixRebalancePreferenceLessMovement = props.getInt(CONTROLLER_HELIX_REBALANCE_PREFERENCE_LESS_MOVEMENT, 1);
+    this.helixRebalancePreferenceEnableForceBaselineConverge =
+        props.getBoolean(CONTROLLER_HELIX_REBALANCE_PREFERENCE_ENABLE_FORCE_BASELINE_CONVERGE, false);
+    this.helixInstanceCapacity = props.getInt(CONTROLLER_HELIX_INSTANCE_CAPACITY, 10000);
+    this.helixResourceCapacityWeight = props.getInt(CONTROLLER_HELIX_RESOURCE_CAPACITY_WEIGHT, 100);
```
nisargthakkar, line 1016: Are all of these mandatory? What if we don't want to set some fields? Can we have a sentinel value to ignore them?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -8777,4 +8777,8 @@ public void setPushJobDetailsStoreClient(AvroSpecificStoreClient<PushJobStatusRe
   public PubSubTopicRepository getPubSubTopicRepository() {
     return pubSubTopicRepository;
   }
+
+  public SafeHelixManager getHelixManager() {
```
nisargthakkar, line 8781: Can we make this package private since it is only for tests?

services/venice-controller/src/main/java/com/linkedin/venice/controller/ZkHelixAdminClient.java
```diff
@@ -215,21 +252,15 @@ public void addClusterToGrandCluster(String clusterName) {
...
    * @see HelixAdminClient#updateRESTConfigs(String, RESTConfig)
    */
   @Override
   public void updateRESTConfigs(String clusterName, RESTConfig restConfig) {
-    HelixConfigScope configScope =
-        new HelixConfigScopeBuilder(HelixConfigScope.ConfigScopeProperty.REST).forCluster(clusterName).build();
-    Map<String, String> helixRestProperties = new HashMap<>(restConfig.getRecord().getSimpleFields());
-    helixAdmin.setConfig(configScope, helixRestProperties);
+    helixConfigAccessor.setRESTConfig(clusterName, restConfig);
   }
```
nisargthakkar, line 264: Good change, thanks

services/venice-controller/src/main/java/com/linkedin/venice/controller/ZkHelixAdminClient.java
```diff
@@ -98,6 +102,52 @@ public void createVeniceControllerCluster() {
...
+              commonConfig.getHelixRebalancePreferenceForceBaselineConverge());
+        }
+
+        if (!globalRebalancePreference.isEmpty()) {
+          clusterConfig.setGlobalRebalancePreference(globalRebalancePreference);
+        }
+
+        if (commonConfig.getHelixInstanceCapacity() > -1 && commonConfig.getHelixResourceCapacityWeight() > -1) {
+          List<String> instanceCapacityKeys = new ArrayList<>();
+          instanceCapacityKeys.add(CONTROLLER_DEFAULT_HELIX_RESOURCE_CAPACITY_KEY);
```
nisargthakkar, line 134: nit: Can we change to `Collections.singletonList(CONTROLLER_DEFAULT_HELIX_RESOURCE_CAPACITY_KEY)` ?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -8835,6 +8835,10 @@ public PubSubTopicRepository getPubSubTopicRepository() {
     return pubSubTopicRepository;
   }
 
+  protected SafeHelixManager getHelixManager() {
```
nisargthakkar, line 8838: Can we remove the `protected` and have nothing? The default access specifier is tighter and allows access only within the same package while protected allows access within the same package and to subclasses of the class where the keyword is specified.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerClusterConfig.java
```diff
@@ -996,6 +1007,13 @@ public VeniceControllerClusterConfig(VeniceProperties props) {
     this.pushJobUserErrorCheckpoints = parsePushJobUserErrorCheckpoints(props);
     this.isHybridStorePartitionCountUpdateEnabled =
         props.getBoolean(ConfigKeys.CONTROLLER_ENABLE_HYBRID_STORE_PARTITION_COUNT_UPDATE, false);
+
+    this.helixRebalancePreferenceEvenness = props.getInt(CONTROLLER_HELIX_REBALANCE_PREFERENCE_EVENNESS, -1);
+    this.helixRebalancePreferenceLessMovement = props.getInt(CONTROLLER_HELIX_REBALANCE_PREFERENCE_LESS_MOVEMENT, -1);
+    this.helixRebalancePreferenceForceBaselineConverge =
+        props.getInt(CONTROLLER_HELIX_REBALANCE_PREFERENCE_FORCE_BASELINE_CONVERGE, -1);
+    this.helixInstanceCapacity = props.getInt(CONTROLLER_HELIX_INSTANCE_CAPACITY, -1);
+    this.helixResourceCapacityWeight = props.getInt(CONTROLLER_HELIX_RESOURCE_CAPACITY_WEIGHT, -1);
```
nisargthakkar, line 1016: Can we have the basic validations here instead of in `ZkHelixAdminClient` so that we can fail fast on controller startup rather than identifying issues only midway through new cluster creation. Also, instead of exposing these raw config values, we can expose materialized objects instead. Like: `helixGlobalRebalancePreference`, `helixInstanceCapacityKeys`, `helixDefaultInstanceCapacityMap`, `helixDefaultPartitionWeightMap`, etc.

nisargthakkar: And we don't have to expose the `-1` sentinel values that are unsupported by Helix out of this config interface

nisargthakkar (commented):
A couple of minor nits. LGTM

services/venice-controller/src/test/java/com/linkedin/venice/controller/TestZkHelixAdminClient.java
```diff
@@ -51,6 +59,10 @@ public void setUp() throws NoSuchFieldException, IllegalAccessException {
         helixAdminField.setAccessible(true);
         helixAdminField.set(zkHelixAdminClient, mockHelixAdmin);
 
+        Field helixConfigAccessorField = ZkHelixAdminClient.class.getDeclaredField("helixConfigAccessor");
+        helixConfigAccessorField.setAccessible(true);
+        helixConfigAccessorField.set(zkHelixAdminClient, mockHelixConfigAccessor);
+
```
nisargthakkar, line 65: Can we make `helixConfigAccessor` package-private (default access specifier) instead of reflection?

nisargthakkar (commented):
Mostly LGTM. A few test nits

services/venice-controller/src/test/java/com/linkedin/venice/controller/TestZkHelixAdminClient.java
kvargha: This would require making `helixConfigAccessor` non-final, and the two functions invoking it are assuming that it is defined. If a future PR manages to set `helixConfigAccessor` to null we can run into a NPE.

What if I added a getter to `helixConfigAccessor` and mock the return?

nisargthakkar: I think how you have it is better in that case

nisargthakkar (approved):
LGTM. Thanks a lot @kvargha!

PR #1477, opened 2025-01-27
[controller] Dual write push job details to both parent and child region

sixpluszero:
## [controller] Dual write push job details to both parent and child region
Add a dual write to child region for push job details, so when the system store is converted to A/A it will still be able to consume the updates.

The hard-coding of AGG mode in system store initialization routine will not be changed, as it will only kick in if no such store is found. We will update this after second phase when we remove the parent RT write.
## How was this PR tested?

Added new unit test and new integration test

### DIFF
--- a/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
@@ -180,6 +180,7 @@ import com.linkedin.venice.schema.rmd.RmdSchemaEntry;
 import com.linkedin.venice.schema.writecompute.DerivedSchemaEntry;
 import com.linkedin.venice.security.SSLFactory;
 import com.linkedin.venice.serialization.avro.AvroProtocolDefinition;
+import com.linkedin.venice.serialization.avro.InternalAvroSpecificSerializer;
 import com.linkedin.venice.serialization.avro.VeniceAvroKafkaSerializer;
 import com.linkedin.venice.serializer.AvroSerializer;
 import com.linkedin.venice.serializer.RecordDeserializer;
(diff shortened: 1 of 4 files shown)


services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -1292,10 +1295,40 @@ static void emitPushJobStatusMetrics(
...
+              value,
+              response.getError());
+          lastDualWriteError = response.getError();
+        } else {
+          // When A/A is enabled, we only need to write to one child region and it will be replicated to all regions.
+          return;
+        }
+      }
+      throw new VeniceException(
+          "Unable to dual write push job details to any child region with last error:" + lastDualWriteError);
```
ZacAttack, line 1323: This looks a little weird unless I'm counting my brackets wrong.  This looks like it'll error if this a child controller.  Shouldn't we update the message to include that case?

ZacAttack: Nevermind, I counted the brackets wrong

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -1292,10 +1295,40 @@ static void emitPushJobStatusMetrics(
...
+        if (response.isError()) {
+          LOGGER.error(
+              "Unable to send push job details to region: {}, for key: {}, value: {} for error: {}",
+              entry.getKey(),
+              key,
+              value,
+              response.getError());
+          lastDualWriteError = response.getError();
+        } else {
+          // When A/A is enabled, we only need to write to one child region and it will be replicated to all regions.
```
ZacAttack, line 1318: Do we need to do an AA check here then?

ZacAttack (approved):
Overall looks good, just left one question

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
sixpluszero: Hmm what do you mean? Today, when AGG mode is enabled, RT is created in all the fabrics. So even today writing to child region is good. I want to let it write to child for a while (like a day) and then do the store property upgrade and create a new version and then daily push job report won't be broken during the transition. And then in phase 2, I remove the write to parent fabric, and set the A/A mode in the system init routine and make the integration test changes.

PR #1478, opened 2025-01-28
[venice-thin-client] Avoid using common thread pool for store client warming up.

haoxu07:

The `venice-thin-client` utilizes the ForkJoinPool common pool, which can cause spurious timeouts during application startup, potentially resulting in startup failures. The recommended solution is to use a different thread pool to mitigate contention issues. Here we adopt a single thread pool for warm-up logic. 

### DIFF
--- a/clients/venice-thin-client/src/main/java/com/linkedin/venice/client/store/AbstractAvroStoreClient.java
+++ b/clients/venice-thin-client/src/main/java/com/linkedin/venice/client/store/AbstractAvroStoreClient.java
@@ -75,8 +75,7 @@ public abstract class AbstractAvroStoreClient<K, V> extends InternalAvroStoreCli
   private RecordDeserializer<StreamingFooterRecordV1> streamingFooterRecordDeserializer;
   private TransportClient transportClient;
   private final Executor deserializationExecutor;
-  private final Executor veniceClientWarmUpExecutor =
-      Executors.newFixedThreadPool(1, new DaemonThreadFactory("Venice-Client-Warmup"));
+  private final Executor veniceClientWarmUpExecutor;
   private final CompressorFactory compressorFactory;
   private final String storageRequestPath;
   private final String computeRequestPath;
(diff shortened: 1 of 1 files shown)


clients/venice-thin-client/src/main/java/com/linkedin/venice/client/store/AbstractAvroStoreClient.java
```diff
@@ -75,6 +75,8 @@ public abstract class AbstractAvroStoreClient<K, V> extends InternalAvroStoreCli
   private RecordDeserializer<StreamingFooterRecordV1> streamingFooterRecordDeserializer;
   private TransportClient transportClient;
   private final Executor deserializationExecutor;
+  private final Executor veniceClientWarmUpExecutor =
+      Executors.newFixedThreadPool(1, new DaemonThreadFactory("Venice-Client-Warmup"));
```
gaojieliu, line 79: So we will allocate a single threaded pool for store client, can you append the store name to the thread pool name prefix?

It is fine to have a pool per store client as the warmup logic is only needed at startup time.

haoxu07: Added store name.

haoxu07:
> Could you please update the commit message for why this is bad? What problem is this PR solving for us?

Added already.

clients/venice-thin-client/src/main/java/com/linkedin/venice/client/store/AbstractAvroStoreClient.java
```diff
@@ -714,7 +716,7 @@ private void warmUpVeniceClient() {
          * If the D2 client isn't retry in the async warm-up phase, it will be delayed to the first query.
          * Essentially, this is a best-effort.
          */
-        CompletableFuture.runAsync(this::getKeySerializerWithRetryWithLongInterval);
+        CompletableFuture.runAsync(this::getKeySerializerWithRetryWithLongInterval, veniceClientWarmUpExecutor);
```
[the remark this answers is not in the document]
haoxu07: Yeah, this 1 thread pool is only for warming up purpose for this client.

gaojieliu (approved):
Thanks!

PR #1479, opened 2025-01-28
[controller] Fix regression in ACL handling for non-secure admin server

sushantmane:
## Do not pass DynamicAccessController when using non-secure channel

VeniceController refactoring, where the DynamicAccessController (DAC) was inadvertently passed to the  
Admin Server even when SSL was disabled. The presence of a DAC was incorrectly interpreted as ACL  
checks being enabled, causing the Admin Server to enforce ACLs on both secure and non-secure  
channels. This happened because a valid DAC was applied universally to both secure and non-secure  
admin servers.  

The fix ensures that the DynamicAccessController is no longer passed to the Admin Server when using  
a non-secure channel. Additionally, when the NoOpDynamicAccessController is set, it is now explicitly  
treated as a signal that ACL checks are disabled for HTTP requests in the controller. These changes  
restore the expected behavior.  

## How was this PR tested?

CI


nisargthakkar (approved):
LGTM

PR #1481, opened 2025-01-28
[protocol] Add KafkaMessageEnvelope schema v12 to replace long offsets with PubSubPosition

sushantmane:
## Add PubSubPosition fields in KafkaMessageEnvelope to replace long offsets

This update introduces a new schema version: `v12` for `KafkaMessageEnvelope` to add fields aimed  
at eventually replacing long-based offsets with `PubSubPosition`. Specifically:
- **`localHighWatermarkPubSubPositions`** is introduced to eventually replace
  `localHighWatermarks`, which stores the latest consumed positions for all real-time topics.
- **`upstreamPubSubPosition`** is introduced to eventually replace `upstreamOffset`, which
  stores the message position in the upstream pubsub system.
  
Misc Addition:
- **`termId`** is introduced to track the unique term in which the message was produced,
  typically using a Helix message timestamp of the leader state transition.

  `termId` is intended to uniquely represent a leader session or term in the replication log.
  In scenarios like split-brain, where multiple writers may exist simultaneously, `termId` will
  help establish a partial order (happens-before relationship) on records. Each replica will
  maintain a `termId` and update it upon becoming the leader, marking the start of a new
  leadership term. The leader will then attach this `termId` to every record it writes to the
  version topic, allowing followers to distinguish between records from the legitimate leader
  and those from outdated or conflicting writers. This mechanism is expected to ensure that
  replicas correctly identify and process only valid records.

```bash
sumane@sumane-mn3406 venice % export VAR1=v11 VAR2=v12
diff internal/venice-common/src/main/resources/avro/KafkaMessageEnvelope/$VAR1/KafkaMessageEnvelope.avsc \
     internal/venice-common/src/main/resources/avro/KafkaMessageEnvelope/$VAR2/KafkaMessageEnvelope.avsc
302a303,310
>                       "name": "localHighWatermarkPubSubPositions",
>                       "doc": "The latest pubsub positions of all real-time topics consumed up until now.",
>                       "type": {
>                         "type": "array",
>                         "items": "bytes"
>                       },
>                       "default": []
>                     }, {
341a350,359
>               "default": -1
>             }, {
>               "name": "upstreamPubSubPosition",
>               "doc": "The position of the message in the upstream pubsub system (usually real-time topic).",
>               "type": "bytes",
>               "default": ""
>             }, {
>               "name": "termId",
>               "doc": "TermId is a unique identifier (usually a Helix message timestamp) for the term in which the message is produced.",
>               "type": "long",

```

## How was this PR tested?
CI

### DIFF
--- a/internal/venice-common/src/main/resources/avro/KafkaMessageEnvelope/v12/KafkaMessageEnvelope.avsc
+++ b/internal/venice-common/src/main/resources/avro/KafkaMessageEnvelope/v12/KafkaMessageEnvelope.avsc
@@ -1,0 +1,371 @@
+{
+  "name": "KafkaMessageEnvelope",
+  "namespace": "com.linkedin.venice.kafka.protocol",
+  "type": "record",
+  "fields": [
+    {
+      "name": "messageType",
+      "doc": "Using int because Avro Enums are not evolvable. Readers should always handle the 'unknown' value edge case, to account for future evolutions of this protocol. The mapping is the following: 0 => Put, 1 => Delete, 2 => ControlMessage, 3 => Update.",
+      "type": "int"
+    }, {
+      "name": "producerMetadata",
+      "doc": "ProducerMetadata contains information that the consumer can use to identify an upstream producer. This is common for all MessageType.",
+      "type": {
+        "name": "ProducerMetadata",
+        "type": "record",
+        "fields": [
+          {
+            "name": "producerGUID",
+            "doc": "A unique identifier for this producer.",
+            "type": {
+              "name": "GUID",
+              "type": "fixed",
+              "size": 16
+            }
+          }, {
+            "name": "segmentNumber",
+            "doc": "A number used to disambiguate between sequential segments sent into a given partition by a given producer. An incremented SegmentNumber should only be sent following an EndOfSegment control message. For finite streams (such as those bulk-loaded from Hadoop), it can be acceptable to have a single SegmentNumber per producer/partition combination, though that is not something that the downstream consumer should assume. For infinite streams, segments should be terminated and begun anew periodically. This number begins at 0.",
+            "type": "int"
+          }, {
+            "name": "messageSequenceNumber",
+            "doc": "A monotonically increasing number with no gaps used to distinguish unique messages produced in this segment (i.e.: by this producer into a given partition). This number begins at 0 (with a StartOfSegment ControlMessage) and subsequent messages (such as Put) will have a SequenceNumber of 1 and so forth.",
+            "type": "int"
+          }, {
+            "name": "messageTimestamp",
+            "doc": "The time of the producer's local system clock, at the time the message was submitted for production. This is the number of milliseconds from the unix epoch, 1 January 1970 00:00:00.000 UTC.",
+            "type": "long"
+          }, {
+            "name": "logicalTimestamp",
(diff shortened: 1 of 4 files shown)


FelixGV:
Thanks for submitting the schema evolution proposal. Can we do the following please?

1. Do not rev up the protocol number yet in `AvroProtocolDefinition`.
2. Use the `build.gradle` SR generation override to specify the version to code gen for (set it to the current latest version, prior to this PR).
3. Post in the PR description the diff between the new proposed version and the one before that.

Thanks!

majisourav99:
> Thanks for submitting the schema evolution proposal. Can we do the following please?
> 
> 1. Do not rev up the protocol number yet in `AvroProtocolDefinition`.
> 2. Use the `build.gradle` SR generation override to specify the version to code gen for (set it to the current latest version, prior to this PR).
> 3. Post in the PR description the diff between the new proposed version and the one before that.
> 
> Thanks!

We plan to reject PRs with both Java and schema file updates, ideally schema changes should be reviewed before code changes.

sushantmane:
@FelixGV @majisourav99  - Thanks for the review. Let me update the PR. I just discovered that if I don't do `./gradlew assemble` then the build fails.

internal/venice-common/src/main/resources/avro/KafkaMessageEnvelope/v12/KafkaMessageEnvelope.avsc
```diff
@@ -0,0 +1,371 @@
...
+              "name": "upstreamPubSubPosition",
+              "doc": "The position of the message in the upstream pubsub system (usually real-time topic).",
+              "type": ["null", "bytes"],
+              "default": null
+            }, {
+              "name": "termId",
+              "doc": "TermId is a unique identifier (usually a Helix message timestamp) for the term in which the message is produced.",
+              "type": "long",
+              "default": -1
+            }
```
FelixGV, line 364: What is this `termId` for? Is it related to pub sub positions, or some separate thing? Is it related to improving leadership handovers?

FelixGV (commented):
Left a question. Thanks.

internal/venice-common/src/main/resources/avro/KafkaMessageEnvelope/v12/KafkaMessageEnvelope.avsc
sushantmane: Yeah, it's is a separate thing intended to improve the leadership handover. Though this is not funded project, I thought since we're doing proto upgrade for KME it might be good idea to start adopting it as we are also going to do dataset refreshes later.

sushantmane: More context around `termId`:

> `termId` is intended to uniquely represent a leader session or term in the replication log. In scenarios like **split-brain**, where multiple writers may exist simultaneously, `termId` will help establish **a partial order (happens-before relationship)** on records. Each replica will maintain a termId and update it upon becoming the leader, marking the start of a new leadership term. The leader will then attach this termId to every record it writes to the version topic, allowing followers to distinguish between records from the legitimate leader and those from outdated or conflicting writers. This mechanism is expected to ensure that replicas correctly identify and process only valid records.

FelixGV (approved):
Thanks Sushant! LGTM!

sushantmane:
Thanks a lot, @FelixGV!

PR #1482, opened 2025-01-29
[protocol] Add PartitionState schema v16 to replace long offset with PubSubPosition

sushantmane:
## Add PubSubPosition fields in PartitionState to replace long offsets

This change introduces fields in `PartitionState.avsc` to replace long 
offsets with PubSub positions. The new fields include:

- `lastProcessedVersionTopicPubSubPosition`: Tracks the last successfully consumed 
  position in the version topic.
- `latestObservedTermId`: Stores the most recent termId observed by this replica.
- `currentTermStartPubSubPosition`: Marks the PubSub position where the current 
  leader's term starts in the version topic.
- `upstreamRealTimeTopicPubSubPositionMap`: A map tracking the last PubSub position 
  consumed from upstream topics, keyed by the bootstrap server URL.
- `upstreamVersionTopicPubSubPosition`: Stores the last persisted upstream version 
  topic PubSub position; remains `null` if the batch native-replication source is the 
  same as the local region.

```bash
sumane@sumane-mn3406 venice %  export VAR1=v15 VAR2=v16 && diff internal/venice-common/src/main/resources/avro/PartitionState/$VAR1/PartitionState.avsc internal/venice-common/src/main/resources/avro/PartitionState/$VAR2/PartitionState.avsc

11a12,26
>       "name": "lastProcessedVersionTopicPubSubPosition",
>       "doc": "The last PubSub position consumed successfully in this partition from version topic",
>       "type": "bytes",
>       "default": ""
>     }, {
>       "name": "latestObservedTermId",
>       "doc": "The most recent termId observed by this replica.",
>       "type": "long",
>       "default": -1
>     }, {
>       "name": "currentTermStartPubSubPosition",
>       "doc": "The pubsub position marking the start of the current leader's term in the version topic.",
>       "type": "bytes",
>       "default": ""
>     }, {
80a96,105
>       "name": "upstreamRealTimeTopicPubSubPositionMap",
>       "type": {
>         "type": "map",
>         "values": "bytes",
>         "java-key-class": "java.lang.String",
>         "avro.java.string": "String"
>       },
>       "doc": "A map of upstream PubSub bootstrap server url -> the last PubSub position consumed from upstream topic.",
>       "default": {}
>     }, {
85a111,115
>       "name": "upstreamVersionTopicPubSubPosition",
>       "type": "bytes",
>       "doc": "The last upstream version topic PubSub position persisted to disk; if the batch native-replication source is the same as local region, this value will always be null",
>       "default": ""
>     }, {

```

## How was this PR tested?

CI

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -314,6 +314,7 @@ subprojects {
     doFirst {
       def versionOverrides = [
 //        project(':internal:venice-common').file('src/main/resources/avro/StoreVersionState/v5', PathValidation.DIRECTORY)
+        project(':internal:venice-common').file('src/main/resources/avro/PartitionState/v15', PathValidation.DIRECTORY)
       ]
 
       def schemaDirs = [sourceDir]
--- a/internal/venice-common/src/main/resources/avro/PartitionState/v16/PartitionState.avsc
+++ b/internal/venice-common/src/main/resources/avro/PartitionState/v16/PartitionState.avsc
@@ -1,0 +1,230 @@
+{
+  "name": "PartitionState",
+  "namespace": "com.linkedin.venice.kafka.protocol.state",
+  "doc": "This record holds the state necessary for a consumer to checkpoint its progress when consuming a Venice partition. When provided the state in this record, a consumer should thus be able to resume consuming midway through a stream.",
+  "type": "record",
+  "fields": [
+    {
+      "name": "offset",
+      "doc": "The last Kafka offset consumed successfully in this partition from version topic.",
+      "type": "long"
+    }, {
+      "name": "lastProcessedVersionTopicPubSubPosition",
+      "doc": "The last PubSub position consumed successfully in this partition from version topic",
+      "type": "bytes",
+      "default": ""
+    }, {
+      "name": "latestObservedTermId",
+      "doc": "The most recent termId observed by this replica.",
+      "type": "long",
+      "default": -1
+    }, {
+      "name": "currentTermStartPubSubPosition",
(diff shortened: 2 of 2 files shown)


ZacAttack:
I think overall this looks good, but I think I want to visit the usage of map.  Can you start a thread in slack and tag myself, Gaojie and Felix?  I remember we had this conversation of indexed list vs. map before but I can't really recall what the original concerns against map were.

sushantmane:
> I think overall this looks good, but I think I want to visit the usage of map. Can you start a thread in slack and tag myself, Gaojie and Felix? I remember we had this conversation of indexed list vs. map before but I can't really recall what the original concerns against map were.





Just to clarify: In this pr `upstreamRealTimeTopicPubSubPositionMap` is of type `map` because the existing field with numeric offset is map as well

```avro
{
      "name": "upstreamOffsetMap",
      "type": {
        "type": "map",
        "values": "long",
        "avro.java.string": "String"
      },
      "doc": "A map of upstream Kafka bootstrap server url -> the last Kafka offset consumed from upstream topic.",
      "default": {}
    },
```

ZacAttack:
upstreamRealTimeTopicPubSubPositionMap: A map tracking the last PubSub position
consumed from upstream topics, keyed by the bootstrap server URL.

I think we got to use brokerId as our key here?

ZacAttack (approved):

sushantmane:
Thanks a lot, @ZacAttack!

PR #1484, opened 2025-01-30
[controller] update real time topic name with UpdateStore admin command

arjun4084346:
## Set real time topic name via `UpdateStore` admin tool command

So far, real time topic name could only be changed during partition count update (https://github.com/linkedin/venice/pull/1404).
In this PR, I am making it changeable directly with update store command. The benefits of this change will that users will be able to delete and recreate stores with a different real time topic name. If real time topic name is not changed, it is observed that the old real time topic, with the same name, sometimes, hang in there longer and does not let the new store create.
 
Resolves #XXX

## How was this PR tested?

added an integration test


ZacAttack (approved):
Looks good, but check on the GA failure?

PR #1492, opened 2025-01-31
[build] Configure dependabot to automatically update duckdb snapshots

kvargha:
## [build] Configure dependabot to automatically update duckdb snapshots

This is necessary because venice-duckdb depends on a snapshot version of duckdb which keeps getting deleted. This PR intends to automatically check/update the snapshot version for duckdb.

## How was this PR tested?

CI


sushantmane (approved):
LGTM. Thanks 🚀

PR #1493, opened 2025-01-31
Fix PR limit workflow to check for schema files limits

majisourav99:
## Fix PR limit workflow to check for different schema files limits

Fix PR limit workflow to check for different schema files limits and total line changes in a PR. Following are the limits a PR must follow.

 MAX_LINES per file is 500 
 MAX_TOTAL_LINES_ADDED is 2000

Also schema changes should be reviewed separately in a different PR before doing any code changes.
 There's an option in build.gradle to isolate schema changes, allowing them to be integrated without affecting the existing code picking up the new generated class.
    def versionOverrides = [
//        project(':internal:venice-common').file('src/main/resources/avro/StoreVersionState/v5', PathValidation.DIRECTORY)
      ]

## How was this PR tested?

CI

### DIFF
--- a/.github/workflows/limit-pr-files-changed.yml
+++ b/.github/workflows/limit-pr-files-changed.yml
@@ -78,6 +78,7 @@ jobs:
           # Fail if there are both schema and Java file changes
           if [[ "JAVA_FILE" == true && "SCHEMA_FILE" == true ]]; then
             echo "The PR has both schema and Java code changes, please make a separate PR for schema changes."
+            echo "For schema file changes, please update build.gradle to update 'versionOverrides' inside compileAvro task to use fixed protocol version"
             exit 1
           fi
           
--- a/.github/workflows/limit-pr-files-changed.yml
+++ b/.github/workflows/limit-pr-files-changed.yml
@@ -1,0 +1,110 @@
+name: Enforce Max Lines Changed Per File
+
+on:
+  pull_request:
+    paths:
+      - '**/*.java'   # Monitor changes to Java files
+      - '**/*.avsc'   # Monitor changes to Avro schema files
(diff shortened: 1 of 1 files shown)


FelixGV:
Can we not have a line limit on schemas? But rather the constraint we talked about where if schemas are changed then code should not be?

majisourav99:
> Can we not have a line limit on schemas? But rather the constraint we talked about where if schemas are changed then code should not be?

Sure, its hard to refactor schema in multiple classes.

huangminchn:
> Can we not have a line limit on schemas? But rather the constraint we talked about where if schemas are changed then code should not be?

@FelixGV Turns out that it's not straightforward. In the "compileAvro" task inside build.gradle, we would always generate avro classes based on the last avro schema file in the resources/avro directory. The fix we could think of so far is to look for the current protocol version number inside Java class "AvroProtocolDefinition", and then build a mapping from the ENUM name in "AvroProtocolDefinition" to the avro file paths somewhere.
We were concern about the extra maintenance overhead of the mapping; please don't hesitate to share your thoughts on this.

huangminchn:
> > name in "AvroProtocolDefinition" to the avro file paths somewhere.
> > We were concern about the extra maintenance overhead of the mapping; please don't hesitate to share your thoughts on this.
> 
> @huangminchn There is a option to override that APD number. Please check def versionOverrides = [ // project(':internal:venice-common').file('src/main/resources/avro/StoreVersionState/v5', PathValidation.DIRECTORY) ]

Right, but then this is the extra mapping; we need to decide whether it's worth maintaining it. It's not a lot of work to change an additional number; but we should put comments here as well as APD file to remind devs.

huangminchn (approved):

PR #1494, opened 2025-02-03
[controller] Set default for `is_abort_migration_cleanup` to false when null

ranwang2024:
## [controller] Set default for `is_abort_migration_cleanup` to false when null

Handle cases where `is_abort_migration_cleanup` is null by setting it to false. This ensures that the system remains backward compatible even when the value is not explicitly provided.

## How was this PR tested?

Tested by adding comprehensive unit tests that cover the new functionality and ensure its correctness.

### DIFF
--- a/internal/venice-common/src/main/java/com/linkedin/venice/utils/Utils.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/utils/Utils.java
@@ -304,27 +304,6 @@ public class Utils {
   }
 
   /**
-   * Since {@link Boolean#parseBoolean(String)} does not throw exception and will always return 'false' for
-   * any string that are not equal to 'true', We validate the string by our own.
-   */
-  public static boolean parseBooleanFromString(String value, String fieldName) {
-    if (value == null) {
-      throw new VeniceHttpException(
-          HttpStatus.SC_BAD_REQUEST,
-          fieldName + " must be a boolean, but value is null",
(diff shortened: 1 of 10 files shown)


internal/venice-common/src/main/java/com/linkedin/venice/utils/Utils.java
sushantmane, line 1: Can we delte old `parseBooleanFromString` method and replace it usage with `parseBooleanOrThrow`?

ranwang2024: Can I make that change in another PR? There are about 19 places using the `parseBooleanFromString` method, and I feel it's a bit risky to make that change in the PR that addresses the null exception.

sushantmane: The behavior of `parseBooleanOrThrow` and `parseBooleanFromString` is exactly the same. So it should be safe to replace `parseBooleanFromString` with  `parseBooleanOrThrow` and delete `parseBooleanFromString`

ranwang2024: Got it. I updated the PR to replace `parseBooleanFromString` with `parseBooleanOrThrow`.

sushantmane (approved):
LGTM. Thanks, @ranwang2024!

PR #1497, opened 2025-02-03
[changelog] Make changes for venice view consumption

ZacAttack:
[tests incoming]

These changes make it so that the venice after image consumer is able to now consume view topics. View topics internally very much resemble version topics, which the after image consumer understands today.  So the logic is largely the same.  The only difference we add here is that we now have to consult store repositories which are view aware, and we have to chnage how we maintain local highwatermark information.  Highwatermarks are meant to be compared on RT partitions, and new view types map 1:N and N:N RT partitions to view partitions.  This means we need to add an additional dimensionality to highwatermark filtering where instead of keeping a single map of topic partitions to offsets, but a map of topic partitions to maps of RT partitions to highwatermarks.

There's still some pending work here, we need to figure out chunk assembly (as the current implementation of chunk assembly doesn't account for interleaving writes) and we need to put a bow on how we articulate to the consumers what the upstream RT partition is (right now we're just hardcoding the 1:1 pairing).

Resolves #XXX

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/InternalLocalBootstrappingVeniceChangelogConsumer.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/InternalLocalBootstrappingVeniceChangelogConsumer.java
@@ -94,7 +94,8 @@ class InternalLocalBootstrappingVeniceChangelogConsumer<K, V> extends VeniceAfte
     bootstrapStateMap = new VeniceConcurrentHashMap<>();
     syncBytesInterval = changelogClientConfig.getDatabaseSyncBytesInterval();
     metricsRepository = changelogClientConfig.getInnerClientConfig().getMetricsRepository();
-    String localStateTopicNameTemp = changelogClientConfig.getStoreName() + LOCAL_STATE_TOPIC_SUFFIX;
+    String viewNamePath = changelogClientConfig.getViewName() == null ? "" : "- " + changelogClientConfig.getViewName();
+    String localStateTopicNameTemp = changelogClientConfig.getStoreName() + viewNamePath + LOCAL_STATE_TOPIC_SUFFIX;
     String bootstrapFileSystemPath = changelogClientConfig.getBootstrapFileSystemPath();
     if (StringUtils.isNotEmpty(consumerId)) {
       localStateTopicNameTemp += "-" + consumerId;
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/InternalLocalBootstrappingVeniceChangelogConsumer.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/InternalLocalBootstrappingVeniceChangelogConsumer.java
@@ -94,7 +94,7 @@ class InternalLocalBootstrappingVeniceChangelogConsumer<K, V> extends VeniceAfte
     bootstrapStateMap = new VeniceConcurrentHashMap<>();
     syncBytesInterval = changelogClientConfig.getDatabaseSyncBytesInterval();
     metricsRepository = changelogClientConfig.getInnerClientConfig().getMetricsRepository();
-    String viewNamePath = changelogClientConfig.getViewName() == null ? "" : "- " + changelogClientConfig.getViewName();
+    String viewNamePath = changelogClientConfig.getViewName() == null ? "" : "-" + changelogClientConfig.getViewName();
     String localStateTopicNameTemp = changelogClientConfig.getStoreName() + viewNamePath + LOCAL_STATE_TOPIC_SUFFIX;
     String bootstrapFileSystemPath = changelogClientConfig.getBootstrapFileSystemPath();
     if (StringUtils.isNotEmpty(consumerId)) {
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/InternalLocalBootstrappingVeniceChangelogConsumer.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/InternalLocalBootstrappingVeniceChangelogConsumer.java
@@ -508,10 +508,9 @@ class InternalLocalBootstrappingVeniceChangelogConsumer<K, V> extends VeniceAfte
 
     storageService.start();
     try {
-      // storeRepository.start();
       storeRepository.subscribe(storeName);
     } catch (InterruptedException e) {
-      throw new RuntimeException(e);
+      throw new VeniceException("Failed to start bootstrapping changelog consumer with error:", e);
     }
 
     return seekWithBootStrap(partitions);
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/InternalLocalBootstrappingVeniceChangelogConsumer.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/InternalLocalBootstrappingVeniceChangelogConsumer.java
@@ -183,7 +183,8 @@ class InternalLocalBootstrappingVeniceChangelogConsumer<K, V> extends VeniceAfte
   protected boolean handleVersionSwapControlMessage(
       ControlMessage controlMessage,
       PubSubTopicPartition pubSubTopicPartition,
-      String topicSuffix) {
+      String topicSuffix,
+      Integer upstreamPartition) {
     ControlMessageType controlMessageType = ControlMessageType.valueOf(controlMessage);
     if (controlMessageType.equals(ControlMessageType.VERSION_SWAP)) {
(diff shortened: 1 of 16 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/InternalLocalBootstrappingVeniceChangelogConsumer.java
```diff
@@ -518,7 +520,13 @@ public CompletableFuture<Void> start(Set<Integer> partitions) {
...
   public CompletableFuture<Void> start() {
     Set<Integer> allPartitions = new HashSet<>();
-    for (int partition = 0; partition < partitionCount; partition++) {
+    try {
+      storeRepository.subscribe(storeName);
+    } catch (InterruptedException e) {
+      throw new VeniceException("Failed to start bootstrapping changelog consumer with error:", e);
+    }
+    Store store = storeRepository.getStore(storeName);
+    for (int partition = 0; partition < store.getPartitionCount(); partition++) {
```
xunyin8, line 529: `ReadOnlyViewStore` did not override the `getPartitionCount()` method so `store.getPartitionCount()` will return the original store's partition count even for views. Is this the desired behavior?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumerClientFactory.java
```diff
@@ -78,6 +78,9 @@ public <K, V> VeniceChangelogConsumer<K, V> getChangelogConsumer(String storeNam
       String viewClass = getViewClass(newStoreChangelogClientConfig, storeName);
       String consumerName = suffixConsumerIdToStore(storeName + "-" + viewClass.getClass().getSimpleName(), consumerId);
       if (viewClass.equals(ChangeCaptureView.class.getCanonicalName())) {
+        // TODO: This is a little bit of a back. This is to deal with the an issue where the before image change
```
xunyin8, line 81: NIT: `hack` instead of `back`?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumerImpl.java
```diff
@@ -182,16 +199,13 @@ public VeniceChangelogConsumerImpl(
...
-        storeName,
-        partitionCount,
-        viewClassName);
+    LOGGER.info("Start a change log consumer client for store: {}", storeName);
   }
 
   @Override
   public int getPartitionCount() {
-    return partitionCount;
+    Store store = storeRepository.getStore(storeName);
```
xunyin8, line 207: Similar here, this will return the original store's partition count in the case of a view store.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumerImpl.java
```diff
@@ -687,7 +702,14 @@ protected boolean handleControlMessage(
...
     if (controlMessageType.equals(ControlMessageType.VERSION_SWAP)) {
-      return handleVersionSwapControlMessage(controlMessage, pubSubTopicPartition, topicSuffix);
+      // TODO: In view topics, we need to know the partition of the upstream RT
+      // how we transmit this information has yet to be determined, so once we finalize
+      // that, we'll need to tweak this. For now, we'll just pass in the same partition number
+      return handleVersionSwapControlMessage(
+          controlMessage,
+          pubSubTopicPartition,
+          topicSuffix,
+          pubSubTopicPartition.getPartitionNumber());
```
xunyin8, line 712: Could you elaborate a bit how will this work in this PR? i.e. without the proper upstream RT partition. Since we don't know the mapping shouldn't we set the `currentVersionHighWatermarks` with the lowest watermark from version swap? Otherwise wouldn't we filter out messages unintentionally?

xunyin8 (commented):
Mostly questions since I'm still trying to understand the expected behavior of message filtering 🤓

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/InternalLocalBootstrappingVeniceChangelogConsumer.java
ZacAttack: That is not the desired behavior.  I guess I was hoping the abstraction would return the view partition count.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumerImpl.java
ZacAttack: Unless we know the upstream RT partition no offset highwatermark is actually comparable.  So I don't think keeping the 'lowest' is the right answer.  Really without knowing the source RT partition all answers are wrong

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumerClientFactory.java
ZacAttack: ah tuck

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/consumer/TestChangelogConsumer.java
```diff
@@ -102,7 +102,7 @@
 
 
 public class TestChangelogConsumer {
```
xunyin8, line 104: Shall we add a CC test with a materialized view either here or in `TestMaterializedViewEndToEnd` to do some simple verification? Like hybrid writes can be captured in the expected view partitions.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumerImpl.java
```diff
@@ -135,7 +138,22 @@ public VeniceChangelogConsumerImpl(
...
+    // they only contain nearline data whereas views are full version topics. So for now it seems legit to have this
+    // caveat, but we should perhaps think through this in the future if we think we'll have more nearline data only
+    // cases for view topics. Change capture topic naming also doesn't follow the view_store_store_name_view_name
+    // naming pattern, making tracking it even weirder.
+    if (changelogClientConfig.getViewName() == null || changelogClientConfig.getViewName().isEmpty()
+        || changelogClientConfig.isBeforeImageView()) {
+      this.storeName = changelogClientConfig.getStoreName();
+    } else {
+      this.storeName =
+          ChangeCaptureView.getViewStoreName(changelogClientConfig.getStoreName(), changelogClientConfig.getViewName());
```
xunyin8, line 154: Does it matter that `localStateTopicNameTemp` and `localStateTopicName` in `InternalLocalBootstrappingVeniceChangelogConsumer` will still use the corresponding store name `changelogClientConfig.getStoreName()` even when view is specified?

xunyin8 (commented):
Just a question about usages of `changelogClientConfig.getStoreName()` in `InternalLocalBootstrappingVeniceChangelogConsumer` when view is specified.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumerImpl.java
ZacAttack: yeah.... admittedly it looks sketchy, but, it's just for naming of filepaths.  I don't see an obvious bug excluding it, HOWEVER, there is potential issue if you ever had local state and you changed your view source.  In that case, we should probably pad the view name if only to make certain that state doesn't drag over between runs.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/consumer/TestChangelogConsumer.java
ZacAttack: ehhhh.  Unfortunately until we do the source partition thing this isn't really doable.  The MaterializedView enforces us to have a different partition count or different partitioner.  This means the record count is gonna be wonky because we'll filter messages incorrectly.  So I'd rather wait until we square that part away.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumerImpl.java
xunyin8: Do you foresee users using CC for both the store and its views in the same instance/host? The same problem could apply right?

xunyin8 (approved):

PR #1498, opened 2025-02-04
[fast-client] Fixed retry budget freshness issue

gaojieliu:
This PR fixed a bug in RetryBudget, and previously, if the request volume becomes 0, the retry budget won't be refreshed any more even there are more traffic later on.

This PR also re-enables retry budget by default.

## How was this PR tested?
CI


xunyin8 (approved):
Good catch Gaojie, so the new behavior is now if we encounter a period without KPS we will continue to use the retry budget calculated in the most recent period with KPS.

PR #1499, opened 2025-02-05
[controller] Skip checking RT for deleted stores

majisourav99:
## Skip checking RT for deleted stores

When deleting stores, we check for hybridness in child colos, if one of the colo has already deleted the store, the check returns false, which can block RT deletion forever.
This PR fixes such cases.

## How was this PR tested?

GHCI

### DIFF
--- a/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
@@ -3895,33 +3895,42 @@ public class VeniceHelixAdmin implements Admin, StoreCleaner {
     String rtTopicName = Utils.composeRealTimeTopic(storeName);
     Map<String, ControllerClient> controllerClientMap = getControllerClientMap(clusterName);
     for (Map.Entry<String, ControllerClient> controllerClientEntry: controllerClientMap.entrySet()) {
-      StoreResponse storeResponse = RetryUtils.executeWithMaxAttemptAndExponentialBackoff(() -> {
-        StoreResponse response = controllerClientEntry.getValue().getStore(storeName);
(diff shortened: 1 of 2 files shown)


services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -3896,6 +3896,13 @@ public boolean isRTTopicDeletionPermittedByAllControllers(String clusterName, St
     for (Map.Entry<String, ControllerClient> controllerClientEntry: controllerClientMap.entrySet()) {
       StoreResponse storeResponse = controllerClientEntry.getValue().getStore(storeName);
       if (storeResponse.isError()) {
+        if (storeResponse.getError().contains("does not exist")) {
```
sushantmane, line 3899: Can we make `does not exist` constant and use it here as well as where this error is sent? 

Also, should we add few retries for get store?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -3894,15 +3895,42 @@ public boolean isRTTopicDeletionPermittedByAllControllers(String clusterName, St
...
+              storeName,
+              clusterName,
+              controllerClientEntry.getKey(),
+              storeResponse.getError());
+          return false;
+        }
+      } catch (VeniceNoStoreException e) {
         LOGGER.warn(
-            "Skipping RT cleanup check for store: {} in cluster: {} due to unable to get store from fabric: {} Error: {}",
+            "Store {} does not exist in fabric {}, probably deleted already, skipping RT check",
```
sushantmane, line 3930: Can we log clusterName as well?

sushantmane (approved):
Thanks for the quick fix, @majisourav99! 🚀

PR #1500, opened 2025-02-05
[da-vinci][server] Fixed the race condition during BlobDB rolling out/back

gaojieliu:
Because of the blob-db integration, `SIT` will forcibly set `sorted` to `false` for hybrid stores and inconsistent `sorted` can happen during the rolling out/rolling back blob-db features. Here is one case how it can happen during the rolling out of blob-db:
1. P1 processes `SOP` with `sorted=true` and persist it in `StoreVersionState`.
2. P2 hasn't started processing `SOP` yet.
3. Restart the cluster and roll out blob-db feature.
4. when P2 processes `SOP` with `sorted=true`, it will forcibly set `sorted=false`, and it will be different from the previously persisted `StoreVersionState`.
5. This is fine as long as we just follow the previously persisted `StoreVersionState`.
6. Here are the reasons why step 5 is true: a. If the input for a given partition is truly sorted, it can always be ingested as unsorted data. b. If the input for a given partition is not sorted, the underlying `SSTFileWriter` will throw exception when we try to ingest it as sorted data.

## How was this PR tested?
CI

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
@@ -2983,7 +2983,9 @@ public abstract class StoreIngestionTask implements Runnable, Closeable {
               // Something very wrong is going on ): ...
               throw new VeniceException(
                   "Unexpected: received multiple " + ControlMessageType.START_OF_PUSH.name()
-                      + " control messages with inconsistent 'sorted' fields within the same topic!");
+                      + " control messages with inconsistent 'sorted' fields within the same topic!"
+                      + " And persisted sorted value: " + previousStoreVersionState.sorted
+                      + " is different from the current one: " + sorted);
             }
             /**
              * Because of the blob-db integration, `SIT` will forcibly set `sorted` to `false` for hybrid stores (check the
@@ -3000,7 +3002,7 @@ public abstract class StoreIngestionTask implements Runnable, Closeable {
              *    b. If the input for a given partition is not sorted, the underlying `SSTFileWriter` will throw exception
              *       when we try to ingest it as sorted data.
              */
-            LOGGER.info(
+            LOGGER.warn(
                 "Store version state for topic {} has already been initialized with a different value of 'sorted': {}",
                 kafkaVersionTopic,
                 previousStoreVersionState.sorted);
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
@@ -2948,47 +2948,70 @@ public abstract class StoreIngestionTask implements Runnable, Closeable {
     } else {
       sorted = startOfPush.sorted;
     }
-    beginBatchWrite(partition, sorted, partitionConsumptionState);
-    partitionConsumptionState.setStartOfPushTimestamp(startOfPushKME.producerMetadata.messageTimestamp);
+
+    StoreVersionState persistedStoreVersionState =
+        storageMetadataService.computeStoreVersionState(kafkaVersionTopic, previousStoreVersionState -> {
+          if (previousStoreVersionState == null) {
+            // No other partition of the same topic has started yet, let's initialize the StoreVersionState
+            StoreVersionState newStoreVersionState = new StoreVersionState();
+            newStoreVersionState.sorted = sorted;
+            newStoreVersionState.chunked = startOfPush.chunked;
+            newStoreVersionState.compressionStrategy = startOfPush.compressionStrategy;
+            newStoreVersionState.compressionDictionary = startOfPush.compressionDictionary;
+            if (startOfPush.compressionStrategy == CompressionStrategy.ZSTD_WITH_DICT.getValue()) {
+              if (startOfPush.compressionDictionary == null) {
+                throw new VeniceException(
+                    "compression Dictionary should not be empty if CompressionStrategy is ZSTD_WITH_DICT");
+              }
+            }
+            newStoreVersionState.batchConflictResolutionPolicy = startOfPush.timestampPolicy;
+            newStoreVersionState.startOfPushTimestamp = startOfPushKME.producerMetadata.messageTimestamp;
+
+            LOGGER.info(
+                "Persisted {} for the first time following a SOP for topic {} with sorted: {}.",
+                StoreVersionState.class.getSimpleName(),
+                kafkaVersionTopic,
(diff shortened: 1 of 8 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -2948,47 +2948,70 @@ private void processStartOfPush(
...
+          } else if (previousStoreVersionState.chunked != startOfPush.chunked) {
+            // Something very wrong is going on ): ...
             throw new VeniceException(
-                "compression Dictionary should not be empty if CompressionStrategy is ZSTD_WITH_DICT");
+                "Unexpected: received multiple " + ControlMessageType.START_OF_PUSH.name()
+                    + " control messages with inconsistent 'chunked' fields within the same topic!");
+          } else if (previousStoreVersionState.sorted != sorted) {
+            if (!isHybridMode()) {
+              // Something very wrong is going on ): ...
+              throw new VeniceException(
```
majisourav99, line 2984: nit: can we add the previous and current sorted value?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -2948,47 +2948,70 @@ private void processStartOfPush(
...
+             * 3. Restart the cluster and roll out blob-db feature.
+             * 4. when P2 processes `SOP` with `sorted=true`, it will forcibly set `sorted=false`, and it will be different
+             *    from the previously persisted `StoreVersionState`.
+             * 5. This is fine as long as we just follow the previously persisted `StoreVersionState`.
+             * 6. Here are the reasons why step 5 is true:
+             *    a. If the input for a given partition is truly sorted, it can always be ingested as unsorted data.
+             *    b. If the input for a given partition is not sorted, the underlying `SSTFileWriter` will throw exception
+             *       when we try to ingest it as sorted data.
+             */
+            LOGGER.info(
```
majisourav99, line 3003: should be a warn level log?

gaojieliu: Well, in the above scenario, it is kind of expected, but it is not expected all the time.
I don't mind to change it to warn level.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -2948,47 +2948,70 @@ private void processStartOfPush(
...
+                "Store version state for topic {} has already been initialized with a different value of 'sorted': {}",
+                kafkaVersionTopic,
+                previousStoreVersionState.sorted);
           }
-        }
-        newStoreVersionState.batchConflictResolutionPolicy = startOfPush.timestampPolicy;
-        newStoreVersionState.startOfPushTimestamp = startOfPushKME.producerMetadata.messageTimestamp;
+          // No need to mutate it, so we return it as is
+          return previousStoreVersionState;
```
majisourav99, line 3009: should we check it only blobDB enabled we allow this ? and after full rollout of blobdb if still there is mismatch we fail?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
gaojieliu: Sure

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
gaojieliu: No, we shouldn't as blobdb can be rolled back and rolling back can trigger the same situation.

We have batch-only store to detect such behavior if it does happen, but hybrid store, sometimes, it is hard to determine which is right/wrong.
Just check the javadoc.

majisourav99: So we do not plan to fail ever once we fully rollout blob-db ?

gaojieliu: Until we don't have a plan to rollback blob db, such as removing the config option and blob db will always be on, also we need to make sure all the existing store versions finish the conversion, which is a fairly challenging assumption.

sixpluszero (approved):

FelixGV:
Thanks for having produced this hot fix quickly after discovering the issue.

Now that we are out of hot water, I would like to explore whether there might be a better way of fixing this, however.

Up until recently, the `SVS` could be expected to deterministically contain whatever came in from the various broadcast control messages (getting initially bootstrapped from a `SOP`, and then getting updated upon receiving the `EOP` and `TS` CMs). This meant that any process consuming any of the partitions should be expected to arrive at the same `SVS` (assuming they have seen the same CMs, e.g. a consumer which has only seen the `SOP` will have a less fresh `SVS` as compared to one which has consumed all the way until the `EOP`, but that is just a matter of freshness, not of permanent divergence).

Since the blob db changes were done (the first of which I reviewed myself, so I acknowledge that I had not thought of all this yet, back then), this is now different. The configs of the consuming process can now alter the state of the `SVS`. While this can be made to still work, as demonstrated in the current PR, I think it makes the system more complicated to reason about. Essentially, with this change, the `SVS` no longer reflects the content of the `SOP`, but rather the `SOP` modified by whatever configs were present when the first `SOP` was consumed...

I think it would be better to go back to the way it was before, where the `SVS` is a deterministic representation of whatever the `SOP` is. But, let the `RocksDBStoragePartition` decide whether the ignore the `sorted` flag or not. As a matter of fact, we already have exactly this kind of logic today in the `RocksDBStoragePartition` (see the [constructor](https://github.com/linkedin/venice/blob/28b29890611e116cb7a8dd88719f579d4ffb8f07/clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBStoragePartition.java#L190)).

This I think would be cleaner for several reasons:

1. The `SIT` should be agnostic of storage engine configs. The right place to deal with RocksDB-related behavior is in the RocksDB storage engine classes. This is important as we are trying to make DVRT work well with non-RocksDB engines as well.
2. All of the logic related to whether the `RocksDBSstFileWriter` should be used or not, ought to be located in the same place.
3. The `SVS` being generated deterministically no matter what the consumer settings are would help make it easier to implement certain improvements to the checkpointing logic.

Please LMK what you think, and let's see if we could iterate on this.

Thanks!

PR #1501, opened 2025-02-05
[dvc][doc] Refactor DaVinciRecordTransformerConfig to use Builder pattern and update DVRT docs

kvargha:
## [dvc][doc] Refactor DaVinciRecordTransformerConfig to use Builder pattern and update DVRT docs

1.  Previously, the configs for `DaVinciRecordTransformerConfig` were set by passing in all of the values into its constructor. This means that anytime we wanted to add another config, it would be a breaking change. To resolve this, I refactored `DaVinciRecordTransformerConfig` to use a Builder pattern.
2. Fixed a bug where if a user set `storeRecordsInDaVinci`, it would always bootstrap from VT and ignore the on-disk state. This isn't deseriable behavior. To fix this, I introduced a new config called `alwaysBootstrapFromVersionTopic`. Although most users won't use this, this can serve use-cases where the user is only storing their records in-memory. It is more preferable for users to at least have `storeRecordsInDaVinci` set to true so they can bootstrap from disk and utilize blob transfer.
3. If a user doesn't intend on modifying the schema of their records, the user won't need to pass in`outputValueSchema` and `outputValueClass` to the `DaVinciRecordTransformerConfig` anymore. This can serve use cases where the user just wants to redirect their records to an alternative storage such as DuckDB.
4. Updated the DVRT docs to include the latest changes.

## How was this PR tested?

Added unit and integration tests

This PR modifies how to instantiate a DVRT, but since this isn't in-use at the moment it doesn't break anyone's changes.

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/AvroGenericDaVinciClient.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/AvroGenericDaVinciClient.java
@@ -700,6 +700,12 @@ public class AvroGenericDaVinciClient<K, V> implements DaVinciClient<K, V>, Avro
       kafkaBootstrapServers = backendConfig.getString(KAFKA_BOOTSTRAP_SERVERS);
     }
 
+    String recordTransformerOutputValueSchema = "null";
+
+    if (daVinciConfig.isRecordTransformerEnabled() && recordTransformerConfig.getOutputValueSchema() != null) {
+      recordTransformerOutputValueSchema = recordTransformerConfig.getOutputValueSchema().toString();
+    }
+
     VeniceProperties config = new PropertyBuilder().put(KAFKA_ADMIN_CLASS, ApacheKafkaAdminAdapter.class.getName())
         .put(ROCKSDB_LEVEL0_FILE_NUM_COMPACTION_TRIGGER, 4) // RocksDB default config
         .put(ROCKSDB_LEVEL0_SLOWDOWN_WRITES_TRIGGER, 20) // RocksDB default config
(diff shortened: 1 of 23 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/AvroGenericDaVinciClient.java
```diff
@@ -700,6 +700,11 @@ private VeniceConfigLoader buildVeniceConfig() {
       kafkaBootstrapServers = backendConfig.getString(KAFKA_BOOTSTRAP_SERVERS);
     }
 
+    String recordTransformerOutputValueSchema = "null";
+    if (daVinciConfig.isRecordTransformerEnabled() && recordTransformerConfig.getOutputValueSchema() != null) {
+      recordTransformerOutputValueSchema = recordTransformerConfig.getOutputValueSchema().toString();
```
ZacAttack, line 705: nit: This is more personal taste, but:

recordTransformerOutputValueSchema = Objects.toString(recordTransformerConfig.getOutputValueSchema(), null);

This just ties the intention into the line as opposed to expanding the above if statement.

You can optionally take this, it's just a personal style preference.

ZacAttack (dismissed):
Looks ok overall despite highlighted nit

clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/AvroGenericDaVinciClient.java
```diff
@@ -700,6 +701,10 @@ private VeniceConfigLoader buildVeniceConfig() {
       kafkaBootstrapServers = backendConfig.getString(KAFKA_BOOTSTRAP_SERVERS);
     }
 
+    String recordTransformerOutputValueSchema = daVinciConfig.isRecordTransformerEnabled()
+        ? Objects.toString(recordTransformerConfig.getOutputValueSchema(), "null")
+        : "null";
```
ZacAttack, line 706: The ternary makes sense if you wanted to set with proper null, but as is this is redundant.

ZacAttack (approved):
Looks good!

PR #1505, opened 2025-02-05
[test] Add test logic for verifying thread pool metric emitted before server shutdown during integration test.

haoxu07:

In production, we previously observed no state transition metrics emitted, here we add test logic to verify these metrics will be emitted for every Venice server instances. 

## How was this PR tested?

ci

### DIFF
--- a/internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/integration/utils/VeniceServerWrapper.java
+++ b/internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/integration/utils/VeniceServerWrapper.java
@@ -51,6 +51,7 @@ import com.linkedin.venice.server.VeniceServer;
 import com.linkedin.venice.server.VeniceServerContext;
 import com.linkedin.venice.servicediscovery.ServiceDiscoveryAnnouncer;
 import com.linkedin.venice.tehuti.MetricsAware;
+import com.linkedin.venice.tehuti.MockTehutiReporter;
 import com.linkedin.venice.utils.ForkedJavaProcess;
 import com.linkedin.venice.utils.PropertyBuilder;
 import com.linkedin.venice.utils.SslUtils;
@@ -58,6 +59,7 @@ import com.linkedin.venice.utils.TestUtils;
 import com.linkedin.venice.utils.Utils;
 import com.linkedin.venice.utils.VeniceProperties;
 import com.linkedin.venice.utils.metrics.MetricsRepositoryUtils;
+import io.tehuti.Metric;
 import io.tehuti.metrics.MetricsRepository;
 import java.io.File;
 import java.io.IOException;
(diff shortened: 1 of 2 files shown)


internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/integration/utils/VeniceServerWrapper.java
```diff
@@ -454,12 +457,28 @@ protected void internalStart() throws Exception {
...
       veniceServer.shutdown();
     } else {
       serverProcess.destroy();
     }
   }
 
+  private void verifyHelixParticipantServicePoolMetricsReporting(VeniceServer veniceServer) {
+    MetricsRepository metricsRepository = veniceServer.getMetricsRepository();
+    MockTehutiReporter reporter = new MockTehutiReporter();
+    metricsRepository.addReporter(reporter);
```
m-nagarajan, line 470: Curious why can't we verify without adding a reporter?
```
MetricsRepository metricsRepository = veniceServer.getMetricsRepository();
Metric activeThreadNumber = metricsRepository.getMetric(".Venice_L/F_ST_thread_pool--active_thread_number.LambdaStat");
    Assert.assertNotNull(activeThreadNumber);
    Assert.assertTrue(activeThreadNumber.value() >= 0);
    Metric maxThreadNumber = metricsRepository.getMetric(".Venice_L/F_ST_thread_pool--max_thread_number.LambdaStat");
    Assert.assertNotNull(maxThreadNumber);
    Assert.assertTrue(maxThreadNumber.value() > 0);
    Metric queuedTaskNumber = metricsRepository.getMetric(".Venice_L/F_ST_thread_pool--queued_task_number.LambdaStat");
    Assert.assertNotNull(queuedTaskNumber);
    Assert.assertTrue(queuedTaskNumber.value() >= 0);
```

haoxu07: Yeah, without reporter we could still get metric value. Here is just for simulating real-prod scenario to use real reporter.

m-nagarajan: Makes sense.

m-nagarajan (approved):
Thanks @haoxu07

huangminchn (approved):
Thanks Hao!

FelixGV:
These new assertions get tested every single time a `VeniceServerWrapper` gets closed, which is in nearly every integration test... and it turns out the assertions are flaky (IDK why yet).

If this is a super critical functionality, then I could see why it warrants getting tested as part of every single integration test...

But if not, then it seems kind of noisy to risk having some test failure which is completely unrelated to the intent of the test.

Has this been considered? Or did we just place the assertions here at random while they could just as well have gotten a dedicated test?

PR #1506, opened 2025-02-06
[WIP] [Not Ready For Review] Adding poll rate to check how slow partition consuming.

haoxu07:

Resolves #XXX


haoxu07:
Close it temporarily, will resume it very soon to add rate feature.

PR #1507, opened 2025-02-06
[controller] Log remote address in audit log for better troubleshooting

sushantmane:
## Log remote address in audit log for better troubleshooting

Sometimes it's necessary to identify the source of requests to the Venice controller.  
Without logging the remote address, debugging can be difficult. This change adds  
the client IP to the audit log, to make troubleshooting easier.

## How was this PR tested?

UT and E2E Tests


lluwm (approved):
Thanks and LGTM!

sushantmane:
Thanks, @lluwm!

PR #1508, opened 2025-02-06
[server] Validate block-cache config against system memory

majisourav99:
##  Validate block-cache config against system memory

If operator deploys config with exceeds system resource limits, it makes the cluster into inoperable state and finally destroys the cluster. This PR fails rolling out such config for blockcache.

## How was this PR tested?

GHCI

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBServerConfig.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBServerConfig.java
@@ -237,6 +237,8 @@ public class RocksDBServerConfig {
       "rocksdb.blob.garbage.collection.force.threshold";
   public static final String ROCKSDB_BLOB_FILE_STARTING_LEVEL = "rocksdb.blob.file.starting.level";
 
+  public static final String ROCKSDB_BLOCK_CACHE_MEMORY_LIMIT = "rocksdb.block.cache.memory.limit";
+
   private final boolean rocksDBUseDirectReads;
 
   private final int rocksDBEnvFlushPoolSize;
@@ -312,6 +314,7 @@ public class RocksDBServerConfig {
   private final double blobGarbageCollectionAgeCutOff;
   private final double blobGarbageCollectionForceThreshold;
   private final int blobFileStartingLevel;
+  private final double rocksdbBlockCacheMemoryLimit;
 
   public RocksDBServerConfig(VeniceProperties props) {
     // Do not use Direct IO for reads by default
@@ -450,6 +453,7 @@ public class RocksDBServerConfig {
     this.blobGarbageCollectionAgeCutOff = props.getDouble(ROCKSDB_BLOB_GARBAGE_COLLECTION_AGE_CUTOFF, 0.25);
     this.blobGarbageCollectionForceThreshold = props.getDouble(ROCKSDB_BLOB_GARBAGE_COLLECTION_FORCE_THRESHOLD, 0.8);
     this.blobFileStartingLevel = props.getInt(ROCKSDB_BLOB_FILE_STARTING_LEVEL, 0);
+    this.rocksdbBlockCacheMemoryLimit = props.getDouble(ROCKSDB_BLOCK_CACHE_MEMORY_LIMIT, 0.8);
   }
 
   public int getLevel0FileNumCompactionTriggerWriteOnlyVersion() {
@@ -685,5 +689,9 @@ public class RocksDBServerConfig {
 
   public int getBlobFileStartingLevel() {
     return blobFileStartingLevel;
+  }
+
+  public double getRocksdbBlockCacheMemoryLimit() {
+    return rocksdbBlockCacheMemoryLimit;
   }
 }
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBServerConfig.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBServerConfig.java
@@ -338,9 +338,9 @@ public class RocksDBServerConfig {
     }
 
     this.rocksDBBlockCacheSizeInBytes =
-        props.getSizeInBytes(ROCKSDB_BLOCK_CACHE_SIZE_IN_BYTES, 16 * 1024 * 1024 * 1024L); // 16GB
+        props.getSizeInBytes(ROCKSDB_BLOCK_CACHE_SIZE_IN_BYTES, 2 * 1024 * 1024 * 1024L); // 2GB
     this.rocksDBRMDBlockCacheSizeInBytes =
-        props.getSizeInBytes(ROCKSDB_RMD_BLOCK_CACHE_SIZE_IN_BYTES, 2 * 1024 * 1024 * 1024L); // 2GB
+        props.getSizeInBytes(ROCKSDB_RMD_BLOCK_CACHE_SIZE_IN_BYTES, 1 * 1024 * 1024 * 1024L); // 1GB
 
     this.rocksDBBlockCacheImplementation = RocksDBBlockCacheImplementations
         .valueOf(props.getString(ROCKSDB_BLOCK_CACHE_IMPLEMENTATION, RocksDBBlockCacheImplementations.LRU.toString()));
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBServerConfig.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBServerConfig.java
@@ -338,9 +338,9 @@ public class RocksDBServerConfig {
     }
 
     this.rocksDBBlockCacheSizeInBytes =
-        props.getSizeInBytes(ROCKSDB_BLOCK_CACHE_SIZE_IN_BYTES, 2 * 1024 * 1024 * 1024L); // 2GB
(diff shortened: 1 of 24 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBStorageEngineFactory.java
```diff
@@ -139,6 +140,15 @@ public RocksDBStorageEngineFactory(
     this.env.setBackgroundThreads(rocksDBServerConfig.getRocksDBEnvFlushPoolSize(), Priority.HIGH);
     this.env.setBackgroundThreads(rocksDBServerConfig.getRocksDBEnvCompactionPoolSize(), Priority.LOW);
 
+    long cacheBytesNeeded =
+        rocksDBServerConfig.getRocksDBBlockCacheSizeInBytes() + (rocksDBServerConfig.isUseSeparateRMDCacheEnabled()
+            ? rocksDBServerConfig.getRocksDBRMDBlockCacheSizeInBytes()
+            : 0);
+    if (Runtime.getRuntime().maxMemory() * 0.8 < cacheBytesNeeded) {
```
m-nagarajan, line 147: can we make this threshold configurable?
Also can multiple rocksDB be instantiated resulting in a combined usage of more than the allowed memory?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBServerConfig.java
```diff
@@ -338,9 +338,9 @@ public RocksDBServerConfig(VeniceProperties props) {
     }
 
     this.rocksDBBlockCacheSizeInBytes =
-        props.getSizeInBytes(ROCKSDB_BLOCK_CACHE_SIZE_IN_BYTES, 16 * 1024 * 1024 * 1024L); // 16GB
+        props.getSizeInBytes(ROCKSDB_BLOCK_CACHE_SIZE_IN_BYTES, 2 * 1024 * 1024 * 1024L); // 2GB
```
m-nagarajan, line 341: why the jump from 16 -> 2?

m-nagarajan (commented):
Thanks @majisourav99. Left a couple of questions.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBServerConfig.java
ZacAttack: Glancing at our config repo, we're already explicitly setting these in all colos, so the only users who this 'might' impact at davinci users?  We should check that.  I don't think shrinking it is going to be a problem for server deployments.  Though I too am curious to the logic for why we'd like to shrink these defaults in this PR

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBStorageEngineFactory.java
ZacAttack: I don't think this will work.  The Runtime will tell you information about the JVM, but the rocksdb data usage won't be billed against that?

majisourav99: blockcache is shared amongst all rocksdb engines.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBServerConfig.java
majisourav99: To make all the tests pass as mac jvm allocates 4G of memory

majisourav99: DVC users pick up from global-config or can override it, which is also set to 16G.

ZacAttack: Can we please verify how this config is getting defaulted for Davinci users?  If we're covered there I think this is fine.

gaojieliu: Emm.. Why not just override it in `VeniceServerWrapper` instead of here to let test pass in Mac?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBServerConfig.java
```diff
@@ -338,9 +338,9 @@ public RocksDBServerConfig(VeniceProperties props) {
...
     this.rocksDBRMDBlockCacheSizeInBytes =
-        props.getSizeInBytes(ROCKSDB_RMD_BLOCK_CACHE_SIZE_IN_BYTES, 2 * 1024 * 1024 * 1024L); // 2GB
+        props.getSizeInBytes(ROCKSDB_RMD_BLOCK_CACHE_SIZE_IN_BYTES, 1 * 1024 * 1024 * 1024L); // 1GB
```
gaojieliu, line 343: Same question as above?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBStorageEngineFactory.java
gaojieliu: Yeah, Zac is correct.
One way in Oracle JVM (getting this from stackoverflow):
```
        long memorySize = ((com.sun.management.OperatingSystemMXBean) ManagementFactory.getOperatingSystemMXBean()).getTotalPhysicalMemorySize();
```
Not sure whether it works with OpenJDK or not.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/rocksdb/RocksDBStorageEngineFactory.java
```diff
@@ -139,6 +141,17 @@ public RocksDBStorageEngineFactory(
     this.env.setBackgroundThreads(rocksDBServerConfig.getRocksDBEnvFlushPoolSize(), Priority.HIGH);
     this.env.setBackgroundThreads(rocksDBServerConfig.getRocksDBEnvCompactionPoolSize(), Priority.LOW);
 
+    long cacheBytesNeeded =
+        rocksDBServerConfig.getRocksDBBlockCacheSizeInBytes() + (rocksDBServerConfig.isUseSeparateRMDCacheEnabled()
+            ? rocksDBServerConfig.getRocksDBRMDBlockCacheSizeInBytes()
+            : 0);
+
+    long systemMemorySize = getOSMemorySize();
+    if (systemMemorySize > 0 && (systemMemorySize * 0.8 < cacheBytesNeeded)) {
```
gaojieliu, line 150: As @m-nagarajan mentioned, it is better to make the ratio configurable.

internal/venice-common/src/main/java/com/linkedin/venice/utils/Utils.java
```diff
@@ -1141,4 +1142,16 @@ public static long parseDateTimeToEpoch(String dateTime, String dateTimeFormat,
     dateFormat.setTimeZone(TimeZone.getTimeZone(timeZone));
     return dateFormat.parse(dateTime).getTime();
   }
+
+  public static long getOSMemorySize() {
+    OperatingSystemMXBean osBean = ManagementFactory.getOperatingSystemMXBean();
+
+    if (osBean instanceof com.sun.management.OperatingSystemMXBean) {
+      com.sun.management.OperatingSystemMXBean extendedOsBean = (com.sun.management.OperatingSystemMXBean) osBean;
```
gaojieliu, line 1150: Have you tested this function in Mac or Linux with OpenJDK?
If it doesn't work in these envs, this feature will be useless...

majisourav99: Yes, this works in mac, and linux container/helix code uses same mechanism to fetch system memory.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/integration/utils/VeniceServerWrapper.java
```diff
@@ -264,6 +266,8 @@ static StatefulServiceProvider<VeniceServerWrapper> generateService(
           .put(SERVER_INGESTION_HEARTBEAT_INTERVAL_MS, 5000)
           .put(SERVER_LEADER_COMPLETE_STATE_CHECK_IN_FOLLOWER_VALID_INTERVAL_MS, 5000)
           .put(SERVER_RESUBSCRIPTION_TRIGGERED_BY_VERSION_INGESTION_CONTEXT_CHANGE_ENABLED, true)
+          .put(ROCKSDB_BLOCK_CACHE_SIZE_IN_BYTES, 8 * 1024 * 1024 * 1024L)
```
gaojieliu, line 269: Why do we need so big cache for integration tests and why we override this config in other integration tests?
Shall we just change it to be something like 256MB here and it will be good by default?

majisourav99: Sure, but venice code has been using this much cache in tests currently.

gaojieliu (approved):
Thanks for the change!

PR #1509, opened 2025-02-07
[server][common][vpj] Introduce ComplexVenicePartitioner to materialized view

xunyin8:
## [server][common][vpj] Introduce ComplexVenicePartitioner to materialized view

The change will not work if record is actually large and chunked. Proper chunking support is needed and will be addressed in a separate PR.

1. Introduced ComplexVenicePartitioner which extends VenicePartitioner and offer a new API to partition by value and provide possible one-to-many partition mapping.

2. Added value provider of type Lazy<GenericRecord> to VeniceViewWriter's processRecord API to access deserialized value if needed. e.g. when a ComplexVenicePartitioner is involved.

3. MergeConflictResultWrapper and WriteComputeResultWrapper will now provide deserialized value in a best effort manner. This is useful when we already deserialized the value for a partial update operation so that the deserialized value can be provided directly to the materialized view writer.

4. Refactored VeniceWriter to expose some APIs to child class. Introduced ComplexVeniceWriter which extends VeniceWriter. Reasoning here is that the ComplexVeniceWriter will have different APIs to be used in MaterializedViewWriter and CompositeVeniceWriter to write to materialized view partition(s) and potentially involving a ComplexVenicePartitioner. Alternatively we could push common logic from VeniceWriter to AbstractVeniceWriter. However, ComplexVeniceWriter needs/shares too much common logic with VeniceWriter (chunking, DIV support, pubSubAdapter, etc...) it will make AbstractVeniceWriter too specialized and unable to offer the flexibility it needs to support something like the CompositeVeniceWriter.

5. Override putLargeValue in ComplexVeniceWriter to skip chunking and writing large messages. Once we have proper chunking support we need to be careful to not re-chunk when writing the same value to different partition in ComplexVeniceWriter.

## How was this PR tested?
Added new integration test with A/A, W/C and a new test value based partitioner.
**Will add new unit tests once we have some consensus on the API changes**


internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
```diff
@@ -2153,4 +2203,44 @@ public String getDestination() {
...
+   * it could return a single, multiple or no partition(s). The delete behavior is also undefined, so we will ignore
+   * deletes via key for now. Proper delete can be performed via re-push or new version push.
+   */
+  public CompletableFuture<PubSubProduceResult> writeWithComplexPartitioner(
+      K key,
+      V value,
+      int valueSchemaId,
+      Lazy<GenericRecord> valueProvider,
+      VeniceComplexPartitioner complexPartitioner,
+      int numPartitions) {
```
gaojieliu, line 2218: I am concerned about this API as it takes more params:
1. Extra partitioner.
2. Extra valueProvider.
For 1), in theory, one VeniceWriter should only use one partitioner at all times and it shouldn't need to take a partitioner at write time.
The meaning of the new valueProvider is not very clear.

I am thinking whether we should evolve `VeniceWriter` API to support a composite partition key object or not.
Essentially, all the `put/update/delete` methods will take a wrapper object: `PartitioningKeyWrapper`, which can contain:
```
1. key bytes/object.
2. Lazy<GenericRecord> value.
```
And the VeniceWriter only takes one partitioner at construction time, and we should continue to support the existing API, but internally, it will wrap the input into the new wrapper and in `VeniceWriter#getPartition`, it will always return a set and inside this function, it can do a type check (can be done at construction time and maintain a flag) to decide whether it should call the new complex partitioner API or the original partition API.
If it is a complex partitioner, but the wrapper doesn't contain a valid `value`, it should fail fast.

So in VPJ, it can invoke the new methods, which are taking the new wrapper param: `PartitioningKeyWrapper` and it will be easier in `CompositeVeniceWriter`, as it can forward the wrapper to all the child writers, and each child writer can make a decision on itself to use just `key` bytes or both `key` bytes and `value` depending on the partitioner.

It will be a slightly big refactoring, but I think it will make the interface cleaner and more maintanable.
I would like to get more opinions from others.
@FelixGV @ZacAttack 
Can you share your opinions here?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/view/MaterializedViewWriter.java
```diff
@@ -48,6 +50,11 @@ public MaterializedViewWriter(
     this.veniceWriter = Lazy.of(
         () -> new VeniceWriterFactory(props.getCombinedProperties().toProperties(), pubSubProducerAdapterFactory, null)
             .createVeniceWriter(buildWriterOptions()));
+    if (internalView.getViewPartitioner() instanceof VeniceComplexPartitioner) {
+      complexPartitioner = (VeniceComplexPartitioner) internalView.getViewPartitioner();
```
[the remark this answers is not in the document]
xunyin8: So one of the ways we thought about how to support proper chunking during NR is to forward chunks based on partition specified by upstream producer. i.e. VPJ writers will include partitioning result so the NR leader during pass-through mode can forward the chunks without needing to re-assemble them. This will need VeniceWriter to have the ability to send messages to arbitrary partition(s).

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/task/datawriter/ComplexPartitionerWriterAdapter.java
```diff
@@ -0,0 +1,95 @@
...
+        key,
+        value,
+        valueSchemaId,
+        Lazy.of(() -> deserializer.apply(value)),
+        complexPartitioner,
+        numPartitions);
+  }
+
+  @Override
+  public CompletableFuture<PubSubProduceResult> delete(
```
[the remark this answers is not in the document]
xunyin8: I have a few questions/concerns about supporting delete this way:
1. We don't have any mechanism to ensure the values that we partition on are actually immutable so deletion via previous full value is going to be best effort. i.e. there will still be resource leak if upstream changes the value(s) multiple times.
2. Does today's L/F (without A/A) have this read back capability? Or do we need a refactor to be able to read the old value upon encountering a deletion event for the original store (RT/VT).

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/view/MaterializedViewWriter.java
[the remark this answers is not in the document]
gaojieliu: Got it, and we can leave it there for now.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/task/datawriter/ComplexPartitionerWriterAdapter.java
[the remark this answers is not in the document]
gaojieliu: For 1), best effort is fine from my understanding.
For 2), yes, we need some refactoring to support this new behavior.

Today, it is guaranteed to be leaking without 1).

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/task/datawriter/AbstractPartitionWriter.java
```diff
@@ -403,6 +410,22 @@ private AbstractVeniceWriter<byte[], byte[], byte[]> createCompositeVeniceWriter
         VeniceView view = ViewUtils
             .getVeniceView(viewConfig.getViewClassName(), new Properties(), storeName, viewConfig.getViewParameters());
         String viewTopic = view.getTopicNamesAndConfigsForVersion(versionNumber).keySet().stream().findAny().get();
+        if (view instanceof MaterializedView) {
+          MaterializedView materializedView = (MaterializedView) view;
+          if (materializedView.getViewPartitioner() instanceof VeniceComplexPartitioner) {
+            // We need to build a ComplexPartitionerWriterAdapter to handle writes with complex partitioner.
+            Schema valueSchema = new Schema.Parser().parse(props.getString(VALUE_SCHEMA_STRING_PROP));
```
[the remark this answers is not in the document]
xunyin8: Probably not then, I need to double check re-push logic and potentially lock-in a working value schema during job setup.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/replication/merge/MergeConflictResult.java
```diff
@@ -56,4 +68,14 @@ public boolean doesResultReuseInput() {
...
     return rmdRecord;
   }
+
+  /**
+   * Provide the deserialized new value on a best-effort approach. Meaning that it's acceptable to return an empty
+   * Optional. e.g. MergeConflictResult of full PUTs will not contain deserialized new value since we don't need to
+   * deserialize the value to generate the MCR.
+   * @return deserialized new value if possible.
+   */
+  public Optional<GenericRecord> getNewValueDeserialized() {
```
ZacAttack, line 78: What is the expected type of this GenericRecord?  I'm assuming it's the user record sans Venice enveloping?

xunyin8: It should be the user value record without the Venice enveloping, see integration test example.

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
xunyin8: I agree that taking partitioner as a parameter for an instance of VeniceWriter which is constructed based on some pre-defined partitioner config is wrong and need to be fixed. I think the suggested refactor is good.

internal/venice-client-common/src/main/java/com/linkedin/venice/partitioner/VeniceComplexPartitioner.java
```diff
@@ -0,0 +1,33 @@
...
+  }
+
+  /**
+   * A complex partitioner API that could be used in materialized views to partition based on value. The resulting
+   * partition could also be a set of partition Ids instead of just a single partition (one-to-many).
+   * @param value that will be mapped to partition(s)
+   * @param numPartitions of total available partitions
+   * @return
+   */
+  public abstract Set<Integer> getPartitionId(Object value, int numPartitions);
```
FelixGV, line 32: A couple of nitpicks...

1. `Set<Integer>` can be expensive... if we insist on this being a set (which of course makes sense sematically...) then let's at least use fast-util's `IntSet` instead. Alternatively, returning a `int[]` could be perfectly acceptable since this is a low-level API that will be called on the hot path, so making it as lean as possible is useful...
2. Having just the `value` as input seems not ideal. We should also pass in the key, IMHO. There is nothing wrong with partitioning by both key and value, or by one or the other (e.g. if one field of the key is 'xyz', then partition by that other field of the key, otherwise partition by so and so fields of the value).
3. `Object` seems not right... we should either pass the serialized value (e.g. `byte[]` or `ByteBuffer`) and leave the concern of deser to the implementer (probably not ideal, but there could be reasons to...) or use generics for both they key and the value...

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
```diff
@@ -2153,4 +2203,44 @@ public String getDestination() {
   public PubSubProducerAdapter getProducerAdapter() {
     return this.producerAdapter;
   }
+
+  /**
+   * {@link VeniceComplexPartitioner} offers a more sophisticated getPartitionId API. It takes value as a parameter, and
+   * it could return a single, multiple or no partition(s). The delete behavior is also undefined, so we will ignore
+   * deletes via key for now. Proper delete can be performed via re-push or new version push.
+   */
+  public CompletableFuture<PubSubProduceResult> writeWithComplexPartitioner(
```
FelixGV, line 2212: What does the returned CF mean? That all partitions have been successfully produced into? What if there is a partial failure? Should this be a collection of results instead?

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
FelixGV: Copying my 2 cents here from an internal conversation, which might be in-line with what Gaojie is saying (at least partially):

IMHO the `VW` should not support that at all. It increases the scope too much and we use the VW in a bunch of places, so adding more available APIs to all these locations, that are not going to exercise them, is not ideal…

We already have a `AbstractVeniceWriter` but it’s currently very lean… I think potentially what we should do is this:

Create a new class, e.g. `SprayingVeniceWriter` (but let’s think carefully and pick a good name), then have that extend `AbstractVeniceWriter`. Then, whatever internal utility functions exist inside `VW` that the `SVW` need, move these functions into `AVW` as protected. Add the new APIs into the `SVW`, and make its constructor take in the `VeniceComplexPartitioner` instead of normal partitioner…

Something along those lines…

FelixGV (commented):
Just some early thoughts... did not read the whole PR yet. But hopefully useful in terms of discussing the API changes.

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
xunyin8: Correct, the current implementation returns a CF that's completed if all partitions have been successfully produced into. If there is a partial failure then the returned CF is completed exceptionally and there will be no way for caller to know which partition succeeded and which partitions failed.

FelixGV: Ok well I guess that is fine, BUT... we should explain that in the JavaDoc, and IMHO it doesn't make sense to return a `PubSubProduceResult` which inherently represents just one partition. It can be a `CompletableFuture<Void>` I guess, or else return some new type which makes sense in that context, if there is anything we really need in there...

internal/venice-client-common/src/main/java/com/linkedin/venice/partitioner/VeniceComplexPartitioner.java
xunyin8: 1 and 2 makes sense but I have some concerns regarding to 3:
Asking the implementer of the partitioner to figure out how to deserialize the value could be challenging and error prone. In the Venice backend we have access to things like writer schema version to help deserialize values properly when multiple value schema versions are involved. Unless we also provide similar capabilities and expect the implementers of `VenicePartitioner` to know about these details it's going to be difficult for implementer to do serialization properly by just throwing them a `byte[]` or `ByteBuffer`

xunyin8: Since there is a cost to deserialize the key and value I think if we want to support 2 then we should make both Lazy in case they only need one or the other. Something like:
`getPartitionId(Lazy<GenericRecord> key, Lazy<GenericRecord> value, int numPartitions);`

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
xunyin8: The common logic needed between the new ComplexVeniceWriter and VeniceWriter is too much (we need chunking, DIV, put, delete, etc...) We also need to keep AbstractVeniceWriter relatively thin and flexible to support something like the CompositeVeniceWriter. Therefore, instead of doing the refactor suggested above we can make ComplexVeniceWriter extend VeniceWriter for now to keep the change minimal and keep the APIs separated.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -4026,4 +4028,11 @@ private void maybeQueueCMWritesToVersionTopic(
       produceCall.run();
     }
   }
+
+  protected Lazy<GenericRecord> getNewValueProvider(ByteBuffer newValue, int schemaId) {
+    if (newValue == null) {
+      return Lazy.of(() -> null);
+    }
+    return Lazy.of(() -> storeDeserializerCache.getDeserializer(schemaId, schemaId).deserialize(newValue));
```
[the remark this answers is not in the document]
xunyin8: moving the logic into `MergeConflictResultWrapper` and `WriteComputeResultWrapper` will actually introduce duplicated code fragments and additional work when views are not involved. Let me try to solve this null/empty check in another way.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -3331,7 +3338,19 @@ private PubSubMessageProcessedResult processMessage(
         if (isWriteComputationEnabled && partitionConsumptionState.isEndOfPushReceived()) {
           partitionConsumptionState.setTransientRecord(kafkaClusterId, consumerRecord.getOffset(), keyBytes, -1, null);
         }
-        return new PubSubMessageProcessedResult(new WriteComputeResultWrapper(null, null, false));
+        // Best-effort to provide the old value for delete operation in case needed by a ComplexVeniceWriter to generate
```
ZacAttack, line 3341: This is actually pretty interesting.  There.... would be many edge cases where this wouldn't work out...  But....  I don't think we could do any better lol.

Really, this should be a great big warning label that view logic needs to be idempotent.  Otherwise all bets are off.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/MergeConflictResultWrapper.java
```diff
@@ -64,4 +67,31 @@ public ByteBuffer getUpdatedValueBytes() {
   public ByteBuffer getUpdatedRmdBytes() {
     return updatedRmdBytes;
   }
+
+  /**
+   * Return a best-effort value provider with the following behaviors:
+   *   1. returns the new value provider for PUT and UPDATE.
+   *   2. returns the old value for DELETE (null for non-existent key).
+   *   3. returns null if the value is not available.
```
ZacAttack, line 75: What is the case where the value would be totally not available?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/view/VeniceViewWriter.java
```diff
@@ -62,18 +62,19 @@ public VeniceViewWriter(
    * @param newValue the incoming fully specified value which hasn't yet been committed to Venice
```
ZacAttack, line 62: Let's update the javadoc a bit here as newValueProvider and newValue sound similar but are certainly different.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
xunyin8: I agree... but like Gaojie mentioned in earlier comments, it could be useful in the event where we could propagate the delete properly to view topics even when a value based partitioner is used.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/MergeConflictResultWrapper.java
xunyin8: Today there's nothing to prevent users trying to delete a non-existent key. In that case the value would be totally not available.

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriterFactory.java
```diff
@@ -61,35 +61,47 @@ public VeniceWriterFactory(
...
-        ? venicePropertiesLazy.get()
-        : venicePropertiesWithCompressionDisabledLazy.get();
+    VeniceProperties props = getVeniceProperties(options);
+    Supplier<PubSubProducerAdapter> producerAdapterSupplier =
+        () -> producerAdapterFactory.create(props, options.getTopicName(), options.getBrokerAddress());
+    PubSubProducerAdapter producerAdapter = buildPubSubProducerAdapter(options, producerAdapterSupplier);
+    return new VeniceWriter<>(options, props, producerAdapter);
+  }
 
+  public <K, V, U> ComplexVeniceWriter<K, V, U> createComplexVeniceWriter(VeniceWriterOptions options) {
```
gaojieliu, line 71: Shall we enforce that `ComplexVeniceWriter` will always use `ComplexVenicePartitioner`? otherwise, the logic is a little confusing to me.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/task/datawriter/ComplexPartitionerWriterAdapter.java
```diff
@@ -0,0 +1,98 @@
...
+        });
+    return wraper;
+  }
+
+  @Override
+  public CompletableFuture<PubSubProduceResult> delete(
+      K key,
+      PubSubProducerCallback callback,
+      DeleteMetadata deleteMetadata) {
+    // No-op, delete by key is undefined for complex partitioner.
```
gaojieliu, line 75: Actually, this is the expected behavior and it is good to call it out.
In VPJ, only repush can trigger this function and during re-push, the deletion to view topics are useless, which can be safely ignored.

internal/venice-common/src/main/java/com/linkedin/venice/writer/ComplexVeniceWriter.java
```diff
@@ -0,0 +1,163 @@
...
+
+  public ComplexVeniceWriter(
+      VeniceWriterOptions params,
+      VeniceProperties props,
+      PubSubProducerAdapter producerAdapter) {
+    super(params, props, producerAdapter);
+    if (partitioner instanceof ComplexVenicePartitioner) {
+      complexPartitioner = (ComplexVenicePartitioner) partitioner;
+    } else {
+      complexPartitioner = null;
```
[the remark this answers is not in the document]
xunyin8: Yes, my plan was always use `ComplexVeniceWriter`'s complex put and delete API in `MaterializedViewWriter` regardless of the underlying partitioner. The `ComplexVeniceWriter` internally will figure out what to do if it's a `ComplexVenicePartitioner` or just a `VenicePartitioner`.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/task/datawriter/ComplexPartitionerWriterAdapter.java
```diff
@@ -0,0 +1,98 @@
...
+import com.linkedin.venice.writer.DeleteMetadata;
+import com.linkedin.venice.writer.PutMetadata;
+import java.io.IOException;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.Future;
+import java.util.function.Function;
+import org.apache.avro.generic.GenericRecord;
+
+
+public class ComplexPartitionerWriterAdapter<K, V, U> extends AbstractVeniceWriter<K, V, U> {
```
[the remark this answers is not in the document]
xunyin8: Yes, since this class is only used by VPJ and we don't support delete by key via VPJ.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/task/datawriter/AbstractPartitionWriter.java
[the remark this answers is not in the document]
xunyin8: Ended up using `HDFSSchemaSource` and involving `VeniceCompressor` just like what we do in `VeniceRmdTTLFilter` in order to properly deserialize and decompress for both regular push and re-push.

internal/venice-common/src/main/java/com/linkedin/venice/writer/ComplexVeniceWriter.java
```diff
@@ -0,0 +1,163 @@
...
+      int valueSchemaId,
+      PubSubProducerCallback callback,
+      int partition,
+      LeaderMetadataWrapper leaderMetadataWrapper,
+      long logicalTs,
+      PutMetadata putMetadata,
+      ChunkedValueManifest oldValueManifest,
+      ChunkedValueManifest oldRmdManifest) {
+    skippedLargeRecords.incrementAndGet();
+    if (!REDUNDANT_LOGGING_FILTER.isRedundantException(topicName, SKIP_LARGE_RECORD)) {
```
gaojieliu, line 123: Do we need to check whether it is a complex partitioner or not before skipping?
For the regular partitioner, I assume we can let large value go to viewer topic, right?

internal/venice-common/src/main/java/com/linkedin/venice/writer/ComplexVeniceWriter.java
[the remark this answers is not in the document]
gaojieliu: I think mixing the purpose of this class carries some risks as the regular API: put/delete without passing valueProvider are still available, and in theory, the user of this class can invoke these regular APIs, and the result will be wrong or the call will be broken, which will be troublesome.

If we really want this class to be able to handle both regular partitioner and complex partitioner, we need to override all the public APIs and create branching logics for regular and complex partitioners.
Or just throw `UnsupportedOperationException` in other public APIs, which makes the purpose clear that only the `complex` APIs can be used even for the regular partitioner.

internal/venice-common/src/main/java/com/linkedin/venice/writer/CompositeVeniceWriter.java
```diff
@@ -23,13 +20,13 @@
 @NotThreadsafe
 public class CompositeVeniceWriter<K, V, U> extends AbstractVeniceWriter<K, V, U> {
   private final VeniceWriter<K, V, U> mainWriter;
-  private final VeniceWriter<K, V, U>[] childWriters;
+  private final AbstractVeniceWriter<K, V, U>[] childWriters;
```
gaojieliu, line 23: This class currently is only being used in VPJ, so shall we move it to `venice-push-job` module instead to avoid confusion?

internal/venice-common/src/main/java/com/linkedin/venice/writer/ComplexVeniceWriter.java
xunyin8: No, we still cannot let large value to go to viewer topic even for regular partitioner. This is because even for regular partitioner a single partition can have multiple leader writing to it and breaks the assumption of single leader for chunk assembling. The consumer support for that is still missing.

gaojieliu: This is bad...
We do have DaVinci use cases, who might want to use the regular partitioner and DaVinci can handle chunking in a correct way.
I wondered whether the large message skipping can be done in CDC instead of here for regular partitioner.
It can be done in a later PR, but I do think we need to revaluate such decision.

internal/venice-common/src/main/java/com/linkedin/venice/writer/ComplexVeniceWriter.java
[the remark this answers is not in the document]
xunyin8: I think the suggested throw `UnsupportedOperationException` in other public APIs is better because:
1. Even if we enforce `ComplexVeniceWriter` to be only allowed to initialize with a `ComplexVenicePartitioner` the public APIs of the `AbstractVeniceWriter` can still be called and produce the wrong result (it will use partition by key and not involve value provider).
2. More branching and casting logic in `MaterializedViewWriter`

internal/venice-common/src/main/java/com/linkedin/venice/writer/ComplexVeniceWriter.java
xunyin8: Sure, we can revisit this but wanted to point out that even for DVC chunking is not fully supported today if we have large values in the batch push. This is because during NR pass-through we don't know how to forward the chunks to view topic partitions correctly. The logic as of now will partition and send the chunks as regular records which will most likely be wrong.

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
```diff
@@ -96,7 +96,7 @@ public class VeniceWriter<K, V, U> extends AbstractVeniceWriter<K, V, U> {
   private final ThreadPoolExecutor threadPoolExecutor;
 
   // log4j logger
-  private final Logger logger;
+  protected final Logger logger;
```
ZacAttack, line 99: I'm not finding it, why open this up?

internal/venice-common/src/main/java/com/linkedin/venice/writer/ComplexVeniceWriter.java
```diff
@@ -0,0 +1,163 @@
...
+
+  /**
+   * {@link ComplexVenicePartitioner} offers a more sophisticated getPartitionId API. It also takes value as a
+   * parameter, and could return a single, multiple or no partition(s).
+   */
+  public CompletableFuture<Void> complexPut(K key, V value, int valueSchemaId, Lazy<GenericRecord> valueProvider) {
+    CompletableFuture<Void> finalCompletableFuture = new CompletableFuture<>();
+    if (value == null) {
+      // Ignore null value
+      finalCompletableFuture.complete(null);
```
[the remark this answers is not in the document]
ZacAttack: uhhh.  Would a null value not potentially indicate a delete?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/MergeConflictResultWrapper.java
```diff
@@ -64,4 +67,31 @@ public ByteBuffer getUpdatedValueBytes() {
...
+   */
+  public Lazy<GenericRecord> getNewValueProvider(
```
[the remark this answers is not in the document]
xunyin8: Good point since everything should be determined and finalized during constructor time.

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
xunyin8: In `ComplexVeniceWriter`'s `putLargeValue`
```
  @Override
  protected CompletableFuture<PubSubProduceResult> putLargeValue(
      byte[] serializedKey,
      byte[] serializedValue,
      int valueSchemaId,
      PubSubProducerCallback callback,
      int partition,
      LeaderMetadataWrapper leaderMetadataWrapper,
      long logicalTs,
      PutMetadata putMetadata,
      ChunkedValueManifest oldValueManifest,
      ChunkedValueManifest oldRmdManifest) {
    skippedLargeRecords.incrementAndGet();
    if (!REDUNDANT_LOGGING_FILTER.isRedundantException(topicName, SKIP_LARGE_RECORD)) {
      logger.warn("Skipped writing {} large record(s) to topic: {}", skippedLargeRecords.getAndSet(0), topicName);
    }
    return CompletableFuture.completedFuture(null);
  }
```

internal/venice-common/src/main/java/com/linkedin/venice/writer/ComplexVeniceWriter.java
```diff
@@ -0,0 +1,163 @@
...
+   *   2. Previous writes
+   */
+  public CompletableFuture<Void> complexDelete(K key, Lazy<GenericRecord> valueProvider) {
+    CompletableFuture<Void> finalCompletableFuture = new CompletableFuture<>();
+    if (complexPartitioner == null) {
+      // No VeniceComplexPartitioner involved, perform simple delete.
+      propagateVeniceWriterFuture(this.delete(key, null), finalCompletableFuture);
+    } else {
+      GenericRecord value = valueProvider.get();
+      if (value == null) {
```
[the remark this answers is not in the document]
xunyin8: I don't think these events should be propagated to view topics but rather just ensure it can be handled properly and not throw exceptions (no-op).

gaojieliu (commented):
Code change looks good overall.
I do think we need to take care of the comment I just left, which is very tricky as it is a race condition.

gaojieliu (commented):
I think we need to actively lookup the previous value when there is a complex view writer:
https://github.com/linkedin/venice/blob/main/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java#L458

```
  if (hasChangeCaptureView) {
      /**
       * Since this function will update the transient cache before writing the view, and if there is
       * a change capture view writer, we need to lookup first, otherwise the transient cache will be populated
       * when writing to the view after this function.
       */
      oldValueProvider.get();
    }
```
I guess the condition can be `delete` op + complex partitioner.

gaojieliu (approved):
Thanks for the change!

PR #1511, opened 2025-02-07
[controller ]Avoid client side store deletion timeout due to admin message is not consumed.

haoxu07:

Store deletion is not fast operation for controller, but it could happen frequently in a short time range. Previously there is 30 seconds default wait for admin message of deletion to be consumed, client side easily get timeout when multiple store deletions happened at the same time for a same cluster. As store deletion is an asynchronous operation, we should not fail client side with time out exception. 

### DIFF
--- a/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
@@ -170,6 +170,7 @@ import com.linkedin.venice.exceptions.ResourceStillExistsException;
 import com.linkedin.venice.exceptions.VeniceException;
 import com.linkedin.venice.exceptions.VeniceHttpException;
 import com.linkedin.venice.exceptions.VeniceNoStoreException;
+import com.linkedin.venice.exceptions.VeniceTimeoutException;
 import com.linkedin.venice.exceptions.VeniceUnsupportedOperationException;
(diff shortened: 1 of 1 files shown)


services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
```diff
@@ -978,8 +979,13 @@ public void deleteStore(
       AdminOperation message = new AdminOperation();
       message.operationType = AdminMessageType.DELETE_STORE.getValue();
       message.payloadUnion = deleteStore;
-
-      sendAdminMessageAndWaitForConsumed(clusterName, storeName, message);
+      // When multiple store deletions happened in a short time range, it is possible that the store deletion message
+      // will not be consumed in time, as previous store deletion message is still being processed.
+      try {
+        sendAdminMessageAndWaitForConsumed(clusterName, storeName, message);
```
ZacAttack, line 985: so.... the effect here is that we will wait up to a timeout, then if we go above that timeout, we'll log an error and then we'll return to the client like everything is fine (as opposed to barfing an error).

....At that point, why wait at all?  Why not fire and forget?

haoxu07: Yeah, actually that's a good point. Even we waited until consumed still do not make too much sense.  Consumed does not mean admin execution successfully.

PR #1514, opened 2025-02-10
[duckdb] Bumped duckdb_jdbc to 1.2.0

FelixGV:
Miscellaneous:

- Got rid of the build tech debt related to the previous dependency on a SNAPSHOT release.

- Moved the asm dependency from all sub-modules to DVC, which is the only module that actually uses it.

Resolves #XXX

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,17 +1,11 @@
 version: 2
 registries:
-  # Helps find updates for non Maven Central dependencies'
-  duckdb-snapshots:
-    type: maven-repository
-    url: https://oss.sonatype.org/content/repositories/snapshots/
 
 updates:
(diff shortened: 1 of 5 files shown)


kvargha:
Can we also delete the dependabot snapshot updater I added here: https://github.com/linkedin/venice/pull/1492

kvargha:
Or we can modify the dependabot to always keep duckdb up to date

nisargthakkar (dismissed):
LGTM

FelixGV:
> Or we can modify the dependabot to always keep duckdb up to date, and just remove the snapshot stuff

Oh yeah, thanks. I missed that. I'll let dependabot stay, but without the snapshot repo. We'll see how it work I guess...

kvargha (approved):

PR #1517, opened 2025-02-10
[router][server] ACL handler efficiency improvement

FelixGV:
Re-ordered some operations to minimize ACL handling overhead, based on profiling.

### DIFF
--- a/internal/venice-common/src/main/java/com/linkedin/venice/acl/handler/AbstractStoreAclHandler.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/acl/handler/AbstractStoreAclHandler.java
@@ -62,7 +62,6 @@ public abstract class AbstractStoreAclHandler<REQUEST_TYPE> extends SimpleChanne
     }
 
     String uri = req.uri();
-    String method = req.method().name();
 
     // Parse resource type and store name
     String[] requestParts = URI.create(uri).getPath().split("/");
@@ -91,7 +90,7 @@ public abstract class AbstractStoreAclHandler<REQUEST_TYPE> extends SimpleChanne
     }
 
     X509Certificate clientCert = extractClientCert(ctx);
-
+    String method = req.method().name();
     AccessResult accessResult = checkAccess(uri, clientCert, storeName, method);
     switch (accessResult) {
       case GRANTED:
@@ -127,17 +126,12 @@ public abstract class AbstractStoreAclHandler<REQUEST_TYPE> extends SimpleChanne
   protected abstract REQUEST_TYPE validateRequest(String[] requestParts);
 
   protected AccessResult checkAccess(String uri, X509Certificate clientCert, String storeName, String method) {
-    if (VeniceSystemStoreUtils.isSystemStore(storeName)) {
-      return AccessResult.GRANTED;
-    }
-
-    String client = identityParser.parseIdentityFromCert(clientCert);
     try {
-      /**
-       * TODO: Consider making this the first check, so that we optimize for the hot path. If rejected, then we
-       *       could check whether the request is for a system store, METADATA, etc.
-       */
       if (accessController.hasAccess(clientCert, storeName, method)) {
+        return AccessResult.GRANTED;
+      }
+
+      if (VeniceSystemStoreUtils.isSystemStore(storeName)) {
         return AccessResult.GRANTED;
       }
 
@@ -146,7 +140,6 @@ public abstract class AbstractStoreAclHandler<REQUEST_TYPE> extends SimpleChanne
       // Possible Reasons:
       // A. ACL not found. OR,
       // B. ACL exists but caller does not have permission.
-      String errLine = String.format("%s requested %s %s", client, method, uri);
 
       if (!accessController.isFailOpen() && !accessController.hasAcl(storeName)) { // short circuit, order matters
         // Case A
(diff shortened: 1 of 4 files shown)


internal/venice-common/src/main/java/com/linkedin/venice/acl/handler/AbstractStoreAclHandler.java
```diff
@@ -160,7 +153,13 @@ protected AccessResult checkAccess(String uri, X509Certificate clientCert, Strin
...
         // return 401 Unauthorized
-        LOGGER.warn("Requested store does not have ACL: {}", errLine);
+        if (LOGGER.isWarnEnabled()) {
+          LOGGER.warn(
+              "Requested store does not have ACL: {} requested {} {}",
+              identityParser.parseIdentityFromCert(clientCert),
+              method,
+              uri);
+        }
         LOGGER.debug(
```
xunyin8, line 163: Do we want to check for `isDebugEnabled` for these lines too?

internal/venice-common/src/main/java/com/linkedin/venice/acl/handler/AbstractStoreAclHandler.java
```diff
@@ -186,17 +185,25 @@ protected AccessResult checkAccess(String uri, X509Certificate clientCert, Strin
...
+          ? String.format("%s requested %s %s", identityParser.parseIdentityFromCert(clientCert), method, uri)
+          : "";
 
       if (accessController.isFailOpen()) {
-        LOGGER.warn("Exception occurred! Access granted: {} {}", errLine, e);
+        LOGGER.warn("Exception occurred! Access granted: {}", errLine, e);
         return AccessResult.GRANTED;
       } else {
-        LOGGER.warn("Exception occurred! Access rejected: {} {}", errLine, e);
+        LOGGER.warn("Exception occurred! Access rejected: {}", errLine, e);
```
KaiSernLim, line 206: Does this remove the `Exception` being logged?

internal/venice-common/src/main/java/com/linkedin/venice/acl/handler/AbstractStoreAclHandler.java
```diff
@@ -127,26 +126,20 @@ protected boolean isAccessAlreadyApproved(ChannelHandlerContext ctx) {
...
-    if (VeniceSystemStoreUtils.isSystemStore(storeName)) {
-      return AccessResult.GRANTED;
-    }
-
-    String client = identityParser.parseIdentityFromCert(clientCert);
     try {
-      /**
-       * TODO: Consider making this the first check, so that we optimize for the hot path. If rejected, then we
-       *       could check whether the request is for a system store, METADATA, etc.
-       */
```
KaiSernLim, line 139: Hah.

Do we want to include a comment saying that this function should be as lightweight as possible, because it's on the hot path for retry storms or something like that?

KaiSernLim (dismissed):

internal/venice-common/src/main/java/com/linkedin/venice/acl/handler/AbstractStoreAclHandler.java
FelixGV: Added.

internal/venice-common/src/main/java/com/linkedin/venice/acl/handler/AbstractStoreAclHandler.java
FelixGV: Actually, it will now log the stacktrace, whereas previously it would only append the exception's name at the end (which seems unintended...).

internal/venice-common/src/main/java/com/linkedin/venice/acl/handler/AbstractStoreAclHandler.java
FelixGV: Not strictly necessary, but just to keep the style consistent, I added it systematically throughout this function, in my latest commit.

internal/venice-common/src/main/java/com/linkedin/venice/acl/handler/AbstractStoreAclHandler.java
```diff
@@ -126,27 +125,27 @@ protected boolean isAccessAlreadyApproved(ChannelHandlerContext ctx) {
...
       if (accessController.hasAccess(clientCert, storeName, method)) {
         return AccessResult.GRANTED;
       }
 
+      if (VeniceSystemStoreUtils.isSystemStore(storeName)) {
```
lluwm, line 140: Is `isSystemStore` cheaper than `hasAccess`? Shall we do the cheaper checks first for saving cpu cycles?

KaiSernLim: Most stores are not system stores, right? So wouldn't that harm the performance of regular stores? This is all very negligible stuff, IMO.

KaiSernLim (approved):

internal/venice-common/src/main/java/com/linkedin/venice/acl/handler/AbstractStoreAclHandler.java
FelixGV: `isSystemStore` is definitely cheaper than `hasAccess`, but we don’t care about the performance of the system store calls, and we do care about that of the user stores. So the main goal of this PR is not to have a very slim chance of doing little work, but rather to do less work in the common case.

lluwm: Gotcha. That makes sense

PR #1519, opened 2025-02-11
[server][dvc] Remove KafkaAdminClass references and merge KafkaSSLUtils into ApacheKafkaUtils

sushantmane:
## Remove KafkaAdminClass references and merge KafkaSSLUtils into ApacheKafkaUtils

- Removed all references to `KafkaAdminClass` and related configurations.
- Moved methods from `KafkaSSLUtils` to `ApacheKafkaUtils` for protocol validation.
- Introduced overload of `clipAndFilterNamespace` to support multi namespace clipping.
- Updated `VeniceClusterConfig`, `VeniceServerConfig`, and `AvroGenericDaVinciClient` to remove dependencies on `KafkaAdminClass`.
- Updated `KafkaStoreIngestionService` and `KafkaInputUtils` to use `ApacheKafkaUtils`.
- Removed `KAFKA_ADMIN_CLASS` constants from `ConfigKeys` and added `PUBSUB_CLIENT_CONFIG_PREFIX`.


internal/venice-client-common/src/main/java/com/linkedin/venice/utils/VeniceProperties.java
```diff
@@ -107,6 +107,50 @@ public VeniceProperties clipAndFilterNamespace(String nameSpace) {
...
+  /**
+   * This method looks for all properties that begin with either of the given
+   * namespaces. Once those properties are identified, it removes the matching
+   * namespace and returns the properties.
+   *
+   * This enables support for dynamic Kafka configurations. All Kafka
+   * properties can follow a convention of namespace, and the properties
+   * are extracted and supplied to the Kafka Producer/Consumer.
+   *
+   * @param nameSpace1 The first namespace to look for.
```
ZacAttack, line 119: Is the only difference between clipAndFilterNamespace is that this takes two namespaces?

Could we just do an end all and make this take a list?

internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
```diff
@@ -31,6 +30,27 @@ private ConfigKeys() {
...
+  /**
+   * Prefix for all Pub/Sub client-specific configurations.
+   *
+   * <p>This prefix ensures that all Pub/Sub-related configurations are passed to the
+   * respective Pub/Sub client adapters. Each adapter will then use its own prefix
+   * to extract and process the relevant properties.</p>
+   *
+   * <p>Configurations should follow this convention:</p>
+   * <pre>{@code
+   * PUBSUB_CLIENT_CONFIG_PREFIX + "PubSubClientSpecificConfigPrefix" + "actual.config.name"
```
ZacAttack, line 42: PubSubClientSpecificConfigPrefix ---> literally this?  Or would I put like, "kafka" ?

ZacAttack: and if it's "kafka", maybe put like "kafka" and/or "pulsar" or even "warpstream" as these are pubsub providers we've run Venice on in OS

ZacAttack (dismissed):
Overall this looks fine, just left a few questions.

internal/venice-client-common/src/main/java/com/linkedin/venice/utils/VeniceProperties.java
sushantmane: Yeah they are same but with diff args. 

>Could we just do an end all and make this take a list?

I like this idea. Will update to use `Set`

internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
sushantmane: `PubSubClientSpecificConfigPrefix` is a placeholder and acutally values will look like `kafka` or `pulsar` etc. Let me add `<>` to make it bit more clearer

internal/venice-client-common/src/main/java/com/linkedin/venice/utils/VeniceProperties.java
sushantmane: Done

internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
sushantmane: Done

sushantmane:
Thanks, @ZacAttack for the review! Addressed your comments.  Please take another look.

ZacAttack (approved):

PR #1520, opened 2025-02-11
[protocol] Add StoreMetaValue schema v28 to introduce largestUsedRTVersionNumber for RT version tracking

arjun4084346:
## Adding a field largestUsedRTVersionNumber in StoreMetaValue schema

When we delete and recreate a store, we do not want to use the same old real time topic name, because that might not have been fully cleaned up. We want to use a new name, and in order to do so we need to know the previous name that was ever used. This config will keep the largestUsedRTVersionNumber that was ever used and help in creating a new name whenever needed.

schema difference -

```
abora-mn3:StoreMetaValue abora$ diff v27/StoreMetaValue.avsc v28/StoreMetaValue.avsc 
34a35
>             {"name": "largestUsedRTVersionNumber", "type": "int", "default": 0, "doc": "The largest version number for the real time topic ever used before for this store."},
266a268
>                     {"name": "largestUsedRTVersionNumber", "type": "int", "default": 0},
```

Resolves #XXX


build.gradle
```diff
@@ -313,7 +313,7 @@ subprojects {
 
     doFirst {
       def versionOverrides = [
-//        project(':internal:venice-common').file('src/main/resources/avro/StoreVersionState/v5', PathValidation.DIRECTORY)
+        project(':internal:venice-common').file('src/main/resources/avro/StoreVersionState/v27', PathValidation.DIRECTORY),
```
sushantmane, line 316: Use this path instead: `internal/venice-common/src/main/resources/avro/StoreMetaValue/v28/StoreMetaValue.avsc`

sushantmane (approved):
LGTM.  When doing `Squash and merge`, please don't forget to cleanup the commit message and format it so that there are less than 100 chars per line. Thanks.

PR #1521, opened 2025-02-11
[router][server] ACL optimization

gaojieliu:
High-level idea:
1. For a keep-alive connection, the client cert will never change, so for the same store, it is useless to validate each request on the same connection.
2. In Server, there is an Acl Handler called `ServerAclHandler`, which is used to validate whether the connection is from Venice Router or not via static ACL. For each connection, Server will buffer the ACL check result in this attribute key: `SERVER_ACL_APPROVED_ATTRIBUTE_KEY`. And the ACL check result won't change during the lifetime of the Server instance.
3. In both Router and Server, there is a store-level ACL check, which can change during the lifetime of the Router/Server (ACL added/removed). The caching idea is a little different from #2, and it will maintain a cache map in the original connection, and for all the requests coming from this particular connection, it will check the acl check cache map first, and it will follow the previous result if the cache entry is not expired. If there is no such entry or the cache entry is expired, it will resort to the underlying access control to update the cache map. In theory, the acl check against the access controller will be minimized a lot.

New config:
acl.in.memory.cache.ttl.ms: 60000 (by default)

## How was this PR tested?
CI

### DIFF
--- a/internal/venice-common/src/main/java/com/linkedin/venice/acl/handler/AbstractStoreAclHandler.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/acl/handler/AbstractStoreAclHandler.java
@@ -7,17 +7,24 @@ import com.linkedin.venice.acl.AclException;
 import com.linkedin.venice.acl.DynamicAccessController;
 import com.linkedin.venice.authorization.IdentityParser;
 import com.linkedin.venice.common.VeniceSystemStoreUtils;
+import com.linkedin.venice.listener.ServerHandlerUtils;
 import com.linkedin.venice.meta.ReadOnlyStoreRepository;
 import com.linkedin.venice.meta.Store;
 import com.linkedin.venice.utils.NettyUtils;
+import com.linkedin.venice.utils.SystemTime;
+import com.linkedin.venice.utils.Time;
+import com.linkedin.venice.utils.concurrent.VeniceConcurrentHashMap;
+import io.netty.channel.Channel;
 import io.netty.channel.ChannelHandler;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.SimpleChannelInboundHandler;
 import io.netty.handler.codec.http.HttpRequest;
 import io.netty.handler.codec.http.HttpResponseStatus;
+import io.netty.util.AttributeKey;
 import io.netty.util.ReferenceCountUtil;
 import java.net.URI;
 import java.security.cert.X509Certificate;
+import java.util.Objects;
 import java.util.stream.Collectors;
 import javax.net.ssl.SSLPeerUnverifiedException;
 import org.apache.logging.log4j.LogManager;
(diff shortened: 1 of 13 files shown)


internal/venice-common/src/main/java/com/linkedin/venice/acl/handler/AbstractStoreAclHandler.java
```diff
@@ -89,9 +156,28 @@ public void channelRead0(ChannelHandlerContext ctx, HttpRequest req) throws SSLP
...
+        originalChannel.attr(STORE_ACL_CHECK_RESULT_ATTRIBUTE_KEY).get();
+    if (storeAclCache == null) {
+      originalChannel.attr(STORE_ACL_CHECK_RESULT_ATTRIBUTE_KEY).setIfAbsent(new VeniceConcurrentHashMap<>());
+      storeAclCache = originalChannel.attr(STORE_ACL_CHECK_RESULT_ATTRIBUTE_KEY).get();
+    }
+    AclKey aclKey = new AclKey(storeName, method);
+
+    AccessResult accessResult = storeAclCache.compute(aclKey, (ignored, value) -> {
+      long currentTimestamp = time.getMilliseconds();
+      if (value == null || currentTimestamp - value.timestamp > cacheTTLMs) {
```
FelixGV, line 170: So... if we set the `cacheTTLMs` to -1, then the code will always go into this branch which means the caching is effectively disabled... but it might be cleaner if there is a way to completely short-circuit the caching logic, rather than let it run in a way where every query evaluates to a cache miss...

internal/venice-common/src/main/java/com/linkedin/venice/acl/handler/AbstractStoreAclHandler.java
```diff
@@ -55,14 +118,18 @@ public AbstractStoreAclHandler(
...
+    if (isAccessAlreadyApproved(originalChannel)) {
       ReferenceCountUtil.retain(req);
       ctx.fireChannelRead(req);
       return;
     }
 
     String uri = req.uri();
-
     // Parse resource type and store name
     String[] requestParts = URI.create(uri).getPath().split("/");
```
FelixGV, line 134: Unrelated to this PR so feel free to ignore, but... the parsing of the URI into request parts happens in several handlers... it may be good to have some utility function that can cache it into the channel (though it would need to be on a per-request basis only, not longer-lived than that; we do a similar thing for when we extract the key count in the router, IIRC). Then we would do the request parsing just once in the whole request.

internal/venice-common/src/main/java/com/linkedin/venice/acl/handler/AbstractStoreAclHandler.java
gaojieliu: Sure, I will add a new config flag to completely disable this feature.

internal/venice-common/src/main/java/com/linkedin/venice/acl/handler/AbstractStoreAclHandler.java
```diff
@@ -29,7 +36,51 @@
  */
 @ChannelHandler.Sharable
 public abstract class AbstractStoreAclHandler<REQUEST_TYPE> extends SimpleChannelInboundHandler<HttpRequest> {
+  private static class AclKey {
+    String storeName;
+    String methodName;
```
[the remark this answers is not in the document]
gaojieliu: Sure, we can remove the method name for caching logic.

internal/venice-common/src/main/java/com/linkedin/venice/acl/handler/AbstractStoreAclHandler.java
gaojieliu: This will be a little tricky as we don't maintain a shared object across all the handlers until `RouterRequestHttpHandler`.
I will leave such optimization outside of this PR.

FelixGV (approved):
LGTM! Thanks!

PR #1522, opened 2025-02-11
[protocol] Add adminOperationProtocolVersion into proto schema AdminTopicMetadata

minhmo1620:
[protocol] Add adminOperationProtocolVersion into proto schema AdminTopicMetadata

This is the preparation for PR: #1418 

## Why?
Since we are introducing a new field in AdminTopicMetadata, we will need to update the API `getAdminTopicMetadata` to return the full metadata map on remote. To do that, we need to add this key into the response object. 

## How was this PR tested?
Only schema changed


sushantmane (approved):
LGTM. Let's add more context why we're adding this field in the commit message. And please don't forget to cleanup the commit message and format it so that there are less than 100 chars per line. Thanks.

huangminchn (approved):
Thanks a lot Minh!

PR #1525, opened 2025-02-12
[server] Allocated a dedicated thread pool for ingestion DB lookup

gaojieliu:
Based on the prod experiment, the SSD being used is not performing well when enabling parallel processing for AA/WC workload as the database lookup concurrency will be roughly same as the concurrency of parallel processing and we saw a higher lookup latency with high cpu wait.
This PR introduces a separated thread pool for ingestion database lookup (value and RMD) and the concurrency can be controlled separately from the parallel processing concurrency, and based on a canary test, a lower concurrency of DB lookup can reduce cpu wait while delivering a slightly better overall throughput.

New Config:
server.aa.wc.ingestion.storage.lookup.thread.pool.size: default 4

## How was this PR tested?
CI

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/config/VeniceServerConfig.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/config/VeniceServerConfig.java
@@ -54,6 +54,7 @@ import static com.linkedin.venice.ConfigKeys.PARTICIPANT_MESSAGE_CONSUMPTION_DEL
 import static com.linkedin.venice.ConfigKeys.PUBSUB_TOPIC_MANAGER_METADATA_FETCHER_CONSUMER_POOL_SIZE;
 import static com.linkedin.venice.ConfigKeys.PUBSUB_TOPIC_MANAGER_METADATA_FETCHER_THREAD_POOL_SIZE;
 import static com.linkedin.venice.ConfigKeys.ROUTER_PRINCIPAL_NAME;
+import static com.linkedin.venice.ConfigKeys.SERVER_AA_WC_INGESTION_STORAGE_LOOKUP_THREAD_POOL_SIZE;
 import static com.linkedin.venice.ConfigKeys.SERVER_AA_WC_LEADER_QUOTA_RECORDS_PER_SECOND;
 import static com.linkedin.venice.ConfigKeys.SERVER_AA_WC_WORKLOAD_PARALLEL_PROCESSING_ENABLED;
 import static com.linkedin.venice.ConfigKeys.SERVER_AA_WC_WORKLOAD_PARALLEL_PROCESSING_THREAD_POOL_SIZE;
@@ -587,6 +588,8 @@ public class VeniceServerConfig extends VeniceClusterConfig {
   private final boolean deleteUnassignedPartitionsOnStartup;
   private final int aclInMemoryCacheTTLMs;
 
+  private final int aaWCIngestionStorageLookupThreadPoolSize;
+
   public VeniceServerConfig(VeniceProperties serverProperties) throws ConfigurationException {
     this(serverProperties, Collections.emptyMap());
   }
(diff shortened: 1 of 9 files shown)


sushantmane:
I haven’t reviewed this PR, but I want to share an idea we might want to consider. IMO, we should have a single common thread pool for RocksDB reads, used for both the read and write paths. This would allow better concurrency control: reads for writes can be made to have lower priority than user reads. This way, we can manage read concurrency more effectively without impacting regular reads.

gaojieliu:
> I haven’t reviewed this PR, but I want to share an idea we might want to consider. IMO, we should have a single common thread pool for RocksDB reads, used for both the read and write paths. This would allow better concurrency control: reads for writes can be made to have lower priority than user reads. This way, we can manage read concurrency more effectively without impacting regular reads.

The main goal of this PR is not trying to solve the conflict between read and write, but just write only.
There is another initiative to address the priority issue, which is called dynamic throttling since the read path can be affected by the whole ingestion pipeline, not just the ingestion read.
Even without this enhancement, the read latency with batch processing is comparable or slightly better than other nodes, and this is mainly to address the high cpu wait issue.

sushantmane:
> The main goal of this PR is not trying to solve the conflict between read and write, but just write only. 


I understand that. However, these concerns are not mutually exclusive. We need to take a comprehensive look at the storage layer's IO impact. If we focus solely on the write path without considering the read path, the overall improvement may still be suboptimal.






These are just my thoughts—this isn't a blocker for this PR.

gaojieliu:
> > The main goal of this PR is not trying to solve the conflict between read and write, but just write only.
> 
> I understand that. However, these concerns are not mutually exclusive. We need to take a comprehensive look at the storage layer's IO impact. If we focus solely on the write path without considering the read path, the overall improvement may still be suboptimal.
> 
> These are just my thoughts—this isn't a blocker for this PR.

Certainly, but another important aspect is to separate of the concerns, and different layers needs to be relatively independent and there will be a coordination layer, which will orchestrate the read and write, which is the dynamic throttling I was referring to in the previous comment.

sushantmane:
> Certainly, but another important aspect is to separate of the concerns, and different layers needs to be relatively independent and there will be a coordination layer, which will orchestrate the read and write, which is the dynamic throttling I was referring to in the previous comment.


We do't need co-ordination layer for the idea I was referring to.

gaojieliu:
> > Certainly, but another important aspect is to separate of the concerns, and different layers needs to be relatively independent and there will be a coordination layer, which will orchestrate the read and write, which is the dynamic throttling I was referring to in the previous comment.
> 
> We do't need co-ordination layer for the idea I was referring to.

I think we do since the read/write path are not just competing IO resources, but all others, such as CPU, GC, network, and that is the holistic view we should look at instead of this narrow scope.
One example is that rebootstrapping can affect read perf even all the writes are full put.

sushantmane:
> Can you share how you going to control the priority of write-path read and regular read within single pool? My understanding is unless you control/throttle the issuer of the read (the consumer thread) you cannot directly control that right? And then if this is the case then it can also be achieved by separate pool then..


All read requests to RocksDB will be routed through the storage executor thread pool, which will be updated to use a priority-based queue for scheduling tasks. Requests from the read path will be assigned a higher priority than those triggered by ingestion, with additional safeguards to prevent ingestion reads from being indefinitely starved. This approach implicitly throttles writes, as ingestion will be blocked if queued writes are not completed in time.

sixpluszero:
> All read requests to RocksDB will be routed through the storage executor thread pool, which will be updated to use a priority-based queue for scheduling tasks. Requests from the read path will be assigned a higher priority than those triggered by ingestion, with additional safeguards to prevent ingestion reads from being indefinitely starved. This approach implicitly throttles writes, as ingestion will be blocked until queued reads are not finished.

It can be promising, but I have a concern of starving the write-path read though.  Maybe if we configure the thread pool large enough then we can avoid that.

sixpluszero (approved):

sushantmane:
> It can be promising, but I have a concern of starving the write-path read though. Maybe if we configure the thread pool large enough then we can avoid that.


Do you think priority + time-based throttling is not enough? Could you please elaborate?

PR #1530, opened 2025-02-13
[admin-tool] Modify the response of updateAdminTopicMetadata

minhmo1620:
[admin-tool] Modify the response of updateAdminTopicMetadata

Modify the response of updateAdminTopicMetadata to have more meaningful response and unify the response of APIs relating to `AdminTopicMetadata`.

Currently, there are 2 APIs related to `AdminTopicMetadata`: `getAdminTopicMetadata` and `updateAdminTopicMetadata`.
While `getAdminTopicMetadata` is returning metadata itself, `updateAdminTopicMetadata` only returns `clusterName` and `storeName`, even though we actually updating many metadata in this `updateAdminTopicMetadata` request.

To reflect better response, we will now return the metadata that we updating inside the response of `updateAdminTopicMetadata`.

This effort will also align with the upcoming API `updateAdminOperationProtocolVersion` in #1418, where we will return the admin operation version + cluster name inside the response.

In this PR, I make `executionId` as optional since `updateAdminOperationProtocolVersion` will also return `AdminTopicMetadataGrpcResponse` but execution id is not necessary inside the response. 

## How was this PR tested?
UTs

### DIFF
--- a/services/venice-controller/src/main/java/com/linkedin/venice/controller/server/ClusterAdminOpsRequestHandler.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/controller/server/ClusterAdminOpsRequestHandler.java
@@ -15,7 +15,6 @@ import com.linkedin.venice.protocols.controller.AdminTopicMetadataGrpcResponse;
 import com.linkedin.venice.protocols.controller.LastSuccessfulAdminCommandExecutionGrpcRequest;
 import com.linkedin.venice.protocols.controller.LastSuccessfulAdminCommandExecutionGrpcResponse;
 import com.linkedin.venice.protocols.controller.UpdateAdminTopicMetadataGrpcRequest;
-import com.linkedin.venice.protocols.controller.UpdateAdminTopicMetadataGrpcResponse;
 import com.linkedin.venice.utils.Pair;
 import java.util.Map;
 import java.util.Optional;
@@ -108,11 +107,14 @@ public class ClusterAdminOpsRequestHandler {
     return AdminTopicMetadataGrpcResponse.newBuilder().setMetadata(adminMetadataBuilder.build()).build();
   }
 
-  public UpdateAdminTopicMetadataGrpcResponse updateAdminTopicMetadata(UpdateAdminTopicMetadataGrpcRequest request) {
+  public AdminTopicMetadataGrpcResponse updateAdminTopicMetadata(UpdateAdminTopicMetadataGrpcRequest request) {
     AdminTopicGrpcMetadata metadata = request.getMetadata();
     String clusterName = metadata.getClusterName();
+    ControllerRequestParamValidator.validateAdminTopicMetadataRequest(clusterName);
+    if (!metadata.hasExecutionId()) {
+      throw new VeniceException("Execution id is required to update admin topic metadata");
+    }
     long executionId = metadata.getExecutionId();
-    ControllerRequestParamValidator.validateAdminCommandExecutionRequest(clusterName, executionId);
     String storeName = metadata.hasStoreName() ? metadata.getStoreName() : null;
     Long offset = metadata.hasOffset() ? metadata.getOffset() : null;
     Long upstreamOffset = metadata.hasUpstreamOffset() ? metadata.getUpstreamOffset() : null;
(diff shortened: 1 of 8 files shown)


services/venice-controller/src/main/java/com/linkedin/venice/controller/server/ClusterAdminOpsRequestHandler.java
```diff
@@ -108,11 +107,14 @@ public AdminTopicMetadataGrpcResponse getAdminTopicMetadata(AdminTopicMetadataGr
     return AdminTopicMetadataGrpcResponse.newBuilder().setMetadata(adminMetadataBuilder.build()).build();
   }
 
-  public UpdateAdminTopicMetadataGrpcResponse updateAdminTopicMetadata(UpdateAdminTopicMetadataGrpcRequest request) {
+  public AdminTopicMetadataGrpcResponse updateAdminTopicMetadata(UpdateAdminTopicMetadataGrpcRequest request) {
     AdminTopicGrpcMetadata metadata = request.getMetadata();
     String clusterName = metadata.getClusterName();
+    ControllerRequestParamValidator.validateAdminTopicMetadataRequest(clusterName);
+    if (!metadata.hasExecutionId()) {
+      throw new VeniceException("Execution id is required to update admin topic metadata");
```
sushantmane, line 118: We should move `metadata.hasExecutionId()` check inside `ControllerRequestParamValidator.validateAdminTopicMetadataRequest()` and use `IllegalArgumentException` to make sure the proper error code is sent to the clients

services/venice-controller/src/test/java/com/linkedin/venice/controller/grpc/server/ClusterAdminOpsGrpcServiceImplTest.java
```diff
@@ -148,18 +148,33 @@ public void testGetAdminTopicMetadataSuccess() {
...
+        .setOffset(100L)
+        .setUpstreamOffset(-1L);
+    AdminTopicMetadataGrpcResponse response =
+        AdminTopicMetadataGrpcResponse.newBuilder().setMetadata(adminTopicGrpcMetadataBuilder.build()).build();
     doReturn(response).when(requestHandler).updateAdminTopicMetadata(any(UpdateAdminTopicMetadataGrpcRequest.class));
     doReturn(true).when(accessManager).isAllowListUser(anyString(), any());
 
     UpdateAdminTopicMetadataGrpcRequest request = UpdateAdminTopicMetadataGrpcRequest.newBuilder()
-        .setMetadata(AdminTopicGrpcMetadata.newBuilder().setClusterName(TEST_CLUSTER).setExecutionId(EXECUTION_ID))
+        .setMetadata(
```
sushantmane, line 162: Why do we need to change/update request?

services/venice-controller/src/main/java/com/linkedin/venice/controller/server/ClusterAdminOpsRequestHandler.java
sushantmane: Or just use `ControllerRequestParamValidator.validateAdminCommandExecutionRequest(clusterName, executionId)`

sushantmane (commented):
Thanks, @minhmo1620! Left few comments. Let me know if you have any questions about them

services/venice-controller/src/main/java/com/linkedin/venice/controller/server/ClusterAdminOpsRequestHandler.java
minhmo1620: Take a closer look, we can use `ControllerRequestParamValidator.validateAdminCommandExecutionRequest(clusterName, executionId)` since if the executionId is not provided, the default value for getExecutionId will be 0, which will be caught in the validator.

services/venice-controller/src/test/java/com/linkedin/venice/controller/grpc/server/ClusterAdminOpsGrpcServiceImplTest.java
minhmo1620: I think it doesn't represent a valid request since we required the offsets to be provided if no store is provided.

```
    if (storeName != null && (offset != null || upstreamOffset != null)) {
      throw new VeniceException("Updating offsets is not allowed for store-level admin topic metadata");
    } else if (storeName == null && (offset == null || upstreamOffset == null)) {
      throw new VeniceException("Offsets must be provided to update cluster-level admin topic metadata");
    }
```

services/venice-controller/src/main/java/com/linkedin/venice/controller/server/ClusterAdminOpsRequestHandler.java
minhmo1620: I have updated the code to use `ControllerRequestParamValidator.validateAdminCommandExecutionRequest(clusterName, executionId)`. Thanks for your suggestion.

services/venice-controller/src/test/java/com/linkedin/venice/controller/grpc/server/ClusterAdminOpsGrpcServiceImplTest.java
sushantmane: If that's the case existing test should be failing with an exception, no?

services/venice-controller/src/main/java/com/linkedin/venice/controller/server/ControllerRequestParamValidator.java
```diff
@@ -39,7 +39,7 @@ public static void validateClusterStoreInfo(ClusterStoreGrpcInfo rpcContext) {
 
   public static void validateAdminCommandExecutionRequest(String clusterName, long executionId) {
     if (StringUtils.isBlank(clusterName)) {
-      throw new IllegalArgumentException("Cluster name is required for getting admin command execution status");
+      throw new IllegalArgumentException("Cluster name is required for admin command execution");
```
sushantmane, line 42: I like original message better

services/venice-controller/src/test/java/com/linkedin/venice/controller/grpc/server/ClusterAdminOpsGrpcServiceImplTest.java
minhmo1620: The test is not failing since the test is not to test the logic since the response is mocked, it is testing grpc end. Aligned offline to keep this.

sushantmane (approved):
LGTM. Thanks, @minhmo1620!

PR #1531, opened 2025-02-13
[dvc] Fix logging in DVRT

kvargha:
## [dvc] Fix logging in DVRT

The logging doesn't match up with the logic, so I swapped them.

## How was this PR tested?

CI


KaiSernLim (approved):

sushantmane (commented):
nit: for logging common pattern in our code is  var: {}  instead of var = {}

KaiSernLim (approved):

sushantmane (approved):
LGTM. Thanks

PR #1532, opened 2025-02-13
[common][router] Cache dimensions for otel 

m-nagarajan:
## Summary
Problem:
Currently, every `metric.record()` call creates a new `Attributes` object with all the dimensions needed for that metric. As it also happens on the happy path and as `Attributes` creation does `mergesort` and `deduplication`, it will lead to high rate of object churn, potentially affecting GC and performance issues. 

Solution:
This PR in its original form dealt with constructing a global cache with the key to the cache being the concatenation of all possible dimension values which proved to be inefficient as the new `Attributes` creation seemed to be faster than constructing the `String` key to the cache. So we decided to switch up the design to handle different cases separately in multiple sub classes of `MetricEntityState` and use`EnumMap` where ever possible (but with multiple levels based on how many dynamic dimensions are needed for that metric). Lookup in `EnumMap` happens to be a lot faster than `HashMap` with keys based on `Object`/`String`, especially trying to compute a composite `String` key on the fly.

There are multiple cases that needs to be handled:
1. No dynamic dimensions: it should just use a precreated `Attributes` object which is based on the base dimensions which will always be the same for this instance. 
2. Some dynamic dimensions with bound values which can be enumerated: we can consider precreating the `Attributes` for all combinations and fetch it during runtime. But that might be a huge memory overhead, especially if many of the combinations will never be used. so this PR relies on lazy initialization of `Attributes` for each set of dimensions in `MetricEntityState` whenever the first time record() is called with those dimensions resulting in creating and caching only the needed combinations. Note that, at least in the router context, `storeName` and `clusterName` which are unbounded strings falls under the base dimensions with how the existing stats objects are structured, so they remain the same for each `RouterHttpRequestStats`, so it doesn't fall under case 3.
3. Some dynamic dimensions which are unbounded: For instance some values that are `String` like `upstream_service` which identifies the caller. This PR doesn't have a sub class for that, but it fits in naturally with one another layer of `HashMap` on top of one of the case 2.

This PR creates multiple subclasses to MetricEntityState like below
1. `MetricEntityStateBase` (Base dimensions + 0 additonal dimensions)
2. `MetricEntityStateOneEnum` (Base dimensions +  1 additonal dimension which can be an `Enum`)
3. `MetricEntityStateTwoEnums` (Base dimensions + 2 additonal dimensions which can be `Enums`)
4. `MetricEntityStateThreeEnums` (Base dimensions + 3 additonal dimensions which can be `Enums`)

Note that the number of dimensions mentioned here are on top of some of the common/base dimensions which this instance would be associated with. For example, an instance of `RouterHttpRequestStats#healthyRequestMetric` will always have the same storeName, clusterName, and RequestType which will be considered base dimensions and it will have 3 additional dimenions like `HttpResponseStatusCode`, `HttpResponseStatusCodeCategory`, `VeniceResponseStatusCodeCategory`. This metric will now use `MetricEntityStateThreeEnums`.

image
 
## How was this PR tested?
GH CI 


internal/venice-client-common/src/main/java/com/linkedin/venice/stats/VeniceOpenTelemetryDimensionsCache.java
```diff
@@ -0,0 +1,143 @@
...
+   * extending {@link AbstractVeniceHttpStats} where the common dimensions can be reused
+   */
+  private final Attributes baseMetricDimensions;
+  /** part of cache key from the base dimensions: To reuse */
+  private final Set<VeniceMetricsDimensions> baseMetricDimensionsSet;
+  /** the keys in the base dimensions to ignore when creating the cache key from all other dimensions */
+  private final String baseMetricDimensionsKey;
+
+  /** used to pass in the dimension and its values to create {@link Attributes} and avoid creating temp maps/arrays */
+  private static final ThreadLocal<Map<VeniceMetricsDimensions, String>> threadLocalReusableDimensionsMap =
```
lluwm, line 46: Is this correct that, ThreadLocal requires every router thread servicing a client query have to create a copy of this map? If it is the case, do we know the upper limit of how large this number could be?

services/venice-router/src/main/java/com/linkedin/venice/router/stats/RouterHttpRequestStats.java
```diff
@@ -494,21 +510,19 @@ public void recordErrorRetryCount() {
   public void recordRetryTriggeredSensorOtel(RequestRetryType retryType) {
     Attributes dimensions = null;
     if (emitOpenTelemetryMetrics) {
-      dimensions = Attributes.builder()
-          .putAll(commonMetricDimensions)
-          .put(getDimensionName(VENICE_REQUEST_RETRY_TYPE), retryType.getRetryType())
-          .build();
+      Map<VeniceMetricsDimensions, String> reusableDimensionsMap = getThreadLocalReusableDimensionsMap();
+      reusableDimensionsMap.put(VENICE_REQUEST_RETRY_TYPE, retryType.getRetryType());
+      dimensions = otelDimensionsCache.checkCacheAndGetDimensions(retryCountMetric.getMetricEntity());
```
FelixGV, line 515: This logic looks extremely complex to me, both in terms of my ability to understand it, but also in terms of time complexity (so many map operations, string building, etc...).

Since the `MetricEntityState retryCountMetric` is a private property of this class, and this class has the per-store scope that we're interested in, I don't understand, why not cache the `Attributes dimensions` inside of the `MetricEntityState` object itself? Then this whole function can become simply `retryCountMetric.record(1);` and we let the dimension-passing be completely handled on the inside, by simply taking it from some `private final Attributes dimensions` property of the `MetricEntityState`... No map lookups, no string building, no threadlocal state... all of that disappears completely. And by hiding the OTel complexity in this way, we should greatly simplify the migration from Tehuti to OTel.

WDYT?

FelixGV (commented):
Left a high level comment... Please take a look and LMK what you think.

services/venice-router/src/main/java/com/linkedin/venice/router/stats/RouterHttpRequestStats.java
m-nagarajan: Thanks @FelixGV  for the comment. Packing the cache inside `MetricEntityState` by relying on the existing class with the per-store scope sounds good as well, but it doesn't eliminate the string building (for cache key) completely as each state can have multiple `Attributes` keyed by one/more of the varying dimensions. For instance,
1. `retryCountMetric`  can have multiple Attributes based on values of `RequestRetryType`  
2. `healthyRequestMetric`  can have multiple Attributes based on values of `HttpResponseStatus`  and `HttpResponseStatusCodeCategory`.
Also, when we move away from `tehuti` we will regroup some of the current `MetricEntityState`s into 1 (for instance: `healthy_request`, `unhealthy_request`, `tardy_request`, etc to just 1) which is going to increase this cardinality further. We can continue keeping each of these combinations to be a separate `MetricEntityState`, but I feel like its too much denormalizing and we have to further denormalize it to keep things 1:1. In other alternative routes, we will need some form of key to access the cache (`global` or per `RouterHttpRequestStats` or per `MetricEntityState` ). The content of the key gets smaller as we move further away from global cache, the easier being just 1 dimension like in `retryCountMetric` where a cache per `RouterHttpRequestStats` like below would work. 
```
VeniceConcurrentHashMap<RequestRetryType, Attributes> otelDimensionCacheForRequestRetryType;
```
If not, if we have combinations of two or more dimensions as keys, we need to construct some form of key for the cache. What do you think?

FelixGV: I don't see a problem with the denormalization... We can make each `MetricEntityState` instance carry a single `Attributes` property, but have multiple `MetricEntityState` hold a reference to the same `otelMetric`. With this design, the complete attribute specification is done upfront, in each `MetricEntityState`, and we have (IMHO) a clear decoupling of the underlying metric stack. IOW, different `MetricEntityState` instances can hold references to a single `otelMetric` while each also holds a reference to a different `tehutiSensor` (since in Tehuti the dimensions are expressed as part of the sensor's name).

The best cache is no cache at all, just an object pointer. That is the fastest. I may be missing context about cases where this will not be good enough and where we will still need a cache, but even if such cases do exist, we still ought to make the simple cases as fast as possible.

gaojieliu: I like the data-driven approach.
There is a trade-off regarding performance vs maintainability.
The complex logic is not always good unless it can prove much better performance as complex logic incurs more maintenance overhead in the future and more error prone.

FelixGV: So... thinking about it some more, I'd like to propose the following:

1. Let's make OTel dimensions fully encapsulated inside of `MetricEntityState` and its subclasses (see next point). No other code path should ever instantiate an OTel `Attributes`. This will help make the migration easier.
2. Let's make `MES` abstract, with an API intended for use (but NOT for extension) by subclasses: `protected final void record(value, attributes)`.
3. Let's provide a few subclass options for common cases. For example, one subclass will have just one immutable set of dimensions, stored as a `private final Attributes` class prop, and expose a `public void record(value)` function which internally passes the class prop attributes into the parent's protected function. Another subclass could take an enum as constructor param and instantiate an `EnumMap` with `Attributes` value (an `EnumMap` lookup is cheaper than a Map lookup, and it's inherently threadsafe in the sense that the backing array will never resize); a `public void record(value, enum)` function would perform the `EnumMap` lookup and pass the attributes into the parent's protected function. We can also do other subclasses as needed (e.g. based on 2 enums, 3 enums...).
4. No thread local state.
5. No map lookups (besides `EnumMap`).
6. No instantiation (of strings or anything else).

IMHO, such an approach will be simple to reason about and faster than both the current impl and the one in this PR.

internal/venice-client-common/src/main/java/com/linkedin/venice/stats/metrics/MetricEntityStateOneEnum.java
```diff
@@ -0,0 +1,143 @@
...
+        enumTypeClass);
+  }
+
+  private void createAttributesEnumMap(
+      MetricEntity metricEntity,
+      VeniceOpenTelemetryMetricsRepository otelRepository,
+      Map<VeniceMetricsDimensions, String> baseDimensionsMap) {
+    Map<VeniceMetricsDimensions, String> additionalDimensionsMap = new HashMap<>();
+    for (E enumValue: enumTypeClass.getEnumConstants()) {
+      additionalDimensionsMap.put(enumValue.getDimensionName(), enumValue.getDimensionValue());
```
lluwm, line 91: Is it the case that all `enumValue` share the same `getDimensionName()` and because of that, in fact, later `enumValue.getDimensionName()` in iteration always overwrites the earlier ones?

internal/venice-client-common/src/main/java/com/linkedin/venice/stats/metrics/MetricEntityStateOneEnum.java
```diff
@@ -0,0 +1,143 @@
...
+      attributesEnumMap
+          .put(enumValue, otelRepository.createAttributes(metricEntity, baseDimensionsMap, additionalDimensionsMap));
+    }
+    if (attributesEnumMap.isEmpty()) {
+      throw new VeniceException(
+          "The dimensions map is empty. Please check the enum types and ensure they are properly defined.");
+    }
```
lluwm, line 98: Nit, can we do the sanity check (e.g., based on the size of `enumTypeClass.getEnumConstants()`) before the for-loop? Would that be more intuitive?

internal/venice-client-common/src/main/java/com/linkedin/venice/stats/metrics/MetricEntityStateThreeEnums.java
```diff
@@ -0,0 +1,195 @@
...
+      if (mapE3 != null) {
+        attributes = mapE3.get(key3);
+      }
+    }
+
+    if (attributes == null) {
+      throw new IllegalArgumentException(
+          "No dimensions found for keys: " + key1 + "," + key2 + "," + key3 + " for metric Entity: "
+              + getMetricEntity().getMetricName());
+    }
```
lluwm, line 166: Seems that we don't perform this check for `MetricEntityStateOneEnum`, but here. Can we make them consistent?

internal/venice-client-common/src/main/java/com/linkedin/venice/stats/metrics/MetricEntityState.java
```diff
@@ -102,16 +111,65 @@ void recordTehutiMetric(double value) {
...
-  public void record(long value, Attributes otelDimensions) {
-    recordOtelMetric(value, otelDimensions);
+  protected final void record(long value, Attributes attributes) {
+    recordOtelMetric(value, attributes);
     recordTehutiMetric(value);
   }
 
-  public void record(double value, Attributes otelDimensions) {
-    recordOtelMetric(value, otelDimensions);
+  protected final void record(double value, Attributes attributes) {
```
lluwm, line 119: Nit, can we make these two generic? We have such separate calls in almost every subclass of MetricEntityState. It would be great if we can generalize them and save many similar codes.
```
record(long, Attributes attributes)
record(double, Attributes attributes)
```

services/venice-router/src/main/java/com/linkedin/venice/router/stats/RouterHttpRequestStats.java
```diff
@@ -415,38 +468,40 @@ public void recordIncomingRequest() {
...
-  }
-
   public void recordHealthyRequest(Double latency, HttpResponseStatus responseStatus, int keyNum) {
-    Attributes dimensions = getRequestMetricDimensions(responseStatus, VeniceResponseStatusCategory.SUCCESS);
-    healthyRequestMetric.record(1, dimensions);
-    keyCountMetric.record(keyNum, dimensions);
+    HttpResponseStatusCode httpResponseStatusCode = transformHttpResponseStatusToHttpResponseStatusCode(responseStatus);
+    HttpResponseStatusCodeCategory httpResponseStatusCodeCategory =
+        getVeniceHttpResponseStatusCodeCategory(responseStatus);
+    VeniceResponseStatusCategory veniceResponseStatusCategory = VeniceResponseStatusCategory.SUCCESS;
```
lluwm, line 475: The var name `veniceResponseStatusCategory` is too generic. Let's either rename it to just `success` or use `VeniceResponseStatusCategory.SUCCESS` directly in later function calls.

services/venice-router/src/main/java/com/linkedin/venice/router/stats/RouterHttpRequestStats.java
```diff
@@ -415,38 +468,40 @@ public void recordIncomingRequest() {
...
     }
   }
 
   public void recordUnhealthyRequest(HttpResponseStatus responseStatus) {
-    Attributes dimensions = getRequestMetricDimensions(responseStatus, VeniceResponseStatusCategory.FAIL);
-    unhealthyRequestMetric.record(1, dimensions);
+    HttpResponseStatusCode httpResponseStatusCode = transformHttpResponseStatusToHttpResponseStatusCode(responseStatus);
+    HttpResponseStatusCodeCategory httpResponseStatusCodeCategory =
+        getVeniceHttpResponseStatusCodeCategory(responseStatus);
+    VeniceResponseStatusCategory veniceResponseStatusCategory = VeniceResponseStatusCategory.FAIL;
```
lluwm, line 490: Ditto and some places later.

internal/venice-client-common/src/main/java/com/linkedin/venice/stats/VeniceOpenTelemetryMetricsRepository.java
```diff
@@ -221,14 +228,61 @@ public Object createInstrument(MetricEntity metricEntity) {
...
+      attributesBuilder.put(getDimensionName(dimension), dimensionValue);
+    }
+
+    // add additional dimensions
+    for (Map.Entry<VeniceMetricsDimensions, String> entry: additionalDimensionsMap.entrySet()) {
+      VeniceMetricsDimensions dimension = entry.getKey();
+      String dimensionValue = entry.getValue();
+      validateDimensionValuesAndThrow(metricEntity, dimension, dimensionValue);
+      attributesBuilder.put(getDimensionName(dimension), dimensionValue);
+    }
```
lluwm, line 269: Nit, some duplicate code with the above block. Can we extract them into a small helper function?

lluwm (commented):
@m-nagarajan, thanks for this change. Left some comments and it makes sense to me overall.

services/venice-router/src/main/java/com/linkedin/venice/router/stats/RouterHttpRequestStats.java
m-nagarajan: These also had some duplicate code. Extracted it to a common method in the last diff.

internal/venice-client-common/src/main/java/com/linkedin/venice/stats/metrics/MetricEntityState.java
lluwm: No, not really. We can go with it for now.

internal/venice-client-common/src/main/java/com/linkedin/venice/stats/metrics/MetricEntityStateTwoEnums.java
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    EnumMap<E1, EnumMap<E2, Attributes>> attributesEnumMap = new EnumMap<>(enumTypeClass1);
+    Map<VeniceMetricsDimensions, String> additionalDimensionsMap = new HashMap<>();
+    for (E1 enumConst1: enumTypeClass1.getEnumConstants()) {
+      additionalDimensionsMap.put(enumConst1.getDimensionName(), enumConst1.getDimensionValue());
+      EnumMap<E2, Attributes> mapE2 = new EnumMap<>(enumTypeClass2);
+      attributesEnumMap.put(enumConst1, mapE2);
+      for (E2 enumConst2: enumTypeClass2.getEnumConstants()) {
+        additionalDimensionsMap.put(enumConst2.getDimensionName(), enumConst2.getDimensionValue());
```
FelixGV, line 116: Is it possible that `E1` and `E2` have dimension names that clash? What would happen then? Can you write a unit test for this scenario, if we don't already have one?

internal/venice-client-common/src/main/java/com/linkedin/venice/stats/metrics/MetricEntityStateTwoEnums.java
```diff
@@ -0,0 +1,179 @@
...
+      throw new IllegalArgumentException(
+          "The key for otel dimension cannot be null for metric Entity: " + getMetricEntity().getMetricName());
+    }
+    if (!enumTypeClass1.isInstance(key1) || !enumTypeClass2.isInstance(key2)) {
+      // defensive check: This can only happen if the instance is declared without the explicit types
+      // and passed in wrong args
+      throw new IllegalArgumentException(
+          "The key for otel dimension is not of the correct type: " + key1.getClass() + "," + key2.getClass()
+              + " for metric Entity: " + getMetricEntity().getMetricName());
+    }
```
FelixGV, line 138: What does it mean that "the instance is declared without the explicit types"? Do we need to care?

internal/venice-client-common/src/main/java/com/linkedin/venice/stats/metrics/MetricEntityStateTwoEnums.java
```diff
@@ -0,0 +1,179 @@
...
+import java.util.EnumMap;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+
+/**
+ * Similar to {@link MetricEntityStateOneEnum} but with two dynamic dimensions and 2 level EnumMap
+ */
+public class MetricEntityStateTwoEnums<E1 extends Enum<E1> & VeniceDimensionInterface, E2 extends Enum<E2> & VeniceDimensionInterface>
```
FelixGV, line 20: That's interesting! I had never seen `&` used in an `extends` clause before...!

internal/venice-client-common/src/main/java/com/linkedin/venice/stats/metrics/MetricEntityStateThreeEnums.java
```diff
@@ -0,0 +1,198 @@
...
+        EnumMap<E3, Attributes> mapE3 = new EnumMap<>(enumTypeClass3);
+        mapE2.put(enumConst2, mapE3);
+        for (E3 enumConst3: enumTypeClass3.getEnumConstants()) {
+          additionalDimensionsMap.put(enumConst3.getDimensionName(), enumConst3.getDimensionValue());
+          mapE3.put(
+              enumConst3,
+              otelRepository.createAttributes(metricEntity, baseDimensionsMap, additionalDimensionsMap));
+        }
+      }
+    }
```
FelixGV, line 136: The algo here looks a bit confusing to me... the `additionalDimensionsMap` gets populated in three different places, and so as we keep iterating through these nested loops, the map will get more and more filled up. But that means that the content of this map will be different each time we call `otelRepository.createAttributes` in the most inner loop. This seems to indicate one of two things...

1. `otelRepository.createAttributes` will be called with inconsistent inputs and that would be a bug, OR
2. There is only a subset of the map elements which matter at any point in time, and somehow `otelRepository.createAttributes` knows which of these elements to look for, and therefore the extra baggage that gets accumulated as we iterate is harmless (with perhaps the sole purpose of confusing me personally 😅 ...). If that is the correct interpretation, then it seems that the map could be re-instantiated (or cleared) as the iteration progresses and the end result would still be correct.

Which of these interpretations is the correct one? (Or is there another?)

FelixGV (commented):
Thanks Manoj for adopting this suggestion! The new benchmark results look great! I left a few questions, which may or may not reflect problems (or perhaps only my own misunderstanding). Please take a look, thanks!

internal/venice-client-common/src/main/java/com/linkedin/venice/stats/metrics/MetricEntityStateTwoEnums.java
m-nagarajan: Thanks for the catch. Added some extra checks to detect this clashes and tests for the same.

internal/venice-client-common/src/main/java/com/linkedin/venice/stats/metrics/MetricEntityStateTwoEnums.java
m-nagarajan: This is where I ended up to have `getDimensionName()` and `getDimensionValue()` on every enums via implemeting `VeniceDimensionInterface` and to be able to call these functions inside this class on the enum classes passed in 😄

m-nagarajan (commented):
Thanks @FelixGV for the review. Made some changes to address your comments and answered some. Please let me know what you think.

internal/venice-client-common/src/main/java/com/linkedin/venice/stats/metrics/MetricEntityStateTwoEnums.java
FelixGV: So what would be the runtime error for case 2? `ClassCastException`?

internal/venice-client-common/src/main/java/com/linkedin/venice/stats/metrics/MetricEntityStateThreeEnums.java
FelixGV: Hmm... I see... so the important point is that `getDimensionName()` is expected to return the same value no matter which enum value we are dealing with. I had not understood that, and just assumed that someone implementing `VeniceDimensionInterface` could decide to return different values for each enum value. And indeed it seems that they can, since the class design does not prevent it... The expectation here would need to be enforced by convention, or maybe at best by unit tests. So then let me ask you this... why does the `additionalDimensionsMap` use a `String` as its key? Why not, for example, use a `Class` instead? The following would seem to indicate the intent much more clearly, IMHO:

`additionalDimensionsMap.put(enumConst2.getClass(), enumConst2.getDimensionValue());`

And it would not be subject to abuse/breakage by implementations of the interface which fail to comply with the expected convention.

FelixGV: Alternatively... get rid of the map entirely, and use a vararg in `createAttributes`, e.g., something like:

```
  public Attributes createAttributes(
      MetricEntity metricEntity,
      Map<VeniceMetricsDimensions, String> commonDimensionsMap,
      VeniceMetricsDimensions... additionalDimensions) {
    AttributesBuilder attributesBuilder = Attributes.builder();

    // add common dimensions
    commonDimensionsMap.forEach(
        (key, value) -> validateDimensionValuesAndBuildAttributes(metricEntity, key, value, attributesBuilder));

    // add additional dimensions
    for (VeniceMetricsDimensions dim: additionalDimensions) {
      validateDimensionValuesAndBuildAttributes(metricEntity, dim.getDimensionName(), dim.getDimensionValue(), attributesBuilder);
    }

    // add custom dimensions passed in by the user
    getMetricsConfig().getOtelCustomDimensionsMap().forEach(attributesBuilder::put);

    return attributesBuilder.build();
  }
```

Why carry the map at all, when instances of the enum itself is enough? The map will only be used as an iterable collection anyway, without any lookups performed on it, which indicates that there may be no need for a map at all.

Then the code calling this would look like:

```
    EnumMap<E1, EnumMap<E2, EnumMap<E3, Attributes>>> attributesEnumMap = new EnumMap<>(enumTypeClass1);
    for (E1 enumConst1: enumTypeClass1.getEnumConstants()) {
      EnumMap<E2, EnumMap<E3, Attributes>> mapE2 = new EnumMap<>(enumTypeClass2);
      attributesEnumMap.put(enumConst1, mapE2);
      for (E2 enumConst2: enumTypeClass2.getEnumConstants()) {
        EnumMap<E3, Attributes> mapE3 = new EnumMap<>(enumTypeClass3);
        mapE2.put(enumConst2, mapE3);
        for (E3 enumConst3: enumTypeClass3.getEnumConstants()) {
          mapE3.put(
              enumConst3,
              otelRepository.createAttributes(metricEntity, baseDimensionsMap, enumConst1, enumConst2, enumConst3));
        }
      }
    }
```

FelixGV (dismissed):
Thanks a lot Manoj. Looks really good now! LFG 🚀 🚀 🚀 !!!

internal/venice-client-common/src/main/java/com/linkedin/venice/stats/dimensions/HttpResponseStatusCode.java
```diff
@@ -0,0 +1,89 @@
...
+
+import io.netty.handler.codec.http.HttpResponseStatus;
+import java.util.HashMap;
+import java.util.Map;
+
+
+/**
+ * This enum is used to map the HttpResponseStatus to an Enum implementing VeniceDimensionInterface.
+ */
+public enum HttpResponseStatusCode implements VeniceDimensionInterface {
```
gaojieliu, line 11: Venice normally only uses a very small subset of http response codes, since we define a full list of response code as Netty, I wondered whether these dimensions will be pre-created or not even we don't really use them. If yes, I recommend to only create the ones being used by Venice to reduce the total number of dimensions to reduce cost and make it easy to navigate via UI.

gaojieliu: If these enum definitions are only placeholders, that would be fine

internal/venice-client-common/src/main/java/com/linkedin/venice/stats/metrics/MetricEntityStateThreeEnums.java
```diff
@@ -0,0 +1,195 @@
...
+import java.util.Collections;
+import java.util.EnumMap;
+import java.util.List;
+import java.util.Map;
+
+
+/**
+ * Similar to {@link MetricEntityStateOneEnum} but with three dynamic dimensions and 3 level EnumMap
+ */
+public class MetricEntityStateThreeEnums<E1 extends Enum<E1> & VeniceDimensionInterface, E2 extends Enum<E2> & VeniceDimensionInterface, E3 extends Enum<E3> & VeniceDimensionInterface>
```
gaojieliu, line 19: I wondered whether we can make it generic or not.
So far, we define `MetricEntityState` implementation with 1/2/3 enums and if we need to support 4 enums, we need to create one new class, but the logic is roughly the same.

I am thinking the following generic implementation:
```
public class MetricEntityStateWithEnums<E extends Enum<E> & VeniceDimensionInterface>  extends MetricEntityState {

private MetricEntityStateThreeEnums(
      MetricEntity metricEntity,
      VeniceOpenTelemetryMetricsRepository otelRepository,
      Map<VeniceMetricsDimensions, String> baseDimensionsMap,
      List<Class<E>> enumTypeClasses) {
      ...
      }
}
```

Inside this generic class, it will re-order the enum class by class names, so that it can return the correct attribute cache entry regardless of the order of the classes passed.
Even we want to have a dedicate definition for 1/2/3 enums, these implementations can easily just extend this class and pass a class list with 1/2/3 enums.
I haven't fully validated this idea, but it seems doable.
WDYT?

gaojieliu (commented):
Left some questions and please help validate whether these comments are valid or not.

internal/venice-client-common/src/main/java/com/linkedin/venice/stats/dimensions/HttpResponseStatusCode.java
m-nagarajan: Its some where inbetween. 
1. We will precreate the dimensions for all these codes (which might be heavy in some cases) 
2. but only the dimensions associated with whatever codes emitted will be used, so the UI will be clean with only emitted status.

I was thinking about the 1st problem and it needs some dynamic creation and caching (which is very easy to add on top of this existing code and make it configurable) or reduce the number of the values in this enums. Below are the last 6 months worth of status in routers. What do you think?
image

FelixGV: The current implementation (in this PR) will pre-create everything, which results in an amount of heap state proportional to the cartesian product of all the enums passed into the classes. If state size becomes an issue, I think this can be optimized without too much trouble by introducing some lazy init behavior, either always or as an option. I think for now we can shelf this and wait until we have more concrete data points to evaluate if the state size is an issue or not, but that's just my own opinion, and feel free to disagree of course.

internal/venice-client-common/src/main/java/com/linkedin/venice/stats/metrics/MetricEntityStateThreeEnums.java
FelixGV: IDK if your example code works or not. IDK if we would be allowed to pass different types of `E` in that list. But... maybe that param need not even be a generic, maybe it can be hardcoded to `List<Class< VeniceDimensionInterface>>` or similar. We can look into refactoring with an eye towards maximizing code re-use, but I think the current PR is probably ok as is for now.

internal/venice-client-common/src/main/java/com/linkedin/venice/stats/dimensions/HttpResponseStatusCode.java
gaojieliu: This is an additional dimension on top of the base dimensions, right?
Base dimensions must include store name, which means the impact can be amplified by the total number of stores (at least).
Do you know how much value it can bring to keep so many unused codes?
Shall we just place the known-to-be-useful codes here and leave others as unknowns?

gaojieliu: To me, we should start with a small set and gradually increase it if we don't want to build the lazy solution now..

internal/venice-client-common/src/main/java/com/linkedin/venice/stats/metrics/MetricEntityStateThreeEnums.java
gaojieliu: This is not a blocker per se, but mainly improve the maintainability of the code since I can duplicate codes across 1/2/3 enum implementations.

internal/venice-client-common/src/main/java/com/linkedin/venice/stats/metrics/MetricEntityStateTwoEnums.java
FelixGV: I guess I would suggest the following:

Look up in the `EnumMap` directly. Then do the `if (attributes == null)` check you currently have at the end. Only if that is true, then do the other checks (null key, wrong class) in order to provide the clearest possible error message.

There is no point to pre-emptively check all of these things which the `EnumMap` already checks on your behalf later. In general, we want to optimize for the correct path, then add extra overhead only in the error handling path.

internal/venice-client-common/src/main/java/com/linkedin/venice/stats/dimensions/HttpResponseStatusCode.java
FelixGV: Lazy init should be easy to build using `computeIfAbsent`. In terms of the thread safety of `EnumMap`, my understanding is the following:

- `EnumMap` is NOT threadsafe, but in practice I don't think it matters.
- `EnumMap` should never resize its internal array, like a regular map would, so there is no chance of hitting `ConcurrentModificationException`.
- When using `computeIfAbsent`, there is a chance that the passed in function gets executed more than once, rather than once atomically only if the key is missing. But in our use case, the end result is idempotent, so it shouldn't matter. The result of concurrently calling `computeIfAbsent` would just be that some garbage is generated transiently, which is an effect that would go away very quickly.

Given the above, I would recommend NOT using any synchronization and just accepting the transient increase in garbage in case of concurrent usage.

It would be good if someone else could read the `EnumMap` impl to confirm my assumptions above.

m-nagarajan: I looked at the `EnumMap` implementation as I was already planning to add this lazy initialization and I had the same understanding. Originally, I was thinking that we will need either a `synchronized` block (double checked locking) or something more granular within the overridden `computeIfAbsent`, but looks like we don't have to do all that because of the idempotent nature.

internal/venice-client-common/src/main/java/com/linkedin/venice/stats/metrics/MetricEntityStateThreeEnums.java
m-nagarajan: Thanks Gaojie for bringing this up. Agree with the concern, although I would like to park it for now to be a next refactor to fix the perf concerns we had and the memory concerns you brought up so we can start with the other components. In parallel I will look to see how we can refactor this to remove some duplicate code as I agree maintainability is as important as well. WDYT?

internal/venice-client-common/src/main/java/com/linkedin/venice/stats/dimensions/HttpResponseStatusCode.java
m-nagarajan: Made some changes around this suggestion.
Added 2 configs that basically enables 3 modes of optimization for `Attributes`.
1. pre-creation (`OTEL_VENICE_PRE_CREATE_ATTRIBUTES`): All `Attributes` are precreated.
- if this config is enabled and 
- dimensions are bound.
2. lazy initialization (`OTEL_VENICE_LAZY_INITIALIZE_ATTRIBUTES`):  `Attributes` are created when needed for the first time and then resused.
- If this config is enabled and
- pre-creation config is disabled or if the dimension values are not bound (in the context of future use cases)
3. no optimization: `Attributes` are created for every record() call.
- if both the above configs are disabled. This can also act as potential fall back for any issues the datastructures used for pre-creation.

internal/venice-client-common/src/main/java/com/linkedin/venice/stats/VeniceMetricsConfig.java
```diff
@@ -166,6 +196,18 @@ public class VeniceMetricsConfig {
...
+   * Config to enable pre-compute Attributes for Venice metrics.
+   * check {@link #OTEL_VENICE_PRE_CREATE_ATTRIBUTES} for more details
+   */
+  private final boolean preCreateAttributes;
+
+  /**
+   * Config to enable lazy initialization of Attributes and caching them.
+   * check {@link #OTEL_VENICE_LAZY_INITIALIZE_ATTRIBUTES} for more details
+   */
+  private final boolean lazyInitializeAttributes;
```
gaojieliu, line 209: These two should be mutually exclusive strategies, right? Shall we use a single config option, so that users can specify different strategies?
Honestly-speaking, having 3 modes is an overkill IMO, and lazy-init should be good enough and in theory the regular test should be able to certify it.

internal/venice-client-common/src/main/java/com/linkedin/venice/stats/VeniceMetricsConfig.java
```diff
@@ -381,6 +437,14 @@ public Builder extractAndSetOtelConfigs(Map<String, String> configs) {
         }
       }
 
+      if ((configValue = configs.get(OTEL_VENICE_PRE_CREATE_ATTRIBUTES)) != null) {
```
lusong64, line 440: Small suggestion: 
```java
Optional.ofNullable(configs.get(OTEL_VENICE_PRE_CREATE_ATTRIBUTES))
    .map(this::setPreCreateAttributes);
```

Optional is quite cheap and it would help all the nullity check.

internal/venice-client-common/src/main/java/com/linkedin/venice/stats/VeniceMetricsConfig.java
```diff
@@ -381,6 +437,14 @@ public Builder extractAndSetOtelConfigs(Map<String, String> configs) {
...
+        setPreCreateAttributes(Boolean.parseBoolean(configValue));
+      }
+
+      if ((configValue = configs.get(OTEL_VENICE_LAZY_INITIALIZE_ATTRIBUTES)) != null) {
```
lusong64, line 444: ditto. But feel free to drop it.

internal/venice-client-common/src/main/java/com/linkedin/venice/stats/VeniceOpenTelemetryMetricsRepository.java
```diff
@@ -40,16 +45,76 @@
...
+    preCreateAttributes = metricsConfig.preCreateAttributes();
+    lazyInitializeAttributes = metricsConfig.lazyInitializeAttributes();
+    if (!emitOpenTelemetryMetrics) {
+      LOGGER.info("OpenTelemetry metrics are disabled");
+      return;
+    }
+    LOGGER.info(
+        "OpenTelemetry initialization for {} started with config: {}",
+        metricsConfig.getServiceName(),
+        metricsConfig.toString());
```
lusong64, line 71: the toString() can be removed.

internal/venice-client-common/src/main/java/com/linkedin/venice/stats/VeniceOpenTelemetryMetricsRepository.java
```diff
@@ -40,16 +45,76 @@
...
+    this.metricPrefix = "venice." + metricsConfig.getMetricPrefix();
```
lusong64, line 72: Why can't we include "venice." into the metricsConfig itself? Unless the metricsConfig is used by other product but it is VeniceMetricsConfig.

internal/venice-client-common/src/main/java/com/linkedin/venice/stats/VeniceOpenTelemetryMetricsRepository.java
```diff
@@ -40,16 +45,76 @@
...
+    try {
+      SdkMeterProviderBuilder builder = SdkMeterProvider.builder();
+      if (metricsConfig.exportOtelMetricsToEndpoint()) {
+        MetricExporter httpExporter = getOtlpHttpMetricExporter(metricsConfig);
+        builder.registerMetricReader(
+            PeriodicMetricReader.builder(httpExporter)
+                .setInterval(metricsConfig.getExportOtelMetricsIntervalInSeconds(), TimeUnit.SECONDS)
+                .build());
+      }
+      if (metricsConfig.exportOtelMetricsToLog()) {
```
lusong64, line 83: Is it possible that both exportOtelMetricsToEndpoint and exportOtelMetricsToLog are true? If it is the case, the builder's build() would be called twice. Is it an expected behavior?

internal/venice-client-common/src/main/java/com/linkedin/venice/stats/VeniceOpenTelemetryMetricsRepository.java
```diff
@@ -40,16 +45,76 @@
...
+            PeriodicMetricReader.builder(new LogBasedMetricExporter(metricsConfig))
+                .setInterval(metricsConfig.getExportOtelMetricsIntervalInSeconds(), TimeUnit.SECONDS)
+                .build());
+      }
+
+      if (metricsConfig.useOtelExponentialHistogram()) {
+        setExponentialHistogramAggregation(builder, metricsConfig);
+      }
+
+      builder.setResource(Resource.empty());
```
lusong64, line 95: The logic here is weird by unconditionally setting the resource as empty. Maybe moving the setting empty resource to the builder's resource field initialization part?

internal/venice-client-common/src/main/java/com/linkedin/venice/stats/VeniceOpenTelemetryMetricsRepository.java
lusong64: never mind. I was wrong.

lusong64: Actually if a user set both exportOtelMetricsToEndpoint and exportOtelMetricsToLog true, the metric would set the metricReader twice and would the latter one override the fist one?

internal/venice-client-common/src/main/java/com/linkedin/venice/stats/VeniceOpenTelemetryMetricsRepository.java
```diff
@@ -40,16 +45,76 @@
...
+      sdkMeterProvider = builder.build();
+
+      // Register MeterProvider with the OpenTelemetry instance
+      OpenTelemetry openTelemetry = OpenTelemetrySdk.builder().setMeterProvider(sdkMeterProvider).build();
+
+      this.meter = openTelemetry.getMeter(transformMetricName(getMetricPrefix(), metricFormat));
+      LOGGER.info(
+          "OpenTelemetry initialization for {} completed with config: {}",
+          metricsConfig.getServiceName(),
+          metricsConfig.toString());
```
lusong64, line 105: again, toString() is unnecessary here. And if you call toString(), it would increase the chance of NPE.

internal/venice-client-common/src/main/java/com/linkedin/venice/stats/metrics/MetricEntityState.java
```diff
@@ -102,16 +126,127 @@ void recordTehutiMetric(double value) {
     }
   }
 
-  public void record(long value, Attributes otelDimensions) {
-    recordOtelMetric(value, otelDimensions);
+  protected final void record(long value, Attributes attributes) {
```
lusong64, line 129: Why making these two method generic? Would package protection suffice for your test?

internal/venice-client-common/src/main/java/com/linkedin/venice/stats/metrics/MetricEntityStateOneEnum.java
```diff
@@ -0,0 +1,188 @@
...
+      // If preCreateAttributes is enabled, then the attributes should have been created during the constructor
+      attributes = attributesEnumMap.get(key);
+    } else if (lazyInitializeAttributes()) {
+      // If lazyInitializeAttributes is enabled, then the attributes should be created on demand for first time and
+      // cache it
+      attributes = attributesEnumMap.get(key);
+      if (attributes != null) {
+        return attributes; // Return from cache if found
+      }
+      attributes = attributesEnumMap.computeIfAbsent(key, this::createAttributes);
```
lusong64, line 147: The computeIfAbsent, imho, already covers statements starting from "if (attributes != null)".

internal/venice-client-common/src/main/java/com/linkedin/venice/stats/VeniceOpenTelemetryMetricsRepository.java
m-nagarajan: Yes we can register multiple readers.

internal/venice-client-common/src/main/java/com/linkedin/venice/stats/metrics/MetricEntityStateOneEnum.java
m-nagarajan: Good point. I modified this to just rely on `computeIfAbsent`

internal/venice-client-common/src/main/java/com/linkedin/venice/stats/VeniceOpenTelemetryMetricsRepository.java
m-nagarajan: This is the only place we need it, so both ways can work. Leaving like this for now.

internal/venice-client-common/src/main/java/com/linkedin/venice/stats/VeniceMetricsConfig.java
m-nagarajan: Following [this](https://venicedb.org/docs/dev_guide/how_to/style_guide#avoid-optionals) to aviod Optionals.

m-nagarajan (commented):
Addressed review comments. Thanks @FelixGV @gaojieliu and @lusong64 for the review.

gaojieliu (approved):

PR #1533, opened 2025-02-13
[protocol] Add gRPC endpoint for updateAdminOperationProtocolVersion in proto

minhmo1620:
## [protocol] Add gRPC endpoint for updateAdminOperationProtocolVersion
Within #1418, we have added the HTTP endpoint for `--update-admin-operation-protocol-version` in admin-tool to update admin operation protocol version, and we would do the same for gRPC endpoint. Due to the CI limitation, we are cutting proto change into this PR. The real implementation will be provided later.

## How was this PR tested?
CI


sushantmane (approved):
LGTM. Thanks @minhmo1620!

sushantmane (approved):
🚀

PR #1534, opened 2025-02-13
[dvc][server] verifying discovered host connectivity before initiating blob transfer request

jingy-li:
## [dvc][server] verifying discovered host connectivity before initiating blob transfer request

During testing on the dark hosts, we observed a few issues:

1. Previously, the per-partition bootstrap process worked as follows: it finds peers (e.g., 1500+ peers) and attempts to connect to each one. If the connection is successful, it sends the bootstrap request; if the connection fails, it moves on to the next candidate host. 
When there are many partitions, such as 200, this results in 200 * 1500 connection attempts. This process is unnecessarily time-consuming (waiting time = reject response time * 200 * 1500). Some partitions share the same peer-finding results, leading to redundant connection attempts to unreachable hosts. To address this, this PR introduces two map (a connectable host map and an unconnectable host map) to avoid redundant reconnections. The connectivity status need to be within certain preset config, if exceed the freshness, Netty client will retry to check the connectivity. 

2. During testing, we encountered a case where a partition bootstrap thread stalled after sending the connection request. It was found that the sender had already sent a reject response for the connection, but the receiver did not receive this response at all. As a result, the thread continued waiting for a response, and the partition was never successfully bootstrapped, whether via blob transfer or Kafka ingestion. To resolve this, this PR introduces a timeout to ensure that if the blob transfer or host connection is not completed, the process will fall back to Kafka ingestion instead of waiting indefinitely.

## How was this PR tested?
Tested on dark hosts.

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/NettyP2PBlobTransferManager.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/NettyP2PBlobTransferManager.java
@@ -95,10 +95,10 @@ public class NettyP2PBlobTransferManager implements P2PBlobTransferManager<Void>
     }
 
     List<String> discoverPeers = response.getDiscoveryResult();
-    Set<String> uniqueConnectablePeers = getUniqueConnectableHosts(discoverPeers, storeName, version, partition);
+    Set<String> connectablePeers = getConnectableHosts(discoverPeers, storeName, version, partition);
 
     // 2: Process peers sequentially to fetch the blob
-    processPeersSequentially(uniqueConnectablePeers, storeName, version, partition, tableFormat, resultFuture);
+    processPeersSequentially(connectablePeers, storeName, version, partition, tableFormat, resultFuture);
 
     return resultFuture;
   }
@@ -252,18 +252,14 @@ public class NettyP2PBlobTransferManager implements P2PBlobTransferManager<Void>
   }
 
   /**
-   * Get the unique connectable hosts for the given storeName, version, and partition
+   * Get the connectable hosts for the given storeName, version, and partition
    * @param discoverPeers the list of discovered peers
    * @param storeName the name of the store
    * @param version the version of the store
    * @param partition the partition of the store
    * @return the set of unique connectable hosts
    */
-  private Set<String> getUniqueConnectableHosts(
-      List<String> discoverPeers,
-      String storeName,
-      int version,
-      int partition) {
+  private Set<String> getConnectableHosts(List<String> discoverPeers, String storeName, int version, int partition) {
     // Extract unique hosts from the discovered peers
     Set<String> uniquePeers = discoverPeers.stream().map(peer -> peer.split("_")[0]).collect(Collectors.toSet());
 
@@ -278,10 +274,6 @@ public class NettyP2PBlobTransferManager implements P2PBlobTransferManager<Void>
     // Get the connectable hosts for this store, version, and partition
     Set<String> connectablePeers =
         nettyClient.getConnectableHosts((HashSet<String>) uniquePeers, storeName, version, partition);
-
-    // Exclude the current host
-    String currentHost = Utils.getHostName();
-    connectablePeers.remove(currentHost);
 
     LOGGER.info(
         "Total {} unique connectable peers for store {} version {} partition {}, peers are {}",
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/NettyP2PBlobTransferManager.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/NettyP2PBlobTransferManager.java
@@ -17,9 +17,12 @@ import com.linkedin.venice.utils.Utils;
 import java.io.InputStream;
 import java.time.Duration;
 import java.time.Instant;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
+import java.util.stream.Collectors;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
@@ -92,11 +95,10 @@ public class NettyP2PBlobTransferManager implements P2PBlobTransferManager<Void>
     }
 
     List<String> discoverPeers = response.getDiscoveryResult();
-    LOGGER
-        .info("Discovered peers {} for store {} version {} partition {}", discoverPeers, storeName, version, partition);
+    Set<String> uniqueConnectablePeers = getUniqueConnectableHosts(discoverPeers, storeName, version, partition);
 
     // 2: Process peers sequentially to fetch the blob
-    processPeersSequentially(discoverPeers, storeName, version, partition, tableFormat, resultFuture);
+    processPeersSequentially(uniqueConnectablePeers, storeName, version, partition, tableFormat, resultFuture);
 
     return resultFuture;
   }
@@ -121,7 +123,7 @@ public class NettyP2PBlobTransferManager implements P2PBlobTransferManager<Void>
    *  - Success case:
    *  1. If the blob is successfully fetched from a peer, an InputStream of the blob is returned.
    *
-   * @param peers the list of peers to process
+   * @param uniqueConnectablePeers the set of peers to process
    * @param storeName the name of the store
    * @param version the version of the store
    * @param partition the partition of the store
@@ -129,7 +131,7 @@ public class NettyP2PBlobTransferManager implements P2PBlobTransferManager<Void>
    * @param resultFuture the future to complete with the InputStream of the blob
    */
   private void processPeersSequentially(
-      List<String> peers,
+      Set<String> uniqueConnectablePeers,
       String storeName,
       int version,
       int partition,
@@ -142,11 +144,9 @@ public class NettyP2PBlobTransferManager implements P2PBlobTransferManager<Void>
     CompletableFuture<Void> chainOfPeersFuture = CompletableFuture.completedFuture(null);
 
     // Iterate through each peer and chain the futures
-    for (int currentPeerIndex = 0; currentPeerIndex < peers.size(); currentPeerIndex++) {
-      final int peerIndex = currentPeerIndex;
+    for (String chosenHost: uniqueConnectablePeers) {
       // Chain the next operation to the previous future
       chainOfPeersFuture = chainOfPeersFuture.thenCompose(v -> {
-        String chosenHost = peers.get(peerIndex).split("_")[0];
 
         if (resultFuture.isDone()) {
           // If the result future is already completed, skip the current peer
@@ -154,7 +154,13 @@ public class NettyP2PBlobTransferManager implements P2PBlobTransferManager<Void>
         }
 
         // Attempt to fetch the blob from the current peer asynchronously
-        LOGGER.info("Attempting to connect to host: {}", chosenHost);
+        LOGGER.info(
+            "Attempting to connect to host: {} for store {} version {} partition {} table format {}",
+            chosenHost,
+            storeName,
+            version,
+            partition,
+            tableFormat);
 
         return nettyClient.get(chosenHost, storeName, version, partition, tableFormat)
             .toCompletableFuture()
@@ -243,5 +249,47 @@ public class NettyP2PBlobTransferManager implements P2PBlobTransferManager<Void>
           partition,
           e);
     }
+  }
+
+  /**
+   * Get the unique connectable hosts for the given storeName, version, and partition
+   * @param discoverPeers the list of discovered peers
+   * @param storeName the name of the store
+   * @param version the version of the store
+   * @param partition the partition of the store
+   * @return the set of unique connectable hosts
+   */
+  private Set<String> getUniqueConnectableHosts(
+      List<String> discoverPeers,
+      String storeName,
+      int version,
+      int partition) {
+    // Extract unique hosts from the discovered peers
+    Set<String> uniquePeers = discoverPeers.stream().map(peer -> peer.split("_")[0]).collect(Collectors.toSet());
+
+    LOGGER.info(
+        "Discovered {} unique peers store {} version {} partition {}, peers are {}",
+        uniquePeers.size(),
+        storeName,
+        version,
+        partition,
+        uniquePeers);
+
+    // Get the connectable hosts for this store, version, and partition
+    Set<String> connectablePeers =
+        nettyClient.getConnectableHosts((HashSet<String>) uniquePeers, storeName, version, partition);
+
+    // Exclude the current host
+    String currentHost = Utils.getHostName();
+    connectablePeers.remove(currentHost);
+
+    LOGGER.info(
+        "Total {} unique connectable peers for store {} version {} partition {}, peers are {}",
+        connectablePeers.size(),
+        storeName,
+        version,
+        partition,
+        connectablePeers);
+    return connectablePeers;
   }
 }
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/client/NettyFileTransferClient.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/client/NettyFileTransferClient.java
@@ -203,6 +203,7 @@ public class NettyFileTransferClient {
   public void close() {
     workerGroup.shutdownGracefully();
     executorService.shutdown();
+    scheduledExecutorService.shutdown();
     unconnectableHosts.clear();
     connectedHosts.clear();
   }
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/client/NettyFileTransferClient.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/client/NettyFileTransferClient.java
@@ -1,8 +1,10 @@
 package com.linkedin.davinci.blobtransfer.client;
 
+import com.linkedin.alpini.base.concurrency.Executors;
 import com.linkedin.davinci.blobtransfer.BlobTransferUtils.BlobTransferTableFormat;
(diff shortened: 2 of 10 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/NettyP2PBlobTransferManager.java
```diff
@@ -92,11 +95,10 @@ public CompletionStage<InputStream> get(
     }
 
     List<String> discoverPeers = response.getDiscoveryResult();
-    LOGGER
-        .info("Discovered peers {} for store {} version {} partition {}", discoverPeers, storeName, version, partition);
+    Set<String> uniqueConnectablePeers = getUniqueConnectableHosts(discoverPeers, storeName, version, partition);
```
gaojieliu, line 98: So even for a single partition, the backend can return duplicate peers?

gaojieliu: I read more codes and realized that this function is used to discover connectable hosts concurrently, so why do you put `unique` keyword into the function name?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/client/NettyFileTransferClient.java
```diff
@@ -51,6 +65,69 @@ public void initChannel(SocketChannel ch) {
...
+   */
+  public Set<String> getConnectableHosts(
+      HashSet<String> discoveredHosts,
+      String storeName,
+      int version,
+      int partition) {
+    List<CompletableFuture<String>> futures = new ArrayList<>();
+    Set<String> connectableHosts = new HashSet<>();
+
+    discoveredHosts.removeAll(unconnectableHosts);
```
gaojieliu, line 88: Emm.. 
Is it possible this node is not connectable for the first partition request, but become connectable for the second request?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/client/NettyFileTransferClient.java
```diff
@@ -51,6 +65,69 @@ public void initChannel(SocketChannel ch) {
...
+          return null;
+        }
+      }, executorService);
+
+      futures.add(future);
+    }
+
+    // Wait for all futures to complete
+    CompletableFuture<Void> allConnections = CompletableFuture.allOf(futures.toArray(new CompletableFuture[0]));
+    allConnections.join();
```
gaojieliu, line 116: Do we need to specify a timeout here?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/client/NettyFileTransferClient.java
```diff
@@ -60,8 +137,11 @@ public CompletionStage<InputStream> get(
       int partition,
       BlobTransferTableFormat requestedTableFormat) {
     CompletionStage<InputStream> inputStream = new CompletableFuture<>();
+    ScheduledExecutorService scheduledExecutorService = Executors.newSingleThreadScheduledExecutor();
```
gaojieliu, line 140: Can this be shared across all the requests?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/client/NettyFileTransferClient.java
```diff
@@ -32,6 +42,10 @@ public class NettyFileTransferClient {
   private final String baseDir;
   private final int serverPort;
   private StorageMetadataService storageMetadataService;
+  private final ExecutorService executorService;
```
gaojieliu, line 45: We will create one `NettyFileTransferClient` for one store, right?
In theory, this thread pool can be shared across multiple stores.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/NettyP2PBlobTransferManager.java
jingy-li: Even for a single partition, the router may return the same host name but with different tags (for example, host1_app1_i000, host1_app1_i001). Therefore, we changed the list to a set to ensure uniqueness. 
However, the term 'unique' seems misleading, so I will remove it.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/client/NettyFileTransferClient.java
jingy-li: Yes, it's possible, but we haven't found an effective way to avoid it. For instance, with 1500 hosts and 200 partitions, the implementation bootstraps each partition in parallel, nearly simultaneously. If we attempt to reconnect all hosts, it becomes unnecessary to retry connectivity for a single host, as this would involve attempting to connect to all 1500+ hosts again.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/client/NettyFileTransferClient.java
jingy-li: Good call, added a timeout.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/client/NettyFileTransferClient.java
jingy-li: Good call! Initiated `ScheduledExecutorService scheduledExecutorService` when create NettyFileTransferClient.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/client/NettyFileTransferClient.java
jingy-li: NettyFileTransferClient is create for multiple store, it was created at getP2PBlobTransferManagerForDVCAndStart which is generated at DaVinciBackend.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/NettyP2PBlobTransferManager.java
gaojieliu: Emm..
For different tags, I assume DaVinci will use different listening ports for blob transfer, right?
Different instances in the same node can host different partitions in a sharded env.

jingy-li: In theory, if a single host is running two DVC instances, it should be able to perform the transfer, as long as their ports are different. This is how we conduct our integration tests.

However, in practice, this is not the case. We default both the client and server ports to 27017, so if instance 1 is running on port 27017 and instance 2 on port 27018, we will only be able to connect to instance 1 when another host try to send connect request.

Here is code in global-config, da-vinci.src:
`<property name="daVinciClient.davinci.p2p.blob.transfer.client.port" value="27017"/>
    <property name="daVinciClient.davinci.p2p.blob.transfer.server.port" value="27017"/>`

gaojieliu: This is certainly an issue and we can follow up offline.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/client/NettyFileTransferClient.java
gaojieliu: In theory, parallel requests should make it possible to reconnect all the hosts for every partition as I assume connect request should be fairly cheap.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/client/NettyFileTransferClient.java
gaojieliu: I see.
Let us make the variable names self descriptive. 
```
executorService
scheduledExecutorService
```
The names are too generic to infer the purpose.
Also, let us use `DaemonThreadFactory` to avoid blocking shutdown.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/NettyP2PBlobTransferManager.java
sixpluszero: Do we have production case where they deploy different tag to the same host? There should be way to set port by instance id per JVM but I wonder if we should do it or not in the first phase

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/client/NettyFileTransferClient.java
```diff
@@ -131,18 +144,42 @@ public Set<String> getConnectableHosts(
...
+  }
+
+  /**
+   * Check the freshness of the connectivity records and purge the stale records
+   */
+  private void purgeStaleConnectivityRecords() {
+    // Purge the unconnectable hosts
+    for (String host: unconnectableHostsToTimestamp.keySet()) {
+      Long lastAttempt = unconnectableHostsToTimestamp.get(host);
+      if (lastAttempt == null || System.currentTimeMillis() - lastAttempt > TimeUnit.SECONDS
```
gaojieliu, line 169: To be more efficient, we can use an iterator.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/client/NettyFileTransferClient.java
```diff
@@ -131,18 +144,42 @@ public Set<String> getConnectableHosts(
...
+      }
+    }
+
+    // Purge the connected hosts
+    for (String host: connectedHostsToTimestamp.keySet()) {
+      Long lastConnected = connectedHostsToTimestamp.get(host);
+      if (lastConnected == null || System.currentTimeMillis() - lastConnected > TimeUnit.SECONDS
+          .toMillis(peersConnectivityFreshnessInSeconds)) {
+        connectedHostsToTimestamp.remove(host);
+      }
```
gaojieliu, line 181: Same here

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/client/NettyFileTransferClient.java
```diff
@@ -18,27 +21,49 @@
...
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
 
 public class NettyFileTransferClient {
   private static final Logger LOGGER = LogManager.getLogger(NettyFileTransferClient.class);
   private static final int MAX_METADATA_CONTENT_LENGTH = 1024 * 1024 * 100;
+  private static final int TIMEOUT_IN_MINUTES = 5;
```
gaojieliu, line 41: 5 mins seems to be long, and typically, will the connect request take more than 1min?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/NettyP2PBlobTransferManager.java
jingy-li: We indeed observe this case in flip testing. 

Synced with @gaojieliu offline, for short term, we ask the client to reserve the port 27017 before they onboard to use blob transfer, for long term, we need to update host address and the listener(server-side) port num in the push status system. Then when trying to connect to server, the port info is also obtained from the peer finding.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/client/NettyFileTransferClient.java
gaojieliu: And duplicate logic in these two for loops.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/client/NettyFileTransferClient.java
jingy-li: Synced offline, we will have a map like <host name, last attempt to connect timestamp>. If it is fresh, we will not reconnect, otherwise, will reconnect.

gaojieliu (commented):
Left some more minor comments.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/client/NettyFileTransferClient.java
```diff
@@ -51,6 +76,101 @@ public void initChannel(SocketChannel ch) {
...
+
+    return connectableHostsResult;
+  }
+
+  /**
+   * Check the freshness of the connectivity records and purge the stale records
+   * @param hostsToTimestamp the map of hosts to the timestamp of the last connection attempt
+   */
+  public void purgeStaleConnectivityRecords(VeniceConcurrentHashMap<String, Long> hostsToTimestamp) {
+    hostsToTimestamp.forEach((host, lastAttempt) -> {
```
gaojieliu, line 168: Actually I meant something like the following:
```
Iterator<Map.Entry<<String, Long>>> iterator = hostsToTimestamp.entrySet().iterator();
    while (iterator.hasNext()) {
      Map.Entry<String, Long> entry = iterator.next();
      if (entry.getValue() == null || System.currentTimeMillis() - entry.getValue() > TimeUnit.SECONDS
          .toMillis(peersConnectivityFreshnessInSeconds)) {
        iterator.remove();
      }
    }
```

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/client/NettyFileTransferClient.java
jingy-li: Updated the connect timeout to 1 min.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/client/NettyFileTransferClient.java
```diff
@@ -51,6 +79,103 @@ public void initChannel(SocketChannel ch) {
...
+    purgeStaleConnectivityRecords(unconnectableHostsToTimestamp);
+    purgeStaleConnectivityRecords(connectedHostsToTimestamp);
+
+    discoveredHosts.removeAll(unconnectableHostsToTimestamp.keySet());
+    for (String host: discoveredHosts) {
+      CompletableFuture<String> future = CompletableFuture.supplyAsync(() -> {
+        try {
+          if (connectedHostsToTimestamp.keySet().contains(host)) {
+            return host; // already verified via previous connection
+          }
```
gaojieliu, line 113: A small enhancement:
We don't need to spin up a thread for the already verified hosts and we can get the diff between `discoveredHosts` and `connectedHostsToTimestamp.keySet()` before connecting, right?

Here is some sample code:
```
Set<String> connectableHostsResult = new HashSet<>();
    Iterator<String> iterator = discoveredHosts.iterator();
    while (iterator.hasNext()) {
       String host = iterator.next();
      if (connectedHostsToTimestamp.contains(host)) {
        resultSet.add(host);
        iterator.remove();
      }
    }
```

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/client/NettyFileTransferClient.java
```diff
@@ -51,6 +79,103 @@ public void initChannel(SocketChannel ch) {
...
+    // Collect only the successfully connected hosts
+    Set<String> connectableHostsResult = new HashSet<>();
+    for (CompletableFuture<String> future: futures) {
+      try {
+        String host = future.get();
+        if (host != null) {
+          connectableHostsResult.add(host);
+        }
+      } catch (Exception e) {
+        LOGGER.error("Error getting result from future", e);
```
gaojieliu, line 159: Is it possible to print out the un-reachable host here?

gaojieliu (commented):
I left a couple of minor comments and other than that, the code looks good.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/client/NettyFileTransferClient.java
jingy-li: Sure! Remove the connectable hosts before start connecting threads.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/client/NettyFileTransferClient.java
jingy-li: I might doubt it because if the host is unreachable, the future.get will return null and not include the host information, so we won't have the unreachable host here.

gaojieliu: Can we return a special object containing the requested hostname when the host is timeout during connecting?
We can wrap up the result into an object:
```
public class ConnectStatus {
   private final boolean connectable;
   private final String host;
}
```
So here, we can print out the hostname if the host doesn't finish connecting within 1 min by checking `ConnectionStatus.connectable`, right?

jingy-li: Added unconnectable hosts logs at channel active check, because even if we add the `ConnectStatus` class, when there is exception on `future.get()`, we won't know the host info as well.

gaojieliu (approved):
Thanks a lot for the improvement!

PR #1535, opened 2025-02-14
[admin-tool] Add gRPC endpoint for updateAdminOperationProtocolVersion

minhmo1620:
## [admin-tool] Add gRPC endpoint for updateAdminOperationProtocolVersion
Within #1418, we have added the HTTP endpoint for --update-admin-operation-protocol-version in admin-tool to update admin operation protocol version, and we would do the same for gRPC endpoint.

## How was this PR tested?
CI

### DIFF
--- a/services/venice-controller/src/main/java/com/linkedin/venice/controller/server/AdminTopicMetadataRoutes.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/controller/server/AdminTopicMetadataRoutes.java
@@ -20,6 +20,7 @@ import com.linkedin.venice.exceptions.VeniceException;
 import com.linkedin.venice.protocols.controller.AdminTopicGrpcMetadata;
 import com.linkedin.venice.protocols.controller.AdminTopicMetadataGrpcRequest;
 import com.linkedin.venice.protocols.controller.AdminTopicMetadataGrpcResponse;
+import com.linkedin.venice.protocols.controller.UpdateAdminOperationProtocolVersionGrpcRequest;
 import com.linkedin.venice.protocols.controller.UpdateAdminTopicMetadataGrpcRequest;
 import java.util.Optional;
 import org.apache.http.HttpStatus;
@@ -105,7 +106,7 @@ public class AdminTopicMetadataRoutes extends AbstractRoute {
     };
   }
 
-  public Route updateAdminOperationProtocolVersion(Admin admin) {
+  public Route updateAdminOperationProtocolVersion(Admin admin, ClusterAdminOpsRequestHandler requestHandler) {
     return (request, response) -> {
       AdminTopicMetadataResponse responseObject = new AdminTopicMetadataResponse();
       response.type(HttpConstants.JSON);
@@ -120,11 +121,15 @@ public class AdminTopicMetadataRoutes extends AbstractRoute {
         AdminSparkServer.validateParams(request, UPDATE_ADMIN_OPERATION_PROTOCOL_VERSION.getParams(), admin);
         String clusterName = request.queryParams(CLUSTER);
         Long adminOperationProtocolVersion = Long.parseLong(request.queryParams(ADMIN_OPERATION_PROTOCOL_VERSION));
+        AdminTopicMetadataGrpcResponse internalResponse = requestHandler.updateAdminOperationProtocolVersion(
+            UpdateAdminOperationProtocolVersionGrpcRequest.newBuilder()
+                .setClusterName(clusterName)
+                .setAdminOperationProtocolVersion(adminOperationProtocolVersion)
+                .build());
 
-        responseObject.setCluster(clusterName);
-        responseObject.setAdminOperationProtocolVersion(adminOperationProtocolVersion);
-
-        admin.updateAdminOperationProtocolVersion(clusterName, adminOperationProtocolVersion);
+        responseObject.setCluster(internalResponse.getMetadata().getClusterName());
+        responseObject
+            .setAdminOperationProtocolVersion(internalResponse.getMetadata().getAdminOperationProtocolVersion());
       } catch (Throwable e) {
         responseObject.setError(e);
         AdminSparkServer.handleError(new VeniceException(e), request, response);
--- a/services/venice-controller/src/main/java/com/linkedin/venice/controller/server/ClusterAdminOpsRequestHandler.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/controller/server/ClusterAdminOpsRequestHandler.java
@@ -153,15 +153,9 @@ public class ClusterAdminOpsRequestHandler {
   public AdminTopicMetadataGrpcResponse updateAdminOperationProtocolVersion(
       UpdateAdminOperationProtocolVersionGrpcRequest request) {
     String clusterName = request.getClusterName();
-    if (StringUtils.isBlank(clusterName)) {
-      throw new IllegalArgumentException("Cluster name is required for updating admin operation protocol version");
-    }
-
     long adminOperationProtocolVersion = request.getAdminOperationProtocolVersion();
-    if (adminOperationProtocolVersion == 0 || adminOperationProtocolVersion < -1) {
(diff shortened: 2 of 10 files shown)


internal/venice-common/src/main/proto/controller/ClusterAdminOpsGrpcService.proto
```diff
@@ -14,6 +14,7 @@ service ClusterAdminOpsGrpcService {
   // AdminTopicMetadata
   rpc getAdminTopicMetadata(AdminTopicMetadataGrpcRequest) returns (AdminTopicMetadataGrpcResponse) {}
   rpc updateAdminTopicMetadata(UpdateAdminTopicMetadataGrpcRequest) returns (UpdateAdminTopicMetadataGrpcResponse) {}
+  rpc updateAdminOperationProtocolVersion(UpdateAdminOperationProtocolVersionGrpcRequest) returns (AdminTopicMetadataGrpcResponse) {}
```
sushantmane, line 17: I wonder if we should reuse  `UpdateAdminTopicMetadataGrpcRequest` instead of adding new `UpdateAdminOperationProtocolVersionGrpcRequest` type?

minhmo1620: I was thinking the same! Technically, they should all the same. I was adding new type is because I think it is much clearer to see what variables they can send in. I don't mind reusing the UpdateAdminTopicMetadataGrpcRequest, I will just need extra validation.

sushantmane: >. I was adding new type is because I think it is much clearer to see what variables they can send in

I agree. Let's keep it then

services/venice-controller/src/main/java/com/linkedin/venice/controller/server/ClusterAdminOpsRequestHandler.java
```diff
@@ -148,4 +149,30 @@ public AdminTopicMetadataGrpcResponse updateAdminTopicMetadata(UpdateAdminTopicM
         AdminTopicMetadataGrpcResponse.newBuilder().setMetadata(adminTopicGrpcMetadataBuilder.build());
     return responseBuilder.build();
   }
+
+  public AdminTopicMetadataGrpcResponse updateAdminOperationProtocolVersion(
+      UpdateAdminOperationProtocolVersionGrpcRequest request) {
+    String clusterName = request.getClusterName();
+    if (StringUtils.isBlank(clusterName)) {
```
sushantmane, line 156: Can we follow the pattern used in the other methods in this class, i.e., move param validation in the separate method of validator class

sushantmane (commented):
Let's add simple test for gRPC as well. Refer to tests for other endpoints for that service.

services/venice-controller/src/main/java/com/linkedin/venice/controller/server/AdminTopicMetadataRoutes.java
```diff
@@ -120,11 +121,15 @@ public Route updateAdminOperationProtocolVersion(Admin admin) {
         AdminSparkServer.validateParams(request, UPDATE_ADMIN_OPERATION_PROTOCOL_VERSION.getParams(), admin);
         String clusterName = request.queryParams(CLUSTER);
         Long adminOperationProtocolVersion = Long.parseLong(request.queryParams(ADMIN_OPERATION_PROTOCOL_VERSION));
+        AdminTopicMetadataGrpcResponse internalResponse = requestHandler.updateAdminOperationProtocolVersion(
+            UpdateAdminOperationProtocolVersionGrpcRequest.newBuilder()
+                .setClusterName(clusterName)
+                .setAdminOperationProtocolVersion(adminOperationProtocolVersion)
+                .build());
```
huangminchn, line 128: This is new to me. I don't find the builder function somehow. Do we need to proactively close it? I guess no, since it's not a thread and should be GC automatically? cc: @sushantmane

services/venice-controller/src/main/java/com/linkedin/venice/controller/server/ClusterAdminOpsRequestHandler.java
```diff
@@ -148,4 +149,30 @@ public AdminTopicMetadataGrpcResponse updateAdminTopicMetadata(UpdateAdminTopicM
...
+        "Updating admin operation protocol version for cluster: {} to version: {}",
+        clusterName,
+        adminOperationProtocolVersion);
+
+    admin.updateAdminOperationProtocolVersion(clusterName, adminOperationProtocolVersion);
+
+    AdminTopicGrpcMetadata.Builder adminMetadataBuilder = AdminTopicGrpcMetadata.newBuilder()
+        .setClusterName(clusterName)
+        .setAdminOperationProtocolVersion(adminOperationProtocolVersion);
+    return AdminTopicMetadataGrpcResponse.newBuilder().setMetadata(adminMetadataBuilder.build()).build();
```
huangminchn, line 176: Is it the common way to write gRpc? I am curious what's reasoning behind building yet another metadata instance. Is it for improving modularization and readability? cc: @sushantmane

services/venice-controller/src/main/java/com/linkedin/venice/controller/server/AdminTopicMetadataRoutes.java
sushantmane: `UpdateAdminOperationProtocolVersionGrpcRequest` is generate based on `gRPC` proto definition by protobuf compiler. We don't need to close it as it's like any other regular java clas

services/venice-controller/src/main/java/com/linkedin/venice/controller/server/ClusterAdminOpsRequestHandler.java
sushantmane: > Is it for improving modularization 
Generally, yes!

I think @minhmo1620 is trying to reuse AdminTopicMetadataGrpcResponse for this endpoint. The response includes an inner field, AdminTopicGrpcMetadata, which was separated to enable reuse across other APIs.

services/venice-controller/src/main/java/com/linkedin/venice/controller/server/AdminTopicMetadataRoutes.java
minhmo1620: Adding to @sushantmane , this is the [PR](https://github.com/linkedin/venice/pull/1533) which we define the proto rpc. Then at compile time, it will generate the java classes based on the definition.

services/venice-controller/src/main/java/com/linkedin/venice/controller/server/ClusterAdminOpsRequestHandler.java
minhmo1620: I was trying to reuse `AdminTopicGrpcMetadata` so we don't need to duplicate the available fields inside metadata multiple times. The metadata is compacted, which I think is cleaner. 

In case you ask why don't we bring it out, let's say return `AdminTopicGrpcMetadata` directly instead of `AdminTopicMetadataGrpcResponse`. We can do that, however, the benefit of the current setup is clearing out the request and response object. In case we want to extend the current response object to have something else (that not part of metadata), we can easily add it in the same level with the metadata. I don't have strong preference here, I'm open for discussion on this!

huangminchn: The current setup is okay. Thanks!

huangminchn (approved):
Thanks Minh!

services/venice-controller/src/main/java/com/linkedin/venice/controller/server/ClusterAdminOpsRequestHandler.java
minhmo1620: Updated, please take another look! Thank you!

sushantmane (approved):
LGTM. Thanks!

PR #1536, opened 2025-02-14
[changelog] Fix metadata refresh deadlock scenario

ZacAttack:
There exists a deadlocking race condition when a version swap happens. The version swap data change listener locks itself and then proceeds with the version swap on the consumer.  However, other function calls would also call refresh() on the metadata repository.  This in turn would trigger more callbacks on the data change listener.  In some scenarios, this can cause deadlocking.

This change makes sure that we no longer leverage manual refreshes and instead let the whole process get driven by the scheduled metadata refresh.  This also makes the refresh time configurable, and seems to speed up tests (yay!) as there were frequent retries on consuming events as we had to wait for the version swap to happen internally before we could consume the events we expected.

Resolves #XXX

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/ChangelogClientConfig.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/ChangelogClientConfig.java
@@ -9,6 +9,7 @@ import org.apache.avro.specific.SpecificRecord;
 
 
 public class ChangelogClientConfig<T extends SpecificRecord> {
+  private Long metaDataRefreshIntervalInSeconds = 60L;
   private Properties consumerProperties;
   private SchemaReader schemaReader;
   private String viewName;
@@ -166,12 +167,12 @@ public class ChangelogClientConfig<T extends SpecificRecord> {
     return this.bootstrapFileSystemPath;
   }
 
-  public long getVersionSwapDetectionIntervalTimeInMs() {
+  public long getVersionSwapDetectionIntervalTimeInSeconds() {
     return versionSwapDetectionIntervalTimeInMs;
   }
 
-  public ChangelogClientConfig setVersionSwapDetectionIntervalTimeInMs(long intervalTimeInMs) {
-    this.versionSwapDetectionIntervalTimeInMs = intervalTimeInMs;
+  public ChangelogClientConfig setVersionSwapDetectionIntervalTimeInSeconds(long intervalTimeInSeconds) {
+    this.versionSwapDetectionIntervalTimeInMs = intervalTimeInSeconds;
     return this;
   }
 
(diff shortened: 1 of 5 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumerImpl.java
```diff
@@ -1,5 +1,6 @@
 package com.linkedin.davinci.consumer;
 
+import static com.linkedin.venice.ConfigKeys.*;
```
KaiSernLim, line 3: nit: static import `CLIENT_SYSTEM_STORE_REPOSITORY_REFRESH_INTERVAL_SECONDS`?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceAfterImageConsumerImpl.java
```diff
@@ -101,11 +92,6 @@ public CompletableFuture<Void> subscribe(Set<Integer> partitions) {
     if (!versionSwapThreadScheduled.get()) {
       // schedule the version swap thread and set up the callback listener
       this.storeRepository.registerStoreDataChangedListener(versionSwapListener);
-      versionSwapExecutorService.scheduleAtFixedRate(
-          new VersionSwapDetectionThread(),
```
KaiSernLim, line 105: nit: Can the `VersionSwapDetectionThread` at the bottom of the file also be removed?

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/consumer/TestChangelogConsumer.java
```diff
@@ -175,6 +170,99 @@ public void cleanUp() {
...
+      // run push job
+      TestWriteUtils.runPushJob("Run push job", props);
+
+      // Assert the push has gone through
+      int expectedVersion = i + 2;
+      TestUtils.waitForNonDeterministicAssertion(
+          5,
+          TimeUnit.SECONDS,
+          () -> Assert
+              .assertEquals(controllerClient.getStore(storeName).getStore().getCurrentVersion(), expectedVersion));
```
KaiSernLim, line 255: Seems like the current version is not getting set to 2...

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceAfterImageConsumerImpl.java
ZacAttack: sure can

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/consumer/TestChangelogConsumer.java
ZacAttack: I'm not sure I follow this one?  What do you mean?

ZacAttack: oh nevermind I know what you mean.  This is the parent controllers controller client, which will always have a current version of zero.  I had to swap it in because static analysis was mad that I had a variable that I wasn't using, but one of the methods for store setup returns to you a controller client, and then static analysis was mad that I wasn't looking at a functions return value.  So made a child controller client for this check and then did a useless lookup with the parent controller client to keep everyone happy.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/ChangelogClientConfig.java
```diff
@@ -166,12 +166,12 @@ public String getBootstrapFileSystemPath() {
     return this.bootstrapFileSystemPath;
   }
 
-  public long getVersionSwapDetectionIntervalTimeInMs() {
+  public long getVersionSwapDetectionIntervalTimeInSeconds() {
     return versionSwapDetectionIntervalTimeInMs;
```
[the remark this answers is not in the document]
ZacAttack: good idea.  Also fixed the default.

KaiSernLim (approved):

xunyin8 (approved):
Thanks Zac!

PR #1537, opened 2025-02-14
[server] Produce heartbeat to separate RT topic

sixpluszero:
## [server] Produce heartbeat to separate RT topic
Produce heartbeat to separate RT topic at the same time as producing to RT.  
This is needed in the edge case where if the rewind time is shorter than the  
previous separate RT write, there will be no data to replay in separate RT,  
and the ready-to-serve check will never pass because no offset has been  
persisted for the separate RT region.  

## How was this PR tested?
Add test in integration test. Without this patch, due to short rewind time in test setup, it will fail complete new push.

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
@@ -3824,19 +3824,13 @@ public class LeaderFollowerStoreIngestionTask extends StoreIngestionTask {
       // Not time for another heartbeat yet.
       return null;
     }
-    Set<String> failedPartitions = sendHeartbeatToRealtimeTopic(false);
-    if (isSeparatedRealtimeTopicEnabled()) {
-      failedPartitions.addAll(sendHeartbeatToRealtimeTopic(true));
-    }
-    lastSendIngestionHeartbeatTimestamp.set(currentTimestamp);
-    return failedPartitions;
-  }
 
-  Set<String> sendHeartbeatToRealtimeTopic(boolean isSeparateTopic) {
-    AtomicInteger numHeartBeatSuccess = new AtomicInteger(0);
     List<CompletableFuture<PubSubProduceResult>> heartBeatFutures = new ArrayList<>();
+    List<CompletableFuture<PubSubProduceResult>> heartBeatFuturesForSepRT = new ArrayList<>();
     Set<String> failedPartitions = VeniceConcurrentHashMap.newKeySet();
+    Set<String> failedPartitionsForSepRT = VeniceConcurrentHashMap.newKeySet();
     AtomicReference<CompletionException> completionException = new AtomicReference<>(null);
+    AtomicReference<CompletionException> completionExceptionForSepRT = new AtomicReference<>(null);
     for (PartitionConsumptionState pcs: partitionConsumptionStateMap.values()) {
       PubSubTopic leaderTopic = pcs.getOffsetRecord().getLeaderTopic(pubSubTopicRepository);
       if (isLeader(pcs) && leaderTopic != null && leaderTopic.isRealTime()) {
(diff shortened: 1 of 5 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -2259,23 +2259,20 @@ protected void recordHeartbeatReceived(
       return;
     }
 
-    // Separate incremental push pubsub entries has the same pubsub url but different cluster id, which creates
```
sushantmane, line 2261: Is this not required for IncPush topic anymore? OR is it the case that URL trickled down from `StorePartitionDataReceiver` is already resolved?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -395,6 +396,7 @@ public StoreIngestionTask(
     this.storeName = versionTopic.getStoreName();
     this.isUserSystemStore = VeniceSystemStoreUtils.isUserSystemStore(storeName);
     this.realTimeTopic = pubSubTopicRepository.getTopic(Utils.getRealTimeTopicName(version));
+    this.separateRealTimeTopic = pubSubTopicRepository.getTopic(Version.composeSeparateRealTimeTopic(storeName));
```
sushantmane, line 399: Can we set this topic iff `isSeparatedRealtimeTopicEnabled` OR we can remove boolean for `isSeparatedRealtimeTopicEnabled` and use nullability of this field to determine if `isSeparatedRealtimeTopicEnabled`

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -3830,9 +3824,17 @@ protected Set<String> maybeSendIngestionHeartbeat() {
       // Not time for another heartbeat yet.
       return null;
     }
+    Set<String> failedPartitions = sendHeartbeatToRealtimeTopic(false);
```
sushantmane, line 3827: Should we call this method just once and then send HB to both reg RT and sep RT if the conditions meet.  This way we don't need to do double looping and verification. Additionally we can make return `maybeSendIngestionHeartbeat` since the return value doesn't seem to be used anywhere

sushantmane (commented):
Few comments but overall LGTM! Thanks, @sixpluszero!

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
gaojieliu: Good question.
I read the code a little bit and I am guessing the Kafka url for inc push RT topic is having `_sep` suffix, so does `getKafkaClusterUrlToAliasMap` contain the right alias for such special urls?

sixpluszero: So for original RT this does not change anything; For separate RT, it will find the correct cluster ID in the record heartbeat method. 
We will only register metric and sensor that these new regions when version level config is enabled.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
sixpluszero: I think about that actually, then I will need to do send HB internal method manually for twice, but maybe this is more clean, I don't have a strong opinion for that.

> Additionally we can make return maybeSendIngestionHeartbeat since the return value doesn't seem to be used anywhere
I checked it is still used in test, we should keep it.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
sixpluszero: isSeparatedRealtimeTopicEnabled is used in multiple places, I can change this to be based on the flag. I created this variable here mainly for the fact that I don't want to create it everytime it is needed.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
sixpluszero: Actually I know why I did that, it is because the below logging requires the topic to be specified...let me think if I can separate the log only and do that send in one loop

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
sushantmane: We can keep the method: `isSeparatedRealtimeTopicEnabled` itself and change the impl to use nullness check? and get rid of var `isSeparatedRealtimeTopicEnabled`

gaojieliu (commented):
Looks good overall, left one minor comment.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -3824,19 +3824,13 @@ protected Set<String> maybeSendIngestionHeartbeat() {
...
-    return failedPartitions;
-  }
 
-  Set<String> sendHeartbeatToRealtimeTopic(boolean isSeparateTopic) {
-    AtomicInteger numHeartBeatSuccess = new AtomicInteger(0);
     List<CompletableFuture<PubSubProduceResult>> heartBeatFutures = new ArrayList<>();
+    List<CompletableFuture<PubSubProduceResult>> heartBeatFuturesForSepRT = new ArrayList<>();
     Set<String> failedPartitions = VeniceConcurrentHashMap.newKeySet();
+    Set<String> failedPartitionsForSepRT = VeniceConcurrentHashMap.newKeySet();
     AtomicReference<CompletionException> completionException = new AtomicReference<>(null);
```
sushantmane, line 3832: Can we init collections size == `pcs.values().size()` and  sepRT collections empty if the feature is not enabled?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
gaojieliu: If I understand your comment correctly, `kafkaClusterUrlToAliasMap` does contain the alias for the special urls for inc push RT topics, right?

sushantmane (approved):
Thanks, @sixpluszero!

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
sixpluszero: yes it does, if we deployed the config change. So far it is deployed and verified in two test clusters.

PR #1538, opened 2025-02-15
[server][vpj][controller] Remove Kafka serialization classes and improve producer flexibility

sushantmane:
## Cleanup and refactor producer adapter configuration

- Removed redundant VeniceWriterFactory for meta store writes in KafkaStoreIngestionTask  
- Moved serialization out of Kafka producer to Venice layer, added API for raw byte messages  
- Deprecated old create API in PubSubProducerAdapterFactory, added a new single-argument API  
- Removed unused serialization configs from Admin tool message dumper  
- Added PubSubProduceResult::getPubSubPosition to return PubSubPosition  
- Deleted SharedKafkaProducerConfig and other dead code from MetaStoreWriter  
- Introduced PubSubMessageSerializer to pass down serializers for KafkaKey and KME  
- Added PubSubProducerAdapterContext to centralize common configs for producer adapters  
- Refactored producer compression config handling in VeniceWriterFactory  
- Moved broker address resolution up in VeniceWriterFactory  


ZacAttack (changes requested):
This looks ok aside from the test failures.

com.linkedin.venice.exceptions.VeniceException: Broker address must be provided to create a pub-sub producer

^^ I think if you just patch that all the failing cases will pass

FelixGV:
Just some high-level comments...

1. I think for the server/DVC code, it is almost certainly a good idea to use `byte[]` and let us control the serde on our own. This will open up some new opportunities for optimization.
2. I think the current API where we can interact with the pub sub via POJOs is probably useful in the rest of the code (e.g. controller certainly, possibly VPJ as well).

So rather than making the old API deprecated, can we just support both? It might make the size of this change smaller as well...

sushantmane:
> Just some high-level comments...
> 
> 1. I think for the server/DVC code, it is almost certainly a good idea to use `byte[]` and let us control the serde on our own. This will open up some new opportunities for optimization.
> 2. I think the current API where we can interact with the pub sub via POJOs is probably useful in the rest of the code (e.g. controller certainly, possibly VPJ as well).
> 
> So rather than making the old API deprecated, can we just support both? It might make the size of this change smaller as well...




Sounds good.

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/api/PubSubProducerAdapterContext.java
```diff
@@ -0,0 +1,166 @@
...
+      if (brokerAddress == null) {
+        throw new VeniceException("Broker address must be provided to create a pub-sub producer");
+      }
+      if (veniceProperties == null) {
+        veniceProperties = VeniceProperties.empty();
+      }
+      if (pubSubMessageSerializer == null) {
+        pubSubMessageSerializer = PubSubMessageSerializer.DEFAULT_PUBSUB_SERIALIZER;
+      }
+      if (compressionType == null) {
```
ZacAttack, line 157: wait... is this right?  Couldn't it be NONE?

ZacAttack (dismissed):
Conditional approve pending one comment

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/api/PubSubProducerAdapterContext.java
sushantmane: I've changed this to make it straightforward as per you suggestion. Thanks!

ZacAttack (approved):

sushantmane:
Thanks a lot, @ZacAttack for the review!

PR #1539, opened 2025-02-15
[controller] Add test timeout for deferred version swap test

misyel:

Adding a test timeout for deferred version swap test

## How was this PR tested?

Existing tests


sushantmane (approved):
Thank you!

PR #1541, opened 2025-02-18
[server] Add pool type information to Kafka consumer client ID

sixpluszero:
## [server] Add pool type information to Kafka consumer client ID
A simple fix for Kafka consumer ID generation.
Added the pool type so that different pool's consumer won't share same client ID.
This is helpful when debugging consumption with JMX.

## How was this PR tested?

Added a new unit test


sushantmane (approved):
LGTM. Thanks

PR #1543, opened 2025-02-18
[admin-tool] Add store list filter to cluster-batch-task

xunyin8:
## [admin-tool] Add store list filter to cluster-batch-task

Add new optional parameter STORE_FILTER_FILE to CLUSTER_BATCH_TASK to perform cluster batch task only on the intersection of provided stores and stores in the cluster. This is useful when you have a list of problematic stores that you'd like to perform some repair on, e.g. re-push system store but instead of doing it to all stores in a cluster you can just run CLUSTER_BATCH_TASK with this store list for all clusters and it will only do what's necessary.

If needed in the future we can also add other set operations like complement. Which will do some operation on the stores in the cluster excluding stores in the provided filter file.

## How was this PR tested?
verified in test environment.

### DIFF
--- a/clients/venice-admin-tool/src/main/java/com/linkedin/venice/AdminTool.java
+++ b/clients/venice-admin-tool/src/main/java/com/linkedin/venice/AdminTool.java
@@ -868,15 +868,33 @@ public class AdminTool {
     String task = getRequiredArgument(cmd, Arg.TASK_NAME, Command.CLUSTER_BATCH_TASK);
     String checkpointFile = getRequiredArgument(cmd, Arg.CHECKPOINT_FILE, Command.CLUSTER_BATCH_TASK);
     int parallelism = Integer.parseInt(getOptionalArgument(cmd, Arg.THREAD_COUNT, "1"));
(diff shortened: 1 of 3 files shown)


clients/venice-admin-tool/src/main/java/com/linkedin/venice/AdminTool.java
```diff
@@ -868,24 +868,50 @@ private static void clusterBatchTask(CommandLine cmd) {
     String task = getRequiredArgument(cmd, Arg.TASK_NAME, Command.CLUSTER_BATCH_TASK);
     String checkpointFile = getRequiredArgument(cmd, Arg.CHECKPOINT_FILE, Command.CLUSTER_BATCH_TASK);
     int parallelism = Integer.parseInt(getOptionalArgument(cmd, Arg.THREAD_COUNT, "1"));
+    String storeFilterFile = getOptionalArgument(cmd, Arg.STORE_FILTER_FILE, "");
+    Set<String> storeList = new HashSet<>();
```
majisourav99, line 872: nit: rename to stores, storseSet

lluwm (approved):
LGTM!

PR #1545, opened 2025-02-19
[controller] Do not throw exception on retrying empty push

majisourav99:
## Do not throw exception on retrying empty push

If the empty push of system store fails, the current code retries another empty push. We should not throw exception to block the admin channel on such cases.

## How was this PR tested?

GHCI

### DIFF
--- a/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
@@ -6360,7 +6360,7 @@ public class VeniceHelixAdmin implements Admin, StoreCleaner {
         LOGGER.warn(
             "System store: {} pushed failed. Issuing a new empty push to create version: {} ",
             systemStoreName,
-            +versionNumber);
+            versionNumber);
       } else {
         throw new VeniceRetriableException(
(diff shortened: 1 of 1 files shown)


kvargha (dismissed):

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -6357,9 +6357,10 @@ public void validateAndMaybeRetrySystemStoreAutoCreation(
             getReplicationFactor(clusterName, systemStoreName));
         int versionNumber = version.getNumber();
         writeEndOfPush(clusterName, systemStoreName, versionNumber, true);
-        throw new VeniceException(
-            "System store: " + systemStoreName + " pushed failed. Issuing a new empty push to create version: "
-                + versionNumber);
+        LOGGER.warn(
+            "System store: {} pushed failed. Issuing a new empty push to create version: {} ",
+            systemStoreName,
+            +versionNumber);
```
kvargha, line 6363: Should there be a plus before versionNumber

majisourav99: good catch!

kvargha (approved):

PR #1549, opened 2025-02-21
[controller] Allow AdminOperationSerializer to use lower version to serialize the AdminOperation message

minhmo1620:
## [controller] Allow AdminOperationSerializer to use lower version to serialize the AdminOperation message
- This PR allows AdminTopicConsumptionTask to use the config `adminOperationProtocolVersion` at AdminTopicMetadata on ZK (could be lower version) as the SOT for serializer to serialize the message and send that writer schema version along with the bytes to deserialize back at consumption stage.

Previous PR: #1418 

## Project description
TLDR version:
Initial start: Producers (parent controllers) need to generate `AdminConsumptionTask` to put into the admin topic queue. Messages will be consumed by consumers (parent controllers + child controllers). If producers produce the message that consumers cannot read, we are having troubles.

Current solution: Child controllers when upgraded to the newer version, will need to be completed BEFORE parent controllers. (Since parent controllers is currently using their latest version code, and child controller don't have the schema for latest version yet).

New solution:
We put a version number in parents' ZK metadata, specifying the version that producers will use (let's called it `X`)
- Producer: OK, I will always use version `X` to produce message **(This PR)**

In deployment, parent and child controllers can freely upgrade to `X+1`, without any delay or order.

Then, when they are all upgraded, we send a request to ZK:
- OK, let's use `X+1` from now on

=> Complete the deployment.

## Serialization issues when serializing the object by lower version
This issue is figured during the effort of serialize the object with different schema that it was created with.

The use case that I used:
- UpdateStore is created with AdminOperation schema v84 (current latest version)
- I tried to serialize this object with schema v74
=> Even though v84 and v74 is compatible, the serialization failed due to ClassCastException (cannot cast Boolean to Number).

Debug process:
- The writer will loop through the field inside schema v74 and get the index position, then use that index position to retrieve the value of the field.

image

- However, comparing to v74, v84 (update store) has one more field (separateRealTimeTopicEnabled) in the middle of the schema 

image

=> Then the field and position of the writer schema (v74) is basically not matched anymore (expecting pos 21 to be `replicationMetadataVersionID` (int), but got `writeComputationEnabled` (boolean) from v84, causing cast exception `java.lang.Boolean cannot be cast to java.lang.Number`

image

### There are two options:
**Option 1**: Override `getField` to get the field by fieldName
**Option 2**: Turn the current AdminOperation (with latest version schema) to GenericRecord (with writer schema id) and serialize the GenericRecord with writer schema id.

Option 1 was implemented ([PR](https://github.com/linkedin/venice/pull/1546)) but then closed after getting the suggestion for option 2.

Option 2, though very "bulky", it works and cut the future maintenance cost since we were modifying Avro API. We chose this option over option 1 and the implementation can be found specifically [here](https://github.com/linkedin/venice/pull/1549/files#diff-c22c4d06f7c68623a9bf1fcc6956f9dc1ad5e828883c1d40484a94363d0915deR50).

## How was this PR tested?
UTs and integration tests

### DIFF
--- a/clients/venice-admin-tool/src/test/java/com/linkedin/venice/TestAdminToolConsumption.java
+++ b/clients/venice-admin-tool/src/test/java/com/linkedin/venice/TestAdminToolConsumption.java
@@ -99,8 +99,9 @@ public class TestAdminToolConsumption {
       adminMessage.operationType = AdminMessageType.STORE_CREATION.getValue();
       adminMessage.payloadUnion = storeCreation;
       adminMessage.executionId = 1;
-      deserializer.serialize(adminMessage);
-      byte[] putValueBytes = deserializer.serialize(adminMessage);
+      deserializer.serialize(adminMessage, AdminOperationSerializer.LATEST_SCHEMA_ID_FOR_ADMIN_OPERATION);
+      byte[] putValueBytes =
+          deserializer.serialize(adminMessage, AdminOperationSerializer.LATEST_SCHEMA_ID_FOR_ADMIN_OPERATION);
       put.putValue = ByteBuffer.wrap(putValueBytes);
       put.replicationMetadataPayload = ByteBuffer.allocate(0);
       messageEnvelope.payloadUnion = put;
--- a/clients/venice-admin-tool/src/test/java/com/linkedin/venice/TestAdminToolConsumption.java
+++ b/clients/venice-admin-tool/src/test/java/com/linkedin/venice/TestAdminToolConsumption.java
@@ -99,7 +99,6 @@ public class TestAdminToolConsumption {
       adminMessage.operationType = AdminMessageType.STORE_CREATION.getValue();
       adminMessage.payloadUnion = storeCreation;
       adminMessage.executionId = 1;
-      deserializer.serialize(adminMessage, AdminOperationSerializer.LATEST_SCHEMA_ID_FOR_ADMIN_OPERATION);
       byte[] putValueBytes =
           deserializer.serialize(adminMessage, AdminOperationSerializer.LATEST_SCHEMA_ID_FOR_ADMIN_OPERATION);
       put.putValue = ByteBuffer.wrap(putValueBytes);
--- a/internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/controller/kafka/consumer/AdminConsumptionTaskIntegrationTest.java
+++ b/internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/controller/kafka/consumer/AdminConsumptionTaskIntegrationTest.java
@@ -5,7 +5,6 @@ import static com.linkedin.venice.ConfigKeys.ADMIN_CONSUMPTION_MAX_WORKER_THREAD
 
 import com.linkedin.venice.controller.Admin;
 import com.linkedin.venice.controller.kafka.AdminTopicUtils;
-import com.linkedin.venice.controller.kafka.protocol.admin.AddVersion;
 import com.linkedin.venice.controller.kafka.protocol.admin.AdminOperation;
 import com.linkedin.venice.controller.kafka.protocol.admin.DeleteStore;
 import com.linkedin.venice.controller.kafka.protocol.admin.DisableStoreRead;
@@ -22,6 +21,7 @@ import com.linkedin.venice.integration.utils.ServiceFactory;
 import com.linkedin.venice.integration.utils.VeniceControllerWrapper;
 import com.linkedin.venice.integration.utils.VeniceMultiRegionClusterCreateOptions;
 import com.linkedin.venice.integration.utils.VeniceTwoLayerMultiRegionMultiClusterWrapper;
+import com.linkedin.venice.meta.StoreInfo;
 import com.linkedin.venice.pubsub.PubSubProducerAdapterFactory;
 import com.linkedin.venice.pubsub.PubSubTopicRepository;
 import com.linkedin.venice.pubsub.api.PubSubTopic;
@@ -274,8 +274,9 @@ public class AdminConsumptionTaskIntegrationTest {
   }
 
   @Test(timeOut = 2 * TIMEOUT)
-  public void testParallelAdminExecutionTasksWithSpecificVersion() throws IOException, InterruptedException {
+  public void testAdminConsumptionTaskWithSpecificWriterId() throws IOException {
     int adminConsumptionMaxWorkerPoolSize = 3;
+    // Use a specific version to test the serialization and deserialization of admin operation.
     int writerSchemaId = AdminOperationSerializer.LATEST_SCHEMA_ID_FOR_ADMIN_OPERATION - 10;
 
     Properties parentControllerProps = new Properties();
@@ -322,18 +323,14 @@ public class AdminConsumptionTaskIntegrationTest {
 
         // Update store
         executionId++;
-        byte[] updateStoreMessage =
-            getStoreUpdateMessage(clusterName, storeName, owner, executionId, adminOperationSerializer, writerSchemaId);
+        byte[] updateStoreMessage = getStoreUpdateMessage(clusterName, storeName, owner, executionId, writerSchemaId);
         writer.put(new byte[0], updateStoreMessage, writerSchemaId);
-
-        // Create a new store version
-        executionId++;
-        byte[] addVersionMessage =
-            getAddVersionMessage(clusterName, storeName, "empty_push", 2, 20, executionId, writerSchemaId);
-        writer.put(new byte[0], addVersionMessage, writerSchemaId);
 
         TestUtils.waitForNonDeterministicAssertion(TIMEOUT, TimeUnit.MILLISECONDS, () -> {
           Assert.assertFalse(parentControllerClient.getStore(storeName).isError());
+          StoreInfo storeInfo = parentControllerClient.getStore(storeName).getStore();
+          Assert.assertTrue(storeInfo.isEnableStoreWrites());
+          Assert.assertTrue(storeInfo.isEnableStoreReads());
         });
       }
     }
@@ -424,47 +421,20 @@ public class AdminConsumptionTaskIntegrationTest {
       String storeName,
       String owner,
       long executionId,
-      AdminOperationSerializer adminOperationSerializer,
       int writerSchemaId) {
     UpdateStore updateStore = (UpdateStore) AdminMessageType.UPDATE_STORE.getNewInstance();
     updateStore.clusterName = clusterName;
     updateStore.storeName = storeName;
     updateStore.owner = owner;
     updateStore.partitionNum = 20;
-    updateStore.currentVersion = 1;
+    updateStore.currentVersion = AdminConsumptionTask.IGNORED_CURRENT_VERSION;
     updateStore.enableReads = true;
     updateStore.enableWrites = true;
     updateStore.replicateAllConfigs = true;
     updateStore.updatedConfigsList = Collections.emptyList();
-    updateStore.incrementalPushEnabled = true;
     AdminOperation adminMessage = new AdminOperation();
     adminMessage.operationType = AdminMessageType.UPDATE_STORE.getValue();
     adminMessage.payloadUnion = updateStore;
-    adminMessage.executionId = executionId;
-    return adminOperationSerializer.serialize(adminMessage, writerSchemaId);
-  }
-
-  private byte[] getAddVersionMessage(
-      String clusterName,
-      String storeName,
-      String pushJobId,
-      int versionNum,
-      int numberOfPartitions,
-      long executionId,
-      int writerSchemaId) {
-    AddVersion addVersion = (AddVersion) AdminMessageType.ADD_VERSION.getNewInstance();
-    addVersion.clusterName = clusterName;
-    addVersion.storeName = storeName;
-    addVersion.pushJobId = pushJobId;
-    addVersion.versionNum = versionNum;
-    addVersion.numberOfPartitions = numberOfPartitions;
-    addVersion.rewindTimeInSecondsOverride = -1;
-    addVersion.timestampMetadataVersionId = 1;
-    addVersion.versionSwapDeferred = false;
-
-    AdminOperation adminMessage = new AdminOperation();
-    adminMessage.operationType = AdminMessageType.ADD_VERSION.getValue();
-    adminMessage.payloadUnion = addVersion;
     adminMessage.executionId = executionId;
     return adminOperationSerializer.serialize(adminMessage, writerSchemaId);
   }
--- a/internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/controller/kafka/consumer/AdminConsumptionTaskIntegrationTest.java
+++ b/internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/controller/kafka/consumer/AdminConsumptionTaskIntegrationTest.java
@@ -138,8 +138,9 @@ public class AdminConsumptionTaskIntegrationTest {
         AdminOperationSerializer.LATEST_SCHEMA_ID_FOR_ADMIN_OPERATION);
     writer.put(new byte[0], goodMessage, AdminOperationSerializer.LATEST_SCHEMA_ID_FOR_ADMIN_OPERATION);
 
-    Utils.sleep(5000); // Non-deterministic, but whatever. This should never fail.
-    Assert.assertTrue(parentControllerClient.getStore(storeName).isError());
+    TestUtils.waitForNonDeterministicAssertion(TIMEOUT, TimeUnit.MILLISECONDS, () -> {
+      Assert.assertTrue(parentControllerClient.getStore(storeName).isError());
+    });
 
     parentControllerClient.skipAdminMessage(Long.toString(badOffset), false);
     TestUtils.waitForNonDeterministicAssertion(TIMEOUT * 3, TimeUnit.MILLISECONDS, () -> {
--- a/internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/controller/kafka/consumer/AdminConsumptionTaskIntegrationTest.java
+++ b/internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/controller/kafka/consumer/AdminConsumptionTaskIntegrationTest.java
@@ -371,6 +371,7 @@ public class AdminConsumptionTaskIntegrationTest {
       String owner,
       long executionId,
       int writerSchemaId) {
+    // TODO: Check back this step when we introduce the semantic check for new version in milestone 3
     UpdateStore updateStore = (UpdateStore) AdminMessageType.UPDATE_STORE.getNewInstance();
     updateStore.clusterName = clusterName;
     updateStore.storeName = storeName;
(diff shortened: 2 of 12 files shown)


services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/protocol/serializer/AdminOperationSerializer.java
```diff
@@ -5,50 +5,81 @@
...
+    if (writerSchemaId == LATEST_SCHEMA_ID_FOR_ADMIN_OPERATION) {
+      return serializedBytes;
     }
+
+    // Get the writer schema.
+    Schema writerSchema = getSchema(writerSchemaId);
+
+    // If writerSchema is not the latest schema, we need to deserialize the serialized bytes to GenericRecord with
+    // the writer schema, then serialize it to bytes with the writer schema.
+    GenericRecord genericRecord = deserialize(serializedBytes, LATEST_SCHEMA, writerSchema);
```
huangminchn, line 76: We used writerSchema on the readerSchema argument position of the deserialize API, which is correct but could be confusing.
Talked to @minhmo1620 ; we are going to change it to `targetSchema`

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/controller/kafka/consumer/AdminConsumptionTaskIntegrationTest.java
```diff
@@ -243,6 +272,73 @@ public void testUpdateAdminOperationVersion() {
     }
   }
 
+  @Test(timeOut = 2 * TIMEOUT)
+  public void testAdminConsumptionTaskWithSpecificWriterId() throws IOException {
+    int adminConsumptionMaxWorkerPoolSize = 3;
+    // Use a specific version to test the serialization and deserialization of admin operation.
+    int writerSchemaId = AdminOperationSerializer.LATEST_SCHEMA_ID_FOR_ADMIN_OPERATION - 10;
```
huangminchn, line 279: Let's change it to `AdminOperationSerializer.LATEST_SCHEMA_ID_FOR_ADMIN_OPERATION - 1` please, so that when dev adds a new version, we will test it against the previous latest schema directly.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/controller/kafka/consumer/AdminConsumptionTaskIntegrationTest.java
```diff
@@ -243,6 +272,73 @@ public void testUpdateAdminOperationVersion() {
...
+                    .numberOfParentControllers(1)
+                    .numberOfChildControllers(1)
+                    .numberOfServers(1)
+                    .numberOfRouters(1)
+                    .replicationFactor(1)
+                    .parentControllerProperties(parentControllerProps)
+                    .build());
+        ControllerClient parentControllerClient = new ControllerClient(
+            venice.getClusterNames()[0],
+            venice.getParentControllers().get(0).getControllerUrl())) {
```
huangminchn, line 299: Could you please make it a shared cluster for this test file?

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/controller/kafka/consumer/AdminConsumptionTaskIntegrationTest.java
```diff
@@ -243,6 +272,73 @@ public void testUpdateAdminOperationVersion() {
...
+      // Setup
+      String clusterName = venice.getClusterNames()[0];
+      Admin admin = venice.getParentControllers().get(0).getVeniceAdmin();
+      PubSubTopicRepository pubSubTopicRepository = admin.getPubSubTopicRepository();
+      TopicManager topicManager = admin.getTopicManager();
+      PubSubTopic adminTopic = pubSubTopicRepository.getTopic(AdminTopicUtils.getTopicNameFromClusterName(clusterName));
+      topicManager.createTopic(adminTopic, 1, 1, true);
+      String storeName = "test-store";
```
huangminchn, line 307: Recommend using Utls.randomString.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/controller/kafka/consumer/AdminConsumptionTaskIntegrationTest.java
```diff
@@ -319,6 +418,30 @@ private byte[] getStoreDeletionMessage(String clusterName, String storeName, lon
...
+      int writerSchemaId) {
+    UpdateStore updateStore = (UpdateStore) AdminMessageType.UPDATE_STORE.getNewInstance();
+    updateStore.clusterName = clusterName;
+    updateStore.storeName = storeName;
+    updateStore.owner = owner;
+    updateStore.partitionNum = 20;
+    updateStore.currentVersion = AdminConsumptionTask.IGNORED_CURRENT_VERSION;
+    updateStore.enableReads = true;
+    updateStore.enableWrites = true;
+    updateStore.replicateAllConfigs = true;
```
huangminchn, line 439: Could you please leave a TODO message here and see if the tests starting to fail once we introduce new fields semantic checks in milestone 3?

services/venice-controller/src/test/java/com/linkedin/venice/controller/kafka/protocol/admin/AdminOperationSerializerTest.java
```diff
@@ -0,0 +1,92 @@
...
+      adminMessage.payloadUnion = updateStore;
+      adminMessage.executionId = 1;
+
+      doCallRealMethod().when(adminOperationSerializer).serialize(any(), anyInt());
+      doCallRealMethod().when(adminOperationSerializer).deserialize(any(), anyInt());
+
+      // Serialize the AdminOperation object with writer schema id v74
+      byte[] serializedBytes = adminOperationSerializer.serialize(adminMessage, 74);
+
+      // Deserialize the serialized bytes back into an AdminOperation object
```
huangminchn, line 76: Let's leave a TODO message here to test that all the new fields only have default values, after the final deserialization.

services/venice-controller/src/test/java/com/linkedin/venice/controller/TestVeniceParentHelixAdmin.java
```diff
@@ -301,7 +299,7 @@ public void testAddStore() {
     when(zkClient.readData(zkMetadataNodePath, null)).thenReturn(null)
         .thenReturn(
             AdminTopicMetadataAccessor
-                .generateMetadataMap(Optional.of(1L), Optional.of(-1L), Optional.of(1L), Optional.empty()));
+                .generateMetadataMap(Optional.of(1L), Optional.of(-1L), Optional.of(1L), Optional.of(84L)));
```
huangminchn, line 302: Let's replace the hardcode version number to the latest version number in AvroProtocolDefinition.

clients/venice-admin-tool/src/test/java/com/linkedin/venice/TestAdminToolConsumption.java
```diff
@@ -99,8 +99,9 @@ private List<PubSubMessage<KafkaKey, KafkaMessageEnvelope, Long>> prepareAdminPu
       adminMessage.operationType = AdminMessageType.STORE_CREATION.getValue();
       adminMessage.payloadUnion = storeCreation;
       adminMessage.executionId = 1;
-      deserializer.serialize(adminMessage);
-      byte[] putValueBytes = deserializer.serialize(adminMessage);
+      deserializer.serialize(adminMessage, AdminOperationSerializer.LATEST_SCHEMA_ID_FOR_ADMIN_OPERATION);
```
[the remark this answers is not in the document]
minhmo1620: removed the duplicated line.

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/protocol/serializer/AdminOperationSerializer.java
```diff
@@ -69,4 +100,41 @@ public static Map<Integer, Schema> initProtocolMap() {
...
+  private <T> byte[] serialize(T object, Schema writerSchema) {
+    try {
+      GenericDatumWriter<T> datumWriter = new GenericDatumWriter<>(writerSchema);
+      ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream();
+      Encoder encoder = AvroCompatibilityHelper.newBinaryEncoder(byteArrayOutputStream, true, null);
+      datumWriter.write(object, encoder);
+      encoder.flush();
+      return byteArrayOutputStream.toByteArray();
+    } catch (IOException e) {
+      throw new VeniceMessageException("Could not serialize object", e);
```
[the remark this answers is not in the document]
minhmo1620: Updated to both serializer and deserializer error message.

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/protocol/serializer/AdminOperationSerializer.java
minhmo1620: Updated. Please help check the new signature for serializer/deserializer.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/controller/kafka/consumer/AdminConsumptionTaskIntegrationTest.java
minhmo1620: Changed to verify for the latest 5 versions.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/controller/kafka/consumer/AdminConsumptionTaskIntegrationTest.java
minhmo1620: Updated, bring the init into setup()

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/controller/kafka/consumer/AdminConsumptionTaskIntegrationTest.java
minhmo1620: Added.

services/venice-controller/src/test/java/com/linkedin/venice/controller/kafka/protocol/admin/AdminOperationSerializerTest.java
minhmo1620: Added.

services/venice-controller/src/test/java/com/linkedin/venice/controller/TestVeniceParentHelixAdmin.java
minhmo1620: Updated.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
```diff
@@ -627,10 +627,16 @@ private void sendAdminMessageAndWaitForConsumed(String clusterName, String store
             adminCommandExecutionTracker.createExecution(AdminMessageType.valueOf(message).name());
         message.executionId = execution.getExecutionId();
         VeniceWriter<byte[], byte[], byte[]> veniceWriter = veniceWriterMap.get(clusterName);
-        byte[] serializedValue = adminOperationSerializer.serialize(message);
+        Map<String, Long> metadata = adminTopicMetadataAccessor.getMetadata(clusterName);
```
gaojieliu, line 630: In practice, all the Controllers are being shared by all the clusters, so when the Controllers are fully deployed, all the clusters will receive the upgrade at the same time.
It is fine to maintain the writer schema id per cluster, but it would cause some maintenance overhead and potentially, some misleading when adding new clusters.
I wondered whether we can use a separate Znode, which is outside of any cluster instead or not, and with this way, when updating the writer schema id via tooling, we don't need to update it cluster by cluster.

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/protocol/serializer/AdminOperationSerializer.java
```diff
@@ -69,4 +104,46 @@ public static Map<Integer, Schema> initProtocolMap() {
...
+   */
+  private GenericRecord deserialize(byte[] serializedRecord, Schema readerSchema, int readerSchemaId) {
+    try {
+      GenericDatumReader<GenericRecord> datumReader = new GenericDatumReader<>(LATEST_SCHEMA, readerSchema);
+      InputStream in = new ByteArrayInputStream(serializedRecord);
+      BinaryDecoder decoder = AvroCompatibilityHelper.newBinaryDecoder(in, true, null);
+      return datumReader.read(null, decoder);
+    } catch (IOException e) {
+      throw new VeniceMessageException(
+          "Could not deserialize bytes back into GenericRecord object with reader version: " + readerSchemaId,
```
gaojieliu, line 145: Can we combine these two functions into one function called something like: `serializeWithDownConversion`?
These two functions are sharing the same names as the existing ones, which can cause confusions, and it seems no user will use the new functions separately, so I think consolidating them with a different name to make it less confusing.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
gaojieliu: Also shall we embed this writer schema retrieval logic into `AdminOperationSerializer`? so that the caller doesn't need to bother about how to fetch the correct writer schema id, which will keep `serialize` interface same as before and internally, it will figure out which writer schema is to use.
Not a big concern, and mainly for better encapsulation and reduce the complexity of `VeniceParentHelixAdmin`, which is already lengthy today.

gaojieliu (commented):
Code change looks good overall and left some craftsmanship/operation related comments.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
minhmo1620: > shall we embed this writer schema retrieval logic into AdminOperationSerializer
I need to send the version along with the message itself ([code](https://github.com/linkedin/venice/pull/1549/files#diff-ed8e93ee89eeb6383b0604c2d5cc0bb816bce922d0e66cd593a51b88e62b67b2R639)), so I don't think we can put the get inside AdminOperationSerializer.

minhmo1620: > I wondered whether we can use a separate Znode, which is outside of any cluster instead or not, and with this way, when updating the writer schema id via tooling, we don't need to update it cluster by cluster.

This is a valid suggestion. I don't think we will cover that change in this PR. We can discuss more on this offline. Since the ZK needs to sit somewhere, I am not really clear whether we can get it out of "cluster" concept. Do we have any existing usage for the znode you are talking about?

gaojieliu: Regarding ZNode, we can certainly can create a new Znode outside of any cluster and currently `storeConfigs` are outside of any cluster since the info stored is global and I think the admin operation schema id is also global info and we can create a new znode in the same layer as `storeConfigs`.

gaojieliu: > shall we embed this writer schema retrieval logic into AdminOperationSerializer
> 
> I need to send the version along with the message itself (code), so I don't think we can put the get inside AdminOperationSerializer.
We can let `serialize` function return a pair to keep both bytes and writer schema id.
It is not a big concern, and it is up to you.
The main goal is to reduce the complexity of `VeniceParentHelixAdmin` and make the code more modular.

minhmo1620: Thanks @gaojieliu , let me discuss more with @huangminchn on this and get back to you offline. I will send out follow-up PR if we decide to move things. 

cc: @huangminchn for the option for znode location.

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/protocol/serializer/AdminOperationSerializer.java
minhmo1620: Discussed offline. We keep the `serialize` method to avoid duplication, while cut the `deserialize` to the main method. I have updated the code, please help check again. Thanks!

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
minhmo1620: > > shall we embed this writer schema retrieval logic into AdminOperationSerializer
> > I need to send the version along with the message itself (code), so I don't think we can put the get inside AdminOperationSerializer.
> 
> We can let `serialize` function return a pair to keep both bytes and writer schema id. It is not a big concern, and it is up to you. The main goal is to reduce the complexity of `VeniceParentHelixAdmin` and make the code more modular.

Discussed offline. I decided to keep it as it is since other options do not appear strongly better than the current option.

gaojieliu: What about the global write schema instead of per-cluster?
Have you got a chance to talk to @huangminchn regarding ^?

huangminchn: We synced up offline on this topic, and decided to keep per cluster config to reduce blast radius. However, tooling will be provided to update all clusters at the same time.

huangminchn (approved):
Thanks for the hardwork, Minh!

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/controller/kafka/consumer/AdminConsumptionTaskIntegrationTest.java
```diff
@@ -57,285 +60,222 @@ public class AdminConsumptionTaskIntegrationTest {
...
+        storeName,
+        owner,
+        keySchema,
+        valueSchema,
+        nextExecutionId(),
+        AdminOperationSerializer.LATEST_SCHEMA_ID_FOR_ADMIN_OPERATION);
+    writer.put(new byte[0], goodMessage, AdminOperationSerializer.LATEST_SCHEMA_ID_FOR_ADMIN_OPERATION);
+
+    Utils.sleep(5000); // Non-deterministic, but whatever. This should never fail.
+    Assert.assertTrue(parentControllerClient.getStore(storeName).isError());
```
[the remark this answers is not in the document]
minhmo1620: This comes from the original test so I did not change. If you think we should just clean it up, I can do that.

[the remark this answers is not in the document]
minhmo1620: Updated! Thank you for your suggestion.

huangminchn (approved):
Thanks for the hard work, Minh!

PR #1550, opened 2025-02-22
[changelog][client] Some clean up and bug fixes in CDC

ZacAttack:
## [changelog][client] Some clean up and bug fixes in CDC
A few fixes that include:

**RMDUtils.hasOffsetAdvanced**: This method did not handle correctly some cases where the offset vectors dimensionality didn't line up 

**VeniceAfterImageConsumer**: Version swap could get stuck if the store repo for in the internal consumer and external consumer were out of sync. This change adds a new method where the outer consumer can instruct the inner consumer directly which version it wants to seek on.

**Heartbeats in CDC**: There was it seemed a forgotten and not so clever lambda attempt to record the timestamp.  This did some unfortunate math and failed to perform a System.currentTimeMillis - timestamp, which in turn recorded a very large value. This fixes that and it now reports the lag correctly.

Resolves #XXX

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceAfterImageConsumerImpl.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceAfterImageConsumerImpl.java
@@ -105,8 +105,7 @@ public class VeniceAfterImageConsumerImpl<K, V> extends VeniceChangelogConsumerI
     return internalSeekToTail(partitions, "");
   }
 
-  @Override
-  public CompletableFuture<Void> seekToEndOfPush(Set<Integer> partitions) {
+  protected CompletableFuture<Void> internalSeekToEndOfPush(Set<Integer> partitions, PubSubTopic targetTopic) {
     if (partitions.isEmpty()) {
       return CompletableFuture.completedFuture(null);
     }
@@ -118,7 +117,7 @@ public class VeniceAfterImageConsumerImpl<K, V> extends VeniceChangelogConsumerI
           // we'd like to do is instead add the offset of the EOP message in the VT, and then just seek to that offset.
           // We'll do that in a future patch.
           internalSeekConsumer.get().unsubscribeAll();
-          internalSeekConsumer.get().subscribe(partitions).get();
+          internalSeekConsumer.get().internalSubscribe(partitions, targetTopic).get();
 
           // We need to get the internal consumer as we have to intercept the control messages that we would normally
           // filter out from the user
@@ -134,9 +133,18 @@ public class VeniceAfterImageConsumerImpl<K, V> extends VeniceChangelogConsumerI
           }
 
           // poll until we get EOP for all partitions
+          LOGGER.info("Polling for EOP messages for partitions: " + partitions.toString());
           synchronized (consumerAdapter) {
+            LOGGER.info("GOT LOCK");
+            int counter = 0;
             while (true) {
+              counter++;
+              if (counter > 500) {
+                LOGGER.info("This is taking a while.....");
+              }
+              LOGGER.info("POLLING " + consumerAdapter.getAssignment().toString());
               polledResults = consumerAdapter.poll(5000L);
+              LOGGER.info("POLLED");
               // Loop through all polled messages
               for (Map.Entry<PubSubTopicPartition, List<PubSubMessage<KafkaKey, KafkaMessageEnvelope, Long>>> entry: polledResults
                   .entrySet()) {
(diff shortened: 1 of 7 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceAfterImageConsumerImpl.java
```diff
@@ -134,9 +133,18 @@ public CompletableFuture<Void> seekToEndOfPush(Set<Integer> partitions) {
...
 
           // poll until we get EOP for all partitions
+          LOGGER.info("Polling for EOP messages for partitions: " + partitions.toString());
           synchronized (consumerAdapter) {
+            LOGGER.info("GOT LOCK");
+            int counter = 0;
             while (true) {
+              counter++;
+              if (counter > 500) {
+                LOGGER.info("This is taking a while.....");
```
xunyin8, line 143: Should we add a bit more info for this log? e.g. the partitions we are still waiting for in `endOfPushConsumedPerPartitionMap`?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceAfterImageConsumerImpl.java
```diff
@@ -134,9 +133,18 @@ public CompletableFuture<Void> seekToEndOfPush(Set<Integer> partitions) {
...
+              }
+              LOGGER.info("POLLING " + consumerAdapter.getAssignment().toString());
```
xunyin8, line 145: Was this for debugging? Do we need to log this for every poll?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VersionSwapDataChangeListener.java
```diff
@@ -69,7 +71,11 @@ public void handleStoreChanged(Store store) {
...
         LOGGER.info(
             "New Version detected!  Seeking consumer to version: " + currentVersion + " in consumer: " + consumerName);
-        this.consumer.seekToEndOfPush(partitions).get();
+        this.consumer
+            .internalSeekToEndOfPush(
+                partitions,
+                pubSubTopicRepository.getTopic(store.getVersion(currentVersion).kafkaTopicName()))
+            .get();
       } catch (Exception e) {
         LOGGER.error(
```
xunyin8, line 80: Should we try to do a bit more than just log the error here since there could be transient errors and we might not get another store change event for a while (maybe around the time when the currently subscribed VT is getting deleted). We could schedule or perform an immediate retry attempt to check for new current version and perform version swap?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceAfterImageConsumerImpl.java
```diff
@@ -164,12 +173,15 @@ public CompletableFuture<Void> seekToEndOfPush(Set<Integer> partitions) {
...
                 // We polled all EOP messages, stop polling!
                 break;
               }
             }
           }
+          LOGGER.info("Seeking to EOP for partitions: " + partitions.toString());
           this.seekToCheckpoint(checkpoints).get();
+          LOGGER.info("Seeked to EOP for partitions: " + partitions.toString());
         } catch (InterruptedException | ExecutionException | VeniceCoordinateOutOfRangeException e) {
           throw new VeniceException(
```
xunyin8, line 186: What will happen if the VT gets deleted before we can reach EOP for all partitions? Will it be caught here and handled gracefully?

xunyin8 (commented):
Thanks for the quick debug and fix, left some comments. Mostly questions.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceAfterImageConsumerImpl.java
ZacAttack: oh crap, let me remove those, they were something I was putting in for debugging

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceAfterImageConsumerImpl.java
ZacAttack: it was for debugging.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceAfterImageConsumerImpl.java
ZacAttack: Probably not.  This will bubble up an exception, but the storeRepository only logs exceptions for callbacks, it doesn't try again.  So what I'll do is I'll add a retry to the top of the callback in the VersionSwapDetector class

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VersionSwapDataChangeListener.java
ZacAttack: yeah let's do that

xunyin8 (approved):
Thanks for the fix Zac.

PR #1551, opened 2025-02-25
[server][da-vinci-client] Return proper error code from getTopicPartitionEndOffSet

majisourav99:
##  Return proper error code from getTopicPartitionEndOffSet

 Return value from `getTopicPartitionEndOffSet` used to indicate the status of the invalid/non-existent topic vs valid LOWEST_OFFSET return value. But in PR https://github.com/linkedin/venice/pull/1232 that assumption was violated. This caused some partitions to think it has caught up even when it has not due to some transient issues. This PR fixes that.

## How was this PR tested?

GHCI

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
@@ -2413,12 +2413,9 @@ public abstract class StoreIngestionTask implements Runnable, Closeable {
           Duration.ofMillis(500),
           Duration.ofSeconds(5),
           Arrays.asList(PubSubTopicDoesNotExistException.class, PubSubOpTimeoutException.class, VeniceException.class));
-    } catch (PubSubTopicDoesNotExistException | PubSubOpTimeoutException e) {
+    } catch (Exception e) {
       LOGGER.error("Failed to get end offset for topic-partition: {} even after 10 retries", topicPartition, e);
       return StatsErrorCode.LAG_MEASUREMENT_FAILURE.code;
-    } catch (Exception e) {
-      LOGGER.error("Could not find latest offset for {} even after 10 retries", pubSubTopic.getName());
-      return -1;
     }
   }
 
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
@@ -2407,7 +2407,7 @@ public abstract class StoreIngestionTask implements Runnable, Closeable {
           throw new VeniceException("Latest offset is unknown. Check if the topic: " + topicPartition + " exists.");
         }
         return offset;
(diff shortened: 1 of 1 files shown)


kvargha (changes requested):
Can we log the kafkaUrl on failure too. It will help us debug why it's failing in the first place.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -2405,9 +2408,12 @@ protected long getTopicPartitionEndOffSet(String kafkaUrl, PubSubTopic pubSubTop
...
           Duration.ofMillis(500),
           Duration.ofSeconds(5),
-          Collections.singletonList(VeniceException.class));
+          Arrays.asList(PubSubTopicDoesNotExistException.class, PubSubOpTimeoutException.class, VeniceException.class));
+    } catch (PubSubTopicDoesNotExistException | PubSubOpTimeoutException e) {
+      LOGGER.error("Failed to get end offset for topic-partition: {} even after 10 retries", topicPartition, e);
+      return StatsErrorCode.LAG_MEASUREMENT_FAILURE.code;
     } catch (Exception e) {
-      LOGGER.error("Could not find latest offset for {} even after 5 retries", pubSubTopic.getName());
+      LOGGER.error("Could not find latest offset for {} even after 10 retries", pubSubTopic.getName());
```
sushantmane, line 2416: let's log PubSubTopicPartition

Besides timeouts are pretty tight on this API IMHO

kvargha (dismissed):

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -2405,10 +2408,14 @@ protected long getTopicPartitionEndOffSet(String kafkaUrl, PubSubTopic pubSubTop
           Duration.ofMillis(10),
           Duration.ofMillis(500),
           Duration.ofSeconds(5),
-          Collections.singletonList(VeniceException.class));
+          Arrays.asList(PubSubTopicDoesNotExistException.class, PubSubOpTimeoutException.class, VeniceException.class));
```
sushantmane, line 2411: Let's make this static and capture all Exceptions

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -2400,15 +2403,14 @@ protected long getTopicPartitionEndOffSet(String kafkaUrl, PubSubTopic pubSubTop
           throw new VeniceException("Latest offset is unknown. Check if the topic: " + topicPartition + " exists.");
         }
         return offset;
-      },
-          10,
-          Duration.ofMillis(10),
-          Duration.ofMillis(500),
-          Duration.ofSeconds(5),
-          Collections.singletonList(VeniceException.class));
+      }, 10, Duration.ofMillis(10), Duration.ofMillis(500), Duration.ofSeconds(5), RETRY_FAILURE_TYPES);
```
kvargha, line 2406: Nit: Can we move the number of retries to a static variable, so that when we change it it'll automatically reflect in the error logs too

kvargha (changes requested):

kvargha (approved):

PR #1552, opened 2025-02-25
[controller] Fix dvc heartbeat value in store and version config

misyel:

The current dvc heartbeat is set from checking if there is a dvc push status and if the status is `NOT_CREATED` This, however, is inaccurate because the status is set to `NOT_STARTED` when there is no status found  if we're under the num of polling retry attempts. We should, instead, check if `daVinciStatus.isNoDaVinciStatusReport` returns false before marking if there is a heartbeat reported and mark it as false otherwise if we've exceeded the dvc status retry attempts

Other minor changes included:
* Gate dvc heartbeat check behind controller config `skip.deferred.version.swap.for.dvc.enabled`
* Do not set target regions for stores with a dvc heartbeat reported if `skip.deferred.version.swap.for.dvc.enabled` is enabled
* Check version target region config instead of store target region config in deferred version swap service

## How was this PR tested?

Unit and e2e tests

### DIFF
--- a/internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
@@ -2456,10 +2456,22 @@ public class ConfigKeys {
    */
   public static final String CONTROLLER_HELIX_RESOURCE_CAPACITY_WEIGHT = "controller.helix.default.instance.capacity";
 
+  /**
+   * Specifies how frequently the deferred version swap service runs. Default value is 1 minute
+   */
   public static final String CONTROLLER_DEFERRED_VERSION_SWAP_SLEEP_MS = "controller.deferred.version.swap.sleep.ms";
+
+  /**
+   * Enables / disables the deferred version swap service. Default value is disabled
+   */
   public static final String CONTROLLER_DEFERRED_VERSION_SWAP_SERVICE_ENABLED =
       "controller.deferred.version.swap.service.enabled";
 
+  /**
+   * Enables / disables allowing dvc clients to perform a target region push with deferred swap. When enabled, dvc clients
+   * will be skipped and target regions will not be set and the deferred version swap service will skip checking stores with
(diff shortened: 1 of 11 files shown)


internal/venice-common/src/main/java/com/linkedin/venice/ConfigKeys.java
```diff
@@ -2456,10 +2456,25 @@ private ConfigKeys() {
...
   public static final String CONTROLLER_DEFERRED_VERSION_SWAP_SLEEP_MS = "controller.deferred.version.swap.sleep.ms";
+
+  /**
+   * Enables / disables the deferred version swap service. Default value is disabled
+   */
   public static final String CONTROLLER_DEFERRED_VERSION_SWAP_SERVICE_ENABLED =
       "controller.deferred.version.swap.service.enabled";
 
+  /**
+   * Enables / disables allowing dvc clients to perform a target region push with deferred swap. When enabled, dvc clients
```
majisourav99, line 2471: So this is temporary till we support delayed DVC ingestion right? can you mention that

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceControllerClusterConfig.java
```diff
@@ -555,6 +556,7 @@ public class VeniceControllerClusterConfig {
   private boolean isHybridStorePartitionCountUpdateEnabled;
   private final long deferredVersionSwapSleepMs;
   private final boolean deferredVersionSwapServiceEnabled;
+  private final boolean deferredVersionSwapServiceWithDvcHeartbeatEnabled;
```
majisourav99, line 559: nit: rename to something like `skipDeferredVersionSwapForDVCEnabled` ?

majisourav99 (approved):
LGTM. Thanks @misyel

majisourav99 (dismissed):

majisourav99 (approved):

PR #1553, opened 2025-02-25
[router] Fix in-flight request check in router shutdown

majisourav99:
## Fix in-flight request draining in router shutdown

Fix: Properly Track In-Flight Requests During Router Shutdown

When the router is shutting down, it checks for in-flight requests to ensure they are properly drained — otherwise, clients may encounter errors.

Currently, this uses a Rate sensor to increment and decrement the request count. However, calling record(-1) on a Rate sensor does not actually decrease the sensor’s rate value.  This can lead to incorrect tracking of in-flight requests and potentially leave requests unaccounted for during shutdown. 

This PR addresses that behavior to ensure request counts are accurately managed and clients don’t see unexpected errors during router shutdown by replacing the `Rate` with `Gauge` stat.

## How was this PR tested?

CI tests

### DIFF
--- a/services/venice-router/src/main/java/com/linkedin/venice/router/InFlightRequestStat.java
+++ b/services/venice-router/src/main/java/com/linkedin/venice/router/InFlightRequestStat.java
@@ -18,6 +18,7 @@ public class InFlightRequestStat {
   private final VeniceMetricsRepository localMetricRepo;
 
   private final Sensor totalInflightRequestSensor;
+  private final Metric metric;
 
   public InFlightRequestStat(VeniceRouterConfig config) {
     metricConfig = new MetricConfig().timeWindow(config.getRouterInFlightMetricWindowSeconds(), TimeUnit.SECONDS);
@@ -28,6 +29,7 @@ public class InFlightRequestStat {
             .build());
     totalInflightRequestSensor = localMetricRepo.sensor("total_inflight_request");
     totalInflightRequestSensor.add(TOTAL_INFLIGHT_REQUEST_COUNT, new Rate());
+    metric = localMetricRepo.getMetric(TOTAL_INFLIGHT_REQUEST_COUNT);
   }
 
   public Sensor getTotalInflightRequestSensor() {
@@ -35,7 +37,6 @@ public class InFlightRequestStat {
   }
 
   public double getInFlightRequestRate() {
-    Metric metric = localMetricRepo.getMetric(TOTAL_INFLIGHT_REQUEST_COUNT);
     // max return -infinity when there are no samples. validate only against finite value
     return Double.isFinite(metric.value()) ? metric.value() : 0.0;
   }
--- a/services/venice-router/src/main/java/com/linkedin/venice/router/InFlightRequestStat.java
+++ b/services/venice-router/src/main/java/com/linkedin/venice/router/InFlightRequestStat.java
@@ -1,0 +1,42 @@
+package com.linkedin.venice.router;
+
+import static com.linkedin.venice.router.RouterServer.ROUTER_SERVICE_METRIC_PREFIX;
+import static com.linkedin.venice.router.RouterServer.ROUTER_SERVICE_NAME;
+
+import com.linkedin.venice.stats.VeniceMetricsConfig;
+import com.linkedin.venice.stats.VeniceMetricsRepository;
+import io.tehuti.Metric;
+import io.tehuti.metrics.MetricConfig;
+import io.tehuti.metrics.Sensor;
+import io.tehuti.metrics.stats.Rate;
+import java.util.concurrent.TimeUnit;
+
+
+public class InFlightRequestStat {
+  public static final String TOTAL_INFLIGHT_REQUEST_COUNT = "total_inflight_request_count";
+  private final MetricConfig metricConfig;
+  private final VeniceMetricsRepository localMetricRepo;
+
+  private final Sensor totalInflightRequestSensor;
+
+  public InFlightRequestStat(VeniceRouterConfig config) {
+    metricConfig = new MetricConfig().timeWindow(config.getRouterInFlightMetricWindowSeconds(), TimeUnit.SECONDS);
+    localMetricRepo = new VeniceMetricsRepository(
+        new VeniceMetricsConfig.Builder().setServiceName(ROUTER_SERVICE_NAME)
+            .setMetricPrefix(ROUTER_SERVICE_METRIC_PREFIX)
+            .setTehutiMetricConfig(metricConfig)
+            .build());
+    totalInflightRequestSensor = localMetricRepo.sensor("total_inflight_request");
+    totalInflightRequestSensor.add(TOTAL_INFLIGHT_REQUEST_COUNT, new Rate());
+  }
+
+  public Sensor getTotalInflightRequestSensor() {
+    return totalInflightRequestSensor;
+  }
+
+  public double getInFlightRequestRate() {
+    Metric metric = localMetricRepo.getMetric(TOTAL_INFLIGHT_REQUEST_COUNT);
+    // max return -infinity when there are no samples. validate only against finite value
+    return Double.isFinite(metric.value()) ? metric.value() : 0.0;
+  }
+}
--- a/services/venice-router/src/main/java/com/linkedin/venice/router/RouterServer.java
+++ b/services/venice-router/src/main/java/com/linkedin/venice/router/RouterServer.java
@@ -859,10 +859,7 @@ public class RouterServer extends AbstractVeniceService {
         if (hasInFlightRequest()) {
           throw new VeniceException("There are still in-flight requests in router");
         }
-      },
-          10,
-          Duration.ofSeconds(config.getRouterNettyGracefulShutdownPeriodSeconds()),
-          Collections.singletonList(VeniceException.class));
+      }, 30, Duration.ofSeconds(1), Collections.singletonList(VeniceException.class));
     } catch (VeniceException e) {
       LOGGER.error(
           "There are still in-flight request during router shutdown, still continuing shutdown, it might cause unhealthy request in client");
--- a/services/venice-router/src/main/java/com/linkedin/venice/router/RouterServer.java
+++ b/services/venice-router/src/main/java/com/linkedin/venice/router/RouterServer.java
@@ -86,6 +86,7 @@ import com.linkedin.venice.service.AbstractVeniceService;
 import com.linkedin.venice.servicediscovery.ServiceDiscoveryAnnouncer;
 import com.linkedin.venice.stats.ThreadPoolStats;
 import com.linkedin.venice.stats.VeniceJVMStats;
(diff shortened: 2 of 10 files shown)


sushantmane (commented):
Actually you don't need to change sensor type right? You can just stop calling recording when the request is processed. This way after some time the rate will drop to zero and inflight request count will be zero. 


that being said, shouldn't it be a gauge type of metric is we want to track inflight requests properly?

sushantmane (commented):
Could you please do two small changes in this PR:
1. Can you replace sensor name: total_inflight_request_count literal with a defined constant?
2. Use Double.compare for real number comparison ? `Double.compare(metric.value(), 0.0) > 0)`

```
  static public boolean hasInFlightRequests() {
    Metric metric = localMetricRepo.getMetric("total_inflight_request_count");
    // max return -infinity when there are no samples. validate only against finite value
    return Double.isFinite(metric.value()) ? metric.value() > 0.0 : false;
  }
```

sushantmane (dismissed):
LGTM. Thanks!

services/venice-router/src/main/java/com/linkedin/venice/router/RouterServer.java
```diff
@@ -852,10 +852,11 @@ public void stopInner() throws Exception {
      * correctly.
      */
 
+    LOGGER.info("Waiting to make sure all in-flight requests are drained");
     // Graceful shutdown: Wait till all the requests are drained
     try {
       RetryUtils.executeWithMaxAttempt(() -> {
-        if (RouterHttpRequestStats.hasInFlightRequests()) {
+        if (hasInFlightRequest()) {
           throw new VeniceException("There are still in-flight requests in router");
```
eldernewborn, line 860: Do we have to throw exception ? 
this is a completely expected situation, can we just log that message ?

services/venice-router/src/main/java/com/linkedin/venice/router/stats/RouterHttpRequestStats.java
```diff
@@ -74,7 +75,7 @@ public class RouterHttpRequestStats extends AbstractVeniceHttpStats {
 
   private final static Sensor totalInflightRequestSensor = localMetricRepo.sensor("total_inflight_request");
   static {
-    totalInflightRequestSensor.add("total_inflight_request_count", new Rate());
+    totalInflightRequestSensor.add(TOTAL_INFLIGHT_REQUEST_COUNT, new Gauge());
```
gaojieliu, line 78: `Gauge` doesn't honor any timed window, but just store the most recent value.
So the increment/decrement won't work.

If it is not windowed, who will update the metric to be `0` then?

Removing `record(-1)` should just work and if we want to speed up the stop, we should adjust time window size for this particular metric and currently, it is 10s, which means the Router can only be stopped after 20s since no traffic (by default, there are 2 windows).

services/venice-router/src/main/java/com/linkedin/venice/router/RouterServer.java
majisourav99: yes we need to throw exception, thats how the retryUtils knows when to retry.

sushantmane (commented):
Could you add a unit test to verify the inflight request recording logic and ensure the max wait time for the rate to decrease does not exceed 2 * windowDuration?

sushantmane:
> Could you add a unit test to verify that inflight request recording behaves as expected, i.e., ensuring requests are tracked during the recording window, the rate remains non-zero until the recordingWindowDuration has elapsed, and the max wait time for the rate to drop to zero does not exceed 2 * recordingWindowDuration?



One possible (fee free to update as you see fit) test:
```java
@Test
public void testInflightRequestsGoDownAfterDefinedThreshold() throws InterruptedException {
  long startTime = System.currentTimeMillis();
  long deltaTime = 5000; // Keep recording for 5 seconds
  long recordingWindowDuration = 10_000; // Duration to expect inflight requests
  AtomicLong lastRecorded = new AtomicLong(0);
  CountDownLatch latch = new CountDownLatch(1);

  assertFalse(RouterHttpRequestStats.hasInFlightRequests());

  // Thread to simulate inflight request additions
  Thread inflightThread = new Thread(() -> {
    while (System.currentTimeMillis() < startTime + deltaTime) {
      RouterHttpRequestStats.totalInflightRequestSensor.record();
      lastRecorded.set(System.currentTimeMillis());
      latch.countDown();
    }
  });

  inflightThread.start();
  latch.await(); // Ensure at least one inflight request is recorded

  // Verify inflight requests exist during the expected window
  while (System.currentTimeMillis() < lastRecorded.get() + recordingWindowDuration) {
    assertTrue(RouterHttpRequestStats.hasInFlightRequests(),
        "Expected inflight requests - elapsed: " + (System.currentTimeMillis() - lastRecorded.get()));
  }
  System.out.println("A. Total elapsed time since last recorded: " + (System.currentTimeMillis() - lastRecorded.get()) + " total time elapsed: " + (System.currentTimeMillis() - startTime));

  // Wait until after the expected window to verify requests drop to zero
  Utils.sleep(2 * recordingWindowDuration);
  System.out.println("B. Total elapsed time since last recorded: " + (System.currentTimeMillis() - lastRecorded.get()) + " total time elapsed: " + (System.currentTimeMillis() - startTime));

  assertFalse(RouterHttpRequestStats.hasInFlightRequests(),
      "Expected no inflight requests - elapsed: " + (System.currentTimeMillis() - lastRecorded.get()));

  System.out.println("X. Total elapsed time since last recorded: " + (System.currentTimeMillis() - lastRecorded.get()) + " total time elapsed: " + (System.currentTimeMillis() - startTime));


  // Stop the recording thread
  inflightThread.interrupt();
  inflightThread.join();
}
```

services/venice-router/src/main/java/com/linkedin/venice/router/RouterServer.java
```diff
@@ -852,10 +852,11 @@ public void stopInner() throws Exception {
      * correctly.
      */
 
+    LOGGER.info("Waiting to make sure all in-flight requests are drained");
     // Graceful shutdown: Wait till all the requests are drained
     try {
       RetryUtils.executeWithMaxAttempt(() -> {
-        if (RouterHttpRequestStats.hasInFlightRequests()) {
+        if (hasInFlightRequest()) {
```
gaojieliu, line 859: The default retry delay is 10s, which means the total time taken to stop after draining all the ongoing requests can be 10s + 2 * time windows, which will be 20s, right?
can we reduce the retry delay, but increase the retry attempts? For example reduce the retry delay to 1s, but increase the retry attempts to 60?

Also how much time did you observe in your test case?

services/venice-router/src/main/java/com/linkedin/venice/router/RouterServer.java
```diff
@@ -852,20 +852,19 @@ public void stopInner() throws Exception {
...
-        if (RouterHttpRequestStats.hasInFlightRequests()) {
-          throw new VeniceException("There are still in-flight requests in router");
+        if (hasInFlightRequest()) {
+          throw new VeniceException("There are still in-flight requests in router :" + getInFlightRequestCount());
         }
-      },
-          10,
-          Duration.ofSeconds(config.getRouterNettyGracefulShutdownPeriodSeconds()),
-          Collections.singletonList(VeniceException.class));
+      }, 30, Duration.ofSeconds(1), Collections.singletonList(VeniceException.class));
```
gaojieliu, line 862: Shall we increase the retry attempts to 60 to be safe?

services/venice-router/src/main/java/com/linkedin/venice/router/stats/RouterHttpRequestStats.java
```diff
@@ -658,9 +658,15 @@ private Sensor registerSensorFinal(String sensorName, MeasurableStat... stats) {
   }
 
   static public boolean hasInFlightRequests() {
-    Metric metric = localMetricRepo.getMetric("total_inflight_request_count");
+    Metric metric = localMetricRepo.getMetric(TOTAL_INFLIGHT_REQUEST_COUNT);
```
gaojieliu, line 661: Should this function be instead?
```
return getInFlightRequestCount() > 0;
```

services/venice-router/src/main/java/com/linkedin/venice/router/RouterServer.java
```diff
@@ -852,20 +852,19 @@ public void stopInner() throws Exception {
      * correctly.
      */
 
+    LOGGER.info("Waiting to make sure all in-flight requests are drained");
     // Graceful shutdown: Wait till all the requests are drained
     try {
       RetryUtils.executeWithMaxAttempt(() -> {
-        if (RouterHttpRequestStats.hasInFlightRequests()) {
-          throw new VeniceException("There are still in-flight requests in router");
+        if (hasInFlightRequest()) {
```
gaojieliu, line 859: I think we should just make one call instead two:
```
long inFlightRequestCount = getInFlightRequestCount();
if (inFlightRequestCount > 0) {
    throw new VeniceException("There are still in-flight requests in router :" + inFlightRequestCount);
}
```
Two calls can return inconsistent results and we can completely remove function: `hasInFlightRequest()`.

services/venice-router/src/main/java/com/linkedin/venice/router/InFlightRequestStat.java
```diff
@@ -0,0 +1,42 @@
...
+    totalInflightRequestSensor = localMetricRepo.sensor("total_inflight_request");
+    totalInflightRequestSensor.add(TOTAL_INFLIGHT_REQUEST_COUNT, new Rate());
+  }
+
+  public Sensor getTotalInflightRequestSensor() {
+    return totalInflightRequestSensor;
+  }
+
+  public double getInFlightRequestRate() {
+    Metric metric = localMetricRepo.getMetric(TOTAL_INFLIGHT_REQUEST_COUNT);
```
gaojieliu, line 38: Nit, 
we can maintain a reference in class level to this metric, right?

services/venice-router/src/test/java/com/linkedin/venice/router/RouteHttpRequestStatsTest.java
```diff
@@ -20,20 +31,30 @@ public class RouteHttpRequestStatsTest {
...
 
+  private VeniceMetricsRepository localMetricRepo;
+
   @BeforeSuite
   public void setUp() {
     MetricsRepository metrics = MetricsRepositoryUtils.createSingleThreadedVeniceMetricsRepository();
     reporter = new MockTehutiReporter();
     metrics.addReporter(reporter);
-
+    MetricConfig metricConfig = new MetricConfig().timeWindow(1, TimeUnit.SECONDS);
```
gaojieliu, line 41: To avoid the duplicate code, we can use `InFlightRequestStat` to construct the in-flight-sensor here, right?

majisourav99: Sure

gaojieliu (approved):

PR #1554, opened 2025-02-26
[changelog][client] Add viewname to changelog client factory

ZacAttack:
## [changelog][client] Add viewname to changelog client factory

This is to enable cases where we need to be able to consume different views for different stores.

Resolves #XXX

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumerClientFactory.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumerClientFactory.java
@@ -71,12 +71,28 @@ public class VeniceChangelogConsumerClientFactory {
    * each consumer can only subscribe to certain partitions. Multiple such consumers can work in parallel.
    */
   public <K, V> VeniceChangelogConsumer<K, V> getChangelogConsumer(String storeName, String consumerId, Class clazz) {
-    return storeClientMap.computeIfAbsent(suffixConsumerIdToStore(storeName, consumerId), name -> {
+    return getChangelogConsumer(storeName, consumerId, clazz, globalChangelogClientConfig.getViewName());
+  }
+
+  public <K, V> VeniceChangelogConsumer<K, V> getChangelogConsumer(
+      String storeName,
+      String consumerId,
+      Class clazz,
+      String viewNameOverride) {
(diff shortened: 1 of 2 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumerClientFactory.java
```diff
@@ -71,12 +71,28 @@ public <K, V> VeniceChangelogConsumer<K, V> getChangelogConsumer(String storeNam
...
+  }
+
+  public <K, V> VeniceChangelogConsumer<K, V> getChangelogConsumer(
+      String storeName,
+      String consumerId,
+      Class clazz,
+      String viewNameOverride) {
+    String adjustedConsumerId;
+    if (!StringUtils.isEmpty(viewNameOverride)) {
+      adjustedConsumerId = consumerId + "-" + viewNameOverride;
```
xunyin8, line 84: If `consumerId` is `null` this concat will actually resolve to `null-<viewNameOverride>` which functionally I don't think it makes a difference when `consumerId` is `null` in:
```
  private String suffixConsumerIdToStore(String storeName, String consumerId) {
    return StringUtils.isEmpty(consumerId) ? storeName : storeName + "-" + consumerId;
  }
```
but I think we should still fix it to avoid confusion and potential side effects in the future.

kcepmada: +1 to simply mirroring the existing consumerName as key
- `suffixConsumerIdToStore(storeName + "-" + viewNameOverride.getClass().getSimpleName(), consumerId);` 

What about a similar overload for getBootstrappingChangelogConsumer?

ZacAttack: re: bootstrappingConsumer

I 'think' but aren't sure that there maybe other gotchas if we allow a bootstrapping consumer to consume from different views simultaneously in the same client.  I think I'd need to research that a bit, but I'm willing to be we don't handle that well w.r.t. the on disk persistence.  So until I'm sure I'd rather leave it out of the interface.

xunyin8 (approved):
Thanks Zac

PR #1558, opened 2025-02-27
[server][controller] Superset schema generation should carry default if new value schema does not have default value

sixpluszero:
## [server][controller] Superset schema generation should carry default if new value schema does not have default value
This issue was detected via an legacy check in Active/Active Partial Update code that resulted in ingestion error:
1. User created store with schema v1(no default), v2(has default)
2. User put data with old value schema. 
3. User enabled and disabled AAWC, but user continue to put data with new value schema (or doesn't matter new->old, as long as old value and new value is using different schema id) 
4. When the logic checks incoming write, since it has field level RMD already, it will go to AAWC path to check and there we need to resolve the superset schema and there it will check if superset(v1, v2) is v1's superset schema, and superset(v2,v1) is v2's superset schema. 
In current implementation, new schema's default will always overwrite old schema's, but if new schema does not have default, it will also erase existing default, which I think is wrong. With this, the check in 4 throws exception.

This PR fixes it by porting the old value schema's default, if new schema does not have default, during superset schema generation logic. Although it is guarded in partial update enabled case, this is to make sure other feature - read compute will still achieve correct superset schema.

## How was this PR tested?

New integration test and new unit test.

### DIFF
--- a/internal/venice-client-common/src/main/java/com/linkedin/venice/utils/AvroSupersetSchemaUtils.java
+++ b/internal/venice-client-common/src/main/java/com/linkedin/venice/utils/AvroSupersetSchemaUtils.java
@@ -15,6 +15,7 @@ import java.util.Objects;
 import java.util.stream.Collectors;
 import org.apache.avro.Schema;
 import org.apache.commons.lang.StringUtils;
+import org.apache.logging.log4j.LogManager;
 
 
 public class AvroSupersetSchemaUtils {
@@ -158,8 +159,18 @@ public class AvroSupersetSchemaUtils {
       if (fieldInExistingSchema != null) {
         fieldBuilder.setSchema(generateSupersetSchema(fieldInExistingSchema.schema(), fieldInNewSchema.schema()))
             .setDoc(fieldInNewSchema.doc() != null ? fieldInNewSchema.doc() : fieldInExistingSchema.doc());
+        if (!fieldInNewSchema.hasDefaultValue() && fieldInExistingSchema.hasDefaultValue()) {
+          fieldBuilder.setDefault(getFieldDefault(fieldInExistingSchema));
+        }
       }
-      fields.add(fieldBuilder.build());
+      Schema.Field geneartedField = fieldBuilder.build();
+      LogManager.getLogger(AvroSupersetSchemaUtils.class)
+          .info(
+              "generated: {} {} {}",
+              fieldInNewSchema.hasDefaultValue(),
+              geneartedField.hasDefaultValue(),
+              fieldInExistingSchema == null ? null : fieldInExistingSchema.hasDefaultValue());
+      fields.add(geneartedField);
     }
 
     for (Schema.Field fieldInExistingSchema: existingSchema.getFields()) {
--- a/internal/venice-client-common/src/main/java/com/linkedin/venice/utils/AvroSupersetSchemaUtils.java
+++ b/internal/venice-client-common/src/main/java/com/linkedin/venice/utils/AvroSupersetSchemaUtils.java
@@ -15,7 +15,6 @@ import java.util.Objects;
 import java.util.stream.Collectors;
 import org.apache.avro.Schema;
 import org.apache.commons.lang.StringUtils;
-import org.apache.logging.log4j.LogManager;
 
 
 public class AvroSupersetSchemaUtils {
@@ -142,9 +141,10 @@ public class AvroSupersetSchemaUtils {
   }
 
   /**
-   * Merge field schema from two schema object. The rule is: If a field exist in both new schema and old schema, we should
-   * generate the superset schema of these two versions of the same field, with new schema's information taking higher
-   * priority.
+   * Merge field schema from two schema object.
+   * The rule is: If a field exist in both new schema and old schema, we should generate the superset schema of these
+   * two versions of the same field, with new schema's information taking higher priority.
+   * For default value, if new schema does not have default value, we will still preserve the old default value.
    * @param newSchema new schema
    * @param existingSchema old schema
    * @return merged schema field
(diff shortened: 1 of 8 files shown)


internal/venice-client-common/src/main/java/com/linkedin/venice/utils/AvroSupersetSchemaUtils.java
```diff
@@ -158,8 +159,18 @@ private static List<Schema.Field> mergeFieldSchemas(Schema existingSchema, Schem
...
             .setDoc(fieldInNewSchema.doc() != null ? fieldInNewSchema.doc() : fieldInExistingSchema.doc());
+        if (!fieldInNewSchema.hasDefaultValue() && fieldInExistingSchema.hasDefaultValue()) {
+          fieldBuilder.setDefault(getFieldDefault(fieldInExistingSchema));
+        }
       }
-      fields.add(fieldBuilder.build());
+      Schema.Field geneartedField = fieldBuilder.build();
+      LogManager.getLogger(AvroSupersetSchemaUtils.class)
+          .info(
+              "generated: {} {} {}",
```
minhmo1620, line 169: Can you update this log to have more information such as `generated: {fieldName}, new schema has default value: {}, ...`?

internal/venice-client-common/src/main/java/com/linkedin/venice/utils/AvroSupersetSchemaUtils.java
```diff
@@ -158,8 +159,18 @@ private static List<Schema.Field> mergeFieldSchemas(Schema existingSchema, Schem
       if (fieldInExistingSchema != null) {
         fieldBuilder.setSchema(generateSupersetSchema(fieldInExistingSchema.schema(), fieldInNewSchema.schema()))
             .setDoc(fieldInNewSchema.doc() != null ? fieldInNewSchema.doc() : fieldInExistingSchema.doc());
+        if (!fieldInNewSchema.hasDefaultValue() && fieldInExistingSchema.hasDefaultValue()) {
```
minhmo1620, line 162: out of curiosity: I wonder whether this would agaisnt the purpose of the method since, in the doc, we mention that we prioritize the new schema over the old schema so we should respect the no default ? (You should update the doc for this new functionality if so I guess)

sixpluszero: Thank you for the review! I will think more about this and add more details in the doc in the next iteration.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/PartialUpdateTest.java
```diff
@@ -388,6 +388,86 @@ public void testIncrementalPushPartialUpdateClassicFormat() throws IOException {
...
+    }
+    VeniceClusterWrapper veniceClusterWrapper = childDatacenters.get(0).getClusters().get(CLUSTER_NAME);
+    SystemProducer veniceProducer = getSamzaProducer(veniceClusterWrapper, storeName, Version.PushType.STREAM);
+    String key = "highway";
+    // Send first value with old schema;
+    GenericRecord record = new GenericData.Record(schemaV1);
+    record.put("id", "101");
+    record.put("name", "U.S. 101");
+    sendStreamingRecord(veniceProducer, storeName, key, record);
+    // Send first value with new schema;
```
lluwm, line 450: typo: first -> second?

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/PartialUpdateTest.java
```diff
@@ -388,6 +388,86 @@ public void testIncrementalPushPartialUpdateClassicFormat() throws IOException {
...
+    GenericRecord recordV2 = new GenericData.Record(schemaV2);
+    recordV2.put("id", "280");
+    recordV2.put("name", "Interstate 280");
+    sendStreamingRecord(veniceProducer, storeName, key, recordV2);
```
lluwm, line 454: could you clarify that without the fix which write would fail, with old schema or with new schema, or both?

internal/venice-client-common/src/main/java/com/linkedin/venice/utils/AvroSupersetSchemaUtils.java
lluwm: Even looking at existing code, it doesn't seems to be always the case that new field overwrites the old. More like a union of the two and when there is any conflict, the new overwrites the old. Agree that we should update the javadoc to make it clear.

e.g. 
`setDoc(fieldInNewSchema.doc() != null ? fieldInNewSchema.doc() : fieldInExistingSchema.doc());`

internal/venice-client-common/src/main/java/com/linkedin/venice/utils/AvroSupersetSchemaUtils.java
sixpluszero: Thanks for pointing out! This is actually a debug log, I removed it.

internal/venice-client-common/src/main/java/com/linkedin/venice/utils/AvroSupersetSchemaUtils.java
sixpluszero: Yeah, updated the doc to reflect the change.

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/PartialUpdateTest.java
sixpluszero: It will failed 2nd PUT, I will add to the java doc line.

lluwm (approved):
Thanks @sixpluszero, LGTM!

PR #1561, opened 2025-02-27
[server] Do not register separate RT region heartbeat metrics for follower

sixpluszero:
## [server] Do not register separate RT region heartbeat metrics for follower
Separate HB metrics should only be applied to leader, not for follower. This will reduce unnecessary metric count.

## How was this PR tested?

New Unit Test.


ZacAttack (approved):
Looks good!  Thanks!

PR #1562, opened 2025-02-28
[dvc][server] Add config to control file receive/send speeds for blob transfer throttling

jingy-li:
## [dvc][server] Add config to control file receive/send speeds for blob transfer throttling

This PR includes several enhancements to improve blob transfer performance:

1. Speed Throttling: Previously, file transfer speeds were unregulated. With Netty's built-in GlobalChannelTrafficShapingHandler added to the pipeline, we now control the write/read speed at the network level using configurable read and write limits per host level.

2. Host Shuffling: Host information is now randomized before sending requests.

3. Connection Timeout: A maximum timeout (in milliseconds) has been set when establishing connection per host.

4. Channel Reuse Check: After establishing a connection, it verifies that no existing file or metadata handler is present, avoiding unintended reuse of an active channel.

5. Metadata Size Check: A validation step has been added to ensure that the metadata is transferred successfully.

## How was this PR tested?
integration test


clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/BlobTransferUtil.java
```diff
@@ -117,4 +144,24 @@ public static BlobTransferManager<Void> getP2PBlobTransferManagerForServerAndSta
...
+   */
+  static GlobalChannelTrafficShapingHandler getGlobalTrafficShapingHandler(
+      long blobTransferClientReadLimitBytesPerSec,
+      long blobTransferServiceWriteLimitBytesPerSec) {
+    LOGGER.info(
+        "Global traffic shaping configured with read limit: {} bytes/sec, write limit: {} bytes/sec",
+        blobTransferClientReadLimitBytesPerSec,
+        blobTransferServiceWriteLimitBytesPerSec);
+
+    return BlobTransferGlobalTrafficShapingHandlerHolder
```
gaojieliu, line 164: Do we need this function here?
Maybe we can just expose a `static` method in `BlobTransferGlobalTrafficShapingHandlerHolder` to return a singleton instance instead of spreading the logic in multiple places.

gaojieliu (commented):

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/BlobTransferGlobalTrafficShapingHandlerHolder.java
```diff
@@ -0,0 +1,56 @@
...
+      long readLimit,
+      long writeLimit) {
+    // Set limits before getting the instance to ensure they're used during initialization
+    BlobTransferGlobalTrafficShapingHandlerHolder.readLimit = readLimit;
+    BlobTransferGlobalTrafficShapingHandlerHolder.writeLimit = writeLimit;
+
+    GlobalChannelTrafficShapingHandler handler = GLOBAL_CHANNEL_TRAFFIC_SHAPING_HANDLER.get();
+
+    // Update limits in case the handler was already created with different limits
+    handler.configure(writeLimit, readLimit);
```
gaojieliu, line 52: Emm...
Changing the global limit per invocation is not desired, and do you know any use cases, which would do that?
The shared instance should remain immutable once it is created, otherwise, the behavior will be non-deterministic.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/P2PBlobTransferManagerFactory.java
```diff
@@ -0,0 +1,136 @@
...
+import com.linkedin.venice.client.store.ClientConfig;
+import com.linkedin.venice.helix.HelixCustomizedViewOfflinePushRepository;
+import com.linkedin.venice.meta.ReadOnlyStoreRepository;
+import io.netty.handler.traffic.GlobalChannelTrafficShapingHandler;
+import java.util.concurrent.CompletableFuture;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+
+public class P2PBlobTransferManagerFactory {
```
gaojieliu, line 21: Can we convert it into a builder pattern?
Some examples:
https://github.com/linkedin/venice/blob/main/clients/venice-client/src/main/java/com/linkedin/venice/fastclient/ClientConfig.java

With the builder pattern, I think we can merge the following two functions.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/BlobTransferGlobalTrafficShapingHandlerHolder.java
jingy-li: Yes, I was thinking if there are 2 instances in one host: one instance has a read/write speed, while another has a different speed. Maybe it is unnecessary to differentiate them. Remove this configure line.

gaojieliu: Two instances in one host should create two separated processes, so they are independent.

jingy-li: Got it, the handler.configure is removed.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/P2PBlobTransferManagerFactory.java
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    public P2PBlobTransferManagerFactoryBuilder setAggVersionedBlobTransferStats(
+        AggVersionedBlobTransferStats aggVersionedBlobTransferStats) {
+      this.aggVersionedBlobTransferStats = aggVersionedBlobTransferStats;
+      return this;
+    }
+
+    public P2PBlobTransferManagerFactory build() {
+      return new P2PBlobTransferManagerFactory(
```
gaojieliu, line 173: In theory, we can pass `builder` as a param to `P2PBlobTransferManagerFactory` constructor to avoid the long param list.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/blobtransfer/P2PBlobTransferManagerFactory.java
```diff
@@ -0,0 +1,183 @@
...
+    this.readOnlyStoreRepository = readOnlyStoreRepository;
+    this.storageEngineRepository = storageEngineRepository;
+    this.aggVersionedBlobTransferStats = aggVersionedBlobTransferStats;
+  }
+
+  /**
+   * Get the P2P blob transfer manager and start it.
+   * @return the blob transfer manager
+   */
+  public BlobTransferManager<Void> getP2PBlobTransferManagerAndStart() {
```
gaojieliu, line 70: Do we expect to create multiple `BlobTransferManager` per process?
If not, we may just build an instance per invocation, and no need to use the factory pattern IMO.
We can chat offline.

jingy-li: Sync offline, move the blob transfer manager init to contractor. Remove the getP2PBlobTransferManagerAndStart method.

gaojieliu (approved):
Thanks for the change!

PR #1566, opened 2025-03-03
[server][vpj][common][changelog] Chunking support for materialized view

xunyin8:
## [server][vpj][common][changelog] Chunking support for materialized view

1. Removed ComplexVeniceWriterAdapter and force CompositeVeniceWriter to use ComplexVeniceWriter for child writers. This is because we need a new write API for writing to materialized view in VPJ that will supply the partitions that it wrote to. This way VPJ and MaterializedViewWriter are both writing to materialized view topic using the same APIs (CompositeVeniceWriter).

2. Expanded ComplexVeniceWriter#complexPut to take a nullable Consumer<int[]> partitionConsumer to provide a partition array that it wrote to. This information is used to build the view partition map that will be attached to the records in source region version topic.

3. The view partition map is attached as a PubSubMessageHeader to the records. Introduced new PubSubMessageHeader: VENICE_VIEW_PARTITIONS_MAP_HEADER. Expanded LeaderMetadataWrapper to also carry the view partition map. VeniceWriter will write records with VENICE_VIEW_PARTITIONS_MAP_HEADER if the provided LeaderMetadataWrapper contains a view partition map. If the record is a large record then all chunks and the manifest are all expected to carry this view partition map.

4. In L/F SIT we will only expect the view partition map in the PubSubMessageHeader during NR pass-through. Ingestion exception will be thrown if corresponding partition info is not found for a given materialized view. In addition, we only expect PUT messages during NR pass-through.

5. Added new method forwardPut in ComplexVeniceWriter to write records or chunks of records to pre-determined materialized view topic partition(s) with new DIV but without any other processing. This is like the materialized view equivalent of NR pass-through write. Forwarded records will not contain view partition map. This is intentional so we only pay the overhead in NR source.

6. Introduced RocksDBChunkAssembler and fixed previous issues with the ChunkAssembler. There were two issues:
    a. Using valueBytes.array() instead of ByteUtils.extractByteArray(valueBytes) which could cause deser issues.
    b. Not retrieving the proper writer schema ID from the manifest when assembling chunks. Put.schemaId will be -20.
New ChangelogClientConfig rocksDBChunkAssemblerStoragePath to specify the storage path for the rocksDB buffer.

## How was this PR tested?
Integration test, unit tests will be added once we have consensus on the API changes


clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ActiveActiveStoreIngestionTask.java
```diff
@@ -674,14 +674,15 @@ protected void processMessageAndMaybeProduceToKafka(
         Lazy<GenericRecord> valueProvider = mergeConflictResultWrapper.getValueProvider();
         queueUpVersionTopicWritesWithViewWriters(
             partitionConsumptionState,
-            (viewWriter) -> viewWriter.processRecord(
+            (viewWriter, ignored) -> viewWriter.processRecord(
```
lusong64, line 677: I believe you must have a legit reason not using ignored while adding it. Mind if adding a one liner comment to explain it?

xunyin8: Sure, will add it as an in-line comment too. Basically the view partition set is only expected and used during L/F SIT's NR pass-through mode. A/A SIT's ingestion of the RT topic will never expect/encounter records with view partition set info provided as part of their pub sub message headers.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/task/datawriter/CompositeVeniceWriter.java
```diff
@@ -1,41 +1,66 @@
...
 
   public CompositeVeniceWriter(
       String topicName,
       VeniceWriter<K, V, U> mainWriter,
-      AbstractVeniceWriter<K, V, U>[] childWriters,
-      PubSubProducerCallback childCallback) {
+      ComplexVeniceWriter<K, V, U>[] childWriters,
+      PubSubProducerCallback childCallback,
+      BiFunction<V, Integer, GenericRecord> deserializeFunction,
+      Function<V, V> decompressFunction) {
```
gaojieliu, line 54: Nit,
Can we consolidate these two params into one called something like `valueExtractor`?

internal/venice-common/src/main/java/com/linkedin/venice/writer/VeniceWriter.java
```diff
@@ -1611,6 +1643,20 @@ public static PubSubMessageHeader getLeaderCompleteStateHeader(LeaderCompleteSta
     return new PubSubMessageHeader(VENICE_LEADER_COMPLETION_STATE_HEADER, val);
   }
 
+  public static PubSubMessageHeader getViewDestinationPartitionHeader(
```
gaojieliu, line 1646: Can we move this util function to `ViewUtils` to reduce the length of this file?

gaojieliu (commented):
Code change looks good to me and left some minor comments.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -3373,19 +3376,45 @@ protected void processMessageAndMaybeProduceToKafka(
...
+          viewPartitionMap,
           produceToVersionTopic);
     } else {
       produceToVersionTopic.run();
     }
   }
 
+  protected Map<String, Set<Integer>> extractViewPartitionMap(PubSubMessageHeaders pubSubMessageHeaders) {
+    Map<String, Set<Integer>> viewPartitionMap = null;
+    for (PubSubMessageHeader header: pubSubMessageHeaders) {
```
ZacAttack, line 3398: Do we really have to iterate through these? PubsubMessageHeaders already has a map in it.  Couldn't we just ask for that header specifically?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/view/MaterializedViewWriter.java
```diff
@@ -67,34 +67,35 @@ public CompletableFuture<Void> processRecord(
...
       int newValueSchemaId,
-      boolean isChunkedKey,
+      Set<Integer> viewPartitionSet,
       Lazy<GenericRecord> newValueProvider) {
-    byte[] viewTopicKey = key;
-    if (isChunkedKey) {
-      viewTopicKey = keyWithChunkingSuffixSerializer.getKeyFromChunkedKey(key);
-    }
     byte[] newValueBytes = newValue == null ? null : ByteUtils.extractByteArray(newValue);
+    if (viewPartitionSet != null) {
```
ZacAttack, line 85: So..... a pending problem we have in incremental push is that we don't have support for sending DELETE events from inc push.  We used to, but there were problems in adding the DELETE type to the union.

The way I had kind of figured we'd handle this is we could just send null values....  This work isn't planned, implemented, or even design reviewed.  But I'm trying to think through if this puts us in a corner or not.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/task/datawriter/AbstractPartitionWriter.java
```diff
@@ -414,80 +417,66 @@ private AbstractVeniceWriter<byte[], byte[], byte[]> createCompositeVeniceWriter
...
-                    return ByteUtils.extractByteArray(compressor.decompress(valueBytes, 0, valueBytes.length));
-                  } catch (IOException e) {
-                    throw new VeniceException("Unable to decompress value bytes", e);
-                  }
-                });
-            continue;
+            deserializeFunction = (valueBytes, valueSchemaId) -> valueDeserializerCache
+                .computeIfAbsent(valueSchemaId, this::getValueDeserializer)
+                .deserialize(valueBytes);
+            compressor.get();
```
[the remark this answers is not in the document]
xunyin8: The idea was we have everything we need to initialize the compressor and it's going to be used eventually so we can try to force out any initialization errors as early as possible.

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/task/datawriter/CompositeVeniceWriter.java
```diff
@@ -108,29 +128,48 @@ public void close() throws IOException {
...
+   * Helper function to perform a composite put where put is first executed for all childWriters and then put is
+   * executed for mainWriter. The returned completable future is completed when the mainWriter completes the put.
    */
-  private CompletableFuture<PubSubProduceResult> compositeOperation(
-      BiFunction<AbstractVeniceWriter<K, V, U>, PubSubProducerCallback, CompletableFuture<PubSubProduceResult>> writerOperation,
+  private CompletableFuture<PubSubProduceResult> compositePut(
+      K key,
+      V value,
+      int valueSchemaId,
       PubSubProducerCallback childWriterCallback,
```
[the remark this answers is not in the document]
xunyin8: Good catch, some refactor misses.

internal/venice-common/src/main/java/com/linkedin/venice/writer/ComplexVeniceWriter.java
```diff
@@ -69,13 +82,45 @@ public CompletableFuture<Void> complexPut(K key, V value, int valueSchemaId, Laz
...
   }
 
+  /**
+   * Used during NR pass-through in remote region to forward records or chunks of records to corresponding view
+   * partition based on provided view partition map. This way the producing leader don't need to worry about large
+   * record assembly or chunking for view topic(s) when ingesting from source VT during NR pass-through.
+   */
+  public CompletableFuture<Void> forwardPut(K key, V value, int valueSchemaId, Set<Integer> partitions) {
+    if (partitions.isEmpty()) {
+      return CompletableFuture.completedFuture(null);
```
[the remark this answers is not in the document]
xunyin8: No, this is expected behavior right? I mentioned in comment that we differentiate unable to retrieve view partition map and expected view partitioner returning empty partitions by explicitly passing empty partition set.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/view/MaterializedViewWriter.java
xunyin8: I thought we moved incremental push to writing to RT exclusively a long time already. DELETE via incremental push will not be in this branch and should be supported by L/F or A/A SIT's RT consumption code path.

internal/venice-common/src/main/java/com/linkedin/venice/writer/ComplexVeniceWriter.java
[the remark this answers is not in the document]
gaojieliu: This function will only be used by NR passthrough mode in remote region, so do we expect this map can be empty for some cases?
I don't quite get your reply, so can you share more details?

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/task/datawriter/AbstractPartitionWriter.java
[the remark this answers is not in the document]
xunyin8: It won't cause an error but rather if something goes wrong in the `compressor.get()` we won't have to wait until we actually decompress a record to find out. Another benefit is technically `props` could be changed later so getting it now is safer. 
```
    compressor = Lazy.of(() -> {
      if (props.containsKey(KAFKA_INPUT_TOPIC)) {
        // Configure compressor using kafka input configs
        String sourceVersion = props.getString(KAFKA_INPUT_TOPIC);
        String kafkaInputBrokerUrl = props.getString(KAFKA_INPUT_BROKER_URL);
        CompressionStrategy strategy =
            CompressionStrategy.valueOf(props.getString(KAFKA_INPUT_SOURCE_COMPRESSION_STRATEGY));
        return KafkaInputUtils
            .getCompressor(compressorFactory.get(), strategy, kafkaInputBrokerUrl, sourceVersion, props);
      } else {
        CompressionStrategy strategy = CompressionStrategy.valueOf(props.getString(COMPRESSION_STRATEGY));
        if (strategy == CompressionStrategy.ZSTD_WITH_DICT) {
          String topicName = props.getString(TOPIC_PROP);
          ByteBuffer dict = DictionaryUtils.readDictionaryFromKafka(topicName, props);
          return compressorFactory.get()
              .createVersionSpecificCompressorIfNotExist(strategy, topicName, ByteUtils.extractByteArray(dict));
        } else {
          return compressorFactory.get().getCompressor(strategy);
        }
      }
    });
```

internal/venice-common/src/main/java/com/linkedin/venice/writer/ComplexVeniceWriter.java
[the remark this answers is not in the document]
xunyin8: Sure, it's possible for an materialized view user to provide a `ComplexVenicePartitioner` implementation that would return an empty array (`new int[0]`) for some values for many reasons. One could be the partitioning field is yet to be set (i.e. current value is the default value).  Another could be we'd like to filter records with certain value out, see `TestValueBasedVenicePartitioner` for example. In such cases the expected behavior is to not write this record to materialized view so we pass along an empty view partition set here to explicitly state that.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -3777,35 +3779,33 @@ private int processKafkaDataMessage(
           recordAssembledRecordSize(keyLen, put.getPutValue(), put.getReplicationMetadataPayload(), currentTimeMs);
         }
 
-        // Check if put.getSchemaId is positive, if not default to 1
-        // TODO: Write a test for chunked records... it does not seem right to transform negative schemas IDs into 1
-        int putSchemaId = put.getSchemaId() > 0 ? put.getSchemaId() : 1;
+        int writerSchemaId = put.getSchemaId();
```
kvargha, line 3782: How does this avoid the issue in the comments? When I was working on this, the first record received wouldn't have a valid schemaId.

xunyin8: Good question, so if the records are chunks then we should short circuit and return 0 for the whole `processKafkaDataMessage` function. We will only proceed if the record is a chunk manifest. In which case the writerSchemaId will be set to -20 first. However, later in the code when we do assemble the chunk we will be able to retrieve the correct user schema from the manifest and set it.
```
          ByteBufferValueRecord<ByteBuffer> assembledRecord = chunkAssembler.bufferAndAssembleRecord(
              consumerRecord.getTopicPartition(),
              put.getSchemaId(),
              keyBytes,
              put.getPutValue(),
              consumerRecord.getOffset(),
              compressor.get());

          // Current record is a chunk. We only write to the storage engine for fully assembled records
          if (assembledRecord == null) {
            return 0;
          }

          ByteBuffer assembledObject = assembledRecord.value();
          writerSchemaId = assembledRecord.writerSchemaId();
```

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/ChangelogClientConfig.java
```diff
@@ -33,9 +33,11 @@ public class ChangelogClientConfig<T extends SpecificRecord> {
   private long databaseSyncBytesInterval = 32 * 1024 * 1024L;
 
   /**
-   * RocksDB block cache size per BootstrappingVeniceChangelogConsumer. Default is 1 MB.
+   * RocksDB block cache size per BootstrappingVeniceChangelogConsumer. Default is 1 MB. This config is used for both
+   * the internal bootstrapping change log consumer and chunk assembler's RocksDB usage.
    */
   private long rocksDBBlockCacheSizeInBytes = 1024 * 1024L;
+  private String rocksDBChunkAssemblerStoragePath;
```
ZacAttack, line 40: we have a bootstrapFileSystemPath which is intended for shoving rocksdb stuff.  We could use that.

internal/venice-common/src/main/java/com/linkedin/venice/writer/ComplexVeniceWriter.java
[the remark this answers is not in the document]
gaojieliu: SG, can you add these details into the javadoc?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -482,7 +484,7 @@ public StoreIngestionTask(
     this.ingestionNotificationDispatcher =
         new IngestionNotificationDispatcher(notifiers, kafkaVersionTopic, isCurrentVersion);
     this.missingSOPCheckExecutor.execute(() -> waitForStateVersion(kafkaVersionTopic));
-    this.chunkAssembler = new ChunkAssembler(storeName);
+    this.chunkAssembler = new InMemoryChunkAssembler(new InMemoryStorageEngine(storeName));
```
gaojieliu, line 487: This instance is mainly for DVRT without an local state, right?
I think we are considering keeping a local state in some cases, so I wondered whether we need to construct an `InMemoryChunkAssembler` all the time or not.
Or shall the `ChunkAssembler` be smart enough to decide whether to use an InMemoryStorageEngine or on-disk RocksDB internally or not.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -3777,35 +3779,33 @@ private int processKafkaDataMessage(
...
 
           // Current record is a chunk. We only write to the storage engine for fully assembled records
-          if (assembledObject == null) {
+          if (assembledRecord == null) {
             return 0;
           }
 
+          ByteBuffer assembledObject = assembledRecord.value();
+          writerSchemaId = assembledRecord.writerSchemaId();
+          final int deserSchemaId = writerSchemaId;
```
gaojieliu, line 3801: Nit,
We often call it `readerSchemaId` instead of `deserSchemaId`.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/store/view/MaterializedViewWriter.java
```diff
@@ -67,34 +67,40 @@ public CompletableFuture<Void> processRecord(
...
+        throw new VeniceException(
+            "Encountered a null PUT record while having view partition map in the message header");
+      }
+      // Forward the record to corresponding view partition without any processing (NR pass-through mode).
+      return veniceWriter.get().forwardPut(key, newValueBytes, newValueSchemaId, viewPartitionSet);
+    }
     if (newValue == null) {
       // this is a delete operation
-      return veniceWriter.get().complexDelete(viewTopicKey, newValueProvider);
+      return veniceWriter.get().complexDelete(key, newValueProvider);
```
gaojieliu, line 96: Nit,
maybe we should add more details for `delete` since `newValueProvider` is actually providing the previous value.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumerImpl.java
```diff
@@ -152,9 +165,39 @@ public VeniceChangelogConsumerImpl(
...
     } else {
       this.storeName =
           VeniceView.getViewStoreName(changelogClientConfig.getStoreName(), changelogClientConfig.getViewName());
+      Properties rocksDBBufferProperties = new Properties();
+      String rocksDBBufferPath = changelogClientConfig.getRocksDBChunkAssemblerStoragePath();
+      if (rocksDBBufferPath == null || rocksDBBufferPath.isEmpty()) {
+        throw new VeniceException(
+            "rocksDBChunkAssemblerStoragePath must be configured for consuming view store: " + storeName);
+      }
+      rocksDBBufferProperties.put(DATA_BASE_PATH, rocksDBBufferPath);
```
gaojieliu, line 183: Shall we create a new folder in the path specified by the application?
I worried about that users might specify an non-empty folder, which contains other important data and the CDC consumer can drop all the files in that folder.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/ChangelogClientConfig.java
```diff
@@ -221,7 +223,8 @@ public static <V extends SpecificRecord> ChangelogClientConfig<V> cloneConfig(Ch
         .setConsumerName(config.consumerName)
         .setDatabaseSyncBytesInterval(config.getDatabaseSyncBytesInterval())
         .setShouldCompactMessages(config.shouldCompactMessages())
-        .setIsBeforeImageView(config.isBeforeImageView());
+        .setIsBeforeImageView(config.isBeforeImageView())
```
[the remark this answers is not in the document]
xunyin8: I think so? If we are certain about not supporting before image then we could do a separate PR to remove all before image related code. This way we will avoid removing configs for certain feature but leak related code elsewhere.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumerImpl.java
```diff
@@ -152,9 +165,39 @@ public VeniceChangelogConsumerImpl(
     if (changelogClientConfig.getViewName() == null || changelogClientConfig.getViewName().isEmpty()
         || changelogClientConfig.isBeforeImageView()) {
       this.storeName = changelogClientConfig.getStoreName();
+      rocksDBStorageEngineFactory = null;
+      // The in memory storage engine only relies on the name of store and nothing else. We use an unversioned store
+      // name
+      // here in order to reduce confusion (as this storage engine can be used across version topics).
+      InMemoryStorageEngine inMemoryStorageEngine = new InMemoryStorageEngine(storeName);
```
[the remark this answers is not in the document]
xunyin8: I did not implement the DVC record transformer originally but I'd imagine is for performance reasons? I just kept the original InMemoryStorageEngine support for DVC record transformer but can consult @kvargha and @ZacAttack to see if it can be replaced by disk based RocksDB.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
xunyin8: Maybe I'm missing some context about DVRT but are you worried about the cost where we instantiate a `InMemoryChunkAssembler` but never actually use it?

gaojieliu: Two parts:
1. Unnecessary instance creation, and overhead is minimal, but mainly for best practice of coding and if we always create many unnecessary objects, eventually, they can accumulate to incur a bigger cost.
2. Whether InMemoryStorageEngine should be used when DVRT is also trying to keep a local RocksDB snapshot, and I assume when DVRT does keep a local snapshot, we don't have to use InMemoryStorageEngine since all chunks are already in local disk.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumerImpl.java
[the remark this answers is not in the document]
ZacAttack: Original intention was performance, but that's not really backed by anything concrete.  Truthfully, we encounter chunking so rarely, I think it'd actually be permissable to just ditch the in memory one.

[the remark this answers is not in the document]
xunyin8: The maintainability of the two ChunkAssembler implementation shouldn't be an issue right? Most of the implementation is the same and the only branching is the cleanup logic. The in-memory chunk assembler actually has a better/bullet proof cleanup strategy than the rocksDB backed. Before we figure out a better strategy the rocksDB chunk assembler WILL leak broken chunks due to producer failure or starting to consume in between chunks.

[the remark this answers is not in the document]
xunyin8: Discussed offline and decided to defer the decision.

ZacAttack (dismissed):
Looks good!  Thanks!

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/ChangelogClientConfig.java
```diff
@@ -33,7 +33,8 @@ public class ChangelogClientConfig<T extends SpecificRecord> {
   private long databaseSyncBytesInterval = 32 * 1024 * 1024L;
 
   /**
-   * RocksDB block cache size per BootstrappingVeniceChangelogConsumer. Default is 1 MB.
+   * RocksDB block cache size per BootstrappingVeniceChangelogConsumer. Default is 1 MB. This config is used for both
+   * the internal bootstrapping change log consumer and chunk assembler's RocksDB usage.
    */
   private long rocksDBBlockCacheSizeInBytes = 1024 * 1024L;
```
gaojieliu, line 39: Nit,
1MB seems to be too small..
Can we use something like 64MB?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumerImpl.java
```diff
@@ -153,9 +168,40 @@ public VeniceChangelogConsumerImpl(
...
     } else {
       this.storeName =
           VeniceView.getViewStoreName(changelogClientConfig.getStoreName(), changelogClientConfig.getViewName());
+      Properties rocksDBBufferProperties = new Properties();
+      String rocksDBBufferPath = changelogClientConfig.getBootstrapFileSystemPath();
+      if (rocksDBBufferPath == null || rocksDBBufferPath.isEmpty()) {
+        throw new VeniceException("bootstrapFileSystemPath must be configured for consuming view store: " + storeName);
+      }
+      // Create a new folder in user provided path so if the path contains other important files we don't drop it.
+      rocksDBBufferProperties
```
gaojieliu, line 186: Question:
Do we expect this folder carry the previous state after a rolling bounce?
If yes, we need to flush the storage engine at checkpointing time, and if no, we can potentially clean up all the previous files at startup time, right?

gaojieliu (commented):
Left one question regarding the RocksDB buffer state during a restart event.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumerImpl.java
xunyin8: Good question, the change log consumer's checkpointing is interesting. The `seekToCheckpoint` API takes in user provided `Set<VeniceChangeCoordinate> checkpoints`. IIUC this means we can't guarantee not dropping records unless we flush after every poll. Otherwise even if the user seek to the exact spot after restart we could be missing chunks during assembly and we simply catch all errors and log:
```
LOGGER.warn(
            "Encountered error assembling chunked record, this can happen when seeking between chunked records. Skipping offset {} on topic {}",
            recordOffset,
            pubSubTopicPartition.getPubSubTopic().getName());
```

gaojieliu (approved):
Thanks!

PR #1569, opened 2025-03-04
[controller] log compaction service follow new optional service creation pattern

WhitneyDeng:
- private create() method
- LogCompactionService wrapped in Optional
- start() & close() service with Optional.ifPresent() rather than if/else

## How was this PR tested?
`TestHybrid::testHybridStoreLogCompaction()` passed

- [ x ] No. You can skip the rest of this section.


huangminchn (approved):
Thanks a lot Whitney!

PR #1570, opened 2025-03-04
[server][vpj][controller] Fix ClassNotFoundException in ApacheKafkaProducer when serializers are passed as strings

sushantmane:
##  Fix ClassNotFoundException in ApacheKafkaProducer when serializers are passed as strings  

When serializer classes were specified as strings in the producer configuration, Kafka attempted  
to load them using `Thread.currentThread().getContextClassLoader()`, which led to  
ClassNotFoundException issues, particularly in JDK 11 environments. To address this, the producer  
adapter now explicitly loads serializer classes, ensuring they are correctly resolved regardless of  
the class loader context. This change improves reliability and prevents unexpected failures due to  
class loading inconsistencies.   

## How was this PR tested?
CI


mynameborat (approved):
Do you need to update any tests?

majisourav99 (commented):
Can we add any tests for this?

sushantmane:
Sorry @bharathkk @majisourav99 - I had enabled auto merging for this PR. I missed your comments. 

> Do you need to update any tests?
All our E2E tests will run with these configs. If they pass that means change is at least correct.

PR #1571, opened 2025-03-04
[server] Do not record heartbeat lag for sep RT region if it is not enabled in version config

sixpluszero:
## [server] Do not record heartbeat lag for sep RT region if it is not enabled in version config
Do not record heartbeat lag for sep RT region if it is not enabled in version config
## How was this PR tested?

changed unit test


ZacAttack (approved):
Nice job!  Thanks!

PR #1572, opened 2025-03-04
[test] Add test to ensure Kafka serializer and deserializer configs remain consistent

sushantmane:
## Add test to ensure Kafka serializer and deserializer configs remain consistent

This test verifies that the key and value serializer and deserializer configurations for the Kafka 
clients are correctly set as `Class` objects rather than string class names. It helps catch  
unintended modifications that could lead to serialization issues in specific environments,   
even if tests pass in a controlled CI environment. A failure indicates a misconfiguration that   
could cause incompatibilities due to class loading behavior differences across JVMs.  

## How was this PR tested?
UT


sixpluszero (approved):

sushantmane:
Thanks, @sixpluszero!

PR #1573, opened 2025-03-05
[server] Improve participant store task handling and fix log pollution issues  

sushantmane:
## Improve participant store task handling, reduce log pollution, and update test logging  

Venice servers were creating the participant store consumption task even when it was not  
enabled or created by the Venice controller. This led to unnecessary queries to the store,  
polluting logs with redundant requests.  

This update ensures the participant store consumption task is only created when  
PARTICIPANT_MESSAGE_STORE_ENABLED is enabled, preventing excessive router queries  
and reducing log noise.  

It also includes test improvements, resource leak fixes in TestServerStorePropertiesEndpoint,  
and enhancements to Kafka consumer/producer configurations for better uniqueness in client IDs.  
This prevents log pollution caused by InstanceAlreadyExistsException when the same ID is used  
for multiple producers.  

Additionally, the unused StoreMetadataType class has been removed as part of the cleanup.  

For testing, excessive log output has been reduced by setting noisy loggers (Zookeeper, Helix,  
Kafka, D2, and Eclipse) to ERROR or OFF in the E2E test log configuration file.  

## How was this PR tested?
CI


internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/integration/utils/VeniceMultiClusterWrapper.java
```diff
@@ -129,6 +130,11 @@ static ServiceProvider<VeniceMultiClusterWrapper> generateService(VeniceMultiClu
       extraProperties.put(SYSTEM_SCHEMA_CLUSTER_NAME, clusterNames[0]);
       extraProperties.putAll(KafkaTestUtils.getLocalCommonKafkaSSLConfig(SslUtils.getTlsConfiguration()));
       pubBrokerDetails.forEach((key, value) -> extraProperties.putIfAbsent(key, value));
+      if (controllerProperties.containsKey(PARTICIPANT_MESSAGE_STORE_ENABLED)) {
+        extraProperties
```
sixpluszero, line 134: I think we should make this by default enabled for all the test. This feature is always turned on and our test should keep exercising the code path.

sushantmane: The problem is if we turn it on lot of tests start failing. Unless we make them pass always we can turn it on

sixpluszero: I see..this is a bit concerning. But maybe we can improve that incrementally.

sixpluszero (approved):

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/integration/utils/VeniceMultiClusterWrapper.java
sushantmane: I previously attempted to enable it in this [PR](https://github.com/linkedin/venice/pull/1338) but had to disable it again due to failing tests. However, I recall keeping it enabled for most tests while disabling it for specific ones. I’ll re-enable it in a follow-up PR, and if all tests pass, we can enable it by default.

PR #1574, opened 2025-03-05
[controller] New Semantic Usage Detector

minhmo1620:
## [controller] New Semantic Usage Detector
The class to help detect whether the object is using non-allowed fields in the target schema.

When object value is different from the default value of that field, we know someone is trying to use new semantic. When the system is not ready for this yet, we will fail the request (and not include the message to the queue at all).

Within this PR, we mainly focus on crafting the class. The actual usage inside serializer will be coming in the next PR with integration tests.

Semantic check - Overall
Semantic check - Page 1

## Project description TLDR version
TLDR version:
Initial start: Producers (parent controllers) need to generate AdminConsumptionTask to put into the admin topic queue. Messages will be consumed by consumers (parent controllers + child controllers). If producers produce the message that consumers cannot read, we are having troubles.

Current solution: Child controllers when upgraded to the newer version, will need to be completed BEFORE parent controllers. (Since parent controllers is currently using their latest version code, and child controller don't have the schema for latest version yet).

New solution:
We put a version number in parents' ZK metadata, specifying the version that producers will use (let's called it X)

Producer: OK, I will always use version X to produce message.
In deployment, parent and child controllers can freely upgrade to X+1, without any delay or order.

(**this PR**) During the serialization, producers may figure out that the message with newer schema is using non-default value for the new field (the field that child controller don't know yet). 

(**next PR**) Producers fail the task:
- Sorry, we are not ready for this message 👎 

Then, when they are all upgraded, we send a request to ZK:

OK, let's use X+1 from now on
=> Complete the deployment.

## How was this PR tested?
UT


services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/protocol/serializer/SchemaDiffTraverser.java
```diff
@@ -0,0 +1,235 @@
...
+        || type == Schema.Type.MAP;
+  }
+
+  /**
+   * Determines if a value is different from its default.
+   */
+  public boolean isNonDefaultValue(Object value, Schema.Field field) {
+    System.out.println("Checking Field: " + field.name() + ", Value: " + value);
+    Map<Schema.Type, Object> defaultValues = new HashMap<>();
+    defaultValues.put(Schema.Type.STRING, "");
```
huangminchn, line 143: I think the default for string in Java is null; can null be put into this Map<Schema.Type, Object>?

minhmo1620: Should be, I can pass in null in the position of Object in the traverse method. Do you want to put null in this default list?

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/protocol/serializer/SchemaDiffTraverser.java
```diff
@@ -0,0 +1,246 @@
...
+  private boolean isNonDefaultValue(Object value, Schema.Field field, AtomicReference<String> errorMessage) {
+    Map<Schema.Type, Object> defaultValues = new HashMap<>();
+    defaultValues.put(Schema.Type.STRING, "");
+    defaultValues.put(Schema.Type.BOOLEAN, false);
+    defaultValues.put(Schema.Type.INT, 0);
+    defaultValues.put(Schema.Type.LONG, 0L);
+    defaultValues.put(Schema.Type.FLOAT, 0.0f);
+    defaultValues.put(Schema.Type.DOUBLE, 0.0d);
+    defaultValues.put(Schema.Type.BYTES, new byte[0]);
+    if (value != null && !value.equals(defaultValues.getOrDefault(field.schema().getType(), null))
```
majisourav99, line 234: can we move the null-check on the value before creating the map?

services/venice-controller/src/test/java/com/linkedin/venice/controller/kafka/protocol/admin/SchemaDiffTraverserTest.java
```diff
@@ -0,0 +1,164 @@
...
+  public void testNestedArrayTraverse() {
+    String schemaJson = "{" + "\"type\": \"array\"," + "\"items\": {" + "  \"type\": \"record\","
+        + "  \"name\": \"ExampleRecord\"," + "  \"fields\": [" + "    {\"name\": \"field1\", \"type\": \"string\"},"
+        + "    {\"name\": \"field2\", \"type\": \"int\"}" + "  ]" + "}" + "}";
+
+    String targetSchemaJson =
+        "{" + "\"type\": \"array\"," + "\"items\": {" + "  \"type\": \"record\"," + "  \"name\": \"ExampleRecord\","
+            + "  \"fields\": [" + "    {\"name\": \"field1\", \"type\": \"string\"}" + "  ]" + "}" + "}";
+
+    Schema.Parser parser1 = new Schema.Parser();
```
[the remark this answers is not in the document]
minhmo1620: Thank you, I ended with `AvroCompatibilityHelper.parse`

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/protocol/serializer/SchemaDiffTraverser.java
minhmo1620: Updated, thank you for your suggestion.

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/protocol/serializer/SchemaDiffTraverser.java
```diff
@@ -0,0 +1,421 @@
...
+    }
+
+    return false;
+  }
+
+  private boolean isNonDefaultValue(Object value, Schema.Field field, AtomicReference<String> errorMessage) {
+    if (value == null)
+      return false;
+
+    Map<Schema.Type, Object> defaultValues = new HashMap<>();
```
majisourav99, line 401: actually instead of allocating a new map just to check for defaults, can you use switch statement?

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/protocol/serializer/NewSemanticUsageValidator.java
```diff
@@ -0,0 +1,253 @@
...
+      case UNION:
+        return isNonDefaultValueUnion(object, currentField);
+      case ENUM:
+        return isNewEnumValue(object, currentField, targetField);
+      case FIXED:
+        if (currentField.schema().getFixedSize() != targetField.schema().getFixedSize()) {
+          return returnTrueAndLogError(
+              String.format("Field %s: has different fixed size", formatFieldName(currentField.name())));
+        }
+        return false;
```
huangminchn, line 72: Fixed is a tricky one; I believe with the current implementation, if the new field is a FIXED schema, we are enforcing the usage of new semantics.
Currently we are using the latest schema to serialize, and then do another round of deserialization and serialization.
If the codes put null to the new FIXED field, the first serialization will fail with NPE, so developers must put a fixed size array to this field.

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/protocol/serializer/NewSemanticUsageValidator.java
```diff
@@ -0,0 +1,253 @@
...
+      }
+    }
+    return null;
+  }
+
+  /**
+   * Check if the value is non-default for the given field.
+   * If the value is null, it is considered default.
+   * Default value of each type are only applicable for non-nested fields.
+   * If the field is nested-field, the default value is always null.
```
huangminchn, line 140: What if nested field is also a record schema?

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/protocol/serializer/NewSemanticUsageValidator.java
```diff
@@ -0,0 +1,253 @@
...
+    if (object == null)
+      return false;
+
+    Object fieldDefaultValue = AvroSchemaUtils.getFieldDefault(field);
+    Object typeDefaultValue = getDefaultForType(field.schema().getType());
+
+    if (!Objects.equals(object, typeDefaultValue) && !Objects.equals(object, fieldDefaultValue)) {
+      return returnTrueAndLogError(
+          String.format(
+              "Field: %s contains non-default value. Actual value: %s. Default value: %s or %s",
```
huangminchn, line 156: Just curious, if default is null, would there be NPE when printing this string?

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/protocol/serializer/NewSemanticUsageValidator.java
```diff
@@ -0,0 +1,253 @@
...
+   * Check if the value is a new enum value.
+   * @param object the value to check
+   * @param currentField the field to check with enum type and its current symbols
+   * @param targetField the field to check with enum type and its target symbols
+   * @return true if the value is a new enum value, false otherwise
+   */
+  public boolean isNewEnumValue(Object object, Schema.Field currentField, Schema.Field targetField) {
+    List<String> currentSymbols = currentField.schema().getEnumSymbols();
+    List<String> targetSymbols = targetField.schema().getEnumSymbols();
+    if (targetSymbols.stream().anyMatch(symbol -> !currentSymbols.contains(symbol) && symbol.equals(object))) {
```
huangminchn, line 88: `symbol` is always String; what does "symbol.equals(object)" indicate here?

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/protocol/serializer/NewSemanticUsageValidator.java
```diff
@@ -0,0 +1,253 @@
...
+    Schema subSchema = findObjectSchemaInsideUnion(object, currentField);
+    if (subSchema == null) {
+      return returnTrueAndLogError(
+          String.format(
+              "Field %s: Cannot find the match schema for value %s from schema union",
+              formatFieldName(currentField.name()),
+              object));
+    }
+
+    Schema.Field newCurrentField = AvroCompatibilityHelper.createSchemaField(currentField.name(), subSchema, "", null);
```
huangminchn, line 117: Here we are using "null" as default value; if subSchema is int or double, "null" would be invalid default value.

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/protocol/serializer/NewSemanticUsageValidator.java
```diff
@@ -0,0 +1,253 @@
...
+
+  /**
+   * Get the default value for the given type.
+   * @param type the type
+   * @return the default value if it is primitive type, else null for nested fields.
+   */
+  private static Object getDefaultForType(Schema.Type type) {
+    switch (type) {
+      case STRING:
+        return "";
```
huangminchn, line 190: I believe default for String in java would be null.

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/protocol/serializer/NewSemanticUsageValidator.java
minhmo1620: `If symbol in targetSymbols. symbol not in currentSymbols and symbol equal to object
`means

If symbol in new_enum, symbol not in prev_enum, and symbol equal to object --> object is the new enum value.

I understand it is confusing, let me fix this.

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/protocol/serializer/NewSemanticUsageValidator.java
```diff
@@ -0,0 +1,253 @@
...
+ * <ol>
+ *   <li>It is null.</li>
+ *   <li>It is a nested type and equals its schema's default value.</li>
+ *   <li>It is a non-nested field and matches either the schema's default value or its type's default.</li>
+ * </ol>
+ *
+ * If the object is not using new semantic, we return False. Otherwise, we return True along with the message
+ * in errorMessage. The traverser should stop traversing the object if the return value is True.
+ */
+public class NewSemanticUsageValidator {
```
FelixGV, line 26: Thanks for working on this functionality! IMHO, this is a fairly generic utility and should not be Venice-specific.

It is fine for it to start as such, but once it is stabilized and we feel ready, it may be more appropriate to contribute it to [avro-util](https://github.com/linkedin/avro-util), or maybe even Apache Avro itself.

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/protocol/serializer/SchemaDiffTraverser.java
```diff
@@ -0,0 +1,300 @@
+package com.linkedin.venice.controller.kafka.protocol.serializer;
+
+import com.linkedin.avroutil1.compatibility.AvroCompatibilityHelper;
+import com.linkedin.venice.utils.AvroSchemaUtils;
+import com.linkedin.venice.utils.Pair;
```
FelixGV, line 5: Note that `Pair` is deprecated and we prefer using dedicated classes instead. There is a similar class which already exists called `SchemaPairAndClassContainer`, though it contains one extra field that you likely don't need (though it supports being null, so technically it would still work...). You can also create a new `SchemaReaderWriterPair` class, or similar.

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/protocol/serializer/NewSemanticUsageValidator.java
minhmo1620: nope. this goes in normally. example http://github.com/linkedin/venice/pull/1574/files#diff-ab9e2fbeeb18564c19f00c706cf6f4659aec783e10fd5447d00ec3718601dec1R289

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/protocol/serializer/NewSemanticUsageValidator.java
minhmo1620: If the object is null, it should not go into this far. So putting the default value for string as null doesn't matter much I think.

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/protocol/serializer/SemanticDetector.java
```diff
@@ -0,0 +1,438 @@
...
+      throw new VeniceProtocolException(
+          String.format(
+              "Field %s: Type %s is not the same as %s",
+              formatFieldName(name),
+              currentSchema.getType(),
+              targetSchema.getType()));
+    }
+
+    // If the current schema and target schema are the same, we do not need to traverse the object.
+    // @code{compareSchemaIgnoreFieldOrder} method is used to compare the two schemas (but cannot take FIXED type)
```
huangminchn, line 60: Please add a more context here to explain why we need to do more checks for FIXED type even if both current and target schema has FIXED type.

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/protocol/serializer/SemanticDetector.java
```diff
@@ -0,0 +1,438 @@
...
+    return parent.isEmpty() ? field : parent + "_" + field;
+  }
+
+  /**
+   * Format the field name to replace "_" with ".". Reason is that we use "_" as a delimiter in the field name since "." is
+   * invalid character in field name.
+   */
+  private static String formatFieldName(String fieldName) {
+    return fieldName.replace("_", ".");
+  }
```
huangminchn, line 437: Just curious, why not use "." at the beginning?

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/protocol/serializer/SemanticDetector.java
```diff
@@ -0,0 +1,438 @@
...
+    if (currentSchema.getType() != Schema.Type.FIXED
+        && AvroSchemaUtils.compareSchemaIgnoreFieldOrder(currentSchema, targetSchema))
+      return;
+
+    switch (currentSchema.getType()) {
+      case RECORD:
+        traverseFields((GenericRecord) object, currentSchema, targetSchema, name, defaultValue);
+        break;
+      case ARRAY:
+        traverseCollections((List<Object>) object, currentSchema, targetSchema, name, defaultValue);
```
huangminchn, line 70: Does this casting to List always work, or does it depend on how the avro classes are compiled? Could you please double check the avro doc and confirm that array field would always be compiled into Java List?

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/protocol/serializer/SemanticDetector.java
```diff
@@ -0,0 +1,438 @@
...
+      String name,
+      Object defaultArrayEntry) {
+    String arrayName = buildFieldPath(name, currentSchema.getName());
+    String nestedSchemaName = buildFieldPath(arrayName, currentSchema.getElementType().getName());
+
+    Schema currentElementSchema = currentSchema.getElementType();
+    Schema targetElementSchema = targetSchema.getElementType();
+
+    for (int i = 0; i < array.size(); i++) {
+      String indexArrayName = buildFieldPath(nestedSchemaName, String.valueOf(i));
```
huangminchn, line 164: Nice! It will help with debugging for sure :D

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/protocol/serializer/SemanticDetector.java
```diff
@@ -0,0 +1,438 @@
...
+        return value instanceof byte[];
+      case RECORD:
+        return value instanceof GenericRecord && ((GenericRecord) value).getSchema().equals(schema);
+      case ARRAY:
+        return value instanceof List
+            && ((List<?>) value).stream().allMatch(element -> isCorrectSchema(element, schema.getElementType()));
+      case MAP:
+        return value instanceof Map && ((Map<?, ?>) value).values()
+            .stream()
+            .allMatch(element -> isCorrectSchema(element, schema.getValueType()));
```
huangminchn, line 299: We probably don't need to loop through the entire collection in this check. `anyMatch` should be good enough. WDYT?

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/protocol/serializer/SemanticDetector.java
```diff
@@ -0,0 +1,438 @@
...
+      throw new VeniceProtocolException(
+          String.format(
+              "Field %s: Object %s does not match any schema within union schemas %s",
+              formatFieldName(name),
+              object,
+              schemasInUnion));
+    }
+
+    Map<String, Schema> targetSchemaMap =
+        targetSchema.getTypes().stream().collect(Collectors.toMap(Schema::getName, s -> s));
```
huangminchn, line 249: Just for my education, primitive avro types also have names?

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/protocol/serializer/SemanticDetector.java
```diff
@@ -0,0 +1,438 @@
...
+    List<String> prevEnumSymbols = currentSchema.getEnumSymbols();
+    List<String> targetEnumSymbols = targetSchema.getEnumSymbols();
+    // If the value is not in the previous enum symbols but in the target enum symbols, it is a new enum value
+    if (!prevEnumSymbols.contains(enumValue) && targetEnumSymbols.contains(enumValue)) {
+      throw new VeniceProtocolException(
+          String.format(
+              "Field %s: Enum value %s is not in the previous enum symbols but in the target enum symbols",
+              formatFieldName(name),
+              enumValue));
+    }
```
huangminchn, line 424: It should be the opposite? It's bad if current schema has this entry but target schema doesn't have it?
If so, do we have a test for ENUM type?

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/protocol/serializer/SemanticDetector.java
```diff
@@ -0,0 +1,438 @@
...
+          throw new VeniceProtocolException(
+              String.format(
+                  "Field %s: Long value %s is not the default value 0 or %s",
+                  formatFieldName(name),
+                  object,
+                  defaultValue));
+        }
+        break;
+      case FLOAT:
+        if (!((object instanceof Float) && (float) object == 0.0f)) {
```
huangminchn, line 350: BTW, for double and float, I would recommend checking the difference between object and 0.0f or 0.0d, and confirm that the difference is smaller than a very small number like 0.00001.
It's not accurate to check object equals to 0.0

huangminchn (commented):
Thanks for the hard work, Minh! This version of implementation is much more readable and easy to reason.

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/protocol/serializer/SemanticDetector.java
minhmo1620: Example for generated code for array field:
```json
{
              "name": "targetedRegions",
              "doc": "The list of regions that is separated by comma for targeted region push. If set, this admin message should only be consumed by the targeted regions",
              "type": [
                "null",
                {
                  "type": "array",
                  "items":  "string"
                }
              ],
              "default": null
            },
```
```java
  /**
   * Sets the value of the 'targetedRegions' field.
   * The list of regions that is separated by comma for targeted region push. If set, this admin message should only be consumed by the targeted regions
   * @param value the value to set.
   */
  public void setTargetedRegions(java.util.List<java.lang.CharSequence> value) {
    this.targetedRegions = value;
  }
```

minhmo1620: Similar for map
```json
                   {
                      "name": "partitionerParams",
                      "type": {
                        "type": "map",
                        "values": "string"
                      }
                    },
```

```java
  /**
   * Sets the value of the 'partitionerParams' field.
   * Partitioner parameters
   * @param value the value to set.
   */
  public void setPartitionerParams(java.util.Map<java.lang.CharSequence,java.lang.CharSequence> value) {
    this.partitionerParams = value;
  }
```

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/protocol/serializer/SemanticDetector.java
minhmo1620: primitive type does not have name, but the `Schema` has name.  For primitive types, the name is the same with type
```java
  public void testFieldName() {
    Schema unionSchema = Schema.createUnion(Arrays.asList(stringSchema, intSchema));
    Map<String, Schema> targetSchemaMap =
        unionSchema.getTypes().stream().collect(Collectors.toMap(Schema::getName, s -> s));

    System.out.println(targetSchemaMap); // {string="string", int="int"}
  }
```

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/protocol/serializer/SemanticDetector.java
minhmo1620: `anyMatch` works. Updating.

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/protocol/serializer/SemanticDetector.java
minhmo1620: good catch, you are correct. Updating the test, search `testEnumNewValue`

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/protocol/serializer/SemanticDetector.java
minhmo1620: Good catch, before this change, I need to feed this fieldName into Schema.Field, which is not accepting `.` so I need to switch to `_`. Now we don't need to make it into a field anymore, I can use dot directly.

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/protocol/serializer/SchemaDiffTraverser.java
lusong64: unless the map is concurrentMap, then the null will trigger an exception.

services/venice-controller/src/test/java/com/linkedin/venice/controller/kafka/protocol/admin/SemanticDetectorTest.java
```diff
@@ -0,0 +1,530 @@
...
+        currentSchema,
+        "fieldA",
+        targetSchema,
+        "Field fieldA: Invalid enum value value5 is not accepted in the current schema [value1, value2, value3, value4]");
+
+    testBadSemanticUsage(1, currentSchema, "fieldB", targetSchema, "Field fieldB: Enum value 1 is not a string");
+  }
+
+  @Test
+  public void testFindObjectSchemaInsideUnion() {
```
huangminchn, line 378: Could you please add a bad semantic usage test for union schema? For example, target schema only has one record schema in the union, while the current schema has two different record schemas in the union; we test that traverse would fail if we start using the second record in the union.

huangminchn (changes requested):

services/venice-controller/src/test/java/com/linkedin/venice/controller/kafka/protocol/admin/SemanticDetectorTest.java
```diff
@@ -0,0 +1,530 @@
...
+    array.add(record1);
+    array.add(record2);
+
+    // Traverse the array
+    testBadSemanticUsage(
+        array,
+        schema,
+        "",
+        targetSchema,
+        "Field array.0.ExampleRecord.field2: Integer value 123 is not the default value 0 or null");
```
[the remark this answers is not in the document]
minhmo1620: Added.

services/venice-controller/src/test/java/com/linkedin/venice/controller/kafka/protocol/admin/SemanticDetectorTest.java
minhmo1620: Added.

huangminchn (approved):
Thanks a lot Minh!

services/venice-controller/src/main/java/com/linkedin/venice/controller/kafka/protocol/serializer/NewSemanticUsageValidator.java
minhmo1620: Thank you. When the class is stable, I will consider to bring it to more general place.

PR #1575, opened 2025-03-05
[protocol] GlobalRtDivState: Schema Change (Offset)

KaiSernLim:
DIFFSIZEOVERRIDE because the test file needs to be amended.

## Summary
Added `long latestOffset` and `bytes latestPubSubPosition` (for future use) to the `GlobalRtDivState` schema for the latest consumed rt offset (LCRO). This is needed in times of leadership handover, so a follower / leader knows the offset to subscribe at.

This is an in-place modification because the schema was never previously used. It was guarded behind a feature flag that was never enabled, and there were some tests that used the `GlobalRtDivState`, but otherwise this shouldn't be a problem.


internal/venice-common/src/main/resources/avro/GlobalRtDivState/v2/GlobalRtDivState.avsc
```diff
@@ -0,0 +1,90 @@
...
+                "values": "string"
+              }
+            }
+          ]
+        }
+      },
+      "default": {}
+    },
+    {
+      "name": "latestOffset",
```
sushantmane, line 85: Let's also add `byte` offset field.

```
>             }, {
>               "name": "latesPubSubPosition",
>               "doc": "The last PubSubPosition consumed successfully in this partition."",
>               "type": "bytes",
>               "default": ""
>             }, {
```

Samples:
https://github.com/linkedin/venice/pull/1482
https://github.com/linkedin/venice/pull/1481

sushantmane: Going forward, we need to consistently follow this practice as numeric offset usage will be deprecated.

sushantmane:
1. Could you please diff in PR summary? `diff oldSchemaFile newSchemaFile`

2. We no longer allow bumping schema version and new schema additions in the same PR. So'll have to undo those changes

internal/venice-common/src/main/java/com/linkedin/venice/serialization/avro/AvroProtocolDefinition.java
```diff
@@ -186,7 +186,7 @@ public enum AvroProtocolDefinition {
   /**
    * Global Realtime Topic Data Integrity Validator is the RT DIV snapshot propagated from the leader to followers.
    */
-  GLOBAL_RT_DIV_STATE(1, GlobalRtDivState.class);
+  GLOBAL_RT_DIV_STATE(2, GlobalRtDivState.class);
```
sushantmane, line 189: You need to change this in a follow up PR where you'll actually use this schema

sixpluszero:
> 1. Could you please add diff in PR summary? `diff oldSchemaFile newSchemaFile`
> 2. We no longer allow bumping schema version and new schema additions in the same PR. So'll have to undo those changes
> 3. Some fields do not have default values.. that could make schema evolution (esp field deletion) challenging or impossible ?
> 4. Instead of bootstrap url we should use region index from the kafka/pubsub cluster map. This will also be consistent with PCS/OffsetRecord and other schemas

For 4, it should be URL -> Long mapping for PCS and OffsetRecord. I think we can keep the convension and since it is per-partition metadata, it won't save much space and it will be more easy to debug TBH

sushantmane (approved):
LGTM. Thanks, @KaiSernLim!

sushantmane (approved):

PR #1576, opened 2025-03-05
[server] Global RT DIV: Leader

KaiSernLim:
## Summary

DIV-Producer

When the `isGlobalRtDivEnabled()` feature flag is enabled, the drainer will no longer sync the VT DIV to the `OffsetRecord` in the StorageEngine. In the `ConsumptionTask` stage, leader will produce a (Global) RT DIV to VT to replicate to the followers. On the callback, it will enqueue a `DivSnapshot` (VT DIV + RT DIV) to the `Drainer`. `Drainer` will put the RT DIV (possibly chunked) in the storage engine, and sync the `OffsetRecord` with VT DIV. Note that the `GlobalRtDivState` will be compressed because it's being sent to VT.

The Drainer component will be handled in a future PR.

### Per-Broker Global RT DIV

Note that the Global RT DIV is on a per-broker / region basis. `PartitionTracker#rtSegments` is a map of `Broker URL -> Producer GUID -> Segment` whereas the VT version is only `Producer GUID -> Segment`. There should only be one `ConsumptionTask` per broker / region and only one per VT, so there should not be any contention with the `Map` with this setup. The latest consumed RT offset (LCRO) is not stored in-memory, it is simply the offset of the last consumed message that triggered the DIV to be synced.

### When to Sync?
The `syncBytesInterval` from the drainer syncing `OffsetRecord` is reused.

However, since the Global RT DIV is per-broker, the tally for how many bytes have been consumed since the last Global RT DIV sync also needs to be per-broker, and needs to be divided by the number of brokers in order to be proportional.

### Changes

* [functional] Added the send Global RT DIV code inside `LeaderFollowerStoreIngestionTask#produceToLocalKafka()`.
* [class] Created container class `DivSnapshot` which contains VT DIV + RT DIV + `latestConsumedVtOffset` + `latestConsumedRtOffset` for the consumer to enqueue to the drainer via `LeaderProducerCallback`.
* [functional] Created method to convert `Segment` into `ProducerPartitionState`. Below is a diagram describing the relationship between these two entities.
* [functional] Created method to clone the VT + RT DIV in `PartitionTracker`.
* [functional] Added `latestConsumedVtOffset` (LCVO) to `PartitionTracker` which is updated as VT is being consumed.

### VT + RT DIV Details
VT + RT DIV

### `Segment` and `ProducerPartitionState` Conversion
pps

### Global RT DIV Message Payload
DIV-Payload
Identical to the existing `PUT` requests other than a few changes:
* **In general:** the Key's `MessageType` is changed from `PUT` to `GLOBAL_RT_DIV`.
* If chunking is necessary, the `schemaId` of the `GlobalRtDivState` is in the `ChunkedValueManifest`. The usage of `ChunkedValueManifest` and `Chunk` `schemaId` remains.

## How was This PR Tested?

1. `testShouldSendGlobalRtDiv()` verifies that upon enabling the `GlobalRtDiv` feature flag, the DIV / `OffsetRecord` is synced by the consumer rather than the drainer.
2. `testSendGlobalRtDivMessage()` verifies that `LeaderFollowerStoreIngestionTask` correctly serializes `GlobalRtDivState` from `VeniceWriter#put()` and includes `DivSnapshot` in the callback for the drainer to use.
3. `testPutGlobalRtDiv()` verifies the output when `put()` is called with `GlobalRtDivState`.
4. `testUpdateLatestConsumedVtOffset() verifies that LCVO is being updated correctly.
5. `testToProducerPartitionState()` verifies the conversion between `Segment` and `ProducerPartitionState`.
TODO: trying to add an integration test.

## Does this PR Introduce Any User-Facing Changes?

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
@@ -2412,7 +2412,7 @@ public class LeaderFollowerStoreIngestionTask extends StoreIngestionTask {
 
         if (isGlobalRtDivEnabled() && !consumerRecord.getTopicPartition().getPubSubTopic().isRealTime()
             && msgType != MessageType.GLOBAL_RT_DIV) {
-          final long offset = consumerRecord.getOffset();
+          final long offset = consumerRecord.getPosition().getNumericOffset();
           getKafkaDataIntegrityValidatorForLeaders().updateLatestConsumedVtOffset(partition, offset);
         }
 
@@ -3581,7 +3581,7 @@ public class LeaderFollowerStoreIngestionTask extends StoreIngestionTask {
    * @param previousMessage the last message validated and produced to kafka before this GlobalRtDiv will be produced
    */
   void sendGlobalRtDivMessage(
-      PubSubMessage<KafkaKey, KafkaMessageEnvelope, Long> previousMessage,
+      DefaultPubSubMessage previousMessage,
       PartitionConsumptionState partitionConsumptionState,
       int partition,
       String brokerUrl,
@@ -3599,7 +3599,8 @@ public class LeaderFollowerStoreIngestionTask extends StoreIngestionTask {
 
     // Create GlobalRtDivState (RT DIV + latestOffset) which will be serialized into a byte array
     ByteBuffer emptyBuffer = ByteBuffer.allocate(0);
-    GlobalRtDivState globalRtDiv = new GlobalRtDivState(brokerUrl, rtDiv, previousMessage.getOffset(), emptyBuffer);
+    final long offset = previousMessage.getPosition().getNumericOffset();
+    GlobalRtDivState globalRtDiv = new GlobalRtDivState(brokerUrl, rtDiv, offset, emptyBuffer);
 
     // Create PubSubMessage for the LeaderProducerCallback to enqueue the RT + VT DIV to the drainer
     KafkaKey divKey = new KafkaKey(MessageType.GLOBAL_RT_DIV, keyBytes);
@@ -3612,12 +3613,12 @@ public class LeaderFollowerStoreIngestionTask extends StoreIngestionTask {
             true,
             DEFAULT_LEADER_METADATA_WRAPPER,
             APP_DEFAULT_LOGICAL_TS);
-    divEnvelope.payloadUnion = new DivSnapshot(divClone, previousMessage.getOffset()); // contains both VT + RT DIV
-    PubSubMessage<KafkaKey, KafkaMessageEnvelope, Long> divMessage = new ImmutablePubSubMessage<>(
+    divEnvelope.payloadUnion = new DivSnapshot(divClone, offset); // contains both VT + RT DIV
+    DefaultPubSubMessage divMessage = new ImmutablePubSubMessage(
         divKey,
         divEnvelope,
         previousMessage.getTopicPartition(),
-        previousMessage.getOffset(), // TODO: are these reused fields correct?
+        previousMessage.getPosition(), // TODO: are these reused fields correct?
         previousMessage.getPubSubMessageTime(),
         divKey.getHeapSize()); // TODO: should the envelope size also be estimated?
     LeaderProducerCallback divCallback = createProducerCallback(
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
@@ -48,12 +48,15 @@ import com.linkedin.venice.exceptions.validation.FatalDataValidationException;
 import com.linkedin.venice.guid.GuidUtils;
 import com.linkedin.venice.kafka.protocol.ControlMessage;
 import com.linkedin.venice.kafka.protocol.Delete;
+import com.linkedin.venice.kafka.protocol.GUID;
 import com.linkedin.venice.kafka.protocol.KafkaMessageEnvelope;
 import com.linkedin.venice.kafka.protocol.Put;
 import com.linkedin.venice.kafka.protocol.TopicSwitch;
 import com.linkedin.venice.kafka.protocol.Update;
 import com.linkedin.venice.kafka.protocol.enums.ControlMessageType;
 import com.linkedin.venice.kafka.protocol.enums.MessageType;
+import com.linkedin.venice.kafka.protocol.state.GlobalRtDivState;
+import com.linkedin.venice.kafka.protocol.state.ProducerPartitionState;
 import com.linkedin.venice.kafka.protocol.state.StoreVersionState;
 import com.linkedin.venice.message.KafkaKey;
 import com.linkedin.venice.meta.DataReplicationPolicy;
@@ -76,6 +79,8 @@ import com.linkedin.venice.pubsub.api.PubSubTopicPartition;
 import com.linkedin.venice.schema.SchemaEntry;
 import com.linkedin.venice.schema.writecompute.DerivedSchemaEntry;
 import com.linkedin.venice.serialization.AvroStoreDeserializerCache;
+import com.linkedin.venice.serialization.avro.AvroProtocolDefinition;
+import com.linkedin.venice.serialization.avro.InternalAvroSpecificSerializer;
 import com.linkedin.venice.serializer.RecordDeserializer;
 import com.linkedin.venice.stats.StatsErrorCode;
 import com.linkedin.venice.storage.protocol.ChunkedValueManifest;
@@ -190,6 +195,9 @@ public class LeaderFollowerStoreIngestionTask extends StoreIngestionTask {
    * for this leader validator as well, if deemed necessary.
    */
   private final KafkaDataIntegrityValidator kafkaDataIntegrityValidatorForLeaders;
+
+  private static final InternalAvroSpecificSerializer<GlobalRtDivState> globalRtDivStateSerializer =
+      AvroProtocolDefinition.GLOBAL_RT_DIV_STATE.getSerializer();
 
   /**
    * N.B.:
@@ -335,7 +343,7 @@ public class LeaderFollowerStoreIngestionTask extends StoreIngestionTask {
       this.veniceWriterForRealTime = this.veniceWriter;
     }
 
-    this.kafkaClusterIdToUrlMap = serverConfig.getKafkaClusterIdToUrlMap();
+    this.kafkaClusterIdToUrlMap = serverConfig.getKafkaClusterIdToUrlMap(); // ?
     this.kafkaDataIntegrityValidatorForLeaders = new KafkaDataIntegrityValidator(kafkaVersionTopic);
     if (builder.getVeniceViewWriterFactory() != null && !store.getViewConfigs().isEmpty()) {
       viewWriters = builder.getVeniceViewWriterFactory()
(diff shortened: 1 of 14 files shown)


sixpluszero (commented):
Thanks for the change, I did a pass but there are some confusion I also need to check about it and I left a few comments.

Regarding "the drainer will no longer sync the VT DIV ", why would this be the case ? If you put the leader consumed record into drainer, during drainer process, it should perform DIV as well right

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -3518,6 +3542,82 @@ private void produceToLocalKafkaHelper(
...
+   */
+  void sendGlobalRtDivMessage(
+      PubSubMessage<KafkaKey, KafkaMessageEnvelope, Long> previousMessage,
+      PartitionConsumptionState partitionConsumptionState,
+      int partition,
+      String brokerUrl,
+      long beforeProcessingRecordTimestampNs,
+      LeaderMetadataWrapper leaderMetadataWrapper,
+      LeaderProducedRecordContext leaderProducedRecordContext) {
+    final InternalAvroSpecificSerializer<GlobalRtDivState> serializer =
```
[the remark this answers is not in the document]
KaiSernLim: I would love to, but I don't think that the `InternalAvroSpecificSerializer` is thread-safe.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -3518,6 +3542,82 @@ private void produceToLocalKafkaHelper(
...
+    // Create PubSubMessage for the LeaderProducerCallback to enqueue the RT + VT DIV to the drainer
+    KafkaKey divKey = new KafkaKey(MessageType.GLOBAL_RT_DIV, keyBytes);
+    // TODO: incrementSequenceNumber? are the pubsubmessage fields correct?
+    KafkaMessageEnvelope divEnvelope = getVeniceWriter(partitionConsumptionState).get()
+        .getKafkaMessageEnvelope(
+            MessageType.PUT,
+            false,
+            partition,
+            true,
+            leaderMetadataWrapper,
```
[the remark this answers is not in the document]
KaiSernLim: Changed to `DEFAULT_LEADER_METADATA_WRAPPER`.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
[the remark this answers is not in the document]
sixpluszero: Can we actually embed it inside VW so it's then it will only be used by the writer?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -3518,6 +3545,89 @@ private void produceToLocalKafkaHelper(
...
+            false,
+            partition,
+            true,
+            DEFAULT_LEADER_METADATA_WRAPPER,
+            APP_DEFAULT_LOGICAL_TS);
+    divEnvelope.payloadUnion = new DivSnapshot(divClone, previousMessage.getOffset()); // contains both VT + RT DIV
+    PubSubMessage<KafkaKey, KafkaMessageEnvelope, Long> divMessage = new ImmutablePubSubMessage<>(
+        divKey,
+        divEnvelope,
+        previousMessage.getTopicPartition(),
```
sixpluszero, line 3601: For this PR in current status, it might be OK, but I wonder if we really drain it in the drainer, will it treat it as a duplicate message and not save it?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -3518,6 +3545,89 @@ private void produceToLocalKafkaHelper(
...
+        .put(
+            keyBytes,
+            ByteUtils.extractByteArray(serializer.serialize(globalRtDiv)),
+            partition,
+            1, // dummy value schema id which shouldn't be used for MessageType.GLOBAL_RT_DIV
+            divCallback,
+            leaderMetadataWrapper,
+            APP_DEFAULT_LOGICAL_TS,
+            null,
+            null, // TODO: do oldValueManifest and rmd need to be populated?
```
sixpluszero, line 3624: Since this will likely going to be chunked, then we should as well get the old manifest and perform the actual deletion.

sixpluszero (commented):
Thank you for your change. Overall it is good, but I left another couple questions in the producing part which is still confusing to me.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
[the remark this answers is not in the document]
KaiSernLim: ~~VW is on the PCS, so that would make it per-partition. There will be multiple consumers (per-broker) for a partition, so I still don't think that would make things thread-safe?~~

Talked with @lluwm and @sixpluszero offline. We'll try it per-SIT as it seems to be thread-safe?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
KaiSernLim: `validateMessage()` in `internalProcessRecord()` using the DIV would definitely throw a duplicate message exception.

But when the feature flag is turned on, the message validation should only be on the consumer and not the drainer. This still needs to be added, and it shouldn't be a problem afterwards.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -3536,6 +3563,101 @@ private void produceToLocalKafkaHelper(
...
+        partition,
+        brokerUrl,
+        beforeProcessingRecordTimestampNs);
+
+    // Get the old value manifest which contains the list of old chunks, so they can be deleted
+    ChunkedValueManifestContainer valueManifestContainer = new ChunkedValueManifestContainer();
+    final int schemaId = AvroProtocolDefinition.GLOBAL_RT_DIV_STATE.getCurrentProtocolVersion();
+    readStoredValueRecord(partitionConsumptionState, keyBytes, schemaId, topicPartition, valueManifestContainer);
+
+    // Produce to local kafka for the Global RT DIV + latestOffset (GlobalRtDivState)
```
lluwm, line 3643: Can we update the comments a bit. e.g.
// Produce to local kafka for the Global RT DIV + latestOffset (GlobalRtDivState)
// VeniceWriter.put() will schedule DELETEs for the old chunks in the old manifest after the new PUT

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -2700,18 +2705,38 @@ public void processConsumerRecord(
...
+  protected boolean shouldSendGlobalRtDiv(DefaultPubSubMessage record, PartitionConsumptionState pcs, String kafkaUrl) {
+    if (!isGlobalRtDivEnabled()) {
+      return false;
+    }
+
+    // The Global RT DIV is sent on a per-broker basis, so divide the size limit by the number of brokers
+    final long syncBytesInterval = getSyncBytesInterval(pcs) / getConsumedBytesSinceLastSync().size();
+    boolean shouldSync = false;
+    if (!record.getKey().isControlMessage()) {
+      shouldSync = (syncBytesInterval > 0 && (getConsumedBytesSinceLastSync().get(kafkaUrl) >= syncBytesInterval));
```
lluwm, line 2735: Nit, remove outermost parentheses?
`shouldSync = (syncBytesInterval > 0 && (getConsumedBytesSinceLastSync().get(kafkaUrl) >= syncBytesInterval));`

clients/da-vinci-client/src/main/java/com/linkedin/davinci/validation/DivSnapshot.java
```diff
@@ -0,0 +1,23 @@
+package com.linkedin.davinci.validation;
+
+/**
+ * Wrapper class for {@link PartitionTracker} with latestConsumedRtOffset (LCRO) for the ConsumptionTask to enqueue
+ * to the Drainer. Contains the VT DIV (Segments) + LCVO and RT DIV (Segments) + LCRO.
+ */
+public class DivSnapshot {
+  private PartitionTracker partitionTracker;
+  private long latestConsumedRtOffset; // LCVO is in PartitionTracker
```
lluwm, line 9: Nit, can we make them `final`? This snapshot should be read-only.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -3536,6 +3563,101 @@ private void produceToLocalKafkaHelper(
...
+    getVeniceWriter(partitionConsumptionState).get()
+        .put(
+            keyBytes,
+            valueBytes,
+            partition,
+            1, // dummy value schema id which shouldn't be used for MessageType.GLOBAL_RT_DIV
```
sixpluszero, line 3649: Should it be schemaId instead?

Just would like to get the clarification here: Do we plan to resolve the downstream dependency in this PR ? 
I think this new message type will break consumption of DVC and CDC and other utils. I wonder if we want to roll them all out first before we enable it on server; or we can try out making this version level config and selectively enable it on certain store?

sixpluszero: OK I found this 
    int schemaId =
        (isPutMessage) ? valueSchemaId : AvroProtocolDefinition.GLOBAL_RT_DIV_STATE.getCurrentProtocolVersion();
in the VW code, so maybe 1 doesn't matter, but still the major concern is the rollout process.

KaiSernLim: The dependency issue will not be resolved in this PR. The store-level config is a great idea, and I'll do that in a future PR.

lluwm (approved):
Thanks for this PR, @KaiSernLim . It looks very good to me!

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
sixpluszero: OK yeah. Because I think if we don't do this, you will have to gone through the wait Lei has gone through for feature rollout. And you simply cannot even turn it on in server side to validate the behavior before all the rollout

lluwm: Also like this idea to make the rollout process easier. Let's do it in a follow-up PR.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -3536,6 +3563,102 @@ private void produceToLocalKafkaHelper(
...
+            DEFAULT_LEADER_METADATA_WRAPPER,
+            APP_DEFAULT_LOGICAL_TS);
+    divEnvelope.payloadUnion = new DivSnapshot(divClone, offset); // contains both VT + RT DIV
+    DefaultPubSubMessage divMessage = new ImmutablePubSubMessage(
+        divKey,
+        divEnvelope,
+        topicPartition,
+        previousMessage.getPosition(),
+        previousMessage.getPubSubMessageTime(),
+        divKey.getHeapSize() + valueBytes.length);
```
FelixGV, line 3629: This is not right... the heap size is the size within the Java heap... whereas the payload size is supposed to represent the serialized size within Kafka...

KaiSernLim: Should it be `divKey.getKeyLength()`, instead?

PR #1577, opened 2025-03-05
[controller] Incorporating Get Dead Store Controller Endpoint

MikeDafi:
## Summary: [controller] Introducing `/get_dead_stores` Endpoint
This PR introduces the `/get_dead_stores` endpoint to the Venice Controller, along with supporting infrastructure to determine when a store is considered "dead." It adds an abstraction (`DeadStoreStats`) that allows plugging in custom logic to identify stores that are no longer active.
The endpoint is feature-gated and only available on parent controllers. This work lays the foundation for dead store detection without tying the logic to any specific internal system.

## How was this PR tested?

Several unit and integration tests were added to validate the components end-to-end:

- Store metadata (`isStoreDead`, `storeDeadStatusReasons`) was verified in `TestStoreInfo` to ensure proper default values and update behavior.

- Controller instantiation was tested in `TestVeniceHelixAdminWithIsolatedEnvironment` by configuring a mock `DeadStoreStats` implementation and confirming the `deadStoreStats` field initializes correctly when the controller is a parent and the feature flag is enabled.

- The endpoint’s logic was verified using a fake stats implementation that labels one store as “dead” and another as active. The endpoint correctly filters out the active store and returns dead store metadata with appropriate reasons.

- A background task that periodically invokes the `preFetchStats` method was also tested to ensure it starts, runs a few cycles, and shuts down cleanly without errors or resource leakage.

✅ Yes

### New Admin Endpoint (Feature-Gated)
```
GET /get_dead_stores
```

### Behavior
Returns a list of stores flagged as “dead” based on custom logic from a configured `DeadStoreStats` class. Each entry includes:
- `storeName`
- `isStoreDead = true`
- `storeDeadStatusReasons = [...]`

### Query Parameters
| Parameter                 | Required | Description                                                              |
|---------------------------|----------|--------------------------------------------------------------------------|
| `cluster=<clusterName>`   | One of `cluster` or `name` is required | Filters to stores in the specified cluster           |
| `name=<storeName>`        | One of `cluster` or `name` is required | Filters to a specific store                          |
| `includeSystemStores=true`| Optional | Include internal/system stores (default: excluded)                       |

### Availability & Config
The endpoint is only active if:
- The controller is a **parent**
- The following config is set:
  ```
  controller.dead.store.endpoint.enabled=true
  controller.dead.store.stats.class.name=com.example.MyDeadStoreStatsImpl
  ```

### Example Response
```json
[
  {
      ...
      "name": "test-store",
      "isStoreDead": true,
      "storeDeadStatusReasons": ["No traffic detected"]
      ...
    },
]
```


services/venice-controller/src/main/java/com/linkedin/venice/controller/stats/DeadStoreStats.java
```diff
@@ -0,0 +1,17 @@
...
+/*
+ * Interface for determining dead stores stats
+ */
+public interface DeadStoreStats {
+
+  /*
+   * Intends to only return the dead stores.
+   * Side effect operation of populating isStoreDead and reasonsStoreIsDead in the StoreInfo object.
+   */
+  List<StoreInfo> getDeadStores(List<StoreInfo> storeInfos) throws Exception;
```
mynameborat, line 16: Is there a reason that this needs to throw a checked exception? If so what is the expectation of the callers in terms of handling this exception.

Also, any pointers to the implementors on how this needs to be implemented, what are the considerations? like thread safety? invocation patterns etc

mynameborat, line 16: Lastly, seems like you want to also propagate the reason for deletion. Which component is providing that information?

Is `List<StoreInfo>` sufficient to capture all the details. Should we have `DeadStoreStats` provide additional information on why a store is selected for deletion which can then be propagated as needed.

Given this is public interface, might be good to think through this and decide so that we don't modify the API/break its compatibility.

mynameborat (commented):
Have few questions. Let me know if you want to chat.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -8104,6 +8105,17 @@ public Map<String, RegionPushDetails> listStorePushInfo(
     throw new UnsupportedOperationException("This function has not been implemented.");
   }
 
+  /**
+   * @see Admin#getDeadStores(String)
+   */
+  @Override
+  public List<StoreInfo> getDeadStores(String clusterName) {
```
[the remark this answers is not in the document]
MikeDafi: Updated the documentation to describe a dead store
```
  /**
   * @return list of stores infos that are considered dead. A store is considered dead if it exists but has no
   * user traffic in it's read or write path.
   *
   * Linkedin deduces a store is dead if the acls associated with the store's
   * current, future, and backup versions have no activity by non Venice services (excluding venice services to avoid
   * venice ingestion traffic which doesn't reflect user traffic).
   */
```

I don't think there's significance in having the logic in the parent controller vs not.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -8104,6 +8105,17 @@ public Map<String, RegionPushDetails> listStorePushInfo(
...
+    checkControllerLeadershipFor(clusterName);
+    HelixVeniceClusterResources resources = getHelixVeniceClusterResources(clusterName);
+    List<Store> storeList = resources.getStoreMetadataRepository().getAllStores();
```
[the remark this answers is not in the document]
MikeDafi: Agreed, updated to 
```java
  public List<StoreInfo> getDeadStores(String clusterName) {
    return new DeadStoreStats().getDeadStores(getAllStores(clusterName))
  }
```

services/venice-controller/src/main/java/com/linkedin/venice/controller/server/StoresRoutes.java
```diff
@@ -1014,6 +1015,21 @@ public void internalHandle(Request request, VersionResponse veniceResponse) {
...
+   * @see Admin#getDeadStores(String)
+   */
+  public Route getDeadStores(Admin admin) {
+    return new VeniceRouteHandler<MultiStoreInfoResponse>(MultiStoreInfoResponse.class) {
+      @Override
+      public void internalHandle(Request request, MultiStoreInfoResponse veniceResponse) {
+        AdminSparkServer.validateParams(request, GET_DEAD_STORES.getParams(), admin);
+        String cluster = request.queryParams(CLUSTER);
+        List<StoreInfo> storeList = admin.getDeadStores(cluster);
+        veniceResponse.setStoreInfoList(storeList);
```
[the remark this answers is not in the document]
MikeDafi: I don't think so, I hadn't seen others place metrics on the request count or latency of a controller request

services/venice-controller/src/main/java/com/linkedin/venice/controller/stats/DeadStoreStats.java
MikeDafi: No reason to throw an exception, I'll remove it

MikeDafi: Yeah, I had initially considered having a custom `MultiStoreResponse` but I think having information about the store is significant as well so adding additional info to StoreInfo seemed appropriate (correct me if I'm wrong).

I'm propagating the reasons the store is dead and the fact that the store is dead or not via `isStoreDead` and `reasonsStoreIsDead`

internal/venice-common/src/main/java/com/linkedin/venice/meta/StoreInfo.java
```diff
@@ -338,6 +339,8 @@ public static StoreInfo fromStore(Store store) {
   private String targetRegionSwap;
   private int targetRegionSwapWaitTime;
   private boolean isDavinciHeartbeatReported;
+  private final List<String> reasonsStoreIsDead = new ArrayList<>();
```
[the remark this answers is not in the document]
MikeDafi: In the case we're not calling the get dead store endpoint, then when we decide to return back a `List<StoreInfo>` via another controller endpoint, we should have the store info's reasonsStoreIsDead to be an empty list or null (I had seen syntax checker was complaining if object is set to null so set to empty list)

services/venice-controller/src/main/java/com/linkedin/venice/controller/Admin.java
```diff
@@ -958,6 +958,16 @@ default boolean isAdminTopicConsumptionEnabled(String clusterName) {
    */
   int getLargestUsedVersionFromStoreGraveyard(String clusterName, String storeName);
 
+  /**
+   * @return list of stores infos that are considered dead. A store is considered dead if it exists but has no
+   * user traffic in it's read or write path.
+   *
+   * Linkedin deduces a store is dead if the acls associated with the store's
```
ZacAttack, line 965: Let's nix the linkedIn specific stuff since we're in open source

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -8104,6 +8105,22 @@ public Map<String, RegionPushDetails> listStorePushInfo(
     throw new UnsupportedOperationException("This function has not been implemented.");
   }
 
+  /**
+   * @see Admin#getDeadStores(String)
+   */
+  @Override
+  public List<StoreInfo> getDeadStores(String clusterName, String storeName) {
+    if (storeName == null) {
+      return new DeadStoreStats().getDeadStores(getAllStores(clusterName));
```
ZacAttack: So, I'm not sure how this will work?  What does this do?  DeadStoreStats is an interface, but you're not providing an implementation in line here.

Take a look at ReflectionUtils and how we use it.  This is how we typically do dependency injection, we'll need to provide for a LinkedIn version of this which interfaces with the stats/ACL system.

MikeDafi: Yeah, the latest PR had the logic to use `ReflectionUtils`  and the class name that will be implemented in livenice

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -8108,6 +8108,14 @@ public Map<String, RegionPushDetails> listStorePushInfo(
     throw new UnsupportedOperationException("This function has not been implemented.");
   }
 
+  /**
+   * @see Admin#getDeadStores(String)
+   */
+  @Override
+  public List<StoreInfo> getDeadStores(String clusterName, String storeName) {
+    throw new UnsupportedOperationException("This function is implemented in VeniceHelixAdmin.");
```
xunyin8, line 8116: This exception message looks wrong/misleading. Why are we throwing `UnsupportedOperationException("This function is implemented in VeniceHelixAdmin.");` inside of `VeniceHelixAdmin`?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
```diff
@@ -5029,6 +5046,27 @@ public CompactionManager getCompactionManager() {
...
   }
 
+  @Override
+  public List<StoreInfo> getDeadStores(String clusterName, String storeName) {
+    if (!multiClusterConfigs.isDeadStoreEndpointEnabled()) {
+      throw new VeniceUnsupportedOperationException("Dead store stats is not enabled.");
+    }
+
+    if (storeName == null) {
+      List<StoreInfo> clusterStoreInfos = getAllStores(clusterName).stream()
```
xunyin8, line 5056: This by default returns all the stores in the cluster including the per user system stores. Is this desired/expected behavior? If not we may want to filter out the system stores.

MikeDafi: Why would we want to filter out system stores?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
MikeDafi: The message should be  in`This function is implemented in VeniceParentHelixAdmin.`

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
xunyin8: In some scenarios the result of per user system store's dead state is no meaningful. e.g. A basic user store with no FC/DVC/CDC is happy and active. You are not going to delete its system stores or the actual user store even if the corresponding system stores are "dead" according to `DeadStoreStats`

internal/venice-common/src/main/java/com/linkedin/venice/controllerapi/ControllerRoute.java
```diff
@@ -273,6 +273,8 @@ public enum ControllerRoute {
   GET_STORES_FOR_COMPACTION("/get_stores_for_compaction", HttpMethod.GET, Collections.singletonList(CLUSTER)),
   COMPACT_STORE("/trigger_repush", HttpMethod.POST, Arrays.asList(NAME)),
   GET_STORE_LARGEST_USED_VERSION("/get_store_largest_used_version", HttpMethod.GET, Arrays.asList(CLUSTER, NAME)),
+
+  GET_DEAD_STORES("/get_dead_stores", HttpMethod.GET, Arrays.asList(CLUSTER, NAME)),
```
mynameborat, line 277: shouldn't this be only `CLUSTER`? `NAME` often refers to store name and might not be applicable here.

services/venice-controller/src/main/java/com/linkedin/venice/controller/Admin.java
```diff
@@ -959,6 +959,12 @@ default boolean isAdminTopicConsumptionEnabled(String clusterName) {
    */
   int getLargestUsedVersionFromStoreGraveyard(String clusterName, String storeName);
 
+  /**
+   * @return list of stores infos that are considered dead. A store is considered dead if it exists but has no
+   * user traffic in it's read or write path.
+   */
+  List<StoreInfo> getDeadStores(String clusterName, String storeName);
```
mynameborat, line 966: why does this API need storeName?

internal/venice-common/src/main/java/com/linkedin/venice/controllerapi/ControllerRoute.java
MikeDafi: `CLUSTER` will be an argument passed in from the batch deletion tooling
`NAME` will be the single store query made from Nuage.

We had also talked about this in person, thank you

services/venice-controller/src/main/java/com/linkedin/venice/controller/Admin.java
MikeDafi: In the case where we could be querying store name for nuage or by clustername for batch deletion purposes. We could have two separate endpoints instead of overloading an endpoint but this storeName over clusterName priority  is a streamlined approach as well

MikeDafi:
We're seeing a diffCoverage failure:
```
java.lang.Exception: Rule violated for bundle venice-controller: branches covered ratio is 0.00, but expected minimum is 0.45
```

Screenshot 2025-04-03 at 1 37 11 AM

To reach the required 45% branch coverage, we’d need to add unit tests for `HelixVeniceClusterResources`, specifically around the start/stop logic of `DeadStoreStatsPreFetchTask`. However, this logic is already exercised by our integration tests—though those don't count toward diffCoverage.

Writing unit tests for this would require heavy mocking of many internal components, just to cover ~10 lines that are already well-tested in other new tests. In my opinion, given the limited added value and high maintenance cost, I’d recommend skipping unit-level mocks against `HelixVeniceClusterResources` unless there's a strong need.

services/venice-controller/src/main/java/com/linkedin/venice/controller/DeadStoreStatsPreFetchTask.java
```diff
@@ -0,0 +1,68 @@
...
+    this.refreshIntervalMs = refreshIntervalMs;
+    this.taskId = String.format(TASK_ID_FORMAT, clusterName);
+    this.logger = LogManager.getLogger(taskId);
+  }
+
+  @Override
+  public void run() {
+    logger.info("Started {}", taskId);
+    isRunning.set(true);
+    admin.preFetchDeadStoreStats(clusterName, getStoresInCluster()); // Fetch before cycling
```
xunyin8, line 41: What if the initial fetch before cycling fails. Should we at least catch the exception in case it's transient so we don't just end up with a dead task and subsequent refreshes will eventually populate the dead store stats?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -690,6 +693,20 @@ public VeniceHelixAdmin(
               ParticipantMessageKey.getClassSchema()));
     }
 
+    if (isParent() && multiClusterConfigs.isDeadStoreEndpointEnabled()) {
```
xunyin8, line 696: I think in integration test we should at least introduce a test implementation of DeadStoreStats. Enable `CONTROLLER_DEAD_STORE_ENDPOINT_ENABLED` and configure `CONTROLLER_DEAD_STORE_STATS_CLASS_NAME` appropriately. Verify that the corresponding prefetch task is triggered and called during controller start and leadership handover. Otherwise this change basically have 0 integration test to verify the newly added behavior.

services/venice-controller/src/main/java/com/linkedin/venice/controller/DeadStoreStatsPreFetchTask.java
MikeDafi: You're point makes a lot of sense. We now have a try catch for the initial fetch. Thank you

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
MikeDafi: Great point. Now, we have a test implementation of the store endpoint enabled and asserting whether the class name for deadStoreStats is indeed the mocked class name. 

As for the prefetch task, was able to call the stop/start dead store tasks in the HelixVeniceClusterResource's integration tests. The mature caller of HelixVeniceClusterResource (VeniceControllerStateModel) is part of the exclude list but given the lightweight nature of the stop/start dead tasks, I think the integration test in TestHelixVeniceClusterResources.java presents a safe assessment of stop/start dead tasks

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/controller/TestVeniceHelixAdminWithIsolatedEnvironment.java
```diff
@@ -311,4 +315,43 @@ public static boolean resourceMissingTopState(SafeHelixManager helixManager, Str
...
     return false;
   }
+
+  @Test
+  public void testDeadStoreStatsInitialization() {
+    int newAdminPort = controllerConfig.getAdminPort() + 100;
+    PropertyBuilder builder = new PropertyBuilder().put(controllerProps.toProperties())
+        .put("admin.port", newAdminPort)
+        .put("controller.dead.store.endpoint.enabled", true)
+        .put("controller.dead.store.stats.class.name", MockDeadStoreStats.class.getName());
```
xunyin8, line 325: Does this test even do anything? We are not enabling `controller.dead.store.endpoint.enabled` or `controller.dead.store.stats.pre.fetch.enabled`... If we want to actually verify that the `MockDeadStoreStats`'s `preFetchStats` is called we should add some implementation and methods so we can assert that. e.g. a boolean named `hasCache` that initializes to false. Set to true in preFetchStats so you can verify. If you are setting VeniceHelixAdmin's deadStoreStats directly via some mock you can also use `verify` to ensure certain method is invoked for a given mock.

xunyin8 (approved):

PR #1578, opened 2025-03-05
[Draft][controller]skip the version got deleted when migratinhg per fabrics

ranwang2024:

When the store is migrating and version numbers are out of sync per fabrics, each fabric should skip old versions that got deleted. 

Resolves #XXX

### DIFF
--- a/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
@@ -2738,7 +2738,7 @@ public class VeniceHelixAdmin implements Admin, StoreCleaner {
     LOGGER.info(
         "During store migration, version configs from source cluster {}: {}",
         migrationSourceCluster,
-        srcStoreResponse.getStore().getVersion(versionNumber).get());
+        srcStoreInfo.getVersion(versionNumber).get());
     return srcStoreResponse.getStore().getVersion(versionNumber);
   }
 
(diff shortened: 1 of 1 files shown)


ZacAttack (approved):
Ok.  This took me a minute to think through, but I think this is sound.  Thanks!

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestStoreMigrationMultiRegion.java
```diff
@@ -0,0 +1,288 @@
...
+        Assert.assertNotNull(storeInfo);
+      });
+    }
+    return props;
+  }
+
+  private void readFromStore(AvroGenericStoreClient<String, Object> client)
+      throws ExecutionException, InterruptedException {
+    int key = ThreadLocalRandom.current().nextInt(RECORD_COUNT) + 1;
+    client.get(Integer.toString(key)).get();
```
xunyin8, line 281: We should at least add a `notNull` assertion for the result. Otherwise it will always pass even if we tried to query non-existing keys.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -1614,6 +1614,8 @@ public void migrateStore(String srcClusterName, String destClusterName, String s
     }
 
     if (!isParent()) {
+      // This will never be executed because the `migrateStore` function is not called in the `AdminConsumptionTask`
```
xunyin8, line 1617: This new comment is confusing. I think the original intent of this code is so we can still perform migration in a single region Venice deployment. i.e. there is no admin topic and parent controller. Store migration call will go directly to the child controller of the single region.

xunyin8 (commented):
Overall LGTM, just some minor comments. Also would suggest renaming the PR to "Skip deleted/missing version(s) due to regional failure in store migration"

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestStoreMigrationMultiRegion.java
ranwang2024: Thanks for the suggestion! I've addressed the issue by adding an `assertEquals` to ensure the correct value is returned for the existing keys.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
ranwang2024: Thanks for the clarification. I have deleted the confusing comment to avoid any misunderstandings

xunyin8 (approved):

PR #1579, opened 2025-03-05
remove SystemSchemaInitializationRoutine for store properties.

pthirun:

Remove SystemSchemaInitializationRoutine for store properties in order to temporarily fix the schema parsing issue located in `AvroProtocolDefinition.SERVER_STORE_PROPERTIES_RESPONSE`. This section will be re-added to `initRoutines` once a long term fix is determined.

## How was this PR tested?

Tests will be written when this section is re added.


sushantmane (approved):
Thanks!

PR #1580, opened 2025-03-05
[server] Fix resubscription issue for partitions of stale store version to avoid useless ingestion. 

haoxu07:

When a store version is retired, the controller attempts to delete the store version from the ZooKeeper-based `StoreRepository` and sends a signal through the participant store to the server host to terminate the `StoreIngestionTask` (SIT). However, it is possible that the SIT thread, which checks the ingestion role logic, acts faster than the KILL SIT command taking effect on the server host.

Currently, store resubscription is triggered when a store version is deleted from the `StoreRepository`. This resubscription process can result in an ingestion role change. For example, if a store version is active-active enabled and later becomes stale and gets deleted, fetching this version from the `StoreRepository` will not provide information on whether it was active-active enabled. As a result, we assume it is active-active disabled, leading to an unintended ingestion role change.

Such resubscription triggers an unsubscribe and subscribe operation, which can occur before or after the SIT termination takes effect on removing all subscriptions of this SIT. As a result, some partitions may remain on the consumer.

For version topic partitions, the `ConsumerSubscriptionCleaner` can eventually remove them by checking if the version topic still exists. However, for leader real-time topic partitions of a stale version, the `ConsumerSubscriptionCleaner` is unable to clean them up.

Fixes Introduced in This PR:
Prevent unintended resubscription – Ensure that SIT does not trigger resubscription due to an ingestion role change when a stale store version is deleted from the `StoreRepository` and store is system store.

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ConsumerSubscriptionCleaner.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ConsumerSubscriptionCleaner.java
@@ -1,10 +1,16 @@
 package com.linkedin.davinci.kafka.consumer;
 
+import com.linkedin.davinci.ingestion.consumption.ConsumedDataReceiver;
+import com.linkedin.venice.kafka.protocol.KafkaMessageEnvelope;
+import com.linkedin.venice.message.KafkaKey;
+import com.linkedin.venice.pubsub.api.PubSubMessage;
 import com.linkedin.venice.pubsub.api.PubSubTopicPartition;
 import com.linkedin.venice.utils.Time;
 import it.unimi.dsi.fastutil.objects.Object2LongMap;
 import it.unimi.dsi.fastutil.objects.Object2LongOpenHashMap;
 import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
 import java.util.Set;
 import java.util.function.Consumer;
 import java.util.function.IntConsumer;
@@ -83,7 +89,8 @@ public class ConsumerSubscriptionCleaner {
    * If yes, this function will remove the topic partition subscriptions to these non-existing topics.
    */
   Set<PubSubTopicPartition> getTopicPartitionsToUnsubscribe(
-      Set<PubSubTopicPartition> returnSetOfTopicPartitionsToUnsub) {
+      Set<PubSubTopicPartition> returnSetOfTopicPartitionsToUnsub,
+      Map<PubSubTopicPartition, ConsumedDataReceiver<List<PubSubMessage<KafkaKey, KafkaMessageEnvelope, Long>>>> dataReceiverMap) {
     returnSetOfTopicPartitionsToUnsub.clear();
     if (++pollTimesSinceLastSanitization < sanitizeTopicSubscriptionAfterPollTimes) {
       /**
@@ -156,10 +163,13 @@ public class ConsumerSubscriptionCleaner {
 
     recordNumberOfTopicsToUnsub.accept(topicsToUnsubscribe.size());
 
+    Set<PubSubTopicPartition> topicPartitionsForNonAliveDataReceiver =
+        getTopicPartitionsForNonAliveDataReceiver(dataReceiverMap);
     // Get the current subscription for this topic and unsubscribe them
     Set<PubSubTopicPartition> newAssignment = new HashSet<>();
     for (PubSubTopicPartition pubSubTopicPartition: currentAssignment) {
-      if (topicsToUnsubscribe.contains(pubSubTopicPartition.getPubSubTopic().getName())) {
+      if (topicsToUnsubscribe.contains(pubSubTopicPartition.getPubSubTopic().getName())
+          || topicPartitionsForNonAliveDataReceiver.contains(pubSubTopicPartition)) {
         returnSetOfTopicPartitionsToUnsub.add(pubSubTopicPartition);
       } else {
         newAssignment.add(pubSubTopicPartition);
@@ -168,7 +178,26 @@ public class ConsumerSubscriptionCleaner {
     if (newAssignment.size() != currentAssignment.size()) {
       batchUnsubscribeFunction.accept(returnSetOfTopicPartitionsToUnsub);
     }
+    return returnSetOfTopicPartitionsToUnsub;
+  }
 
+  /**
+   * This function is used to find whether there is any subscription to the non-alive data receivers. As non-alive
+   * data receivers could not process polled records properly.
+   */
+  private Set<PubSubTopicPartition> getTopicPartitionsForNonAliveDataReceiver(
+      Map<PubSubTopicPartition, ConsumedDataReceiver<List<PubSubMessage<KafkaKey, KafkaMessageEnvelope, Long>>>> dataReceiverMap) {
+    Set<PubSubTopicPartition> returnSetOfTopicPartitionsToUnsub = new HashSet<>();
+    for (Map.Entry<PubSubTopicPartition, ConsumedDataReceiver<List<PubSubMessage<KafkaKey, KafkaMessageEnvelope, Long>>>> entry: dataReceiverMap
+        .entrySet()) {
+      if (!entry.getValue().isDataReceiverAlive()) {
+        returnSetOfTopicPartitionsToUnsub.add(entry.getKey());
+        LOGGER.warn(
+            "Detected the non-alive data receiver for data receiver of: {} for: {}",
+            entry.getValue().destinationIdentifier(),
+            entry.getKey());
+      }
+    }
     return returnSetOfTopicPartitionsToUnsub;
   }
 
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ConsumerSubscriptionCleaner.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ConsumerSubscriptionCleaner.java
@@ -1,15 +1,10 @@
 package com.linkedin.davinci.kafka.consumer;
 
-import com.linkedin.davinci.ingestion.consumption.ConsumedDataReceiver;
-import com.linkedin.venice.pubsub.api.DefaultPubSubMessage;
-import com.linkedin.venice.pubsub.api.PubSubTopic;
 import com.linkedin.venice.pubsub.api.PubSubTopicPartition;
 import com.linkedin.venice.utils.Time;
 import it.unimi.dsi.fastutil.objects.Object2LongMap;
 import it.unimi.dsi.fastutil.objects.Object2LongOpenHashMap;
 import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
 import java.util.Set;
 import java.util.function.Consumer;
 import java.util.function.IntConsumer;
@@ -45,7 +40,7 @@ public class ConsumerSubscriptionCleaner {
    * 2. The non-existing period lasts longer than {@link #nonExistingTopicCleanupDelayMS}, and the corresponding task
    *    will fail.
    */
-  private final Object2LongMap<PubSubTopic> nonExistingTopicDiscoverTimestampMap = new Object2LongOpenHashMap<>();
+  private final Object2LongMap<String> nonExistingTopicDiscoverTimestampMap = new Object2LongOpenHashMap<>();
 
   private final TopicExistenceChecker topicExistenceChecker;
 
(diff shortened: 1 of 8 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -1580,10 +1580,18 @@ protected void checkIngestionProgress(Store store) throws InterruptedException {
   }
 
   protected void updateIngestionRoleIfStoreChanged(Store store) throws InterruptedException {
+    Version version = store.getVersion(versionNumber);
```
sushantmane, line 1583: Can we use version that is already available in SIT instead of looking it up from store

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -1580,10 +1580,18 @@ protected void checkIngestionProgress(Store store) throws InterruptedException {
...
+          storeName);
+      return;
+    }
     PartitionReplicaIngestionContext.VersionRole newVersionRole =
-        PartitionReplicaIngestionContext.getStoreVersionRole(versionTopic, store);
+        PartitionReplicaIngestionContext.getStoreVersionRole(version, store);
     PartitionReplicaIngestionContext.WorkloadType newWorkloadType =
-        PartitionReplicaIngestionContext.getWorkloadType(versionTopic, store);
+        PartitionReplicaIngestionContext.getWorkloadType(version, store);
     if (serverConfig.isResubscriptionTriggeredByVersionIngestionContextChangeEnabled() && isHybridMode()) {
```
sushantmane, line 1595: Let’s move the newVersionRole and newWorkloadType lookups inside the if block to prevent unnecessary memory allocations.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
sushantmane, line 1: Here’s another optimization: Instead of passing the versionTopic, we can pass the version number directly. This eliminates the need to call `Version.parseVersionFromKafkaTopicName` on each invocation, reducing redundant processing.
```java

public static VersionRole getStoreVersionRole(int versionNumber, Store store) {
    int currentVersionNumber = store.getCurrentVersion();
    if (currentVersionNumber < versionNumber) {
        return VersionRole.FUTURE;
    } else if (currentVersionNumber > versionNumber) {
        return VersionRole.BACKUP;
    } else {
        return VersionRole.CURRENT;
    }
}

```

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
haoxu07: After discussed, we could use that version information inside SIT.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ConsumerSubscriptionCleaner.java
```diff
@@ -156,10 +161,13 @@ Set<PubSubTopicPartition> getTopicPartitionsToUnsubscribe(
 
     recordNumberOfTopicsToUnsub.accept(topicsToUnsubscribe.size());
 
+    Set<PubSubTopicPartition> topicPartitionsForNonAliveDataReceiver =
```
sushantmane, line 164: Actually we don't need to do this here, we can just to data receiver liveness check in for loop (L#168+). This way we can avoid double iterations

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/PartitionReplicaIngestionContext.java
```diff
@@ -61,22 +58,14 @@ public WorkloadType getWorkloadType() {
...
 
-  public static WorkloadType getWorkloadType(PubSubTopic versionTopic, Store store) {
-    checkVersionTopicStoreOrThrow(versionTopic, store);
-    int versionNumber = Version.parseVersionFromKafkaTopicName(versionTopic.getName());
-    Version version = store.getVersion(versionNumber);
-    if (version == null) {
-      return WorkloadType.NON_AA_OR_WRITE_COMPUTE;
-    }
-    if (store.isWriteComputationEnabled() || version.isActiveActiveReplicationEnabled()) {
+  public static WorkloadType getWorkloadType(boolean isActiveActiveReplicationEnabled, Store store) {
```
sushantmane, line 61: This is not sufficient condition right? If even batch portion of data will land in  `AA_OR_WRITE_COMPUTE` pool

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/PartitionReplicaIngestionContext.java
```diff
@@ -61,22 +58,14 @@ public WorkloadType getWorkloadType() {
...
+    if (store.isWriteComputationEnabled() || isActiveActiveReplicationEnabled) {
       return WorkloadType.AA_OR_WRITE_COMPUTE;
     }
     return WorkloadType.NON_AA_OR_WRITE_COMPUTE;
   }
 
-  public static VersionRole getStoreVersionRole(PubSubTopic versionTopic, Store store) {
-    checkVersionTopicStoreOrThrow(versionTopic, store);
-    int versionNumber = Version.parseVersionFromKafkaTopicName(versionTopic.getName());
+  public static VersionRole getStoreVersionRole(int versionNumber, Store store) {
```
sushantmane, line 68: What happens in case of version rollbacks? Backup and future version pools might be incorrect? 

Since there are returns in each conditions we don't need else statements right
Can we do
```
int currentVersionNumber = store.getCurrentVersion();

if (currentVersionNumber < versionNumber) {
    return VersionRole.FUTURE;
}
if (currentVersionNumber > versionNumber) {
    return VersionRole.BACKUP;
}
return VersionRole.CURRENT;
```

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -1581,12 +1581,16 @@ protected void checkIngestionProgress(Store store) throws InterruptedException {
   }
 
   protected void updateIngestionRoleIfStoreChanged(Store store) throws InterruptedException {
-    PartitionReplicaIngestionContext.VersionRole newVersionRole =
-        PartitionReplicaIngestionContext.getStoreVersionRole(versionTopic, store);
-    PartitionReplicaIngestionContext.WorkloadType newWorkloadType =
-        PartitionReplicaIngestionContext.getWorkloadType(versionTopic, store);
     if (serverConfig.isResubscriptionTriggeredByVersionIngestionContextChangeEnabled() && isHybridMode()) {
```
sushantmane, line 1584: Can we always disable pool switching for user system stores?

sushantmane (changes requested):
Could you please rebase your PR to use `Map<PubSubTopicPartition, ConsumedDataReceiver<List<DefaultPubSubMessage>>>

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/PartitionReplicaIngestionContext.java
sushantmane: This is handled during subscription. So no change required

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -1581,12 +1581,16 @@ protected void checkIngestionProgress(Store store) throws InterruptedException {
   }
 
   protected void updateIngestionRoleIfStoreChanged(Store store) throws InterruptedException {
```
sushantmane, line 1583: We should rename `updateIngestionRoleIfStoreChanged` since it doesn’t accurately reflect what the method does. The ingestion role isn’t being changed directly. Instead, the method refreshes the ingestion context when necessary. A more suitable name could be `refreshIngestionContextIfChanged` or please use something that better captures its purpose.

haoxu07: Yeah, refresh is better term.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/ConsumerSubscriptionCleaner.java
sushantmane: It also looks like this might unintentionally remove some subscriptions from the consumer task.

For example, suppose:
	•	test_store_rt-0 is ingesting data for SIT_v1
	•	test_store_rt-1 is ingesting data for SIT_v2

If SIT_v1 is no longer alive, we might mistakenly remove the subscription for test_store_rt-1 as well due to the check in the later part of the code.

haoxu07: After discussion, we decided to revert the change for data receiver temporarily, as we plan to refactor the the clearner in future PR.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -1580,13 +1580,18 @@ protected void checkIngestionProgress(Store store) throws InterruptedException {
...
 
-  protected void updateIngestionRoleIfStoreChanged(Store store) throws InterruptedException {
-    PartitionReplicaIngestionContext.VersionRole newVersionRole =
-        PartitionReplicaIngestionContext.getStoreVersionRole(versionTopic, store);
-    PartitionReplicaIngestionContext.WorkloadType newWorkloadType =
-        PartitionReplicaIngestionContext.getWorkloadType(versionTopic, store);
-    if (serverConfig.isResubscriptionTriggeredByVersionIngestionContextChangeEnabled() && isHybridMode()) {
+  protected void refreshIngestionContextIfChanged(Store store) throws InterruptedException {
+    if (serverConfig.isResubscriptionTriggeredByVersionIngestionContextChangeEnabled() && isHybridMode()
+        && !isUserSystemStore) {
```
sushantmane, line 1585: This will still let participant systems store and push details store switch pools?

sushantmane: Consider using early return to reduce unnecessary indentation and improve readability. By negating the conditions at the beginning, we can exit early and avoid deep nesting. This makes it easier to follow the logic, especially when handling multiple conditions like feature flags, hybrid mode checks, and system store exclusions.

```java
protected void refreshIngestionContextIfChanged(Store store) throws InterruptedException {
    if (!serverConfig.isResubscriptionTriggeredByVersionIngestionContextChangeEnabled() 
        || !isHybridMode() || isUserSystemStore) {
        return;
    }

    PartitionReplicaIngestionContext.VersionRole newVersionRole =
        PartitionReplicaIngestionContext.getStoreVersionRole(versionNumber, store);
    PartitionReplicaIngestionContext.WorkloadType newWorkloadType =
        PartitionReplicaIngestionContext.getWorkloadType(isActiveActiveReplicationEnabled, store);

    if (newVersionRole.equals(versionRole) && newWorkloadType.equals(workloadType)) {
        return;
    }

    // If the store has no current version, we do not need to resubscribe.
    if (store.getCurrentVersion() == Store.NON_EXISTING_VERSION) {
        return;
    }

    LOGGER.info(
        "Trigger for version topic: {} due to Previous: version role: {}, workload type: {} "
            + "changed to New: version role: {}, workload type: {}",
        versionTopic,
        versionRole,
        workloadType,
        newVersionRole,
        newWorkloadType);

    versionRole = newVersionRole;
    workloadType = newWorkloadType;

    try {
        resubscribeForAllPartitions();
    } catch (Exception e) {
        LOGGER.error("Error happened during resubscription when store version ingestion context changed.", e);
        hostLevelIngestionStats.recordResubscriptionFailure();
        throw e;
    }
}
```

sushantmane (changes requested):

sushantmane (approved):
LGTM. Thanks, @haoxu07!

PR #1581, opened 2025-03-05
[server]  Adjust dynamic throttler limiting at faster rate

majisourav99:
##  Adjust dynamic throttler limiting at faster rate

For dynamic throttling, during limiting signals, such as high read latency or heartbeat lag, step down should be faster than step up to prevent further breach of SLA and quicker recovery. This PR also adds throttler for leader pool type for AAWC ingestion (both non-current/current versions).

## How was this PR tested?

GHCI

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/VeniceAdaptiveIngestionThrottler.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/VeniceAdaptiveIngestionThrottler.java
@@ -4,6 +4,7 @@ import com.linkedin.venice.throttle.EventThrottler;
 import java.text.DecimalFormat;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.BooleanSupplier;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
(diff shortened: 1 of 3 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/VeniceAdaptiveIngestionThrottler.java
```diff
@@ -71,13 +72,15 @@ public void checkSignalAndAdjustThrottler() {
         hasLimitedRate = true;
         isSignalIdle = false;
         signalIdleCount = 0;
-        if (currentThrottlerIndex > 0) {
-          currentThrottlerIndex--;
+        int currentIndex = currentThrottlerIndex.get();
+        // for limiter signal, step down limit in faster rate.
+        if (currentIndex >= 1) {
+          currentThrottlerIndex.set(Math.max(currentIndex - 2, 0));
```
haoxu07, line 78: Is it possible that we want even faster than 2x? But I think it is OK to hardcode as 2 for testing now.

haoxu07 (approved):
LGTM! We can keep iterating and testing this.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/VeniceAdaptiveIngestionThrottler.java
majisourav99: We can, let me try the decrement by 2 first.

PR #1583, opened 2025-03-06
[server] Close idle ingestion tasks are batchpush unsubscription

majisourav99:
## Close idle ingestion tasks are batchpush unsubscription

After batch push the ingestion task unsubscribes, but keeps the task alive, this PR closes such tasks, prevents unnecessary resource wastage.

## How was this PR tested?

CI


sushantmane:
1. What happens when KILL message is received for the batch version after SIT is closed?
 2. Similarly what happens when replica gets dropped message

PR #1584, opened 2025-03-06
[vpj][controller] Fix push job status for target region push w/ deferred swap

misyel:

For a target region push with deferred swap, we only poll for the job status of the target regions. This, however, is incorrect because we ingest data in all regions concurrently so the push completion should mean that >50% of all regions are in terminal status. Once the push is in terminal status, we need to truncate the kafka topic so that future pushes can start, but for target region push with deferred swap, this isn't the case currently. We only truncate the kafka topic once we roll forward in the remaining non target regions. This is also incorrect because when a push is marked as complete (when >50% of the regions are in terminal status), we should be able to start another push regardless if the version is swapped in all regions

Changes included for a target region push w/ deferred swap:
- Poll job status for *all* regions
- Truncate kafka topic once >50% of all regions are in terminal status

## How was this PR tested?

Integration test

### DIFF
--- a/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
+++ b/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
@@ -410,15 +410,16 @@ public class VenicePushJob implements AutoCloseable {
 
     pushJobSettingToReturn.isTargetedRegionPushEnabled = props.getBoolean(TARGETED_REGION_PUSH_ENABLED, false);
     pushJobSettingToReturn.isSystemSchemaReaderEnabled = props.getBoolean(SYSTEM_SCHEMA_READER_ENABLED, false);
-    if (pushJobSettingToReturn.isIncrementalPush && pushJobSettingToReturn.isTargetedRegionPushEnabled) {
+    pushJobSettingToReturn.isTargetRegionPushWithDeferredSwapEnabled =
+        props.getBoolean(TARGETED_REGION_PUSH_WITH_DEFERRED_SWAP, false);
+    if (pushJobSettingToReturn.isIncrementalPush && (pushJobSettingToReturn.isTargetedRegionPushEnabled
+        || pushJobSettingToReturn.isTargetRegionPushWithDeferredSwapEnabled)) {
       throw new VeniceException("Incremental push is not supported while using targeted region push mode");
     }
 
-    // If target region push with deferred version swap is enabled, enable deferVersionSwap and
-    // isTargetedRegionPushEnabled
-    if (props.getBoolean(TARGETED_REGION_PUSH_WITH_DEFERRED_SWAP, false)) {
+    // If target region push with deferred version swap is enabled, enable deferVersionSwap
+    if (pushJobSettingToReturn.isTargetRegionPushWithDeferredSwapEnabled) {
       pushJobSettingToReturn.deferVersionSwap = true;
-      pushJobSettingToReturn.isTargetRegionPushWithDeferredSwapEnabled = true;
     }
 
     if (pushJobSettingToReturn.isTargetRegionPushWithDeferredSwapEnabled
--- a/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
+++ b/clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
@@ -854,7 +854,8 @@ public class VenicePushJob implements AutoCloseable {
             controllerClient,
             pushJobSetting,
             pushJobSetting.targetedRegions,
-            pushJobSetting.isTargetedRegionPushEnabled);
+            pushJobSetting.isTargetedRegionPushEnabled,
+            pushJobSetting.isTargetRegionPushWithDeferredSwapEnabled);
       }
 
       updatePushJobDetailsWithCheckpoint(PushJobCheckpoints.JOB_STATUS_POLLING_COMPLETED);
@@ -999,6 +1000,7 @@ public class VenicePushJob implements AutoCloseable {
           controllerClient,
           pushJobSetting,
           RegionUtils.composeRegionList(candidateRegions),
+          false,
           false);
     }
     pushJobDetails.overallStatus.add(getPushJobDetailsStatusTuple(PushJobDetailsStatus.COMPLETED.getValue()));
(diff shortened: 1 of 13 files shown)


internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestDeferredVersionSwap.java
```diff
@@ -131,4 +132,72 @@ public void testDeferredVersionSwap() throws IOException {
...
+            Assert.assertEquals((int) version, 1);
+          } else {
+            Assert.assertEquals((int) version, 0);
+          }
+        });
+      });
+
+      TestUtils.waitForNonDeterministicAssertion(30, TimeUnit.SECONDS, () -> {
+        StoreInfo parentStore = parentControllerClient.getStore(storeName).getStore();
+        Assert.assertEquals(parentStore.getVersion(1).get().getStatus(), VersionStatus.PUSHED);
```
majisourav99, line 178: can we also check the child colo stores have version set to 1 in both colo after some wait ?

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestDeferredVersionSwap.java
```diff
@@ -131,4 +132,72 @@ public void testDeferredVersionSwap() throws IOException {
...
+      });
+
+      // Start a normal push
+      props.put(TARGETED_REGION_PUSH_WITH_DEFERRED_SWAP, false);
```
majisourav99, line 182: also does this test capture the case when the one of non-target child colo fails and still we are able to start a new job ?

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestDeferredVersionSwap.java
misyel: We do that in the previous test, but for this test we explicitly don't want to wait because we want to verify that we can start a second push if the first push is complete even if the version hasn't been swapped yet

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestDeferredVersionSwap.java
misyel: It should trigger the same case as the one the test checks for. The test checks for the case where the push is `COMPLETE` and the version hasn't been swapped yet in the non target region, then we should be able to start a new push. If we have the case where we have 2 colo and the following is true:

- target region: `COMPLETED`
- non target region: `ERROR`

The final push status is `COMPLETED` so the truncate topic logic should trigger similar to the case above and we should be able to start a new push

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
```diff
@@ -3845,8 +3856,11 @@ private void truncateTopicsOptionally(
       boolean incPushEnabledBatchPushSuccess = !incrementalPushVersion.isPresent() && store.isIncrementalPushEnabled();
```
majisourav99, line 3856: GH does not allow me to post comment on code far way from changed lines :)  can you  please update the comments before this to indicate clearly when we can truncate the parent topic/

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
```diff
@@ -3717,13 +3727,15 @@ private OfflinePushStatusInfo getOffLineJobStatus(
...
+          boolean isTargetRegionPushWithDeferredSwapForCurrentVersion = isTargetRegionPush && !isVersionPushed;
           // Truncate topic after push is in terminal state if
           // 1. Its a hybrid store or regular push. (Hybrid store target push uses repush where isTargetRegionPush is
           // false)
           // 2. If target region push is enabled and job to push data only to target region completed (status == PUSHED)
+          // 3. If it is a target region push with deferred swap and a majority of regions have reached terminal status
           if (!isTargetRegionPush // regular push
               || isVersionPushed // target region push
-              || isHybridStore) {
+              || isHybridStore || isTargetRegionPushWithDeferredSwapForCurrentVersion) {
```
majisourav99, line 3738: Are we supporting hybrid stores for deferred swap target push? If so can you add an integ test for that feature?

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestDeferredVersionSwap.java
```diff
@@ -131,4 +132,134 @@ public void testDeferredVersionSwap() throws IOException {
...
+    String inputDirPath = "file://" + inputDir.getAbsolutePath();
+    String storeName = Utils.getUniqueString("testDeferredVersionSwapWithHybridStore");
+    Properties props =
+        IntegrationTestPushUtils.defaultVPJProps(multiRegionMultiClusterWrapper, inputDirPath, storeName);
+    String keySchemaStr = "\"string\"";
+    UpdateStoreQueryParams storeParams = new UpdateStoreQueryParams().setUnusedSchemaDeletionEnabled(true)
+        .setHybridOffsetLagThreshold(TEST_TIMEOUT)
+        .setHybridRewindSeconds(2L)
+        .setActiveActiveReplicationEnabled(true)
+        .setIncrementalPushEnabled(true)
```
majisourav99, line 217: I mean we should have an unit test for inc push, and we should fail that in vpj as we do not support inc push in deferred push,

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/integration/utils/VeniceTwoLayerMultiRegionMultiClusterWrapper.java
```diff
@@ -448,4 +452,28 @@ public void logMultiCluster() {
       }
     }
   }
+
+  public void useLeaderErrorNotifier(String regions) {
```
majisourav99, line 456: Cool!! Thanks for adding this. nit: can rename it to something like: `failPushInRegion()`

clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/VenicePushJob.java
```diff
@@ -705,6 +706,13 @@ public void run() {
       validateStoreSettingAndPopulate(controllerClient, pushJobSetting);
       inputStorageQuotaTracker = new InputStorageQuotaTracker(pushJobSetting.storeStorageQuota);
 
+      Set<String> nonTargetRegionsList = getNonTargetRegions();
```
majisourav99, line 709: can we move this check into validateStoreSettingAndPopulate.

majisourav99 (approved):
LGTM. Thanks @misyel !

majisourav99 (approved):

PR #1585, opened 2025-03-07
[producer] Make ProducerTool prioritize newer schemas over older ones

nisargthakkar:
## Make `ProducerTool` prioritize newer schemas over older ones

In `ProducerTool`, the input is Avro's JSON encoded data, and we use Avro serialization test to see if the json encoded data maps to a particular Avro schema. This seems to not work as expected, and we end up picking an incorrect schema to write. As a short-term improvement, this PR changes the order in which schemas are tested for to be looked at from newest to oldest, as users are more likely to write data with newer schemas.

## How was this PR tested?

Tried manually

### DIFF
--- a/clients/venice-producer/src/main/java/com/linkedin/venice/producer/online/ProducerTool.java
+++ b/clients/venice-producer/src/main/java/com/linkedin/venice/producer/online/ProducerTool.java
@@ -12,13 +12,16 @@ import com.linkedin.venice.client.store.ClientConfig;
 import com.linkedin.venice.client.store.ClientFactory;
 import com.linkedin.venice.client.store.StatTrackingStoreClient;
 import com.linkedin.venice.exceptions.VeniceException;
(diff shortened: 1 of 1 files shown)


sushantmane (approved):
LGTM, but you’ll need to add tests to satisfy the CC checks.

nisargthakkar:
Since this is a non-critical tool, it has been excluded from tests when it was added. We can revisit that decision, but for now, I think it's okay to merge as is. WDYT?

sushantmane (approved):
> Since this is a non-critical tool, it has been excluded from tests when it was added. We can revisit that decision, but for now, I think it's okay to merge as is. WDYT?




Sounds good. Though, eventually, we should remove it from the excluded files and add tests for it. Several customers seem to rely on this. Also, since we add tests for the admin tool, we might follow the same approach here?

PR #1588, opened 2025-03-07
[router] Enable SSL handshake offload to a thread pool and DNS resolve through a thread pool at the same time

huangminchn:
## Summary
Previously a separate thread pool is only used for DNS resolution before SSL handshakes; it turns out that
we can assign a thread pool for SSL handshake as well. If both features are enabled, the execution order in
the Netty pipeline would be:
Decode, verify SSL is enabled -> Run DNS resolution asynchronously -> Retry DNS resolution within a budget
and if connection is active -> Regardless if resolution is successful, start SSL handshake if connection is still
active by swapping current handler to SSL handler -> SSL handshake -> log after handshake completes

Commends with "Step" prefix are added to improve readability of alpini SslInitializer.

Configs updates:
1. Changed "router.resolve.before.ssl" to "router.resolve.threads": 0 means DNS resolution before SSL is disabled; positive number represents the number of threads for DNS resolution.
2. Changed "router.max.concurrent.resolutions" to "router.max.concurrent.ssl.handshakes": the same config value has been used for max concurrent SSL handshakes limit, but it was wrongly named as resolution limit.

Other changes:
1. Measure latency of DNS resolution, add warning log when latency is higher than 5 seconds
3. Add debug level log when SSL handshake starts
4. First attempt to fix leaked pending SSL handshake metric, when connection is closed mid handshake, increase handshake failure counter.
5. Fixed a bug that max concurrent resolution limit is set to max ssl handshake limit
6. Added error log and return 429 when connection limits exceed; fixed the metric for connection count
7. Added request level SSL handshake task queue size metrics for both router and server.

## How was this PR tested?
Updated integration tests to enable both thread pools for DNS resolution and SSL handshake.

### DIFF
--- a/internal/alpini/netty4/alpini-netty4-base/src/main/java/com/linkedin/alpini/netty4/handlers/ConnectionControlHandler.java
+++ b/internal/alpini/netty4/alpini-netty4-base/src/main/java/com/linkedin/alpini/netty4/handlers/ConnectionControlHandler.java
@@ -15,8 +15,8 @@ import io.netty.handler.codec.http.HttpVersion;
 import java.nio.charset.StandardCharsets;
 import java.util.concurrent.Semaphore;
 import java.util.concurrent.atomic.LongAdder;
+import java.util.function.Consumer;
 import java.util.function.IntSupplier;
-import javax.annotation.Nonnegative;
 import javax.annotation.Nonnull;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
@@ -50,17 +50,16 @@ public class ConnectionControlHandler extends ConnectionLimitHandler {
   private final Semaphore _activeSemaphore = new Semaphore(1);
   private final Semaphore _inactiveSemaphore = new Semaphore(1);
 
-  public ConnectionControlHandler(@Nonnegative int connectionLimit) {
-    this(() -> connectionLimit);
-  }
-
-  public ConnectionControlHandler(@Nonnull IntSupplier connectionLimit) {
-    super(connectionLimit);
+  public ConnectionControlHandler(
+      @Nonnull IntSupplier connectionLimit,
+      @Nonnull Consumer<Integer> connectionCountRecorder) {
+    super(connectionLimit, connectionCountRecorder);
   }
 
   @Override
   public void channelActive(ChannelHandlerContext ctx) throws Exception {
     _activeCount.increment();
+    _connectionCountRecorder.accept(getConnectedCount());
 
     Channel parent = ctx.channel().parent();
     if (parent != null && _activeSemaphore.tryAcquire()) {
@@ -91,6 +90,7 @@ public class ConnectionControlHandler extends ConnectionLimitHandler {
   @Override
   public void channelInactive(ChannelHandlerContext ctx) throws Exception {
     _inactiveCount.increment();
+    _connectionCountRecorder.accept(getConnectedCount());
 
     Channel parent = ctx.channel().parent();
     if (parent != null && _inactiveSemaphore.tryAcquire()) {
--- a/internal/alpini/netty4/alpini-netty4-base/src/main/java/com/linkedin/alpini/netty4/handlers/ConnectionControlHandler.java
+++ b/internal/alpini/netty4/alpini-netty4-base/src/main/java/com/linkedin/alpini/netty4/handlers/ConnectionControlHandler.java
@@ -1,18 +1,8 @@
 package com.linkedin.alpini.netty4.handlers;
 
-import static io.netty.handler.codec.http.HttpHeaderValues.TEXT_PLAIN;
-
-import io.netty.buffer.Unpooled;
 import io.netty.channel.Channel;
-import io.netty.channel.ChannelFutureListener;
 import io.netty.channel.ChannelHandler;
 import io.netty.channel.ChannelHandlerContext;
-import io.netty.handler.codec.http.DefaultFullHttpResponse;
-import io.netty.handler.codec.http.FullHttpResponse;
-import io.netty.handler.codec.http.HttpHeaderNames;
-import io.netty.handler.codec.http.HttpResponseStatus;
-import io.netty.handler.codec.http.HttpVersion;
-import java.nio.charset.StandardCharsets;
 import java.util.concurrent.Semaphore;
 import java.util.concurrent.atomic.LongAdder;
 import java.util.function.Consumer;
@@ -70,16 +60,6 @@ public class ConnectionControlHandler extends ConnectionLimitHandler {
               getConnectedCount(),
               getConnectionLimit());
           parent.config().setAutoRead(false);
-          Channel clientChannel = ctx.channel(); // Get the current client's channel
-          if (clientChannel.isActive()) {
-            FullHttpResponse response = new DefaultFullHttpResponse(
-                HttpVersion.HTTP_1_1,
-                HttpResponseStatus.TOO_MANY_REQUESTS,
-                Unpooled.copiedBuffer("Connection limit exceeded on Venice routers", StandardCharsets.US_ASCII));
-            response.headers().set(HttpHeaderNames.CONTENT_LENGTH, response.content().readableBytes());
-            response.headers().set(HttpHeaderNames.CONTENT_TYPE, TEXT_PLAIN);
-            clientChannel.writeAndFlush(response).addListener(ChannelFutureListener.CLOSE);
-          }
         }
       }).addListener(ignored -> _activeSemaphore.release());
     }
--- a/internal/alpini/netty4/alpini-netty4-base/src/main/java/com/linkedin/alpini/netty4/handlers/ConnectionControlHandler.java
+++ b/internal/alpini/netty4/alpini-netty4-base/src/main/java/com/linkedin/alpini/netty4/handlers/ConnectionControlHandler.java
@@ -1,13 +1,25 @@
 package com.linkedin.alpini.netty4.handlers;
 
+import static io.netty.handler.codec.http.HttpHeaderValues.TEXT_PLAIN;
+
+import io.netty.buffer.Unpooled;
 import io.netty.channel.Channel;
+import io.netty.channel.ChannelFutureListener;
 import io.netty.channel.ChannelHandler;
 import io.netty.channel.ChannelHandlerContext;
+import io.netty.handler.codec.http.DefaultFullHttpResponse;
+import io.netty.handler.codec.http.FullHttpResponse;
+import io.netty.handler.codec.http.HttpHeaderNames;
+import io.netty.handler.codec.http.HttpResponseStatus;
+import io.netty.handler.codec.http.HttpVersion;
+import java.nio.charset.StandardCharsets;
 import java.util.concurrent.Semaphore;
 import java.util.concurrent.atomic.LongAdder;
 import java.util.function.IntSupplier;
 import javax.annotation.Nonnegative;
 import javax.annotation.Nonnull;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 
 /**
@@ -32,6 +44,7 @@ import javax.annotation.Nonnull;
  */
 @ChannelHandler.Sharable
 public class ConnectionControlHandler extends ConnectionLimitHandler {
+  private static final Logger LOGGER = LogManager.getLogger(ConnectionControlHandler.class);
   private final LongAdder _activeCount = new LongAdder();
   private final LongAdder _inactiveCount = new LongAdder();
   private final Semaphore _activeSemaphore = new Semaphore(1);
@@ -53,7 +66,21 @@ public class ConnectionControlHandler extends ConnectionLimitHandler {
     if (parent != null && _activeSemaphore.tryAcquire()) {
       parent.eventLoop().submit(() -> {
         if (getConnectedCount() > getConnectionLimit()) {
+          LOGGER.error(
+              "Connection limit exceeded! Active connections: {}, Limit: {}",
+              getConnectedCount(),
+              getConnectionLimit());
           parent.config().setAutoRead(false);
+          Channel clientChannel = ctx.channel(); // Get the current client's channel
+          if (clientChannel.isActive()) {
+            FullHttpResponse response = new DefaultFullHttpResponse(
+                HttpVersion.HTTP_1_1,
+                HttpResponseStatus.TOO_MANY_REQUESTS,
+                Unpooled.copiedBuffer("Connection limit exceeded on Venice routers", StandardCharsets.US_ASCII));
+            response.headers().set(HttpHeaderNames.CONTENT_LENGTH, response.content().readableBytes());
+            response.headers().set(HttpHeaderNames.CONTENT_TYPE, TEXT_PLAIN);
+            clientChannel.writeAndFlush(response).addListener(ChannelFutureListener.CLOSE);
+          }
         }
       }).addListener(ignored -> _activeSemaphore.release());
     }
--- a/internal/alpini/netty4/alpini-netty4-base/src/main/java/com/linkedin/alpini/netty4/handlers/ConnectionLimitHandler.java
+++ b/internal/alpini/netty4/alpini-netty4-base/src/main/java/com/linkedin/alpini/netty4/handlers/ConnectionLimitHandler.java
@@ -7,6 +7,7 @@ import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.ChannelInboundHandlerAdapter;
 import java.nio.charset.StandardCharsets;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.Consumer;
 import java.util.function.IntSupplier;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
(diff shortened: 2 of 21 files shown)


internal/alpini/netty4/alpini-netty4-base/src/main/java/com/linkedin/alpini/netty4/ssl/SslInitializer.java
```diff
@@ -577,9 +591,19 @@ public void operationComplete(Future<String> future) {
...
         private void resolved(Future<? super Void> future) {
+          // Step 3.5: This function is called after DNS resolution completes (if _resolveClient enabled); next step is
+          // to replace itself with SSL handler. If _resolveClient not enabled, SslDetect would be replaced
+          // with SSL handler at the decode function directly, before it has a chance to reach this function.
           // This is executed on the channel executor thread
           assert _channelHandlerContext.channel().eventLoop().inEventLoop();
 
+          Long dnsStartTime = (Long) _channelHandlerContext.channel().attr(DNS_START_TS).get();
+          long dnsLatencyMs = TimeUnit.NANOSECONDS.toMillis(Time.nanoTime() - dnsStartTime);
+          LOG.log(
```
gaojieliu, line 602: Can we add a debug in func: `initChannel`? That is supposed to be the first endpoint before starting ssl handshake.

services/venice-router/src/main/java/com/linkedin/venice/router/RouterServer.java
```diff
@@ -693,34 +693,35 @@ public boolean startInner() throws Exception {
...
+            config.getClientSslHandshakeQueueCapacity(),
+            LINKED_BLOCKING_QUEUE);
+        new ThreadPoolStats(metricsRepository, sslHandshakeExecutor, "ssl_handshake_thread_pool");
+        sslInitializer.enableSslTaskExecutor(sslHandshakeExecutor);
+      }
+      if (config.getResolveThreads() > 0) {
+        ThreadPoolExecutor dnsResolveExecutor = ThreadPoolFactory.createThreadPool(
+            config.getResolveThreads(),
+            "DNSResolveThread",
+            config.getResolveThreads(),
```
gaojieliu, line 708: This param is for queue length, so I suppose we should use a high number than the thread count.

internal/alpini/netty4/alpini-netty4-base/src/main/java/com/linkedin/alpini/netty4/ssl/SslInitializer.java
huangminchn: Let me see if I can get some information about the channel into the log.

services/venice-router/src/main/java/com/linkedin/venice/router/RouterServer.java
huangminchn: Good catch! Updated

internal/alpini/netty4/alpini-netty4-base/src/main/java/com/linkedin/alpini/netty4/handlers/ConnectionControlHandler.java
```diff
@@ -32,28 +44,42 @@
...
+              "Connection limit exceeded! Active connections: {}, Limit: {}",
+              getConnectedCount(),
+              getConnectionLimit());
           parent.config().setAutoRead(false);
+          Channel clientChannel = ctx.channel(); // Get the current client's channel
+          if (clientChannel.isActive()) {
+            FullHttpResponse response = new DefaultFullHttpResponse(
+                HttpVersion.HTTP_1_1,
+                HttpResponseStatus.TOO_MANY_REQUESTS,
+                Unpooled.copiedBuffer("Connection limit exceeded on Venice routers", StandardCharsets.US_ASCII));
```
majisourav99, line 78: can we include the connection limit in the response?

internal/alpini/netty4/alpini-netty4-base/src/main/java/com/linkedin/alpini/netty4/ssl/SslInitializer.java
```diff
@@ -577,9 +604,19 @@ public void operationComplete(Future<String> future) {
...
```
majisourav99, line 615: can this blowup during new connection setup and slow dns ? Do we need to rate limit it?

internal/venice-client-common/src/main/java/com/linkedin/venice/stats/ThreadPoolStats.java
```diff
@@ -26,7 +26,19 @@ public ThreadPoolStats(MetricsRepository metricsRepository, ThreadPoolExecutor t
...
+    registerSensor(
+        new LambdaStat((ignored, ignored2) -> this.threadPoolExecutor.getQueue().size(), "queued_task_number_gauge"));
+    /**
+     * If only registered as Gauge, the metric would show the queue size at the time of the metric collection, which is not
+     * very useful. It can provide a better view of the queue size if we record the average and max queue size within
+     * the metric reporting time window which is usually 1 minute.
+     * As a result, we need the users of the thread pool to explicitly call the record function to record the queue size
+     * during each new task submission.
+     */
+    queuedTasksNumberSensor = registerSensor("queued_task_number", avgAndMax());
```
majisourav99, line 38: nit: rename to `queued_task_size` or `queued_task_count` ?

internal/alpini/netty4/alpini-netty4-base/src/main/java/com/linkedin/alpini/netty4/handlers/ConnectionControlHandler.java
lusong64: nitpick: I feel that we overload 429 too much.

huangminchn: We probably don't need to expose internal details for users; 429 should be good enough.
We have a error log about the limit on router side.

internal/alpini/netty4/alpini-netty4-base/src/main/java/com/linkedin/alpini/netty4/ssl/SslInitializer.java
huangminchn: This one is probably okay, since it's only one line per request only when DNS resolution is very slow.

I added an error log with stacktrace when SSL handshake fails, which is concerning. Planning to remove the stacktrace after finishing all the reproduction tests. Will probably keep this one as it is. Notice that there is a log with latency for every successful handshake too.

internal/alpini/netty4/alpini-netty4-base/src/main/java/com/linkedin/alpini/netty4/ssl/SslInitializer.java
```diff
@@ -370,6 +379,11 @@ public void channelInactive(ChannelHandlerContext ctx) throws Exception {
       if (!ctx.isRemoved()) {
         next(ctx);
       }
+      if (ctx.channel().hasAttr(SSL_HANDSHAKE_START_TS)) {
+        // If SSL_HANDSHAKE_START_TS is set, then the handshake was started.
+        // If the connection closes mid-handshake, userEventTriggered might never fire.
+        _handshakesFailed.increment();
```
lluwm, line 385: QQ, do we clear this attribute in `userEventTriggered`?
`ctx.channel().attr(SSL_HANDSHAKE_START_TS).getAndSet(null)`

Just a heads-up, Netty [doc](https://netty.io/4.0/api/io/netty/util/Attribute.html#getAndRemove--) seems to say that `getAndSet(null)` might still keep the attribute in the AttribyteMap, so hasAttr(SSL_HANDSHAKE_START_TS) still returns true? Wondering if this could actually be an issue or not. 
> If you only want to return the old value and clear the [Attribute](https://netty.io/4.0/api/io/netty/util/Attribute.html) while still keep it in the [AttributeMap](https://netty.io/4.0/api/io/netty/util/AttributeMap.html) use [getAndSet(Object)](https://netty.io/4.0/api/io/netty/util/Attribute.html#getAndSet-T-) with a value of null.

services/venice-router/src/main/java/com/linkedin/venice/router/stats/SecurityStats.java
```diff
@@ -46,10 +41,10 @@ public void recordNonSslRequest() {
   }
 
   /**
-   * This function will be triggered in every ssl event.
+   * This function will be triggered in {@link com.linkedin.alpini.netty4.handlers.ConnectionControlHandler}.
    */
-  private void recordLiveConnectionCount() {
-    this.sslLiveConnectionCount.record(secureConnectionCountSupplier.getAsInt());
+  public void recordLiveConnectionCount(int connectionCount) {
```
gaojieliu, line 46: The end result might not be as what we expected..
For example, if there is no new connections within 2 time windows, the value of the metric will be `NaN`.
One way is to maintain an integer field here, so that `ConnectionControllerHandler` or `ConnectionLimitHandler` can update it when creating new conns or closing any existing conns, and this method will be triggered by any new request.

That means, when there is no traffic, this metric would report `NaN`, which might be acceptable, or we can do min/max/avg/gauge, so when there are traffic, `min/max/avg` will be more meaningful, and when there is no traffic, `gauge` will give us per-min data point.

internal/alpini/router/alpini-router-impl/src/main/java/com/linkedin/alpini/router/impl/Router.java
```diff
@@ -107,6 +107,8 @@ default Builder timeoutProcessor(@Nonnull TimeoutProcessor timeoutProcessor) {
 
     Builder connectionLimit(@Nonnull IntSupplier connectionLimit);
 
+    Builder connectionCountRecorder(@Nonnull Consumer<Integer> recorder);
```
gaojieliu, line 110: Since would like to keep both stalling and fail-fast connection handlers, shall we add a config option here to control which one to use?

services/venice-server/src/main/java/com/linkedin/venice/listener/ListenerService.java
```diff
@@ -102,7 +102,6 @@ public ListenerService(
           serverConfig.getSslHandshakeThreadPoolSize(),
           "SSLHandShakeThread",
           serverConfig.getSslHandshakeQueueCapacity());
-      new ThreadPoolStats(metricsRepository, this.sslHandshakeExecutor, "ssl_handshake_thread_pool");
```
gaojieliu, line 105: Why this change?

gaojieliu: Also Server is not using the dns offload logic, and I wondered whether it is necessary or not..
Maybe it doesn't since Server is using H2.

internal/alpini/netty4/alpini-netty4-base/src/main/java/com/linkedin/alpini/netty4/handlers/ConnectionControlHandler.java
huangminchn: Yes, since we are aligned with the approach now, I am going to remove all the changes in ConnectionControlHandler.

To more context: ConnectionControlHandler would stall the client request, while ConnectionLimitHander would fail fast. We are going to remove all the changes in ConnectionControlHandler and keep the stalling behavior. We might need it to prevent DDOS attack in future.
For now, by default, Venice will use ConnectionLimitHandler.

internal/venice-client-common/src/main/java/com/linkedin/venice/stats/ThreadPoolStats.java
huangminchn: For sure, picked `queued_task_count`

internal/alpini/netty4/alpini-netty4-base/src/main/java/com/linkedin/alpini/netty4/handlers/ConnectionControlHandler.java
```diff
@@ -32,28 +44,42 @@
...
       parent.eventLoop().submit(() -> {
         if (getConnectedCount() > getConnectionLimit()) {
+          LOGGER.error(
+              "Connection limit exceeded! Active connections: {}, Limit: {}",
+              getConnectedCount(),
+              getConnectionLimit());
           parent.config().setAutoRead(false);
+          Channel clientChannel = ctx.channel(); // Get the current client's channel
+          if (clientChannel.isActive()) {
+            FullHttpResponse response = new DefaultFullHttpResponse(
```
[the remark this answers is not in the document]
huangminchn: Discussed offline; we are going to remove the changes in ConnectionControlHandler, and keep the stalling behavior for this implementation. Venice will switch to use ConnectionLimitHandler by default.

internal/alpini/netty4/alpini-netty4-base/src/main/java/com/linkedin/alpini/netty4/ssl/SslInitializer.java
huangminchn: Yes! Good catch. I believe it would be an issue.
If the channel becomes inactive after SSL handshakes completed exceptionally, we would end up incrementing the failed counter twice.

internal/alpini/netty4/alpini-netty4-base/src/main/java/com/linkedin/alpini/netty4/handlers/ConnectionControlHandler.java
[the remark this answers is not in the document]
huangminchn: Sounds good.

internal/alpini/netty4/alpini-netty4-base/src/main/java/com/linkedin/alpini/netty4/handlers/ConnectionLimitHandler.java
```diff
@@ -75,7 +75,7 @@ public int getConnectionLimit() {
   public void channelActive(ChannelHandlerContext ctx) throws Exception {
     int count = _activeCount.incrementAndGet();
     int limit = _connectionLimit.getAsInt();
-
+    _connectionCountRecorder.accept(count);
     if (count > limit) {
       LOG.debug("Connection count {} exceeds {}", count, limit);
```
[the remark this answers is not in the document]
huangminchn: Right, since we were not using this handler, I kept it as it is.

I think it should be error level, WDYT? Since it indicates a potential capacity issue and client could see major failures.

services/venice-router/src/main/java/com/linkedin/venice/router/stats/SecurityStats.java
huangminchn: Oh, good point. Do you mean the below implementation:
1. Maintain two counters, one for avg and the other for max; ConnectionLimitHander would update these two counters
2. Register two gauge metrics which get metric values from these two counters?

internal/alpini/router/alpini-router-impl/src/main/java/com/linkedin/alpini/router/impl/Router.java
huangminchn: Yes, will do.

services/venice-server/src/main/java/com/linkedin/venice/listener/ListenerService.java
huangminchn: It's not DNS offload, but SSL offload. Server is doing SSL offload properly. You could double check "HttpChannelInitializer#initChannel"

services/venice-router/src/main/java/com/linkedin/venice/router/stats/SecurityStats.java
gaojieliu: Let us sync up offline on ^.

services/venice-server/src/main/java/com/linkedin/venice/listener/ListenerService.java
gaojieliu: Got it, but why removing the stats?

internal/alpini/netty4/alpini-netty4-base/src/main/java/com/linkedin/alpini/netty4/handlers/ConnectionLimitHandler.java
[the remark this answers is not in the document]
huangminchn: Correct.

services/venice-server/src/main/java/com/linkedin/venice/listener/ListenerService.java
huangminchn: It's now initialized in "HttpChannelInitializer"; otherwise, I need to pass the stats from ListenerService into HttpChannelInitializer.

internal/alpini/router/alpini-router-impl/src/main/java/com/linkedin/alpini/router/impl/Router.java
huangminchn: Added new config.

internal/alpini/netty4/alpini-netty4-base/src/main/java/com/linkedin/alpini/netty4/handlers/ConnectionControlHandler.java
```diff
@@ -32,27 +34,31 @@
  */
 @ChannelHandler.Sharable
 public class ConnectionControlHandler extends ConnectionLimitHandler {
+  private static final Logger LOGGER = LogManager.getLogger(ConnectionControlHandler.class);
   private final LongAdder _activeCount = new LongAdder();
   private final LongAdder _inactiveCount = new LongAdder();
```
gaojieliu, line 39: Just a question about why using `LongAddr` here? I think `AtomicInt` or `AtomicLong` is quite convenient and this handler is shared, so at most we only need to create one atomic variable.

Not a blocker.

gaojieliu (approved):
LGTM.

internal/alpini/netty4/alpini-netty4-base/src/main/java/com/linkedin/alpini/netty4/handlers/ConnectionControlHandler.java
huangminchn: These are implementations from Antony, previous Alpini owner. It appears that LongAdder performs better than AtomicLong during high contention on updating the value because LongAdder would maintain multiple counters and only merge then during read.

PR #1590, opened 2025-03-07
[dvc][server][controller][cc][doc][test] DVRT CDC MVP

kvargha:
## [dvc][server][controller][cc][doc][test] DVRT CDC MVP

This PR introduces a new implementation of `BootstrappingVeniceChangelogConsumer` that leverages blob transfer, enabling a CDC client to bootstrap its state from another peer. The Da Vinci Client (DVC) and the Da Vinci Record Transformer (DVRT) are utilized in conjunction to achieve this.

Key changes:
1. Version swap messages (VSM) are now being emitted to the Version Topic (VT). Previously, they were only produced to the Real Time Topic (RT). This was done so DVRT CDC knows when to perform a version swap by consuming the VSM from the future VT. For hybrid stores, the VSM will be produced to the RT by the child controller, then the leader will consume it who will produce it to the VT. For batch stores, the child controller will directly produce the VSM to the VT. To not break the current CDC client, a check was added to ignore VSMs consumed from a VT, as it currently relies on consuming them from the change-capture topic.
2. Modified `ChangelogClientConfig` to make using this new client configurable. For the sake of testing, blob transfer is also configurable. This allows us to compare the performance of when blob transfer is enabled or not. These configs will eventually be removed in the future once DVRT CDC has stabilized and has become the default.
3. Introduced a new configuration parameter, `maxBufferSize`, to `ChangelogClientConfig`. DVRT CDC maintains a message buffer that users need to drain. To prevent an OutOfMemory error, we set a maximum size on this buffer. The default setting might not be suitable for all use cases, so it must be configurable.
4. Made the `DaVinciRecordTransformer` interface partition aware. Additionally, added a new api to `BlockingDaVinciRecordTransformer`, `onVersionSwap`. This will get triggered when a VSM is consumed from the future VT by DVC. These are necessary to know whether a specific partition is ready to serve.
5. Added a new config `skipCompatibilityChecks` to `DaVinciRecordTransformerConfig`. If a user isn't modifying the contents of the records and is making frequent changes to their DVRT implementation, setting this to true will prevent DVRT from wiping local state everytime a change is deployed. DVRT CDC is setting this to true due to these reasons.
6. Updated DVRT docs to make the examples partition aware.
7. Modified the setup in `TestBootstrappingChangelogConsumer` to use a single region as the tests don't depend on it being multi-region. This was done to simplify the setup logic. Additionally, created helper functions to reduce code duplication.
8. Turned off compression in `testRecordTransformerChunking` to ensure chunking threshold is hit.

## How was this PR tested?

The DVRT interface has been modified to become partition aware, so any implementations will need to reflect this.


services/venice-controller/src/main/java/com/linkedin/venice/ingestion/control/RealTimeTopicSwitcher.java
```diff
@@ -222,42 +221,53 @@ public void transmitVersionSwapMessage(Store store, int previousVersion, int nex
...
+       * race-condition inside DaVinci. If we only emit the VSM to the previous version, it's possible that DaVinci
+       * can complete consumption of the next version before the previous can consume the VSM. When this happens,
+       * DaVinci will delete the previous version inside StoreBackend::setDaVinciCurrentVersion. Thus, the previous
+       * version will never be able to consume the VSM.
+       * Because of this, we need to emit the VSM to both the previous and next version.
+       */
+      broadcastVersionSwap(previousStoreVersion, nextStoreVersion, previousStoreVersion.kafkaTopicName());
+      broadcastVersionSwap(previousStoreVersion, nextStoreVersion, nextStoreVersion.kafkaTopicName());
+    } else {
+      broadcastVersionSwap(previousStoreVersion, nextStoreVersion, Utils.getRealTimeTopicName(store));
```
kvargha, line 241: Producing the VSM to the RT causes `PartialUpdateWithParallelProcessingTest::testNonAARewind` to fail. This is because DIV encounters `DuplicateDataException`. Not sure how to address this.

internal/venice-common/src/main/java/com/linkedin/venice/pubsub/PubSubTopicImpl.java
```diff
@@ -11,13 +11,7 @@ public class PubSubTopicImpl implements PubSubTopic {
...
   private final String storeName;
 
-  /**
-   * N.B. leaving this constructor package-private, as the intent is not necessarily the instantiate this object
-   * from anywhere in the code. Most likely, it would be preferable to have some form of singleton topic-registry where
-   * instances can be gotten from, so that the internal state is computed just once, and then reused for the lifetime
-   * of the topic.
-   */
-  PubSubTopicImpl(String name) {
+  public PubSubTopicImpl(String name) {
```
kvargha, line 14: Had to make this public so this object can be created by DVRT CDC

services/venice-controller/src/main/java/com/linkedin/venice/ingestion/control/RealTimeTopicSwitcher.java
ZacAttack: Not sure I do either.  Maybe we can have Lei take a look?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/client/DaVinciRecordTransformer.java
```diff
@@ -142,25 +144,35 @@ public boolean useUniformInputValueSchema() {
     return false;
   }
 
+  /**
+   * Lifecycle event triggered when a version swap is detected for partitionId
+   *
+   * By default, it performs no operation.
+   */
+  public void onVersionSwap(int currentVersion, int futureVersion, int partitionId) {
```
ZacAttack, line 152: This one is kind of interesting and I'm not sure how to feel about it.  I.... think this might pollute the interface a bit.

Thinking aloud about it, how should a normal user of DVRT deal with this kind of event?  I assume this is triggered when we consume the VSM and that makes sense for CDC, but what about for other use cases.  Consuming version swap on v1, doesn't imply that local consumption of v2 has progressed far enough to serve traffic.

kvargha: Right now, the code only invokes this on v2 (future version). It originally did this on v1, but I encountered the issue you mentioned.

As for polluting the interface, we could remove this from the public DVRT interface and only have this method inside DVRT CDC. Then, inside BlockingDVRT we can check if the instance of the DVRT is a change capture one and invoke `onVersionSwap`. I originally wanted to avoid a similar approach, but after thinking about this new method of custom invocation I don't think it's too ugly.

services/venice-controller/src/main/java/com/linkedin/venice/ingestion/control/RealTimeTopicSwitcher.java
```diff
@@ -222,42 +221,63 @@ public void transmitVersionSwapMessage(Store store, int previousVersion, int nex
...
-    if (!hasViewConfigs(nextStoreVersion, previousStoreVersion)) {
-      // NoOp for now
-      return;
+    // If there exists an RT, then broadcast the Version Swap message to it, otherwise broadcast it to the VT
+    String storeName = store.getName();
+    if (!topicManager.containsTopic(pubSubTopicRepository.getTopic(Version.composeRealTimeTopic(storeName)))) {
+      LOGGER.info("RT topic doesn't exist for store: {}. Broadcasting Version Swap message directly to VT.", storeName);
+
+      /*
+       * In a hybrid mode, the VSM gets emitted to the previous and next version topics by the Leader.
```
ZacAttack, line 230: why do we need the VSM in batch mode?

services/venice-controller/src/main/java/com/linkedin/venice/ingestion/control/RealTimeTopicSwitcher.java
```diff
@@ -222,42 +221,63 @@ public void transmitVersionSwapMessage(Store store, int previousVersion, int nex
     Version previousStoreVersion = store.getVersionOrThrow(previousVersion);
     Version nextStoreVersion = store.getVersionOrThrow(nextVersion);
 
-    // Only transmit version swap message to RT's if there is a view config (temporary check)
-    if (!hasViewConfigs(nextStoreVersion, previousStoreVersion)) {
-      // NoOp for now
-      return;
+    // If there exists an RT, then broadcast the Version Swap message to it, otherwise broadcast it to the VT
+    String storeName = store.getName();
+    if (!topicManager.containsTopic(pubSubTopicRepository.getTopic(Version.composeRealTimeTopic(storeName)))) {
```
ZacAttack, line 226: I wonder if we should add a TODO here for broadcasting to batch only view topics?  I don't think we should do that with this PR just to keep things small.  There shouldn't be any user of DVRT/CDC on a view which is batch only, so then the only thing that would remain is if DVC consumption on a view topic needs this message for some reason?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/BootstrappingVeniceChangelogConsumerDaVinciRecordTransformerImpl.java
```diff
@@ -0,0 +1,333 @@
...
+    public void processDelete(Lazy<K> key, int partitionId) {
+      // When value is null, it indicates it's a delete
+      addMessageToBuffer(key.get(), null, partitionId);
+    };
+
+    @Override
+    public void onVersionSwap(int currentVersion, int futureVersion, int partitionId) {
+      /*
+       * Only the futureVersion should act on the version swap message (VSM).
+       * The previousVersion will consume the VSM earlier than the futureVersion, leading to an early version swap.
```
ZacAttack, line 314: This is only true if the previousVersion is caught up right?  Does this change anything?

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/consumer/TestBootstrappingChangelogConsumer.java
```diff
@@ -243,73 +214,228 @@ public void testVeniceChangelogConsumer(int consumerCount) throws Exception {
...
-            polledChangeEventsMap,
-            bootstrappingVeniceChangelogConsumerList.get(0));
-        Assert.assertEquals(polledChangeEventsMap.size(), 0);
-      });
+      verifyNoRecordsProduced(polledChangeEventsMap, polledChangeEventsList, bootstrappingVeniceChangelogConsumerList);
 
-      VeniceChangelogConsumer<Utf8, Utf8> afterImageChangelogConsumer =
-          veniceChangelogConsumerClientFactory.getChangelogConsumer(storeName);
-      afterImageChangelogConsumer.subscribe(new HashSet<>(Arrays.asList(0, 1, 2))).get();
+      // Create new version
```
ZacAttack, line 314: this is good, but can we add a case for repush?  I've been finding repush can do wonky things for CDC because it reorders events.  You may find bugs, but we can discuss them because I believe the current implementation has scenarios where it doesn't handle repush gracefully.  Let's at least start by seeing what shakes out and add the test code for it.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/ChangelogClientConfig.java
```diff
@@ -251,4 +258,46 @@ public ChangelogClientConfig setIsBeforeImageView(Boolean beforeImageView) {
...
+    return maxBufferSize;
+  }
+
+  /**
+   * This is used by the experimental client to determine the max buffer size that's returned to the user when
+   * calling poll. When the max size is reached, ingestion will be paused, until the buffer is drained.
+   * In order for this feature to be used, {@link #setIsExperimentalClientEnabled(Boolean)} must be set to true.
+   * It is currently only supported for {@link BootstrappingVeniceChangelogConsumer}.
+   */
+  public ChangelogClientConfig setMaxBufferSize(int maxBufferSize) {
```
ZacAttack, line 299: We should probably mention here that this buffer size is in addition to the buffer allocated to any pubsub clients under the hood configured via consumerProperties.

ZacAttack: Also that the unit is in elements I think, not memory size.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/BootstrappingVeniceChangelogConsumerDaVinciRecordTransformerImpl.java
```diff
@@ -0,0 +1,333 @@
...
+        .put(ROCKSDB_BLOCK_CACHE_SIZE_IN_BYTES, changelogClientConfig.getRocksDBBlockCacheSizeInBytes())
+        .put(changelogClientConfig.getConsumerProperties())
+        .put(DATA_BASE_PATH, changelogClientConfig.getBootstrapFileSystemPath())
+        .put(ROCKSDB_PLAIN_TABLE_FORMAT_ENABLED, false)
+        .put(PERSISTENCE_TYPE, ROCKS_DB)
+        // Turning this off, so users don't subscribe to unwanted partitions automatically
+        .put(DA_VINCI_SUBSCRIBE_ON_DISK_PARTITIONS_AUTOMATICALLY, false)
+        .put(BLOB_TRANSFER_MANAGER_ENABLED, changelogClientConfig.isBlobTransferEnabled())
+        .put(CLIENT_USE_SYSTEM_STORE_REPOSITORY, true)
+        .put(PUSH_STATUS_STORE_ENABLED, true)
```
ZacAttack, line 225: I 'think' we should also set PARTICIPANT_MESSAGE_STORE_ENABLED to false.  Since this CDC user is going to be throttling ingestion based on the user calls to poll, it may block version pushes as it's not ready to serve based on customer code.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/BootstrappingVeniceChangelogConsumerDaVinciRecordTransformerImpl.java
```diff
@@ -0,0 +1,333 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.CountDownLatch;
+import org.apache.avro.Schema;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+
+@Experimental
+public class BootstrappingVeniceChangelogConsumerDaVinciRecordTransformerImpl<K, V>
+    extends VeniceAfterImageConsumerImpl<K, V> implements BootstrappingVeniceChangelogConsumer<K, V> {
```
ZacAttack, line 60: This is kind of interesting.  Why do we need to extend the afterimage consumer?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/BootstrappingVeniceChangelogConsumerDaVinciRecordTransformerImpl.java
```diff
@@ -0,0 +1,333 @@
...
+     * will wait forever for capacity to become available. This leads to a deadlock.
+     */
+    daVinciClient.subscribe(partitions).whenComplete((result, error) -> {
+      if (error != null) {
+        LOGGER.error("Failed to subscribe to partitions: {} for store: {}", partitions, storeName, error);
+        throw new VeniceException(error);
+      }
+    });
+
+    return CompletableFuture.supplyAsync(() -> {
```
ZacAttack, line 134: Based on recently learned lessons, let's pass this an executor.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/BootstrappingVeniceChangelogConsumerDaVinciRecordTransformerImpl.java
```diff
@@ -0,0 +1,333 @@
...
+  @Override
+  public void stop() throws Exception {
+    daVinciClientFactory.close();
+    isStarted = false;
+  }
+
+  @Override
+  public Collection<PubSubMessage<K, ChangeEvent<V>, VeniceChangeCoordinate>> poll(long timeoutInMs) {
+    try {
+      Thread.sleep(timeoutInMs);
```
ZacAttack, line 172: Why are we sleeping here?

kvargha: This is to respect the timeout passed in by the user, and to give some time for messages to make it into the buffer. If we don't sleep and the user is calling this in an infinite loop, they may get a lot of empty responses and have poll acquire the lock on the buffer more often.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/ChangelogClientConfig.java
kvargha:
```  /**
   * This config is used to control the maximum records returned by every poll request.
   * So far, Store Ingestion is throttling per poll, so if the configured value is too big,
   * the throttling could be inaccurate and it may impact GC as well.
   *
   * We should try to avoid too many long-lasting objects in JVM to minimize GC overhead.
   */
  public static final String SERVER_KAFKA_MAX_POLL_RECORDS = "server.kafka.max.poll.records";
```

Is this the config you're referring to?

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/consumer/TestBootstrappingChangelogConsumer.java
kvargha: I added a test case for repush and I don't see anything abnormal. What are some of the issues the current CDC implementation sees?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/BootstrappingVeniceChangelogConsumerDaVinciRecordTransformerImpl.java
kvargha: TLDR: Yes there is. A user may not call poll fast enough and blocks ingestion of the current version. This can allow the future version to get ahead.

kvargha: Added a todo for this.

ZacAttack (approved):
I think this is ready.  Great work!  We have a short list of TODO's yet, but we can iterate.

ZacAttack (approved):

PR #1592, opened 2025-03-08
[controller] check rt topic existence before trying to truncate it

arjun4084346:
## Check existence of RT topic before trying to truncate it to avoid timing out the `deleteStore` operation
Some implementations of PubSubAdminAdapter may retry multiple times to get/update topic's config in order to truncate it. Because of this, `truncateKafkaTopic` may take long enough time to make `delete store` operation time out. Therefore, we need to check for the existence of RT topic before trying to truncate it. No known implementation of PubSubAdminAdapter does retries for `containsTopic`, so that should not increase the risk of timing out.

Resolves #XXX


sushantmane (approved):
Great find. Thanks, @arjun4084346! 🚀

PR #1594, opened 2025-03-10
[dvc] Do not unsubscribe store from store repository during Da Vinci shutdown

sixpluszero:
## [dvc] Do not unsubscribe store from store repository during Da Vinci shutdown
Da Vinci client throws a lot of exception during shutdown due to the fact that it is unsubscribing to store repo while doing other checkpointing in async way. They will throw exception when store is not found from the repo.
There are two ways to fix it: (1) Just don't unsubscribe it from store repository, as the subscription is fixed during client life cycle and there is no harm to keep the existing subscription during shutdown (2) Make all the shutdown work first and then perform unsubscribe, which will have more changes and potentially slow down. This PR goes with (1) as it is simpler and cleaner to achieve.

## How was this PR tested?

Existing test

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/StoreBackend.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/StoreBackend.java
@@ -90,6 +90,7 @@ public class StoreBackend {
       version.delete();
     }
 
+    backend.getStoreRepository().unsubscribe(storeName);
   }
 
   public boolean isManaged() {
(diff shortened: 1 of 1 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/StoreBackend.java
```diff
@@ -92,7 +90,6 @@ synchronized void delete() {
       version.delete();
     }
 
-    backend.getStoreRepository().unsubscribe(storeName);
```
gaojieliu, line 95: Why do we want to remove this?
Keeping the subscription only will let DaVinci application keep polling the backend to retrieve store metadata, which is not good for a deleted store.

sixpluszero: Yeah I think this makes sense. We can keep this one and only remove the above one, during the regular close.

gaojieliu (approved):
Thanks!

PR #1599, opened 2025-03-11
[server][controller][dvc] Cache VeniceSystemStoreType for faster lookup

sushantmane:
## Cache VeniceSystemStoreType for faster lookup
CPU profiles for venice-server showed that approximately 5% of execution time was spent  
in VeniceSystemStoreType::getSystemStoreType(). This change improves performance by  
introducing a cache to store computed store types for quick retrieval, eliminating redundant  
iterations over all system store types in repeated lookups.  

image

## How was this PR tested?
UT


ZacAttack (approved):
Heck yes

PR #1605, opened 2025-03-12
do not try to get mr job counters when running in local mode in tests

arjun4084346:

Resolves #XXX


clients/venice-push-job/src/main/java/com/linkedin/venice/hadoop/mapreduce/datawriter/reduce/VeniceReducer.java
```diff
@@ -134,23 +137,21 @@ private long getTotalIncomingDataSizeInBytes(JobConf jobConfig) {
...
     } catch (Exception e) {
-      /**
-       * Note that this will catch a NPE during tests unless the storage quota is set to
-       * {@link Store.UNLIMITED_STORAGE_QUOTA}, which seems quite messed up. It manifests as {@link runningJob}
-       * being null, which then prevents us from calling {@link RunningJob#getCounters()}. Obviously, this is
-       * not happening in prod, though it's not completely clear why...
-       *
-       * TODO: Fix this so that tests are more representative of prod
-       */
+      if (jobConfig.get("mapreduce.framework.name").equals("local")) {
```
sushantmane, line 149: Can we replace these two literals with constants `MRConfig.FRAMEWORK_NAME`, `MRConfig.LOCAL_FRAMEWORK_NAME`?  Also is `0` correct return value?

PR #1608, opened 2025-03-12
[controller] remove unnecessary code

arjun4084346:
## removing some unnecessary code
1) `createNewVersion` flag is always `false` in the code, so removing it
2) `peekNextVersion` in Admin has no caller
3) All callers of `Store.peekNextVersion()` want version number, not version

Resolves #XXX


lluwm (approved):
Looks good to me.

PR #1611, opened 2025-03-13
[controller] Set target regions to null in update store if not present

misyel:
## Summary

Currently, in updateStore, targetSwapRegions is always set to "" if it is not present. The type of it is a nullable string so it's default value is expected to be null, not an an empty string, and this fails with the new semantic usage detector. This pr changes it to not set the targetSwapRegion value if it is not present and default to null instead

## How was this PR tested?

Updated unit test

### DIFF
--- a/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
+++ b/services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
@@ -2683,11 +2683,8 @@ public class VeniceParentHelixAdmin implements Admin {
           .map(addToUpdatedConfigList(updatedConfigsList, NEARLINE_PRODUCER_COUNT_PER_WRITER))
           .orElseGet(currStore::getNearlineProducerCountPerWriter);
 
-      if (params.getTargetSwapRegion().isPresent()) {
-        setStore.targetSwapRegion = params.getTargetSwapRegion()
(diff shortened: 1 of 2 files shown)


services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceParentHelixAdmin.java
```diff
@@ -2683,9 +2683,11 @@ public void updateStore(String clusterName, String storeName, UpdateStoreQueryPa
           .map(addToUpdatedConfigList(updatedConfigsList, NEARLINE_PRODUCER_COUNT_PER_WRITER))
           .orElseGet(currStore::getNearlineProducerCountPerWriter);
 
-      setStore.targetSwapRegion = params.getTargetSwapRegion()
-          .map(addToUpdatedConfigList(updatedConfigsList, TARGET_SWAP_REGION))
-          .orElseGet(currStore::getTargetSwapRegion);
+      if (params.getTargetSwapRegion().isPresent()) {
+        setStore.targetSwapRegion = params.getTargetSwapRegion()
+            .map(addToUpdatedConfigList(updatedConfigsList, TARGET_SWAP_REGION))
+            .orElseGet(currStore::getTargetSwapRegion);
```
majisourav99, line 2689: could replace orElseGet to `orElse(null)`

majisourav99 (approved):
Thanks for the quick fix

PR #1337, opened 2024-11-22
[controller][server][da-vinci-client] Mark a helix replica to error state during ingestion error

majisourav99:
## Mark a helix replica to error state during ingestion error

If ingestion task encounters any error, it just marks the CV to error state which never remediates. This PR makes user of a new [API](https://github.com/apache/helix/pull/2792) helix added to annotate a replica to ERROR state. This will later be picked by the error replica reset task `ErrorPartitionResetTask`  which will attempt to recover those replicas. 
Also added system stores to the controller error replica reset task.

## How was this PR tested?

GHCI
## Does this PR introduce any user-facing changes?

- [x] No. You can skip the rest of this section.
- [ ] Yes. Make sure to explain your proposed changes and call out the behavior change.

PR #1613, opened 2025-03-14
[server] Fix multi-threading bug in shared multi-region PubSub configs

nisargthakkar:
## Problem Statement

In production, we found that servers had been querying PubSub offsets from an completely unintended region in some cases. This is pretty gnarly and can lead to data-loss.

## Solution

We found a concurrency issue in `KafkaStoreIngestionService` where the shared server config object would be replaced by temporarily created configs for different regions. The expectation of this code was to only create this object for local use.

###  Code changes
  - [X] Confirmed if logs need to be **rate limited** to avoid excessive logging.

- [X] Code has **no race conditions** or **thread safety issues**.
- [X] Proper **synchronization mechanisms** (e.g., `synchronized`, `RWLock`) are used where needed.
- [X] No **blocking calls** inside critical sections that could lead to deadlocks or performance degradation.
- [X] Verified **thread-safe collections** are used (e.g., `ConcurrentHashMap`, `CopyOnWriteArrayList`).
- [X] Validated proper exception handling in multi-threaded code to avoid silent thread termination.

## How was this PR tested?

- [ ] New unit tests added.

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaStoreIngestionService.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaStoreIngestionService.java
@@ -1177,28 +1177,28 @@ public class KafkaStoreIngestionService extends AbstractVeniceService implements
   }
 
   private VeniceProperties getPubSubSSLPropertiesFromServerConfig(String kafkaBootstrapUrls) {
-    final VeniceServerConfig serverConfigForCluster;
+    final VeniceServerConfig serverConfigForPubSubCluster;
     if (kafkaBootstrapUrls.equals(serverConfig.getKafkaBootstrapServers())) {
-      serverConfigForCluster = serverConfig;
+      serverConfigForPubSubCluster = serverConfig;
     } else {
       Properties clonedProperties = serverConfig.getClusterProperties().toProperties();
       clonedProperties.setProperty(KAFKA_BOOTSTRAP_SERVERS, kafkaBootstrapUrls);
-      serverConfigForCluster =
+      serverConfigForPubSubCluster =
           new VeniceServerConfig(new VeniceProperties(clonedProperties), serverConfig.getKafkaClusterMap());
     }
 
-    VeniceProperties clusterProperties = serverConfigForCluster.getClusterProperties();
-    Properties properties = serverConfigForCluster.getClusterProperties().getPropertiesCopy();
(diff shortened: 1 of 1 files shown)


ZacAttack (approved):

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaStoreIngestionService.java
```diff
@@ -1177,23 +1177,28 @@ private Properties getCommonKafkaConsumerProperties(VeniceServerConfig serverCon
...
   private VeniceProperties getPubSubSSLPropertiesFromServerConfig(String kafkaBootstrapUrls) {
-    if (!kafkaBootstrapUrls.equals(serverConfig.getKafkaBootstrapServers())) {
+    final VeniceServerConfig serverConfigForPubSubCluster;
+    if (kafkaBootstrapUrls.equals(serverConfig.getKafkaBootstrapServers())) {
+      serverConfigForPubSubCluster = serverConfig;
+    } else {
       Properties clonedProperties = serverConfig.getClusterProperties().toProperties();
       clonedProperties.setProperty(KAFKA_BOOTSTRAP_SERVERS, kafkaBootstrapUrls);
-      serverConfig = new VeniceServerConfig(new VeniceProperties(clonedProperties), serverConfig.getKafkaClusterMap());
+      serverConfigForPubSubCluster =
```
lusong64, line 1186: is the serverConfigForPubSubCluster completely a deep copy of the serverConfig? Or we don't care as long as the reference of serverConfig is immutable now? In other words, is sharing some properties between serverConfigForPubSubCluster and serverConfig OK?

nisargthakkar: I don't expect there to be any sharing now that it is local. The reasoning is that:
1. the constructor args are a `new VeniceProperties` which is essentially an immutable `Map<String, String>` with some helper functions
2. `clonedProperties` is created through `.toProperties()` which is also a deep copy of a `Map<String, String>`
3. And we only reuse `kafkaBootstrapUrls` which is also an immutable object (Strings are immutable in Java)

That being said, I do not like the code pattern of creating a `new ServerConfig`, but this code is essentially a partial revert from a previous change (#1337) with variable name changed for readability, and making things final so this doesn't reoccur

lusong64: Cool. as long as there is a deep copy.

PR #1555, opened 2025-02-26
[controller] versioning support for RT topics

arjun4084346:
## add versioning in RT topics

This PR will make RT topic to have version numbers in their name. It will help RT topic names to be unique when older RT is not compatible with the new RT.
Largest used RT version number will be saved in store config and will remain with it in the graveyard, if store is deleted.
When a new store is created with the same name, we will use the `largestUsedRTVersionNumber` to name the first RT in the recreated store, this way this RT will have different name from the RT of the deleted store. This will also avoid the problem when store recreation fails because the RT topic of the previous store is not fully cleaned up.
`largestUsedRTVersionNumber` will be increased in two cases :
- when store is recreated
- when partition count of the RT topic is changed and `controller.enable.hybrid.store.partition.count.update` is set (see PR#1404)

RT topics of SystemStores, UserSystemStores, ParticipationStores do not follow this versioning. Regular RT topic of stores and separate RT topics follow this versioning.

RT topic name is stored in store-version hybrid configs. If it is not found there, which would be the case for the existing stores, old RT naming convention is used. This way, old stores will work with the new code. Test `testOldStoresWithHybridStoreVersioning` verifies this.

## How was this PR tested?

updated test `testHybridStoreRepartitioning`, `testOldStoresWithHybridStoreVersioning`

## Does this PR introduce any user-facing changes?

- [x] No. You can skip the rest of this section.
- [ ] Yes. Make sure to explain your proposed changes and call out the behavior change.

PR #1616, opened 2025-03-15
[compat] [controller] add largestUsedRTVersionNumber to updateStore and addVersion admin commands

arjun4084346:
## Solution

Adding a field for `largestUsedRTVersionNumber` in `addVersion` and `updateStore` operations.
- It is added in `addVersion` so that parent controller will pass this value to child controllers and we do not end up different child controllers having different value for this field.
- It is added in `updateStore` for two reasons - a) write test case `testOldStoresWithHybridStoreVersioning` of future PR https://github.com/linkedin/venice/pull/1555/  , I am verifying the effect of that work on existing stores that do not have `largestUsedRTVersionNumber` value (backward compatibility testing). So I have to update this value from 1 to 0 to make a test store act like an existing store in prod. b) disaster recovery , if we create a wrong RT name at any component of Venice, we can update `largestUsedRTVersionNumber` to 0 and `realTimeTopicName` to "" (empty) and then, that will behave like the old store, without RT versioning.

schema diff is  this -

```abora-mn3:venice abora$  diff services/venice-controller/src/main/resources/avro/AdminOperation/v84/AdminOperation.avsc services/venice-controller/src/main/resources/avro/AdminOperation/v85/AdminOperation.avsc 
482a483,487
>             {
>               "name": "largestUsedRTVersionNumber",
>               "doc": "largestUsedRTVersionNumber that should be used to formulate real time topic name during add version",
>               "type": ["null", "int"],
>               "default": null
>             },
825a832,837
>             {
>               "name": "largestUsedRTVersionNumber",
>               "type": "int",
>               "doc": "largestUsedRTVersionNumber that should be used to formulate real time topic name during add version",
>               "default": 0
>             },
```

###  Code changes

## How was this PR tested?

- [ ] New unit tests added.

### DIFF
--- a/services/venice-controller/src/main/resources/avro/AdminOperation/v85/AdminOperation.avsc
+++ b/services/venice-controller/src/main/resources/avro/AdminOperation/v85/AdminOperation.avsc
@@ -486,7 +486,7 @@
             {
               "name": "largestUsedRTVersionNumber",
               "type": ["null", "int"],
-              "doc": "largestUsedRTVersionNumber that should be used to formulate real time topic name during add version",
+              "doc": "Largest used RT version number used by this store. This is used to create real time topic name while creating a new store-version",
               "default": null
             },
             {
--- a/services/venice-controller/src/main/resources/avro/AdminOperation/v85/AdminOperation.avsc
+++ b/services/venice-controller/src/main/resources/avro/AdminOperation/v85/AdminOperation.avsc
@@ -1,0 +1,1177 @@
+{
+  "name": "AdminOperation",
+  "namespace": "com.linkedin.venice.controller.kafka.protocol.admin",
+  "type": "record",
+  "fields": [
+    {
+      "name": "operationType",
+      "doc": "0 => StoreCreation, 1 => ValueSchemaCreation, 2 => PauseStore, 3 => ResumeStore, 4 => KillOfflinePushJob, 5 => DisableStoreRead, 6 => EnableStoreRead, 7=> DeleteAllVersions, 8=> SetStoreOwner, 9=> SetStorePartitionCount, 10=> SetStoreCurrentVersion, 11=> UpdateStore, 12=> DeleteStore, 13=> DeleteOldVersion, 14=> MigrateStore, 15=> AbortMigration, 16=>AddVersion, 17=> DerivedSchemaCreation, 18=>SupersetSchemaCreation, 19=>EnableNativeReplicationForCluster, 20=>MetadataSchemaCreation, 21=>EnableActiveActiveReplicationForCluster, 25=>CreatePersona, 26=>DeletePersona, 27=>UpdatePersona, 28=>RollbackCurrentVersion, 29=>RollforwardCurrentVersion",
+      "type": "int"
+    }, {
+      "name": "executionId",
+      "doc": "ID of a command execution which is used to query the status of this command.",
+      "type": "long",
+      "default": 0
+    }, {
+      "name": "payloadUnion",
+      "doc": "This contains the main payload of the admin operation",
+      "type": [
+        {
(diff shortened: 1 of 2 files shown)


services/venice-controller/src/main/resources/avro/AdminOperation/v85/AdminOperation.avsc
```diff
@@ -0,0 +1,1177 @@
...
+              "default": null
+            },
+            {
+              "name": "repushSourceVersion",
+              "doc": "Indicates the source version from which a repush version is created",
+              "type": "int",
+              "default": -1
+            },
+            {
+              "name": "largestUsedRTVersionNumber",
```
nisargthakkar, line 834: Does the version config need to be named `largestUsedRTVersionNumber`? Can it be named something like `currentRTVersionNumber` or `rtVersionNumber`?

nisargthakkar: It makes for the store-level config to be called `largestUsedRTVersionNumber`.

But version configs are immutable. So, whatever goes in there stays there for the lifetime of the version. So, it is better to have a more representative name in version configs in such cases because it will always be associated with only that version

arjun4084346: ok, that makes sense, i will rename it. but note that we are not storing this `largestUsedRTVersionNumber/currentRTVersionNumber` in version config, just passing it in `addVersion` admin command.

services/venice-controller/src/main/resources/avro/AdminOperation/v85/AdminOperation.avsc
```diff
@@ -0,0 +1,1177 @@
...
+            },
+            {
+              "name": "largestUsedVersionNumber",
+              "type": ["null", "int"],
+              "default": null
+            },
+            {
+              "name": "largestUsedRTVersionNumber",
+              "type": ["null", "int"],
+              "doc": "largestUsedRTVersionNumber that should be used to formulate real time topic name during add version",
```
nisargthakkar, line 489: Sorry for the late comment, but can we make this more human-readable?

services/venice-controller/src/main/resources/avro/AdminOperation/v85/AdminOperation.avsc
nisargthakkar: Got it. Thanks for making the change

services/venice-controller/src/main/resources/avro/AdminOperation/v85/AdminOperation.avsc
nisargthakkar: (referring to the doc)

nisargthakkar (approved):
LGTM. Thanks

services/venice-controller/src/main/resources/avro/AdminOperation/v85/AdminOperation.avsc
```diff
@@ -0,0 +1,1177 @@
...
+            },
+            {
+              "name": "repushSourceVersion",
+              "doc": "Indicates the source version from which a repush version is created",
+              "type": "int",
+              "default": -1
+            },
+            {
+              "name": "currentRTVersionNumber",
+              "type": "int",
```
nisargthakkar, line 835: What will the value be for a batch-only store?

arjun4084346: for new stores, it will start from 1. however, it will not be used, because the only use for this is to create rt name.
existing stores will have the default value - 0. and the value of this field (0 or 1+) will determine if a store predates rt-versioning changes

PR #1617, opened 2025-03-16
[build] Make line check run only on main branch

majisourav99:
## Problem Statement

PR commit change check should run only against main branch, otherwise hotfix releases might fail.

## Solution

This PR ignore the workflow for non-main branch.

###  Code changes

## How was this PR tested?

- [ ] New unit tests added.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -18,6 +18,8 @@ Describe
 - What issues or limitations exist in the current code
 - Why this change is necessary 
 -->
+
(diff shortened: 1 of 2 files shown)


nisargthakkar (commented):
Change looks good, but could you please update the PR title? Seems a bit too vague right now. Also, could you add the `[build]` tag?

nisargthakkar (approved):
LGTM. Thanks!

PR #1618, opened 2025-03-16
[compat][vpj] Cleanup extra MR job to validate schemas and build dictionary

nisargthakkar:
## Problem Statement

VPJ has supported Spark as a compute engine for a while now, and it is considered STABLE. However, unit tests still use only MapReduce as the compute engine.

## Solution

Before adding Spark in unit tests, this PR cleans up the dedicated MR job to validate schemas and build dictionary.

The primary motivation to add this additional MR job was that Zstd library would cause ungraceful crashes of the JVM. These errors have now been fixed, and we have experience with running Zstd dict building in the driver itself.

###  Code changes
  - [X] Confirmed if logs need to be **rate limited** to avoid excessive logging.

## How was this PR tested?

Regression tests will be run in GH CI

- [ ] New unit tests added.
- [X] Modified or extended existing tests.
- [X] Verified backward compatibility (if applicable).

- [X] Yes. Clearly explain the behavior change and its impact.
    - Certain configs related to "ValidateSchemaAndBuildDict" MR job will no longer be honored.


m-nagarajan (approved):
Thanks @nisargthakkar 🚀

PR #1619, opened 2025-03-16
[changelog][client] Clean up cdc locking

ZacAttack:
This is mostly a clean up PR, though there may be some bugs getting fixed here.  The locking structure of the ChangelogConsumer was a little too convoluted, and didn't seem to offer adequate safety for some scenarios.  For example, some users have reported seeking exceptions when polling when subscriptions change on the client (as there is a period of time where the consumer would have no subscriptions before getting new subscriptions).

This accomplishes the above with three main refactors.  They are:

1.  **Use pubsubAdapter directly instead of a seperate CDC client** 
When doing seek to EOP The implementation was just directly accessing the pubsubAdapter in the wrapped CDC client anyway, and the CDC client is significantly heavier.  So we swap that type out.

2. **Unify all locking with a ReentrentReadWriteLock** 
The previous implementation made liberal use of synchronized() on the pubsubadapter as well as a few other objects.  The downside of doing this was that it could it sieze up client calls to poll() unnecessarily.  Using a proper readwritelock should allow for better concurrency and concurrency that makes more sense.

3. **Add sync api's to reduce deadlock risk** 
So, an observed problem was that if you go N degrees deep on async functions, it's easy to lose track of who holds what lock.  It was actually pretty easy to get into scenarios where a main function call would syncrhonize on an object, and then call a function which would spawn an async thread which would need to lock the same object in order to do any work.  To solve this, we introduce syncApi's which can be called directly and avoid introducing extra locking threads.

###  Code changes

## How was this PR tested?

- [ ] New unit tests added.

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceAfterImageConsumerImpl.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceAfterImageConsumerImpl.java
@@ -1,9 +1,12 @@
 package com.linkedin.davinci.consumer;
 
+import static com.linkedin.davinci.consumer.VeniceChangelogConsumerClientFactory.*;
+
 import com.linkedin.davinci.repository.NativeMetadataRepositoryViewAdapter;
 import com.linkedin.venice.exceptions.VeniceException;
 import com.linkedin.venice.kafka.protocol.ControlMessage;
 import com.linkedin.venice.kafka.protocol.enums.ControlMessageType;
+import com.linkedin.venice.offsets.OffsetRecord;
 import com.linkedin.venice.pubsub.api.DefaultPubSubMessage;
 import com.linkedin.venice.pubsub.api.PubSubConsumerAdapter;
 import com.linkedin.venice.pubsub.api.PubSubMessage;
@@ -11,13 +14,13 @@ import com.linkedin.venice.pubsub.api.PubSubTopic;
 import com.linkedin.venice.pubsub.api.PubSubTopicPartition;
 import com.linkedin.venice.utils.lazy.Lazy;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.ExecutionException;
 import java.util.concurrent.atomic.AtomicBoolean;
 import org.apache.kafka.common.errors.UnknownTopicOrPartitionException;
 import org.apache.logging.log4j.LogManager;
@@ -30,7 +33,7 @@ public class VeniceAfterImageConsumerImpl<K, V> extends VeniceChangelogConsumerI
   // in the context of seeking to EOP in the event of the user calling that seek or a version push.
   // TODO: We shouldn't use this in the long run. Once the EOP position is queryable from venice and version
   // swap is produced to VT, then we should remove this as it's no longer needed.
-  final private Lazy<VeniceChangelogConsumerImpl<K, V>> internalSeekConsumer;
+  final private Lazy<PubSubConsumerAdapter> internalSeekConsumer;
   AtomicBoolean versionSwapThreadScheduled = new AtomicBoolean(false);
   private final VersionSwapDataChangeListener<K, V> versionSwapListener;
 
@@ -39,17 +42,15 @@ public class VeniceAfterImageConsumerImpl<K, V> extends VeniceChangelogConsumerI
         changelogClientConfig,
         consumer,
         Lazy.of(
-            () -> new VeniceChangelogConsumerImpl<K, V>(
-                changelogClientConfig,
-                VeniceChangelogConsumerClientFactory.getConsumer(
-                    changelogClientConfig.getConsumerProperties(),
-                    changelogClientConfig.getStoreName() + "-" + "internal"))));
+            () -> getConsumer(
+                changelogClientConfig.getConsumerProperties(),
+                changelogClientConfig.getStoreName() + "-" + "internal")));
   }
 
   protected VeniceAfterImageConsumerImpl(
       ChangelogClientConfig changelogClientConfig,
       PubSubConsumerAdapter consumer,
-      Lazy<VeniceChangelogConsumerImpl<K, V>> seekConsumer) {
+      Lazy<PubSubConsumerAdapter> seekConsumer) {
     super(changelogClientConfig, consumer);
     internalSeekConsumer = seekConsumer;
     versionSwapListener = new VersionSwapDataChangeListener<K, V>(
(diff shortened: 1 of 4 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumerImpl.java
```diff
@@ -388,7 +395,9 @@ public void pause() {
 
   @Override
   public void resume(Set<Integer> partitions) {
-    synchronized (pubSubConsumer) {
+    // We're not changing the subscription set, so we can just lock the read lock
+    subscriptionLock.readLock().lock();
+    try {
       Set<PubSubTopicPartition> currentSubscriptions = getTopicAssignment();
       synchronized (currentSubscriptions) {
```
kvargha, line 402: Does this need to be synchronized if we using a read lock?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumerImpl.java
```diff
@@ -411,7 +422,8 @@ public void resume() {
 
   @Override
   public void pause(Set<Integer> partitions) {
-    synchronized (pubSubConsumer) {
+    subscriptionLock.readLock().lock();
+    try {
       Set<PubSubTopicPartition> currentSubscriptions = getTopicAssignment();
       synchronized (currentSubscriptions) {
```
kvargha, line 428: Same here

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumerImpl.java
```diff
@@ -468,19 +482,23 @@ public CompletableFuture<Void> seekToTail() {
...
 
+  protected void synchronousSeekToCheckpoint(Set<VeniceChangeCoordinate> checkpoints) {
+    for (VeniceChangeCoordinate coordinate: checkpoints) {
+      checkLiveVersion(coordinate.getTopic());
+      PubSubTopic topic = pubSubTopicRepository.getTopic(coordinate.getTopic());
+      PubSubTopicPartition pubSubTopicPartition = new PubSubTopicPartitionImpl(topic, coordinate.getPartition());
+      internalSeek(Collections.singleton(coordinate.getPartition()), topic, foo -> {
+        Long topicOffset = coordinate.getPosition().getNumericOffset();
+        pubSubConsumerSeek(pubSubTopicPartition, topicOffset);
+      });
```
kvargha, line 498: The previous implementation had a join. Do we need that here too?

kvargha (changes requested):

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumerImpl.java
ZacAttack: No.  Actually the intention is to get rid of the join here as the join.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceAfterImageConsumerImpl.java
```diff
@@ -108,22 +109,32 @@ protected CompletableFuture<Void> internalSeekToEndOfPush(Set<Integer> partition
...
+        // filter out from the user
+        synchronized (internalSeekConsumer) {
+          PubSubConsumerAdapter consumerAdapter = internalSeekConsumer.get();
+          consumerAdapter.batchUnsubscribe(consumerAdapter.getAssignment());
+          List<PubSubTopicPartition> topicPartitionList =
+              getPartitionListToSubscribe(partitions, Collections.EMPTY_SET, targetTopic);
+
+          for (PubSubTopicPartition topicPartition: topicPartitionList) {
+            consumerAdapter.subscribe(topicPartition, OffsetRecord.LOWEST_OFFSET);
+          }
```
kvargha, line 135: We're modifying the subscriptions here. Don't we need a write lock instead of a read lock?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceAfterImageConsumerImpl.java
```diff
@@ -132,53 +143,48 @@ protected CompletableFuture<Void> internalSeekToEndOfPush(Set<Integer> partition
...
+                    LOGGER.info("Found EOP message for partition: " + pubSubTopicPartition.getPartitionNumber());
+                    // note down the partition and offset and mark that we've got the thing
+                    endOfPushConsumedPerPartitionMap.put(pubSubTopicPartition.getPartitionNumber(), true);
+                    VeniceChangeCoordinate coordinate = new VeniceChangeCoordinate(
+                        pubSubTopicPartition.getPubSubTopic().getName(),
+                        message.getPosition(),
+                        pubSubTopicPartition.getPartitionNumber());
+                    checkpoints.add(coordinate);
+                    // No need to look at the rest of the messages for this partition that we might have polled
+                    consumerAdapter.unSubscribe(pubSubTopicPartition);
```
kvargha, line 166: Same thing here

kvargha (changes requested):

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceChangelogConsumerImpl.java
ZacAttack: Actually, you are right, this should use synchronousSeek, good catch!

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceAfterImageConsumerImpl.java
ZacAttack: Great question!  So, here's the rub.  We're changing the subscription on the seekConsumer which we get a seperate lock on.  In this path, I want to prevent OTHER writers from changing the subcription BUT I want user poll requests to go through.  Write is an exclusive lock, so by getting a read lock initially, I prevent other writers getting in.  Once I've gone and done the long work of finding the EOP message, I then call synchronous seek.  The purpose of doing this is to keep the context in the current thread and then grab the writeLock to update the main consumer's subscriptions.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/consumer/VeniceAfterImageConsumerImpl.java
ZacAttack: see above.  We lock on both the internalConsumer and the read lock.

kvargha (approved):

kvargha (approved):

PR #1620, opened 2025-03-17
[router] tightening store version readiness checks for dictionary decompressor readiness

WhitneyDeng:
## Problem Statement
For full problem scope, see [Design Doc::Problem Statement](https://docs.google.com/document/d/1HHK_v4OAG4Wz-mgtPdTMijNiece5JoJTz0EONXadZt4/edit?tab=t.0#heading=h.vady9t8qn7xg)
Initial problem scope: [Bug Basher Ticket](https://jira01.corp.linkedin.com:8443/browse/VENG-11843)

## Solution
For full solution w/ pseudocode and tabular presentation of behaviour changes, see [Design Doc::Solution](https://docs.google.com/document/d/1HHK_v4OAG4Wz-mgtPdTMijNiece5JoJTz0EONXadZt4/edit?tab=t.0#heading=h.1q8lj5wqcnbl)
Initial solution scope: continue work of [this PR](https://github.com/linkedin/venice/pull/652)

###  Code changes
- [x] Introduced new **log lines**. 

## How was this PR tested?
For full list of tests added/modified/extended with commit links, see [Design Doc::Tests](https://docs.google.com/document/d/1HHK_v4OAG4Wz-mgtPdTMijNiece5JoJTz0EONXadZt4/edit?tab=t.0#heading=h.b5jlqdn01mdh)
- [x] New unit tests added.
- [x] Verified backward compatibility (if applicable).

- [x] Yes. Clearly explain the behavior change and its impact.
  - For full table of behaviour changes, see [Design Doc::Solution::Local Solution::Truth Tables of Scenarios & Router Responses](https://docs.google.com/document/d/1HHK_v4OAG4Wz-mgtPdTMijNiece5JoJTz0EONXadZt4/edit?tab=t.0#heading=h.8s6r7ydl1y26)


services/venice-router/src/main/java/com/linkedin/venice/router/api/VeniceVersionFinder.java
```diff
@@ -98,17 +98,22 @@ public int getVersion(String storeName, BasicFullHttpRequest request) throws Ven
       throw new StoreDisabledException(storeName, "read");
     }
 
+    // TODO: put in Map.computeIfAbsent()
```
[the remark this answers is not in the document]
WhitneyDeng: I concur with your sentiments, hence the TODO note.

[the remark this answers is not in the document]
lusong64: Please do not let it slip. Putting todo here is nothing as no one would take a look. It is an invite for another hard to troubleshoot concurrency bug. And why we removed the lastCurrentVersionMap.put?

lusong64 (changes requested):

services/venice-router/src/main/java/com/linkedin/venice/router/api/VeniceVersionFinder.java
```diff
@@ -90,146 +92,104 @@ public int getVersion(String storeName, BasicFullHttpRequest request) throws Ven
...
-    String errorMessage = "Unable to serve new active version: " + kafkaTopic + ".";
-    if (!currentVersionPartitionResourcesReady) {
-      errorMessage += " Partition resources not ready for new active version.";
-      stats.recordStalenessReason(StaleVersionReason.OFFLINE_PARTITIONS);
-    }
+      return existingVersion;
+    }
+
+    // version swap: new version
+    int servingVersion = lastCurrentVersionMap.computeIfPresent(storeName, (k, v) -> {
```
FelixGV, line 129: This right here is not safe... `computeIfPresent` will return null if no mapping currently exists for the provided key.

The `lastCurrentVersionMap` is a `Map<String, Integer>` which means `computeIfPresent` will return `Integer` values. An `Integer` can be cast to an `int` only if it is not null, otherwise you'll get a NPE.

services/venice-router/src/main/java/com/linkedin/venice/router/api/VeniceVersionFinder.java
```diff
@@ -77,8 +79,8 @@ public int getVersion(String storeName, BasicFullHttpRequest request) throws Ven
      * TODO: clone a store object is too expensive, and we could choose to expose the necessary methods
      * in {@link ReadOnlyStoreRepository}, such as 'isEnableReads' and others.
      */
-    Store store = metadataRepository.getStore(storeName);
-    if (store == null || store.isMigrating()) {
+    AtomicReference<Store> store = new AtomicReference<>(metadataRepository.getStore(storeName));
```
FelixGV, line 82: Why use `AtomicReference`? If the point is just to access it from within the `computeIfPresent` closure, then it is likely cheaper to leave it as a unwrapped `Store store` variable on the stack, and then instantiate a `Store finalStore = store` prior to the map function call, then use the latter one within the closure.

services/venice-router/src/main/java/com/linkedin/venice/router/api/VeniceVersionFinder.java
```diff
@@ -90,146 +92,104 @@ public int getVersion(String storeName, BasicFullHttpRequest request) throws Ven
...
+      String existingVersionKafkaTopic = Version.composeKafkaTopic(storeName, existingVersion);
+      boolean existingVersionStatusOnline = existingVersionStatus.equals(VersionStatus.ONLINE);
+      boolean existingVersionPartitionResourcesReady = isPartitionResourcesReady(existingVersionKafkaTopic);
+      boolean existingVersionDecompressorReady = isDecompressorReady(store.get(), existingVersion);
+      if (existingVersionStatusOnline && existingVersionPartitionResourcesReady && existingVersionDecompressorReady) {
+        // existing version ready to serve
+        String message = errorMessage + " Continuing to serve existing version: " + existingVersion + ".";
+        if (!EXCEPTION_FILTER.isRedundantException(message)) {
+          LOGGER.warn(message);
+        }
```
FelixGV, line 167: This seems like a pretty common case, and you already have a metric recorded anyway, so let's not log here.

WhitneyDeng: Thank you for providing the bigger picture context on version swap. Nixed the "Continuing to serve existing version" line. 

Want to learn more: does the common case refer to `no version swap, thus serve existing version` or specifically `new version exists but not ready to serve, thus serve existing version`?

services/venice-router/src/main/java/com/linkedin/venice/router/api/VeniceVersionFinder.java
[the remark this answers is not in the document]
WhitneyDeng: Implemented `computeIfAbsent` in this commit https://github.com/linkedin/venice/pull/1620/commits/afe7714cddf01e46cd7c67fb8bccf7c371e19191

> why we removed the lastCurrentVersionMap.put
After getting context from @majisourav , we decided to not put the `(storeName, versionNumber)` map entry before validating that `versionNumber != Store.NON_EXISTING_VERSION` as the purpose of `lastCurrentVersionMap` is to record the most recent serving version. `Store.NON_EXISTING_VERSION` can not be a serving version.

services/venice-router/src/main/java/com/linkedin/venice/router/api/VeniceVersionFinder.java
WhitneyDeng: Thanks for helping me understand the the tradeoff of `AtomicReference` vs a new variable.

I tried to implement `Store finalStore = store` but my IDE complains and insists on using a final array `Store[]`...

services/venice-router/src/main/java/com/linkedin/venice/router/api/VeniceVersionFinder.java
WhitneyDeng: TIL. Thank you. 

I nixed `computeIfPresent` for leanness and refactored the mapping function into the parent method `getVersion` main logic.

But theoretically, IIUC, we'd want to type `Integer servingVersion` to hedge against NPE.

services/venice-router/src/main/java/com/linkedin/venice/router/api/VeniceVersionFinder.java
majisourav99: Its the latter one, but ideally if we manage the dictionary download during SOP, that case where compressor is not ready in router should be extremely rare.

majisourav99: But IMO, having the log to track if moving dictionary download early helped is fine, and later we can remove it altogether.

services/venice-router/src/main/java/com/linkedin/venice/router/api/VeniceVersionFinder.java
WhitneyDeng: replaced `computerIfAbsent` with `putIfAbsent` to balance both leanness & atomicity in this commit 7b5c9f3f4d8f246909b0f67510be9a814226b7ba

services/venice-router/src/main/java/com/linkedin/venice/router/api/VeniceVersionFinder.java
```diff
@@ -100,86 +98,85 @@ public int getVersion(String storeName, BasicFullHttpRequest request) throws Ven
...
+    errorMessage += " Unable to serve existing version: " + existingVersion + ".";
+    if (!existingVersionStatusOnline) {
+      errorMessage += " Previous version has status: " + existingVersionStatus + ".";
+    }
+    if (!existingVersionDecompressorReady) {
+      errorMessage += " Decompressor not ready for previous version (Has dictionary downloaded?).";
+    }
+    errorMessage += " No version ready to serve.";
+    if (!EXCEPTION_FILTER.isRedundantException(errorMessage)) {
+      LOGGER.warn(errorMessage);
```
FelixGV, line 172: I think we should remove all logging from this function... We already have metrics, no need for logging as well. If we do absolutely need logging, then let's flip things around a bit... perform the redundant exception check ASAP on the smallest possible string (e.g., just `"dictionary" + storeName`), and only if it says to log, then concatenate that giant string and log it. Otherwise we're assembling this big string with a ton of stuff in it and just throwing it away...

services/venice-router/src/main/java/com/linkedin/venice/router/api/DictionaryRetrievalService.java
```diff
@@ -119,18 +119,16 @@ public void handleStoreChanged(Store store) {
...
-      // For versions that went into non ONLINE states, delete dictionary.
+      // For versions that went into nonONLINE states, delete dictionary.
       versions.stream()
           .filter(
-              version -> version.getCompressionStrategy() == CompressionStrategy.ZSTD_WITH_DICT
-                  && version.getStatus() != VersionStatus.ONLINE)
+              version -> version.getCompressionStrategy() == CompressionStrategy.ZSTD_WITH_DICT && version
+                  .getStatus() != VersionStatus.ONLINE /*TODO: do not delete future versions. either status==STARTED or check version number > current*/)
           .forEach(
               version -> handleVersionRetirement(version.kafkaTopicName(), "Version status " + version.getStatus()));
```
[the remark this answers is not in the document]
majisourav99: +1, we need not have this block as we discussed before.

services/venice-router/src/main/java/com/linkedin/venice/router/api/VeniceVersionFinder.java
```diff
@@ -44,7 +42,7 @@ public class VeniceVersionFinder {
   private final ReadOnlyStoreConfigRepository storeConfigRepo;
   private final Map<String, String> clusterToD2Map;
   private final String clusterName;
-  private final ConcurrentMap<String, Integer> lastCurrentVersionMap = new ConcurrentHashMap<>();
+  private final VeniceConcurrentHashMap<String, Integer> lastCurrentVersionMap = new VeniceConcurrentHashMap<>();
```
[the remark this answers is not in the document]
lusong64: Second here. Usually we use interface to declare the variable and assign it with concrete class. But I would argue to keep ConcurrentMap instead of Map. 

Your call.

services/venice-router/src/main/java/com/linkedin/venice/router/api/VeniceVersionFinder.java
```diff
@@ -100,86 +98,85 @@ public int getVersion(String storeName, BasicFullHttpRequest request) throws Ven
...
+        lastCurrentVersionMap.putIfAbsent(storeName, metadataCurrentVersion);
+      } else {
+        // new store not ready to serve
+        return Store.NON_EXISTING_VERSION;
+      }
     }
-    int lastCurrentVersion = lastCurrentVersionMap.get(storeName);
-    if (lastCurrentVersion == metadataCurrentVersion) {
+
+    int existingVersion = lastCurrentVersionMap.get(storeName);
```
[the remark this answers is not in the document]
lusong64: It was my idea to use: "if (null != concurrentMap.putIfAbsent(....))" to leverage the return Value being null or not to detect whether the Integer exists before. 

My argument is that, if we don't need a concurrentMap here, we should totally change the variable and use it in a non-thread safe way. That said, since we declare the map being a concurrentMap since 2022, we got to have a reason, unless we are absolutely sure that this block is protected in its caller.

services/venice-router/src/main/java/com/linkedin/venice/router/api/DictionaryRetrievalService.java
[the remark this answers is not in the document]
WhitneyDeng: Will do. Completed the short term solution to fail requests where dictionary is not ready. Refining the long term solution to get future version dictionary.

services/venice-router/src/main/java/com/linkedin/venice/router/api/DictionaryRetrievalService.java
```diff
@@ -119,18 +119,16 @@ public void handleStoreChanged(Store store) {
...
-              .filter(
-                  version -> version.getCompressionStrategy() == CompressionStrategy.ZSTD_WITH_DICT
-                      && version.getStatus() == VersionStatus.ONLINE)
+              .filter(version -> version.getCompressionStrategy() == CompressionStrategy.ZSTD_WITH_DICT)
               .filter(version -> !downloadingDictionaryFutures.containsKey(version.kafkaTopicName()))
               .map(Version::kafkaTopicName)
               .collect(Collectors.toList()));
 
-      // For versions that went into non ONLINE states, delete dictionary.
+      // For versions that went into nonONLINE states, delete dictionary.
```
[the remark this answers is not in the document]
WhitneyDeng: oops good catch. a65bcead26fd69bc21861158c111c705ff1e35a2

services/venice-router/src/main/java/com/linkedin/venice/router/api/VeniceVersionFinder.java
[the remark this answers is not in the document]
WhitneyDeng: agreed. done 6e02011ba71d69522ef0ec317998d75be0f87848

services/venice-router/src/main/java/com/linkedin/venice/router/api/VeniceVersionFinder.java
```diff
@@ -100,86 +98,85 @@ public int getVersion(String storeName, BasicFullHttpRequest request) throws Ven
...
-    boolean currentVersionPartitionResourcesReady = isPartitionResourcesReady(kafkaTopic);
-    if (currentVersionPartitionResourcesReady && currentVersionDecompressorReady) {
-      // all partitions are online and decompressor is initialized with dictionary
-      lastCurrentVersionMap.put(storeName, newCurrentVersion);
-      stats.recordNotStale();
-      return newCurrentVersion;
+    // version swap: new version
+    String newVersionKafkaTopic = Version.composeKafkaTopic(storeName, metadataCurrentVersion);
+    boolean newVersionPartitionResourcesReady = isPartitionResourcesReady(newVersionKafkaTopic);
+    boolean newVersionDecompressorReady = isDecompressorReady(store, metadataCurrentVersion);
```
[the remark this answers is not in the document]
WhitneyDeng: Good shout. Done. 1258b4fe60f84730e70bca9b7a964f19bcbfbfac

services/venice-router/src/main/java/com/linkedin/venice/router/api/VeniceVersionFinder.java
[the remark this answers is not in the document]
WhitneyDeng: Neat solution, Felix. Done. e0e32c9c7c04f5a5aa167a3253ac25969568c34b

services/venice-router/src/main/java/com/linkedin/venice/router/api/VeniceVersionFinder.java
WhitneyDeng: log discussion continued here https://github.com/linkedin/venice/pull/1620#discussion_r2033864172

services/venice-router/src/main/java/com/linkedin/venice/router/api/DictionaryRetrievalService.java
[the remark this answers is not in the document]
WhitneyDeng: nixed it.

services/venice-router/src/main/java/com/linkedin/venice/router/api/DictionaryRetrievalService.java
```diff
@@ -142,6 +117,20 @@ public void handleStoreChanged(Store store) {
...
     }
+
+    // download dictionary of future & current store versions
+    private void queueDictionaryDownloadOfFutureAndCurrentVersions(Store store) {
+      dictionaryDownloadCandidates.addAll(
+          store.getVersions()
+              .stream()
+              .filter(
+                  version -> version.getCompressionStrategy() == CompressionStrategy.ZSTD_WITH_DICT
+                      && (version.getStatus() == VersionStatus.STARTED || version.getStatus() == VersionStatus.PUSHED
```
majisourav99, line 128: dont think we need to handle PUSHED status here, which is used only in parent controller.

services/venice-router/src/main/java/com/linkedin/venice/router/api/DictionaryRetrievalService.java
```diff
@@ -247,6 +237,8 @@ private CompletableFuture<byte[]> getDictionary(String store, int version) {
               "Dictionary download for resource: " + kafkaTopic + " from: " + instanceUrl
                   + " returned unexpected response.");
         } else {
+          long elapsedTime = System.currentTimeMillis() - startTime;
```
majisourav99, line 240: nit: may use `getElapsedTimeFromMsToMs`

services/venice-router/src/main/java/com/linkedin/venice/router/api/DictionaryRetrievalService.java
```diff
@@ -333,7 +325,7 @@ private boolean getAllDictionaries() {
    * @param dictionaryDownloadTopics A Collection of topics (representing store and version) to download the dictionaries for.
    * @return false if the dictionary download timed out, true otherwise.
    */
-  private boolean downloadDictionaries(Collection<String> dictionaryDownloadTopics) {
+  protected boolean downloadDictionaries(Collection<String> dictionaryDownloadTopics) {
```
majisourav99, line 328: why this change?

services/venice-router/src/main/java/com/linkedin/venice/router/api/VeniceVersionFinder.java
```diff
@@ -38,13 +37,14 @@ public class VeniceVersionFinder {
...
       RedundantExceptionFilter.getRedundantExceptionFilter();
+  private static final String UNINITALISED_KAFKA_TOPIC_STRING = "uninit-kafka-topic";
 
   private final ReadOnlyStoreRepository metadataRepository;
   private final StaleVersionStats stats;
   private final ReadOnlyStoreConfigRepository storeConfigRepo;
   private final Map<String, String> clusterToD2Map;
   private final String clusterName;
-  private final ConcurrentMap<String, Integer> lastCurrentVersionMap = new ConcurrentHashMap<>();
+  private final ConcurrentMap<String, Integer> lastCurrentVersionMap = new VeniceConcurrentHashMap<>();
```
majisourav99, line 47: nit: can we use `Map<>` lastCurrentVersionMap = new VeniceConcurrentHashMap...

majisourav99: actually its fine. I see you are using `ConcurrentMap`

services/venice-router/src/main/java/com/linkedin/venice/router/api/VeniceVersionFinder.java
```diff
@@ -38,13 +37,14 @@ public class VeniceVersionFinder {
   private static final Logger LOGGER = LogManager.getLogger(VeniceVersionFinder.class);
   private static final RedundantExceptionFilter EXCEPTION_FILTER =
       RedundantExceptionFilter.getRedundantExceptionFilter();
+  private static final String UNINITALISED_KAFKA_TOPIC_STRING = "uninit-kafka-topic";
```
majisourav99, line 40: can we just use empty or null string instead of this ?

services/venice-router/src/main/java/com/linkedin/venice/router/api/VeniceVersionFinder.java
```diff
@@ -99,87 +99,96 @@ public int getVersion(String storeName, BasicFullHttpRequest request) throws Ven
...
+    if (existingVersion == null) {
       if (metadataCurrentVersion == Store.NON_EXISTING_VERSION) {
-        /** This should happen at most once per store, since we are adding the mapping to {@link lastCurrentVersionMap} */
+        /** new store push has completed but the routers are not up-to-date on the latest metadata yet
+         * This should happen at most once per store, since we are adding the mapping to {@link lastCurrentVersion} */
         store = metadataRepository.refreshOneStore(storeName);
         metadataCurrentVersion = store.getCurrentVersion();
       }
+      // check if new store is ready to serve
+      newVersionKafkaTopic = Version.composeKafkaTopic(storeName, metadataCurrentVersion);
```
majisourav99, line 115: are we guaranteed here that metaDataCurrentVersion is not `NON_EXISTING_VERSION` or need to put these in a if check ?

services/venice-router/src/main/java/com/linkedin/venice/router/api/DictionaryRetrievalService.java
WhitneyDeng: done ecb2e721f54152d2a26f32374a8606306f447da9

services/venice-router/src/main/java/com/linkedin/venice/router/api/DictionaryRetrievalService.java
WhitneyDeng: hm must be IDE misbehaving. thanks for spotting this, sourav. 5d0575ec35e22f37269b202abbb501bf6a28a509

services/venice-router/src/main/java/com/linkedin/venice/router/api/DictionaryRetrievalService.java
WhitneyDeng: good call on making the variable name more descriptive, IIUC. change made: 2a462ef4da48bb70db56649e906824974c720ec6

services/venice-router/src/main/java/com/linkedin/venice/router/api/VeniceVersionFinder.java
WhitneyDeng: done fcd4a09ee87da69b69514480ec0827de9f1e5ae5

services/venice-router/src/main/java/com/linkedin/venice/router/api/VeniceVersionFinder.java
WhitneyDeng: A good sanity check that I've also marinated on myself, Sourav. Allow me to break down and explicate my thoughts.

# Setup:
- scenario where `metadataCurrentVersion` = `NON_EXISTING_VERSION` after refreshing the store: the new store version is not ready

# Why I decided against an if check:
1. original code logic doesn't check for `NON_EXISTING_VERSION` either. It even `stats.recordNotStale()` and returns `NON_EXISTING_VERSION`
2. since `NON_EXISTING_VERSION = 0`, the checks for the serving readiness of `NON_EXISTING_VERSION` would fail anyways. 

# Performance consideration
I can see why we want to check for and fail `NON_EXISTING_VERSION` after refreshing store early for performance since this is on the hot path. The performance tradeoff: if statement for `NON_EXISTING_VERSION` _versus_ let store readiness checks fail

services/venice-router/src/main/java/com/linkedin/venice/router/api/VeniceVersionFinder.java
WhitneyDeng: After discussing offline, compromised on `LOGGER.debug`. changes made here https://github.com/linkedin/venice/pull/1620/commits/531142987177e27cfdf14f3c3f2951012486c613

Can note down follow up PR to nix these logs once this change is fully stabilised and rolled out.

services/venice-router/src/main/java/com/linkedin/venice/router/api/DictionaryRetrievalService.java
```diff
@@ -416,6 +411,7 @@ private CompletableFuture<Void> fetchCompressionDictionary(Version version) {
             }
             return null;
           }, executor);
+      System.out.println("put dictionary of this topic to downloadingDictionaryFutures: " + kafkaTopic);
```
majisourav99, line 414: remove it?

services/venice-router/src/main/java/com/linkedin/venice/router/api/DictionaryRetrievalService.java
```diff
@@ -247,6 +236,11 @@ private CompletableFuture<byte[]> getDictionary(String store, int version) {
               "Dictionary download for resource: " + kafkaTopic + " from: " + instanceUrl
                   + " returned unexpected response.");
         } else {
+          long elapsedTimeForDictionaryRetrievalMs = System.currentTimeMillis() - startTime;
```
majisourav99, line 239: nit: use LatencyUtils::getElapsedTimeFromMsToMs ?

services/venice-router/src/main/java/com/linkedin/venice/router/api/VeniceVersionFinder.java
```diff
@@ -99,87 +99,90 @@ public int getVersion(String storeName, BasicFullHttpRequest request) throws Ven
     }
 
     int metadataCurrentVersion = store.getCurrentVersion();
-    if (!lastCurrentVersionMap.containsKey(storeName)) {
-      lastCurrentVersionMap.put(storeName, metadataCurrentVersion);
+    String newVersionKafkaTopic = UNINITALISED_KAFKA_TOPIC_STRING;
```
majisourav99, line 102: Can we use empty string check instead?

services/venice-router/src/main/java/com/linkedin/venice/router/api/VeniceVersionFinder.java
```diff
@@ -99,87 +99,90 @@ public int getVersion(String storeName, BasicFullHttpRequest request) throws Ven
...
+      return metadataCurrentVersion;
     }
 
-    String errorMessage = "Unable to serve new active version: " + kafkaTopic + ".";
-    if (!currentVersionPartitionResourcesReady) {
-      errorMessage += " Partition resources not ready for new active version.";
+    // new version not ready to serve
+    LOGGER.debug("Unable to serve new version: {}", newVersionKafkaTopic);
+    if (!newVersionPartitionResourcesReady) {
+      LOGGER.debug("Partition resources not ready for new version");
```
majisourav99, line 151: add the topic name to the log, logs with fixed string are generally useless.

services/venice-router/src/main/java/com/linkedin/venice/router/api/VeniceVersionFinder.java
```diff
@@ -99,87 +99,90 @@ public int getVersion(String storeName, BasicFullHttpRequest request) throws Ven
...
+    if (!existingVersionStatusOnline) {
+      LOGGER.debug("Existing version has status: {}", existingVersionStatus);
+    }
+    if (!existingVersionDecompressorReady) {
+      LOGGER.debug("Decompressor not ready for existing version (Has dictionary downloaded?)");
     }
+
+    stats.recordStale(metadataCurrentVersion, Store.NON_EXISTING_VERSION);
+    lastCurrentVersionMap.put(storeName, Store.NON_EXISTING_VERSION);
+    return Store.NON_EXISTING_VERSION;
```
majisourav99, line 182: is it better to return `existingVersion` instead?

services/venice-router/src/main/java/com/linkedin/venice/router/api/VeniceVersionFinder.java
FelixGV, line 151: So... I have doubts whether this bunch of debug logs are useful... Prior to this commit, there was one giant `errorMessage` getting concatenated with a bunch of stuff, and it was logged just once towards the end. Now this has been split into a bunch of separate logger calls, some of which carry no variables at all (e.g., the one highlighted by this comment...). As someone who turns on the debug logs, do I get any meaningful benefit from seeing just `Partition resources not ready for new version` without knowing what store-version (or any other info) it related to? Also, even if we did add the relevant info to each and every debug log, we would still need to piece together a ~half-dozen separate statements to understand what is going on. That doesn't seem sufficiently helpful...

So anyway... here are a few options to consider:

1. Just delete the damn logs 😅 ... we already have metrics, so that could be sufficient.
2. If we must have the logs, then do at most one logging call with all of the relevant info, while still paying the cost for it only if the logging level is enabled. For example:

```
String errorMessage = null;
if (LOGGER.isDebugEnabled) {
  errorMessage = "Unable to serve new version: " + newVersionKafkaTopic + ".";
}

  if (!currentVersionPartitionResourcesReady) {
    if (LOGGER.isDebugEnabled) {
      errorMessage += " Partition resources not ready for new active version.";
    }
    stats.recordStalenessReason(StaleVersionReason.OFFLINE_PARTITIONS);
  }

// ... and so on every time you need to pack a bit more info...

if (LOGGER.isDebugEnabled) {
  // N.B.: At this point, you've already paid the price to concatenate the message, so you might as well pass it in the redundant filter as well...
  if (!EXCEPTION_FILTER.isRedundantException(errorMessage)) {
    LOGGER.debug(message);
  }
}
```

FelixGV: It should also be possible to achieve option 2 with a single `isDebugEnabled` check at the end of the function... Just gather all the info you need there at the last moment, rather than accumulating as you go.

majisourav99: Yes, I am not saying we scatter the logs all over, just concat the strings and log at the end.

services/venice-router/src/main/java/com/linkedin/venice/router/api/VeniceVersionFinder.java
WhitneyDeng: discussed offline. This is the desired behaviour change: to return `NON_EXISTING_VERSION` when no version is ready to serve, rather than to return the not-ready-to-serve `existingVersion`

services/venice-router/src/main/java/com/linkedin/venice/router/api/VeniceVersionFinder.java
```diff
@@ -99,87 +99,90 @@ public int getVersion(String storeName, BasicFullHttpRequest request) throws Ven
...
+    if (existingVersion == null) {
       if (metadataCurrentVersion == Store.NON_EXISTING_VERSION) {
-        /** This should happen at most once per store, since we are adding the mapping to {@link lastCurrentVersionMap} */
+        /** new store push has completed but the routers are not up-to-date on the latest metadata yet
+         * This should happen at most once per store, since we are adding the mapping to {@link lastCurrentVersion} */
         store = metadataRepository.refreshOneStore(storeName);
         metadataCurrentVersion = store.getCurrentVersion();
       }
+      // check if new store is ready to serve
+      lastCurrentVersionMap.putIfAbsent(storeName, metadataCurrentVersion);
```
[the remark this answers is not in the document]
WhitneyDeng: `putIfAbsent` to prevent race condition.

services/venice-router/src/main/java/com/linkedin/venice/router/api/VeniceVersionFinder.java
WhitneyDeng: aligned synced offline

services/venice-router/src/main/java/com/linkedin/venice/router/api/DictionaryRetrievalService.java
WhitneyDeng: good catch, thanks. nixed it https://github.com/linkedin/venice/pull/1620/commits/230d5e14763ef680a6632b2f744ba7ee99c78ed1

services/venice-router/src/main/java/com/linkedin/venice/router/api/DictionaryRetrievalService.java
WhitneyDeng: TIL! Thanks for suggesting this. done. 51e9c07fa99f02a6b169e96ee7be01a7ac9472b8

services/venice-router/src/main/java/com/linkedin/venice/router/api/VeniceVersionFinder.java
WhitneyDeng: Thanks for the suggestion. I gave it some thought:
- today, `INITIALISED_KAFKA_TOPIC_STRING` is an empty string, which would work with empty string `newVersionKafkaTopic.isEmpty()` check. However, there's an off chance that `INITIALISED_KAFKA_TOPIC_STRING` might be modified in future (perhaps a unique string). Therefore, it seems to me that it's safer to directly check/compare `newVersionKafkaTopic` against `UNINITALISED_KAFKA_TOPIC_STRING`.

services/venice-router/src/main/java/com/linkedin/venice/router/api/VeniceVersionFinder.java
WhitneyDeng: 1. add `LOGGER.isDebugEnabled()` at string concatenation/info gathering stage
2. increase `RedundantExceptionFilter::noRedundantExceptionDurationMs` from 1 min to 5 mins.

WhitneyDeng: continued in https://github.com/linkedin/venice/pull/1620#discussion_r2047402161

services/venice-router/src/main/java/com/linkedin/venice/router/api/DictionaryRetrievalService.java
```diff
@@ -424,7 +419,8 @@ private CompletableFuture<Void> fetchCompressionDictionary(Version version) {
 
   private void initCompressorFromDictionary(Version version, byte[] dictionary) {
     String kafkaTopic = version.kafkaTopicName();
-    if (version.getStatus() != VersionStatus.ONLINE || !downloadingDictionaryFutures.containsKey(kafkaTopic)) {
+    if ((version.getStatus() != VersionStatus.ONLINE && version.getStatus() != VersionStatus.STARTED)
```
majisourav99, line 422: can you refactor this logic to a method and use it everywhere else?

WhitneyDeng: good shout. done 2762e5609f41bc3797eb4098fe61912e598cb523

services/venice-router/src/main/java/com/linkedin/venice/router/api/DictionaryRetrievalService.java
```diff
@@ -448,6 +443,10 @@ private void handleVersionRetirement(String kafkaTopic, String exceptionReason)
     compressorFactory.removeVersionSpecificCompressor(kafkaTopic);
   }
 
+  private boolean isCurrentOrFutureVersion(Version version) {
```
majisourav99, line 446: why not everything into `shouldDownloadDictionaryForVersion` including this check `version.getCompressionStrategy() == CompressionStrategy.ZSTD_WITH_DICT`

WhitneyDeng: good call. done. 
renaming: 9b90c2b7aa32a26aaa63aa80e3b0623c77aa80bd 
include compression strategy check: 5e6dcd3016cd39d3f8b542b72777a412bb337c71

services/venice-router/src/main/java/com/linkedin/venice/router/api/VeniceVersionFinder.java
```diff
@@ -191,26 +206,25 @@ private boolean isPartitionResourcesReady(String kafkaTopic) {
...
-          LOGGER.warn("Failed to get partition assignment for resource: {}", kafkaTopic, e);
+          LOGGER.debug("Failed to get partition assignment for resource: {}", kafkaTopic, e);
           partitionAssignment = "unknown";
         }
-        String message = "No online replica exists for partition " + partitionId + " of " + kafkaTopic
-            + ", partition assignment: " + partitionAssignment;
-        if (!EXCEPTION_FILTER.isRedundantException(message)) {
-          LOGGER.warn(message);
-        }
+        LOGGER.debug(
```
majisourav99, line 212: make it back to warn with log limiter. this is helpful log

services/venice-router/src/main/java/com/linkedin/venice/router/api/VeniceVersionFinder.java
```diff
@@ -98,88 +99,102 @@ public int getVersion(String storeName, BasicFullHttpRequest request) throws Ven
...
+        isDecompressorReady(existingVersion, Version.composeKafkaTopic(storeName, existingVersionNumber));
+    if (existingVersionStatusOnline && existingVersionDecompressorReady) {
+      // existing version ready to serve
+      if (LOGGER.isDebugEnabled()) {
+        String errorMessage = "Unable to serve new version: " + newVersionKafkaTopic + ".";
+        errorMessage += newVersionPartitionResourcesReady ? "" : " Partition resources not ready for new version.";
+        errorMessage += newVersionDecompressorReady ? "" : " Decompressor not ready for new version.";
+        errorMessage += " Continuing to serve existing version: " + existingVersionNumber + ".";
+        if (!EXCEPTION_FILTER.isRedundantException(errorMessage)) {
+          LOGGER.debug(errorMessage);
```
majisourav99, line 171: this one too, make it warn, as you have 5mins rate limiting already and they will help with debugging.

services/venice-router/src/main/java/com/linkedin/venice/router/api/VeniceVersionFinder.java
```diff
@@ -98,88 +99,102 @@ public int getVersion(String storeName, BasicFullHttpRequest request) throws Ven
...
+      if (!EXCEPTION_FILTER.isRedundantException(errorMessage)) {
+        LOGGER.debug(errorMessage);
       }
-      lastCurrentVersionMap.put(storeName, newCurrentVersion);
-      stats.recordNotStale();
-      return newCurrentVersion;
     }
+    stats.recordStale(metadataCurrentVersionNumber, Store.NON_EXISTING_VERSION);
+    lastCurrentVersionMap.put(storeName, Store.NON_EXISTING_VERSION);
+    return Store.NON_EXISTING_VERSION;
```
majisourav99, line 194: Why are we diverging from previous behavior of 

       * When the router has only one available version, despite offline partitions, or dictionary not yet downloaded,
       * it will return it as the available version.
       * If the partitions are still unavailable or the dictionary is not downloaded by the time the records needs to
       * be decompressed, then the router will return an error response.

WhitneyDeng: Thank you for your question. The rationale is, if the partitions are offline, the version is not ready to serve, the router should report that version as not ready to serve and fail fast rather than kick the can down the road.

services/venice-router/src/main/java/com/linkedin/venice/router/api/VeniceVersionFinder.java
WhitneyDeng: done 2ff32d7f0af3bac473453d4176444d086f64acbb

services/venice-router/src/main/java/com/linkedin/venice/router/api/VeniceVersionFinder.java
WhitneyDeng: done 9f6c0c7b332bbbfc30ba03c82981d2ce3a63577a

services/venice-router/src/main/java/com/linkedin/venice/router/api/VeniceVersionFinder.java
WhitneyDeng: Discussed offline. Conclusion: if both new and existing version are not ready, serve the existing version + log store as stale. Reason: serving something is better than nothing. eaf8a7c53290abde6715757f4f63377f5fbf15c8

services/venice-router/src/main/java/com/linkedin/venice/router/api/VeniceVersionFinder.java
```diff
@@ -98,119 +98,122 @@ public int getVersion(String storeName, BasicFullHttpRequest request) throws Ven
...
-       * be decompressed, then the router will return an error response.
-       */
-      String message = errorMessage + " Switching to serve new active version.";
-      if (!EXCEPTION_FILTER.isRedundantException(message)) {
-        LOGGER.warn(message);
-      }
-      lastCurrentVersionMap.put(storeName, newCurrentVersion);
-      stats.recordNotStale();
-      return newCurrentVersion;
+    // check if existing version ready to serve
```
majisourav99, line 156: nit: we are just logging, not really checking the existing version for RTS right?

services/venice-router/src/main/java/com/linkedin/venice/router/api/VeniceVersionFinder.java
```diff
@@ -98,119 +98,122 @@ public int getVersion(String storeName, BasicFullHttpRequest request) throws Ven
...
+    Version existingVersion = store.getVersion(existingVersionNumber);
+    boolean existingVersionDecompressorReady =
+        isDecompressorReady(existingVersion, Version.composeKafkaTopic(storeName, existingVersionNumber));
+    // existing version ready to serve
+    if (!EXCEPTION_FILTER.isRedundantException(storeName)) {
+      LOGGER.warn(
+          "Unable to serve new version: {}." + " Partition resources ready for new version? {}."
+              + " Decompressor not ready for new version? {}." + " Continuing to serve existing version: {}."
+              + " Decompressor ready for existing version? (Has dictionary downloaded?) {}.",
```
majisourav99, line 165: nit: can you remove '(Has dictionary downloaded)` part ?

WhitneyDeng: This was in the original code. I am with you on this. done. d9f510084398f48c2a67dd210260700f3f90390c

services/venice-router/src/main/java/com/linkedin/venice/router/api/VeniceVersionFinder.java
WhitneyDeng: good eye. done. 67b90a99deb2a303fc8aaee27af5f690dd690b03

services/venice-router/src/main/java/com/linkedin/venice/router/api/VeniceVersionFinder.java
```diff
@@ -98,119 +98,122 @@ public int getVersion(String storeName, BasicFullHttpRequest request) throws Ven
...
+            errorMessage += "Failed to get partition assignment for resource: " + kafkaTopic + " " + e;
+          }
           partitionAssignment = "unknown";
         }
-        String message = "No online replica exists for partition " + partitionId + " of " + kafkaTopic
-            + ", partition assignment: " + partitionAssignment;
-        if (!EXCEPTION_FILTER.isRedundantException(message)) {
-          LOGGER.warn(message);
+
+        if (LOGGER.isDebugEnabled()) {
```
majisourav99, line 199: remove the if check.

WhitneyDeng: I concur with @FelixGV's previous [comment and suggestion](https://github.com/linkedin/venice/pull/1620#discussion_r2047402161) and am inclined to keep `LOGGER.isDebugEnabled` check and using `debug()` rather than `warn()`. 

However, in the interest of merging the code changes and testing the new behaviour rather than bike-shedding, this check has been removed. See this commit 8ae9c56d3430b8a29bfff8c9d83642d29594cf8f

@majisourav99's rationale for deleting the check and using `warn`:
1. This particular log will only be emitted in the scenario where both new and existing versions are not ready to serve. According to @majisourav99, this is a rare scenario and thus this log will rarely be emitted.
2. Logs in this class `VeniceVersionFinder` is rate limited to 5 minutes.

services/venice-router/src/main/java/com/linkedin/venice/router/api/VeniceVersionFinder.java
```diff
@@ -98,119 +97,123 @@ public int getVersion(String storeName, BasicFullHttpRequest request) throws Ven
...
+     * it will return it as the available version.
+     * If the partitions are still unavailable or the dictionary is not downloaded by the time the records needs to
+     * be decompressed, then the router will return an error response.
+     */
+    // log if existing version ready to serve
+    Version existingVersion = store.getVersion(existingVersionNumber);
+    boolean existingVersionDecompressorReady =
+        isDecompressorReady(existingVersion, Version.composeKafkaTopic(storeName, existingVersionNumber));
+    // existing version ready to serve
+    if (!EXCEPTION_FILTER.isRedundantException(storeName)) {
```
majisourav99, line 166: nit: make it redundant on isRedundantException(store_version) ? should be a minor as we do not expect store version changes are not that frequent.

majisourav99 (approved):
Left a minor comment, other than that LGTM! Thanks for patiently iterating over the comments.

PR #1621, opened 2025-03-17
[controller][repush] Add controller client path to manually trigger repush from parent controller

ZacAttack:
## Problem

We need a way for the controller to be able to trigger repushes outside of the normal compaction scheduling.

## Solution

This change allows for an operator with a controller client to be able to trigger repush via the controller. Additionally add configs to disable controller scheduled based compactions while still allowing for repushes to be triggered from the controller.

###  Code changes

## How was this PR tested?

- [ ] New unit tests added.


internal/venice-client-common/src/main/java/com/linkedin/venice/controllerapi/RepushJobResponse.java
```diff
@@ -0,0 +1,22 @@
...
+
+  public RepushJobResponse() {
+    // TODO: ControlleResponse types need to have a default constructor which takes no arguments,
+    // we either need to refactor that or make it so this class can populate executionId in a different way
+    this.executionId = "-1";
+  }
+
+  public RepushJobResponse(String executionId) {
+    this.executionId = executionId;
+  }
```
mynameborat, line 17: I believe the pattern used is empty constructor with setters across the board. Hence the need to not support empty constructors. 
While I think setter vs constructor injections are both valid and widely used patterns, any reasons why this needs to take the executionId as part of construction (to pivot from existing pattern)

internal/venice-common/src/main/java/com/linkedin/venice/controllerapi/ControllerClient.java
```diff
@@ -1218,13 +1218,13 @@ public MultiStoreInfoResponse getStoresForCompaction(String clusterName) {
    * @param storeName
    * @return //TODO LC:
    */
-  public ControllerResponse triggerRepush(String storeName, byte[] repushJobDetails) {
+  public RepushJobResponse triggerRepush(String storeName, byte[] repushJobDetails, byte[] data) {
```
mynameborat, line 1221: why introduce additional parameter data? `repushJobDetails` exists for this reason? If we want to keep this consistent, maybe just rename as `data`?

internal/venice-common/src/main/java/com/linkedin/venice/controllerapi/ControllerClient.java
```diff
@@ -1218,13 +1218,13 @@ public MultiStoreInfoResponse getStoresForCompaction(String clusterName) {
...
     QueryParams params = newParams().add(NAME, storeName);
     // TODO repush: Use byte[] to pass parameters instead of QueryParams as it is a post method. see
     // (https://github.com/linkedin/venice/pull/1282#discussion_r1871510627)
     // TODO repush: add params from admin tool for repush: e.g. version, fabric etc.
     // TODO repush: add admin.repush()
-    return request(ControllerRoute.COMPACT_STORE, params, ControllerResponse.class, repushJobDetails);
+    return request(ControllerRoute.COMPACT_STORE, params, RepushJobResponse.class, null);
```
mynameborat, line 1227: Did you intend to 
s/repushJobDetails/data?

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -8068,6 +8068,9 @@ public Map<String, StoreDataAudit> getClusterStaleStores(String clusterName) {
    */
   @Override
   public List<StoreInfo> getStoresForCompaction(String clusterName) {
+    if (compactionManager == null) {
+      throw new VeniceException("CompactionManager is not initialized.");
+    }
```
mynameborat, line 8073: Should we instead assert of `isLogCompactionEnabled` and return `API not supported/enabled`?

Throwing uninitialized could only mean the feature is disabled and returning meaningful response might be helpful to the client.

internal/venice-client-common/src/main/java/com/linkedin/venice/controllerapi/RepushJobResponse.java
ZacAttack: So this class was actually copied over with this other constructor, all I did was move it to a new package so that it was accessible from Controllerclient.  I have no idea why RepushJobResponse(String executionId) exists.  Maybe @WhitneyDeng can tell us?

internal/venice-common/src/main/java/com/linkedin/venice/controllerapi/ControllerClient.java
ZacAttack: oh my bad, this was an oversight.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
ZacAttack: sounds better.

WhitneyDeng (commented):
speedy! some comments.

internal/venice-client-common/src/main/java/com/linkedin/venice/controllerapi/RepushJobResponse.java
```diff
@@ -0,0 +1,22 @@
+package com.linkedin.venice.controllerapi;
+
+/**
+ * Data model of response from a repush job trigger request for a store
+ */
+public class RepushJobResponse extends ControllerResponse {
+  private String executionId;
+
+  public RepushJobResponse() {
+    // TODO: ControlleResponse types need to have a default constructor which takes no arguments,
```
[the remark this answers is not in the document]
ZacAttack: This class today can't have a constructor with arguments because it's instantiated using reflection in the controller.

internal/venice-client-common/src/main/java/com/linkedin/venice/controllerapi/RepushJobResponse.java
WhitneyDeng: Let's go with the empty constructors + setters.

mynameborat (dismissed):
lgtm

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestHybrid.java
```diff
@@ -1096,6 +1082,22 @@ public void testHybridStoreLogCompaction() throws Exception {
       LOGGER.error("Log compaction job failed");
       throw new RuntimeException(e);
     }
+
+    // ok, now do it again manually
```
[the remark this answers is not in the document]
ZacAttack: Yeah, so I'm not really sure how we've instrumented things in this test.  It seems like to do that configuration I'd have to spin up an entirely new cluster to tweak that configuration.  Which we could do, but seems excessive for what's going on here.  If you recall that thread we had with sushant w.r.t. setting up the defaults for the cluster for some of the hybrid testing, it'd probably make sense to configure it by default as you've described it, just to get all the bits moving.

So I have a compromise that I've added here.  The implementation is such that the controllerClient's call is synchronouse with the compactionManager ticking the latch down.  As such, I've removed the code which waited on the latch, instead checking the latch's value immediately after it's returned.  I think this verifies that the next run was "probably" done manually.

mynameborat (dismissed):

internal/venice-test-common/src/integrationTest/java/com/linkedin/venice/endToEnd/TestHybrid.java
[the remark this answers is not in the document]
WhitneyDeng: Yea... this test might be a bit involved to set up. My concern here is that we want to be sure that all the components (CompactionManager, RepushOrchestration implementation) are set up for manual repush if scheduling is turned off. The code changes look good to me, but just on the off chance we missed something.

I trust your judgement and if you're confident, feel free to move forward :)

mynameborat (approved):

WhitneyDeng (approved):
🚀

PR #1623, opened 2025-03-18
[controller] use largestUsedRTVersionNumber in addVersion and updateStore commands

arjun4084346:
## Problem

If child controller fails to add version, there may be inconsistency between local RT version numbers of child controllers

## Solution

In this PR parent controller will pass RT version number in addVersion admin command and child controllers are not going to use the local copy of that.
In https://github.com/linkedin/venice/pull/1616 I introduced a new field `largestUsedRTVersionNumber` and `currentRTVersionNumber` in admin tool commands. In this PR, I am using them.

###  Code changes
- [x] Added new code behind **a config**. If so list the config names and their default values in the PR description.

- [x] Code has **no race conditions** or **thread safety issues**.

## How was this PR tested?

- [ ] New unit tests added.

### DIFF
--- a/internal/venice-common/src/main/java/com/linkedin/venice/meta/AbstractStore.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/meta/AbstractStore.java
@@ -1,5 +1,7 @@
 package com.linkedin.venice.meta;
 
+import static com.linkedin.venice.meta.Version.DEFAULT_RT_VERSION_NUMBER;
+
 import com.linkedin.venice.exceptions.StoreDisabledException;
 import com.linkedin.venice.exceptions.StoreVersionNotFoundException;
 import com.linkedin.venice.exceptions.VeniceException;
@@ -78,7 +80,7 @@ public abstract class AbstractStore implements Store {
 
   @Override
   public void addVersion(Version version) {
-    addVersion(version, true, false, 0);
+    addVersion(version, true, false, DEFAULT_RT_VERSION_NUMBER);
   }
 
   @Override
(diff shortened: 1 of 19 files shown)


internal/venice-common/src/main/java/com/linkedin/venice/meta/AbstractStore.java
```diff
@@ -107,7 +107,11 @@ public void checkDisableStoreWrite(String action, int version) {
    *                        any store level config on it; if false, the version being added is new version, so the new version
    *                        config should be the same as store config.
    */
-  private void addVersion(Version version, boolean checkDisableWrite, boolean isClonedVersion) {
+  private void addVersion(
+      Version version,
+      boolean checkDisableWrite,
+      boolean isClonedVersion,
+      int currentRTVersionNumber) {
```
arjun4084346, line 114: not using `currentRTVersionNumber` in the method yet. That will be the part of future PR#1555
because the logic is closer to that PR's work.
Some changes in this PR are going to make that long PR more reviewable.

services/venice-controller/src/main/java/com/linkedin/venice/controller/VeniceHelixAdmin.java
```diff
@@ -2129,7 +2129,8 @@ public void addVersionAndStartIngestion(
         replicationMetadataVersionId,
         versionSwapDeferred,
         null,
-        repushSourceVersion);
+        repushSourceVersion,
+        getStore(clusterName, storeName).getLargestUsedRTVersionNumber());
```
lluwm, line 2133: The caller of this method `CreateVersion.addVersionAndStartIngestion` is only used in the store migration. I feel that we should double check if it's correct to use the `LargestUsedRTVersionNumber` from the destination here. Can we verify its behavior with an existing or new integration test?

arjun4084346: I changed it use DEFAULT_RT_VERSION_NUMBER that will not change anything in the version and the migrated version just copies whatever is there in the source version.

lluwm (approved):
Thanks @arjun4084346 , LGTM!

PR #1625, opened 2025-03-18
[server] Thread-Safe Segment

KaiSernLim:
## Problem Statement

Resubscription can cause different `ConsumptionTask` threads to read and write to the `KafkaDataIntegrityValidator` and its subsequent `Segment` maps. Therefore, it would be safer if the `Segment` data structure was thread-safe.

### DIFF
--- a/internal/venice-common/src/main/java/com/linkedin/venice/kafka/validation/Segment.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/kafka/validation/Segment.java
@@ -55,11 +55,11 @@ public class Segment {
   // Immutable state
   private final int partition;
   private final int segmentNumber;
-  private final CheckSum checkSum;
   private final Map<CharSequence, CharSequence> debugInfo;
   private final Map<CharSequence, Long> aggregates;
 
   // Mutable state
+  private final CheckSum checkSum;
   private int sequenceNumber;
   private boolean registered;
   private boolean started;
@@ -121,23 +121,27 @@ public class Segment {
   }
 
   public Segment(Segment segment) {
-    this.partition = segment.partition;
-    this.segmentNumber = segment.segmentNumber;
-    this.checkSum = CheckSum.getInstance(segment.getCheckSumType(), segment.getCheckSumState());
-    this.sequenceNumber = segment.sequenceNumber;
+    synchronized (segment) {
+      this.partition = segment.partition;
+      this.segmentNumber = segment.segmentNumber;
+      this.checkSum = CheckSum.getInstance(segment.getCheckSumType(), segment.getCheckSumState());
+      this.sequenceNumber = segment.getSequenceNumber();
 
-    this.started = segment.started;
-    this.ended = segment.ended;
-    this.finalSegment = segment.finalSegment;
-    this.newSegment = false;
-    /**
-     * N.B. No need to call {@link #getDedupedDebugInfo(Map)} here since we assume the other {@link Segment} instance we
-     * are copying from was already deduped, having come from one of the other constructors.
-     */
-    this.debugInfo = segment.debugInfo;
-    this.aggregates = segment.aggregates;
-    this.registered = segment.registered;
-    this.lastRecordProducerTimestamp = segment.lastRecordProducerTimestamp;
+      this.started = segment.isStarted();
+      this.ended = segment.isEnded();
+      this.finalSegment = segment.isFinalSegment();
+      this.newSegment = false;
+      /**
+       * N.B. No need to call {@link #getDedupedDebugInfo(Map)} here since we assume the other {@link Segment} instance we
+       * are copying from was already deduped, having come from one of the other constructors.
+       */
+      this.debugInfo = segment.debugInfo;
+      this.aggregates = segment.aggregates;
+      this.registered = segment.isRegistered();
+      this.lastRecordProducerTimestamp = segment.getLastRecordProducerTimestamp();
+      this.lastSuccessfulOffset = segment.getLastSuccessfulOffset();
+      this.lastRecordTimestamp = segment.getLastRecordTimestamp();
+    }
   }
 
   public int getSegmentNumber() {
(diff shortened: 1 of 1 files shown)


internal/venice-common/src/main/java/com/linkedin/venice/kafka/validation/Segment.java
```diff
@@ -55,11 +55,11 @@ public class Segment {
   // Immutable state
   private final int partition;
   private final int segmentNumber;
-  private final CheckSum checkSum;
   private final Map<CharSequence, CharSequence> debugInfo;
   private final Map<CharSequence, Long> aggregates;
 
   // Mutable state
+  private final CheckSum checkSum;
```
FelixGV, line 62: Why move this here? Because the content of the checksum is mutable, even though the pointer is immutable?

internal/venice-common/src/main/java/com/linkedin/venice/kafka/validation/Segment.java
```diff
@@ -148,20 +152,19 @@ public int getPartition() {
     return this.partition;
   }
 
-  /** N.B. This function is not threadsafe. Locking must be handled by the caller. */
-  public int getAndIncrementSequenceNumber() {
+  public synchronized int getAndIncrementSequenceNumber() {
```
FelixGV, line 155: Starting a comment thread here though it's about the PR as a whole, not this specific line.

I wonder if this is necessary? The segment is a low-level primitive which we expect to manipulate at a very high throughput. It ought to, for all intents and purposes, be used by a single thread. Adding the overhead of synchronization "just in case" is unfortunate.

The PR description states that this is because in some scenarios we can have multiple `ConsumptionTask` manipulating the same `Segment`, and I have a bit of trouble with that.

First of all, what does it mean to have multiple `CT` instances for a given store-version? And even if we allow for the possibility that a `CT` is in the process of closing but is maybe not fully done closing yet and we already have a new `CT` coming up... on a more fine-grained level, what does it mean for two `CT` instances to be operating on the exact same partition?

What I'm getting at is: I wonder whether synchronizing access to the `Segment` for a given partition might be the least of our concern? Even if the `Segment`'s bookkeeping was done in a threadsafe way, can it really be correct to have two `CT` instances operating on the same partition?? Will we end up needing to synchronize a bunch of other things as well?

To me (though I'm not 100% sure, it is just an impression...), this seems like a fix in the wrong place. Why would we allow there to be two concurrently-running `CT` instances for the same store-version in the first place?

FelixGV: And one more thing... even if we came to the conclusion that it's the right thing to do to have concurrent manipulation of `Segment` instances... it would still not be right (IMHO) to synchronize at the function-level.

A given message might update up to 3 parts of the `Segment` state: the segment number, the sequence number, and the checksum. These are 3 pieces of state but in reality their mutation belongs to the processing of the same message. If we allow two messages being processed by concurrent threads for the same partition (which, again, that premise is quite shaky, but let's go with it for now), then it is almost certainly incorrect to let their mutations to these three pieces of state interleave. For example, this does not seem right:

1. T1 takes care of segment number.
2. T1 takes care of segment number.
3. T2 takes care of segment number.
4. T2 takes care of sequence number.
5. T2 takes care of checksum.
6. T1 takes care of checksum.

Will the checksum be right if it's updated out of order? Probably not?

KaiSernLim:
Discussed offline. Will be making the fields `volatile` instead of adding `synchronized`. 🫣

internal/venice-common/src/main/java/com/linkedin/venice/kafka/validation/Segment.java
FelixGV: After chatting offline, I realized that I was mistaken on certain things.

There is NOT just one `CT` / store-version. There is one `SIT` / store-version, and there is one `CT` / consumer instance. The number of consumers is determined by the consumer pool size, and there is one consumer pool per pub sub cluster (both local and remote). So there are a bunch of `CT` instances, but as far as what we are concerned with here, there _can be_ multiple `CT` instances for the same partition, _sort of_. These `CT` instances are for the same partition number of the same topic name, but in different pub sub clusters (i.e., in the case of A/A replication consuming the RT topics of multiple regions).

So it can indeed be the case that multiple `CT` instances interact with the same logical partition, and even do so concurrently, _however_ they should still not interact with the same `Segment` instance, because each of these is tied to a specific producer, and the producer GUID should make them globally unique (barring GUID collisions...).

So all in all, it is still the case that we should not need synchronization, since there should only ever be one `CT` interacting with a given `Segment` at any given time. However, there can be handoffs of partitions across consumer pools (as part of the consumer pool priority work), and therefore, even though there is just one `CT` at a time, it could be the case that two `CT` instances deal with the same `Segment` "not very long after each other". For this reason, it is considered prudent to make the state inside of `Segment` volatile, to avoid the Java Memory Model serving stale data to the new `CT`... this is a pretty unlikely edge case, given the delay required in making the new `CT` resubscribe to the partition, but in any case, it can be ok to add volatile out of an abundance of caution...

lluwm: Felix, thanks for reviewing this PR carefully. During the DIV development work, we realized that consumer pool priority work (re-subscription) can cause multiple `CT` threads to modify the same `Segment` instance, not concurrently, but one `CT` thread at one time and another `CT` thread later. Such behavior might introduce a visibility problem for us and the focus here is to make sure that writes made by one `CT` thread is guaranteed to be  visible to new `CT` thread. Based on the [Java Memory Model](https://www.cs.umd.edu/~pugh/java/memoryModel/jsr133.pdf), to achieve such visibility, the options seems to be either we write to a `volatile` viable  or use a lock. Otherwise, as the doc put, "compilers are allowed to reorder the instructions in either thread, when it does not affect the execution of that thread in isolation." We don't really want compiler to be smart, optimize for us and causes us surprise to debug later. Agree that we can use the `volatile` if we don't want the overhead of synchronization. Thanks again!

PR #1626, opened 2025-03-18
[thin-client] Add config for Avro store client to use custom warmup Executor

bbarkley:
## Problem Statement
Our use of the Venice client makes requests to a service which expects metadata to be present in an invocation context which is stored in a ThreadLocal and set up by our executor services.

## Solution
This change makes it possible to customize the Executor used for the client warmup. If none is provided, the existing behavior is maintained.

###  Code changes

## How was this PR tested?
I don't see related tests for other areas where client customizations are done. Since this is a straightforward change, I don't have tests for it, but am open to suggestions.
- [ ] New unit tests added.

- [x ] No. You can skip the rest of this section.


clients/venice-thin-client/src/main/java/com/linkedin/venice/client/store/AbstractAvroStoreClient.java
```diff
@@ -133,8 +133,11 @@ protected AbstractAvroStoreClient(
     this.compressorFactory = new CompressorFactory();
     this.storageRequestPath = TYPE_STORAGE + "/" + clientConfig.getStoreName();
     this.computeRequestPath = TYPE_COMPUTE + "/" + clientConfig.getStoreName();
-    this.veniceClientWarmUpExecutor = Executors
-        .newFixedThreadPool(1, new DaemonThreadFactory("Venice-Client-Warmup-For-" + clientConfig.getStoreName()));
+    this.veniceClientWarmUpExecutor = Optional.ofNullable(clientConfig.getWarmupExecutor())
```
haoxu07, line 136: Could you helpt o add a unit test to verify this work as expected? Example: `AbstractAvroStoreClientTest`

haoxu07 (commented):
Thanks for preparing this, @bbarkley ! Just feel we need a unit test to see customer executor work as expected.

clients/venice-thin-client/src/main/java/com/linkedin/venice/client/store/AbstractAvroStoreClient.java
bbarkley: Thanks - added a test case to verify the executor is used when provided.

haoxu07 (approved):

PR #1627, opened 2025-03-18
[server] Deprecate OffsetRecord field leaderOffset and its usage to avoid ready-to-serve check using wrong offset; Fix bug that persist VT offset into upstream offset map

sixpluszero:
## [server] Deprecate OffsetRecord field leaderOffset and its usage to avoid ready-to-serve check using wrong offset; Fix bug that persist VT offset into upstream offset map
No we should not use this OffsetRecord#leaderOffset anymore.

**_This PR fixes two issues._**

**The first issue is:**
This is legacy field and we already replaced it with upstreamOffsetMap. However, we are still updating this field from all the regions and our offset fetch logic can fall back can still hit this value in corner case.

Let's say we subscribe to region 1,2,3 when we promote this replica to LEADER role. If one of the consumer failed to subscribe the region (say 3), then it will not consuming any data from that, and in PartitionConsumptionState and OffsetRecord nothing is persisted for this region. 

However, in the RTS check, when we are measuring offset lag for this region, we will eventually fall back to leaderOffset as persisted offset, if all the map in PCS does not contain region offset value. This is **WRONG**, because it is updated by all the regions. So essentially we are doing lag for region 3 = end offset of region 3 - persisted offset of region 1. And if Region 1 has more data and higher offset, this can be < 0 and mark it ready to serve before it is truly ready after consumer repair service kicks in.

**The second issue is:**
In non-AGG bug fix PR: #1547 , there is a regression introduced during refactor - previously we will only checkpoint calculated upstream offset map during TopicSwitch, not during any standby -> leader, but during the refactor we merged two methods, but the checkpoint should not happen unless we are consuming RTs. Actually, we should not checkpoint to OffsetRecord at all in the method, we should only do the in memory update and only sync to OffsetRecord during formal checkpoint time.

The fix is just simply remove the line, and add a unit test to test against the behavior.

## Solution

This PR fixes it by removing the whole concept of this field. We should return -1 for the offset if no offset has checkpointed to this region.

###  Code changes

## How was this PR tested?

- [x] New unit tests added.

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
@@ -1044,8 +1044,6 @@ public class LeaderFollowerStoreIngestionTask extends StoreIngestionTask {
           calculateLeaderUpstreamOffsetWithTopicSwitch(partitionConsumptionState, leaderTopic, Collections.emptyList())
               .getOrDefault(OffsetRecord.NON_AA_REPLICATION_UPSTREAM_OFFSET_MAP_KEY, OffsetRecord.LOWEST_OFFSET);
     }
-    partitionConsumptionState.getOffsetRecord()
-        .setLeaderUpstreamOffset(OffsetRecord.NON_AA_REPLICATION_UPSTREAM_OFFSET_MAP_KEY, upstreamStartOffset);
 
     consumerSubscribe(leaderTopic, partitionConsumptionState, upstreamStartOffset, leaderSourceKafkaURL);
     syncConsumedUpstreamRTOffsetMapIfNeeded(
--- a/internal/venice-common/src/main/java/com/linkedin/venice/offsets/OffsetRecord.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/offsets/OffsetRecord.java
@@ -248,7 +248,7 @@ public class OffsetRecord {
    * call this API to get the latest upstream offset.
    */
   public long getUpstreamOffset(String kafkaURL) {
-    return getUpstreamOffsetFromPartitionState(this.partitionState, kafkaURL);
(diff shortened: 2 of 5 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
```diff
@@ -1044,8 +1044,6 @@ void prepareOffsetCheckpointAndStartConsumptionAsLeader(
           calculateLeaderUpstreamOffsetWithTopicSwitch(partitionConsumptionState, leaderTopic, Collections.emptyList())
               .getOrDefault(OffsetRecord.NON_AA_REPLICATION_UPSTREAM_OFFSET_MAP_KEY, OffsetRecord.LOWEST_OFFSET);
     }
-    partitionConsumptionState.getOffsetRecord()
```
sushantmane, line 1047: I wonder if we should persists offsets we get from `offsetForTime` query

internal/venice-common/src/main/java/com/linkedin/venice/offsets/OffsetRecord.java
```diff
@@ -346,10 +336,6 @@ public String toSimplifiedString() {
   }
 
   private String getPartitionUpstreamOffsetString() {
```
sushantmane, line 338: Can we delete this method? when we concat map with string it should call toString on the map internally

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/LeaderFollowerStoreIngestionTask.java
sushantmane: I didn't see syncConsumedUpstreamRTOffsetMapIfNeeded earlier.

sushantmane (approved):
Thanks, @sixpluszero. Just one minor comment.

PR #1628, opened 2025-03-18
[router] Added metric for rejected connection when limit is breached

huangminchn:
## Problem Statement
Missing a metric for rejected connection, which should be easier for
setting up alerts.

## Solution
A rate metric is added when connections are rejected due to breaching limit.
Metric name: rejected_connection_count

###  Code changes

## How was this PR tested?

- [ ] New unit tests added.
- [x] New integration tests added.


gaojieliu (approved):
Thanks for adding the metric!

PR #1632, opened 2025-03-19
[controller] Add leader discovery for parent helix admin for updateAdminOperationProtocolVersion API

minhmo1620:
[controller] Add leader discovery for parent helix admin

## Problem Statement
- Currently, `updateAdminOperationProtocolVersion` API will not fail if the request is sent to non-leader, which can cause the racing issue if Helix has a state transfer in the middle of the request.
- `updateAdminOperationProtocolVersion` for child controller is not necessary since we never try to consume the config from child controller.

## Solution
- Add leader discovery for parent helix admin for updateAdminOperationProtocolVersion API to allow us to fail if the current parent controller is not leader.
- Removing support for `updateAdminOperationProtocolVersion` in child controller
- Convert all execeptions in AdminOperationSerializer to `VeniceProtocolException`

###  Code changes

## How was this PR tested?

- [ ] New unit tests added.


huangminchn (approved):
Thanks Minh!

PR #1637, opened 2025-03-20
[server][da-vinci] Centralized store ingestion task clean up service

huangminchn:
## Problem Statement
Currently both state transition threads and store ingestion threads can close store ingestion tasks;
as a result, race conditions could happen in the below order:
SIT thread decides to close itself
-> ST thread starts subscribing to new partitions
-> new partition subscription cleans and ST returns successfully
-> SIT finishes closing itself

## Solution
This PR fixes the issue by introducing a centralized clean up service inside KafkaStoreIngestionService which reuses the topic level lock manager in it. Besides, SIT will not be able to close itself anymore (unless it's isolated ingestion process); ST threads and the clean-up service will control the life cycle of SIT instances.

New config to enable clean-up version: "server.idle.ingestion.task.cleanup.interval.in.seconds"
-1: feature disabled.
non-negative values: clean-up service enabled, config value is the clean-up delay interval.

###  Code changes
- [x] Added new code behind **a config**. If so list the config names and their default values in the PR description.
- [x] Introduced new **log lines**. 

## How was this PR tested?

- [x] New unit tests added.

### DIFF
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaStoreIngestionService.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaStoreIngestionService.java
@@ -110,6 +110,7 @@ import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.ConcurrentSkipListMap;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.locks.ReentrantLock;
@@ -200,6 +201,8 @@ public class KafkaStoreIngestionService extends AbstractVeniceService implements
   private final Lazy<ZKHelixAdmin> zkHelixAdmin;
 
   private final ExecutorService aaWCIngestionStorageLookupThreadPool;
+
+  private final ScheduledExecutorService idleStoreIngestionTaskKillerExecutor;
 
   public KafkaStoreIngestionService(
       StorageService storageService,
@@ -450,6 +453,10 @@ public class KafkaStoreIngestionService extends AbstractVeniceService implements
       this.aaWCWorkLoadProcessingThreadPool = null;
     }
 
+    this.idleStoreIngestionTaskKillerExecutor = serverConfig.getIdleIngestionTaskCleanupIntervalInSeconds() >= 0
+        ? Executors.newScheduledThreadPool(1, new DaemonThreadFactory("idle-store-ingestion-task-clean-up-thread"))
+        : null;
+
     this.aaWCIngestionStorageLookupThreadPool = Executors.newFixedThreadPool(
         serverConfig.getAaWCIngestionStorageLookupThreadPoolSize(),
         new DaemonThreadFactory("AA_WC_INGESTION_STORAGE_LOOKUP"));
@@ -535,6 +542,14 @@ public class KafkaStoreIngestionService extends AbstractVeniceService implements
       participantStoreConsumerExecutorService =
           Executors.newSingleThreadExecutor(new DaemonThreadFactory("ParticipantStoreConsumptionTask"));
       participantStoreConsumerExecutorService.submit(participantStoreConsumptionTask);
+    }
+    final int idleIngestionTaskCleanupIntervalInSeconds = serverConfig.getIdleIngestionTaskCleanupIntervalInSeconds();
+    if (idleIngestionTaskCleanupIntervalInSeconds >= 0) {
+      this.idleStoreIngestionTaskKillerExecutor.scheduleWithFixedDelay(
+          this::scanAndCloseIdleConsumptionTasks,
+          idleIngestionTaskCleanupIntervalInSeconds,
+          idleIngestionTaskCleanupIntervalInSeconds,
+          TimeUnit.SECONDS);
     }
     // Although the StoreConsumptionTasks are now running in their own threads, there is no async
     // process that needs to finish before the KafkaStoreIngestionService can be considered
@@ -625,7 +640,7 @@ public class KafkaStoreIngestionService extends AbstractVeniceService implements
     Utils.closeQuietlyWithErrorLogged(ingestionThrottler);
     Utils.closeQuietlyWithErrorLogged(participantStoreConsumptionTask);
     shutdownExecutorService(participantStoreConsumerExecutorService, "participantStoreConsumerExecutorService", true);
-
+    shutdownExecutorService(idleStoreIngestionTaskKillerExecutor, "idleStoreIngestionTaskKillerExecutor", true);
     /*
      * We would like to gracefully shutdown {@link #ingestionExecutorService},
      * so that it will have an opportunity to checkpoint the processed offset.
@@ -958,11 +973,26 @@ public class KafkaStoreIngestionService extends AbstractVeniceService implements
       LOGGER.info("Reset consumption offset for topic: {}, partition: {}", topicName, partitionId);
     }
     if (!ingestionTaskHasAnySubscription(topicName)) {
-      if (isIsolatedIngestion) {
-        LOGGER.info("Ingestion task for topic {} will be kept open for the access from main process.", topicName);
-      } else {
-        LOGGER.info("Shutting down ingestion task of topic {}", topicName);
-        shutdownStoreIngestionTask(topicName);
+      shutdownIdleIngestionTask(topicName);
+    }
+  }
+
+  /**
+   * A helper function which checks the idles counter inside StoreIngestionTask; if the counter is higher than the idle
+   * threshold, treat the task as idle and shutdown the task.
+   */
+  private void shutdownIdleIngestionTask(String topicName) {
+    try (AutoCloseableLock ignore = topicLockManager.getLockForResource(topicName)) {
+      // Must check the idle status again after acquiring the lock.
+      StoreIngestionTask ingestionTask = topicNameToIngestionTaskMap.get(topicName);
+      if (ingestionTask != null
+          && !(ingestionTask.consumerHasAnySubscription() || ingestionTask.hasAnyPendingSubscription())) {
+        if (isIsolatedIngestion) {
+          LOGGER.info("Ingestion task for topic {} will be kept open for the access from main process.", topicName);
+        } else {
+          LOGGER.info("Shutting down ingestion task of topic {}", topicName);
+          shutdownStoreIngestionTask(topicName);
+        }
       }
     }
   }
@@ -1018,6 +1048,23 @@ public class KafkaStoreIngestionService extends AbstractVeniceService implements
     LOGGER.info("Start killing the following ingestion tasks: {}", nonCurrentVersions);
     nonCurrentVersions.forEach(topic -> killConsumptionTask(topic));
     LOGGER.info("Finished killing the following ingestion tasks: {}", nonCurrentVersions);
+  }
+
+  private void scanAndCloseIdleConsumptionTasks() {
+    try {
+      LOGGER.info("Number of ingestion tasks before cleaning: {}", topicNameToIngestionTaskMap.size());
+      for (Map.Entry<String, StoreIngestionTask> entry: topicNameToIngestionTaskMap.entrySet()) {
+        String topicName = entry.getKey();
+        StoreIngestionTask task = entry.getValue();
+        if (task.isIdleOverThreshold()) {
+          LOGGER.info("Found idle task for topic {}, shutting it down.", topicName);
+          shutdownIdleIngestionTask(topicName);
+        }
+      }
+      LOGGER.info("Number of active ingestion tasks after cleaning: {}", topicNameToIngestionTaskMap.size());
+    } catch (VeniceException e) {
+      LOGGER.info("Error when attempting to shutdown idle store ingestion tasks", e);
+    }
   }
 
   /**
@@ -1322,7 +1369,8 @@ public class KafkaStoreIngestionService extends AbstractVeniceService implements
   private boolean ingestionTaskHasAnySubscription(String topic) {
     try (AutoCloseableLock ignore = topicLockManager.getLockForResource(topic)) {
       StoreIngestionTask consumerTask = topicNameToIngestionTaskMap.get(topic);
-      return consumerTask != null && consumerTask.hasAnySubscription();
+      return consumerTask != null
+          && (consumerTask.consumerHasAnySubscription() || consumerTask.hasAnyPendingSubscription());
     }
   }
 
--- a/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
+++ b/clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
@@ -754,6 +754,20 @@ public abstract class StoreIngestionTask implements Runnable, Closeable {
   }
 
   /**
+   * Check the consumer action queue to see if there are any pending SUBSCRIBE actions.
+   */
+  public boolean hasAnyPendingSubscription() {
+    return consumerActionsQueue.stream().anyMatch(action -> action.getType() == SUBSCRIBE);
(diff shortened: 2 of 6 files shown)


clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -1902,7 +1916,9 @@ private void internalClose(boolean doFlush) {
     // Only reset Offset and Drop Partition Messages are important, subscribe/unsubscribe will be handled
     // on the restart by Helix Controller notifications on the new StoreIngestionTask.
     this.storeRepository.unregisterStoreDataChangedListener(this.storageUtilizationManager);
-    for (ConsumerAction message: consumerActionsQueue) {
+    // Remove all the actions out of the queue in order
+    while (!consumerActionsQueue.isEmpty()) {
+      ConsumerAction message = consumerActionsQueue.poll();
```
lluwm, line 1921: why we change to remove actions from queue?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -1555,7 +1569,7 @@ protected void checkIngestionProgress(Store store) throws InterruptedException {
      * Check whether current consumer has any subscription or not since 'poll' function will throw
      * {@link IllegalStateException} with empty subscription.
      */
-    if (!consumerHasAnySubscription()) {
+    if (!consumerHasAnySubscription() && !hasAnyPendingSubscription()) {
```
lluwm, line 1572: Looks like We have several places to use these two conditions together, how about we extract them out and put it into a simple helper function?

`consumerHasAnySubscription() || hasAnyPendingSubscription()`

lluwm (commented):
Looks good to me overall. Minor comments.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
huangminchn: In theory, it doesn't matter. Since the queue only exists in the SIT instance, if SIT is closed, no one can access to the queue anyway.
IIRC, a test case related to kill job failed, the test tries to verify that there is no active subscription after killing the push, but since we also check subscription action in the queue, the check would fail.

kvargha (changes requested):

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
huangminchn: Updated in the latest diff.
However, I need to do another refactoring after talking with @majisourav99 .
We can't rely on consumer subscriptions and pending consumer actions to conclude if the SIT is idle or not, but instead, check PartitionConsumptionStatusMap and pending consumer actions. Context:
If unsubscriptionAfterBatchPushComplete is enabled, consumers would be unsubscribed from the VT, but no changes to PCS map on purpose, because we want to keep SIT thread running in this case to process actions like LEADER_TO_STANDBY transition or kill job action.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaStoreIngestionService.java
```diff
@@ -1020,6 +1049,23 @@ private void killConsumptionTaskForNonCurrentVersions() {
     LOGGER.info("Finished killing the following ingestion tasks: {}", nonCurrentVersions);
   }
 
+  private void scanAndCloseIdleConsumptionTasks() {
+    try {
+      LOGGER.info("Number of ingestion tasks before cleaning: {}", topicNameToIngestionTaskMap.size());
+      for (Map.Entry<String, StoreIngestionTask> entry: topicNameToIngestionTaskMap.entrySet()) {
```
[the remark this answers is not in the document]
huangminchn: This is done on purpose, the first check will not synchronize, once the first decision is made, inside "shutdownIdleIngestionTask(String topicName)", we will take the lock and check if the task is still idle again.

Or are you referring to something else?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -755,7 +755,21 @@ private void dropPartitionSynchronously(PubSubTopicPartition topicPartition) {
...
   public boolean hasAnySubscription() {
-    return !partitionConsumptionStateMap.isEmpty();
+    return consumerHasAnySubscription() || hasAnyPendingSubscription();
+  }
+
+  /**
+   * Check the consumer action queue to see if there are any pending SUBSCRIBE actions.
+   */
+  public boolean hasAnyPendingSubscription() {
+    return consumerActionsQueue.stream().anyMatch(action -> action.getType() == SUBSCRIBE);
```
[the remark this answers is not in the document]
huangminchn: There is a config to control scanning interval, I am planning to set it to at least 10 minutes in prod. `PriorityBlockingQueue` is from the "java.util.concurrent" package, I believe it's thread-safe.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaStoreIngestionService.java
```diff
@@ -536,6 +543,14 @@ public boolean startInner() {
           Executors.newSingleThreadExecutor(new DaemonThreadFactory("ParticipantStoreConsumptionTask"));
       participantStoreConsumerExecutorService.submit(participantStoreConsumptionTask);
     }
+    final int idleIngestionTaskCleanupIntervalInSeconds = serverConfig.getIdleIngestionTaskCleanupIntervalInSeconds();
+    if (idleIngestionTaskCleanupIntervalInSeconds > 0) {
```
majisourav99, line 547: nit: idleStoreIngestionTaskKillerExecutor != null check seems cleaner ? ie

```
if (idleStoreIngestionTaskKillerExecutor != null) {
 this.idleStoreIngestionTaskKillerExecutor.scheduleWithFixedDelay(
          this::scanAndCloseIdleConsumptionTasks,
          idleIngestionTaskCleanupIntervalInSeconds,
          idleIngestionTaskCleanupIntervalInSeconds,
          TimeUnit.SECONDS);
    }
```

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaStoreIngestionService.java
```diff
@@ -1020,6 +1049,23 @@ private void killConsumptionTaskForNonCurrentVersions() {
...
+      for (Map.Entry<String, StoreIngestionTask> entry: topicNameToIngestionTaskMap.entrySet()) {
+        String topicName = entry.getKey();
+        StoreIngestionTask task = entry.getValue();
+        if (task.isIdleOverThreshold()) {
+          LOGGER.info("Found idle task for topic {}, shutting it down.", topicName);
+          shutdownIdleIngestionTask(topicName);
+        }
+      }
+      LOGGER.info("Number of active ingestion tasks after cleaning: {}", topicNameToIngestionTaskMap.size());
+    } catch (VeniceException e) {
```
majisourav99, line 1064: can we catch Exception here? Not a good idea to catch sub exception in a background thread.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -755,7 +755,21 @@ private void dropPartitionSynchronously(PubSubTopicPartition topicPartition) {
...
+  }
+
+  /**
+   * This helper function will check if the ingestion task has been idle for a long time.
+   */
+  public boolean isIdleOverThreshold() {
```
majisourav99, line 771: So we reset this counter once it reaches maxIdleCounter. when it will be > maxIdleCounter ?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaStoreIngestionService.java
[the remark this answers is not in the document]
kvargha: No. That makes sense.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -1589,7 +1603,7 @@ protected void checkIngestionProgress(Store store) throws InterruptedException {
       }
       return;
     }
-    idleCounter = 0;
+    idleCounter.set(0);
```
majisourav99, line 1606: nit: maybe call `resetIdleCounter()`

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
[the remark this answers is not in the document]
kvargha: I'm worried that this will result in periodic CPU spikes. Could we instead maintain a counter of pending subscribes that haven't been processed from the `consumerActionsQueue`?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -4192,6 +4208,10 @@ private void deserializeValue(int schemaId, ByteBuffer value, DefaultPubSubMessa
   }
 
   private void maybeCloseInactiveIngestionTask() {
+    if (serverConfig.getIdleIngestionTaskCleanupIntervalInSeconds() > 0 && !isIsolatedIngestion) {
```
majisourav99, line 4211: So previously we did close II SITs?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaStoreIngestionService.java
huangminchn: Yeah, good to be consistent with other service.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaStoreIngestionService.java
huangminchn: Good call.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
huangminchn: There is a "return" at the end of the condition block that handle idle SIT. counter is not reset to 0. Otherwise, the new tests wouldn't work anyway.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
huangminchn: Removed this function instead since there is only one caller. It doesn't seem necessary to switch function context for this update.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
huangminchn: Currently we did.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
```diff
@@ -1915,23 +1928,31 @@ private void internalClose(boolean doFlush) {
...
+            storageMetadataService.clearOffset(topic, partition);
+            message.getFuture().complete(null);
+            break;
+          case DROP_PARTITION:
+            PubSubTopicPartition topicPartition = message.getTopicPartition();
+            LOGGER.info("Processing DROP_PARTITION message for {} in internalClose", topicPartition);
+            dropPartitionSynchronously(topicPartition);
+            message.getFuture().complete(null);
+            break;
+          case SUBSCRIBE:
```
kvargha, line 1951: Do we ever hit this condition? With the updated logic, SIT shouldn't be closed if there's pending SUBSCRIBEs right?

huangminchn: No, kill job command could close SIT too. Server shutdown would also close the SIT.

kvargha (approved):
Approving to unblock, since my review comments were addressed.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaStoreIngestionService.java
```diff
@@ -1020,6 +1051,23 @@ private void killConsumptionTaskForNonCurrentVersions() {
...
+      LOGGER.info("Number of ingestion tasks before cleaning: {}", topicNameToIngestionTaskMap.size());
+      for (Map.Entry<String, StoreIngestionTask> entry: topicNameToIngestionTaskMap.entrySet()) {
+        String topicName = entry.getKey();
+        StoreIngestionTask task = entry.getValue();
+        if (task.isIdleOverThreshold()) {
+          LOGGER.info("Found idle task for topic {}, shutting it down.", topicName);
+          shutdownIdleIngestionTask(topicName);
+        }
+      }
+      LOGGER.info("Number of active ingestion tasks after cleaning: {}", topicNameToIngestionTaskMap.size());
```
majisourav99, line 1065: nit: can we print only if we clean any SIT, depending on how frequent you plan to run it can generate unnecessary excessive logging.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaStoreIngestionService.java
```diff
@@ -1042,12 +1090,13 @@ public boolean killConsumptionTask(String topicName) {
```
majisourav99, line 1090: should we add a try-catch around entire kill method?

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaStoreIngestionService.java
huangminchn: Going to run it at most once every 10 minutes.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/KafkaStoreIngestionService.java
huangminchn: Not aware of any existing issues. No need to be paranoid about unknown issues IMHO.

clients/da-vinci-client/src/main/java/com/linkedin/davinci/kafka/consumer/StoreIngestionTask.java
huangminchn: Checked with both @lluwm and @majisourav99 offline. We can't unify the check condition so far.
1. The cleanup service must check the PCS map instead of consumer subscription states to make decision of cleaning SIT
2. SIT must check consumer subscription state to increase idle counter so that "skipAfterBatchPushUnsubEnabled" short-circuit flag can be set.
Currently there is a bug that skipAfterBatchPushUnsubEnabled will never be set back to False after rebalance happens. Potentially we can unify the check by moving skipAfterBatchPushUnsubEnabled flag update outside of the idle counter maintenance logic after fixing the bug.

PR #1638, opened 2025-03-20
[server] Remote extra layer of retry on getLatestOffsetWithRetriesAsync

majisourav99:
## Problem Statement

In method `getTopicPartitionEndOffSet` we have retries builtin. But the method it calls has another layer of retries leading to retry storm. 

## Solution

This PR removes the lower layer of retries in getTopicPartitionEndOffSet using `getLatestOffsetNoRetry`.

###  Code changes

- [x] Code has **no race conditions** or **thread safety issues**.
- [x] Proper **synchronization mechanisms** (e.g., `synchronized`, `RWLock`) are used where needed.
- [x] No **blocking calls** inside critical sections that could lead to deadlocks or performance degradation.
- [x] Verified **thread-safe collections** are used (e.g., `ConcurrentHashMap`, `CopyOnWriteArrayList`).
- [x] Validated proper exception handling in multi-threaded code to avoid silent thread termination.

## How was this PR tested?

- [ ] New unit tests added.

### DIFF
--- a/internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicMetadataFetcher.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicMetadataFetcher.java
@@ -373,6 +373,13 @@ class TopicMetadataFetcher implements Closeable {
     }, retries, INITIAL_RETRY_DELAY, Duration.ofSeconds(5), Duration.ofMinutes(5), PUBSUB_RETRIABLE_FAILURES);
   }
 
+  CompletableFuture<Long> getLatestOffsetNoRetry(PubSubTopicPartition pubSubTopicPartition) {
+    return CompletableFuture.supplyAsync(() -> {
+      validateTopicPartition(pubSubTopicPartition);
+      return getLatestOffset(pubSubTopicPartition);
+    }, threadPoolExecutor);
+  }
+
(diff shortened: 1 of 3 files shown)


internal/venice-common/src/main/java/com/linkedin/venice/pubsub/manager/TopicMetadataFetcher.java
```diff
@@ -373,6 +373,13 @@ long getLatestOffsetWithRetries(PubSubTopicPartition pubSubTopicPartition, int r
     }, retries, INITIAL_RETRY_DELAY, Duration.ofSeconds(5), Duration.ofMinutes(5), PUBSUB_RETRIABLE_FAILURES);
   }
 
+  CompletableFuture<Long> getLatestOffsetNoRetry(PubSubTopicPartition pubSubTopicPartition) {
+    return CompletableFuture.supplyAsync(() -> {
```
sushantmane, line 377: Can we use `getLatestOffsetWithRetries` with 0 retries?

majisourav99: you mean 1 attempt. I just wanted to remove the log algother.

sushantmane (approved):

sushantmane:
Thanks @majisourav99 !

PR #1639, opened 2025-03-20
[controller] Trim controller logs for DVC instance push status

majisourav99:
## Problem Statement

Excessive logging for stores with many DVC instances ingestion.

## Solution

Skip such instance level logs.

###  Code changes

## How was this PR tested?

- [ ] New unit tests added.

### DIFF
--- a/internal/venice-common/src/main/java/com/linkedin/venice/pushstatushelper/PushStatusStoreReader.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/pushstatushelper/PushStatusStoreReader.java
@@ -101,7 +101,9 @@ public class PushStatusStoreReader implements Closeable {
       if (pushStatusValue == null) {
         return Collections.emptyMap();
       } else {
-        LOGGER.info(" {}/{} Instance status: {}", storeName, partitionId, pushStatusValue.instances);
+        if (pushStatusValue.instances.size() < 50) {
+          LOGGER.info(" {}/{} Instance status: {}", storeName, partitionId, pushStatusValue.instances);
+        }
         return pushStatusValue.instances;
       }
     } catch (Exception e) {
--- a/internal/venice-common/src/main/java/com/linkedin/venice/pushstatushelper/PushStatusStoreReader.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/pushstatushelper/PushStatusStoreReader.java
@@ -101,9 +101,6 @@ public class PushStatusStoreReader implements Closeable {
       if (pushStatusValue == null) {
         return Collections.emptyMap();
       } else {
-        if (pushStatusValue.instances.size() < 20) {
-          LOGGER.info(" {}_v{}/{} Instance status: {}", storeName, version, partitionId, pushStatusValue.instances);
-        }
         return pushStatusValue.instances;
       }
     } catch (Exception e) {
--- a/internal/venice-common/src/main/java/com/linkedin/venice/pushstatushelper/PushStatusStoreReader.java
+++ b/internal/venice-common/src/main/java/com/linkedin/venice/pushstatushelper/PushStatusStoreReader.java
@@ -101,8 +101,8 @@ public class PushStatusStoreReader implements Closeable {
       if (pushStatusValue == null) {
         return Collections.emptyMap();
       } else {
-        if (pushStatusValue.instances.size() < 50) {
-          LOGGER.info(" {}/{} Instance status: {}", storeName, partitionId, pushStatusValue.instances);
+        if (pushStatusValue.instances.size() < 20) {
+          LOGGER.info(" {}_v{}/{} Instance status: {}", storeName, version, partitionId, pushStatusValue.instances);
         }
         return pushStatusValue.instances;
       }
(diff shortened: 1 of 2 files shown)


internal/venice-common/src/main/java/com/linkedin/venice/pushstatushelper/PushStatusStoreReader.java
```diff
@@ -101,7 +101,9 @@ public Map<CharSequence, Integer> getPartitionStatus(
       if (pushStatusValue == null) {
         return Collections.emptyMap();
       } else {
-        LOGGER.info(" {}/{} Instance status: {}", storeName, partitionId, pushStatusValue.instances);
+        if (pushStatusValue.instances.size() < 50) {
+          LOGGER.info(" {}/{} Instance status: {}", storeName, partitionId, pushStatusValue.instances);
```
sushantmane, line 105: Can we  log store version as well? Using

```java
LOGGER.info(" {}_v{}-{} Instance status: {}", storeName, version, partitionId, pushStatusValue.instances);
```

services/venice-controller/src/main/java/com/linkedin/venice/pushmonitor/PushMonitorUtils.java
```diff
@@ -285,11 +285,14 @@ public static ExecutionStatusWithDetails getDaVinciPartitionLevelPushStatusAndDe
         if (instanceStatus.equals(PushStatusStoreReader.InstanceStatus.BOOTSTRAPPING)) {
           // Don't count bootstrapping instance status report.
           totalReplicaCount--;
-          LOGGER.info(
-              "Skipping ingestion status report from bootstrapping node: {} for topic: {}, partition: {}",
-              entry.getKey().toString(),
-              topicName,
-              partitionId);
+          skipLogCount++;
```
sushantmane, line 288: Why do we log only 10 entries?

internal/venice-common/src/main/java/com/linkedin/venice/pushstatushelper/PushStatusStoreReader.java
gaojieliu: Can we log the first 50 `incomplete` instances?
`completed` instances are normally not useful for debugging, and we mostly care about `incomplete` instances.

gaojieliu: Also it is good to provide a summary about the instance status, such as the count for different instance status and maybe log several instances for the `incomplete` status.
@sixpluszero 
Do you have any idea about how to summarize the DVC push status from the debugging POV?

services/venice-controller/src/main/java/com/linkedin/venice/pushmonitor/PushMonitorUtils.java
gaojieliu: What don't we use `RedundantExceptionFilter` here?
Maybe we should rename it to something like `LogRateLimiter`, since it can be used for the regular logging as well.

majisourav99: Sure will use RedundantExceptionFilter

internal/venice-common/src/main/java/com/linkedin/venice/pushstatushelper/PushStatusStoreReader.java
majisourav99: I think the return statuses are logged in `getDaVinciPartitionLevelPushStatusAndDetails` so it should be fine to skip logging it here.

internal/venice-common/src/main/java/com/linkedin/venice/pushstatushelper/PushStatusStoreReader.java
```diff
@@ -101,7 +101,9 @@ public Map<CharSequence, Integer> getPartitionStatus(
       if (pushStatusValue == null) {
         return Collections.emptyMap();
       } else {
-        LOGGER.info(" {}/{} Instance status: {}", storeName, partitionId, pushStatusValue.instances);
+        if (pushStatusValue.instances.size() < 20) {
```
sushantmane, line 104: IMHO, rather than logging this only when there are fewer than 20 instances, we should log all unique DVC client hostnames where ingestion is still incomplete. In most cases, this is far more useful for troubleshooting than just listing the completed partitions.

gaojieliu (approved):

services/venice-controller/src/main/java/com/linkedin/venice/pushmonitor/PushMonitorUtils.java
```diff
@@ -378,9 +379,10 @@ public static ExecutionStatusWithDetails getDaVinciPartitionLevelPushStatusAndDe
           .append(totalReplicaCount);
     }
     int incompleteSize = incompletePartition.size();
-    if (incompleteSize > 0 && incompleteSize <= 5) {
-      statusDetailStringBuilder.append(". Following partitions still not complete ")
-          .append(incompletePartition)
+    if (incompleteSize > 0) {
+      List<Integer> list = new ArrayList<>(incompletePartition);
+      statusDetailStringBuilder.append(". Following partitions still not complete (capped at 10) ")
```
m-nagarajan, line 384: We also need to change this for `OfflinePushStatus#updateStatusDetails`

